PR #1580, opened 2023-04-25
Fix plugin compilation

GumpacG:
### Description
Fixed the build due to upstream refactors (https://github.com/opensearch-project/OpenSearch/pull/7301). 
Gradle version update to accommodate for https://github.com/opensearch-project/OpenSearch/pull/7286 and referred to
https://github.com/FasterXML/jackson-core/issues/955#issuecomment-1480411209 and https://github.com/opensearch-project/sql/pull/1584
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1584, opened 2023-04-26
Update Gradle Wrapper to 7.6.1

dbwiddis:
### Description

Gradle versions before 7.6 are incompatible with JDK 19 and later.

Some GitHub action backend code using jackson-core 2.5.0 is failing builds because it includes an MRJAR with JDK 19 code.  See, for example, [this failed action](https://github.com/opensearch-project/sql/actions/runs/4801615420/jobs/8544063968#step:5:48).

There are other incompatibilities in the build for Gradle 8, so 7.6.1 is the only option to keep a working build.

This commit is the result of executing `./gradlew wrapper`.
 
### Issues Resolved
N/A
 
### Check List
- ~[ ] New functionality includes testing.~
  - ~[ ] All tests pass, including unit test, integration test and doctest~
- ~[ ] New functionality has been documented.~
  - ~[ ] New functionality has javadoc added~
  - ~[ ] New functionality has user manual doc added~


MaxKsyunz:
Thank you @dbwiddis !

@GumpacG has a PR (#1580 ) that updates Gradle and fixes build issues due to refactor in OpenSearch. It makes sense to combine the two.

dbwiddis:
> @GumpacG has a PR (#1580 ) that updates Gradle and fixes build issues due to refactor in OpenSearch. It makes sense to combine the two.

Yep, I didn't see that when I started the process, so it looks like we were both doing it at the same time.

For reference/clarity, simply updating the version in the properties (as #1580 does) is sufficient to get around the bug.  Actually executing `gradle wrapper` also updates some of the supporting script/batch files and provides a local jar that eliminates the need to download it from gradle every time, saving a few seconds off build time.

MaxKsyunz:
Closing this PR as this change is already in main.

PR #1600, opened 2023-05-02
Merge OpenSearchPagedIndexScan and OpenSearchIndexScan

MaxKsyunz:
# High-level Summary
## Core Module
- Added `PushDownPageSize` optimization rule to `LogicalPlanOptimizer`.
- Added `TableScanBuilder.pushDownPageSize` method.
- Removed `Table.createPagedScanBuilder`.
- Removed `CreatePagingTableScanBuilder`.

## opensearch Module
- Simplified `OpenSearchIndexScanQueryBuilder` and `OpenSearchIndexScanAggregationBuilder` by removing index scan dependency.
- `OpenSearchScrollRequest.toCursor` does not generate a cursor if the request needs to be cleaned up -- means OpenSearch will not send any more data.
- Added `ExecutableRequestBuilder` interface that represents operations of an executable request. Used by `OpenSearchIndexScan`.
- Simplified `ContinuePageRequest` to only implement `ExecutableRequestBuilder`.
- Removed
  -  `InitialPageRequestBuilder`, 
  - `OpenSearchPagedIndexScan`,
  -  `OpenSearchPagedIndexScanBuilder`,
  - `PagedRequestBuilder`,
  - `PushDownRequestBuilder`

# Issues Resolved
- Review comments on pagination PR.
 
# Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScanQueryBuilder.java
```diff
@@ -22,41 +22,30 @@
...
-import org.opensearch.sql.storage.TableScanOperator;
-import org.opensearch.sql.storage.read.TableScanBuilder;
 
 /**
  * Index scan builder for simple non-aggregate query used by
  * {@link OpenSearchIndexScanBuilder} internally.
  */
 @VisibleForTesting
-class OpenSearchIndexScanQueryBuilder extends TableScanBuilder {
+@EqualsAndHashCode
```
Yury-Fridlyand, line 42:
```suggestion
@EqualsAndHashCode(callSuper = false)
```

Yury-Fridlyand (commented):
I have two concerns there:
1. Last empty page is not issued anymore. If it is expected, you have to update docs and ... demo.
2. `PushDownPageSize` is performed after all `PUSH_DOWN_*` operations, especially after `PUSH_DOWN_LIMIT`. I think it should be performed before.

docs/user/optimization/optimization.rst
```diff
@@ -287,7 +287,7 @@ The Aggregation operator will merge into OpenSearch Aggregation::
           {
             "name": "OpenSearchIndexScan",
             "description": {
-              "request": "OpenSearchQueryRequest(indexName=accounts, sourceBuilder={\"from\":0,\"size\":0,\"timeout\":\"1m\",\"aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"sources\":[{\"gender\":{\"terms\":{\"field\":\"gender.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}}]},\"aggregations\":{\"avg(age)\":{\"avg\":{\"field\":\"age\"}}}}}}, searchDone=false)"
+              "request": "OpenSearchQueryRequest(indexName=accounts, sourceBuilder={\"from\":0,\"size\":200,\"timeout\":\"1m\",\"aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"sources\":[{\"gender\":{\"terms\":{\"field\":\"gender.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}}]},\"aggregations\":{\"avg(age)\":{\"avg\":{\"field\":\"age\"}}}}}}, searchDone=false)"
```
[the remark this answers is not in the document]
MaxKsyunz: Refactor fixed a bug in explain -- the size of results is always limited but explain call was not showing it.

You can confirm this by running `select * from opensearch_dashboards_sample_data_logs group by ip` in OpenSearch [playground](https://playground.opensearch.org/app/opensearch-query-workbench):

There are 1000 unique `ip` values. Explain for that query is 

```json
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[count(distinct ip)]"
    },
    "children": [
      {
        "description": {
          "request": "OpenSearchQueryRequest(indexName=opensearch_dashboards_sample_data_logs, sourceBuilder={\"from\":0,\"size\":0,\"timeout\":\"1m\",\"aggregations\":{\"count(distinct ip)\":{\"cardinality\":{\"field\":\"ip\"}}}}, searchDone=false)"
        },
        "children": []
      }
    ]
  }
}
```

yet only 200 rows are returned.

integ-test/src/test/resources/expectedOutput/ppl/explain_filter_agg_push.json
```diff
@@ -8,7 +8,7 @@
...
+          "request": "OpenSearchQueryRequest(indexName\u003dopensearch-sql_test_index_account, sourceBuilder\u003d{\"from\":0,\"size\":10000,\"timeout\":\"1m\",\"query\":{\"range\":{\"age\":{\"from\":30,\"to\":null,\"include_lower\":false,\"include_upper\":true,\"boost\":1.0}}},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}],\"aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"sources\":[{\"state\":{\"terms\":{\"field\":\"state.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}},{\"city\":{\"terms\":{\"field\":\"city.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}}]},\"aggregations\":{\"avg_age\":{\"avg\":{\"field\":\"age\"}}}}}}, searchDone\u003dfalse)"
```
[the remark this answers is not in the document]
MaxKsyunz: [It's the same explain bug](https://github.com/opensearch-project/sql/pull/1600#discussion_r1187930632). Previous value was incorrect.

core/src/test/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizerTest.java
```diff
@@ -85,15 +69,15 @@ void setUp() {
   @Test
   void filter_merge_filter() {
     assertEquals(
-        filter(
+        LogicalPlanDSL.filter(
```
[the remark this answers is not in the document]
MaxKsyunz: Must've been done by robots. I'll revert.

MaxKsyunz:
> I have two concerns there:
> 
> 1. Last empty page is not issued anymore. If it is expected, you have to update docs and ... demo.

Why is it needed?  As I understand, the last empty page was benign --  an artifact of how `OpenSearchPagedIndexScan` managed OpenSearch requests but not required by any client and not sent by  legacy pagination. 

If you could point me to the materials that need updating, I'm happy to update them as well. 

> 2. `PushDownPageSize` is performed after all `PUSH_DOWN_*` operations, especially after `PUSH_DOWN_LIMIT`. I think it should be performed before.

Why do you think so? This PR sets up `OpenSearchIndexScan` to handle both limit and pagination. They can be applied in any order.

Assuming it's necessary, how would you implement it?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScanQueryBuilder.java
MaxKsyunz: `false` is default for `callSuper`. Setting `callSuper` is necessary when a class extends another. This is not the case here.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -44,104 +42,102 @@
...
 import org.opensearch.sql.opensearch.data.value.OpenSearchExprValueFactory;
 import org.opensearch.sql.opensearch.response.agg.OpenSearchAggregationResponseParser;
-import org.opensearch.sql.planner.logical.LogicalNested;
+
 
 /**
  * OpenSearch search request builder.
  */
-@EqualsAndHashCode(callSuper = false)
+@EqualsAndHashCode
```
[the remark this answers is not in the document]
MaxKsyunz: Not necessary -- `false` is the default value of `callSuper` and also `OpenSearchRequestBuilder` has no base class so `callSuper` does not apply.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
```diff
@@ -6,91 +6,91 @@
...
 
+  private transient OpenSearchRequest.IndexName indexName;
+  private transient Settings settings;
+  private transient Integer maxResultWindow;
   /** Search request builder. */
   @EqualsAndHashCode.Include
-  @Getter
   @ToString.Include
-  private final OpenSearchRequestBuilder requestBuilder;
+  private transient ExecutableRequestBuilder requestBuilder;
```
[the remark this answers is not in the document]
MaxKsyunz: To indicate that `requestBuilder` field is not serialized.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScanBuilder.java
```diff
@@ -66,10 +70,16 @@ public boolean pushDownAggregation(LogicalAggregation aggregation) {
...
+    delegate = new OpenSearchIndexScanAggregationBuilder(delegate.build(), aggregation);
+    return true;
+  }
 
-    return delegate.pushDownAggregation(aggregation);
+  @Override
+  public boolean pushDownPageSize(LogicalPaginate paginate) {
+    if (isLimitPushedDown) {
+      throw new IllegalStateException("Pagination has to be pushed down before limit.");
+    }
```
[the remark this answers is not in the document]
MaxKsyunz: Yes, the check is unnecessary -- I'll remove it. I am optimistic about pagination and limit :).

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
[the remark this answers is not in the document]
Yury-Fridlyand: Nothing is serialized implicitly if you're using `Externalizable` like here. Do we need to update `ProjectOperator` then to use `transient`?

Yury-Fridlyand:
> Why do you think so? This PR sets up `OpenSearchIndexScan` to handle both limit and pagination. They can be applied in any order.
> 
> Supposing, it's necessary to apply `PushDownPageSize`, how would you implement it?

limit and pagination can't be handled both, because both set the same parameter of a query - `size`.

This starts a scroll search with 5 docs per page:
```
POST http://localhost:9200/calcs/_search?scroll=10m
{
    "size" : 5
}
```
This starts a regular search limited by 42 docs in total:
```
POST http://localhost:9200/calcs/_search
{
    "size" : 42
}
```
It is impossible to do a scrolled search for `N` docs with `M` docs per page.

`pushDownLimit` for paged searches should return `false` to keep `LogicalLimit`/`LimitOperator` in the tree. The only difference between paged and non paged search is a call to `pushDownPageSize` which happens too late now.
I think we need to update Optimizer to preserve the rule order before implementing `LIMIT` clause support for pagination. Maybe it should be done in another PR.

core/src/main/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizer.java
```diff
@@ -52,7 +51,7 @@ public static LogicalPlanOptimizer create() {
          * Phase 2: Transformations that rely on data source push down capability
          */
         new CreateTableScanBuilder(),
-        new CreatePagingTableScanBuilder(),
+        new PushDownPageSize(),
```
acarbonetto, line 54: curious why we aren't using `TableScanPushDown.PUSH_DOWN_PAGESIZE`?  Is pagesize not a concept that the TableScan should know about?

core/src/main/java/org/opensearch/sql/planner/optimizer/PushDownPageSize.java
```diff
@@ -0,0 +1,51 @@
...
+import com.facebook.presto.matching.Captures;
+import com.facebook.presto.matching.Pattern;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.Optional;
+import org.opensearch.sql.planner.logical.LogicalPaginate;
+import org.opensearch.sql.planner.logical.LogicalPlan;
+import org.opensearch.sql.storage.read.TableScanBuilder;
+
+public class PushDownPageSize implements Rule<LogicalPaginate> {
```
acarbonetto, line 17: javadoc please

integ-test/src/test/java/org/opensearch/sql/sql/PaginationWindowIT.java
```diff
@@ -54,15 +55,15 @@ public void testQuerySizeLimitDoesNotEffectTotalRowsReturned() throws IOExceptio
     JSONObject response = executeQueryTemplate("SELECT * FROM %s", TEST_INDEX_PHRASE, 5);
     assertTrue(response.getInt("size") > querySizeLimit);
 
-    String cursor = "";
+    String cursor;
```
acarbonetto, line 58: you can move this into the `do` statement.  There's no need for it here.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/ContinuePageRequestBuilder.java
```diff
@@ -5,94 +5,39 @@
...
 import org.opensearch.sql.opensearch.data.value.OpenSearchExprValueFactory;
-import org.opensearch.sql.opensearch.response.agg.OpenSearchAggregationResponseParser;
 
 /**
  * Builds a {@link ContinuePageRequest} to handle subsequent pagination/scroll/cursor requests.
- * Initial search requests is handled by {@link InitialPageRequestBuilder}.
+ * Initial search requests are handled by {@link OpenSearchRequestBuilder}
  */
-public class ContinuePageRequestBuilder extends PagedRequestBuilder {
+public class ContinuePageRequestBuilder implements ExecutableRequestBuilder {
```
acarbonetto, line 17: I thought the request was to remove this completely?  We still want to keep this around in a reduced form?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -44,104 +42,102 @@
...
+      if (startFrom + size > maxResultWindow) {
+        sourceBuilder.size(maxResultWindow - startFrom);
+        return new OpenSearchScrollRequest(
+            indexName, scrollTimeout, sourceBuilder, exprValueFactory);
+      } else {
+        sourceBuilder.from(startFrom);
+        sourceBuilder.size(querySize);
+        return new OpenSearchQueryRequest(indexName, sourceBuilder, exprValueFactory);
+      }
     } else {
```
acarbonetto, line 121: nit: `else` not needed

acarbonetto (approved):

Yury-Fridlyand:
> > 
> > 1. Last empty page is not issued anymore. If it is expected, you have to update docs and ... demo.
> 
> Why is it needed? As I understand, the last empty page was benign -- an artifact of how `OpenSearchPagedIndexScan` managed OpenSearch requests but not required by any client and not sent by legacy pagination.
> 
> If you could point me to the materials that need updating, I'm happy to update them as well.

I agree that it is odd and could be removed. Just to confirm that you are cleaning up (closing) search context (scroll ID) once last page received.
There is a demo video I recorded (posted on GH and in docs) where I mention the last empty page - it should be patched or re-recorded.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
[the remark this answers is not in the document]
MaxKsyunz: It helped me think about it, but you Yury is right -- it's unnecessary. I removed it.

core/src/main/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizer.java
MaxKsyunz: `PushDownPageSize` does not really fit into the simple pattern of "when node of this type is encountered, call this method" that other `PUSH_DOWN_*` rules use.

It's more similar to `CreateTableScanBuilder` rule.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/ContinuePageRequestBuilder.java
MaxKsyunz: Request was to remove `OpenSearchPagedIndexScan` and use only `OpenSearchIndexScan`. 

Internally, `OpenSearchIndexScan` already used two different request types -- `OpenSearchScrollRequest` and `OpenSearchQueryRequest`. 

`ContinuePagedRequest` is the third one. It has a different use case compare to OpenSearchScrollRequest -- a continuation of a scroll request.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/ExecutableRequestBuilder.java
```diff
@@ -0,0 +1,11 @@
+package org.opensearch.sql.opensearch.request;
+
+import org.opensearch.sql.common.setting.Settings;
+
+public interface ExecutableRequestBuilder {
```
dai-chen, line 5: Could you add JavaDoc for all new interfaces?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PushDownQueryBuilder.java
```diff
@@ -0,0 +1,36 @@
...
+import org.opensearch.sql.planner.logical.LogicalLimit;
+import org.opensearch.sql.planner.logical.LogicalNested;
+import org.opensearch.sql.planner.logical.LogicalPaginate;
+import org.opensearch.sql.planner.logical.LogicalProject;
+import org.opensearch.sql.planner.logical.LogicalSort;
+
+/**
+ * Translates a logical query plan into OpenSearch DSL and an appropriate request.
+ */
+public interface PushDownQueryBuilder {
```
dai-chen, line 20: Could you elaborate the relation between this and `OpenSearchRequestBuilder`?
So this builder builds another builder `OpenSearchRequestBuilder` and meanwhile returns the subclass rather than its interface `ExecutableRequestBuilder`?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
```diff
@@ -152,6 +153,6 @@ public void reset() {
    */
   @Override
   public String toCursor() {
-    return scrollId;
+    return needClean ? "" : scrollId;
```
dai-chen, line 156: Does this change have any impact on the external behavior and protocol that JDBC needs to change accordingly?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
```diff
@@ -168,28 +166,21 @@ public PhysicalPlan implement(LogicalPlan plan) {
...
-  @Override
-  public TableScanBuilder createPagedScanBuilder(int pageSize) {
-    var requestBuilder = new InitialPageRequestBuilder(indexName, pageSize, settings,
-        new OpenSearchExprValueFactory(getFieldOpenSearchTypes()));
-    var indexScan = new OpenSearchPagedIndexScan(client, requestBuilder);
-    return new OpenSearchPagedIndexScanBuilder(indexScan);
+    var builder = new OpenSearchRequestBuilder(
+        settings.getSettingValue(Settings.Key.QUERY_SIZE_LIMIT),
+        new OpenSearchExprValueFactory(allFields));
+    return new OpenSearchIndexScanBuilder(builder) {
```
dai-chen, line 177: Does OpenSearchIndexScanBuilder have other subclass other than this? I'm trying to understand why it becomes abstract class.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScanAggregationBuilder.java
```diff
@@ -8,64 +8,68 @@
...
 import org.opensearch.sql.planner.logical.LogicalSort;
-import org.opensearch.sql.storage.TableScanOperator;
-import org.opensearch.sql.storage.read.TableScanBuilder;
 
 /**
  * Index scan builder for aggregate query used by {@link OpenSearchIndexScanBuilder} internally.
  */
-class OpenSearchIndexScanAggregationBuilder extends TableScanBuilder {
+@EqualsAndHashCode
+class OpenSearchIndexScanAggregationBuilder implements PushDownQueryBuilder {
```
dai-chen, line 36: Could you update the design doc: https://github.com/opensearch-project/sql/blob/main/docs/dev/query-optimizier-improvement.md?
It would be very helpful to understand all these responsibility change as well as the new abstraction you added.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScanAggregationBuilder.java
```diff
@@ -78,6 +82,31 @@ public boolean pushDownSort(LogicalSort sort) {
     return true;
   }
 
+  @Override
+  public boolean pushDownLimit(LogicalLimit limit) {
+    return false;
```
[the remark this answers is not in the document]
dai-chen: or make all these default impl in PushDownQueryBuilder interface?

core/src/main/java/org/opensearch/sql/executor/execution/QueryPlanFactory.java
```diff
@@ -75,15 +78,18 @@ public AbstractPlan createContinuePaginatedPlan(
...
+                             ResponseListener<ExecutionEngine.QueryResponse> queryResponseListener,
+                             ResponseListener<ExecutionEngine.ExplainResponse> explainListener) {
     QueryId queryId = QueryId.queryId();
-    var plan = new ContinuePaginatedPlan(queryId, cursor, queryService,
-            planSerializer, queryResponseListener);
+    var plan = new QueryPlan(queryId, new Cursor(cursor), queryService, queryResponseListener);
     return isExplain ? new ExplainPlan(queryId, plan, explainListener) : plan;
   }
 
+   boolean canConvertToCursor(UnresolvedPlan plan) {
```
Yury-Fridlyand, line 89:
```suggestion
  boolean canConvertToCursor(UnresolvedPlan plan) {
```

core/src/main/java/org/opensearch/sql/planner/ExternalizablePlan.java
```diff
@@ -23,11 +23,11 @@
  *   </li>
  *   <li>
  *     Overwrite {@link #getPlanForSerialization} to return
- *     another instance of {@link SerializablePlan}.
+ *     another instance of {@link ExternalizablePlan}.
```
Yury-Fridlyand, line 26: Why rename?
I feel `SerializablePlan` is self-describing, while `ExternalizablePlan` is not.

core/src/main/java/org/opensearch/sql/planner/LogicalCursor.java
```diff
@@ -0,0 +1,21 @@
+package org.opensearch.sql.planner;
+
+import java.util.List;
+import lombok.Getter;
+import org.opensearch.sql.planner.logical.LogicalPlan;
+import org.opensearch.sql.planner.logical.LogicalPlanNodeVisitor;
+
+public class LogicalCursor extends LogicalPlan {
```
Yury-Fridlyand, line 8: javadoc?

core/src/main/java/org/opensearch/sql/ast/tree/Cursor.java
```diff
@@ -0,0 +1,21 @@
+package org.opensearch.sql.ast.tree;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.opensearch.sql.ast.AbstractNodeVisitor;
+
+@RequiredArgsConstructor
+public class Cursor extends UnresolvedPlan {
```
Yury-Fridlyand, line 8: javadoc?

core/src/test/java/org/opensearch/sql/executor/pagination/PlanSerializerTest.java
```diff
@@ -52,19 +51,19 @@ void setUp() {
...
     var unresolvedPlan = AstDSL.project(AstDSL.relation("table"), AstDSL.allFields());
-    assertTrue(planCache.canConvertToCursor(unresolvedPlan));
+//    assertTrue(planCache.canConvertToCursor(unresolvedPlan));
   }
 
   @Test
   void canConvertToCursor_project_some_fields_relation() {
     var unresolvedPlan = AstDSL.project(AstDSL.relation("table"), AstDSL.field("rando"));
-    Assertions.assertFalse(planCache.canConvertToCursor(unresolvedPlan));
+//    Assertions.assertFalse(planCache.canConvertToCursor(unresolvedPlan));
```
Yury-Fridlyand, line 66: Reminder: uncomment or fix or remove.

core/src/test/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizerTest.java
```diff
@@ -394,19 +392,18 @@ void push_page_size_noop_if_no_sub_plans() {
...
-    when(table.createPagedScanBuilder(anyInt())).thenReturn(pagedTableScanBuilder);
+    when(tableScanBuilder.pushDownPageSize(any())).thenReturn(true);
 
     var relation = new LogicalRelation("schema", table);
     var optimized = LogicalPlanOptimizer.create()
         .optimize(new LogicalPaginate(42, List.of(project(relation))));
     // `optimized` structure: LogicalPaginate -> LogicalProject -> TableScanBuilder
     // LogicalRelation replaced by a TableScanBuilder instance
-    assertEquals(project(pagedTableScanBuilder), optimized);
+    assertEquals(project(tableScanBuilder), optimized);
```
Yury-Fridlyand, line 402: Please update the comment above: `LogicalPaginate` is removed from the optimized tree

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
```diff
@@ -111,15 +115,8 @@ public void clean(Consumer<String> cleanAction) {
...
-   * @return search request
-   */
-  @VisibleForTesting
-  protected SearchRequest searchRequest() {
-    return new SearchRequest()
-        .indices(indexName.getIndexNames())
-        .source(sourceBuilder);
+  @Override
+  public void writeTo(StreamOutput out) throws IOException {
+    throw new UnsupportedOperationException("Not necessary");
```
Yury-Fridlyand, line 120: Maybe: add a comment why

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -37,111 +35,102 @@
...
    */
-  private final Integer maxResultWindow;
+  private int requestedTotalSize;
 
   /**
-   * Search request source builder.
+   * Size of each page request to return.
    */
-  private final SearchSourceBuilder sourceBuilder;
+  private Integer pageSize = null;
```
Yury-Fridlyand, line 72: Use `Optional`

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
```diff
@@ -151,7 +151,40 @@ public void reset() {
...
+
+  @Override
+  public void writeTo(StreamOutput out) throws IOException {
+    initialSearchRequest.writeTo(out);
+    out.writeTimeValue(scrollTimeout);
+    if (!needClean) {
+      // If needClean is true, there is no more data to get from OpenSearch and scrollId is
+      // used only to clean up OpenSearch context.
+
+      out.writeString(scrollId == null? "" : scrollId);
```
Yury-Fridlyand, line 166: I think you need to throw an exception (`NoCursorException`) if scroll is finished

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
```diff
@@ -151,7 +151,40 @@ public void reset() {
...
    */
   @Override
-  public String toCursor() {
-    return scrollId;
+  public boolean hasAnotherBatch() {
+    return !needClean && scrollId != null && !scrollId.equals("");
+  }
+
+  @Override
+  public void writeTo(StreamOutput out) throws IOException {
```
Yury-Fridlyand, line 159: It looks like serializable, but not deserializable

opensearch/src/test/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilderTest.java
```diff
@@ -72,18 +82,13 @@ public class OpenSearchRequestBuilderTest {
...
-    requestBuilder = new OpenSearchRequestBuilder(
-        "test", MAX_RESULT_WINDOW, settings, exprValueFactory);
+    requestBuilder = new OpenSearchRequestBuilder(DEFAULT_LIMIT, exprValueFactory);
   }
 
   @Test
   void build_query_request() {
-    Integer limit = 200;
-    Integer offset = 0;
+    int limit = 200;
```
Yury-Fridlyand, line 90: why?

core/src/main/java/org/opensearch/sql/ast/tree/Cursor.java
```diff
@@ -0,0 +1,21 @@
+package org.opensearch.sql.ast.tree;
```
Yury-Fridlyand, line 1: License header

core/src/main/java/org/opensearch/sql/planner/LogicalCursor.java
```diff
@@ -0,0 +1,21 @@
+package org.opensearch.sql.planner;
```
Yury-Fridlyand, line 1: License header

opensearch/src/main/java/org/opensearch/sql/opensearch/request/ExecutableRequestBuilder.java
```diff
@@ -0,0 +1,9 @@
+package org.opensearch.sql.opensearch.request;
```
Yury-Fridlyand, line 1: License header

Yury-Fridlyand:
Please, update doc on lines 628-634 and remove the duplicate. It became obsolete.
https://github.com/opensearch-project/sql/blob/73f18df8c79653790381e4203652c74d8e274826/docs/dev/Pagination-v2.md?plain=1#L628-L634

core/src/main/java/org/opensearch/sql/ast/AbstractNodeVisitor.java
```diff
@@ -299,4 +300,8 @@ public T visitExplain(Explain node, C context) {
   public T visitPaginate(Paginate paginate, C context) {
     return visitChildren(paginate, context);
   }
+
+  public T visitCursor(Cursor cursor, C context) {
+    return visit(cursor, context);
```
Yury-Fridlyand, line 305: nit
```suggestion
    return visitChildren(cursor, context);
```

core/src/main/java/org/opensearch/sql/planner/LogicalCursor.java
```diff
@@ -0,0 +1,30 @@
...
```
Yury-Fridlyand, line 1: Move to
```suggestion
package org.opensearch.sql.planner.logical;
```

core/src/main/java/org/opensearch/sql/planner/logical/LogicalPlanNodeVisitor.java
```diff
@@ -6,6 +6,7 @@
 
 package org.opensearch.sql.planner.logical;
 
+import org.opensearch.sql.planner.LogicalCursor;
```
Yury-Fridlyand, line 9: You can remove this ref when you move `LogicalCursor`
```suggestion
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -211,6 +213,11 @@ public LogicalPlan visitTableFunction(TableFunction node, AnalysisContext contex
         tableFunctionImplementation.applyArguments());
   }
 
+  @Override
+  public LogicalPlan visitCursor(Cursor cursor, AnalysisContext context) {
```
Yury-Fridlyand, line 217: maybe: move to bottom after `visitPaginate` to keep these things together

sql/src/main/java/org/opensearch/sql/sql/SQLService.java
```diff
@@ -65,10 +65,15 @@ private AbstractPlan plan(
...
       Optional<ResponseListener<QueryResponse>> queryListener,
       Optional<ResponseListener<ExplainResponse>> explainListener) {
+    boolean isExplainRequest = request.isExplainRequest();
     if (request.getCursor().isPresent()) {
       // Handle v2 cursor here -- legacy cursor was handled earlier.
-      return queryExecutionFactory.createContinuePaginatedPlan(request.getCursor().get(),
-          request.isExplainRequest(), queryListener.orElse(null), explainListener.orElse(null));
+      if (isExplainRequest) {
+        throw new UnsupportedOperationException("Explain of a paged query continuation " +
+            "is not supported. Use `explain` for the initial query request.");
```
Yury-Fridlyand, line 73: Why it is here?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
```diff
@@ -68,10 +75,18 @@ public OpenSearchScrollRequest(IndexName indexName,
...
+        .scroll(scrollTimeout)
+        .source(sourceBuilder);
+
+    includes = sourceBuilder.fetchSource() != null && sourceBuilder.fetchSource().includes() != null
+      ? Arrays.asList(sourceBuilder.fetchSource().includes())
+      : List.of();
   }
 
+
   /** Constructor. */
```
Yury-Fridlyand, line 90: Please update or delete this doc

docs/dev/Pagination-v2.md
```diff
@@ -504,6 +504,8 @@ sequenceDiagram
     participant QueryPlanFactory
     participant QueryService
     participant OpenSearchExecutionEngine
+    participant LogicalCursor
+    participant DefaultImplementor
```
Yury-Fridlyand, line 508: These 2 are not used in the graph.
`CreatePagingScanBuilder` in the previous graph is also orphaned.

core/src/main/java/org/opensearch/sql/planner/optimizer/PushDownPageSize.java
```diff
@@ -0,0 +1,55 @@
...
+      throw new IllegalStateException("Failed to push down LogicalPaginate");
+    }
+    return plan.getChild().get(0);
+  }
+
+  private Optional<TableScanBuilder> findTableScanBuilder(LogicalPaginate logicalPaginate) {
+    Deque<LogicalPlan> plans = new ArrayDeque<>();
+    plans.add(logicalPaginate);
+    do {
+      final var plan = plans.removeFirst();
```
penghuo, line 42: final var? should be val?

core/src/main/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizer.java
penghuo: Does it similar to PUSH_DOWN_LIMIT? and should be at the end of rule list?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
penghuo: +1

Yury-Fridlyand:
Please, update graph in `docs/dev/query-optimizer-improvement.md` - it refers to `CreatePagingTableScanBuilder`

sql/src/main/java/org/opensearch/sql/sql/SQLService.java
MaxKsyunz: If we don't check for explain of cursor request, the user gets and error message about invalid cursor. This exception was already being thrown from a different class but that class was removed and the check needed a new home.

Explain in legacy does not work as well so there isn't loss in functionality.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
MaxKsyunz: I did that at first and code analysis got unhappy. Using `Integer` class is reasonable here.

From API docs:
```
Optional is primarily intended for use as a method return type where there is a clear need to represent "no result," and where using null is likely to cause errors.
```

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
MaxKsyunz: It's implementing OpenSearch's serialization API. Deserialization is done in the `StreamInput` constructor.

opensearch/src/test/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilderTest.java
MaxKsyunz: `int` is more appropriate in this case but didn't realize it makes it inconsistent with the rest of the file. I'll change it back for that reason.

A common use case for `Integer` class is store integer values in `List` classes and similar but for simple use cases like `limit` and `offset` in this case it just adds overhead.

core/src/main/java/org/opensearch/sql/planner/ExternalizablePlan.java
```diff
@@ -23,11 +23,11 @@
...
  *   </li>
  * </ul>
  */
-public interface SerializablePlan extends Externalizable {
+public interface ExternalizablePlan extends Externalizable {
```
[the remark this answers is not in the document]
MaxKsyunz: Removed `readExternal` and `writeExternal` -- they were unnecessary here.

`getPlanForSerialization` is used to skip `ResourceMonitorPlan` during serialization.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/ExecutableRequestBuilder.java
MaxKsyunz: Deleted `ExecutableRequestBuilder`  -- it was no longer required.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PushDownQueryBuilder.java
MaxKsyunz: This changed and probably warrants another look.

`PushDownQueryBuilder` is interface a class that can implement push down rules needs to support. 
It's implemented by `OpenSearchIndexScanQueryBuilder` and `OpenSearchIndexScanAggregationBuilder` because simple and aggregation queries have different pushdown logic for limit and sort.

For each push down operation,`OpenSearchRequestBuilder` delegates the actual implementation to either `OpenSearchIndexScanQueryBuilder`. One of those implementations swaps the delegate from `OpenSearchIndexScanQueryBuilder` to `OpenSearchIndexScanAggregationBuilder`.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
MaxKsyunz: This has changed but no, it didn't affect external behaviour -- toCursor was only used for plan serialization for pagination.

core/src/main/java/org/opensearch/sql/planner/ExternalizablePlan.java
MaxKsyunz: I'm of two opinions here -- `ExternalizablePlan` is less obvious but more consistent with Java serialization API name convention. `SerializablePaln` is more obvious but implies that the plan implements `Serializable` which is not the case.

docs/dev/Pagination-v2.md
MaxKsyunz: That was an unfinished revision. It should be better now.

sql/src/main/java/org/opensearch/sql/sql/SQLService.java
Yury-Fridlyand: I think this class isn't a right place of generating response (this exception returned to a user). Probably, you can omit this and let cursor explain query to be processed as explain for a paged request. Take into account that `create` gets `isExplainRequest` argument and creates a proper explain plan for that.

acarbonetto (approved):

core/src/main/java/org/opensearch/sql/ast/AbstractNodeVisitor.java
```diff
@@ -299,4 +300,8 @@ public T visitExplain(Explain node, C context) {
   public T visitPaginate(Paginate paginate, C context) {
     return visitChildren(paginate, context);
   }
+
+  public T visitCursor(FetchCursor cursor, C context) {
```
Yury-Fridlyand, line 304: Rename?
```suggestion
  public T visitFetchCursor(FetchCursor cursor, C context) {
```

core/src/main/java/org/opensearch/sql/ast/tree/FetchCursor.java
```diff
@@ -0,0 +1,28 @@
...
+
+@RequiredArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+public class FetchCursor extends UnresolvedPlan {
+  @Getter
+  final String cursor;
+
+  @Override
+  public <T, C> T accept(AbstractNodeVisitor<T, C> nodeVisitor, C context) {
+    return nodeVisitor.visitCursor(this, context);
```
Yury-Fridlyand, line 21:
```suggestion
    return nodeVisitor.visitFetchCursor(this, context);
```

core/src/main/java/org/opensearch/sql/ast/tree/FetchCursor.java
```diff
@@ -0,0 +1,28 @@
...
+package org.opensearch.sql.ast.tree;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.opensearch.sql.ast.AbstractNodeVisitor;
+
+@RequiredArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+public class FetchCursor extends UnresolvedPlan {
```
Yury-Fridlyand, line 15: javadoc?

core/src/main/java/org/opensearch/sql/planner/DefaultImplementor.java
```diff
@@ -148,6 +150,11 @@ public PhysicalPlan visitRelation(LogicalRelation node, C context) {
         + "implementing and optimizing logical plan with relation involved");
   }
 
+  @Override
+  public PhysicalPlan visitCursor(LogicalFetchCursor plan, C context) {
```
Yury-Fridlyand, line 154:
```suggestion
  public PhysicalPlan visitFetchCursor(LogicalFetchCursor plan, C context) {
```

core/src/main/java/org/opensearch/sql/planner/SerializablePlan.java
```diff
@@ -29,21 +25,6 @@
...
-  /**
-   * Argument is an instance of {@link PlanSerializer.CursorDeserializationStream}.
-   */
-  @Override
-  void readExternal(ObjectInput in) throws IOException, ClassNotFoundException;
-
-  /**
-   * Each plan which has as a child plan should do.
-   * <pre>{@code
-   * out.writeObject(input.getPlanForSerialization());
```
Yury-Fridlyand, line 41: Maybe:  keep this comment (move down)?

core/src/main/java/org/opensearch/sql/planner/logical/LogicalFetchCursor.java
```diff
@@ -0,0 +1,38 @@
...
+   */
+  public LogicalFetchCursor(String cursor, StorageEngine engine) {
+    super(List.of());
+    this.cursor = cursor;
+    this.engine = engine;
+  }
+
+  @Override
+  public <R, C> R accept(LogicalPlanNodeVisitor<R, C> visitor, C context) {
+    return visitor.visitCursor(this, context);
```
Yury-Fridlyand, line 36:
```suggestion
    return visitor.visitFetchCursor(this, context);
```

core/src/main/java/org/opensearch/sql/planner/logical/LogicalFetchCursor.java
```diff
@@ -0,0 +1,38 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import org.opensearch.sql.planner.logical.LogicalPlan;
+import org.opensearch.sql.planner.logical.LogicalPlanNodeVisitor;
+import org.opensearch.sql.storage.StorageEngine;
+
+@EqualsAndHashCode(callSuper = false)
+@ToString
+public class LogicalFetchCursor extends LogicalPlan {
```
Yury-Fridlyand, line 18: javadoc?

core/src/main/java/org/opensearch/sql/planner/logical/LogicalPlanDSL.java
```diff
@@ -32,6 +31,10 @@
 @UtilityClass
 public class LogicalPlanDSL {
 
+  public static LogicalPlan cursor(String cursor, StorageEngine engine) {
```
Yury-Fridlyand, line 34:
```suggestion
  public static LogicalPlan fetchCursor(String cursor, StorageEngine engine) {
```

core/src/main/java/org/opensearch/sql/planner/logical/LogicalPlanNodeVisitor.java
```diff
@@ -108,4 +108,8 @@ public R visitAD(LogicalAD plan, C context) {
   public R visitPaginate(LogicalPaginate plan, C context) {
     return visitNode(plan, context);
   }
+
+  public R visitCursor(LogicalFetchCursor plan, C context) {
```
Yury-Fridlyand, line 112:
```suggestion
  public R visitFetchCursor(LogicalFetchCursor plan, C context) {
```

core/src/main/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizer.java
```diff
@@ -52,11 +51,11 @@ public static LogicalPlanOptimizer create() {
          * Phase 2: Transformations that rely on data source push down capability
          */
         new CreateTableScanBuilder(),
-        new CreatePagingTableScanBuilder(),
         TableScanPushDown.PUSH_DOWN_FILTER,
         TableScanPushDown.PUSH_DOWN_AGGREGATION,
         TableScanPushDown.PUSH_DOWN_SORT,
         TableScanPushDown.PUSH_DOWN_LIMIT,
+        new PushDownPageSize(),
```
Yury-Fridlyand, line 58: Does it really follow the order of rules?

integ-test/src/test/java/org/opensearch/sql/ppl/StandaloneIT.java
```diff
@@ -241,7 +241,7 @@ public QueryPlanFactory queryPlanFactory(ExecutionEngine executionEngine,
               new ExpressionAnalyzer(functionRepository), dataSourceService, functionRepository);
       Planner planner = new Planner(LogicalPlanOptimizer.create());
       QueryService queryService = new QueryService(analyzer, executionEngine, planner);
-      return new QueryPlanFactory(queryService, planSerializer);
+      return new QueryPlanFactory(queryService);
```
Yury-Fridlyand, line 244: Should we update `queryPlanFactory` here and in the plugin to remove the second argument? Probably you can remove function which `@Provides` a `PlanSerializer` instance too.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
```diff
@@ -34,32 +38,38 @@
...
   private boolean needClean = false;
 
-  /** Search request source builder. */
-  private final SearchSourceBuilder sourceBuilder;
+  @Getter
+  private List<String> includes;
+
+  /** Default constructor for Externalizable only.
+   */
+  public OpenSearchScrollRequest() {
```
Yury-Fridlyand, line 71: Make `@Deprecated`

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
```diff
@@ -143,15 +147,50 @@ public SearchScrollRequest scrollRequest() {
...
+  public void writeTo(StreamOutput out) throws IOException {
+    initialSearchRequest.writeTo(out);
+    out.writeTimeValue(scrollTimeout);
+    out.writeBoolean(needClean);
+    if (!needClean) {
+      // If needClean is true, there is no more data to get from OpenSearch and scrollId is
+      // used only to clean up OpenSearch context.
+
+      out.writeString(scrollId);
+    }
```
Yury-Fridlyand, line 172: I guess you need to throw an `NoCursorException` in `else`

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
```diff
@@ -143,15 +147,50 @@ public SearchScrollRequest scrollRequest() {
...
+   * @throws IOException thrown if reading from input {@param in} fails.
+   */
+  public OpenSearchScrollRequest(StreamInput in, OpenSearchStorageEngine engine)
+      throws IOException {
+    initialSearchRequest = new SearchRequest(in);
+    scrollTimeout = in.readTimeValue();
+    needClean = in.readBoolean();
+    if (!needClean) {
+      scrollId = in.readString();
+    }
```
Yury-Fridlyand, line 190: Is it possible?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
```diff
@@ -126,6 +110,54 @@ public void close() {
...
+    request.writeTo(reqOut);
+    reqOut.flush();
+
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    reqOut.bytes().writeTo(baos);
+    baos.flush();
+
+    var reqAsBytes = baos.toByteArray();
+    out.writeInt(reqAsBytes.length);
+    out.write(reqAsBytes);
```
Yury-Fridlyand, line 159: Why intermediate streams are required? May be leave a comment

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
```diff
@@ -126,6 +110,54 @@ public void close() {
...
+  @Deprecated(since = "introduction")
+  public OpenSearchIndexScan() {
+  }
+
+  @Override
+  public void readExternal(ObjectInput in) throws IOException {
+    int reqSize = in.readInt();
+    byte[] requestStream = new byte[reqSize];
+    in.read(requestStream);
+
```
Yury-Fridlyand, line 128: `Please   remove   extra   whitespaces`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScanAggregationBuilder.java
```diff
@@ -8,64 +8,68 @@
...
-    return indexScan;
+    return requestBuilder;
   }
 
   @Override
-  public boolean pushDownAggregation(LogicalAggregation aggregation) {
-    aggregatorList = aggregation.getAggregatorList();
-    groupByList = aggregation.getGroupByList();
-    return true;
+  public boolean pushDownFilter(LogicalFilter filter) {
```
Yury-Fridlyand, line 71: Shouldn't be there a dummy `pushDownAggregation` which just returns `true`?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PushDownQueryBuilder.java
Yury-Fridlyand: Maybe post this as a code comment or in the doc?

plugin/src/main/java/org/opensearch/sql/plugin/config/OpenSearchPluginModule.java
```diff
@@ -108,6 +108,6 @@ public QueryPlanFactory queryPlanFactory(DataSourceService dataSourceService,
     Planner planner = new Planner(LogicalPlanOptimizer.create());
     QueryService queryService = new QueryService(
         analyzer, executionEngine, planner);
-    return new QueryPlanFactory(queryService, planSerializer);
+    return new QueryPlanFactory(queryService);
```
Yury-Fridlyand, line 111: `PlanSerializer` is not needed here anymore as long as a function which `@Provides` it

core/src/test/java/org/opensearch/sql/planner/ExternalizablePlanTest.java
```diff
@@ -18,9 +18,9 @@
 
 @ExtendWith(MockitoExtension.class)
 @DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
-public class SerializablePlanTest {
+public class ExternalizablePlanTest {
```
Yury-Fridlyand: Rename back?

docs/dev/Pagination-v2.md
Yury-Fridlyand: Please remove `CreatePagingScanBuilder` in the previous graph

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScanAggregationBuilder.java
MaxKsyunz: Not needed because `pushDownAggregation` does not apply to non-aggregation queries. 

For example, `pushDownLimit` is on the builder interface because both aggregate and non-aggregate queries can have a limit clause but when aggregation is encountered, query scan builder is converted to an aggregation scan builder.

Yury-Fridlyand: I mean if `pushDownAggregation` of a `OpenSearchIndexScanAggregationBuilder` is called it would go to default implementation which returns `false` or throws.

core/src/main/java/org/opensearch/sql/executor/execution/QueryPlanFactory.java
```diff
@@ -75,15 +76,18 @@ public AbstractPlan createContinuePaginatedPlan(
   /**
    * Creates a ContinuePaginatedPlan from a cursor.
    */
-  public AbstractPlan createContinuePaginatedPlan(String cursor, boolean isExplain,
-      ResponseListener<ExecutionEngine.QueryResponse> queryResponseListener,
-      ResponseListener<ExecutionEngine.ExplainResponse> explainListener) {
+  public AbstractPlan create(String cursor, boolean isExplain,
```
Yury-Fridlyand, line 79: Please, update the doc above.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
MaxKsyunz: Throwing by `OpenSearchIndexScan` is sufficient.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
MaxKsyunz: It's a valid state of the request. 

The only reason scrollId is not serialized here is because when `needClean` is true, then scrollId will be invalid after this request is cleaned up.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
MaxKsyunz: Fair point. At the time didn't see a better way to serialize request to byte stream.

Found a better way and commented this section.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScanAggregationBuilder.java
MaxKsyunz: `OpenSearchIndexScanAggregationBuilder` does not have `pushDownAggregation` as far as I can see.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
Yury-Fridlyand: I see it is not used anymore in deserialization. This class doesn't implement `Readable` or `Externalizable`.
Can you remove this constructor?

core/src/main/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizer.java
MaxKsyunz: Optimizer visits nodes in the query plan in depth-first manner -- table scan is visited first, then the plan that has the table scan as a child, and so on.  For each node, the rules are checked and applied in the oder listed here.

If two rules apply to the same node and the order matters, then this list determines it.

core/src/main/java/org/opensearch/sql/ast/AbstractNodeVisitor.java
acarbonetto: lave for later

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
Yury-Fridlyand: True, I missed that.

core/src/main/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizer.java
MaxKsyunz: Why should limit be at the end of the list? I did move `PushDownPageSize` to be after limit.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchScrollRequest.java
MaxKsyunz: Good catch.

acarbonetto (approved):
Please respond to Chen/Peng on this: 
[OpenSearchIndex.java](https://github.com/opensearch-project/sql/pull/1600/files#diff-fc793365d68310cd7b92007c94343f021e4647fee5bf53a41003374b8810a74e)

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
MaxKsyunz: OpenSearchIndexScanBuilder can be made an inner class of OpenSearchIndex. I can do that as part of follow-up work -- doing so now requires a lot of tests to be updated.

It became abstract because building an OpenSearchIndexScan and OpenSearchRequest depends on some runtime state  which is now provided by this inner class.  

Overall, the new class structure is simpler than the previous one and was required to add serialization to OpenSearchIndexScan.

For reference, new structure is
```mermaid
classDiagram
direction BT
class OpenSearchIndex
class OpenSearchIndexScan
class OpenSearchIndexScanBuilder
class OpenSearchIndexScanQueryBuilder
class OpenSearchQueryRequest
class OpenSearchRequest {
<<Interface>>

}
class OpenSearchRequestBuilder

OpenSearchIndex  ..>  OpenSearchIndexScan : «create»
OpenSearchIndex  ..>  OpenSearchRequestBuilder : «create»
OpenSearchIndexScan "1" *--> "request 1" OpenSearchRequest 
OpenSearchIndexScanBuilder  ..>  OpenSearchIndexScanQueryBuilder : «create»
OpenSearchIndexScanQueryBuilder "1" *--> "requestBuilder 1" OpenSearchRequestBuilder 
OpenSearchQueryRequest  ..>  OpenSearchRequest 
OpenSearchRequestBuilder  ..>  OpenSearchQueryRequest : «create»


```

Previous structure was
```mermaid
classDiagram
direction BT
class OpenSearchIndex
class OpenSearchIndexScan
class OpenSearchIndexScanQueryBuilder
class OpenSearchQueryRequest
class OpenSearchRequest {
<<Interface>>

}
class OpenSearchRequestBuilder

OpenSearchIndex  ..>  OpenSearchIndexScan : «create»
OpenSearchIndexScan "1" *--> "request 1" OpenSearchRequest 
OpenSearchIndexScan  ..>  OpenSearchRequestBuilder : «create»
OpenSearchIndexScan "1" *--> "requestBuilder 1" OpenSearchRequestBuilder 
OpenSearchIndexScanQueryBuilder "1" *--> "indexScan 1" OpenSearchIndexScan 
OpenSearchQueryRequest  ..>  OpenSearchRequest 
OpenSearchRequestBuilder  ..>  OpenSearchQueryRequest : «create»


```

core/src/test/java/org/opensearch/sql/planner/logical/LogicalPlanNodeVisitorTest.java
```diff
@@ -130,9 +131,11 @@ public TableWriteOperator build(PhysicalPlan child) {
 
     LogicalNested nested = new LogicalNested(null, nestedArgs, projectList);
 
+    LogicalFetchCursor cursor = new LogicalFetchCursor("n:test", mock(StorageEngine.class));
     return Stream.of(
         relation, tableScanBuilder, write, tableWriteBuilder, filter, aggregation, rename, project,
-        remove, eval, sort, dedup, window, rareTopN, highlight, mlCommons, ad, ml, paginate, nested
+        remove, eval, sort, dedup, window, rareTopN, highlight, mlCommons, ad, ml, paginate, nested,
+        cursor
```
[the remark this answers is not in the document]
Yury-Fridlyand: I merged this recently to avoid duplicating code in tests...

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
Yury-Fridlyand: Please, add this graph to the doc.

Yury-Fridlyand:
Please, update PR description - for example, you don't have `ExecutableRequestBuilder` anymore.

Yury-Fridlyand (approved):
LGFM
Last few minor things left.

Issue #1686, opened 2023-05-30
[BUG] Date formats are not recognized for nested type fields


GumpacG:
**What is the bug?**
Date formats are not formatted for nested type.

**How can one reproduce the bug?**
1. Load mappings file:
```
{
  "mappings": {
    "properties": {
      "message": {
        "type": "nested",
        "properties": {
          "info": {
            "type": "keyword",
            "index": "true"
          },
          "author": {
            "type": "keyword",
            "fields": {
              "keyword": {
                "type": "keyword",
                "ignore_above": 256
              }
            },
            "index": "true"
          },
          "dayOfWeek": {
            "type": "long"
          },
          "date": {
            "type": "date",
            "format": "date"
          },
          "time": {
            "type": "date",
            "format": "hour_minute_second"
          },
          "timestamp": {
            "type": "date",
            "format": "date_time"
          }
        }
      },
      "comment": {
        "type": "nested",
        "properties": {
          "data": {
            "type": "keyword",
            "index": "true"
          },
          "likes": {
            "type": "long"
          }
        }
      },
      "myNum": {
        "type": "long"
      },
      "someField": {
        "type": "keyword",
        "index": "true"
      }
    }
  }
}

```
3. Load data
```
{"index":{"_id":"1"}}
{"message":{"info":"a","author":"e","dayOfWeek":1,"date":"1984-04-12","time":"08:47:52","blah":"1984-04-12T08:47:52.000Z"},"comment":{"data":"ab","likes":3},"myNum":1,"someField":"b"}
{"index":{"_id":"2"}}
{"message":{"info":"b","author":"f","dayOfWeek":2,"date":"2023-05-30","time":"06:50:16","blah":"2023-05-30T06:50:16.000Z"},"comment":{"data":"aa","likes":2},"myNum":2,"someField":"a"}
{"index":{"_id":"3"}}
{"message":{"info":"c","author":"g","dayOfWeek":1,"date":"1974-12-11","time":"01:25:02","blah":"1974-12-11T01:25:02.000Z"},"comment":{"data":"aa","likes":3},"myNum":3,"someField":"a"}
{"index":{"_id":"4"}}
[...]

Issue #1300, opened 2023-01-27
[DISCUSSION] Properly support array values in new engine


GumpacG:
## What is the bug?
The new engine does not return values in an array while the Legacy engine returns all values in a row as an array. Implementing same support as `V1` does isn't a right way, because legacy engine produces inconsistent value.

## How can one reproduce the bug?
1. Start OpenSearch server
2. Clean index if was created before: `curl -XDELETE 'http://localhost:9200/dbg'`
3. Create a simple index with automatic mapping: `curl -X POST "localhost:9200/dbg/_doc/?pretty" -H 'Content-Type: application/json' -d '{"myNum": 5}'`
4. Query data: `select * from dbg`. Not bas so far.
5. Add new doc: `curl -X POST "localhost:9200/dbg/_doc/?pretty" -H 'Content-Type: application/json' -d '{"myNum": [3, 4]}'`
6. Check mapping: `curl -X GET "localhost:9200/dbg?pretty"`
```json
"mappings" : {
  "properties" : {
	"myNum" : {
	  "type" : "long"
	}
  }
}
```
7. Query in the new engine: `curl -s -XPOST http://localhost:9200/_plugins/_sql -H 'Content-Type: application/json' -d '{"query": "select * from dbg"}'`
```json
{
  "schema": [
    {
      "name": "myNum",
      "type": "long"
    }
  ],
  "datarows": [
    [
      5
    ],
    [
      3
    ]
  ],
  "total": 2,
  "size": 2,
  "status": 200
}
```
(if you have only second doc in the index)
```json
"schema": [
    {
      "name": "myNum",
      "type": "long"
    }
  ],
  "datarows": [
    [
      3
    ],
    [
      4
    ]
  ]
```
8. Query in the legacy engine: `curl -s -XPOST http://localhost:9200/_plugins/_sql -H 'Content-Type: application/json' -d '{"query": "select * from dbg", "fetch_size": 20}'`
```json
{
  "schema": [
    {
      "name": "myNum",
      "type": "long"
    }
  ],
  "total": 2,
  "datarows": [
    [
      5
    ],
    [
      [
        3,
        4
      ]
    ]
  ],
  "size": 2,
  "status": 200
}
```

## What is the expected behavior?
**TBD**

### Why legacy response is incorrect?
It declares data type as long, but returns a number **and** array of numbers. Imagine a user has a parser for response, what should parser do with such values?
You can try our JDBC driver as an example of a customer application.

## What is your host/environment?
[...]

PR #1737, opened 2023-06-14
Support Array and ExprValue Parsing With Inner Hits

forestmvey:
### Description
With the addition of the nested function support in the V2 engine the `inner_hits` portion of the OS response needs to  be parsed for OS types. Previously arrays were not entirely parsed due to lack of support.

### Parsing Functionality
Parsing of each value in arrays needs to be supported for proper `nested` functionality. The following gives examples of the current response formats when parsing arrays.

***

#### Object Type - Array Of Objects
Dataset:
`"obj": [{"key": "val1"}, {"key": "val2"}]`

SQL Query:
```sql
SELECT obj FROM objects;
```
V1 Engine Response:
`[{'key': 'val1'},{'key': 'val2'}]`

V2 Engine Response:
`{'key': 'val1'}`

***

#### Nested Type - Array Of Objects
Dataset:
`"obj": [{"key": "val1"}, {"key": "val2"}]`

SQL Query:
```sql
SELECT obj FROM nested_objects;
```
V1 and V2 Engine Response:
`[{'key': 'val1'},{'key': 'val2'}]`

***

#### Array Of Values (Nested only supports objects and not arrays of values):
Dataset:
`"obj": ["val1", "val2"]`

SQL Query:
```sql
SELECT obj FROM objects;
```
V1 Engine Response:
`['val1','val2']`

V2 Engine Response:
`{'val1'}`

***

### Issues Resolved
Issue: [1686](https://github.com/opensearch-project/sql/issues/1686)

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java
@@ -6,6 +6,7 @@
 package org.opensearch.sql.sql;
 
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_MULTI_NESTED_TYPE;
+import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_NESTED_SIMPLE;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_NESTED_TYPE;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_NESTED_TYPE_WITHOUT_ARRAYS;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_NESTED_WITH_NULLS;
@@ -31,6 +32,7 @@ public class NestedIT extends SQLIntegTestCase {
     loadIndex(Index.NESTED_WITHOUT_ARRAYS);
     loadIndex(Index.EMPLOYEE_NESTED);
     loadIndex(Index.NESTED_WITH_NULLS);
+    loadIndex(Index.NESTED_SIMPLE);
   }
 
   @Test
@@ -365,5 +367,35 @@ public class NestedIT extends SQLIntegTestCase {
     JSONObject result = executeJdbcRequest(query);
     assertEquals(1, result.getInt("total"));
     verifyDataRows(result, rows(10, "a"));
+  }
+
+  @Test
+  public void nested_function_with_date_types_as_object_arrays_within_arrays_test() {
+    String query = "SELECT nested(address.moveInDate) FROM " + TEST_INDEX_NESTED_SIMPLE;
+    JSONObject result = executeJdbcRequest(query);
+
+    assertEquals(11, result.getInt("total"));
+    verifyDataRows(result,
+        rows(new JSONObject(Map.of("dateAndTime","1984-04-12 09:07:42"))),
+        rows(new JSONArray(
+            List.of(
+                Map.of("dateAndTime", "2023-05-03 08:07:42"),
+                Map.of("dateAndTime", "2001-11-11 04:07:44"))
+            )
+        ),
+        rows(new JSONObject(Map.of("dateAndTime", "1966-03-19 03:04:55"))),
+        rows(new JSONObject(Map.of("dateAndTime", "2011-06-01 01:01:42"))),
+        rows(new JSONObject(Map.of("dateAndTime", "1901-08-11 04:03:33"))),
+        rows(new JSONObject(Map.of("dateAndTime", "2023-05-03 08:07:42"))),
+        rows(new JSONObject(Map.of("dateAndTime", "2001-11-11 04:07:44"))),
+        rows(new JSONObject(Map.of("dateAndTime", "1977-07-13 09:04:41"))),
+        rows(new JSONObject(Map.of("dateAndTime", "1933-12-12 05:05:45"))),
+        rows(new JSONObject(Map.of("dateAndTime", "1909-06-17 01:04:21"))),
+        rows(new JSONArray(
+            List.of(
+                Map.of("dateAndTime", "2001-11-11 04:07:44"))
+            )
+        )
+    );
   }
 }
--- a/integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java
@@ -375,6 +375,9 @@ public class NestedIT extends SQLIntegTestCase {
     JSONObject result = executeJdbcRequest(query);
 
     assertEquals(11, result.getInt("total"));
+    verifySchema(result,
+        schema("nested(address.moveInDate)", null, "object")
+    );
     verifyDataRows(result,
         rows(new JSONObject(Map.of("dateAndTime","1984-04-12 09:07:42"))),
         rows(new JSONArray(
--- a/integ-test/src/test/resources/indexDefinitions/nested_simple_index_mapping.json
+++ b/integ-test/src/test/resources/indexDefinitions/nested_simple_index_mapping.json
@@ -21,6 +21,14 @@
                 "ignore_above": 256
               }
             }
+          },
+          "moveInDate" : {
+            "properties": {
+              "dateAndTime": {
+                "type": "date",
+                "format": "basic_date_time"
+              }
+            }
           }
         }
       },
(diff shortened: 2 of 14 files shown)


Yury-Fridlyand (approved):

GumpacG (dismissed):

integ-test/src/test/resources/indexDefinitions/nested_simple_index_mapping.json
```diff
@@ -21,6 +21,14 @@
                 "ignore_above": 256
               }
             }
+          },
+          "moveInDate" : {
```
acarbonetto, line 25: did we want to test this with a `nested` type too?

acarbonetto (approved):

integ-test/src/test/resources/indexDefinitions/nested_simple_index_mapping.json
forestmvey: This is a inner object to the nested object `address`. We have IT tests to test with the `nested` function in  [NestedIT.java#L373-L400](https://github.com/opensearch-project/sql/blob/71147ffc14ff7460f420d083a92b707481570e10/integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java#L373-L400). Is this what you were thinking of testing?

integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java
```diff
@@ -366,4 +368,34 @@ public void test_nested_in_where_as_predicate_expression_with_relevance_query()
...
     verifyDataRows(result, rows(10, "a"));
   }
+
+  @Test
+  public void nested_function_with_date_types_as_object_arrays_within_arrays_test() {
+    String query = "SELECT nested(address.moveInDate) FROM " + TEST_INDEX_NESTED_SIMPLE;
+    JSONObject result = executeJdbcRequest(query);
+
+    assertEquals(11, result.getInt("total"));
+    verifyDataRows(result,
```
penghuo, line 378: what would be the data type of nested(address.moveInDate), object or array?

opensearch/src/test/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactoryTest.java
```diff
@@ -390,22 +401,239 @@ public void constructArray() {
...
+                + "]}"
+        ).get("stringV"));
+  }
+
+  @Test
+  public void constructNestedArraysOfStringsReturnsFirstIndex() {
+    assertEquals(
+        stringValue("zz"),
+        tupleValue(
+            "{\"stringV\":["
```
penghuo, line 448: what if the data is?
```
{
  "stringV": ["z", ["s"], ["zz", "au"]]
}
```

integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java
forestmvey: It would be of type object. address would be of type nested, I'll add the schema verification.

opensearch/src/test/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactoryTest.java
forestmvey: With this array structure the parsed value will be "z". This is the same as prior to this PR. I will add a test with this case.

Yury-Fridlyand: @penghuo your case mostly related to #1300

opensearch/src/test/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactoryTest.java
```diff
@@ -390,22 +401,252 @@ public void constructArray() {
...
+            List.of(
+                collectionValue(
+                  List.of("zz", "au")
+                ),
+                collectionValue(
+                    List.of("ss")
+                )
+            )
+        ),
+        tupleValueWithArraySupport(
```
penghuo, line 435: ignore my comments below. 
what if the data is? the expected stringV is [["zz", "au"], ["ss"], ["z"]], right?
```
{ "stringV": ["z", ["s"], ["zz", "au"]] }
```

integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java
penghuo: i see. one more question. 
some rows is JSONObject, some rows are JSONArray, how does JDBC handle it?

opensearch/src/test/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactoryTest.java
forestmvey: If we execute on this:
```
tupleValueWithArraySupport("{"stringV": ["z", ["s"], ["zz", "au"]]}").get("stringV");
```

The resulting value is:
```
["z", ["s"], ["zz", "au"]]
```

In your example the order is reversed which is not the case. Does this answer your question?

integ-test/src/test/resources/indexDefinitions/nested_simple_index_mapping.json
acarbonetto: Yes.  I think I was wondering if we had nested within nested.

integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java
forestmvey: In some cases this breaks JDBC, but there is no change in the expected output with this PR. An example of this would be in `ObjectFieldSelectIT` where we do a select on an `OBJECT` type array, and a `NESTED` type array.

[ObjectFieldSelectIT.java#L69-L93](https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/java/org/opensearch/sql/legacy/ObjectFieldSelectIT.java#L69-L93)
[people.json](https://github.com/opensearch-project/sql/blob/main/doctest/test_data/people.json)
[people-mapping](https://github.com/opensearch-project/sql/blob/main/doctest/test_mapping/people.json#L18-L30)

When selecting on an array that is not of type `NESTED`, the first value in the array is returned and is functional with JDBC. As shown by the query:
```SQL
SELECT accounts FROM people;
```

While on the other hand the query that does a basic SELECT on an `NESTED` type array returns the whole array which breaks JDBC.

Query:
```SQL
SELECT projects FROM people;
```

A user can get around this by using the `nested` function to flatten the response for a correct JDBC output.

integ-test/src/test/resources/indexDefinitions/nested_simple_index_mapping.json
forestmvey: Yes we already have tests for nested in nested too.
[NestedIT.java#L106-L118](https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/java/org/opensearch/sql/sql/NestedIT.java#L106-L118)

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -338,39 +353,97 @@ private ExprValue createOpenSearchDateType(Content value, ExprType type) {
...
+    } else if (content.isString()) {
+      return parse(content, prefix, Optional.of(OpenSearchDataType.of(STRING)), supportArrays);
+    } else if (content.isLong()) {
+      return parse(content, prefix, Optional.of(OpenSearchDataType.of(LONG)), supportArrays);
+    } else if (content.isFloat()) {
+      return parse(content, prefix, Optional.of(OpenSearchDataType.of(FLOAT)), supportArrays);
+    } else if (content.isDouble()) {
+      return parse(content, prefix, Optional.of(OpenSearchDataType.of(DOUBLE)), supportArrays);
+    } else if (content.isNumber()) {
+      return parse(content, prefix, Optional.of(OpenSearchDataType.of(INTEGER)), supportArrays);
```
matthewryanwells, line 433: If you are using something with type of `SHORT` or `BYTE` does it get cast/treated as an `INTEGER` before this point or does it need to be supported?

GumpacG (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
forestmvey: Treated as an integer. No need to cast or treat differently.

PR #1752, opened 2023-06-17
Design for `LIMIT` in pagination.

Yury-Fridlyand:
### Description
This PR is design review for supporting `LIMIT` in pagination. It includes:
- doc describing changes need for this feature
- MVP PoC code

### TODOs:
- [ ] Close cursor when limit reached
- [ ] Optimize push down operations - don't push down page size if `page_size > limit`
- [ ] Test `MergeFilterAndFilter` and `PushFilterUnderSort` optimizer rules (probably by combining `WHERE` and `HAVING` clauses or with PPL query)

### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/docs/dev/limit-in-pagination.md
+++ b/docs/dev/limit-in-pagination.md
@@ -1,0 +1,62 @@
+## Background
+
+Scrolled search is performed by request to `http://localhost:9200/<index>/_search?scroll=<timeout>` endpoint and it has a mandatory field `size`, which defines page size:
+```json
+{
+    "size" : 5
+}
+```
+Regular (non-scrolled/non-paged) search is performed by request to `http://localhost:9200/<index</_search` endpoint and it has not mandatory fields. Parameter `size` though defines maximum number of docs to be returned by search.
+
+## Problem statement
+
+`LIMIT` clause is being converted to `size` by SQL plugin during push down operation.
+Hereby comes the conflict in using `LIMIT` with pagination.
+
+## Solution
+
+Don't do push down for `LIMIT`. `LimitOperator` Physical Plan Tree node will cut off yielding search results with minimal overhead.
+It seems simple, but here comes another problem.
+
+## Impediments
+
+`Optimizer` which calls push down operations sets page size after limit, which overwrites this settings. `Optimizer` does not apply rules in the order they are given.
+
+## Fix
+
+Rework [`Optimizer`](https://github.com/opensearch-project/sql/blob/57ce303740f64fe0279fc34aab0116f33f11fbe6/core/src/main/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizer.java).
+Current behavior:
+```py
+def Optimize:
+    for node in PlanTree: # Traverse the Logical Plan Tree
+        for rule in rules: # Enumerate rules
+            tryApplyRule()
+```
+
+Expected behavior:
+```py
+def Optimize:
+    for rule in rules: # Enumerate rules
+        for node in PlanTree: # Traverse the Logical Plan Tree
+            tryApplyRule()
+```
+Rules list:
+```
+...
+CreateTableScanBuilder
+PushDownPageSize
+...
+PUSH_DOWN_LIMIT
+...
+```
+
+This gives us warranty that `pushDownLimit` operation would be rejected if `pushPageSize` called before. Then, not optimized Logical Plan Tree node `LogicalLimit` will be converted to `LimitOperator` Physical Plan tree node.
+
+## Other changes
+
+1. Make `Optimizer` rules applied only once.
+2. Reorder `Optimizer` rules.
+3. Make `LimitOperator` properly serialized and deserialized.
+4. Make `OpenSearchIndexScanBuilder::pushDownLimit` return `false` if `pushDownPageSize` was called before.
+5. (Optional) Groom `Optimizer` to reduce amount of unchecked casts and uses raw classes.
+6. (Optional) Rework `Optimizer` to make it a tree visitor.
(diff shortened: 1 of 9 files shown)


docs/dev/limit-in-pagination.md
```diff
@@ -0,0 +1,62 @@
...
+Regular (non-scrolled/non-paged) search is performed by request to `http://localhost:9200/<index</_search` endpoint and it has not mandatory fields. Parameter `size` though defines maximum number of docs to be returned by search.
+
+## Problem statement
+
+`LIMIT` clause is being converted to `size` by SQL plugin during push down operation.
+Hereby comes the conflict in using `LIMIT` with pagination.
+
+## Solution
+
+Don't do push down for `LIMIT`. `LimitOperator` Physical Plan Tree node will cut off yielding search results with minimal overhead.
```
forestmvey, line 18: Rather than have an operator to limit the output with post processing why not include it as part of the base class. We can have an optional limit if push down isn't available that can be performed in the `PhysicalPlan` base class prior to the inheriting class's `next()` call. Another option would be to have the limit post-processing performed in the `ResourceMonitorPlan` prior to calling the `delegate` `next()`.

docs/dev/limit-in-pagination.md
```diff
@@ -0,0 +1,62 @@
...
+```json
+{
+    "size" : 5
+}
+```
+Regular (non-scrolled/non-paged) search is performed by request to `http://localhost:9200/<index</_search` endpoint and it has not mandatory fields. Parameter `size` though defines maximum number of docs to be returned by search.
+
+## Problem statement
+
+`LIMIT` clause is being converted to `size` by SQL plugin during push down operation.
```
acarbonetto, line 13: We need to properly define what we want to accomplish, but discuss what is currently the problem with the code.  I would change this to talk about LIMIT and size conflict, but also include the use cases:
1. When LIMIT > size
2. When size < LIMIT
3. When size == LIMIT (this is kind of naive case, but maybe just mention it anyways)
Then discuss exactly how we expect the system to behave (which does it return a cursor, when does it NOT return a cursor, where does it break (max_window_size), and does this override the default fetch size. 
If there is anything to add in terms of how the JDBC driver will behave, then include it here.

docs/dev/limit-in-pagination.md
acarbonetto, line 18: I don't think this is the solution we talked about in the walk-through... I think you just need to reverse the business logic to execute all the rules on nodes before proceeding to the next node...

acarbonetto: Seems like this section is similar to `Fix` below

docs/dev/limit-in-pagination.md
```diff
@@ -0,0 +1,62 @@
...
+        for node in PlanTree: # Traverse the Logical Plan Tree
+            tryApplyRule()
+```
+Rules list:
+```
+...
+CreateTableScanBuilder
+PushDownPageSize
+...
+PUSH_DOWN_LIMIT
```
acarbonetto, line 49: why is this ordering important?

docs/dev/limit-in-pagination.md
```diff
@@ -0,0 +1,62 @@
...
+...
+```
+
+This gives us warranty that `pushDownLimit` operation would be rejected if `pushPageSize` called before. Then, not optimized Logical Plan Tree node `LogicalLimit` will be converted to `LimitOperator` Physical Plan tree node.
```
acarbonetto, line 53:
```suggestion
This gives us guarantee that `pushDownLimit` operation will be rejected if `pushPageSize` already called. In that case, the not-optimized Logical Plan Tree node `LogicalLimit` will be converted to `LimitOperator` Physical Plan tree node.

```

docs/dev/limit-in-pagination.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Other changes
+
+1. Make `Optimizer` rules applied only once.
+2. Reorder `Optimizer` rules.
+3. Make `LimitOperator` properly serialized and deserialized.
+4. Make `OpenSearchIndexScanBuilder::pushDownLimit` return `false` if `pushDownPageSize` was called before.
+5. (Optional) Groom `Optimizer` to reduce amount of unchecked casts and uses raw classes.
+6. (Optional) Rework `Optimizer` to make it a tree visitor.
```
acarbonetto, line 62: Out of scope work.  Please raise a follow-up ticket to do these later.  You can label this as `maintenance` ticket.

prometheus/src/main/java/org/opensearch/sql/prometheus/planner/logical/PrometheusLogicalPlanOptimizerFactory.java
```diff
@@ -23,10 +28,10 @@ public class PrometheusLogicalPlanOptimizerFactory {
    * Create Prometheus storage specified logical plan optimizer.
    */
   public static LogicalPlanOptimizer create() {
-    return new LogicalPlanOptimizer(Arrays.asList(
+    return new LogicalPlanOptimizer(Stream.of(
         new MergeFilterAndRelation(),
         new MergeAggAndIndexScan(),
         new MergeAggAndRelation()
-    ));
+    ).map(r -> (Rule<LogicalPlan>)r).collect(Collectors.toList()));
```
acarbonetto, line 35: this is messy, and we can update the Rules classes to extend `LogicalPlan` instead.

acarbonetto (changes requested):

Yury-Fridlyand:
Design review meeting notes:
1. Approach is correct, we should proceed in that direction
2. Split implementation into parts:
   1. `Optimizer` rework
   2. Reorder rules and update push down operations
   3. Add support for `limit` in paginated queries

Swiddis:
Is this PR still in progress, or should we close it as stale? Looks like it gets linked a lot from elsewhere but not actively updated

Issue #1478, opened 2023-03-27
Allow individual shards to be targeted during query execution [FEATURE]


akuzin1:
**Is your feature request related to a problem?**

In MySQL one can retrieve partition information about a table, which can later be used to target specific partitions during query execution.

The following is an example of a query that can be used to retrieve partition information of a specific table.
```
"SELECT DISTINCT partition_name FROM INFORMATION_SCHEMA.PARTITIONS WHERE TABLE_NAME = <table_name> AND TABLE_SCHEMA = <table_schema> " +
            "AND partition_name IS NOT NULL"
```
Next is an example of a query targeting a specific partition in a table.

`SELECT * FROM table PARTITION (partitionName);`

When applying this to opensearch, partitions could be treated as the equivalent of shards for our use case.

**What solution would you like?**
It would be great to be able to treat shards in Opensearch as the equivalent to MySQL Partitions and be able to query individual shards.

**What alternatives have you considered?**
We've considered generating splits based on hashing a key or tuple of keys and then modulo that against some fixed number of splits that we want to generate.

```
Example:
For example for 3 splits:

split 1:
SELECT *
FROM some_unpartitioned_table
WHERE hash(col1, col2, col3) % 10 == 0

split 2:
SELECT *
FROM some_unpartitioned_table
WHERE hash(col1, col2, col3) % 10 == 1

split 3:
SELECT *
FROM some_unpartitioned_table
WHERE hash(col1, col2, col3) % 10 == 2
```
However, it doesn't seem like there is a hashing function like that available.

Therefore, the above mentioned solution would be a great behavior to add for all sql users, to more closely mimic the behavior and syntax of MySQL.

dai-chen:
@akuzin1 Thanks for the feature request! This recall an old request previously: https://github.com/opendistro-for-elasticsearch/sql/issues/1151. @acarbonetto is adding OpenSearch meta field support to our SQL engine. I'm thinking can we support this as part of meta field work.

akuzin1:
Sounds good, that would be great to see. Is there an estimated timeline for when this feature would be released?
Thank you

dai-chen:
@acarbonetto @Yury-Fridlyand @MaxKsyunz ^^

acarbonetto:
This will be a dependency for https://github.com/opensearch-project/sql/issues/1441
The API on OpenSearch side is `_routing` which we should consider using instead of `_shard` (or both!)

acarbonetto:
[...]

Issue #277, opened 2021-11-09
[BUG] Self join doesn't work properly


Yury-Fridlyand:
**Describe the bug**
[Self join](https://www.w3schools.com/sql/sql_join_self.asp) doesn't work properly.

**To Reproduce**
```
opensearchsql> SELECT c1.`key`, c1.num1, c2.str0 FROM calcs AS c1, calcs AS c2;
fetched rows / total rows = 17/17
+----------+-----------+-----------+
| c1.key   | c1.num1   | c2.str0   |
|----------+-----------+-----------|
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
| null     | null      | null      |
+----------+-----------+-----------+
opensearchsql> SELECT c1.`key`, c1.num1, c1.str0 FROM calcs AS c1;
fetched rows / total rows = 17/17
+-------+--------+-----------------+
| key   | num1   | str0            |
|-------+--------+-----------------|
| key00 | 8.42   | FURNITURE       |
| key01 | 6.71   | FURNITURE       |
| key02 | 9.78   | OFFICE SUPPLIES |
| key03 | 7.43   | OFFICE SUPPLIES |
| key04 | 9.05   | OFFICE SUPPLIES |
| key05 | 9.38   | OFFICE SUPPLIES |
| key06 | 16.42  | OFFICE SUPPLIES |
| key07 | 11.38  | OFFICE SUPPLIES |
| key08 | 9.47   | TECHNOLOGY      |
| key09 | 12.4   | TECHNOLOGY      |
| key10 | 10.32  | TECHNOLOGY      |
| key11 | 2.47   | TECHNOLOGY      |
| key12 | 12.05  | TECHNOLOGY      |
| key13 | 10.37  | TECHNOLOGY      |
| key14 | 7.1    | TECHNOLOGY      |
| key15 | 16.81  | TECHNOLOGY      |
| key16 | 7.12   | TECHNOLOGY      |
+-------+--------+-----------------+

```

**Expected behavior**
First query should yield the same result as the second one.

chloe-zh:
[...]

PR #1771, opened 2023-06-23
Add _routing to SQL includes list (#277)

acarbonetto:
### Description
Add the `_routing` reserved word as a metafield accessible as an identifier from any opensearch index. 

Example: 
```
select _id, _index, _routing, int0 from calcs limit 5
```

Example response:
```
        [
            "1",
            "calcs",
            "[xoyIAOdqQsS74x2yKSgiEQ][calcs][0]",
            1
        ],
```
 
### Issues Resolved
Partially resolves: [sql#1478](https://github.com/opensearch-project/sql/issues/1478)
 
### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/sql/IdentifierIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/sql/IdentifierIT.java
@@ -102,16 +102,61 @@ public class IdentifierIT extends SQLIntegTestCase {
   }
 
   @Test
-  public void testMetafieldIdentifierRoutingTest() throws IOException {
-    // create an index, but the contents doesn't matter
-    String id = "12345";
-    String index = "test.routing_metafields";
-    new Index(index).addDoc("{\"age\": 30}", id);
+  public void testMetafieldIdentifierRoutingSelectTest() throws IOException {
+    // create an index, but the contents doesn't really matter
+    String index = "test.routing_select";
+    String mapping = "{\"_routing\": {\"required\": true }}";
+    new Index(index, mapping)
+        .addDocWithShardId("{\"age\": 31}", "test1", "test1")
+        .addDocWithShardId("{\"age\": 32}", "test2", "test2")
+        .addDocWithShardId("{\"age\": 33}", "test3", "test3")
+        .addDocWithShardId("{\"age\": 34}", "test4", "test4")
+        .addDocWithShardId("{\"age\": 35}", "test5", "test5")
+        .addDocWithShardId("{\"age\": 36}", "test6", "test6");
 
-    // Execute using field metadata values
+    // Execute using field metadata values filtering on the routing shard hash id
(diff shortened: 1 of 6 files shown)


Yury-Fridlyand (dismissed):
I think we need an IT to prove that filtering by `_routing` works fine, for example:
```sql
select *, _id from <index> where _routing = 'a'
select *, _id from <index> where _routing = 'b'
```
The test should propagate the data properly into different shards and get their names using REST API.

acarbonetto:
> I think we need an IT to prove that filtering by `_routing` works fine, for example:
> 
> ```sql
> select *, _id from <index> where _routing = 'a'
> select *, _id from <index> where _routing = 'b'
> ```
> 
> The test should propagate the data properly into different shards and get their names using REST API.

Right now routing shard names are randomly assigned, so I cannot filter by shard name.  Yet.  
In the following PR, I will be able to assign documents to specific shards, and will be able to filter by routing shard using exactly what you were suggesting. 

Note: this is also why a doctest with _routing doesn't work, because the documents are assigned to random shards.

Yury-Fridlyand:
Yes, but you can get shard names with REST API or CAT API

MitchellGale (approved):

integ-test/src/test/java/org/opensearch/sql/sql/IdentifierIT.java
```diff
@@ -99,6 +101,78 @@ public void testMetafieldIdentifierTest() throws IOException {
...
+    verifySchema(result,
+        schema("_id", null, "keyword"),
+        schema("_index", null, "keyword"),
+        schema("_routing", null, "keyword"));
+    assertTrue(result.getJSONArray("schema").length() == 3);
+
+    var datarows = result.getJSONArray("datarows");
+    assertEquals(1, datarows.length());
+
+    // note that _routing in the SELECT clause returns the shard
```
Yury-Fridlyand, line 172: Maybe add this phrase to docs?

Yury-Fridlyand (dismissed):

integ-test/src/test/java/org/opensearch/sql/sql/IdentifierIT.java
```diff
@@ -99,6 +101,78 @@ public void testMetafieldIdentifierTest() throws IOException {
...
+    assertTrue(datarows.getJSONArray(0).getString(2).contains("[" + index + "]"));
```
dai-chen, line 173: Should we assert doc id is "test4"?

dai-chen (approved):
Thanks for the changes!

integ-test/src/test/java/org/opensearch/sql/sql/IdentifierIT.java
acarbonetto: Updated

Yury-Fridlyand (approved):

MaxKsyunz (approved):

Issue #1722, opened 2023-06-08
[FEATURE] Add spark application


rupal-bq:
**Is your feature request related to a problem?**
This should address step 4 and 5 from below 

<img width="1262" alt="image" src="https://github.com/opensearch-project/sql/assets/56703525/da99e8db-3c79-4d92-8808-46170e096ad3">

**What solution would you like?**
Application should accept sql query, OpenSearch info and write results to index in following format
```
"stepId":"<step-id>",
"schema": "json blob",
"result": "json blob" 
```

PR #1778, opened 2023-06-26
Add spark applicationId

rupal-bq:
### Description
- Added applicationId 
 
### Issues Resolved
https://github.com/opensearch-project/sql/issues/1722
- applicationId will be used instead of stepId for non-emr spark setup
 
### Check List


Yury-Fridlyand:
Is it possible to do an IT for that?

rupal-bq:
> Is it possible to do an IT for that?

Need to check. will address that in follow up PR.

dai-chen:
Could you elaborate a little what's the difference and why this change?

rupal-bq:
> Could you elaborate a little what's the difference and why this change?

In spark connector, we will be able to use different spark clients. In case of EMR client, stepId can be used to fetch the result from opensearch index but stepId concept is specific to EMR.

So when we will use another client e.g. Livy client then we can use spark applicationId to get results from opensearch index because Livy response includes appId.
ref: https://livy.incubator.apache.org/docs/latest/rest-api.html

This is basically just for adding alternate option to stepId if it's not emr in design of https://github.com/opensearch-project/sql/issues/1722

penghuo (approved):

dai-chen (approved):

Issue #1572, opened 2023-04-21
[BUG] nested/object query with raw or csv format inserts unexpected newlines


acarbonetto:
**What is the bug?**
When using the `raw` or `csv` format output in SQL plugin, we expect each row of output to correspond with a row from the response.  

for example, from `SELECT nested(message.info), nested(comment.data) FROM nested_with_nulls`: 

Expected output: 
```
nested(message.info)|nested(comment.data)
|hh
b|aa
c|aa
c|ab
a|ab
zz|[aa, bb]
zz|[aa, bb]
|ee
a|ab
rr|[asdf, sdfg]
```

Actual output: 
```
nested(message.info)|nested(comment.data)
|hh
b|aa
c|aa
c|ab
a|ab
zz|[aa, bb
]
zz|[aa, bb
]
|ee
a|ab
rr|[asdf, sdfg
]
```

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

acarbonetto:
Note: this might be related to errors in Dashboard processing `raw` and `csv` format from the V2 engine (seems to work for the legacy engine).

Yury-Fridlyand:
Detected the same issue with `csv` format without `nested`:
**V1**
```json
POST http://localhost:9200/_plugins/_sql?format=csv
{
    "query" : "select * from account where account_number = 70",
    "legacy": 1
}
```
```csv
account_number,firstname,address,balance,gender,city,employer,state,age,email,lastname
70,Deidre,685 School Lane,38172,F,Chestnut,Netplode,GA,33,deidrethompson@netplode.com,Thompson
```
**V2**
```json
POST http://localhost:9200/_plugins/_sql?format=csv
{
    "query" : "select * from account where account_number = 70"
}
```
```
account_number,firstname,address,balance,gender,city,employer,state,age,email,lastname
70,Deidre,
685 School Lane,
38172,F,Chestnut,Netplode,GA,
33,deidrethompson@netplode.com,Thompson
```

matthewryanwells:
I've narrowed down that this bug is caused by the fact in legacy the header defines the content-type as `text/plain; charset=UTF-8` while in the new engine the content-type is `application/json; charset=UTF-8`

PR #1779, opened 2023-06-26
Fix CSV/RAW output header being application/json rather than plain/text

matthewryanwells:
### Description
When making SQL calls with `csv` or `raw` outputs the formatting would be incorrect as the header sets the format to `application/json` rather than `plain/text` like the legacy engine did.

This PR adds a format getter to both response formatters to allow it to be set correctly.
 
### Issues Resolved
https://github.com/opensearch-project/sql/issues/1572
 
### Check List


Yury-Fridlyand (approved):

dai-chen (approved):
Thanks for the fix!

acarbonetto (approved):

matthewryanwells:
> @matthewryanwells have we determined if this fixes nested and objects in the Dashboard too?

@acarbonetto I can confirm that this fixes the nested, objects, and any other issues with output as the problem was how csv and raw were being handled, rather than the specific thing being processed.

PR #1807, opened 2023-07-04
Added new datetime functions and aliases to PPL

matthewryanwells:
### Description
This PR adds multiple datetime functions and aliases to PPL.

The following aliases are added:
| Alias | Original Function |
| --- | --- |
| day_of_week | dayofweek |
| day_of_month | day |
| hour_of_day | hour |
| minute_of_hour | minute |
| second_of_minute | second |
| day_of_year | dayofyear |
| month_of_year | month |
| week_of_year | week |

The following functions are added:
- minute_of_day
- time_format

### Issues Resolved
No related issue
 
### Check List


acarbonetto (dismissed):

Yury-Fridlyand (approved):

docs/user/ppl/functions/datetime.rst
```diff
@@ -711,6 +738,30 @@ Example::
     +---------------------------------+
 
 
+DAY_OF_WEEK
+-----------
+
+Description
+>>>>>>>>>>>
+
+Usage: day_of_week(date) returns the weekday index for date (1 = Sunday, 2 = Monday, ..., 7 = Saturday).
```
GumpacG, line 747: `DAY_OF_MONTH` states that "The dates with value 0 such as '0000-00-00' or '2008-00-00' are invalid".
This is not included here. Does that mean that these values are valid for this and other functions below?

GumpacG (dismissed):

docs/user/ppl/functions/datetime.rst
matthewryanwells: I missed out on this as I based this off of the SQL documentation for these functions. I believe this isn't mentioned as having a 0 in the month/day slot is an invalid date that can't exist and doesn't work for any other datetime functions. I don't believe that that line should be added here as it would also imply that it should be added to every other description, and might be worth removing form the `DAY_OF_MONTH` function, thoughts?

GumpacG: I think it is worth including this statement to all relevant functions to avoid confusion for the users. If some functions indicate this and some don't even though it applies to these functions, users may think it is safe to assume that they can have 0s as values.

matthewryanwells: As this is a limitation of dates and would be equivalent to specifying that the month has to be 12 or less, I have decided to remove this line from all functions in the file in https://github.com/opensearch-project/sql/pull/1807/commits/cd4293df46372e8d1e36bf92807f273a04a434d6

docs/user/ppl/functions/datetime.rst
```diff
@@ -669,13 +669,13 @@ DAYOFMONTH
 Description
 >>>>>>>>>>>
 
-Usage: dayofmonth(date) extracts the day of the month for date, in the range 1 to 31. The dates with value 0 such as '0000-00-00' or '2008-00-00' are invalid.
```
MitchellGale, line 672: What happens now with zero dates? Kind of a weird old comment.

matthewryanwells: I am taking out these lines because those are invalid dates and I see no need on them being specifically mentioned, it's like saying on time functions that seconds can't be higher than 60, it is true but also implied

MitchellGale: Makes sense!

docs/user/ppl/functions/datetime.rst
```diff
@@ -1298,6 +1508,61 @@ Example::
...
+   * - Specifier
+     - Description
+   * - %f
+     - Microseconds (000000..999999)
+   * - %H
+     - Hour (00..23)
+   * - %h
+     - Hour (01..12)
+   * - %I
+     - Hour (01..12)
```
MitchellGale, line 1536: Is there a difference between `%I` and `%h`? If so, perhaps mention the difference.

MitchellGale (approved):

Yury-Fridlyand (approved):

docs/user/ppl/functions/datetime.rst
matthewryanwells: `%I` and `%h` return the same thing, they are just synonyms for each other.

MitchellGale: Cool! Thanks.

forestmvey (approved):

PR #1815, opened 2023-07-07
Guarantee datasource read api is strong consistent read

### LINKED ISSUE #1801: Compatibility with segment replication
### Summary
With 2.9.0 release, there are lot of enhancements going in for segment replication[1][2] feature (went GA in 2.7.0), we need to ensure different plugins are compatible with current state of this feature. Previously, we ran [tests](https://github.com/opensearch-project/OpenSearch/issues/6602) on plugin repos to verify this compatibility but want plugin owners to be aware of these changes so that required updates (if any) can be made. With `2.10.0` release, [remote store feature](https://github.com/opensearch-project/OpenSearch/issues/2992) is going GA which internally uses SEGMENT replication strategy only i.e. it enforces all indices to use `SEGMENT` replication strategy. So, it is important to validate plugins are compatible with segment replication feature. 

#### 1. Refresh policy behavior
1. RefreshPolicy.IMMEDIATE will only refresh primary shards but not replica shards immediately. Instead post refresh, primary will start  a round of segment replication to update the replica shard copies leading to eventual consistency.
2. RefreshPolicy.WAIT_UNTIL ensures the indexing operation is searchable in your cluster i.e. RAW (Read after write guarantee). With segment replication, this guarantee is not promised due to delay in replica shared updates from asynchronous background refreshes.

####  2. Refresh lag on replicas
With segment replication, there is inherent delay in documents to be searchable on replica shard copies. This is due to the fact that replica shard copies over data (segment) files from primary. Thus, compared to document replication, there will be on average increase in amount of time the replica shards are consistent with primaries.

#### 3. System/hidden indices support
With https://github.com/opensearch-project/OpenSearch/pull/8200, system and hidden indices are now supported with `SEGMENT` replication strategy. We need to ensure there are no bottlenecks which prevents system/hidden indices with segment replication.

### Next steps
With segment replication strong reads are not guaranteed. Thus, if the plugin needs strong reads guarantees specially as alternative to change in behavior of refresh policy and lag on replicas (point 1 and 2 above), we need to update search requests to target primary shard only. With https://github.com/opensearch-project/OpenSearch/pull/7375, core now supports primary shards only based search. Please follow [documentation](https://opensearch.org/docs/latest/api-reference/search/#the-preference-query-parameter) for examples and details

### Open questions
In case of any questions or issues, please post it in [core](https://github.com/opensearch-project/OpenSearch/issues/8211) issue

### Reference
[1] [Design](https://github.com/opensearch-project/OpenSearch/issues/2229)

[2] [Documentation](https://opensearch.org/docs/latest/tuning-your-cluster/availability-and-recovery/segment-replication/index/)

penghuo:
### Description
Will add IT later because current IT RestTestIT does not support configure nodes for each test and I do not want to configure nodes for all the IT now which will be also risky.

Create an issue for impove IT framework.https://github.com/opensearch-project/sql/issues/1816
 
### Issues Resolved
#1801
 
### Check List

### DIFF
--- a/datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
+++ b/datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
@@ -217,6 +217,8 @@ public class OpenSearchDataSourceMetadataStorage implements DataSourceMetadataSt
     searchSourceBuilder.query(query);
     searchSourceBuilder.size(DATASOURCE_QUERY_RESULT_SIZE);
     searchRequest.source(searchSourceBuilder);
+    // strongly consistent reads is requred. more info https://github.com/opensearch-project/sql/issues/1801.
+    searchRequest.preference("_primary");
     ActionFuture<SearchResponse> searchResponseActionFuture;
     try (ThreadContext.StoredContext ignored = client.threadPool().getThreadContext()
         .stashContext()) {


dai-chen (approved):

joshuali925 (approved):

rupal-bq (approved):

datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
```diff
@@ -217,6 +217,8 @@ private List<DataSourceMetadata> searchInDataSourcesIndex(QueryBuilder query) {
     searchSourceBuilder.query(query);
     searchSourceBuilder.size(DATASOURCE_QUERY_RESULT_SIZE);
     searchRequest.source(searchSourceBuilder);
+    // strongly consistent reads is requred. more info https://github.com/opensearch-project/sql/issues/1801.
+    searchRequest.preference("_primary");
```
dreamer-89, line 221: @penghuo : Is there any data around TPS on this search query ?  I am thinking if this can have any performance implications.

PR #1819, opened 2023-07-07
Provide hybrid scan setting for consistency requirement

dai-chen:
### Description

Add hybrid scan mode which covers the latest source files that haven't refreshed to Flint index yet:

1. For skipping index, this means source file unknown in skipping index will not be skipped (because we're not sure if it has the answer for a query <= this is covered in this PR
2. For covered index/MV, this means we need to union all the results from unknown files and index data
 
### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/2
 
### Check List

### DIFF
--- a/flint/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/FlintSparkSkippingFileIndex.scala
+++ b/flint/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/FlintSparkSkippingFileIndex.scala
@@ -17,10 +17,10 @@ import org.apache.spark.sql.types.StructType
  *
  * @param baseFileIndex
  *   original file index
- * @param filterByIndex
- *   pushed down filtering on index data
+ * @param queryIndex
+ *   query skipping index DF with pushed down filters
  */
-case class FlintSparkSkippingFileIndex(baseFileIndex: FileIndex, filterByIndex: DataFrame)
+case class FlintSparkSkippingFileIndex(baseFileIndex: FileIndex, queryIndex: DataFrame)
     extends FileIndex {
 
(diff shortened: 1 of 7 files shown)


flint/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/FlintSparkSkippingFileIndex.scala
```diff
@@ -48,6 +58,44 @@ case class FlintSparkSkippingFileIndex(baseFileIndex: FileIndex, filterByIndex:
...
+   *     OR [indexFilter]
+   */
+  private def selectFilesFromIndexAndSource(partitions: Seq[PartitionDirectory]): Set[String] = {
+    val sparkSession = indexScan.sparkSession
+    import sparkSession.implicits._
+
+    partitions
+      .flatMap(_.files.map(f => f.getPath.toUri.toString))
+      .toDF(FILE_PATH_COLUMN)
+      .join(indexScan, Seq(FILE_PATH_COLUMN), "left")
```
penghuo, line 78: i assume left join is expensive, is there an magic number to avoid left join, which mean scan without index? if it make sense, we can add an issue to track it at perf test stage.

penghuo (approved):

flint/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/FlintSparkSkippingFileIndex.scala
dai-chen: Sure, will cover this in our benchmark after merged. Thanks!

PR #2040, opened 2023-08-30
Add _primary preference only for segment replication enabled indices

### LINKED ISSUE #1801: Compatibility with segment replication

dreamer-89:
### Description
Adds _primary preference only for segment replication enabled indices. 
 
### Issues Resolved
#1801 
 
### Check List

### DIFF
--- a/datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
+++ b/datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
@@ -32,6 +32,7 @@ import org.opensearch.action.support.WriteRequest;
 import org.opensearch.action.update.UpdateRequest;
 import org.opensearch.action.update.UpdateResponse;
 import org.opensearch.client.Client;
+import org.opensearch.cluster.ClusterState;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.action.ActionFuture;
 import org.opensearch.common.util.concurrent.ThreadContext;
@@ -84,7 +85,7 @@ public class OpenSearchDataSourceMetadataStorage implements DataSourceMetadataSt
       createDataSourcesIndex();
       return Collections.emptyList();
     }
-    return searchInDataSourcesIndex(QueryBuilders.matchAllQuery());
+    return searchInDataSourcesIndex(this.clusterService.state(), QueryBuilders.matchAllQuery());
   }
 
   @Override
@@ -93,7 +94,7 @@ public class OpenSearchDataSourceMetadataStorage implements DataSourceMetadataSt
       createDataSourcesIndex();
       return Optional.empty();
     }
-    return searchInDataSourcesIndex(QueryBuilders.termQuery("name", datasourceName)).stream()
+    return searchInDataSourcesIndex(this.clusterService.state(), QueryBuilders.termQuery("name", datasourceName)).stream()
         .findFirst()
         .map(x -> this.encryptDecryptAuthenticationData(x, false));
   }
@@ -217,13 +218,17 @@ public class OpenSearchDataSourceMetadataStorage implements DataSourceMetadataSt
     }
   }
 
-  private List<DataSourceMetadata> searchInDataSourcesIndex(QueryBuilder query) {
+  private List<DataSourceMetadata> searchInDataSourcesIndex(ClusterState state, QueryBuilder query) {
     SearchRequest searchRequest = new SearchRequest();
     searchRequest.indices(DATASOURCE_INDEX_NAME);
     SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
     searchSourceBuilder.query(query);
     searchSourceBuilder.size(DATASOURCE_QUERY_RESULT_SIZE);
     searchRequest.source(searchSourceBuilder);
+    if (state.isSegmentReplicationEnabled(DATASOURCE_INDEX_NAME)) {
+      // https://github.com/opensearch-project/sql/issues/1801.
+      searchRequest.preference("_primary");
+    }
     ActionFuture<SearchResponse> searchResponseActionFuture;
     try (ThreadContext.StoredContext ignored =
         client.threadPool().getThreadContext().stashContext()) {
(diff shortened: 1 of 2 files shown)


datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
```diff
@@ -217,13 +220,18 @@ private void createDataSourcesIndex() {
...
-  private List<DataSourceMetadata> searchInDataSourcesIndex(QueryBuilder query) {
+  private List<DataSourceMetadata> searchInDataSourcesIndex(
+      ClusterState state, QueryBuilder query) {
     SearchRequest searchRequest = new SearchRequest();
     searchRequest.indices(DATASOURCE_INDEX_NAME);
     SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
     searchSourceBuilder.query(query);
     searchSourceBuilder.size(DATASOURCE_QUERY_RESULT_SIZE);
     searchRequest.source(searchSourceBuilder);
+    if (state.isSegmentReplicationEnabled(DATASOURCE_INDEX_NAME)) {
```
penghuo, line 231: why not add parameter in search Request? e.g. consistentRead = true. Current approach expose internal implementation detail to client. In future, in case we add any feature breaking consistent read, all the client require consistent read need to make code change accordingly.

vamsi-amazon: +1 on Peng's comment.
Also are we doing this with all the plugins? 
Now, I am really not sure is it any worth of identifying the replication strategy and set the primary preference. What is the worst that could happen if we do it for all requests? It overwhelms the primary shard only if we cross certain limit? 
By any chance do we have numbers for that?

mch2: Adding a consistency param would be cleaner here for clients/plugins and doesn't expose implementation detail or use internal apis.  However, this needs more vetting before being made api. 

We need to define the behavior here - does this simply route to primary or does it guarantee a strong read every time it is invoked? Because if its the latter we need to perform a refresh internally to guarantee even the primary is serving the latest writes irrespective of what refresh policy may have been set on the write path.  This concerns me because it changes refresh behavior where it may not be intended and have consequences.

I think what we really are asking for here is the former, which is to "read the last refreshed on data" whatever that may be at the time.

mch2: We don't have any data on this @vamsi-amazon so I agree with leaving as is.  The downside of primary preference is the request will fail if the primary is unavailable and not route to replicas that are healthy.

To alleviate some concern, I suggest we use `_primary_first` which will attempt primary first and fallback to replicas only if the node is weighed away.

dreamer-89: @penghuo : Adding a new parameter may need more discussion and product feedback. 
@vamsi-amazon : No, we don't have data around primary shard limits i.e. max request that it can support before causing instability. 

Proceeding with @mch2 suggestion here.

dreamer-89: Updated. Request for review @penghuo @vamsi-amazon @mch2

dreamer-89: @penghuo @vamsi-amazon : Please have a look. Let know if you have any concerns. 

CC @mch2

penghuo (approved):

dreamer-89:
@penghuo @vamsi-amazon : Can you please help with review and merge.

vamsi-amazon (approved):

dreamer-89:
@vamsi-amazon @penghuo : Can you please add `backport 2.x` label to have this backport to 2.x/2.10 branch

PR #2132, opened 2023-09-22
Datasource description

vamsi-amazon:
### Description
* Added description field to datasource metadata.
* Removed restriction on all fields with prefix auth. Now only username, password, access_key, secret_key are the only fields removed from GET datasource API.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/datasource/model/DataSourceMetadata.java
+++ b/core/src/main/java/org/opensearch/sql/datasource/model/DataSourceMetadata.java
@@ -11,25 +11,28 @@ import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableMap;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/opensearch/sql/datasource/model/DataSourceMetadata.java
```diff
@@ -11,26 +11,29 @@
...
 @Setter
 @AllArgsConstructor
-@NoArgsConstructor
 @EqualsAndHashCode
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class DataSourceMetadata {
 
   @JsonProperty private String name;
 
+  @JsonProperty private String description;
```
derek-ho, line 35: Does this annotation provide default getter/setter? Do not see code for that

vamsi-amazon: There is getter setter annotation above class name. JsonProperty is usually not required unless we have some customization in serialization like changing the name of the field to something other than field name/ignoring in serialization.

dai-chen (approved):
Just to confirm the BWC, old data source metadata can still be handled by this new code?

vamsi-amazon:
> Just to confirm the BWC, old data source metadata can still be handled by this new code?

Yes it can handle, BWC is because of different issue.

PR #2156, opened 2023-09-26
Fix response codes returned by JSON formatting them

derek-ho:
### Description
Fix response returned by POST and PUT. Before, front end could not parse the JSON, so used JSONFormatter to format them.
 
### Issues Resolved
Fix: https://github.com/opensearch-project/sql/issues/2157
 
### Check List


datasources/src/main/java/org/opensearch/sql/datasources/transport/TransportUpdateDataSourceAction.java
```diff
@@ -55,9 +58,15 @@ protected void doExecute(
...
+              new JsonResponseFormatter<String>(PRETTY) {
+                @Override
+                protected Object buildJsonObject(String response) {
+                  return response;
+                }
+              }.format("Updated DataSource with name " +  request.getDataSourceMetadata().getName());
       actionListener.onResponse(
-          new UpdateDataSourceActionResponse(
-              "Updated DataSource with name " + request.getDataSourceMetadata().getName()));
+          new UpdateDataSourceActionResponse(responseContent));
```
derek-ho, line 69: Delete does not need this change since there is no body - since it is using HTTP 204 response. Do we want to change that and return a message similar to POST and PUT?

datasources/src/main/java/org/opensearch/sql/datasources/transport/TransportCreateDataSourceAction.java
```diff
@@ -56,9 +60,15 @@ protected void doExecute(
     try {
       DataSourceMetadata dataSourceMetadata = request.getDataSourceMetadata();
       dataSourceService.createDataSource(dataSourceMetadata);
+      String responseContent =
+              new JsonResponseFormatter<String>(PRETTY) {
+                @Override
+                protected Object buildJsonObject(String response) {
+                  return response;
+                }
+              }.format("Created DataSource with name " + dataSourceMetadata.getName());
```
vamsi-amazon, line 69: Is this working? I feel it gives the same response. What are we trying to fix here?

vamsi-amazon: For context:: https://stackoverflow.com/questions/42981632/return-simple-string-in-web-api-producing-application-json

If we want to send a json response: we should change the response to 

{
       "status": "Created DataSource with name {name}"
}

derek-ho: Addressed in: [96f3ef0](https://github.com/opensearch-project/sql/pull/2156/commits/96f3ef0098a4fd9207c86ebe32bffb0950d55019)

derek-ho:
Screenshot 2023-09-26 at 4 11 38 PM

datasources/src/main/java/org/opensearch/sql/datasources/transport/CreateUpdateDatasourceResponse.java
```diff
@@ -0,0 +1,12 @@
+package org.opensearch.sql.datasources.transport;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class CreateUpdateDatasourceResponse {
```
penghuo, line 5: Add comments. this class is used by Create and Update Datasource response?

core/src/main/java/org/opensearch/sql/planner/physical/datasource/DataSourceTableScan.java
```diff
@@ -45,7 +45,7 @@ public String explain() {
   @Override
   public void open() {
     List<ExprValue> exprValues = new ArrayList<>();
-    Set<DataSourceMetadata> dataSourceMetadataSet = dataSourceService.getDataSourceMetadata(true);
+    Set<DataSourceMetadata> dataSourceMetadataSet = dataSourceService.getDataSourceMetadata(false);
```
penghuo, line 48: why change to false?

datasources/src/test/java/org/opensearch/sql/datasources/transport/TransportUpdateDataSourceActionTest.java
```diff
@@ -59,8 +60,18 @@ public void testDoExecute() {
...
+    JsonResponseFormatter<CreateUpdateDatasourceResponse> jsonResponseFormatter =
+        new JsonResponseFormatter<>(JsonResponseFormatter.Style.PRETTY) {
+          @Override
+          protected Object buildJsonObject(CreateUpdateDatasourceResponse response) {
+            return response;
+          }
+        };
     Assertions.assertEquals(
-        "Updated DataSource with name test_datasource", updateDataSourceActionResponse.getResult());
+        jsonResponseFormatter.format(
```
penghuo, line 72: assert value should be literal string. it is much readable.

datasources/src/main/java/org/opensearch/sql/datasources/transport/TransportCreateDataSourceAction.java
```diff
@@ -56,9 +59,16 @@ protected void doExecute(
     try {
       DataSourceMetadata dataSourceMetadata = request.getDataSourceMetadata();
       dataSourceService.createDataSource(dataSourceMetadata);
-      actionListener.onResponse(
-          new CreateDataSourceActionResponse(
-              "Created DataSource with name " + dataSourceMetadata.getName()));
+      String responseContent =
```
penghuo, line 62: just return string in json format, not schema required?

core/src/main/java/org/opensearch/sql/planner/physical/datasource/DataSourceTableScan.java
derek-ho: This change is to remove @opensearch default datasource from output of `show datasources`. Since UI cannot show/Vamsi said that UI should not show this value

datasources/src/main/java/org/opensearch/sql/datasources/transport/CreateUpdateDatasourceResponse.java
derek-ho: Removed the class - respond in your other comment

datasources/src/test/java/org/opensearch/sql/datasources/transport/TransportUpdateDataSourceActionTest.java
derek-ho: Fixed in: [767a9cb](https://github.com/opensearch-project/sql/pull/2156/commits/767a9cbd4e2465506f1a9e007d632c6ca06662c5)

datasources/src/main/java/org/opensearch/sql/datasources/transport/TransportCreateDataSourceAction.java
derek-ho: The reasoning behind this change was to have a response class and later on, if we have time we can do something like return the DataSourceMetadata that was Updated/Created. Don't think we will have time for that/I don't know what the standard is for SQL repo - but I reverted it back to just a JSON string for now - it should be fine as long as the UI can parse it.

penghuo (approved):

YANG-DB (approved):

penghuo (approved):

PR #2435, opened 2023-11-21
[WIP][Prometheus] PrometheusClient use Labels API for "describe" resultset

pjfitzgibbons:
### Description
Modify PrometheusClient to use Prometheus Labels API for PPL "describe" results, instead of Prometheus "metadata" API.

Prometheus series will only produce a "metadata" record if the Prometheus server actively scrapes the configured metrics source server.

In the case of backfilling via `promtool tsdb create-blocks-from`, this scraping event does not occur.  In the case of development or test environment, this scraping event may never occur.

All series are exposed by the Prometheus Labels API, via  endpoint `/api/v1/label/__name__/values`.   Prometheus naming conventions can be used to determine the "unit" and "type" of each series.   This is the expected internal behaviour of the Prometheus service itself.

Use case : 
Script-generated backfill files are created, with content similar to :
```
# HELP test metric
# TYPE test gauge
test_usage{instance="2"} 8.723671517308066 1640995200
# EOF
 ```

Upon ingestion of this backfill via the cli `promtool tsdb create-blocks-from /tmp/test_usage_backfill.txt`, the series is visible in the Promethus Web UI (http://localhost:9090) via the "metrics browser".
Yet the series is not visible in the Metadata API (`/api/v1/metdata`).

This result is irrespective of any configuration in `prometheus.yml` for scraping, relabeling, or other configuration feautres.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/prometheus/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
+++ b/prometheus/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
@@ -84,6 +84,17 @@ public class PrometheusClientImpl implements PrometheusClient {
   }
 
   @Override
+  public List<String> getAllSeriesLabels() throws IOException {
+    String queryUrl =
+        String.format("%s/api/v1/label/__name__/values", uri.toString().replaceAll("/$", ""));
+    logger.debug("queryUrl: " + queryUrl);
+    Request request = new Request.Builder().url(queryUrl).build();
+    Response response = this.okHttpClient.newCall(request).execute();
+    JSONObject jsonObject = readResponse(response);
+    return toListOfLabels(jsonObject.getJSONArray("data"));
+  }
+
+  @Override
   public JSONArray queryExemplars(String query, Long start, Long end) throws IOException {
     String queryUrl =
         String.format(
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: prometheus/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java [hunk 1]
```diff
@@ -83,6 +83,17 @@ public Map<String, List<MetricMetadata>> getAllMetrics() throws IOException {
     return new ObjectMapper().readValue(jsonObject.getJSONObject("data").toString(), typeRef);
   }
 
+  @Override
+  public List<String> getAllSeriesLabels() throws IOException {
```
vamsi-amazon, line 87: Can you add use case to the PR description, what are we trying to achieve?

pjfitzgibbons: Updated.  Please review.

pjfitzgibbons: I'm only looking for your review on throughput and generic functionality.
There are some hardcoded values that I will produce dynamically in a fast-follow commit on this PR

Swiddis:
Is this still in progress?

PR #2577, opened 2024-03-20
Handle EMR Exceptions in FlintCancelJob Operation

vamsi-amazon:
### Description
This PR handles different kinds of exception from EMRServerless when cancelling a job. There are frequent scenarios where the streaming job is already in cancelled state and resulting in error of ALTER INDEX command. This PR catches scenarios where the job is already in cancelled state and proceeds with the command execution.

* ResourceNotFoundException: When JobId is not present.
* ValidationException: When JobRun is not in a cancellable state.
* OtherExceptions.
 

* Fixed Datasources IT.
* Fixed Settings documentation: https://github.com/vamsi-amazon/sql/blob/emr-exception/docs/user/admin/settings.rst#pluginsqueryexecutionengineasync_queryenabled
### Issues Resolved
 
### Check List

### DIFF
--- a/spark/src/main/java/org/opensearch/sql/spark/flint/operation/FlintIndexOp.java
+++ b/spark/src/main/java/org/opensearch/sql/spark/flint/operation/FlintIndexOp.java
@@ -5,10 +5,12 @@
 
 package org.opensearch.sql.spark.flint.operation;
 
+import static org.opensearch.sql.spark.client.EmrServerlessClientImpl.GENERIC_INTERNAL_SERVER_ERROR_MESSAGE;
 import static org.opensearch.sql.spark.execution.statestore.StateStore.deleteFlintIndexState;
 import static org.opensearch.sql.spark.execution.statestore.StateStore.getFlintIndexState;
 import static org.opensearch.sql.spark.execution.statestore.StateStore.updateFlintIndexState;
 
+import com.amazonaws.services.emrserverless.model.ValidationException;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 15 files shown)


spark/src/main/java/org/opensearch/sql/spark/flint/operation/FlintIndexOp.java
```diff
@@ -145,11 +148,18 @@ public void cancelStreamingJob(
     String jobId = flintIndexStateModel.getJobId();
     try {
       emrServerlessClient.cancelJobRun(
-          flintIndexStateModel.getApplicationId(), flintIndexStateModel.getJobId());
-    } catch (IllegalArgumentException e) {
-      // handle job does not exist case.
+          flintIndexStateModel.getApplicationId(), flintIndexStateModel.getJobId(), true);
+    } catch (ResourceNotFoundException e) {
```
vamsi-amazon, line 152: This happens when the current JobId is not found?
Should we ignore this or throw exception?

spark/src/main/java/org/opensearch/sql/spark/flint/operation/FlintIndexOp.java
```diff
@@ -145,11 +148,18 @@ public void cancelStreamingJob(
...
+      // JobId Not Found Exception.
       LOG.error(e);
       return;
+    } catch (ValidationException e) {
```
vamsi-amazon, line 156: This is for scenario where the current Job is already in terminal state, where we want to ignore and proceed with the operation.

vamsi-amazon: Currently we are capturing IllegalArgumentException and returning false. 
@penghuo  @dai-chen  Need some inputs on what exceptions should we ignore and propagate others.
/ handle job does not exist case. [@penghuo  I saw this comment. You have any specific case in mind here.]

penghuo: current change make sense to me, plugin can handle validation exception
but any other exception from EMR-S plugin can not handle. throw exception to client looks good.

penghuo (approved):

seankao-az (approved):

PR #2644, opened 2024-04-30
Refactor IndexDMLHandler and related classes

ykmr1224:
### Description
* Refactor IndexDMLHandler and related classes to reduce the direct dependency to StateStore and abstract data store.
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/spark/src/main/java/org/opensearch/sql/spark/flint/IndexDMLResultStorageService.java
+++ b/spark/src/main/java/org/opensearch/sql/spark/flint/IndexDMLResultStorageService.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.spark.flint;
+
+import org.opensearch.sql.spark.dispatcher.model.IndexDMLResult;
+
+public interface IndexDMLResultStorageService {
+  IndexDMLResult createIndexDMLResult(IndexDMLResult result, String datasourceName);
+}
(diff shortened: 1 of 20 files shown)


seankao-az (approved):

spark/src/main/java/org/opensearch/sql/spark/flint/IndexDMLResultStorageService.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.spark.flint;
+
+import org.opensearch.sql.spark.dispatcher.model.IndexDMLResult;
+
+public interface IndexDMLResultStorageService {
```
vamsi-amazon, line 10: [NIT]: can you add description of class.

vamsi-amazon (approved):

spark/src/main/java/org/opensearch/sql/spark/flint/IndexDMLResultStorageService.java
ykmr1224: In my opinion, most of the Java doc comments in this repository are not adding value, and we should not put such comment to reduce visual noise. I think we want to add better comment on IndexDMLResult or IndexDMLHandler so that it explains the use case better. I can do that in later PRs.

vamsi-amazon: Adding Javadoc is part of PR checklist. You can add in a different PR. 

In opensource, generally other projects maintain good Java Documentation. we have missed these earlier for meeting the release timelines. The documentation should be actually good enough for external contributor to understand from looking at classes itself. 
Since we are not in a hurry now, we should add those.

PR #2701, opened 2024-05-31
Fix compiling error by adding meta fields checker before 7DOT8 in legacy SQL

LantaoJin:
### Description
Fix compiling error by adding meta fields checker before 7DOT8 in legacy SQL
 
### Issues Resolved
https://github.com/opensearch-project/sql/issues/2699
 
### Check List

### DIFF
--- a/legacy/src/main/java/org/opensearch/sql/legacy/utils/Util.java
+++ b/legacy/src/main/java/org/opensearch/sql/legacy/utils/Util.java
@@ -24,6 +24,8 @@ import com.alibaba.druid.sql.ast.statement.SQLUnionQuery;
 import com.alibaba.druid.sql.parser.ParserException;
 import com.alibaba.druid.sql.parser.SQLExprParser;
 import com.alibaba.druid.sql.parser.Token;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


LantaoJin:
Rerunning failed CI

ykmr1224:
The original comment for the deleted constant mentions that `isMetadataField` should be used instead. Did you seek the possibility to use that? I think the background is that other plugin could define metadata field, and usage of constant is not the best way.

legacy/src/main/java/org/opensearch/sql/legacy/utils/Util.java
```diff
@@ -41,6 +43,19 @@ public class Util {
 
   public static final String NESTED_JOIN_TYPE = "NestedJoinType";
 
+  public static final Set<String> META_FIELDS_BEFORE_7DOT8 =
```
ykmr1224, line 46: 1. Can we wrap this with a static method so we can change implementation easily?
2. If we go ahead with this approach, please add comment regarding this is not ideal solution.
3. What does the suffix `7D0T8` mean?

legacy/src/main/java/org/opensearch/sql/legacy/utils/Util.java
```diff
@@ -41,6 +43,19 @@ public class Util {
...
+      Collections.unmodifiableSet(
```
ykmr1224, line 47: nit: It's simpler to use ImmutableSet.of

legacy/src/main/java/org/opensearch/sql/legacy/utils/Util.java
cwperks: 7DOT8 is referring to ES 7.8.

"_ttl" and "_timestamp" were removed in ES6

"_type" was renamed to "_nested_path" in OS 2: https://github.com/opensearch-project/OpenSearch/pull/3196

"_size" is not a built-in metadata mapper, it comes from the mapper-size plugin which is not installed by default. Using mapperService.isMetadataField("_size") should return true if the mapper-size plugin is installed in the cluster.

The other values in this list still exist and are built-in metadata fields so mapperService.isMetadataField(<field>) should return true.

ykmr1224:
Revert commit PR is raised and we can be unblocked once that is merged.

LantaoJin:
Yep, revert PR merged, close this one.

Issue #2603, opened 2024-04-03
[FEATURE] Move from scroll to point in time.


vamsimanohar:
**Is your feature request related to a problem?**
Today, we are leveraging scroll api at different places in SQL Plugin. The ask from this issue is to use PIT instead of scroll.

saratvemulapalli:
SQL Plugin relies on scroll API support for : 
- Pagination
- Joins

PR #2703, opened 2024-06-03
Add pit for join queries

rupal-bq:
### Description
- Added option to use search after with pit instead of scroll api for join queries
 
### Issues Resolved
 
### Check List


legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
```diff
@@ -0,0 +1,61 @@
...
+        createPitRequest,
+        new ActionListener<>() {
+          @Override
+          public void onResponse(CreatePitResponse createPitResponse) {
+            pitId = createPitResponse.getId();
+          }
+
+          @Override
+          public void onFailure(Exception e) {
+            LOG.error("Error occurred while creating PIT" + e);
```
ykmr1224, line 36: Should we pass exception as second parameter instead? (I think it won't log stack trace if we do `"..." + e`)
```
LOG.error("Error occurred while creating PIT", e);
```

integ-test/src/test/java/org/opensearch/sql/legacy/JoinIT.java
```diff
@@ -288,6 +288,7 @@ public void hintMultiSearchCanRunFewTimesNL() throws IOException {
     Assert.assertThat(hits.length(), equalTo(42));
   }
 
+  @Ignore
```
ykmr1224, line 291: Should we put comment to indicate why it is ignored? (or maybe TODO comment?)

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/HashJoinElasticExecutor.java
```diff
@@ -124,16 +130,26 @@ private List<SearchHit> createCombinedResults(TableInJoinRequestBuilder secondTa
...
-      searchResponse =
-          secondTableRequest
-              .getRequestBuilder()
-              .setScroll(new TimeValue(60000))
-              .setSize(MAX_RESULTS_ON_ONE_FETCH)
-              .get();
+      SearchRequestBuilder request =
+          secondTableRequest.getRequestBuilder().setSize(MAX_RESULTS_ON_ONE_FETCH);
+
+      if (paginationWithSearchAfter) {
```
ykmr1224, line 143: I am a bit concerned to put this branching logic in many places since it reduces maintainability. Please consider refactoring if possible.

ykmr1224 (commented):
Will you add unit test to cover the new logic?

rupal-bq:
> Will you add unit test to cover the new logic?

Existing logic doesn't have unit tests so I didn't add to cover new logic also.

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/HashJoinElasticExecutor.java
```diff
@@ -311,13 +341,32 @@ private List<SearchHit> scrollTillLimit(
...
+        }
+        response = request.get();
+      } else {
+        response =
+            client
+                .prepareSearchScroll(response.getScrollId())
+                .setScroll(new TimeValue(600000))
+                .execute()
+                .actionGet();
+      }
```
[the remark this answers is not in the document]
rupal-bq: done

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/NestedLoopsElasticExecutor.java
```diff
@@ -111,11 +116,41 @@ protected List<SearchHit> innerRun() throws SqlParseException {
...
+            } else {
+              request.setSize(MAX_RESULTS_ON_ONE_FETCH);
+            }
+
+            firstTableResponse = request.get();
+          } else {
+            firstTableResponse =
+                client
+                    .prepareSearchScroll(firstTableResponse.getScrollId())
+                    .setScroll(new TimeValue(600000))
```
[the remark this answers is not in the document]
rupal-bq: used SQL_CURSOR_KEEP_ALIVE

integ-test/src/test/java/org/opensearch/sql/legacy/JoinIT.java
rupal-bq: added comment

legacy/src/main/java/org/opensearch/sql/legacy/executor/SearchRequestHandler.java
ykmr1224, line 1: Is this file used?

rupal-bq: no I already deleted it in following comimit

ykmr1224:
> > Will you add unit test to cover the new logic?
> 
> Existing logic doesn't have unit tests so I didn't add to cover new logic also.

I'm not sure the consensus in this repository, but I'd generally recommend adding unit test at least covering new logic.

ykmr1224 (dismissed):

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class PointInTimeHandler {
+  private Client client;
+  @Getter private String pitId;
+  private static final Logger LOG = LogManager.getLogger();
+
+  public PointInTimeHandler(Client client, String[] indices) {
+    this.client = client;
+
+    CreatePitRequest createPitRequest = new CreatePitRequest(new TimeValue(600000), false, indices);
```
dai-chen, line 26: Should we use `plugins.sql.cursor.keep_alive` instead of `6000000` here?

integ-test/src/test/java/org/opensearch/sql/legacy/JoinIT.java
```diff
@@ -455,7 +457,7 @@ public void joinParseCheckSelectedFieldsSplitNLConditionOrderGT() throws IOExcep
             "SELECT /*! USE_NL*/ a.firstname, a.lastname, a.gender, d.firstname, d.age  FROM %s a"
                 + " JOIN %s d on a.age < d.age WHERE (d.firstname = 'Lynn' OR d.firstname ="
                 + " 'Obrien') AND a.firstname = 'Mcgee'",
-            TEST_INDEX_PEOPLE,
+            TEST_INDEX_PEOPLE2,
```
dai-chen, line 460: what's the difference between people and people2 test index?

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
```diff
@@ -23,6 +23,7 @@ public enum Key {
     SQL_SLOWLOG("plugins.sql.slowlog"),
     SQL_CURSOR_KEEP_ALIVE("plugins.sql.cursor.keep_alive"),
     SQL_DELETE_ENABLED("plugins.sql.delete.enabled"),
+    SQL_PAGINATION_API_SEARCH_AFTER("plugins.sql.pagination.api"),
```
dai-chen, line 26: So we're not switching to PIT completely but support both PIT and Scroll?

legacy/src/main/java/org/opensearch/sql/legacy/executor/QueryActionElasticExecutor.java
```diff
@@ -45,6 +45,7 @@ public static SearchHits executeJoinSearchAction(
     ElasticJoinExecutor executor =
         ElasticJoinExecutor.createJoinExecutor(client, joinRequestBuilder);
     executor.run();
+    joinQueryAction.getPointInTimeHandler().deletePointInTime();
```
dai-chen, line 48: Following previous comment, if we support both, here how we know if the JOIN is running with PIT or not?

legacy/src/main/java/org/opensearch/sql/legacy/query/join/OpenSearchJoinQueryAction.java
```diff
@@ -34,9 +37,14 @@ public SqlElasticRequestBuilder explain() throws SqlParseException {
     JoinRequestBuilder requestBuilder = createSpecificBuilder();
     fillBasicJoinRequestBuilder(requestBuilder);
     fillSpecificRequestBuilder(requestBuilder);
+    updateRequestWithPitId(requestBuilder);
```
dai-chen, line 40: As I understand, PIT will be created in PIT handler. We need to make sure it won't be created until execution, ex. EXPLAIN shouldn't create a PIT?

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -256,23 +261,44 @@ protected void updateMetaSearchResults(SearchResponse searchResponse) {
...
   }
 
-  protected SearchResponse scrollOneTimeWithMax(
-      Client client, TableInJoinRequestBuilder tableRequest) {
-    SearchRequestBuilder scrollRequest =
-        tableRequest
-            .getRequestBuilder()
-            .setScroll(new TimeValue(60000))
-            .setSize(MAX_RESULTS_ON_ONE_FETCH);
+  public SearchResponse getResponseWithHits(
```
dai-chen, line 264: For all executor changes around this, do you know if we have enough IT to cover?

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
rupal-bq: Current plan is to add for all pagination then remove scroll after performance comparison.

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
rupal-bq: We have total 68 IT for join, 5 are ignored. Test covers setting scroll page size to 2 to test scrolling which will replaced by search_after with pit call so should be good but I am not sure how to check if it's enough

integ-test/src/test/java/org/opensearch/sql/legacy/JoinIT.java
rupal-bq: only name is different but when I use people index, I get hits from both index.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
rupal-bq: yes. thanks for catching this.

legacy/src/main/java/org/opensearch/sql/legacy/executor/QueryActionElasticExecutor.java
rupal-bq: added condition to check

legacy/src/main/java/org/opensearch/sql/legacy/query/join/OpenSearchJoinQueryAction.java
rupal-bq: That make sense. RequestBuilder is filled here and I need pit id in request, that's why I added. Checking what is alternate option, if you have any suggestion on how to get pit id if not here, please comment.

rupal-bq: Added pit to request before creating executor. Can you please take another look?

dai-chen: I still feel we shouldn't create PIT in request builder. Could you check where we created Scroll context previously? I think request builder should just build request for later execution. Probably that's why there is no dependency on `Client` in request builder.

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
dai-chen: So we will raise another PR for this after performance test?

dai-chen (commented):
1 major comment about where we create PIT. Please check if how we did that for Scroll.

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
rupal-bq: yes I was planning to add search_after for all queries (join, multi query, pagination) then a separate PR to remove scroll after performance test.

rupal-bq:
> 1 major comment about where we create PIT. Please check if how we did that for Scroll.

We don't have PIT support for Scroll in sql. Looks like PIT was added from v2.4 (ref: https://opensearch.org/blog/point-in-time/) so when join queries were implemented, PIT didn't exist.

legacy/src/main/java/org/opensearch/sql/legacy/query/join/OpenSearchJoinQueryAction.java
rupal-bq: SQL plugin rebuild the context on the fly, the scroll context will be handled natively by OpenSearch. 

(Please see **_3.6.3.3 Context Rebuild_** from  https://github.com/opensearch-project/sql/blob/main/docs/dev/opensearch-pagination.md#36-different-approaches-for-different-sql-queries )


For scroll, we only needed timeout for first request and scroll_id (got this from previous response)
for subsequent request. Only one request was enough to get datarows.
But with PIT, we need extra request to create PIT and that pitId from response is required in every search request that's why added in request builder.

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/NestedLoopsElasticExecutor.java
```diff
@@ -49,6 +48,8 @@ public NestedLoopsElasticExecutor(Client client, NestedLoopsElasticRequestBuilde
 
   @Override
   protected List<SearchHit> innerRun() throws SqlParseException {
+    createPointInTimeHandler(client, nestedLoopsRequest);
+    pit.create();
```
dai-chen, line 52: Instead of doing this in each subclass, we can do it in parent class's `run()` with try-finally to make sure PIT will be released?

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/NestedLoopsElasticExecutor.java
```diff
@@ -111,16 +112,25 @@ protected List<SearchHit> innerRun() throws SqlParseException {
           if (!BackOffRetryStrategy.isHealthy()) {
             throw new IllegalStateException("Memory circuit is broken");
           }
-          firstTableResponse =
-              client
-                  .prepareSearchScroll(firstTableResponse.getScrollId())
-                  .setScroll(new TimeValue(600000))
-                  .get();
+          Integer hintLimit = nestedLoopsRequest.getFirstTable().getHintLimit();
```
dai-chen, line 115: For such change in legacy code, could you add comment in code to explain why?

rupal-bq: added comment.

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/NestedLoopsElasticExecutor.java
rupal-bq: addressed this in https://github.com/opensearch-project/sql/pull/2703/commits/617d1dbab3875636e522c522f99e42691d587ae1

dai-chen (approved):
Thanks for the changes!

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.action.search.CreatePitRequest;
+import org.opensearch.action.search.CreatePitResponse;
+import org.opensearch.action.search.DeletePitRequest;
+import org.opensearch.client.Client;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.sql.legacy.esdomain.LocalClusterState;
+
+public class PointInTimeHandler {
```
vamsi-amazon, line 20: Can we write the classes to interface? 
Also add java documentation to the class

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
```diff
@@ -0,0 +1,66 @@
...
+  private String[] indices;
+  @Getter private String pitId;
+  private static final Logger LOG = LogManager.getLogger();
+
+  public PointInTimeHandler(Client client, String[] indices) {
+    this.client = client;
+    this.indices = indices;
+  }
+
+  public void create() {
```
vamsi-amazon, line 31: Unit tests??

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -256,23 +280,47 @@ protected void updateMetaSearchResults(SearchResponse searchResponse) {
...
+    SearchRequestBuilder request = tableRequest.getRequestBuilder().setSize(size);
+    SearchResponse responseWithHits;
+    if (LocalClusterState.state().getSettingValue(SQL_PAGINATION_API_SEARCH_AFTER)) {
+      // Set sort field for search_after
+      boolean ordered = tableRequest.getOriginalSelect().isOrderdSelect();
+      if (!ordered) {
+        request.addSort(FieldSortBuilder.DOC_FIELD_NAME, SortOrder.ASC);
+      }
+      // Set PIT
+      request.setPointInTime(new PointInTimeBuilder(pit.getPitId()));
```
vamsi-amazon, line 295: what if pit request didn't complete by the time code execution reaches this line?

vamsi-amazon: As commented in PointInTimeHandler, we are actually listening to the response in listener instead of waiting.

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
rupal-bq: Current default setting is true for using PIT with search after. After benchmark, we can decide if this needs to be changed. We can remove this while removing scroll.

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
rupal-bq: Tried printing current thread ID in run and while getting id from pitResponse, thread id is same so it's waiting. Looks like pit request will complete.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
rupal-bq: added unit test for handler

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
```diff
@@ -0,0 +1,66 @@
...
+          @Override
+          public void onFailure(Exception e) {
+            LOG.error("Error occurred while creating PIT", e);
+          }
+        });
+  }
+
+  public void delete() {
+    DeletePitRequest deletePitRequest = new DeletePitRequest(pitId);
+    client.deletePits(
```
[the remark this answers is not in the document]
rupal-bq: no

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
rupal-bq: done

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import org.opensearch.action.search.CreatePitRequest;
+import org.opensearch.action.search.CreatePitResponse;
+import org.opensearch.action.search.DeletePitRequest;
+import org.opensearch.action.search.DeletePitResponse;
+import org.opensearch.client.Client;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.sql.legacy.esdomain.LocalClusterState;
+
+/** Handler for Point In Time */
+public class PointInTimeHandler implements PIT {
```
vamsi-amazon, line 22: Can we name PointInTimeHandler and PointInTimeHandlerImpl?

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
vamsi-amazon: Ok. Strange if it is working..I am good but I am still not sure...can you add logs somewhere.

rupal-bq: Added logs in pit handler impl

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandler.java
rupal-bq: done

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
vamsi-amazon: I actually mean thread id logs to comment or description. I am fine anyways.

vamsi-amazon (approved):

Issue #2719, opened 2024-06-06
[FEATURE] Support flexible grouping order in aggregation including span


xinyual:
**Is your feature request related to a problem?**
When query using PPL, we find 
`source=opensearch_dashboards_sample_data_logs |stats COUNT() as cnt by span(timestamp, 1d), response` could work but `source=opensearch_dashboards_sample_data_logs |stats COUNT() as cnt by response， span(timestamp, 1d)` fails. It is unreasonable because the order should not influence the execution result. It works well if `span` is not included

**What solution would you like?**
Support flexible order when including   `span`

LantaoJin:
Span will always be the first grouping expression in grouping list in current implementation.
Microsoft Kusto Query Language (KQL) has similar behavior (You can try [here] (https://dataexplorer.azure.com/clusters/help/databases/Samples)) with below quoted statements.
>StormEvents 
| summarize avg(DamageProperty) by bin(StartTime, 1d), State, EventType

and
>StormEvents 
| summarize avg(DamageProperty) by State, EventType, bin(StartTime, 1d)

both use `StartTime` bin as the first grouping key in results. But all queries could be executed.

PPL query
``` 
source=opensearch_dashboards_sample_data_logs | stats COUNT() as cnt by response, span(timestamp, 1d)
```
fails with `SyntaxCheckException` that seems not make sense.

LantaoJin:
The [stats syntax](https://opensearch.org/docs/latest/search-plugins/sql/ppl/functions/#syntax-6) in our official document doesn‘t state that `span` must be the first expression of `by-clause`. But its order is declared in `stats.rst` as a design from https://github.com/opensearch-project/sql/pull/417. https://github.com/opensearch-project/sql/blob/3ab785179db593a3999efffb347d897ea8364853/docs/user/ppl/cmd/stats.rst?plain=1#L42
Based on current implementation, `span` is a [dedicated member](https://github.com/opensearch-project/sql/blob/3ab785179db593a3999efffb347d897ea8364853/core/src/main/java/org/opensearch/sql/ast/tree/Aggregation.java#L50) of `Aggregation`(UnresolvedPlan), rather than a part of `groupExprList`. The `span` member variable will be filled into the first grouping key when converting to `LogicalAggregation`.
Should we support this enhancement? Any thoughts @penghuo ?

dblock:
[...]

PR #2720, opened 2024-06-06
Span in PPL statsByClause could be specified after fields

LantaoJin:
### Description
Below PPL query fails with `SyntaxCheckException`
```
source=opensearch_dashboards_sample_data_logs | stats COUNT() as cnt by response, span(timestamp, 1d)
```
 This PR still keep the current behavior of `span` as the first grouping key. But allow the `span` being specified after fields in `statsByClause`. This fixing also could help on some LLM scenarios.

### Issues Resolved
https://github.com/opensearch-project/sql/issues/2719
 
### Check List

### DIFF
--- a/ppl/src/main/antlr/OpenSearchPPLParser.g4
+++ b/ppl/src/main/antlr/OpenSearchPPLParser.g4
@@ -188,6 +188,7 @@ statsByClause
    : BY fieldList
    | BY bySpanClause
    | BY bySpanClause COMMA fieldList
+   | BY fieldList COMMA bySpanClause
    ;
 
 bySpanClause
--- a/docs/user/ppl/cmd/stats.rst
+++ b/docs/user/ppl/cmd/stats.rst
@@ -43,7 +43,7 @@ stats <aggregation>... [by-clause]
  * Description: The by clause could be the fields and expressions like scalar functions and aggregation functions. Besides, the span clause can be used to split specific field into buckets in the same interval, the stats then does the aggregation by these span buckets.
  * Default: If no <by-clause> is specified, the stats command returns only one row, which is the aggregation over the entire result set.
 
-* span-expression: optional.
+* span-expression: optional, at most one.
 
  * Syntax: span(field_expr, interval_expr)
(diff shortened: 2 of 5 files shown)


ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -188,6 +188,7 @@ statsByClause
    : BY fieldList
    | BY bySpanClause
    | BY bySpanClause COMMA fieldList
+   | BY fieldList COMMA bySpanClause
```
dai-chen, line 191: Is span expression mixed in field list also valid case?

vamsi-amazon: From the changes it looks we are missing that.
```
: (fieldExpression | bySpanClause) (COMMA (fieldExpression | bySpanClause))*
```

Something like this would actually be a good idea or definining a new fieldList.

dai-chen: yeah, if this is valid, probably we should treat span expression as regular expression and do other validation outside ANTLR. Will approve now. Please check if we need to change this in future. Thanks!

dai-chen (approved):

ppl/src/main/antlr/OpenSearchPPLParser.g4
LantaoJin: > probably we should treat span expression as regular expression 

The reason separating Span from field list AFAIK because that grouping by more than one time-span seems meaningless. And there is one span expression in Aggregation currently: https://github.com/opensearch-project/sql/blob/b2403ca4fa1bbba2a1ac8827b6f7aeefe48f9f32/core/src/main/java/org/opensearch/sql/ast/tree/Aggregation.java#L28

In general, it doesn't make sense that grouping by a 10-minutes span and 5-minutes span together.
But I tried in SPL, query group by two time span could be executed:
```
sourcetype=access_* status=200 action=purchase | bin span=10m _time as 10m_span | bin span=5m _time as 5m_span | stats count by 10m_span, 5m_span, clientip
```
Yes, we could refactor in future.

vamsi-amazon (approved):

PR #2723, opened 2024-06-06
Added Setting to Toggle Data Source Management Code Paths

fddattal:
### Description

This commit adds a setting that toggles data source management code paths.
It will disable explicit APIs and internal code paths that implicitly create data sources.
 
### Issues Resolved

https://github.com/opensearch-project/OpenSearch/issues/13274
 
### Check List

### DIFF
--- a/datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
+++ b/datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
@@ -42,11 +42,13 @@ import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.index.query.QueryBuilders;
 import org.opensearch.search.SearchHit;
 import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.sql.common.setting.Settings;
 import org.opensearch.sql.datasource.model.DataSourceMetadata;
 import org.opensearch.sql.datasources.encryptor.Encryptor;
 import org.opensearch.sql.datasources.exceptions.DataSourceNotFoundException;
 import org.opensearch.sql.datasources.service.DataSourceMetadataStorage;
 import org.opensearch.sql.datasources.utils.XContentParserUtils;
+import org.opensearch.sql.opensearch.setting.OpenSearchSettings;
 
 public class OpenSearchDataSourceMetadataStorage implements DataSourceMetadataStorage {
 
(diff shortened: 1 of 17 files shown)


docs/user/admin/settings.rst
```diff
@@ -630,3 +630,84 @@ Request ::
...
       }
     }
+
+plugins.query.datasources.enabled
+=================================
+
+Description
+-----------
+
+This setting controls whether datasources are enabled.
```
penghuo, line 640: does this setiting impact _async_query API? if plugins.query.datasources.enabled=false, _async_query will fail?

datasources/src/main/java/org/opensearch/sql/datasources/storage/OpenSearchDataSourceMetadataStorage.java
```diff
@@ -71,14 +74,21 @@ public class OpenSearchDataSourceMetadataStorage implements DataSourceMetadataSt
...
     this.client = client;
     this.clusterService = clusterService;
     this.encryptor = encryptor;
+    this.settings = settings;
   }
 
   @Override
   public List<DataSourceMetadata> getDataSourceMetadata() {
+    if (!isEnabled()) {
+      return Collections.emptyList();
```
penghuo, line 90: throw exception?

fddattal: Hey Peng, thanks for the comment!

There are a few places in the code which call `getDataSourceMetadata()` and `getDataSourceMetadata(String)` which I felt it would be better to have them noop and continue to work instead of failing with an exception.

In particular here:
1. When performing a SQL query such as `SELECT * FROM test.index` the result should be a 4xx "Data Source test does not exist" -> https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/analysis/Analyzer.java#L153
2. When performing a PPL query such as `show datasources` the result should be a 2xx with an empty list -> https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/planner/physical/datasource/DataSourceTableScan.java#L48

For direct calls to the data source APIs, I agree there should be an exception thrown, which this PR does.

docs/user/admin/settings.rst
fddattal: That's a good question. I did not test this case. I will add an integ test to validate the behavior.

fddattal: Update : I've added additional checks in the async query code path to validate that this setting is also enabled.
This PR also contains integ tests for that case as well.

penghuo (dismissed):
Thx!

rupal-bq (dismissed):
Thanks for the changes!

penghuo (approved):
Thx

PR #2733, opened 2024-06-10
Scaffold async-query-core and async-query module

ykmr1224:
### Description
- Scaffold async-query-core and async-query module.
  - async-query-core will include the core logic async-query without OpenSearch specific implementations
  - async-query will include OpenSearch specific part such as data accessor using OpenSearch index.
- There are dummy classes and its tests to verify the dependency/build works. Those will be deleted when actual classes are moved from `spark` module in the next PR. (separated it to reduce the size of PR)
 
### Issues Resolved
n/a
 
### Check List

### DIFF
--- a/async-query-core/build.gradle
+++ b/async-query-core/build.gradle
@@ -1,0 +1,137 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+plugins {
+    id 'java-library'
+    id "io.freefair.lombok"
+    id 'jacoco'
+    id 'antlr'
+    id 'com.diffplug.spotless' version '6.22.0'
+    id 'com.github.johnrengelman.shadow'
+}
+
+repositories {
+    mavenCentral()
+}
+
+tasks.register('downloadG4Files', Exec) {
+    description = 'Download remote .g4 files from GitHub'
+
+    executable 'curl'
+
+    args '-o', 'src/main/antlr/FlintSparkSqlExtensions.g4', 'https://raw.githubusercontent.com/opensearch-project/opensearch-spark/main/flint-spark-integration/src/main/antlr4/FlintSparkSqlExtensions.g4'
+    args '-o', 'src/main/antlr/SparkSqlBase.g4', 'https://raw.githubusercontent.com/opensearch-project/opensearch-spark/main/flint-spark-integration/src/main/antlr4/SparkSqlBase.g4'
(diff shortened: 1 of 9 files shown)


async-query-core/src/main/java/org/opensearch/sql/asyncquery/Dummy.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.asyncquery;
+
+// This is a dummy class for scaffolding and should be deleted later
+public class Dummy {
```
vamsi-amazon, line 9: Do we really need a commit for this?

async-query-core/build.gradle
```diff
@@ -0,0 +1,137 @@
...
+    implementation group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-cbor', version: "${versions.jackson}"
+    implementation 'com.google.code.gson:gson:2.8.9'
+
+    testImplementation(platform("org.junit:junit-bom:5.9.3"))
+
+    testCompileOnly('org.junit.jupiter:junit-jupiter')
+    testImplementation group: 'org.mockito', name: 'mockito-core', version: '5.7.0'
+    testImplementation group: 'org.mockito', name: 'mockito-junit-jupiter', version: '5.7.0'
+
+    testCompileOnly('junit:junit:4.13.1') {
```
vamsi-amazon, line 59: Can you copy this form another module. We need this for making opnesearchTest class work in spark module. Not sure if this is required in this module.

async-query-core/src/main/java/org/opensearch/sql/asyncquery/Dummy.java
ykmr1224: If we move something, the build would break since something is depending on that. There might be a better way to verify the dependency works, but I thought it is the simplest way.

async-query-core/build.gradle
ykmr1224: Seems there are some usage of junit4 APIs outside of opensearchTest classes. I'll eliminate them after the restructure and remove this dependency.

vamsi-amazon (approved):

async-query-core/src/main/java/org/opensearch/sql/asyncquery/Dummy.java
vamsi-amazon: I am ok even if we don't verify if the dependency works in the first PR and we can verify when the dependency actually comes out.

rupal-bq (approved):

Issue #2689, opened 2024-05-22
[BUG] DROP INDEX cancel FlintREPL job


penghuo:
**What is the bug?**
User submit DROP INDEX query, plugin cancel FlintREPL job. The imapct is if user submit query in same session, submitted query will in waiting state forever.

**How can one reproduce the bug?**
* T1: submit CREATE SKIPPING INDEX query with auto_refresh=false
* T2: submit DROP INDEX query,  make sure (T2-T1) < 3mins

unexpected result is
* FlintREPL EMR-S job is cancelled, indexState is REFRESHING

**What is the expected behavior?**
FlintREPL EMR-S job should not been cancelled.

**Do you have any additional context?**
[In SQL Plugin 2.13](https://github.com/opensearch-project/sql/blob/main/spark/src/main/java/org/opensearch/sql/spark/flint/operation/FlintIndexOpDrop.java#L28), when user submit DROP INDEX query, plugin cancel EMR-S job even index state is EMTPY / ACTIVE. 

```
  public boolean validate(FlintIndexState state) {
    return state == FlintIndexState.REFRESHING
        || state == FlintIndexState.EMPTY
        || state == FlintIndexState.ACTIVE
        || state == FlintIndexState.CREATING;
  }
```

**Suggested solutions**
Alt-1. For CREATE INDEX, we could submit FlintJob instead of FlintREPL. 
Alt-2. Prevalidate indexState, only cancel Job when it is in Refreshing state.

seankao-az:
Before refactoring the DML handler and FlintIndexOp, DROP index involved 2 steps: Cancel op and Delete op, and each has their own validation. ([cancel](https://github.com/opensearch-project/sql/blob/60058ce1de6db0c6c2278a2382b674ff6e1eb78c/spark/src/main/java/org/opensearch/sql/spark/flint/operation/FlintIndexOpCancel.java), [delete](https://github.com/opensearch-project/sql/blob/60058ce1de6db0c6c2278a2382b674ff6e1eb78c/spark/src/main/java/org/opensearch/sql/spark/flint/operation/FlintIndexOpDelete.java))
After refactoring, the two steps are merged and shared one same validation logic, which is incorrect.

Of the two solution proposed, Alt-1 using FlintJob has the disadvantage of sacrificing query latency. Executing the create index statement in FlintREPL will be way faster.
Alt-2 seems better. We can validate the index state to be Refreshing state before cancel.

PR #2734, opened 2024-06-10
Handle create index with batch FlintJob

seankao-az:
### Description
Use batch FlintJob to handle create index query. It's already used for streaming query. This PR makes creating auto_refresh=false index use FlintJob as well.
This would resolve the bug where dropping a manual refresh index could result in cancellation of the FlintREPL job that executed the create index statement.

Alongside this change, also make the RefreshQueryHandler borrow from lease, instead of BatchQueryHandler borrowing, because the concurrent refresh job limit shouldn't apply to batch jobs in general. The behavior for refresh query would remain the same.
Making the change in this same PR because prior to this PR, there is no query that can be used to test BatchQueryHandler.

#### Tests
Added unit tests that cover the code path for dispatching to IndexDMLHandler, and create index queries.

#### End to end test
* create index auto_refresh=false -> drop index
  - [x] index create should be a batch FlintJob
  - [x] drop index should
    - [x] not cancel any job if the job running the create statement is completed
    - [x] cancel the batch job if it isn't completed yet
* create index auto_refresh=true -> alter index auto_refresh=false -> drop index
  - [x] index create should be a streaming FlintJob
  - [x] alter index should cancel the streaming job
  - [x] drop index should not further cancel any job
* create index auto_refresh=false -> alter index auto_refresh=true -> drop index
  - [x] index create should be a batch FlintJob
  - [x] alter index should start a streaming job
  - [x] drop index should cancel the streaming job
 
### Issues Resolved
* #2689 
 
### Check List


spark/src/main/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcher.java
```diff
@@ -79,6 +79,9 @@ private AsyncQueryHandler getQueryHandlerForFlintExtensionQuery(
       return queryHandlerFactory.getIndexDMLHandler();
     } else if (isEligibleForStreamingQuery(indexQueryDetails)) {
       return queryHandlerFactory.getStreamingQueryHandler();
+    } else if (IndexQueryActionType.CREATE.equals(indexQueryDetails.getIndexQueryActionType())) {
+      // create should be handled by batch handler
+      return queryHandlerFactory.getBatchQueryHandler();
```
vamsi-amazon, line 84: Is this handler being used anywhere today? sessionManaged.isEnabled is always true so even getDefaultAsyncQueryHandler would always return InteractiveQueryHandler. I hope there won't be any regression. 

Also RefreshQueryHandler is also using FlintJob Right?

spark/src/main/java/org/opensearch/sql/spark/dispatcher/BatchQueryHandler.java
```diff
@@ -69,8 +68,6 @@ public String cancelJob(AsyncQueryJobMetadata asyncQueryJobMetadata) {
   @Override
   public DispatchQueryResponse submit(
       DispatchQueryRequest dispatchQueryRequest, DispatchQueryContext context) {
-    leaseManager.borrow(new LeaseRequest(JobType.BATCH, dispatchQueryRequest.getDatasource()));
```
vamsi-amazon, line 72: On a side node..we have BatchQueryHandler and Batch Jobtype nomenclature. Even RefreshQueryHandler is also using JobType Batch. We should refactor this nomenclature some time.

vamsi-amazon: Removing this would make us loose control over the number of emr jobs created right. I think we should revisit the reason we introduced the lease manager..is it to control number of emr jobs or number of streaming emr jobs?

spark/src/main/java/org/opensearch/sql/spark/dispatcher/RefreshQueryHandler.java
```diff
@@ -59,6 +60,8 @@ public String cancelJob(AsyncQueryJobMetadata asyncQueryJobMetadata) {
   @Override
   public DispatchQueryResponse submit(
       DispatchQueryRequest dispatchQueryRequest, DispatchQueryContext context) {
+    leaseManager.borrow(new LeaseRequest(JobType.BATCH, dispatchQueryRequest.getDatasource()));
```
vamsi-amazon, line 63: Does this warrant any change in documentation?

vamsi-amazon:
Have we tested this on a working stack?

spark/src/main/java/org/opensearch/sql/spark/dispatcher/RefreshQueryHandler.java
dai-chen: We're moving this from parent class? The new create index job doesn't need this?

spark/src/main/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcher.java
seankao-az: > Is this handler being used anywhere today?

I believe no

> RefreshQueryHandler is also using FlintJob Right?

yes. RefreshQueryHandler extends BatchQueryHandler

spark/src/main/java/org/opensearch/sql/spark/dispatcher/BatchQueryHandler.java
seankao-az: it's for concurrent session / refresh job rule: https://github.com/opensearch-project/sql/blob/3ab785179db593a3999efffb347d897ea8364853/spark/src/main/java/org/opensearch/sql/spark/leasemanager/DefaultLeaseManager.java#L83

it shouldn't be applied to batch job in general. This PR moves this line to RefreshQueryHandler instead

spark/src/main/java/org/opensearch/sql/spark/dispatcher/RefreshQueryHandler.java
seankao-az: behavior doesn't change for refresh query handler.
also according to https://github.com/opensearch-project/sql/blob/main/docs/user/admin/settings.rst#pluginsqueryexecutionenginesparkrefresh_joblimit
it's intended to affect refresh job only

seankao-az: > The new create index job doesn't need this?

DDL shouldn't be affected by the concurrent limit. https://github.com/opensearch-project/sql/blob/3ab785179db593a3999efffb347d897ea8364853/spark/src/test/java/org/opensearch/sql/spark/asyncquery/IndexQuerySpecTest.java#L848

seankao-az:
> Have we tested this on a working stack?

yes in my test domain

spark/src/main/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcher.java
```diff
@@ -79,6 +79,9 @@ private AsyncQueryHandler getQueryHandlerForFlintExtensionQuery(
       return queryHandlerFactory.getIndexDMLHandler();
     } else if (isEligibleForStreamingQuery(indexQueryDetails)) {
       return queryHandlerFactory.getStreamingQueryHandler();
+    } else if (IndexQueryActionType.CREATE.equals(indexQueryDetails.getIndexQueryActionType())) {
+      // create should be handled by batch handler
```
ykmr1224, line 83: nit: This comment might be redundant. If we keep this, should we add more context ?

dai-chen (dismissed):
As I understand, the current approach is make sure cancel job run won't happen on REPL job. I think we can discuss later how to make index operation more clear. For example, instead of separating EMR-S job, should we avoiding cancel logic for index not in refreshing state at all?

ykmr1224 (approved):
Just minor comments. Those can be addressed in separate PR.

spark/src/test/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcherTest.java
```diff
@@ -407,6 +408,49 @@ void testDispatchIndexQuery() {
     verifyNoInteractions(flintIndexMetadataService);
   }
 
+  @Test
+  void testDispatchCreateManualRefreshIndexQuery() {
+    when(emrServerlessClientFactory.getClient()).thenReturn(emrServerlessClient);
+    HashMap<String, String> tags = new HashMap<>();
+    tags.put(DATASOURCE_TAG_KEY, "my_glue");
+    tags.put(CLUSTER_NAME_TAG_KEY, TEST_CLUSTER_NAME);
+    tags.put(JOB_TYPE_TAG_KEY, JobType.BATCH.getText());
```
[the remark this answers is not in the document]
seankao-az: this whole file use this same pattern. Can be a separate refactor PR for this for next release

spark/src/main/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcher.java
seankao-az: will add a oneliner why this should be the case

ykmr1224 (approved):

rupal-bq (approved):

PR #2737, opened 2024-06-11
Move classes from spark to async-query-core and async-query

ykmr1224:
### Description
- Move classes from spark to async-query-core and async-query to separate async-query-core as library.
- There are still some dependency from async-query-core to datasources/legacy/spark, and those will be addressed in following PRs.
- Since many classes were covered by integration tests, classes in async-query-core module are currently not covered by tests in async-query-core. Unit tests and integration tests for async-query-core will be added in following PRs. (uncovered one are excluded from coverage check right now)
 
### Issues Resolved
n/a
 
### Check List


async-query-core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
seankao-az, line 1: what is this for? and nit: add newline in the end

async-query/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
seankao-az, line 1: same

async-query-core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
ykmr1224: This is a magic spell to enable mocking on final class with Mockito. (Mockito disable it by default, and can be enabled with this configuration file. And we are doing the same for other modules.)

async-query/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
ykmr1224: same as above.

seankao-az (approved):

vamsi-amazon (approved):

PR #2753, opened 2024-06-14
Add pit for multi query

rupal-bq:
### Description
- Added option to use search after with pit instead of scroll api for multi query
 
### Issues Resolved
 
### Check List


legacy/src/main/java/org/opensearch/sql/legacy/executor/multi/MinusExecutor.java
```diff
@@ -63,45 +72,63 @@ public MinusExecutor(Client client, MultiQueryRequestBuilder builder) {
...
+        pit =
+            new PointInTimeHandlerImpl(
+                client,
+                ArrayUtils.concat(
+                    builder.getOriginalSelect(true).getIndexArr(),
+                    builder.getOriginalSelect(false).getIndexArr()));
+        pit.create();
+      }
+
+      if (this.useTermsOptimization && this.fieldsOrderFirstTable.length != 1) {
```
dai-chen, line 86: The logic starting from this become even more complex than before. Could you explain why and think about if any better approach?

legacy/src/main/java/org/opensearch/sql/legacy/executor/multi/MinusExecutor.java
```diff
@@ -49,6 +57,7 @@ public class MinusExecutor implements ElasticHitsExecutor {
   private String[] fieldsOrderFirstTable;
   private String[] fieldsOrderSecondTable;
   private String seperator;
+  private PointInTimeHandler pit;
```
dai-chen, line 60: If all executor depends on PIT, `getResponseWithHits` and others like `Client`, probably we can move all common fields to `ElasticHitsExecutor` and make it abstract class?

legacy/src/main/java/org/opensearch/sql/legacy/executor/multi/MinusExecutor.java
dai-chen: I see. I recall we have this similar logic in your previous PR? If all `ElasticHitsExecutor` has such logic, can we move it to parent class?

legacy/src/main/java/org/opensearch/sql/legacy/executor/ElasticHitsExecutor.java
```diff
@@ -5,13 +5,77 @@
...
+import org.opensearch.search.builder.PointInTimeBuilder;
+import org.opensearch.sql.legacy.domain.Select;
+import org.opensearch.sql.legacy.esdomain.LocalClusterState;
 import org.opensearch.sql.legacy.exception.SqlParseException;
+import org.opensearch.sql.legacy.pit.PointInTimeHandler;
 
 /** Created by Eliran on 21/8/2016. */
-public interface ElasticHitsExecutor {
-  void run() throws IOException, SqlParseException;
+public abstract class ElasticHitsExecutor {
```
dai-chen, line 28: np: could you add missing Javadoc?

legacy/src/main/java/org/opensearch/sql/legacy/executor/multi/MinusExecutor.java
rupal-bq: No, not all ElasticHitsExecutor has similar logic. Join and multiquery has this pit logic but there is no scrolling in UnionExecutor so it doesn't have such logic.

dai-chen (approved):
Approved. This https://github.com/opensearch-project/sql/pull/2753/files#r1669287520 is still open.

vamsi-amazon (approved):

legacy/src/main/java/org/opensearch/sql/legacy/executor/ElasticHitsExecutor.java
rupal-bq: done

Issue #2540, opened 2024-03-06
[BUG] Query with UNION operator fails with NullPointerException


StewartWBrown1:
**What is the bug?**
Present in OpenSearch 2.12.0, 2.11.0, 1.3.10
The [documentation ](https://opensearch.org/docs/latest/search-plugins/sql/sql/complex/) states that Union queries are supported. However, when attempting a simple union query, the request fails with: 

`{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "",
    "type": "NullPointerException"
  },
  "status": 500
}`

The query returns the expected result when _?format=csv_ is added to the request URL. 

**How can one reproduce the bug?**
1. Index data into OpenSearch (for this example, I am using simple novels/song data)
2. perform query involving union: 
`{
  "query": "SELECT title FROM novels UNION SELECT title FROM songs"
}`
3. NullPointerException error returned

**What is the expected behavior?**
Union operator to work as expected. In above example adding the request call to be ''http://localhost:9200/_plugins/_sql?format=csv" returns: 
```
title
Animal Farm
The Great Gatsby
Catcher In The Rye
Bohemian Rhapsody
```

**What is your host/environment?**
Present in OpenSearch 2.12, 2.11 and 1.3.10 

**Do you have any screenshots?**
Error: 
![image](https://github.com/opensearch-project/sql/assets/132061861/293d85dd-cc6f-4722-bffb-5fc496a2a5ae)

As expected (When adding ?format=csv to request URL): 
![image](https://github.com/opensearch-project/sql/assets/132061861/c24a2143-2d6c-4689-b28e-f2a11a3d8779)

**Do you have any additional context?**
Add any other context about the problem.

penghuo:
could u share the index mapping.

```
PUT index001/_doc/1
{
  "query_id" : "1"
}

PUT index002/_doc/1
{
  "query_id" : "1"
}

POST /_plugins/_sql?format=csv
{
  "query": "select query_id from index001 UNION select query_id from index002"
}

### result
query_id
1
1

```

StewartWBrown1:
@penghuo error occurs for me following same mapping you have used, but without the '?format=csv' tag at the end of the POST call, but my understanding is this shouldn't be necessary and a response should be returned without it as expected. 

```
$ curl -X PUT -k -s -H "Content-Type: application/json" http://localhost:9200/index003/_doc/1 -u stxbro:$PASSWORD -d $'{"query_id":"1"}'
[...]

PR #2757, opened 2024-06-17
Union query result should work in JDBC format

LantaoJin:
### Description
`Union` query only works with `csv` format, but without the '?format=csv' query parameter at the end of the POST call will throw `NullPointerException`. Reproduce: 
```
PUT index001/_doc/1
{
  "query_id" : "1"
}

PUT index002/_doc/1
{
  "query_id" : "1"
}

POST /_plugins/_sql
{
  "query": "select query_id from index001 UNION select query_id from index002"
}
```
Currently, the `UNION` statement is implemented in legacy engine(v1). It's a bug in legacy engine(v1).
This PR fixes this issue in legacy engine.
 
### Issues Resolved
Resolve https://github.com/opensearch-project/sql/issues/2540
 
### Check List
- ~~[ ] New functionality has been documented.~~
  - ~~[ ] New functionality has javadoc added~~
  - ~~[ ] New functionality has user manual doc added~~


integ-test/src/test/java/org/opensearch/sql/legacy/CsvFormatResponseIT.java
```diff
@@ -706,6 +706,103 @@ public void selectFunctionAsFieldTest() throws IOException {
     Assert.assertEquals(1, headers.size());
   }
 
+  @Test
+  public void unionTest() throws IOException {
+    String query =
+        String.format(
+            Locale.ROOT,
+            "SELECT firstname, lastname FROM %s LIMIT 3 "
```
penghuo, line 714: does UNION works for query with aggregation?

LantaoJin: Actually, UNION with Agg in legacy engine has bugs by my testing whatever format set. This fixing doesn't cover all existed bugs. You can try below queries.
```
POST _plugins/_sql?format=csv
{
  "query": "select customer_gender,count(type) from opensearch_dashboards_sample_data_ecommerce group by customer_gender UNION select customer_gender,count(type) from opensearch_dashboards_sample_data_ecommerce group by customer_gender"
}
POST _plugins/_sql?format=csv
{
  "query": "select count(type) from opensearch_dashboards_sample_data_ecommerce group by customer_gender UNION select count(type) from opensearch_dashboards_sample_data_ecommerce group by customer_gender"
}
```
May be we could open another ticket to fix the problems in Union with Agg? Or migrating to v2?

LantaoJin: And seems current implementation doesn't distinguish UNION and UNION ALL. A better way is to re-implement UNION in v2, PPL also requests multi-search implementation.
Again, this fixing is just for address the NPE problem of UNION query in JDBC format (such a common use case).

vamsi-amazon:
Can you add little more details on where exactly is the bug?

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/DateFieldFormatter.java
```diff
@@ -105,7 +105,7 @@ private List<String> getFormatsForColumn(String columnName) {
   private Set<String> getDateColumns(List<Schema.Column> columns) {
     return columns.stream()
         .filter(column -> column.getType().equals(Schema.Type.DATE.nameLowerCase()))
-        .map(Schema.Column::getName)
+        .map(Schema.Column::getIdentifier)
```
vamsi-amazon, line 108: Is this related to the current issue?

legacy/src/main/java/org/opensearch/sql/legacy/query/multi/MultiQueryAction.java
```diff
@@ -22,7 +23,7 @@ public class MultiQueryAction extends QueryAction {
   private MultiQuerySelect multiQuerySelect;
 
   public MultiQueryAction(Client client, MultiQuerySelect multiSelect) {
-    super(client, null);
+    super(client, new Union(multiSelect.getFirstSelect(), multiSelect.getSecondSelect()));
```
vamsi-amazon, line 26: Is this the root cause of the issue?

LantaoJin: Yes.

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/DateFieldFormatter.java
LantaoJin: Good catching! This change should be reverted. What we need to change is https://github.com/opensearch-project/sql/pull/2757/files#diff-943465139286a5010c19cab380f141e3c9bc26fb02652bedbc6a7bbec5c98cc3R46

LantaoJin:
> Can you add little more details on where exactly is the bug?

The NPE issue caused by calling `loadFromEsState` for both SELECT clauses of UNION children is missing. The fixing is adding https://github.com/opensearch-project/sql/pull/2757/files#diff-744a18b0fa02b144d9aaf91192a6594175eb0bafdca6e01adb841093c36854f5R105-R108 and https://github.com/opensearch-project/sql/pull/2757/files#diff-d81d74cf75806ee8450a8a706d7e71d21787d2cabfcb2c4f2c0e83176e711940R26

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/SelectResultSet.java
```diff
@@ -180,8 +185,12 @@ private void loadFromEsState(Query query) {
     Map<String, FieldMappingMetadata> typeMappings = mappings.get(indexName);
 
     this.indexName = this.indexName == null ? indexName : (this.indexName + "|" + indexName);
-    this.columns.addAll(
-        renameColumnWithTableAlias(query, populateColumns(query, fieldNames, typeMappings)));
+    if (isPartOfUnion(query) && !this.columns.isEmpty()) {
```
dai-chen, line 188: What's the difference between `isUnionQuery` and `isPartOfUnion`? Do we need the new `partOfUnion` field?

LantaoJin: `isUnionQuery` checks if the query type is `Union`, just similar to `isJoinQuery` does, checking if the type is `JoinSelect`.
The children type of Join is `TableOnJoinSelect`, a special type rather than a normal `Select`. But The children type of Union is a normal `Select`, unless we set a new type for Union's children, we can't identify a `Select` is a normal one or a child of Union in `loadFromEsState`, that's the reason I add `partOfUnion` in Select. Or add a new class `class UnionSubSelect extends Select`. Which one would be better?  
https://github.com/opensearch-project/sql/pull/2757/files#diff-0b55e630b732c43e3c8ea8038eb068f21a3c46bcd1744c98be4cc94bad799b84R50

LantaoJin:
> the `UNION` statement is implemented in legacy engine(v1)

Since the UNION statement is implemented in legacy engine(v1), let me close this PR now. The issue should be addressed when the enable Calcite for SQL.

Issue #2700, opened 2024-05-30
[BUG] SQL query doesn't allow different date formats without date casting


rupal-bq:
**What is the bug?**
Getting below error for user defined date format
```
"{\n  \"error\": {\n    \"reason\": \"Invalid SQL query\",\n    \"details\": \"date:03-Jan-24 in unsupported format, please use 'yyyy-MM-dd'\",\n    \"type\": \"SemanticCheckException\"\n  },\n  \"status\": 400\n}"

```

**How can one reproduce the bug?**
```
PUT testindex
{
  "mappings" : {
    "properties" :  {
      "release_date" : {
        "type" : "date",
        "format" : "dd-MMM-yy"
      }
    }
  }
}
```
```
PUT testindex/_doc/21 
{
  "release_date" : "03-Jan-24"
}
```
```
POST _plugins/_sql
{
"query":"select release_date from testindex where release_date='03-Jan-24'"
}
```

**What is the expected behavior?**
Query should work without date casting

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

manasvinibs:
We have one more similar issue reported from users when date in timestamp format in the SQL query is erroring out due to  mismatch in the timestamp format between field mapping and the request query - 

Sample query: 
```
select count(*) from ami where State = 'available' and  Public = 'true' and CallerVersion = "2016-11-15 00:00:00";
```
```
select count(*) from ami where State = 'available' and  Public = 'true' and CallerVersion < "2016-11-15 00:00:00";
```

Error -
```
[INFO] 10f0cef3-4159-4c39-84e1-f4ba8b88be97 (Bobcat-29) com.amazonaws.fleetinsightsopensearchcommons.helper.ResponseConvertor: Original SQL response : {
  "error": {
    "reason": "Error occurred in OpenSearch engine: all shards failed",
[...]

Issue #2741, opened 2024-06-12
[BUG] `str_to_date` does not work with two-digits year


LantaoJin:
**What is the bug?**
`str_to_date` function in https://opensearch.org/docs/latest/search-plugins/sql/functions/#date-and-time does not work with two-digits year ([RFC 822 date-time](http://www.w3.org/Protocols/rfc822/#z28) formatted). 

**How can one reproduce the bug?**
Make a similar query
```SQL
SELECT str_to_date('03-Jan-24', '%d-%b-%y');
```
returns
`0024-01-03 00:00:00`

**What is the expected behavior?**
Return `2024-01-03 00:00:00`

**What is your host/environment?**
 - Version: OpenSearch 2.14
 - Plugins: SQL

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

dblock:
[Catch All Triage - Attendees [1](https://github.com/dblock/), [2](https://github.com/krisfreedain), [3](https://github.com/andrross), [4](https://github.com/gkamat), [5](https://github.com/getsaurabh02)]

PR #70, opened 2021-05-19
remove opensearch.sql.engine.new.enabled setting

penghuo:
Signed-off-by: penghuo <penghuo@gmail.com>

### Description
Remove the opensearch.sql.engine.new.enabled. The opensearch.sql.engine.new.enabled setting is added and set to default true on ODFE 1.13.0.0, the major purpose is to safely publish the new engine, in any case, the customer found issue, they still have the way to rollback to the legacy engine. 
We didn't found major issue with new engine, so we decide to fullly enable it by default. If there are any case the new engine can't handle, it still fallback to the the legacy engine automatically. So, there is no impact for the customer.
 
### Issues Resolved
#58 

### To Reviewers
The PR include the following changes
1. Change the SQL entry point logic, the new engine is default now.
2. Remove the integration test against the legacy engine.
3. Ignore some IT which only work for legacy engine.
4. Remove the opensearch.sql.engine.new.enabled related docs accordingly.
 
### Check List
- [ ] New functionality includes testing.
  - [x] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #44, opened 2021-05-13
OpenSearch function support


dai-chen:
**Is your feature request related to a problem? Please describe.**
Currently both SQL (new engine) and PPL doesn't support any OpenSearch functions. We should prioritize this, especially the most common full text search and aggregate functions.

Here are the feature requests from community:

1. https://github.com/opendistro-for-elasticsearch/sql/issues/1093: Percentile function support in PPL stats

**Describe the solution you'd like**
Expose OpenSearch functions in SQL/PPL.

**Describe alternatives you've considered**
N/A

**Additional context**

dai-chen:
Issues related to meta field:
+ https://github.com/opensearch-project/sql/issues/339
+ https://github.com/opensearch-project/sql/issues/639

PR #2762, opened 2024-06-18
Add support for custom date formats and OpenSearch date formats for date fields as part of Lucene query

manasvinibs:
### Description
This change adds support for OpenSearch Date formats and custom date formats to be part of SQL Lucene queries. 
Without this change, we are supporting only list of selected formats which always gets formatted to ISO local string or epoch before submitting to opensearch dsl query. 

* Refactors `ExprType` to be `OpenSearchDateType` object instead of `ExprCoretype` enums  passed as a param to the Lucene QueryBuilder. 
* Extract `OpenSearchDateNamedFormatters` and `OpenSearchDateCustomFormatters` from the `OpenSearchDateType` object which is set during IndexMapping field parsing for a field type.
* Format information is passed to the `ExprValue` implementation constructor to parse the date string using the field specific formatters.
* Adds unit tests for all the different formats use cases. 
* Adds `OpenSearch` package dependency to the `core` module to use OpenSearch provided `DateFormatter `and `DateFormatters` classes instead of java provided DateFormatter.

Follow-up PR - I'll add IT tests in a separate PR
 
### Issues Resolved
https://github.com/opensearch-project/sql/issues/2700
 
### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java
@@ -62,19 +62,23 @@ public class OpenSearchDataType implements ExprType, Serializable {
   @EqualsAndHashCode.Exclude @Getter protected MappingType mappingType;
 
   // resolved ExprCoreType
-  protected ExprCoreType exprCoreType;
+  @Getter protected ExprCoreType exprCoreType;
 
   /**
    * Get a simplified type {@link ExprCoreType} if possible. To avoid returning `UNKNOWN` for
-   * `OpenSearch*Type`s, e.g. for IP, returns itself.
+   * `OpenSearch*Type`s, e.g. for IP, returns itself. If the `exprCoreType` is {@link
+   * ExprCoreType#DATE}, {@link ExprCoreType#TIMESTAMP}, {@link ExprCoreType#TIME}, or {@link
+   * ExprCoreType#UNKNOWN}, it returns the current instance; otherwise, it returns `exprCoreType`.
    *
    * @return An {@link ExprType}.
    */
   public ExprType getExprType() {
-    if (exprCoreType != ExprCoreType.UNKNOWN) {
-      return exprCoreType;
-    }
-    return this;
+    return (exprCoreType == ExprCoreType.DATE
+            || exprCoreType == ExprCoreType.TIMESTAMP
+            || exprCoreType == ExprCoreType.TIME
+            || exprCoreType == ExprCoreType.UNKNOWN)
+        ? this
+        : exprCoreType;
   }
 
   /**
@@ -228,6 +232,9 @@ public class OpenSearchDataType implements ExprType, Serializable {
   // Called when serializing SQL response
   public String legacyTypeName() {
     if (mappingType == null) {
+      return exprCoreType.typeName();
+    }
+    if (mappingType.toString().equalsIgnoreCase("DATE")) {
       return exprCoreType.typeName();
     }
     return mappingType.toString().toUpperCase();
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
@@ -11,11 +11,16 @@ import static org.opensearch.sql.data.type.ExprCoreType.DATE;
 import static org.opensearch.sql.data.type.ExprCoreType.TIME;
 import static org.opensearch.sql.data.type.ExprCoreType.TIMESTAMP;
 
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.temporal.TemporalAccessor;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import lombok.EqualsAndHashCode;
 import org.opensearch.common.time.DateFormatter;
+import org.opensearch.common.time.DateFormatters;
 import org.opensearch.common.time.FormatNames;
 import org.opensearch.sql.data.type.ExprCoreType;
 import org.opensearch.sql.data.type.ExprType;
@@ -137,6 +142,11 @@ public class OpenSearchDateType extends OpenSearchDataType {
 
   private static final String CUSTOM_FORMAT_DATE_SYMBOLS = "FecEWwYqQgdMLDyuG";
 
+  private static final List<DateFormatter> OPENSEARCH_DEFAULT_FORMATTERS =
+      Stream.of("strict_date_time_no_millis", "strict_date_optional_time", "epoch_millis")
+          .map(DateFormatter::forPattern)
+          .toList();
+
   @EqualsAndHashCode.Exclude private final List<String> formats;
 
   private OpenSearchDateType() {
@@ -233,6 +243,59 @@ public class OpenSearchDateType extends OpenSearchDataType {
     return getAllCustomFormats().stream()
         .map(DateFormatter::forPattern)
         .collect(Collectors.toList());
+  }
+
+  /**
+   * Retrieves a list of custom formatters and OpenSearch named formatters defined by the user, and
+   * attempts to parse the given date/time string using these formatters.
+   *
+   * @param dateTime The date/time string to parse.
+   * @return A ZonedDateTime representing the parsed date/time in UTC, or null if parsing fails.
+   */
+  public ZonedDateTime getParsedDateTime(String dateTime) {
+    List<DateFormatter> dateFormatters =
+        Stream.concat(this.getAllNamedFormatters().stream(), this.getAllCustomFormatters().stream())
+            .collect(Collectors.toList());
+    ZonedDateTime zonedDateTime = null;
+
+    // check if dateFormatters are empty, then set default ones
+    if (dateFormatters.isEmpty()) {
+      dateFormatters = OPENSEARCH_DEFAULT_FORMATTERS;
+    }
+    // parse using OpenSearch DateFormatters
+    for (DateFormatter formatter : dateFormatters) {
+      try {
+        TemporalAccessor accessor = formatter.parse(dateTime);
+        zonedDateTime = DateFormatters.from(accessor).withZoneSameLocal(ZoneOffset.UTC);
+        break;
+      } catch (IllegalArgumentException ignored) {
+        // nothing to do, try another format
+      }
+    }
+    return zonedDateTime;
+  }
+
+  /**
+   * Returns a formatted date string using the internal formatter, if available.
+   *
+   * @param accessor The TemporalAccessor object containing the date/time information.
+   * @return A formatted date string if a formatter is available, otherwise null.
+   */
+  public String getFormattedDate(TemporalAccessor accessor) {
+    if (hasNoFormatter()) {
+      return OPENSEARCH_DEFAULT_FORMATTERS.get(0).format(accessor);
+    }
+    // Use the first available format string to create the formatter
+    return DateFormatter.forPattern(this.formats.get(0)).format(accessor);
+  }
+
+  /**
+   * Checks if the formatter is not initialized.
+   *
+   * @return True if the formatter is not set, otherwise false.
+   */
+  public boolean hasNoFormatter() {
+    return this.formats.isEmpty();
   }
 
   /**
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
@@ -230,7 +230,7 @@ public class OpenSearchExprValueFactory {
   private static ExprValue parseDateTimeString(String value, OpenSearchDateType dataType) {
     List<DateFormatter> formatters = dataType.getAllNamedFormatters();
     formatters.addAll(dataType.getAllCustomFormatters());
-    ExprCoreType returnFormat = (ExprCoreType) dataType.getExprType();
+    ExprCoreType returnFormat = dataType.getExprCoreType();
 
     for (DateFormatter formatter : formatters) {
       try {
@@ -273,8 +273,7 @@ public class OpenSearchExprValueFactory {
 
   private static ExprValue createOpenSearchDateType(Content value, ExprType type) {
     OpenSearchDateType dt = (OpenSearchDateType) type;
-    ExprType returnFormat = dt.getExprType();
-
+    ExprCoreType returnFormat = dt.getExprCoreType();
     if (value.isNumber()) { // isNumber
       var numFormatters = dt.getNumericNamedFormatters();
       if (numFormatters.size() > 0 || !dt.hasFormats()) {
@@ -287,7 +286,7 @@ public class OpenSearchExprValueFactory {
           epochMillis = value.longValue();
         }
         Instant instant = Instant.ofEpochMilli(epochMillis);
-        switch ((ExprCoreType) returnFormat) {
+        switch (returnFormat) {
           case TIME:
             return new ExprTimeValue(LocalTime.from(instant.atZone(ZoneOffset.UTC)));
           case DATE:
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
@@ -8,8 +8,9 @@ package org.opensearch.sql.opensearch.storage.script.filter.lucene;
 import static org.opensearch.sql.analysis.NestedAnalyzer.isNestedFunction;
 
 import com.google.common.collect.ImmutableMap;
+import java.time.ZonedDateTime;
 import java.util.Map;
-import java.util.function.Function;
+import java.util.function.BiFunction;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.sql.data.model.ExprBooleanValue;
 import org.opensearch.sql.data.model.ExprByteValue;
@@ -32,6 +33,7 @@ import org.opensearch.sql.expression.NamedArgumentExpression;
 import org.opensearch.sql.expression.ReferenceExpression;
 import org.opensearch.sql.expression.function.BuiltinFunctionName;
 import org.opensearch.sql.expression.function.FunctionName;
+import org.opensearch.sql.opensearch.data.type.OpenSearchDateType;
 
 /** Lucene query abstraction that builds Lucene query from function expression. */
 public abstract class LuceneQuery {
@@ -105,135 +107,164 @@ public abstract class LuceneQuery {
     ReferenceExpression ref = (ReferenceExpression) func.getArguments().get(0);
     Expression expr = func.getArguments().get(1);
     ExprValue literalValue =
-        expr instanceof LiteralExpression ? expr.valueOf() : cast((FunctionExpression) expr);
+        expr instanceof LiteralExpression ? expr.valueOf() : cast((FunctionExpression) expr, ref);
     return doBuild(ref.getAttr(), ref.type(), literalValue);
   }
 
-  private ExprValue cast(FunctionExpression castFunction) {
+  private ExprValue cast(FunctionExpression castFunction, ReferenceExpression ref) {
     return castMap
         .get(castFunction.getFunctionName())
-        .apply((LiteralExpression) castFunction.getArguments().get(0));
+        .apply((LiteralExpression) castFunction.getArguments().get(0), ref);
   }
 
   /** Type converting map. */
-  private final Map<FunctionName, Function<LiteralExpression, ExprValue>> castMap =
-      ImmutableMap.<FunctionName, Function<LiteralExpression, ExprValue>>builder()
-          .put(
-              BuiltinFunctionName.CAST_TO_STRING.getName(),
-              expr -> {
-                if (!expr.type().equals(ExprCoreType.STRING)) {
-                  return new ExprStringValue(String.valueOf(expr.valueOf().value()));
-                } else {
-                  return expr.valueOf();
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_BYTE.getName(),
-              expr -> {
-                if (ExprCoreType.numberTypes().contains(expr.type())) {
-                  return new ExprByteValue(expr.valueOf().byteValue());
-                } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
-                  return new ExprByteValue(expr.valueOf().booleanValue() ? 1 : 0);
-                } else {
-                  return new ExprByteValue(Byte.valueOf(expr.valueOf().stringValue()));
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_SHORT.getName(),
-              expr -> {
-                if (ExprCoreType.numberTypes().contains(expr.type())) {
-                  return new ExprShortValue(expr.valueOf().shortValue());
-                } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
-                  return new ExprShortValue(expr.valueOf().booleanValue() ? 1 : 0);
-                } else {
-                  return new ExprShortValue(Short.valueOf(expr.valueOf().stringValue()));
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_INT.getName(),
-              expr -> {
-                if (ExprCoreType.numberTypes().contains(expr.type())) {
-                  return new ExprIntegerValue(expr.valueOf().integerValue());
-                } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
-                  return new ExprIntegerValue(expr.valueOf().booleanValue() ? 1 : 0);
-                } else {
-                  return new ExprIntegerValue(Integer.valueOf(expr.valueOf().stringValue()));
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_LONG.getName(),
-              expr -> {
-                if (ExprCoreType.numberTypes().contains(expr.type())) {
-                  return new ExprLongValue(expr.valueOf().longValue());
-                } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
-                  return new ExprLongValue(expr.valueOf().booleanValue() ? 1 : 0);
-                } else {
-                  return new ExprLongValue(Long.valueOf(expr.valueOf().stringValue()));
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_FLOAT.getName(),
-              expr -> {
-                if (ExprCoreType.numberTypes().contains(expr.type())) {
-                  return new ExprFloatValue(expr.valueOf().floatValue());
-                } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
-                  return new ExprFloatValue(expr.valueOf().booleanValue() ? 1 : 0);
-                } else {
-                  return new ExprFloatValue(Float.valueOf(expr.valueOf().stringValue()));
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_DOUBLE.getName(),
-              expr -> {
-                if (ExprCoreType.numberTypes().contains(expr.type())) {
-                  return new ExprDoubleValue(expr.valueOf().doubleValue());
-                } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
-                  return new ExprDoubleValue(expr.valueOf().booleanValue() ? 1 : 0);
-                } else {
-                  return new ExprDoubleValue(Double.valueOf(expr.valueOf().stringValue()));
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_BOOLEAN.getName(),
-              expr -> {
-                if (ExprCoreType.numberTypes().contains(expr.type())) {
-                  return expr.valueOf().doubleValue() != 0
-                      ? ExprBooleanValue.of(true)
-                      : ExprBooleanValue.of(false);
-                } else if (expr.type().equals(ExprCoreType.STRING)) {
-                  return ExprBooleanValue.of(Boolean.valueOf(expr.valueOf().stringValue()));
-                } else {
-                  return expr.valueOf();
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_DATE.getName(),
-              expr -> {
-                if (expr.type().equals(ExprCoreType.STRING)) {
-                  return new ExprDateValue(expr.valueOf().stringValue());
-                } else {
-                  return new ExprDateValue(expr.valueOf().dateValue());
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_TIME.getName(),
-              expr -> {
-                if (expr.type().equals(ExprCoreType.STRING)) {
-                  return new ExprTimeValue(expr.valueOf().stringValue());
-                } else {
-                  return new ExprTimeValue(expr.valueOf().timeValue());
-                }
-              })
-          .put(
-              BuiltinFunctionName.CAST_TO_TIMESTAMP.getName(),
-              expr -> {
-                if (expr.type().equals(ExprCoreType.STRING)) {
-                  return new ExprTimestampValue(expr.valueOf().stringValue());
-                } else {
-                  return new ExprTimestampValue(expr.valueOf().timestampValue());
-                }
-              })
-          .build();
+  private final Map<FunctionName, BiFunction<LiteralExpression, ReferenceExpression, ExprValue>>
+      castMap =
+          ImmutableMap
+              .<FunctionName, BiFunction<LiteralExpression, ReferenceExpression, ExprValue>>
+                  builder()
+              .put(
+                  BuiltinFunctionName.CAST_TO_STRING.getName(),
+                  (expr, ref) -> {
+                    if (!expr.type().equals(ExprCoreType.STRING)) {
+                      return new ExprStringValue(String.valueOf(expr.valueOf().value()));
+                    } else {
+                      return expr.valueOf();
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_BYTE.getName(),
+                  (expr, ref) -> {
+                    if (ExprCoreType.numberTypes().contains(expr.type())) {
+                      return new ExprByteValue(expr.valueOf().byteValue());
+                    } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
+                      return new ExprByteValue(expr.valueOf().booleanValue() ? 1 : 0);
+                    } else {
+                      return new ExprByteValue(Byte.valueOf(expr.valueOf().stringValue()));
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_SHORT.getName(),
+                  (expr, ref) -> {
+                    if (ExprCoreType.numberTypes().contains(expr.type())) {
+                      return new ExprShortValue(expr.valueOf().shortValue());
+                    } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
+                      return new ExprShortValue(expr.valueOf().booleanValue() ? 1 : 0);
+                    } else {
+                      return new ExprShortValue(Short.valueOf(expr.valueOf().stringValue()));
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_INT.getName(),
+                  (expr, ref) -> {
+                    if (ExprCoreType.numberTypes().contains(expr.type())) {
+                      return new ExprIntegerValue(expr.valueOf().integerValue());
+                    } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
+                      return new ExprIntegerValue(expr.valueOf().booleanValue() ? 1 : 0);
+                    } else {
+                      return new ExprIntegerValue(Integer.valueOf(expr.valueOf().stringValue()));
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_LONG.getName(),
+                  (expr, ref) -> {
+                    if (ExprCoreType.numberTypes().contains(expr.type())) {
+                      return new ExprLongValue(expr.valueOf().longValue());
+                    } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
+                      return new ExprLongValue(expr.valueOf().booleanValue() ? 1 : 0);
+                    } else {
+                      return new ExprLongValue(Long.valueOf(expr.valueOf().stringValue()));
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_FLOAT.getName(),
+                  (expr, ref) -> {
+                    if (ExprCoreType.numberTypes().contains(expr.type())) {
+                      return new ExprFloatValue(expr.valueOf().floatValue());
+                    } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
+                      return new ExprFloatValue(expr.valueOf().booleanValue() ? 1 : 0);
+                    } else {
+                      return new ExprFloatValue(Float.valueOf(expr.valueOf().stringValue()));
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_DOUBLE.getName(),
+                  (expr, ref) -> {
+                    if (ExprCoreType.numberTypes().contains(expr.type())) {
+                      return new ExprDoubleValue(expr.valueOf().doubleValue());
+                    } else if (expr.type().equals(ExprCoreType.BOOLEAN)) {
+                      return new ExprDoubleValue(expr.valueOf().booleanValue() ? 1 : 0);
+                    } else {
+                      return new ExprDoubleValue(Double.valueOf(expr.valueOf().stringValue()));
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_BOOLEAN.getName(),
+                  (expr, ref) -> {
+                    if (ExprCoreType.numberTypes().contains(expr.type())) {
+                      return expr.valueOf().doubleValue() != 0
+                          ? ExprBooleanValue.of(true)
+                          : ExprBooleanValue.of(false);
+                    } else if (expr.type().equals(ExprCoreType.STRING)) {
+                      return ExprBooleanValue.of(Boolean.valueOf(expr.valueOf().stringValue()));
+                    } else {
+                      return expr.valueOf();
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_DATE.getName(),
+                  (expr, ref) -> {
+                    if (expr.type().equals(ExprCoreType.STRING)) {
+                      ZonedDateTime zonedDateTime = getParsedDateTime(expr, ref);
+                      if (zonedDateTime != null) {
+                        return new ExprDateValue(zonedDateTime.toLocalDate());
+                      }
+                      return new ExprDateValue(expr.valueOf().stringValue());
+                    } else {
+                      return new ExprDateValue(expr.valueOf().dateValue());
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_TIME.getName(),
+                  (expr, ref) -> {
+                    if (expr.type().equals(ExprCoreType.STRING)) {
+                      ZonedDateTime zonedDateTime = getParsedDateTime(expr, ref);
+                      if (zonedDateTime != null) {
+                        return new ExprTimeValue(zonedDateTime.toLocalTime());
+                      }
+                      return new ExprTimeValue(expr.valueOf().stringValue());
+                    } else {
+                      return new ExprTimeValue(expr.valueOf().timeValue());
+                    }
+                  })
+              .put(
+                  BuiltinFunctionName.CAST_TO_TIMESTAMP.getName(),
+                  (expr, ref) -> {
+                    if (expr.type().equals(ExprCoreType.STRING)) {
+                      ZonedDateTime zonedDateTime = getParsedDateTime(expr, ref);
+                      if (zonedDateTime != null) {
+                        return new ExprTimestampValue(zonedDateTime.toInstant());
+                      }
+                      return new ExprTimestampValue(expr.valueOf().stringValue());
+                    } else {
+                      return new ExprTimestampValue(expr.valueOf().timestampValue());
+                    }
+                  })
+              .build();
+
+  /**
+   * Parses the date/time from the given expression if the reference type is an instance of
+   * OpenSearchDateType.
+   *
+   * @param expr The expression to parse.
+   * @return The parsed ZonedDateTime or null if the conditions are not met.
+   */
+  private ZonedDateTime getParsedDateTime(LiteralExpression expr, ReferenceExpression ref) {
+    if (ref.type() instanceof OpenSearchDateType) {
+      return ((OpenSearchDateType) ref.type()).getParsedDateTime(expr.valueOf().stringValue());
+    }
+    return null;
+  }
 
   /**
    * Build method that subclass implements by default which is to build query from reference and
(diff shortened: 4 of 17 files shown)


core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
```diff
@@ -25,20 +32,44 @@
...
+  }
+
   @Override
   public String value() {
-    return DateTimeFormatter.ISO_LOCAL_DATE.format(date);
+    if (this.datePattern == null) {
+      return DateTimeFormatter.ISO_LOCAL_DATE.format(date);
+    }
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern(this.datePattern , Locale.ROOT);
+    return this.date.format(formatter);
```
penghuo, line 72: what is the issue solved by this? could u add a test to cover this change?

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
```diff
@@ -25,20 +32,44 @@
...
+  // List of cleaned patterns
+  private static final List<String> CLEANED_PATTERNS = Arrays.stream(DATE_TIME_FORMATTER_PATTERNS_NANOS_OPTIONAL.split("\\]\\["))
+          .map(pattern -> pattern.replace("[", "").replace("]", ""))
+          .collect(Collectors.toList());
 
   /** Constructor of ExprDateValue. */
   public ExprDateValue(String date) {
     try {
       this.date = LocalDate.parse(date, DATE_TIME_FORMATTER_VARIABLE_NANOS_OPTIONAL);
+      this.datePattern = determineDatePattern(date);
```
penghuo, line 45: does parese execute 2 times, the reasons is LocalDate.parse already parse the date string

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
penghuo, line 72: what is the issue solved by this? could u add a test to cover this change?

core/src/test/java/org/opensearch/sql/data/model/DateTimeValueTest.java
```diff
@@ -95,6 +95,12 @@ public void dateValueInterfaceTest() {
     assertEquals("invalid to get dateValue from value of type INTEGER", exception.getMessage());
   }
 
+  @Test
+  public void dateValueInterfaceSupportedFormatTest() {
+    ExprValue dateValue = new ExprDateValue("03-Jan-21");
+    assertEquals(LocalDate.of(2021, 1, 3), dateValue.dateValue());
+  }
```
ykmr1224, line 102: Let's cover all the patterns and also the pattern not defined.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
```diff
@@ -25,20 +32,44 @@
 public class ExprDateValue extends AbstractExprValue {
 
   private final LocalDate date;
+  private String datePattern;
+  // List of cleaned patterns
+  private static final List<String> CLEANED_PATTERNS = Arrays.stream(DATE_TIME_FORMATTER_PATTERNS_NANOS_OPTIONAL.split("\\]\\["))
+          .map(pattern -> pattern.replace("[", "").replace("]", ""))
+          .collect(Collectors.toList());
```
ykmr1224, line 39: It was a bit difficult to understand what this is doing. It would be helpful for reading if it is extracted as method like `extractDateTimePatterns`.
I would consider generating pattern from list instead of parsing from string. (That would be more error prune)

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
manasvinibs: On line #44 here, yes we are already parsing the date string to Local date object. But I need to know the exact pattern that matched the input date string. So inside `determineDatePattern` function, using `parse` function (LocalDate/LocalDatetime) I'm determining the matching pattern and returning that to set `this.datePattern`. This date pattern is used to format the LocalDate date object instead of default ISO_LOCAL_DATE string inside `value()`.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
manasvinibs: When we read the value for a literal while building query to opensearch DSL, we are reading the `ExprDateValue.value()` function which by default always converts the local date object into ISO string format. This leads to `OpenSearchParseException ` as it fails to parse date field in some custom format (dd-MM-yy) with ISO format. 
Here I'm formatting the parsed date to the matching input string pattern while converting it to literals in the request.

core/src/test/java/org/opensearch/sql/data/model/DateTimeValueTest.java
manasvinibs: Yes, added more tests covering the patterns and unsupported ones.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
manasvinibs: Sure, I've converted the patterns as list and do not need this cleaning part anymore.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
penghuo: Thx, 
* I understand the reason of add pattern to support parse literal string value.
* But, why we want to value() function keep original format?

penghuo: As discussed offline, when translate Date value to string in term query, we could explore use field format defined in index.

LantaoJin: What in my mind is why we don't know the customized date format during catalog binding? For example, in analyzing the field as a date, the format pattern "dd-MMM-yy" should be recognized and stored in SQL's `Date` dataType.
```
PUT testindex
{
  "mappings" : {
    "properties" :  {
      "release_date" : {
        "type" : "date",
        "format" : "dd-MMM-yy"
      }
    }
  }
}
```

core/src/main/java/org/opensearch/sql/utils/DateTimeFormatters.java
```diff
@@ -130,7 +142,7 @@ public class DateTimeFormatters {
 
   public static final DateTimeFormatter DATE_TIME_FORMATTER_VARIABLE_NANOS_OPTIONAL =
       new DateTimeFormatterBuilder()
-          .appendPattern("[uuuu-MM-dd HH:mm:ss][uuuu-MM-dd HH:mm][HH:mm:ss][HH:mm][uuuu-MM-dd]")
+          .appendPattern("[" + String.join("][", DATE_TIME_FORMATTER_PATTERNS_NANOS_OPTIONAL) + "]")
```
ykmr1224, line 145: Not a blocker, but it is better doing like wrapping each item like `"[" + item + "]"` or `String.format("[%s]")` and then join without delimiter. It would make it more straight forward and clearer.

ykmr1224 (commented):
Seems you need to reformat the code using `./gradlew spotlessApply`.

core/src/main/java/org/opensearch/sql/utils/DateTimeFormatters.java
manasvinibs: This is not needed as I'm using OpenSearch Date formatters to parse the date string.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
manasvinibs: Yes I've updated the PR to use OpenSearch date formatters of a field from its mappings info to parse the date string from Lucene query.

manasvinibs:
@penghuo @ykmr1224 I've updated the PR to now use OpenSearch Date formatters specific to field mapping to parse date string read as part of Lucene query instead of hard coded list of formats. Now instead of `ExprCoreType`, I'm using `OpenSearchDateType` field as part of QueryBuilder. I've 2 of the integration tests failing which I'm looking into fixing it. Meanwhile can you please review my changes.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java
```diff
@@ -62,7 +62,7 @@ public String toString() {
   @EqualsAndHashCode.Exclude @Getter protected MappingType mappingType;
 
   // resolved ExprCoreType
-  protected ExprCoreType exprCoreType;
+  @Getter protected ExprCoreType exprCoreType;
 
   /**
    * Get a simplified type {@link ExprCoreType} if possible. To avoid returning `UNKNOWN` for
```
LantaoJin, line 68: Would you update this javadoc? It's outdate.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -230,7 +230,7 @@ private Optional<ExprType> type(String field) {
   private static ExprValue parseDateTimeString(String value, OpenSearchDateType dataType) {
     List<DateFormatter> formatters = dataType.getAllNamedFormatters();
     formatters.addAll(dataType.getAllCustomFormatters());
-    ExprCoreType returnFormat = (ExprCoreType) dataType.getExprType();
+    ExprCoreType returnFormat = dataType.getExprCoreType();
```
LantaoJin, line 233: The `returnFormat` could be `UNKNOWN` now. Be mind that `getExprCoreType()` may introduce some behaviour changes. Maybe it's safe here.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -32,10 +34,13 @@
 import org.opensearch.sql.expression.ReferenceExpression;
 import org.opensearch.sql.expression.function.BuiltinFunctionName;
 import org.opensearch.sql.expression.function.FunctionName;
+import org.opensearch.sql.opensearch.data.type.OpenSearchDateType;
 
 /** Lucene query abstraction that builds Lucene query from function expression. */
 public abstract class LuceneQuery {
 
+  private List<DateFormatter> dateFormatters;
```
LantaoJin, line 42: Is it safe to store this list variable here? I think we should refactor method 
```java
ExprValue cast(FunctionExpression castFunction)
```
 to 
```java
ExprValue cast(FunctionExpression castFunction), List<DateFormatter> dateFormatters
```
and remove this variable.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/RangeQuery.java
```diff
@@ -28,8 +30,14 @@ public enum Comparison {
   /** Comparison that range query build for. */
   private final Comparison comparison;
 
+  List<String> fieldFormats;
```
LantaoJin, line 33: ditto

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/TermQuery.java
```diff
@@ -5,24 +5,32 @@
...
 import org.opensearch.index.query.QueryBuilders;
 import org.opensearch.sql.data.model.ExprValue;
 import org.opensearch.sql.data.type.ExprCoreType;
 import org.opensearch.sql.data.type.ExprType;
+import org.opensearch.sql.opensearch.data.type.OpenSearchDateType;
 import org.opensearch.sql.opensearch.data.type.OpenSearchTextType;
 
 /** Lucene query that build term query for equality comparison. */
 public class TermQuery extends LuceneQuery {
+  List<String> fieldFormats;
```
LantaoJin, line 19: ditto

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
```diff
@@ -25,6 +31,7 @@
 public class ExprDateValue extends AbstractExprValue {
 
   private final LocalDate date;
+  private String dateFormat;
 
   /** Constructor of ExprDateValue. */
   public ExprDateValue(String date) {
```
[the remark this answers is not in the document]
ykmr1224: +1

core/src/main/java/org/opensearch/sql/data/model/ExprTimeValue.java
```diff
@@ -37,9 +44,43 @@ public ExprTimeValue(String time) {
     }
   }
 
+  /** Constructor of ExprTimeValue to support custom/OpenSearch date formats in mappings. */
+  public ExprTimeValue(String time, List<DateFormatter> dateFormatters) {
```
ykmr1224, line 48: Seems it's very similar to ExprDateValue constructor. Can we extract some part and reduce redundancy?

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
```diff
@@ -36,9 +43,44 @@ public ExprDateValue(String date) {
...
+      try {
+        TemporalAccessor accessor = formatter.parse(date);
+        ZonedDateTime zonedDateTime = DateFormatters.from(accessor);
+        localDate = zonedDateTime.withZoneSameLocal(ZoneOffset.UTC).toLocalDate();
+
+        dateFormat = formatter.format(accessor);
+        break;
+      } catch (IllegalArgumentException ignored) {
+        // nothing to do, try another format
+      }
```
ykmr1224, line 69: What happens if all the formatter failed? Should we throw exception in that case?

core/src/main/java/org/opensearch/sql/data/model/ExprTimestampValue.java
```diff
@@ -42,18 +50,54 @@ public ExprTimestampValue(String timestamp) {
     }
   }
 
+  /**
+   * Constructor of ExprTimestampValue to support custom/OpenSearch dateTime formats in mappings.
+   */
+  public ExprTimestampValue(String timestamp, List<DateFormatter> dateFormatters) {
```
ykmr1224, line 56: Same as above.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
```diff
@@ -25,6 +31,7 @@
 public class ExprDateValue extends AbstractExprValue {
 
   private final LocalDate date;
+  private String dateFormat;
```
penghuo, line 34: add javadoc, what is dateFormat? is it a Format? or formatted date?

integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
```diff
@@ -48,8 +48,8 @@ public void test_nonnumeric_data_types() throws IOException {
         schema("keyword_value", "string"),
         schema("text_value", "string"),
         schema("binary_value", "binary"),
-        schema("date_value", "timestamp"),
-        schema("date_nanos_value", "timestamp"),
+        schema("date_value", "date"),
```
penghuo, line 51: why change timestmap to date? IT is broken?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
penghuo: Lucene query should not ref to dateFormatters.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/RangeQuery.java
penghuo: remove it?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/RangeQuery.java
```diff
@@ -48,7 +56,8 @@ protected QueryBuilder doBuild(String fieldName, ExprType fieldType, ExprValue l
   }
 
   private Object value(ExprValue literal) {
-    if (literal.type().equals(ExprCoreType.TIMESTAMP)) {
+    if (literal.type().equals(ExprCoreType.TIMESTAMP)
+        && (fieldFormats == null || fieldFormats.isEmpty())) {
```
penghuo, line 60: why add this logic fieldFormats == null || fieldFormats.isEmpty()?

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
```diff
@@ -36,9 +43,44 @@ public ExprDateValue(String date) {
     }
   }
 
+  /** Constructor of ExprDateValue to support custom/OpenSearch date formats in mappings. */
+  public ExprDateValue(String date, List<DateFormatter> dateFormatters) {
```
penghuo, line 47: try to understand what we are tyring to achieve. Construct a ExprDateValue from string value, when call getValue, the result should be formatted with provided formatter? If yes, we should move the format logic to getValue instead of in constructor where the value is lazy evaluated.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -32,10 +32,13 @@
 import org.opensearch.sql.expression.ReferenceExpression;
 import org.opensearch.sql.expression.function.BuiltinFunctionName;
 import org.opensearch.sql.expression.function.FunctionName;
+import org.opensearch.sql.opensearch.data.type.OpenSearchDateType;
 
 /** Lucene query abstraction that builds Lucene query from function expression. */
 public abstract class LuceneQuery {
 
+  private ReferenceExpression ref;
```
LantaoJin, line 40: Maybe you could add this field and an override method `build()` in `SingleFieldQuery`, instead of `LuceneQuery`. My another concern is will this class be used in multiple threads? Adding a field in XXQuery will cause bug in multiple threads case.

LantaoJin:
@manasvinibs Try to avoid using force push when the PR contained many review comments.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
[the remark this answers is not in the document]
manasvinibs: Updated

core/src/main/java/org/opensearch/sql/data/model/ExprTimeValue.java
```diff
@@ -26,6 +32,7 @@
 public class ExprTimeValue extends AbstractExprValue {
 
   private final LocalTime time;
+  private String timeFormat;
 
   /** Constructor of ExprTimeValue. */
   public ExprTimeValue(String time) {
```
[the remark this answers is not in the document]
manasvinibs: Done

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
```diff
@@ -36,9 +43,44 @@ public ExprDateValue(String date) {
...
+        dateFormatters.add(DateFormatter.forPattern(pattern));
+      }
+    }
+    for (DateFormatter formatter : dateFormatters) {
+      try {
+        TemporalAccessor accessor = formatter.parse(date);
+        ZonedDateTime zonedDateTime = DateFormatters.from(accessor);
+        localDate = zonedDateTime.withZoneSameLocal(ZoneOffset.UTC).toLocalDate();
+
+        dateFormat = formatter.format(accessor);
```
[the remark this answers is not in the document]
manasvinibs: Cleaned up little bit, avoided adding final keyword as it will break all the other places in the code using default constructor.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
manasvinibs: I think with the current code too returnFormat could have been unknown. Here I didn't see any issues with changing to `ExprCoreType`.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
manasvinibs: I refactored this to use available reference expression to identify the `type`. I have moved the formatter logic to `OpenSearchDateType`  class helper function and just passing that here in the cast map.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/RangeQuery.java
manasvinibs: Removed

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/TermQuery.java
manasvinibs: removed

core/src/main/java/org/opensearch/sql/data/model/ExprTimeValue.java
manasvinibs: Yes I did extracted few date formatting stuff inside DateFormatter util class. I think remaining ones are little more specific to the way we parse the date string in that particular class.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
manasvinibs: Yes if we cannot parse using default/custom or constructor set formats, we are throwing exception on line #70.

core/src/main/java/org/opensearch/sql/data/model/ExprTimestampValue.java
manasvinibs: Done

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
manasvinibs: Renamed this to formattedDate and added java doc.

integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
manasvinibs: Yes now that date fields are type `OpenSearchDateType` and not ` ExprCoreType`, I was having IT tests mismatch, this is to fix that.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/RangeQuery.java
manasvinibs: Removed this.

integ-test/src/test/java/org/opensearch/sql/sql/DateTimeFunctionIT.java
```diff
@@ -1455,8 +1455,8 @@ public void testBracketedEquivalent() throws IOException {
     compareBrackets("timestamp", "ts", "2020-09-16 17:30:00");
     compareBrackets("timestamp", "timestamp", "2020-09-16 17:30:00.123");
     compareBrackets("timestamp", "ts", "2020-09-16 17:30:00.123");
-    compareBrackets("date", "date", "2020-09-16");
-    compareBrackets("date", "d", "2020-09-16");
+    compareBrackets("date", "date", "2020-09-16T00:00:00.000Z");
+    compareBrackets("date", "d", "2020-09-16T00:00:00.000Z");
```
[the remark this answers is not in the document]
manasvinibs: > Another question is why is `sssZ` here, but the `SQL_LITERAL_DEFAULT_DATE_TIME_FORMAT` defined to `ssZ` above.

Sorry for the confusion, I'm not using this. I can remove.

core/src/test/java/org/opensearch/sql/data/model/DateTimeValueTest.java
```diff
@@ -103,6 +110,347 @@ public void dateInUnsupportedFormat() {
         "date:2020-07-07Z in unsupported format, please use 'yyyy-MM-dd'", exception.getMessage());
   }
 
+  @Test
+  void testValidTimestampWithCustomFormatter() {
+    String timestamp = "2021-11-08T17:00:00Z";
+    DateFormatter formatter = DateFormatter.forPattern("yyyy-MM-dd'T'HH:mm:ssX");
```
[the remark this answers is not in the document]
manasvinibs: Not really using the one defined [here](https://github.com/opensearch-project/sql/pull/2762/files#diff-a0471258f6239d679f1993616080bfde30a9f43c336f9305fa49653b7b3dc5c1R120), will be removed. 

Also, here in this test I'm testing the use case for supporting cutom formatters that user can define for date field in the mapping and in this case it is `yyyy-MM-dd'T'HH:mm:ssX` format.  I'm testing multiple different custom formats and named formats in this class along with default ones.

ykmr1224 (approved):

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
penghuo: > I'm parsing date string using OpenSearch DateFormatters

Core module should not directly depend on OpenSearch library. This is why we have OpenSearch module.

penghuo: > There is no lazy variable evaluation in Java

we can simulate lazy evalution, right?

> especially there is a for loop in this logic.

another, could we avoid for...loop here?

core/build.gradle
```diff
@@ -56,6 +56,7 @@ dependencies {
     api "com.fasterxml.jackson.core:jackson-annotations:${versions.jackson}"
     api group: 'com.google.code.gson', name: 'gson', version: '2.8.9'
     api group: 'com.tdunning', name: 't-digest', version: '3.3'
+    api group: 'org.opensearch', name: 'opensearch', version: "${opensearch_version}"
```
penghuo, line 59: Core module should not directly depend on OpenSearch library.

core/src/main/java/org/opensearch/sql/data/model/ExprDateValue.java
penghuo: Instead change value() of ExprDateValue, format the ExprDateValue.value() based on Type. for instance
```
if (this.ref.type instance of OpenSearchDataType) {
  ((OpenSearchDataType) this.ref.type instance).getAllFormatters.format(literal.value())
}
```

manasvinibs: I see what you mean. Instead of bringing OpenSearch library dependency to `core`, I can do all the formatting logic within `OpenSearchDateType ` class which already has access to OpenSearch DateFormatters. 
```
if (this.ref.type instance of OpenSearchDataType) {
  ((OpenSearchDataType) this.ref.type instance).getAllFormatters.format(literal.value())
}
```
But the above logic cannot go inside `ExprDatevalue` class because I cannot import `OpenSearchDateType` class within core as it will make it circular dependency. (OpenSearch module already has dependency on core modules)

Synced offline with @penghuo to make the below changes:

- Move DateFormatter logic inside `OpenSearchDateType` class 
- Lucene query class refers to `OpenSearchDateType.getParsedDateTime()` function to get the parsed date object and pass it to `ExprDateValue` constructor
- Remove new constructors from ExprDateType and similar classes
- Formatted value will be fetched inside term query and range query referring to `OpenSearchDateType` class directly.

core/build.gradle
manasvinibs: Removed the dependency and moved formatting logic inside `OpenSearchDateType` class as discussed offline

integ-test/src/test/java/org/opensearch/sql/sql/SystemFunctionIT.java
```diff
@@ -65,15 +65,6 @@ public void typeof_opensearch_types() {
...
-            "TEXT",
-            "TIMESTAMP",
-            "TIMESTAMP",
-            "BOOLEAN",
-            "OBJECT",
-            "KEYWORD",
-            "IP",
-            "BINARY",
-            "GEO_POINT"));
+        rows("TEXT", "DATE", "DATE", "BOOLEAN", "OBJECT", "KEYWORD", "IP", "BINARY", "GEO_POINT"));
```
penghuo, line 68: @dai-chen @rupal-bq do u rememeber why we use timestamp for JDBC type?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -209,7 +214,16 @@ private ExprValue cast(FunctionExpression castFunction) {
           .put(
               BuiltinFunctionName.CAST_TO_DATE.getName(),
               expr -> {
-                if (expr.type().equals(ExprCoreType.STRING)) {
+                if (expr.type().equals(ExprCoreType.STRING)
```
penghuo, line 217: The code is duplicate as 235-244 and 253-262, could u simplify.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
```diff
@@ -235,6 +247,71 @@ public List<DateFormatter> getAllCustomFormatters() {
...
   }
 
+  /**
+   * Retrieves a list of custom formatters and OpenSearch named formatters defined by the user, and
+   * attempts to parse the given date/time string using these formatters.
+   *
+   * @param dateTime The date/time string to parse.
+   * @return A ZonedDateTime representing the parsed date/time in UTC, or null if parsing fails.
+   */
+  public ZonedDateTime getParsedDateTime(String dateTime) {
```
penghuo, line 257: we should not choose dateFormatter based on input dateTime.
in getFormattedData, can we move this logic to getFormattedDate?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
penghuo: Regarding to parse string as ExprDateValue, we can extend supported formatted in ExprDataValue and update user doc. 
The literay format data value in SQL not necessary follow OpenSearch index dataformat.

for instance, if SQL only support date in format 2024-01-03, it is also fine. unless we can make sure the literary value is rewrite as 03-Jan-24 in DSL query.
```
POST _plugins/_sql
{
    "query":"select release_date from testindex where release_date='03-Jan-24'"
}


POST _plugins/_sql
{
    "query":"select release_date from testindex where release_date='2024-01-03'"
}
```

integ-test/src/test/java/org/opensearch/sql/sql/SystemFunctionIT.java
dai-chen: I found some context here: https://github.com/opendistro-for-elasticsearch/sql/pull/716

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
manasvinibs: Sorry, can you help me understand more here? In this logic, I'm parsing the input string to date object using OpenSearch field dateFormatters. If we don't want to do this, then we won't be able to parse input string such as `03-Jan-21`(dd-MM-yy) as part of term/range query which causes parsing exception. Aren't we supposed to support all valid date formats that OpenSearch can hold values into as part of Lucene query?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
manasvinibs: Does it mean its ok to not support the below query? 
```
POST _plugins/_sql
{
    "query":"select release_date from testindex where release_date='03-Jan-24'"
}
```
Instead, when field mapping format has `dd-MM-yy`, we would still take '2024-01-03' string as part of query and do formatting to match `dd-MM-yy` just before sending request in DSL query?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
manasvinibs: Synced offline with @penghuo to keep the parsing date string to date object using OpenSearch dateFormatter logic as it is but to move formatting date string logic to Term/Range query when needed.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
LantaoJin: Too many code rebases, haven't catch up the latest commit. But if the only target is to support two-digits year, maybe you can refer my poc [here](https://github.com/LantaoJin/search-plugins-sql/commit/265be074cb22a7f70ed730568e5f7f1f44c043f0), it armed to fix the issue [[BUG] str_to_date does not work with two-digits year](https://github.com/opensearch-project/sql/issues/2741)

integ-test/src/test/java/org/opensearch/sql/sql/SystemFunctionIT.java
manasvinibs: I reverted these changes and IT looks good. We don't need this here.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -227,14 +238,31 @@ private ExprValue cast(FunctionExpression castFunction) {
...
 
+  /**
+   * Parses the date/time from the given expression if it is of string type and the reference type
+   * is an instance of OpenSearchDateType.
+   *
+   * @param expr The expression to parse.
+   * @return The parsed ZonedDateTime or null if the conditions are not met.
+   */
+  private ZonedDateTime getParsedDateTime(LiteralExpression expr) {
+    if (expr.type().equals(ExprCoreType.STRING) && this.ref.type() instanceof OpenSearchDateType) {
```
penghuo, line 260: line 220 check the type is STRING. can we simiply the code?

penghuo: can u update on this behaviour. https://github.com/opensearch-project/sql/blob/main/docs/user/general/datatypes.rst#conversion-from-string-to-date-and-time-types

we should clarify in doc what is the valide dateformat can be used for literal value. 
does it means only when the operator include indexDateField and literal value, then the literal value can be indexDateFieldd format?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -120,7 +125,7 @@ private ExprValue cast(FunctionExpression castFunction) {
       ImmutableMap.<FunctionName, Function<LiteralExpression, ExprValue>>builder()
           .put(
               BuiltinFunctionName.CAST_TO_STRING.getName(),
-              expr -> {
+              (expr) -> {
```
penghuo, line 128: revert it

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/RangeQuery.java
```diff
@@ -30,7 +31,8 @@ public enum Comparison {
 
   @Override
   protected QueryBuilder doBuild(String fieldName, ExprType fieldType, ExprValue literal) {
-    Object value = value(literal);
+
```
penghuo, line 34: remove empty line.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/RangeQuery.java
```diff
@@ -47,11 +49,23 @@ protected QueryBuilder doBuild(String fieldName, ExprType fieldType, ExprValue l
...
+            : openSearchDateType.getFormattedDate(literal.timestampValue());
+      } else if (literal.type().equals(ExprCoreType.DATE)) {
+        return openSearchDateType.hasNoFormatter()
+            ? literal.value()
+            : openSearchDateType.getFormattedDate(literal.dateValue());
+      } else if (literal.type().equals(ExprCoreType.TIME)) {
+        return openSearchDateType.hasNoFormatter()
+            ? literal.value()
+            : openSearchDateType.getFormattedDate(literal.timeValue());
+      }
```
penghuo, line 67: can we avoid duplicate code with TermQuery?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
```diff
@@ -235,6 +244,71 @@ public List<DateFormatter> getAllCustomFormatters() {
...
+    List<DateFormatter> dateFormatters = this.getAllNamedFormatters();
+    dateFormatters.addAll(this.getAllCustomFormatters());
+    ZonedDateTime zonedDateTime = null;
+
+    // check if dateFormatters are empty, then set default ones
+    if (dateFormatters.isEmpty()) {
+      dateFormatters = initializeDateFormatters();
+    }
```
penghuo, line 262:
```
List<DateFormatter> dateFormatters = this.getAllNamedFormatters();
dateFormatters.addAll(this.getAllCustomFormatters());
```
this code will duplicate getAllCustomFormatters?

use Stream.concat to concat two list works?

penghuo: formats is fetched from OpenSearch index mapping, if formats is null, does it means the OpenSearchDataType is  a wrapper of CoreType. So in this case, should we use OPENSEARCH_DEFAULT_FORMATS to format to parse the data?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
```diff
@@ -235,6 +244,71 @@ public List<DateFormatter> getAllCustomFormatters() {
...
+   *
+   * @param accessor The TemporalAccessor object containing the date/time information.
+   * @return A formatted date string if a formatter is available, otherwise null.
+   */
+  public String getFormattedDate(TemporalAccessor accessor) {
+    if (hasNoFormatter()) {
+      return DateFormatter.forPattern(OPENSEARCH_DEFAULT_FORMATS.get(0)).format(accessor);
+    }
+    // Use the first available format string to create the formatter
+    return DateFormatter.forPattern(this.formats.get(0)).format(accessor);
```
penghuo, line 287: why we use first format to format the data, but in line 204, use for loop to parse the string?

opensearch/src/test/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateTypeTest.java
```diff
@@ -259,4 +263,158 @@ public void check_if_date_type_compatible() {
...
+    String timestamp = "2021-11-08 17:00:00";
+    String format = "strict_date_time_no_millis || epoch_millis";
+    OpenSearchDateType dateType = OpenSearchDateType.of(format);
+    assertNull(dateType.getParsedDateTime(timestamp));
+  }
+
+  @Test
+  void test_valid_date_with_custom_formatter() {
+    String dateString = "2021-11-08";
+    String format = "yyyy-MM-dd";
```
LantaoJin, line 372: Could you test some date strings such as `23-Oct-17`?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -102,7 +106,7 @@ private boolean literalExpressionWrappedByCast(FunctionExpression func) {
    * @return query
    */
   public QueryBuilder build(FunctionExpression func) {
-    ReferenceExpression ref = (ReferenceExpression) func.getArguments().get(0);
+    this.ref = (ReferenceExpression) func.getArguments().get(0);
```
LantaoJin, line 109: Will this class be used in multi-threads? `ref` in `LuceneQuery` is shared member in multi-threads.
I think you can change to:
1. Add a parameter `ReferenceExpression ref` in `getParsedDateTime()`
```java
private ZonedDateTime getParsedDateTime(LiteralExpression expr, ReferenceExpression ref)
```
2. Change `cast()` method to
```java
  private ExprValue cast(FunctionExpression castFunction, ReferenceExpression ref) {
    return castMap
        .get(castFunction.getFunctionName())
        .apply((LiteralExpression) castFunction.getArguments().get(0), ref);
  }
```
3. Change the `castMap` with BiFunction
```java
private final Map<FunctionName, BiFunction<LiteralExpression, ReferenceExpression, ExprValue>> castMap =
         ...
         .put(
              BuiltinFunctionName.CAST_TO_DATE.getName(),
              (expr, ref) -> {
                if (expr.type().equals(ExprCoreType.STRING)) {
                  ZonedDateTime zonedDateTime = getParsedDateTime(expr, ref);
                  if (zonedDateTime != null) {
                    return new ExprDateValue(zonedDateTime.toLocalDate());
                  }
                  return new ExprDateValue(expr.valueOf().stringValue());
                } else {
                  return new ExprDateValue(expr.valueOf().dateValue());
                }
              })
```

LantaoJin: Since only `ref.type()` used in `getParsedDateTime`,
Pass a `ExprType` parameter instead of a `ReferenceExpression` in above solution would be better in case serialize/deserialize problem in lambda function.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
manasvinibs: Added the doc

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
manasvinibs: When format is null, it can still be `OpenSearchDateType` field as there can be date fields without formats (default ones) defined. So I have used OPENSEARCH_DEFAULT_FORMATS to format the date object in those cases.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java
manasvinibs: In for loop, I'm trying to parse the date string with available formatters and to do that I need to find exact matching pattern of the date string. But here when we format the date object to date string to send to DSL, we can format in any available formats for OpenSearch to accept.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
manasvinibs: Good call! Updated

opensearch/src/test/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateTypeTest.java
manasvinibs: Done

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java
```diff
@@ -230,6 +234,9 @@ public String legacyTypeName() {
     if (mappingType == null) {
       return exprCoreType.typeName();
     }
+    if (mappingType.toString().equalsIgnoreCase("DATE")) {
+      return exprCoreType.typeName();
+    }
```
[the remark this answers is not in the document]
manasvinibs: I think we cannot move this check from `legacyTypeName()` to `typeName()` function as it will fail IT tests. I think `legacyTypeName` is used to serialize SQL response on which few IT tests depends on and `typeName` is used to map and serialize PPL response.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java
```diff
@@ -230,6 +234,9 @@ public String legacyTypeName() {
     if (mappingType == null) {
       return exprCoreType.typeName();
     }
+    if (mappingType.toString().equalsIgnoreCase("DATE")) {
```
penghuo, line 237: change to mappingType == Date?

opensearch/src/test/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataTypeTest.java
```diff
@@ -104,8 +104,8 @@ private static Stream<Arguments> getTestDataWithType() {
         Arguments.of(MappingType.ScaledFloat, "scaled_float", DOUBLE),
         Arguments.of(MappingType.Double, "double", DOUBLE),
         Arguments.of(MappingType.Boolean, "boolean", BOOLEAN),
-        Arguments.of(MappingType.Date, "date", TIMESTAMP),
-        Arguments.of(MappingType.DateNanos, "date", TIMESTAMP),
+        Arguments.of(MappingType.Date, "timestamp", TIMESTAMP),
```
penghuo, line 107: Please double confirm on UT chagne. does all these date -> timestamp change necessary?

penghuo (approved):
Please double confirm on the UT assert change,

opensearch/src/test/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataTypeTest.java
manasvinibs: All these assertions directly comparing with `legacyTypeName` and hence need this update. I think these changes are necessary and only in UT to avoid any breaking changes with type change from `ExprCoreType` to `OpenSearchDateType` inside `fieldMappings` map.

ykmr1224 (approved):

LantaoJin (approved):

opensearch/src/test/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataTypeTest.java
LantaoJin: I only find the legacy name change for `Date` in production code, why the `DateNanos` changed in test either?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java
LantaoJin: Should we check mappingType == DateNanos either?
```java
if (mappingType == Date || mappingType == DateNanos) {
```

LantaoJin:
https://github.com/opensearch-project/sql/pull/2762/files/ea3e205184dda4e8e0d5ba55cd9cf18eac82c30e#r1683753172

opensearch/src/test/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataTypeTest.java
manasvinibs: I think this is because when we create `OpenSearchDataType` instance for every field mapping type, we are creating `OpenSearchDateType` instance for both `Mappingtype.Date` and `MappingType.DateNanos`- https://github.com/opensearch-project/sql/blob/main/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java#L165 
and set the mapping type to Date even for DateNanos (not sure why is it this way) - https://github.com/opensearch-project/sql/blob/main/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDateType.java#L143
So I think here UT assertion still looking for Mapping type date even for date nanos I believe.

PR #2774, opened 2024-06-24
Introduce SparkParameterComposerCollection

ykmr1224:
### Description
- Introduce SparkParameterComposerCollection to abstract the datasource type specific logics from async-query-core
- DatasourceType specific parameters can be implemented in DataSourceSparkParameterComposer and registered to SparkParameterComposerCollection.
- SparkParameterComposerCollection will hold all the implementation of DataSourceSparkParameterComposer and GeneralSparkParameterComposer, and apply them when `composeByDatasource` and `compose` is called.
- I'm planning to remove SparkSubmitParameterModifier in separate PR since Composer will cover the use case.
 
### Issues Resolved
n/a
 
### Check List


async-query-core/src/main/java/org/opensearch/sql/spark/parameter/SparkSubmitParameters.java
```diff
@@ -0,0 +1,52 @@
...
+package org.opensearch.sql.spark.parameter;
+
+import static org.opensearch.sql.spark.data.constants.SparkConstants.DEFAULT_CLASS_NAME;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import lombok.Setter;
+
+/** Define Spark Submit Parameters. */
+public class SparkSubmitParameters {
```
ykmr1224, line 15: Moved from: 
async-query-core/src/main/java/org/opensearch/sql/spark/asyncquery/model/SparkSubmitParameters.java
And builder was extracted as separate class.

async-query-core/src/main/java/org/opensearch/sql/spark/parameter/SparkSubmitParametersBuilder.java
```diff
@@ -0,0 +1,171 @@
...
+
+import lombok.Getter;
+import org.apache.commons.text.StringEscapeUtils;
+import org.opensearch.sql.datasource.model.DataSourceMetadata;
+import org.opensearch.sql.spark.asyncquery.model.AsyncQueryRequestContext;
+import org.opensearch.sql.spark.config.SparkSubmitParameterModifier;
+import org.opensearch.sql.spark.dispatcher.model.DispatchQueryRequest;
+import org.opensearch.sql.spark.execution.statestore.OpenSearchStateStoreUtil;
+
+public class SparkSubmitParametersBuilder {
```
ykmr1224, line 58: The logic in this builder was originally within SparkSubmitParameters class.
S3GLUE related configs are extracted to S3GlueDataSourceSparkParameterComposer

async-query-core/src/main/java/org/opensearch/sql/spark/parameter/DataSourceSparkParameterComposer.java
```diff
@@ -0,0 +1,19 @@
...
+
+package org.opensearch.sql.spark.parameter;
+
+import org.opensearch.sql.datasource.model.DataSourceMetadata;
+import org.opensearch.sql.spark.asyncquery.model.AsyncQueryRequestContext;
+import org.opensearch.sql.spark.dispatcher.model.DispatchQueryRequest;
+
+/** Compose Spark parameter based on DataSourceMetadata */
+public interface DataSourceSparkParameterComposer {
+  void compose(
```
penghuo, line 14: no return value? does it modify sparkSubmitParameters?

async-query-core/src/main/java/org/opensearch/sql/spark/parameter/SparkParameterComposerCollection.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Map;
+import org.opensearch.sql.datasource.model.DataSourceMetadata;
+import org.opensearch.sql.datasource.model.DataSourceType;
+import org.opensearch.sql.spark.asyncquery.model.AsyncQueryRequestContext;
+import org.opensearch.sql.spark.dispatcher.model.DispatchQueryRequest;
+
+/** Stores Spark parameter composers and dispatch compose request to each composer */
+public class SparkParameterComposerCollection {
+  Collection<GeneralSparkParameterComposer> generalComposers = new ArrayList<>();
+  Map<DataSourceType, Collection<DataSourceSparkParameterComposer>> datasourceComposers =
```
penghuo, line 22: why need map? how about Collection<DataSourceSparkParameterComposer>?

async-query-core/src/main/java/org/opensearch/sql/spark/parameter/SparkSubmitParametersBuilder.java
penghuo: is it GeneralSparkParameterComposer? Hive and S3 relation configuration should in datasource composer?

async-query/src/main/java/org/opensearch/sql/spark/config/SparkExecutionEngineConfigClusterSettingLoader.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Optional;
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
+import org.opensearch.sql.common.setting.Settings;
+
+@RequiredArgsConstructor
+public class SparkExecutionEngineConfigClusterSettingLoader {
```
penghuo, line 18: javadoc

async-query/src/main/java/org/opensearch/sql/spark/config/OpenSearchExtraParameterComposer.java
```diff
@@ -0,0 +1,30 @@
...
+@RequiredArgsConstructor
+public class OpenSearchExtraParameterComposer implements GeneralSparkParameterComposer {
+  private final SparkExecutionEngineConfigClusterSettingLoader settingLoader;
+
+  @Override
+  public void compose(
+      SparkSubmitParameters sparkSubmitParameters,
+      DispatchQueryRequest dispatchQueryRequest,
+      AsyncQueryRequestContext context) {
+    settingLoader
```
penghuo, line 24: do we need Setting loader, why not directly access setting in here.

async-query/src/main/java/org/opensearch/sql/spark/config/OpenSearchSparkSubmitParameterModifier.java
```diff
@@ -6,15 +6,15 @@
...
 
 @AllArgsConstructor
 public class OpenSearchSparkSubmitParameterModifier implements SparkSubmitParameterModifier {
 
   private String extraParameters;
 
   @Override
-  public void modifyParameters(SparkSubmitParameters parameters) {
-    parameters.setExtraParameters(this.extraParameters);
+  public void modifyParameters(SparkSubmitParametersBuilder builder) {
```
penghuo, line 17: could u explain more on Modifier and Composer and add to PR description. are these necessary? could we simply the concept?

async-query/src/main/java/org/opensearch/sql/spark/config/SparkExecutionEngineConfigSupplierImpl.java
```diff
@@ -6,20 +6,17 @@
...
-import org.apache.commons.lang3.StringUtils;
 import org.opensearch.cluster.ClusterName;
 import org.opensearch.sql.common.setting.Settings;
 import org.opensearch.sql.spark.asyncquery.model.AsyncQueryRequestContext;
 
 @AllArgsConstructor
 public class SparkExecutionEngineConfigSupplierImpl implements SparkExecutionEngineConfigSupplier {
 
-  private Settings settings;
+  private final Settings settings;
```
penghuo, line 18: in case we proposal setting loader, could we also unify setting loader and config supplier concept?

async-query-core/src/main/java/org/opensearch/sql/spark/parameter/GeneralSparkParameterComposer.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.spark.parameter;
+
+import org.opensearch.sql.spark.asyncquery.model.AsyncQueryRequestContext;
+import org.opensearch.sql.spark.dispatcher.model.DispatchQueryRequest;
+
+/** Compose spark submit parameters. See {@link SparkParameterComposerCollection}. */
+public interface GeneralSparkParameterComposer {
```
penghuo, line 12: why we want to differenate General and DataSource composer, they are both Function<SparkSubmitParamaters, SparkParamaters>, right?

async-query-core/src/main/java/org/opensearch/sql/spark/parameter/SparkParameterComposerCollection.java
ykmr1224: We want to lookup by DataSourceType. Do you mean we want to lookup using linear search?

async-query/src/main/java/org/opensearch/sql/spark/config/OpenSearchExtraParameterComposer.java
ykmr1224: Since it had access from two places, I extracted as a class to avoid redundancy. (It include AccessController.doPrivileged, etc.)

async-query/src/main/java/org/opensearch/sql/spark/config/SparkExecutionEngineConfigSupplierImpl.java
ykmr1224: The setting loader was very specific to the SparkExecutionEngineConfigClusterSetting so that we don't need to write duplicate code to verify privilege and load the setting from two place.

async-query-core/src/main/java/org/opensearch/sql/spark/parameter/GeneralSparkParameterComposer.java
ykmr1224: As discussed offline, DataSource composer is specific to Datasource related parameters, and come with DataSource Type validation (If there is no composer for the DataSource type, it would raise exception).
Added comments to make it clear.

async-query/src/main/java/org/opensearch/sql/spark/config/OpenSearchSparkSubmitParameterModifier.java
ykmr1224: Added description. I am planning to delete Modifier since Generic composer will cover the use case.

async-query-core/src/main/java/org/opensearch/sql/spark/config/SparkSubmitParameterModifier.java
```diff
@@ -1,11 +1,12 @@
...
-import org.opensearch.sql.spark.asyncquery.model.SparkSubmitParameters;
+import org.opensearch.sql.spark.parameter.SparkSubmitParametersBuilder;
 
 /**
  * Interface for extension point to allow modification of spark submit parameter. modifyParameter
- * method is called after the default spark submit parameter is build.
+ * method is called after the default spark submit parameter is build. To be deprecated in favor of
+ * {@link org.opensearch.sql.spark.parameter.GeneralSparkParameterComposer}
  */
 public interface SparkSubmitParameterModifier {
```
vamsi-amazon, line 10: When does modifier come into picture? So we have two concepts composer and modifier? Modifier is used after the completion of building spark submit parameters using composer.

ykmr1224: I am thinking to remove Modifier once I can verify the Composer can satisfy the use cases.

vamsi-amazon:
Please resolve the conflict.

async-query-core/src/main/java/org/opensearch/sql/spark/parameter/SparkParameterComposerCollection.java
penghuo: no, i mean why not use Collection of DataSourceSparkParameterComposer.
understood now,

penghuo (dismissed):

vamsi-amazon (dismissed):

vamsi-amazon (dismissed):

penghuo (approved):

vamsi-amazon (approved):

PR #2783, opened 2024-06-27
Pass accountId to EMRServerlessClientFactory.getClient

ykmr1224:
### Description
- Pass accountId to EMRServerlessClientFactory.getClient so that implementation of EMRServerlessClientFactory can decide the EMR cluster based on that.
 
### Issues Resolved
n/a
 
### Check List
  - [n/a] New functionality has user manual doc added

### DIFF
--- a/async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactory.java
+++ b/async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactory.java
@@ -11,7 +11,8 @@ public interface EMRServerlessClientFactory {
   /**
    * Gets an instance of {@link EMRServerlessClient}.
    *
+   * @param accountId Account ID of the requester. It will be used to decide the cluster.
    * @return An {@link EMRServerlessClient} instance.
    */
-  EMRServerlessClient getClient();
+  EMRServerlessClient getClient(String accountId);
 }
--- a/async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactoryImpl.java
+++ b/async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactoryImpl.java
@@ -18,7 +18,6 @@ import org.opensearch.sql.spark.config.SparkExecutionEngineConfig;
 import org.opensearch.sql.spark.config.SparkExecutionEngineConfigSupplier;
 import org.opensearch.sql.spark.metrics.MetricsService;
 
-/** Implementation of {@link EMRServerlessClientFactory}. */
 @RequiredArgsConstructor
 public class EMRServerlessClientFactoryImpl implements EMRServerlessClientFactory {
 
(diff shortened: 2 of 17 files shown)


async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactoryImpl.java
```diff
@@ -27,13 +26,8 @@ public class EMRServerlessClientFactoryImpl implements EMRServerlessClientFactor
...
   private String region;
 
-  /**
-   * Gets an instance of {@link EMRServerlessClient}.
-   *
-   * @return An {@link EMRServerlessClient} instance.
-   */
   @Override
-  public EMRServerlessClient getClient() {
+  public EMRServerlessClient getClient(String accountId) {
```
vamsi-amazon, line 30: Where are we using the accountId? would it only be used in managed service?

vamsi-amazon: // TODO: It does not handle accountId for now. (it creates client for same account)
There is this TODO in this file, is this related?

noCharger: Also cluster name should include accountID for managed service

async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactory.java
```diff
@@ -11,7 +11,8 @@ public interface EMRServerlessClientFactory {
   /**
    * Gets an instance of {@link EMRServerlessClient}.
    *
+   * @param accountId Account ID of the requester. It will be used to decide the cluster.
    * @return An {@link EMRServerlessClient} instance.
    */
-  EMRServerlessClient getClient();
+  EMRServerlessClient getClient(String accountId);
```
vamsi-amazon, line 17: This seems odd passing a parameter without using it. Can we make existing code use accountId?

async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactoryImpl.java
ykmr1224: Yes, it would be used only for managed service. The TODO is mentioning different thing. I think the intention of TODO is to make it possible to use EMRS cluster in different account.
The accountId passed here is requester's accountId.

> @noCharger
What do you mean?

async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactory.java
ykmr1224: It is not needed for open source use case, since we won't have multiple EMRS cluster.

vamsi-amazon (approved):

rupal-bq (approved):

async-query-core/src/main/java/org/opensearch/sql/spark/client/EMRServerlessClientFactoryImpl.java
noCharger: @ykmr1224  https://github.com/opensearch-project/sql/blob/00f82f52dfca873f083ce0b7791a2b69bba0175c/async-query-core/src/main/java/org/opensearch/sql/spark/asyncquery/model/SparkSubmitParameters.java#L110-L111
clustername is in format of <accountid>:<domain name>

PR #2784, opened 2024-06-27
Fix SQLQueryUtils to extract multiple tables

ykmr1224:
### Description
- Fix SQLQueryUtils to extract multiple tables
  - Originally had some issue where alias was also extracted as part of table name, and this change fixes it.
  - This method will be used to identify datasources from the query.
 
### Issues Resolved
n/a
 
### Check List
- [-] New functionality has been documented.
  - [-] New functionality has user manual doc added

### DIFF
--- a/async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
+++ b/async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.sql.spark.utils;
 
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Locale;
 import lombok.Getter;
 import lombok.experimental.UtilityClass;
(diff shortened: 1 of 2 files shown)


async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
```diff
@@ -32,16 +35,15 @@
 @UtilityClass
 public class SQLQueryUtils {
 
-  // TODO Handle cases where the query has multiple table Names.
-  public static FullyQualifiedTableName extractFullyQualifiedTableName(String sqlQuery) {
+  public static List<FullyQualifiedTableName> extractFullyQualifiedTableNames(String sqlQuery) {
```
vamsi-amazon, line 38: Can we show an example SQL query where there is a bug? What is the usecase for multiple table names.

ykmr1224: The bug was in case the identifier has some more elements such as alias, version, etc.
I added test cases for that ([ref](https://github.com/opensearch-project/sql/pull/2784/files#diff-4aba5e85641e1ff025d49ef65ad0eb29a5db4bb00d4a3d3f90ccf55a925b5a8fR80-R85)).

Simplest example is:
```
SELECT * FROM my_glue.default.http_logs alias;
```
Where table name was extracted as "http_logsalias".

DQS wants to extract datasource names from the input query and verify them, and that would require all the table names to be extracted.

vamsi-amazon: How are we solving the use case with multiple table names?

ykmr1224: With this change, we collect table names to a list in the visitor return it.
I've added test cases for multi table queries.

vamsi-amazon (approved):

rupal-bq (approved):

PR #2786, opened 2024-06-28
Eliminate dependency from async-query-core to legacy

ykmr1224:
### Description
- Eliminate dependency from async-query-core to legacy
  - Extracted XContent related part and moved to async-query module
 
### Issues Resolved
n/a
 
### Check List
- [-] New functionality has been documented.
  - [-] New functionality has user manual doc added


async-query/src/main/java/org/opensearch/sql/spark/transport/format/CreateAsyncQueryRequestConverter.java
```diff
@@ -0,0 +1,46 @@
...
+
+import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+
+import lombok.experimental.UtilityClass;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.sql.spark.rest.model.CreateAsyncQueryRequest;
+import org.opensearch.sql.spark.rest.model.LangType;
+
+@UtilityClass
+public class CreateAsyncQueryRequestConverter {
```
vamsi-amazon, line 16: Can you see this commit: https://github.com/vamsi-amazon/sql/commit/fd75fcd7949240a31c18469b429ccbb2e32ba3e4#diff-b8da527871e1adbe7ec8f4e08dd75bf5b426be4d195c9b0860e09260c5dda127
Also, is this the only model left to have dependency on XContent?
Are there other models in async-query-core which have dependency on XContent.

vamsi-amazon (approved):

async-query/src/main/java/org/opensearch/sql/spark/transport/format/CreateAsyncQueryRequestConverter.java
ykmr1224: Yes, other XContent dependent classes are already in `async-query` module since they are OpenSearch dependent implementation.

rupal-bq (approved):

PR #2797, opened 2024-07-01
Fix regexp filter query

### LINKED ISSUE #2796: [BUG]  REGEXP filter failed due to IllegalStateException
**What is the bug?**
When executing a SQL query with REGEXP filter, get the following exception:
```
{
  "error": {
    "details": "Shard[0]: java.lang.IllegalStateException: Expression has wrong result type instead of boolean: expression [regexp(field1, \"test.*\")], result [1]\n\nFor more details, please send request for Json format to see the raw response from OpenSearch engine.",
  },
  "status": 500
}
```

1. create an index with the following mappings:
```
{
  "mappings": {
    "properties": {
      "field1": {
        "type": "keyword"
      }
    }
  }
}
```
2. write one doc using bulk API:
```
{"index": {"_id": 1}}
{"field1": "test search something"}
```
3. run the following SQL query:
```
{
  "query": "SELECT field1 FROM test WHERE field1 REGEXP 'test.*'"
}
```

The root cause of this bug may be because what `ExpressionFilterScript#evaluateExpression` expects is a boolean value:

https://github.com/opensearch-project/sql/blob/43263963aca2ada949d55ae01cd38788d5d96fbb/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java#L51

but what `OperatorUtils#matchesRegexp` returns is an integer value:

https://github.com/opensearch-project/sql/blob/43263963aca2ada949d55ae01cd38788d5d96fbb/core/src/main/java/org/opensearch/sql/utils/OperatorUtils.java#L37-L39

If we modify the return type of `OperatorUtils#matchesRegexp`, it will impact the return type of the REGEXP function, which could be a breaking change. To avoid this, we could consider converting the integer value to a boolean value in `ExpressionFilterScript#evaluateExpression`.

bugmakerrrrrr:
### Description
Fix #2796 .
 
### Issues Resolved
#2796 
 
### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
@@ -17,7 +17,6 @@ import org.opensearch.sql.expression.Expression;
 import org.opensearch.sql.expression.FunctionExpression;
 import org.opensearch.sql.expression.env.Environment;
 import org.opensearch.sql.expression.function.BuiltinFunctionName;
-import org.opensearch.sql.expression.function.FunctionImplementation;
 import org.opensearch.sql.opensearch.storage.script.core.ExpressionScript;
 
 /**
@@ -72,7 +71,7 @@ class ExpressionFilterScript extends FilterScript {
   }
 
   private boolean isRegexpExpression(Expression expression) {
-    return expression instanceof FunctionImplementation
+    return expression instanceof FunctionExpression
         && ((FunctionExpression) expression)
             .getFunctionName()
             .equals(BuiltinFunctionName.REGEXP.getName());
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
@@ -48,6 +48,15 @@ class ExpressionFilterScript extends FilterScript {
       return ExprBooleanValue.of(false);
     }
 
+    // refer to https://github.com/opensearch-project/sql/issues/2796
+    if (result.type() == ExprCoreType.INTEGER) {
+      if (result.integerValue() == 0) {
+        result = ExprBooleanValue.of(false);
+      } else if (result.integerValue() == 1) {
+        result = ExprBooleanValue.of(true);
+      }
+    }
+
     if (result.type() != ExprCoreType.BOOLEAN) {
       throw new IllegalStateException(
           String.format(
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
@@ -14,7 +14,10 @@ import org.opensearch.sql.data.model.ExprBooleanValue;
 import org.opensearch.sql.data.model.ExprValue;
 import org.opensearch.sql.data.type.ExprCoreType;
 import org.opensearch.sql.expression.Expression;
+import org.opensearch.sql.expression.FunctionExpression;
 import org.opensearch.sql.expression.env.Environment;
+import org.opensearch.sql.expression.function.BuiltinFunctionName;
+import org.opensearch.sql.expression.function.FunctionImplementation;
 import org.opensearch.sql.opensearch.storage.script.core.ExpressionScript;
 
 /**
@@ -49,7 +52,8 @@ class ExpressionFilterScript extends FilterScript {
     }
 
     // refer to https://github.com/opensearch-project/sql/issues/2796
-    if (result.type() == ExprCoreType.INTEGER) {
+    if (isRegexpExpression(expression)) {
+      assert result.type() == ExprCoreType.INTEGER;
       if (result.integerValue() == 0) {
         result = ExprBooleanValue.of(false);
       } else if (result.integerValue() == 1) {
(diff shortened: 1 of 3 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
```diff
@@ -48,6 +48,15 @@ private ExprValue evaluateExpression(
       return ExprBooleanValue.of(false);
     }
 
+    // refer to https://github.com/opensearch-project/sql/issues/2796
+    if (result.type() == ExprCoreType.INTEGER) {
+      if (result.integerValue() == 0) {
+        result = ExprBooleanValue.of(false);
+      } else if (result.integerValue() == 1) {
+        result = ExprBooleanValue.of(true);
```
LantaoJin, line 56: This check will cause additional issue.
`select * from %s where 1 * 1` will be treated as `where true` and `select * from %s where 1 * 0` will be treated as `where false`. They should throw exceptions.
How about change the return value type to `ExprBooleanValue` in https://github.com/opensearch-project/sql/blob/43263963aca2ada949d55ae01cd38788d5d96fbb/core/src/main/java/org/opensearch/sql/utils/OperatorUtils.java#L37-L40

LantaoJin (changes requested):

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
bugmakerrrrrr: If we modify the return type of OperatorUtils#matchesRegexp, it will impact the return type of the REGEXP function, which could be a breaking change. For instance, the query `SELECT 'Hello!' REGEXP '.*', 'a' REGEXP 'b'` will initially yield `1,0` as the output. However, after modifying the return type, the output will be `true,false`.

bugmakerrrrrr: @LantaoJin only when the expression is a REGEXP expression will the conversion be executed. What do you think?

LantaoJin: Sounds good. The conversion scope should be limited in `REGEXP` expression.

bugmakerrrrrr: I have fixed it. Please take another look at your convenience.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
```diff
@@ -57,4 +70,11 @@ private ExprValue evaluateExpression(
     }
     return result;
   }
+
+  private boolean isRegexpExpression(Expression expression) {
+    return expression instanceof FunctionImplementation
+        && ((FunctionExpression) expression)
```
LantaoJin, line 76: minor: `instanceof FunctionExpression`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/ExpressionFilterScript.java
dai-chen: Could you explain why `REGEXP` should return integer? From the Javadoc, the implementation seems wrong? I'm thinking we should avoid adding such special logic on critical path if possible.

```
  /**
   * Checks if text matches regular expression pattern.
   *
   * @param pattern string pattern to match.
   * @return if text matches pattern returns true; else return false.
   */
  public static ExprIntegerValue matchesRegexp(ExprValue text, ExprValue pattern) {
    return new ExprIntegerValue(
        Pattern.compile(pattern.stringValue()).matcher(text.stringValue()).matches() ? 1 : 0);
  }
```

LantaoJin: > Could you explain why REGEXP should return integer?

@dai-chen I guess the reason is it's the behaviour of MySQL (https://dev.mysql.com/doc/refman/8.0/en/regexp.html#operator_regexp). The original request was opened in https://github.com/opendistro-for-elasticsearch/sql/issues/710 and its implementation introduced by https://github.com/opendistro-for-elasticsearch/sql/pull/750. It would be a breaking change if we change its return value type. Besides updating the user doc (https://github.com/opensearch-project/sql/blob/main/docs/user/dql/expressions.rst#regexp-value-test), we need a changelog file to record those breaking changes. cc @penghuo, @chloe-zh

LantaoJin: My proposal are two options:
1. Correct the return type of `matchesRegexp()` to `ExprBooleanValue` in this PR, including user doc updates, then we need a new PR to introduce a changelog to record any breaking changes.
2. Close current RP because it is not an issue. `SELECT field1 FROM test WHERE 1 = (field1 REGEXP 'test.*')` should work. The current fixing not only adds special logic but also introduces new confused behaviour (REGEXP expression returns integer in SELECT clause but boolean in WHERE clause).

bugmakerrrrrr: In my opinion, option 2 lacks user-friendliness.

dai-chen:
CI was blocked. Triggered its run now.

RyanL1997:
Hi @bugmakerrrrrr, do we still need this change?

LantaoJin:
The regexp filter returns boolean in V3. Ref https://github.com/opensearch-project/sql/issues/3487. Close this.

Issue #2859, opened 2024-07-25
[RFC] Change the default value of `plugins.query.size_limit` to 10000 (MAX_RESULT_WINDOW)


LantaoJin:
**Is your feature request related to a problem?**

The `plugins.query.size_limit` configure the maximum amount of rows returned by query. The default value is 200 which inherited from the legacy settings `opendistro.query.size_limit`. `200` is insufficient for many use cases which results incorrect outcomes:
  - When operator pushdown is not able to work
    - https://github.com/opensearch-project/sql/issues/2802 described a scenario that query with eval command caused correctness issue. 
  - When head(PPL) or limit(SQL) set greater than the default value 200.
    - https://github.com/opensearch-project/sql/issues/1774 described a scenario that query with limit 2000 which overwritten by the  default value 200.

On the other hand, `plugins.query.size_limit` is not the only threshold to protect the service from OOM issue. `plugins.query.memory_limit` which default value is 85% could be another gatekeeper.
In OpenSearch, the maximum result for searches on index controlled by `index.max_result_window`, which is 10000 by default. That is the suggested new default value of `plugins.query.size_limit`.

Goals
---
- Mitigate the impacts caused by insufficient default settings. Be aware in the vast majority of cases, users will not adjust the default value.

Non-Goals
---
- Users specify a larger number by head or LIMIT over 10000. 
  - https://github.com/opensearch-project/sql/issues/703#issuecomment-1201420214 proposed a solution that leverage `scrollRequest` to handle the case which requests docs more than 10000.
- Sort and limit operators (TopK) against a large mount of documents (for example, 9000) will consume much memory than 200.
  - https://github.com/opensearch-project/sql/issues/2857 proposed a solution to do a top-K sort.

**What solution would you like?**

Change the default value of `plugins.query.size_limit` to the default value of `index.max_result_window` (10000 by default)

**What alternatives have you considered?**

https://github.com/opensearch-project/sql/issues/703#issuecomment-1201420214 proposed a more complex solution that leverage `scrollRequest` to handle the case which requests docs more than 10000.
E.g. `source=index | head 11000` returns 11000 rows using scroll
but `source=index | fields a,b | head 11000` still returns 200 rows

**Do you have any additional context?**

1. Plugin setting: https://github.com/opensearch-project/sql/blob/main/docs/user/admin/settings.rst#plugins-query-size-limit
2. Current limitation: https://github.com/opensearch-project/sql/blob/main/docs/user/dql/basics.rst#limitation

PR #2860, opened 2024-07-25
Change the default value of plugins.query.size_limit to MAX_RESULT_WINDOW (10000)

LantaoJin:
### Description
The `plugins.query.size_limit` configure the maximum amount of rows returned by query. The default value is 200 which inherited from the legacy settings `opendistro.query.size_limit`. `200` is insufficient for many use cases which results incorrect outcomes. Check https://github.com/opensearch-project/sql/issues/2859 for more details.

This PR changes the default value of `plugins.query.size_limit` to the default value of `index.max_result_window` (10000 by default)
 
### Issues Resolved
Resolves https://github.com/opensearch-project/sql/issues/2859
 
### Check List
- [-] New functionality includes testing.

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
@@ -28,6 +28,7 @@ import org.opensearch.common.settings.SecureSetting;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.unit.MemorySizeValue;
 import org.opensearch.common.unit.TimeValue;
+import org.opensearch.index.IndexSettings;
 import org.opensearch.sql.common.setting.LegacySettings;
 import org.opensearch.sql.common.setting.Settings;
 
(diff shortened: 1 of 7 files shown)


penghuo (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
```diff
@@ -90,7 +91,7 @@ public class OpenSearchSettings extends Settings {
   public static final Setting<?> QUERY_SIZE_LIMIT_SETTING =
       Setting.intSetting(
           Key.QUERY_SIZE_LIMIT.getKeyValue(),
-          LegacyOpenDistroSettings.QUERY_SIZE_LIMIT_SETTING,
+          IndexSettings.MAX_RESULT_WINDOW_SETTING,
```
dai-chen, line 94: I forgot what does this `fallbackSetting` argument mean. What's the difference between this and changing the default value 200 in `LegacyOpenDistroSettings`?

LantaoJin: In don't think we should touch the default value of any LegacyOpenDistroSettings, although the effects are same.

dai-chen (approved):
Thanks for the fix!

Issue #2857, opened 2024-07-24
[FEATURE] Top-K  enhancement when having PPL pattern like `source | sort $field | head k`


qianheng-aws:
**Is your feature request related to a problem?**
Currently, when PPL having a `sort` and a followed `head` in its query, it will using `SortOperator` and `LimitOperator` to calculate the final result if they are not being pushed down into the OpenSearch DSL. As shown in this issue https://github.com/opensearch-project/sql/issues/2802

```
POST _plugins/_ppl/_explain
{
  "explain": true,
  "query": """
    source=opensearch_dashboards_sample_data_flights
    | eval FlightMin = FlightTimeMin 
    | sort -FlightMin | head 5 | fields FlightMin
  """
}

# explain results
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[FlightMin]"
    },
    "children": [
      {
        "name": "LimitOperator",
        "description": {
          "limit": 5,
          "offset": 0
        },
        "children": [
          {
            "name": "SortOperator",
            "description": {
              "sortList": {
                "FlightMin": {
                  "sortOrder": "DESC",
                  "nullOrder": "NULL_LAST"
                }
              }
            },
            "children": [
              {
                "name": "EvalOperator",
                "description": {
                  "expressions": {
                    "FlightMin": "FlightTimeMin"
                  }
                },
                "children": [
                  {
                    "name": "OpenSearchIndexScan",
                    "description": {
[...]

PR #2863, opened 2024-07-26
add TakeOrderedOperator

qianheng-aws:
### Description
add TakeOrderedOperator 
 
### Issues Resolved
https://github.com/opensearch-project/sql/issues/2857
 
### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/planner/DefaultImplementor.java
+++ b/core/src/main/java/org/opensearch/sql/planner/DefaultImplementor.java
@@ -38,6 +38,7 @@ import org.opensearch.sql.planner.physical.RareTopNOperator;
 import org.opensearch.sql.planner.physical.RemoveOperator;
 import org.opensearch.sql.planner.physical.RenameOperator;
 import org.opensearch.sql.planner.physical.SortOperator;
+import org.opensearch.sql.planner.physical.TakeOrderedOperator;
 import org.opensearch.sql.planner.physical.ValuesOperator;
 import org.opensearch.sql.planner.physical.WindowOperator;
 import org.opensearch.sql.storage.read.TableScanBuilder;
@@ -129,7 +130,13 @@ public class DefaultImplementor<C> extends LogicalPlanNodeVisitor<PhysicalPlan, 
 
   @Override
   public PhysicalPlan visitLimit(LogicalLimit node, C context) {
-    return new LimitOperator(visitChild(node, context), node.getLimit(), node.getOffset());
+    PhysicalPlan child = visitChild(node, context);
+    // Optimize sort + limit to take ordered operator
+    if (child instanceof SortOperator sortChild) {
+      return new TakeOrderedOperator(
+          sortChild.getInput(), node.getLimit(), node.getOffset(), sortChild.getSortList());
+    }
+    return new LimitOperator(child, node.getLimit(), node.getOffset());
   }
 
   @Override
--- a/core/src/main/java/org/opensearch/sql/planner/physical/SortHelper.java
+++ b/core/src/main/java/org/opensearch/sql/planner/physical/SortHelper.java
@@ -15,11 +15,21 @@ import org.opensearch.sql.expression.Expression;
 
 public interface SortHelper {
 
+  /**
+   * Construct an expr comparator for sorting on ExprValue.
+   * @param sortList list of sort fields and their related sort options.
+   * @return A comparator for ExprValue
+   */
   static Comparator<ExprValue> constructExprComparator(
       List<Pair<SortOption, Expression>> sortList) {
     return (o1, o2) -> compareWithExpressions(o1, o2, constructComparator(sortList));
   }
 
+  /**
+   * Construct an expr ordering for efficiently taking the top-k elements on ExprValue.
+   * @param sortList list of sort fields and their related sort options.
+   * @return An guava ordering for ExprValue
+   */
   static Ordering<ExprValue> constructExprOrdering(List<Pair<SortOption, Expression>> sortList) {
     return Ordering.from(constructExprComparator(sortList));
   }
--- a/core/src/main/java/org/opensearch/sql/planner/physical/SortHelper.java
+++ b/core/src/main/java/org/opensearch/sql/planner/physical/SortHelper.java
@@ -1,0 +1,52 @@
+package org.opensearch.sql.planner.physical;
+
+import static org.opensearch.sql.ast.tree.Sort.NullOrder.NULL_FIRST;
+import static org.opensearch.sql.ast.tree.Sort.SortOrder.ASC;
+
+import com.google.common.collect.Ordering;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import org.apache.commons.lang3.tuple.Pair;
+import org.opensearch.sql.ast.tree.Sort.SortOption;
+import org.opensearch.sql.data.model.ExprValue;
+import org.opensearch.sql.data.utils.ExprValueOrdering;
+import org.opensearch.sql.expression.Expression;
+
+public class SortHelper {
+
+  public static Comparator<ExprValue> constructExprComparator(List<Pair<SortOption, Expression>> sortList) {
+    return (o1, o2) -> compareWithExpressions(o1, o2, constructComparator(sortList));
+  }
+
+  public static Ordering<ExprValue> constructExprOrdering(List<Pair<SortOption, Expression>> sortList) {
+    return Ordering.from(constructExprComparator(sortList));
+  }
+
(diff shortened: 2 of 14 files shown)


core/src/main/java/org/opensearch/sql/planner/DefaultImplementor.java
```diff
@@ -129,7 +130,13 @@ public PhysicalPlan visitValues(LogicalValues node, C context) {
 
   @Override
   public PhysicalPlan visitLimit(LogicalLimit node, C context) {
-    return new LimitOperator(visitChild(node, context), node.getLimit(), node.getOffset());
+    PhysicalPlan child = visitChild(node, context);
+    // Optimize sort + limit to take ordered operator
+    if (child instanceof SortOperator sortChild) {
```
penghuo, line 135: have u consider add a rule in LogicalPlanOptimizer?

core/src/main/java/org/opensearch/sql/planner/physical/TakeOrderedOperator.java
```diff
@@ -0,0 +1,128 @@
...
+  public ExprValue next() {
+    return iterator.next();
+  }
+
+  @Builder
+  public static class ExprOrdering extends Ordering<ExprValue> {
+    @Singular private final List<Pair<Expression, Comparator<ExprValue>>> comparators;
+
+    @Override
+    public int compare(ExprValue o1, ExprValue o2) {
```
penghuo, line 113: is it possible to reuse code of SortOperator.Sorter?

core/src/main/java/org/opensearch/sql/planner/physical/TakeOrderedOperator.java
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * TakeOrdered Operator Constructor.
+   *
+   * @param input input {@link PhysicalPlan}
+   * @param limit the limit value from LimitOperator
+   * @param offset the offset value from LimitOperator
+   * @param sortList list of sort field from SortOperator
+   */
+  public TakeOrderedOperator(
```
penghuo, line 59: is it possible to combine SortOperator and TakeOrderedOperator. Both operator has next interface return the sorted valued, the difference is TakeOrderedOperator is more efficiency with limit + offset.

core/src/main/java/org/opensearch/sql/planner/DefaultImplementor.java
qianheng-aws: Yeah, I've considered doing this optimization via a rule in logical plan in the beginning, then I should also add an new logical operator like LogicalTakeOrdered to replace LogicalLimit + LogicalSort. 

But I find it may be not proper to do that because this is an physical execution level optimization. The logical plan of PPL or SQL is not only able to be executed on OpenSearch,  it could also be translated to execute on other engine like Spark(opensearch-spark supports this) or something else in the future. But usually, they don't have such logical or physical operator like TakeOrdered and will make this logical plan incompatible.

I think this optimization is all about execution and deeply bound to execution engine, so I choose to do that replacement in physical plan.

core/src/main/java/org/opensearch/sql/planner/physical/TakeOrderedOperator.java
qianheng-aws: TakeOrderedOperator could cover the usage of SortOperator functionally if we set limit to MAX_VALUE and offset to 0.

But it may be a little bit confusing to do that because TakeOrderedOperator aims to do calculating for taking Top-K. So these 2 operators should have different operator name and parameters when users look into plan explain. 

They may be confusing to see explain with TakeOrderedOperator whose parameters contains limit =  MAX_VALUE and offset = 0, if they just want to sort without limit. In turn, it's also confusing to see explain with only SortOperator  whose parameters contain limit and offset while not seeing LimitOperator, if they want to sort + limit k.

core/src/main/java/org/opensearch/sql/planner/physical/TakeOrderedOperator.java
qianheng-aws: It could be, we may need to put this comparator in a class like SortHelper so both TakeOrdered and Sort can access to it. I'll implement that.

core/src/main/java/org/opensearch/sql/planner/physical/TakeOrderedOperator.java
penghuo: i see, make sense

core/src/main/java/org/opensearch/sql/planner/physical/SortHelper.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.List;
+import org.apache.commons.lang3.tuple.Pair;
+import org.opensearch.sql.ast.tree.Sort.SortOption;
+import org.opensearch.sql.data.model.ExprValue;
+import org.opensearch.sql.data.utils.ExprValueOrdering;
+import org.opensearch.sql.expression.Expression;
+
+public interface SortHelper {
+
+  static Comparator<ExprValue> constructExprComparator(
```
penghuo, line 18: nit: add comments.

penghuo, line 18: nit: add comments

core/src/main/java/org/opensearch/sql/planner/physical/TakeOrderedOperator.java
```diff
@@ -0,0 +1,88 @@
...
+      PhysicalPlan input,
+      Integer limit,
+      Integer offset,
+      List<Pair<SortOption, Expression>> sortList) {
+    this.input = input;
+    this.sortList = sortList;
+    this.limit = limit;
+    this.offset = offset;
+    this.ordering = SortHelper.constructExprOrdering(sortList);
```
penghuo, line 60: good job on simpification, code is easy to read

penghuo (approved):
LGTM, thanks for the change!
@dai-chen  please take a look.

core/src/main/java/org/opensearch/sql/planner/DefaultImplementor.java
penghuo: make sense.

core/src/test/java/org/opensearch/sql/planner/physical/TakeOrderedOperatorTest.java
```diff
@@ -0,0 +1,603 @@
...
+                inputPlan, 2, 0, Pair.of(SortOption.DEFAULT_ASC, ref("response", INTEGER)))),
+        contains(
+            tupleValue(ImmutableMap.of("size", 499, "response", 404)),
+            tupleValue(ImmutableMap.of("size", 320, "response", 404))));
+
+    wrapper.setIterator(inputList.iterator());
+    assertThat(
+        execute(
+            takeOrdered(
+                inputPlan, 2, 1, Pair.of(SortOption.DEFAULT_ASC, ref("response", INTEGER)))),
```
LantaoJin, line 141: Could you add a test which its limit equals its offset?

qianheng-aws: Oh, it comes to me that we may not do that test since Limit(offset, limit) in OpenSearch has different semantics with other engine. For OpenSearch, limit stands for the size of `HEAD` and offset stands for the from of `HEAD`.
So it will always return `limit` rows no matter which value `offset` is.
e.g. 
```
POST _plugins/_ppl
{
  "query": """
    source=opensearch_dashboards_sample_data_flights| fields FlightTimeMin | head 2 from 2
  """
}

# response
{
  "schema": [
    {
      "name": "FlightTimeMin",
      "type": "float"
    }
  ],
  "datarows": [
    [
      0
    ],
    [
      222.74905
    ]
  ],
  "total": 2,
  "size": 2
}

```

LantaoJin: > So it will always return limit rows no matter which value offset is.

Will `skip(offset)` in `open()` method of TakeOrderedOperator` change the current behavior?
It comes to me better to add a test to compare the results between take-ordered and limit+head.

qianheng-aws: Sorry, it's not exact right to say `it will always return limit rows no matter which value offset is` actually. It's right only when the size of input rows is more than `offset + limit`. 

I think `skip(offset)` is semantic equal to what LimitOperator do in its `open()`. I'll add test to verify this.

dai-chen (approved):
Thanks for the changes!

In future, I think it'd great to have micro-benchmark in [benchmarks](https://github.com/opensearch-project/sql/tree/main/benchmarks) module to demonstrate such improvements.

LantaoJin (approved):
LGTM

Issue #2901, opened 2024-08-06
[FEATURE] Add more details in explain for ProjectOperator


qianheng-aws:
**Is your feature request related to a problem?**
Currently, in an explain for a sql which produces ProjectOperator, it only shows the field information of ProjectOperator.

However, if a sql-produced ProjectOperator contains alias, its original delegate information is missing in the explain, thus users cannot tell whether the plan is right or not.
e.g.
```
POST _plugins/_sql/_explain
{
  "query": "SELECT (FlightTimeMin + 1) as FlightTimeMinPlusOne FROM opensearch_dashboards_sample_data_flights LIMIT 5"
}


# explain response
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[FlightTimeMinPlusOne]"
    },
    "children": [
      {
        "name": "OpenSearchIndexScan",
        "description": {
          "request": """OpenSearchQueryRequest(indexName=opensearch_dashboards_sample_data_flights, sourceBuilder={"from":0,"size":5,"timeout":"1m","_source":{"includes":["FlightTimeMin"],"excludes":[]}}, searchDone=false)"""
        },
        "children": []
      }
    ]
  }
}
```

**What solution would you like?**
Show the information of `name` which is `(FlightTimeMin + 1)` or `delegate` which is `+(FlightTimeMin, 1)` in the explain of ProjectOperators. 

**What alternatives have you considered?**

**Do you have any additional context?**

dblock:
[Catch All Triage - [1](https://github.com/andrross), [2](https://github.com/dblock), [3](https://github.com/AmiStrn), [4](https://github.com/reta), [5](https://github.com/finnegancarroll)]

PR #2911, opened 2024-08-07
Make explain of ProjectOperator with alias show both name and alias info

qianheng-aws:
### Description
Add name information as well in the explain of ProjectOperator with alias, so user can identify whether the field in his plan comes from the right source.

```
POST _plugins/_sql/_explain
{
  "query": "SELECT (FlightTimeMin + 1) as FlightTimeMinPlusOne FROM opensearch_dashboards_sample_data_flights LIMIT 5"
}


#Before this PR
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[FlightTimeMinPlusOne]"
    },
    "children": [
      {
        "description": {
          "request": """OpenSearchQueryRequest(indexName=opensearch_dashboards_sample_data_flights, sourceBuilder={"from":0,"size":5,"timeout":"1m","_source":{"includes":["FlightTimeMin"],"excludes":[]}}, searchDone=false)"""
        },
        "children": []
      }
    ]
  }
}

#With this PR
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[(FlightTimeMin + 1) AS FlightTimeMinPlusOne]"
    },
    "children": [
      {
        "description": {
          "request": """OpenSearchQueryRequest(indexName=opensearch_dashboards_sample_data_flights, sourceBuilder={"from":0,"size":5,"timeout":"1m","_source":{"includes":["FlightTimeMin"],"excludes":[]}}, searchDone=false)"""
        },
        "children": []
      }
    ]
  }
}

```

### Related Issues
Resolves [#2901](https://github.com/opensearch-project/sql/issues/2901)

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/NamedExpression.java
+++ b/core/src/main/java/org/opensearch/sql/expression/NamedExpression.java
@@ -60,6 +60,6 @@ public class NamedExpression implements Expression {
 
   @Override
   public String toString() {
-    return getNameOrAlias();
+    return Strings.isNullOrEmpty(alias) ? name : name + " AS " + alias;
   }
 }
--- a/integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
@@ -283,7 +283,10 @@ public class ExplainIT extends SQLIntegTestCase {
 
     String result =
         explainQuery(
-            String.format("SELECT (age + 1) AS agePlusOne FROM %s LIMIT 10", TEST_INDEX_ACCOUNT));
+            String.format(
+                "SELECT city, (age + 1) AS agePlusOne, (balance + 1000) as newBalance FROM %s LIMIT"
+                    + " 10",
+                TEST_INDEX_ACCOUNT));
     Assert.assertThat(
         result.replaceAll("\\s+", ""), equalTo(expectedOutput.replaceAll("\\s+", "")));
   }
--- a/integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
@@ -271,4 +271,20 @@ public class ExplainIT extends SQLIntegTestCase {
 
     assertEquals("application/json; charset=UTF-8", response.getHeader("content-type"));
   }
+
+  @Test
+  public void explainAlias() throws IOException {
+
+    String expectedOutputFilePath =
+        TestUtils.getResourceFilePath("src/test/resources/expectedOutput/alias_explain.json");
+    String expectedOutput =
+        Files.toString(new File(expectedOutputFilePath), StandardCharsets.UTF_8)
+            .replaceAll("\r", "");
+
+    String result =
+        explainQuery(
+            String.format("SELECT (age + 1) AS agePlusOne FROM %s LIMIT 10", TEST_INDEX_ACCOUNT));
+    Assert.assertThat(
+        result.replaceAll("\\s+", ""), equalTo(expectedOutput.replaceAll("\\s+", "")));
+  }
 }
--- a/core/src/test/java/org/opensearch/sql/executor/ExplainTest.java
+++ b/core/src/test/java/org/opensearch/sql/executor/ExplainTest.java
@@ -77,7 +77,7 @@ class ExplainTest extends ExpressionTestBase {
         new ExplainResponse(
             new ExplainResponseNode(
                 "ProjectOperator",
-                Map.of("fields", "[name, age]"),
+                Map.of("fields", "[full_name AS name, age]"),
                 singletonList(
                     new ExplainResponseNode(
                         "FilterOperator",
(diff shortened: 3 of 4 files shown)


dai-chen (approved):
Thanks for the fix!

core/src/main/java/org/opensearch/sql/expression/NamedExpression.java
```diff
@@ -60,6 +60,6 @@ public <T, C> T accept(ExpressionNodeVisitor<T, C> visitor, C context) {
 
   @Override
   public String toString() {
-    return getNameOrAlias();
+    return Strings.isNullOrEmpty(alias) ? name : name + " AS " + alias;
```
LantaoJin, line 63: Looks like you should add a new method `getNameWithAlias()` in this class similar to `getNameOrAlias()`.
And I have some concerns when to use `getNameOrAlias()` and when to use `getNameWithAlias()`?
Could you deep dive in it?

integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
```diff
@@ -271,4 +271,20 @@ public void testContentTypeOfExplainRequestShouldBeJson() throws IOException {
...
+
+    String expectedOutputFilePath =
+        TestUtils.getResourceFilePath("src/test/resources/expectedOutput/alias_explain.json");
+    String expectedOutput =
+        Files.toString(new File(expectedOutputFilePath), StandardCharsets.UTF_8)
+            .replaceAll("\r", "");
+
+    String result =
+        explainQuery(
+            String.format("SELECT (age + 1) AS agePlusOne FROM %s LIMIT 10", TEST_INDEX_ACCOUNT));
```
LantaoJin, line 286: Could you overwrite some more complex cases:
1. SELECT \<field-list\> \<alias-list\>
2. SELECT field as alias1, alias1 as alias2, alias2 as alias3 (nested alias)
3. SELECT 1 as existing-field from t (override)

core/src/main/java/org/opensearch/sql/expression/NamedExpression.java
qianheng-aws: `getNameOrAlias` are used to get the field name to use in other operator. While in its `toString`, I want to add more info than the field name.

I think we don't add such a new function as  it will only be used in this `toString` method,  we should add it unless it can be reused in other places.

integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
qianheng-aws: 1. Does SQL support such syntax? Or do you mean `filed1 as alias1, filed2 as alias2`?
2. Seems SQL don't support nested alias.
3. ditto.

core/src/main/java/org/opensearch/sql/expression/NamedExpression.java
LantaoJin: > we should add it unless it can be reused in other places as well.

I doubt are there any other places to replace to `getNameWithAlias` either as `toString` does. Can you help to confirm them?

integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
LantaoJin: 1. Compound case means not only contains pure fields but also contains fields with alias
2. It should be a bug if it's unsupported.
3. Same as above.

qianheng-aws: 1. I see. We have such cases in UT. Maybe make this IT be the compound case as well.
2. yeah. PPL supports while SQL doesn't.

LantaoJin: For 2 and 3, could you create two issues for that?

core/src/main/java/org/opensearch/sql/expression/NamedExpression.java
qianheng-aws: Having scanned over all the usage of `getNameOrAlias`, I don't see any other places need to be replaced. They are all related to execution.

integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
qianheng-aws: Sure, but I think they may be the same cause. It's because we only use ProjectOperator instead of EvalOperator for SQL, while project don't support update and evaluate nested fields like eval.

core/src/main/java/org/opensearch/sql/expression/NamedExpression.java
LantaoJin: Thanks @qianheng-aws .  @dai-chen Can u help to double confirm it since u are the author of this class?

dai-chen: Sorry I can't recall the specific details. I think it should be safe to change the `toString` method only.

integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
qianheng-aws: Create a new issue to bring out this problem here: https://github.com/opensearch-project/sql/issues/2933

RyanL1997:
@qianheng-aws Do we still need this change?

Issue #2864, opened 2024-07-29
[RFC] Push Down Sort/Limit/Project through LogicalEval


qianheng-aws:
**Is your feature request related to a problem?**
Currently in a logical plan of OpenSearch, the LogicalEval is a blocker for many push-down optimization rules like  `TableScanPushDown.PUSH_DOWN_PROJECT`,  `TableScanPushDown.PUSH_DOWN_SORT`, `TableScanPushDown.PUSH_DOWN_LIMIT` and etc. Not only does it reduce the performance of execution, it also lead to wrong results if we cannot do pushdown optimization.

E.g. Below 2 PPL are semantic equal, but the latter one with eval retrieves all fields from OpenSearch(which is unnecessary), while not enough documents due to the limitation of settings of `plugins.query.size_limit`.

```
# no eval
POST _plugins/_ppl/_explain
{
  "query": """
    source=opensearch_dashboards_sample_data_flights |  head 300 | fields FlightTimeMin
  """
}
# explain result
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[FlightTimeMin]"
    },
    "children": [
      {
        "name": "OpenSearchIndexScan",
        "description": {
          "request": """OpenSearchQueryRequest(indexName=opensearch_dashboards_sample_data_flights, sourceBuilder={"from":0,"size":200,"timeout":"1m","_source":{"includes":["FlightTimeMin"],"excludes":[]}}, searchDone=false)"""
        },
        "children": []
      }
    ]
  }
}

--------------------------------------------------------

# with eval
POST _plugins/_ppl/_explain
{
  "query": """
    source=opensearch_dashboards_sample_data_flights | eval FlightMin = FlightTimeMin |  head 300 | fields FlightTimeMin
  """
}
# explain result
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[FlightTimeMin]"
    },
    "children": [
      {
        "name": "LimitOperator",
        "description": {
          "limit": 10,
          "offset": 0
        },
        "children": [
          {
            "name": "EvalOperator",
            "description": {
              "expressions": {
                "FlightMin": "FlightTimeMin"
              }
            },
            "children": [
              {
                "name": "OpenSearchIndexScan",
                "description": {
                  "request": """OpenSearchQueryRequest(indexName=opensearch_dashboards_sample_data_flights, sourceBuilder={"from":0,"size":300,"timeout":"1m"}, searchDone=false)"""
                },
[...]

PR #2937, opened 2024-08-15
Push down sort through eval

### LINKED ISSUE #2904: push down sort

qianheng-aws:
### Description
Add a rule EvalPushDown.PUSH_DOWN_SORT, this rule will push down limit under eval. Thus, sort has chance to be pushed down into TableScanBuilder later.

For now, it only supports push specific sort and eval, which has restriction:
1. The expression in sort and replaced expression are both ReferenceExpression.
2. No internal reference in eval.

As described in this RFC #2864 , we may have chance to extend it.

### Related Issues
Resolves #2904

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/planner/optimizer/rule/EvalPushDown.java
+++ b/core/src/main/java/org/opensearch/sql/planner/optimizer/rule/EvalPushDown.java
@@ -7,6 +7,7 @@ package org.opensearch.sql.planner.optimizer.rule;
 
 import static org.opensearch.sql.planner.optimizer.pattern.Patterns.evalCapture;
 import static org.opensearch.sql.planner.optimizer.pattern.Patterns.limit;
+import static org.opensearch.sql.planner.optimizer.pattern.Patterns.sort;
 import static org.opensearch.sql.planner.optimizer.rule.EvalPushDown.EvalPushDownBuilder.match;
 
 import com.facebook.presto.matching.Capture;
@@ -14,13 +15,21 @@ import com.facebook.presto.matching.Captures;
 import com.facebook.presto.matching.Pattern;
 import com.facebook.presto.matching.pattern.CapturePattern;
 import com.facebook.presto.matching.pattern.WithPattern;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.function.BiFunction;
+import java.util.stream.Collectors;
 import lombok.Getter;
 import lombok.experimental.Accessors;
+import org.apache.commons.lang3.tuple.Pair;
+import org.opensearch.sql.ast.tree.Sort.SortOption;
+import org.opensearch.sql.expression.Expression;
+import org.opensearch.sql.expression.ReferenceExpression;
 import org.opensearch.sql.planner.logical.LogicalEval;
 import org.opensearch.sql.planner.logical.LogicalLimit;
 import org.opensearch.sql.planner.logical.LogicalPlan;
+import org.opensearch.sql.planner.logical.LogicalSort;
 import org.opensearch.sql.planner.optimizer.Rule;
 
 /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/opensearch/sql/planner/optimizer/rule/EvalPushDown.java
```diff
@@ -42,6 +51,38 @@ public class EvalPushDown<T extends LogicalPlan> implements Rule<T> {
...
+                  if (pair.getRight() instanceof ReferenceExpression) {
+                    ReferenceExpression ref = (ReferenceExpression) pair.getRight();
+                    Expression replacedExpr = evalExpressionMap.getOrDefault(ref, ref);
+                    if (replacedExpr instanceof ReferenceExpression) {
+                      ReferenceExpression newRef = (ReferenceExpression) replacedExpr;
+                      if (!evalExpressionMap.containsKey(newRef)) {
+                        newSortList.add(Pair.of(pair.getLeft(), newRef));
+                      } else return sort;
+                    } else return sort;
+                  } else return sort;
```
ykmr1224, line 79: I think you can `return sort` at the end of `for` block once.

core/src/main/java/org/opensearch/sql/planner/optimizer/rule/EvalPushDown.java
```diff
@@ -42,6 +51,38 @@ public class EvalPushDown<T extends LogicalPlan> implements Rule<T> {
...
+                List<Pair<SortOption, Expression>> newSortList = new ArrayList<>();
+                for (Pair<SortOption, Expression> pair : sortList) {
+                  /*
+                   Narrow down the optimization to only support:
+                   1. The expression in sort and replaced expression are both ReferenceExpression.
+                   2. No internal reference in eval.
+                  */
+                  if (pair.getRight() instanceof ReferenceExpression) {
+                    ReferenceExpression ref = (ReferenceExpression) pair.getRight();
+                    Expression replacedExpr = evalExpressionMap.getOrDefault(ref, ref);
```
ykmr1224, line 72: Is it right to return `ref` as default value?

core/src/main/java/org/opensearch/sql/planner/optimizer/rule/EvalPushDown.java
```diff
@@ -42,6 +51,38 @@ public class EvalPushDown<T extends LogicalPlan> implements Rule<T> {
                 return logicalEval;
               });
 
+  public static final Rule<LogicalSort> PUSH_DOWN_SORT =
+      match(sort(evalCapture()))
+          .apply(
+              (sort, logicalEval) -> {
```
ykmr1224, line 57: Looks this Lambda is too large. Please extract to a method or a class. And extract some portion to reduce the size of method for readability.

core/src/test/java/org/opensearch/sql/planner/optimizer/LogicalPlanOptimizerTest.java
```diff
@@ -368,6 +369,48 @@ void push_limit_through_eval_into_scan() {
...
+
+    // don't push sort if replaced expr in eval is not ReferenceExpression
+    Pair<ReferenceExpression, Expression> evalExprWithNonRef = Pair.of(sortRef, nonRefExpr);
+    originPlan = sort(eval(relation("schema", table), evalExprWithNonRef), sortExpr);
+    assertEquals(originPlan, optimize(originPlan));
+
+    // don't push sort if there are internal reference in eval
+    Pair<ReferenceExpression, Expression> evalExpr2 = Pair.of(sortRef, evalRef);
+    originPlan = sort(eval(relation("schema", table), evalExpr, evalExpr2), sortExpr);
+    assertEquals(originPlan, optimize(originPlan));
```
ykmr1224, line 398: Please split the method for each test case.

core/src/main/java/org/opensearch/sql/planner/optimizer/rule/EvalPushDown.java
qianheng-aws: Logically, it's right I think, though the name of `replacedExpr` may be a little bit misleading.  It includes 2 cases:
1. the `ref` is produced by eval operator, then it needs to be replaced
2. the `ref` isn't produced by eval operator, then we don't need to replace it. That's why I use the default value `ref`.

How about changing the name of `replacedExpr` to `newExpr`?

core/src/main/java/org/opensearch/sql/planner/optimizer/rule/EvalPushDown.java
qianheng-aws: Then we need a flag to indicate returning sort or not for each iteration. 

It could be simplified by pattern matching with java version>=14. But since we also need to support java11 for 2.x, I don't use that feature to keep code align.

RyanL1997:
Hi @qianheng-aws , do we still need this change?

PR #2940, opened 2024-08-15
Add pit for pagination query

manasvinibs:
### Description
Added option to use search after with pit instead of scroll api for pagination query

### Related Issues

Manual test on sample queries - https://github.com/opensearch-project/sql/issues/2941

### Check List

### DIFF
--- a/legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
+++ b/legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
@@ -13,10 +13,11 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -26,7 +27,6 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
 import lombok.Setter;
-import lombok.SneakyThrows;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.opensearch.common.settings.Settings;
@@ -99,7 +99,7 @@ public class DefaultCursor implements Cursor {
   private String pitId;
 
   /** To get next batch of result with search after api */
-  public SearchSourceBuilder searchSourceBuilder;
+  private SearchSourceBuilder searchSourceBuilder;
 
   /** To get last sort values * */
   private Object[] sortFields;
@@ -115,7 +115,7 @@ public class DefaultCursor implements Cursor {
    */
   private static final NamedXContentRegistry xContentRegistry =
       new NamedXContentRegistry(
-          new SearchModule(Settings.builder().build(), new ArrayList<>()).getNamedXContents());
+          new SearchModule(Settings.EMPTY, Collections.emptyList()).getNamedXContents());
 
   @Override
   public CursorType getType() {
@@ -124,11 +124,7 @@ public class DefaultCursor implements Cursor {
 
   @Override
   public String generateCursorId() {
-    boolean isCursorValid =
-        LocalClusterState.state().getSettingValue(SQL_PAGINATION_API_SEARCH_AFTER)
-            ? Strings.isNullOrEmpty(pitId)
-            : Strings.isNullOrEmpty(scrollId);
-    if (rowsLeft <= 0 || isCursorValid) {
+    if (rowsLeft <= 0 || isCursorIdNullOrEmpty()) {
       return null;
     }
     JSONObject json = new JSONObject();
@@ -146,51 +142,50 @@ public class DefaultCursor implements Cursor {
                     try {
                       return objectMapper.writeValueAsString(sortFields);
                     } catch (JsonProcessingException e) {
-                      throw new RuntimeException(e);
+                      throw new RuntimeException(
+                          "Failed to parse sort fields from JSON string.", e);
                     }
                   });
       json.put(SORT_FIELDS, sortFieldValue);
+      setSearchRequestString(json, searchSourceBuilder);
     } else {
       json.put(SCROLL_ID, scrollId);
     }
-    return String.format("%s:%s", type.getId(), encodeCursor(json, searchSourceBuilder));
+    return String.format("%s:%s", type.getId(), encodeCursor(json));
   }
 
-  @SneakyThrows
+  private void setSearchRequestString(JSONObject cursorJson, SearchSourceBuilder sourceBuilder) {
+    try {
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      XContentBuilder builder = XContentFactory.jsonBuilder(outputStream);
+      sourceBuilder.toXContent(builder, null);
+      builder.close();
+
+      String searchRequestBase64 = Base64.getEncoder().encodeToString(outputStream.toByteArray());
+      cursorJson.put("searchSourceBuilder", searchRequestBase64);
+    } catch (IOException ex) {
+      throw new RuntimeException("Failed to set search request string on cursor json.", ex);
+    }
+  }
+
+  private boolean isCursorIdNullOrEmpty() {
+    return LocalClusterState.state().getSettingValue(SQL_PAGINATION_API_SEARCH_AFTER)
+        ? Strings.isNullOrEmpty(pitId)
+        : Strings.isNullOrEmpty(scrollId);
+  }
+
   public static DefaultCursor from(String cursorId) {
     /**
      * It is assumed that cursorId here is the second part of the original cursor passed by the
      * client after removing first part which identifies cursor type
      */
-    String[] parts = cursorId.split(":::");
-    JSONObject json = decodeCursor(parts[0]);
+    JSONObject json = decodeCursor(cursorId);
     DefaultCursor cursor = new DefaultCursor();
     cursor.setFetchSize(json.getInt(FETCH_SIZE));
     cursor.setRowsLeft(json.getLong(ROWS_LEFT));
     cursor.setIndexPattern(json.getString(INDEX_PATTERN));
     if (LocalClusterState.state().getSettingValue(SQL_PAGINATION_API_SEARCH_AFTER)) {
-      cursor.setPitId(json.getString(PIT_ID));
-
-      Object[] sortFieldValue =
-          AccessController.doPrivileged(
-              (PrivilegedAction<Object[]>)
-                  () -> {
-                    try {
-                      return objectMapper.readValue(json.getString(SORT_FIELDS), Object[].class);
-                    } catch (JsonProcessingException e) {
-                      throw new RuntimeException(e);
-                    }
-                  });
-      cursor.setSortFields(sortFieldValue);
-
-      byte[] bytes = Base64.getDecoder().decode(parts[1]);
-      ByteArrayInputStream streamInput = new ByteArrayInputStream(bytes);
-      XContentParser parser =
-          XContentType.JSON
-              .xContent()
-              .createParser(xContentRegistry, IGNORE_DEPRECATIONS, streamInput);
-      SearchSourceBuilder sourceBuilder = SearchSourceBuilder.fromXContent(parser);
-      cursor.searchSourceBuilder = sourceBuilder;
+      populateCursorForPit(json, cursor);
     } else {
       cursor.setScrollId(json.getString(SCROLL_ID));
     }
@@ -198,6 +193,39 @@ public class DefaultCursor implements Cursor {
     cursor.setFieldAliasMap(fieldAliasMap(json.getJSONObject(FIELD_ALIAS_MAP)));
 
     return cursor;
+  }
+
+  private static void populateCursorForPit(JSONObject json, DefaultCursor cursor) {
+    cursor.setPitId(json.getString(PIT_ID));
+
+    cursor.setSortFields(getSortFieldsFromJson(json));
+
+    // Retrieve and set the SearchSourceBuilder from the JSON field
+    String searchSourceBuilderBase64 = json.getString("searchSourceBuilder");
+    byte[] bytes = Base64.getDecoder().decode(searchSourceBuilderBase64);
+    ByteArrayInputStream streamInput = new ByteArrayInputStream(bytes);
+    try {
+      XContentParser parser =
+          XContentType.JSON
+              .xContent()
+              .createParser(xContentRegistry, IGNORE_DEPRECATIONS, streamInput);
+      SearchSourceBuilder sourceBuilder = SearchSourceBuilder.fromXContent(parser);
+      cursor.setSearchSourceBuilder(sourceBuilder);
+    } catch (IOException ex) {
+      throw new RuntimeException("Failed to get searchSourceBuilder from cursor Id", ex);
+    }
+  }
+
+  private static Object[] getSortFieldsFromJson(JSONObject json) {
+    return AccessController.doPrivileged(
+        (PrivilegedAction<Object[]>)
+            () -> {
+              try {
+                return objectMapper.readValue(json.getString(SORT_FIELDS), Object[].class);
+              } catch (JsonProcessingException e) {
+                throw new RuntimeException("Failed to parse sort fields from JSON string.", e);
+              }
+            });
   }
 
   private JSONArray getSchemaAsJson() {
@@ -220,18 +248,8 @@ public class DefaultCursor implements Cursor {
     return entry;
   }
 
-  @SneakyThrows
-  private static String encodeCursor(JSONObject cursorJson, SearchSourceBuilder sourceBuilder) {
-    String jsonBase64 = Base64.getEncoder().encodeToString(cursorJson.toString().getBytes());
-
-    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-    XContentBuilder builder = XContentFactory.jsonBuilder(outputStream);
-    sourceBuilder.toXContent(builder, null);
-    builder.close();
-
-    String searchRequestBase64 = Base64.getEncoder().encodeToString(outputStream.toByteArray());
-
-    return jsonBase64 + ":::" + searchRequestBase64;
+  private static String encodeCursor(JSONObject cursorJson) {
+    return Base64.getEncoder().encodeToString(cursorJson.toString().getBytes());
   }
 
   private static JSONObject decodeCursor(String cursorId) {
--- a/legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
+++ b/legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
@@ -5,7 +5,17 @@
 
 package org.opensearch.sql.legacy.cursor;
 
+import static org.opensearch.core.xcontent.DeprecationHandler.IGNORE_DEPRECATIONS;
+import static org.opensearch.sql.common.setting.Settings.Key.SQL_PAGINATION_API_SEARCH_AFTER;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.List;
@@ -16,8 +26,19 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
 import lombok.Setter;
+import lombok.SneakyThrows;
 import org.json.JSONArray;
 import org.json.JSONObject;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.xcontent.XContentFactory;
+import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.search.SearchModule;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.sql.legacy.esdomain.LocalClusterState;
 import org.opensearch.sql.legacy.executor.format.Schema;
 
 /**
@@ -40,6 +61,10 @@ public class DefaultCursor implements Cursor {
   private static final String SCROLL_ID = "s";
   private static final String SCHEMA_COLUMNS = "c";
   private static final String FIELD_ALIAS_MAP = "a";
+  private static final String PIT_ID = "p";
+  private static final String SEARCH_REQUEST = "r";
+  private static final String SORT_FIELDS = "h";
+  private static final ObjectMapper objectMapper = new ObjectMapper();
 
   /**
    * To get mappings for index to check if type is date needed for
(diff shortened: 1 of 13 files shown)


legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -70,42 +95,105 @@ public class DefaultCursor implements Cursor {
...
 
   private Integer limit;
 
+  /**
+   * {@link NamedXContentRegistry} from {@link SearchModule} used for construct {@link QueryBuilder}
+   * from DSL query string.
+   */
+  private static final NamedXContentRegistry xContentRegistry =
+      new NamedXContentRegistry(
+          new SearchModule(Settings.builder().build(), new ArrayList<>()).getNamedXContents());
```
vamsi-amazon, line 118: There should be Settings.EMPTY?
Also use Collections.emptyList().

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -70,42 +95,105 @@ public class DefaultCursor implements Cursor {
...
+
   @Override
   public CursorType getType() {
     return type;
   }
 
   @Override
   public String generateCursorId() {
-    if (rowsLeft <= 0 || Strings.isNullOrEmpty(scrollId)) {
+    boolean isCursorValid =
```
vamsi-amazon, line 127: Move this to logic to a function.

vamsi-amazon: `rowsLeft <= 0 || isCursorValid`

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -132,8 +220,18 @@ private JSONObject schemaEntry(String name, String alias, String type) {
...
+    String jsonBase64 = Base64.getEncoder().encodeToString(cursorJson.toString().getBytes());
+
+    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+    XContentBuilder builder = XContentFactory.jsonBuilder(outputStream);
+    sourceBuilder.toXContent(builder, null);
+    builder.close();
+
+    String searchRequestBase64 = Base64.getEncoder().encodeToString(outputStream.toByteArray());
+
+    return jsonBase64 + ":::" + searchRequestBase64;
```
vamsi-amazon, line 234: Also,, can we add searchSourceBuilder as a filed in json instead of adding it to cursor.
```
 XContentBuilder builder = XContentFactory.jsonBuilder();
        sourceBuilder.toXContent(builder, null);
        String searchRequestJson = builder.string();
```
```
String searchRequestJson = new String(searchRequestBytes);
        
        XContentParser parser = XContentType.JSON.xContent().createParser(
            xContentRegistry, 
            IGNORE_DEPRECATIONS, 
            searchRequestJson
        );
        
        SearchSourceBuilder sourceBuilder = SearchSourceBuilder.fromXContent(parser);
        ```


Add searchRequestJson to JSON

 @SneakyThrows is not good in source code. Catch the exception and log the errors.

File: legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java [hunk 4]
```diff
@@ -37,37 +39,68 @@ public PointInTimeHandlerImpl(Client client, String[] indices) {
...
           }
         });
+    while (createStatus == null) {
+      try {
+        Thread.sleep(100);
+      } catch (InterruptedException e) {
+        LOG.error("Error occurred while creating PIT", e);
+      }
+    }
+    return createStatus;
```
vamsi-amazon, line 87: What is this? This is bad. Are there no synchronous requests?

legacy/src/main/java/org/opensearch/sql/legacy/query/DefaultQueryAction.java
```diff
@@ -100,7 +102,19 @@ public void checkAndSetScroll() {
...
+      request.setSize(fetchSize);
+      // Set scroll or search after for pagination
+      if (LocalClusterState.state().getSettingValue(SQL_PAGINATION_API_SEARCH_AFTER)) {
+        // search after requires results to be in specific order
+        // set sort field for search_after
+        boolean ordered = select.isOrderdSelect();
+        if (!ordered) {
+          request.addSort(FieldSortBuilder.DOC_FIELD_NAME, SortOrder.ASC);
+        }
+        // Request also requires PointInTime, but we should create pit while execution.
```
vamsi-amazon, line 114: Is this addressed?

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -90,20 +98,49 @@ public String execute(Client client, Map<String, String> params, QueryAction que
   private Protocol buildProtocolForDefaultQuery(Client client, DefaultQueryAction queryAction)
```
vamsi-amazon, line 98: When is this method used?

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -5,23 +5,31 @@
...
 import org.opensearch.sql.legacy.executor.QueryActionElasticExecutor;
 import org.opensearch.sql.legacy.executor.RestExecutor;
+import org.opensearch.sql.legacy.pit.PointInTimeHandler;
+import org.opensearch.sql.legacy.pit.PointInTimeHandlerImpl;
 import org.opensearch.sql.legacy.query.DefaultQueryAction;
 import org.opensearch.sql.legacy.query.QueryAction;
+import org.opensearch.sql.legacy.query.SqlOpenSearchRequestBuilder;
 import org.opensearch.sql.legacy.query.join.BackOffRetryStrategy;
 
 public class PrettyFormatRestExecutor implements RestExecutor {
```
vamsi-amazon, line 35: What is the relation between pagination queries and this class?

manasvinibs: I think any default query action with search uses this executor and this executor in turn uses `scrollId` to form cursor Id as part of  search hits while returning search response to the user. Wherever we are using scrollId today needs this refactor I think.

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
manasvinibs: This is called while formatting the search response received by executing any query action plan.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
manasvinibs: Updated to use `Collections.emptyList()`. But I'll be hesitant to change this to `Settings.EMPTY` as I'm not entirely sure if it is same as `Settings.builder().build()` because here we might be trying to retrieve the existing settings and it might not be empty!?

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
manasvinibs: Looking at the previous checked in iteration of this change here https://github.com/opensearch-project/sql/pull/2703/files#diff-b7d29ea36ecb24676a833f1897d949d465f34c1595dbdd81c283a355447a9704R42 I see this change is made after the previous PR and I do not have entire context on why would we need this change. Looks like this might be part of some testing. I took out this unnecessary change from this PR and verified the changes.

legacy/src/main/java/org/opensearch/sql/legacy/query/DefaultQueryAction.java
manasvinibs: I think this is just a comment for the reader to call out executors handle the creation of PIT when we do this check here.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
vamsi-amazon: ``
Debug and see if there is anything in settings. Nothing is being used in Settings...we are creating dummy search module just to use NamedXContents.
``

manasvinibs: That's right! I see both the values are empty actually.. changes it to Settings.Empty.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
manasvinibs: I added completable future to wait for the async thread to complete PIT creation/delete before the process continues with next operation which might be depending on the PIT ID. This way its safe to assume we are not failing during query execution if PIT is still being created.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -132,8 +222,20 @@ private JSONObject schemaEntry(String name, String alias, String type) {
...
-    return Base64.getEncoder().encodeToString(cursorJson.toString().getBytes());
+  private static String encodeCursor(JSONObject cursorJson, SearchSourceBuilder sourceBuilder) {
+    try {
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      XContentBuilder builder = XContentFactory.jsonBuilder(outputStream);
+      sourceBuilder.toXContent(builder, null);
+      builder.close();
+
+      String searchRequestBase64 = Base64.getEncoder().encodeToString(outputStream.toByteArray());
+      cursorJson.put("searchSourceBuilder", searchRequestBase64);
```
ykmr1224, line 233: Is `searchSourceBuilder` the right naming? Also unsure if it is right to serialize the searchSourceBuilder and make it cursor. (builder didn't sound like a persisted object, but it is fine if this is a common way in OpenSearch)

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -70,29 +91,67 @@ public class DefaultCursor implements Cursor {
...
     }
     JSONObject json = new JSONObject();
     json.put(FETCH_SIZE, fetchSize);
     json.put(ROWS_LEFT, rowsLeft);
     json.put(INDEX_PATTERN, indexPattern);
-    json.put(SCROLL_ID, scrollId);
     json.put(SCHEMA_COLUMNS, getSchemaAsJson());
     json.put(FIELD_ALIAS_MAP, fieldAliasMap);
-    return String.format("%s:%s", type.getId(), encodeCursor(json));
+    if (LocalClusterState.state().getSettingValue(SQL_PAGINATION_API_SEARCH_AFTER)) {
```
ykmr1224, line 132: I see this branching logic in many place. And it is making the code more complicated. Can't we implement the logic for PIT in a separate class?

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -105,7 +164,38 @@ public static DefaultCursor from(String cursorId) {
...
+      String searchSourceBuilderBase64 = json.getString("searchSourceBuilder");
+      byte[] bytes = Base64.getDecoder().decode(searchSourceBuilderBase64);
+      ByteArrayInputStream streamInput = new ByteArrayInputStream(bytes);
+      try {
+        XContentParser parser =
+            XContentType.JSON
+                .xContent()
+                .createParser(xContentRegistry, IGNORE_DEPRECATIONS, streamInput);
+        SearchSourceBuilder sourceBuilder = SearchSourceBuilder.fromXContent(parser);
+        cursor.searchSourceBuilder = sourceBuilder;
```
vamsi-amazon, line 192: can we use setter?

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -70,29 +91,67 @@ public class DefaultCursor implements Cursor {
   /** To get next batch of result */
   private String scrollId;
 
+  /** To get Point In Time */
+  private String pitId;
+
+  /** To get next batch of result with search after api */
+  public SearchSourceBuilder searchSourceBuilder;
```
vamsi-amazon, line 98: public fields is a bad practice.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
```diff
@@ -37,31 +39,62 @@ public PointInTimeHandlerImpl(Client client, String[] indices) {
...
+
+  /**
+   * Create PIT for given indices
+   *
+   * @return Point In Time creation status
+   */
   @Override
-  public void create() {
+  public boolean create() {
+    CompletableFuture<Boolean> future = new CompletableFuture<>();
```
vamsi-amazon, line 60: This is not required. The client itself provides this.

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/client/support/AbstractClient.java#L465
Please copy the code here:
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/internalClusterTest/java/org/opensearch/search/slice/SearchSliceIT.java#L177

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
vamsi-amazon: @manasvinibs  Please don't change anything in encodeCursor..before inovking this method...add 
`
 cursorJson.put("searchSourceBuilder", searchRequestBase64);
 `
 Add the above stuff in the invoker.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
manasvinibs: Thanks for catching!

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
manasvinibs: > Is `searchSourceBuilder` the right naming? Also unsure if it is right to serialize the searchSourceBuilder and make it cursor. (builder didn't sound like a persisted object, but it is fine if this is a common way in OpenSearch)
I think Rupal spent sometime on this and after discussing with team came up with this approach. (Now that we need sort info for PIT search_after). I'm not sure how this can be improved, if you think its worth it can be something to take as follow-up.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -70,29 +91,67 @@ public class DefaultCursor implements Cursor {
...
+
   @Override
   public CursorType getType() {
     return type;
   }
 
   @Override
   public String generateCursorId() {
-    if (rowsLeft <= 0 || Strings.isNullOrEmpty(scrollId)) {
+    if (rowsLeft <= 0 || isCursorValid()) {
```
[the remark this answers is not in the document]
manasvinibs: ya I think naming is confusing. I've updated it.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
manasvinibs: I think for now both the approaches are supported and it is temporary till we get benchmark both the performances and remove scroll support at some point. I can create an issue for this to track - https://github.com/opensearch-project/sql/issues/2945

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
manasvinibs: Nice

ykmr1224 (dismissed):

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
```diff
@@ -37,47 +40,55 @@ public PointInTimeHandlerImpl(Client client, String[] indices) {
...
-        });
+    ActionFuture<CreatePitResponse> execute =
+        client.execute(CreatePitAction.INSTANCE, createPitRequest);
+    try {
+      CreatePitResponse pitResponse = execute.get();
+      pitId = pitResponse.getId();
+      LOG.info("Created Point In Time {} successfully.", pitId);
+      return true;
+    } catch (InterruptedException | ExecutionException e) {
+      throw new RuntimeException("Error occurred while creating PIT.", e);
```
vamsi-amazon, line 72: What will be the end user response when we encounter these?

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
```diff
@@ -37,47 +40,55 @@ public PointInTimeHandlerImpl(Client client, String[] indices) {
...
-            LOG.error("Error occurred while creating PIT", e);
-          }
-        });
+    ActionFuture<CreatePitResponse> execute =
+        client.execute(CreatePitAction.INSTANCE, createPitRequest);
+    try {
+      CreatePitResponse pitResponse = execute.get();
+      pitId = pitResponse.getId();
+      LOG.info("Created Point In Time {} successfully.", pitId);
+      return true;
```
vamsi-amazon, line 70: What is this boolean used for?

Where is false scenario?

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
manasvinibs: I don't think this is a blocker. I think supporting both the approaches with default to PIT was the plan from the beginning as user can still set it in the setting. At least I don't see anything called out in the hand-off doc. Let me know if you think this is a blocker.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
manasvinibs: Currently I see this boolean is used only in unit tests for assertions. When PIT creation fails, I'm throwing exception instead of returning false to stop the query execution.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
manasvinibs: User will see exception with the failure message followed by stack trace of underlying execution exception.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
vamsi-amazon: Lets make sure we have something to track this

manasvinibs: Yes I have created the issue already https://github.com/opensearch-project/sql/issues/2945

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
vamsi-amazon: we don't change interface methods just for unit tests.
In unit tests, use some other way by checking no exceptions are made and change this method to void.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
vamsi-amazon: Can you test it and see what is the error code and message that is coming up. You can add fake exception and test it.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -105,7 +184,39 @@ public static DefaultCursor from(String cursorId) {
...
+      try {
+        XContentParser parser =
+            XContentType.JSON
+                .xContent()
+                .createParser(xContentRegistry, IGNORE_DEPRECATIONS, streamInput);
+        SearchSourceBuilder sourceBuilder = SearchSourceBuilder.fromXContent(parser);
+        cursor.setSearchSourceBuilder(sourceBuilder);
+      } catch (IOException ex) {
+        throw new RuntimeException("Failed to get searchSourceBuilder from cursor Id", ex);
+      }
```
vamsi-amazon, line 216: can we refactor this to a new method.

legacy/src/main/java/org/opensearch/sql/legacy/executor/cursor/CursorCloseExecutor.java
```diff
@@ -79,14 +83,25 @@ public String execute(Client client, Map<String, String> params) throws Exceptio
...
-      return SUCCEEDED_TRUE;
+    if (LocalClusterState.state().getSettingValue(SQL_PAGINATION_API_SEARCH_AFTER)) {
+      String pitId = cursor.getPitId();
+      PointInTimeHandler pit = new PointInTimeHandlerImpl(client, pitId);
+      if (pit.delete()) {
+        return SUCCEEDED_TRUE;
+      } else {
+        Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
+        return SUCCEEDED_FALSE;
+      }
```
vamsi-amazon, line 94: Does this makes sense, if we are never returning false from delete method. 91-93 would never get executed.

if you want to throw exception in delete method, catch exception here and emit metric.
If you want to return boolean, don't throw exception in delete method.

vamsi-amazon: I possible how scroll is executed below.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
vamsi-amazon: you can do this after merging the code as well.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
manasvinibs: sure, updated

legacy/src/main/java/org/opensearch/sql/legacy/executor/cursor/CursorCloseExecutor.java
manasvinibs: Yes that's right! Updated the places using delete function to catch exception and log metrics similar to scroll.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -70,31 +95,85 @@ public class DefaultCursor implements Cursor {
...
+      setSearchRequestString(json, searchSourceBuilder);
+    } else {
+      json.put(SCROLL_ID, scrollId);
+    }
     return String.format("%s:%s", type.getId(), encodeCursor(json));
   }
 
+  private void setSearchRequestString(JSONObject cursorJson, SearchSourceBuilder sourceBuilder) {
+    try {
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
vamsi-amazon, line 159: I don't know if we really need output stream here:

try (XContentBuilder builder = XContentFactory.jsonBuilder()) {
    sourceBuilder.toXContent(builder, ToXContent.EMPTY_PARAMS);
    String jsonString = builder.toString();
    searchRequestBase64 = Base64.getEncoder().encodeToString(jsonString.getBytes(StandardCharsets.UTF_8));
}
can you take this after the PR?

vamsi-amazon (approved):

vamsi-amazon:
I am approving for now. please follow up with the issues created out of the review.

legacy/src/main/java/org/opensearch/sql/legacy/cursor/DefaultCursor.java
```diff
@@ -146,58 +142,90 @@ public String generateCursorId() {
...
       json.put(SORT_FIELDS, sortFieldValue);
+      setSearchRequestString(json, searchSourceBuilder);
     } else {
       json.put(SCROLL_ID, scrollId);
     }
-    return String.format("%s:%s", type.getId(), encodeCursor(json, searchSourceBuilder));
+    return String.format("%s:%s", type.getId(), encodeCursor(json));
+  }
+
+  private void setSearchRequestString(JSONObject cursorJson, SearchSourceBuilder sourceBuilder) {
```
ykmr1224, line 157: nit: we should rather return the searchRequestBase64 from this method and set it to json at caller side. (it is not a good practice to modify parameter object in a method for readability/maintainability) Please address it in later PR.

ykmr1224 (approved):

PR #2959, opened 2024-08-30
Adds validation to allow only flint queries and sql SELECT queries to security lake type datasource

### LINKED ISSUE #2907: Limit Spark SQL queries to SELECT  + FLINT commands when Lake Formation is enabled
When using Lake Formation, OpenSearch should limit the Spark SQL queries to only SELECT to limit the capability of users to escalate privilege beyond what Lake Formation actually grants the user.

This proposal includes 2 main pieces:
1. Define a more limited ANTLR grammar that restricts SQL grammar to only SELECT, COVERING INDEX, and MATERIALIZED VIEW
2. If Lake Formation is enabled, validate that SQL is restricted to that grammar.

eirsep:
### Description
Adds validation to allow only flint queries and sql select queries to security lake type datasource

### Related Issues
Resolves #2907

### DIFF
--- a/async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
+++ b/async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
@@ -5,19 +5,18 @@
 
 package org.opensearch.sql.spark.utils;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
 import lombok.Getter;
 import lombok.experimental.UtilityClass;
 import org.antlr.v4.runtime.CommonTokenStream;
 import org.antlr.v4.runtime.misc.Interval;
 import org.antlr.v4.runtime.tree.ParseTree;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.opensearch.sql.common.antlr.CaseInsensitiveCharStream;
 import org.opensearch.sql.common.antlr.SyntaxAnalysisErrorListener;
 import org.opensearch.sql.common.antlr.SyntaxCheckException;
+import org.opensearch.sql.datasource.model.DataSource;
+import org.opensearch.sql.datasource.model.DataSourceType;
 import org.opensearch.sql.spark.antlr.parser.FlintSparkSqlExtensionsBaseVisitor;
 import org.opensearch.sql.spark.antlr.parser.FlintSparkSqlExtensionsLexer;
 import org.opensearch.sql.spark.antlr.parser.FlintSparkSqlExtensionsParser;
@@ -32,13 +31,18 @@ import org.opensearch.sql.spark.dispatcher.model.IndexQueryActionType;
 import org.opensearch.sql.spark.dispatcher.model.IndexQueryDetails;
 import org.opensearch.sql.spark.flint.FlintIndexType;
 
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Locale;
+
 /**
  * This util class parses spark sql query and provides util functions to identify indexName,
  * tableName and datasourceName.
  */
 @UtilityClass
 public class SQLQueryUtils {
-
+  private static final Logger logger = LogManager.getLogger(SQLQueryUtils.class);
   public static List<FullyQualifiedTableName> extractFullyQualifiedTableNames(String sqlQuery) {
     SqlBaseParser sqlBaseParser =
         new SqlBaseParser(
@@ -77,7 +81,8 @@ public class SQLQueryUtils {
     }
   }
 
-  public static List<String> validateSparkSqlQuery(String sqlQuery) {
+  public static List<String> validateSparkSqlQuery(DataSource datasource, String sqlQuery) {
+    List<String> validationErrors = new ArrayList<>();
     SparkSqlValidatorVisitor sparkSqlValidatorVisitor = new SparkSqlValidatorVisitor();
     SqlBaseParser sqlBaseParser =
         new SqlBaseParser(
@@ -86,10 +91,16 @@ public class SQLQueryUtils {
     try {
       SqlBaseParser.StatementContext statement = sqlBaseParser.statement();
       sparkSqlValidatorVisitor.visit(statement);
-      return sparkSqlValidatorVisitor.getValidationErrors();
+      validationErrors.addAll(sparkSqlValidatorVisitor.getValidationErrors());
     } catch (SyntaxCheckException syntaxCheckException) {
-      return Collections.emptyList();
+      logger.error("Failed to parse sql statement context while validating sql query ");
     }
+    if(datasource.getConnectorType().equals(DataSourceType.SECURITY_LAKE)) {
+      if(!isSelectQuery(sqlQuery)) {
+        validationErrors.add("Only SELECT queries are allowed for sql queries submitted to Security Lake datasource");
+      }
+    }
+    return validationErrors;
   }
 
   private static class SparkSqlValidatorVisitor extends SqlBaseParserBaseVisitor<Void> {
@@ -388,5 +399,9 @@ public class SQLQueryUtils {
     public String removeUnwantedQuotes(String input) {
       return input.replaceAll("^\"|\"$", "");
     }
+  }
+
+  private boolean isSelectQuery(String query) {
+    return query.trim().toUpperCase().startsWith("SELECT");
   }
 }
--- a/async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
+++ b/async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
@@ -5,6 +5,11 @@
 
 package org.opensearch.sql.spark.utils;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Locale;
 import lombok.Getter;
 import lombok.experimental.UtilityClass;
 import org.antlr.v4.runtime.CommonTokenStream;
@@ -12,7 +17,6 @@ import org.antlr.v4.runtime.misc.Interval;
 import org.antlr.v4.runtime.tree.ParseTree;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.jetbrains.annotations.NotNull;
 import org.opensearch.sql.common.antlr.CaseInsensitiveCharStream;
 import org.opensearch.sql.common.antlr.SyntaxAnalysisErrorListener;
 import org.opensearch.sql.common.antlr.SyntaxCheckException;
@@ -33,12 +37,6 @@ import org.opensearch.sql.spark.dispatcher.model.IndexQueryActionType;
 import org.opensearch.sql.spark.dispatcher.model.IndexQueryDetails;
 import org.opensearch.sql.spark.flint.FlintIndexType;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-
 /**
  * This util class parses spark sql query and provides util functions to identify indexName,
  * tableName and datasourceName.
@@ -46,6 +44,7 @@ import java.util.Locale;
 @UtilityClass
 public class SQLQueryUtils {
   private static final Logger logger = LogManager.getLogger(SQLQueryUtils.class);
+
   public static List<FullyQualifiedTableName> extractFullyQualifiedTableNames(String sqlQuery) {
     SqlBaseParser sqlBaseParser =
         new SqlBaseParser(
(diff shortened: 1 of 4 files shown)


eirsep:
@opensearch-project/sql team, 

2 points for consideration before I can mark PR as ready
1. I am considering adding a configuration file to add validator methods for combination of datasource type and queries and register them to different datasources via config.
2. What's the right way to determine if sql query is a SELECT query? Checking if query string starts with SELECT keyword seems a bit juvenile.

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
```diff
@@ -77,32 +83,73 @@ public static boolean isFlintExtensionQuery(String sqlQuery) {
...
+      sqlBaseParser.addErrorListener(new SyntaxAnalysisErrorListener());
+      SqlBaseValidatorVisitor sqlParserBaseVisitor = getSparkSqlValidatorVisitor(datasource);
+      StatementContext statement = sqlBaseParser.statement();
+      sqlParserBaseVisitor.visit(statement);
+      return sqlParserBaseVisitor.getValidationErrors();
+    } catch (SyntaxCheckException e) {
+      logger.error(
+          String.format(
+              "Failed to parse sql statement context while validating sql query %s", sqlQuery),
+          e);
```
ykmr1224, line 100: Why do we add this error log?

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
```diff
@@ -77,32 +83,73 @@ public static boolean isFlintExtensionQuery(String sqlQuery) {
...
     try {
-      SqlBaseParser.StatementContext statement = sqlBaseParser.statement();
-      sparkSqlValidatorVisitor.visit(statement);
-      return sparkSqlValidatorVisitor.getValidationErrors();
-    } catch (SyntaxCheckException syntaxCheckException) {
+      SqlBaseParser sqlBaseParser =
+          new SqlBaseParser(
+              new CommonTokenStream(new SqlBaseLexer(new CaseInsensitiveCharStream(sqlQuery))));
+      sqlBaseParser.addErrorListener(new SyntaxAnalysisErrorListener());
+      SqlBaseValidatorVisitor sqlParserBaseVisitor = getSparkSqlValidatorVisitor(datasource);
```
ykmr1224, line 92: Why do we move into the try block? As we don't expect SyntaxCheckException from these, we should put outside.

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
```diff
@@ -38,13 +43,14 @@
  */
 @UtilityClass
 public class SQLQueryUtils {
+  private static final Logger logger = LogManager.getLogger(SQLQueryUtils.class);
```
ykmr1224, line 46: nit: We prefer `LOG` instead of `logger`, though I see some variation in this repository...

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
```diff
@@ -404,15 +407,40 @@ void testAutoRefresh() {
 
   @Test
   void testValidateSparkSqlQuery_ValidQuery() {
-    String validQuery = "SELECT * FROM users WHERE age > 18";
-    List<String> errors = SQLQueryUtils.validateSparkSqlQuery(validQuery);
+    DataSource dataSource = Mockito.mock(DataSource.class);
```
ykmr1224, line 410: Can you use `@Mock` annotation instead?

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
```diff
@@ -404,15 +407,40 @@ void testAutoRefresh() {
 
   @Test
   void testValidateSparkSqlQuery_ValidQuery() {
```
ykmr1224, line 409: nit: Please use blank line to separate Arrange/Act/Assert section for readability.

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
```diff
@@ -404,15 +407,40 @@ void testAutoRefresh() {
...
-    List<String> errors = SQLQueryUtils.validateSparkSqlQuery(validQuery);
+    DataSource dataSource = Mockito.mock(DataSource.class);
+    Mockito.when(dataSource.getConnectorType()).thenReturn(DataSourceType.PROMETHEUS);
+    String validQuery = "DELETE FROM Customers WHERE CustomerName='Alfreds Futterkiste'";
+    List<String> errors = SQLQueryUtils.validateSparkSqlQuery(dataSource, validQuery);
     assertTrue(errors.isEmpty(), "Valid query should not produce any errors");
   }
 
+  @Test
+  void testValidateSparkSqlQuery_SelectQuery_DataSourceSecurityLake() {
```
ykmr1224, line 418: I think we want to have more test cases for positive and negative cases, such as `SELECT` query is included as part of the `CREATE` query, etc. To avoid redundancy, you'd want to extract methods like `verifyValidSecurityLakeQuery`.

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
```diff
@@ -404,15 +407,40 @@ void testAutoRefresh() {
...
+    Mockito.when(dataSource.getConnectorType()).thenReturn(DataSourceType.SECURITY_LAKE);
+    String validQuery = "SELECT * FROM users WHERE age > 18";
+    List<String> errors = SQLQueryUtils.validateSparkSqlQuery(dataSource, validQuery);
+    assertTrue(errors.isEmpty(), "Valid query should not produce any errors ");
+  }
+
+  @Test
+  void testValidateSparkSqlQuery_SelectQuery_DataSourceSecurityLake_ValidationFails() {
+    DataSource dataSource = Mockito.mock(DataSource.class);
+    Mockito.when(dataSource.getConnectorType()).thenReturn(DataSourceType.SECURITY_LAKE);
```
ykmr1224, line 429: nit: Can we static import?

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
```diff
@@ -77,32 +83,73 @@ public static boolean isFlintExtensionQuery(String sqlQuery) {
...
+      return new SparkSqlSecurityLakeValidatorVisitor();
+    } else {
+      return new SparkSqlValidatorVisitor();
+    }
+  }
 
-    @Getter private final List<String> validationErrors = new ArrayList<>();
+  /**
+   * A generic validator impl for Spark Sql Queries that supports accumulating validation errors on
+   * visiting sql statement
```
ykmr1224, line 117: nit: this sounds a bit weird to me. should it say this is a base class instead of generic validator impl?

core/src/main/java/org/opensearch/sql/datasource/model/DataSourceType.java
```diff
@@ -56,4 +57,17 @@ public static DataSourceType fromString(String name) {
...
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    DataSourceType that = (DataSourceType) o;
+    return Objects.equals(name, that.name);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(name);
+  }
```
ykmr1224, line 72: Can we use `@EqualsAndHashCode` annotation instead?

If we need our own implementation, we should add unit tests.

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
eirsep: for better debuggability if there is a syntax error?

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
eirsep: isn't @Mock only used for class level variables?
If yes, is it necessary to declare this variable at class level? I am using `dataSource` variable only in 2-3 methods.

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
eirsep: added

core/src/main/java/org/opensearch/sql/datasource/model/DataSourceType.java
eirsep: changed to use `@EqualsAndHashCode` annotation

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
eirsep: fixed

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
eirsep: changed to static import

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
eirsep: done

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
eirsep: reverted

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
ykmr1224: As normal flow go through here, the log level should be `info` or `debug`.

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
ykmr1224: You can define dataSource with `@Mock` annotation once, and can use it in multiple test cases. MockitoExtension will initialize for each tests.

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
ykmr1224: As I mentioned in the above comment, can we extract method and reduce redundancy?

eirsep: done

async-query-core/src/test/java/org/opensearch/sql/spark/utils/SQLQueryUtilsTest.java
eirsep: done

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
eirsep: I am not sure if this is an info or debug level log
Imagine a bad sql query we are debugging in a log deep dive. It would be easier to trace it correctly if we throw error logs and user can also keep an eye on them.
this is an `error` level log because we have caught an exception from a user input or from another system. I strongly feel this should be retained as `error` level log

ykmr1224: I thought this skips SyntaxCheckException intentionally. @vamsi-amazon any thoughts?

ykmr1224 (dismissed):

async-query-core/src/main/java/org/opensearch/sql/spark/utils/SQLQueryUtils.java
vamsi-amazon: Yeah, I was little apprehensive to add error logs because..sometimes g4 files are not updated to the latest one from spark repo and this can result in syntax errors for valid queries. So, in case of syntax exceptions, I have decided to forward the query to spark in case of validation errors.

Now since we are clear on which spark version we are going to support, we should take g4 files from that branch and throw validation error in case of exception. what do you think @ykmr1224 ?

vamsi-amazon (commented):
LGTM. If possible, can we add more types of queries in SQLQueryUtils test.

vamsi-amazon (approved):

ykmr1224 (approved):

PR #2961, opened 2024-09-03
Flint query scheduler part 2

noCharger:
### Description
1. Abstract `AsyncQueryScheduler` interface, with `AsyncQuerySchedulerRequest` model
2. Inject `AsyncQueryScheduler` to `FlintIndexOpFactory` to support `DROP`, `VACUUM`, and `ALTER ... WITH (auto_refresh = false)`
3. Register OpenSearch scheduler index as system index
4. Refactor refresh job to support any scheduled query, decoupled scheduler job from `AsyncQueryScheduler`
5. FlintStreamingJobHouseKeeperTask uses `flintIndexOpFactory` so it integrates seamlessly

#### Drop should unschedule

```
% curl -X POST -H "Content-Type: application/json" -d '{
  "datasource": "myglue_test",
  "lang": "sql",
  "query": "DROP MATERIALIZED VIEW myglue_test.default.count_by_status"
}' http://localhost:9200/_plugins/_async_query
{
  "queryId": "UDVuUE1jdDFqVm15Z2x1ZV90ZXN0"
}% 

                        
[2024-09-02T15:24:19,262][INFO ][o.o.s.s.a.AsyncQueryExecutorServiceImpl] CreateAsyncQueryRequest: DROP MATERIALIZED VIEW myglue_test.default.count_by_status
[2024-09-02T15:24:19,307][INFO ][o.o.j.s.JobScheduler     ] Descheduling jobId: flint_myglue_test_default_count_by_status
[2024-09-02T15:24:19,348][INFO ][o.o.s.s.s.OpenSearchAsyncQueryScheduler]  Unscheduled job for jobId: flint_myglue_test_default_count_by_status     
```

#### Vacuum should remove schedule

```
% curl -X POST -H "Content-Type: application/json" -d '{
  "datasource": "myglue_test",
  "lang": "sql",
  "query": "VACUUM MATERIALIZED VIEW myglue_test.default.count_by_status"
}' http://localhost:9200/_plugins/_async_query
{
  "queryId": "Zk05NXdZdFNET215Z2x1ZV90ZXN0"
}%

[2024-09-02T15:56:22,959][INFO ][o.o.s.s.a.AsyncQueryExecutorServiceImpl] CreateAsyncQueryRequest: VACUUM MATERIALIZED VIEW myglue_test.default.count_by_status
[2024-09-02T15:56:23,027][INFO ][o.o.s.s.f.o.FlintIndexOpVacuum] Vacuuming Flint index flint_myglue_test_default_count_by_status
[...]

PR #2968, opened 2024-09-04
Fix handler for existing query

ykmr1224:
### Description
- Fix handler for existing query
- We changed handler for CREATE INDEX with manual refresh to batch, but did not update the logic for get/cancel query. This lead to manual refresh job to be handled by RefreshQueryHandler

### Related Issues
n/a

### Check List

### DIFF
--- a/async-query-core/src/main/java/org/opensearch/sql/spark/dispatcher/model/JobType.java
+++ b/async-query-core/src/main/java/org/opensearch/sql/spark/dispatcher/model/JobType.java
@@ -8,6 +8,7 @@ package org.opensearch.sql.spark.dispatcher.model;
 public enum JobType {
   INTERACTIVE("interactive"),
   STREAMING("streaming"),
+  REFRESH("refresh"),
   BATCH("batch");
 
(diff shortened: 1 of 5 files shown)


seankao-az (approved):

async-query-core/src/main/java/org/opensearch/sql/spark/dispatcher/model/JobType.java
```diff
@@ -8,6 +8,7 @@
 public enum JobType {
   INTERACTIVE("interactive"),
```
vamsi-amazon, line 9: Should we revisit the definitions of these jobtypes? What is the difference between a streaming and refresh job type.

vamsi-amazon (approved):

seankao-az:
Found some bug..
in AsyncQueryCoreIntegTest.java
some verifyStoreJobMetadataCalled only use 1 argument, when 2 is expected
I think some calls were added by https://github.com/opensearch-project/sql/pull/2961 so they're missed

PR #2986, opened 2024-09-05
Throw IllegalArgumentException when cancelJob failed with ValidationException

ykmr1224:
### Description
- Throw IllegalArgumentException when cancelJob failed with ValidationException
  - It is caused when the Job is not cancellable state such as FAILED or COMPLETED 
- This is to distinguish the error for 4xx and 5xx

### Related Issues
n/a

### Check List

### DIFF
--- a/async-query-core/src/main/java/org/opensearch/sql/spark/client/EmrServerlessClientImpl.java
+++ b/async-query-core/src/main/java/org/opensearch/sql/spark/client/EmrServerlessClientImpl.java
@@ -116,11 +116,16 @@ public class EmrServerlessClientImpl implements EMRServerlessClient {
                   } catch (Throwable t) {
                     if (allowExceptionPropagation) {
                       throw t;
-                    } else {
-                      logger.error("Error while making cancel job request to emr:", t);
-                      metricsService.incrementNumericalMetric(EMR_CANCEL_JOB_REQUEST_FAILURE_COUNT);
-                      throw new RuntimeException(GENERIC_INTERNAL_SERVER_ERROR_MESSAGE);
                     }
+
+                    logger.error("Error while making cancel job request to emr:", t);
+                    metricsService.incrementNumericalMetric(EMR_CANCEL_JOB_REQUEST_FAILURE_COUNT);
+                    if (t instanceof ValidationException) {
+                      throw new IllegalArgumentException(
+                          "The input fails to satisfy the constraints specified by AWS EMR"
+                              + " Serverless.");
+                    }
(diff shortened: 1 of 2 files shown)


async-query-core/src/main/java/org/opensearch/sql/spark/client/EmrServerlessClientImpl.java
```diff
@@ -116,11 +116,16 @@ public CancelJobRunResult cancelJobRun(
...
                       throw t;
-                    } else {
-                      logger.error("Error while making cancel job request to emr:", t);
-                      metricsService.incrementNumericalMetric(EMR_CANCEL_JOB_REQUEST_FAILURE_COUNT);
-                      throw new RuntimeException(GENERIC_INTERNAL_SERVER_ERROR_MESSAGE);
                     }
+
+                    logger.error("Error while making cancel job request to emr:", t);
+                    metricsService.incrementNumericalMetric(EMR_CANCEL_JOB_REQUEST_FAILURE_COUNT);
+                    if (t instanceof ValidationException) {
```
noCharger, line 123: Just curious, n which cases will async query core cause EMR to return a validation exception when cancelJob?

ykmr1224: It is caused when the Job is not cancellable state such as FAILED or COMPLETED. Updated the PR description.

noCharger: What about [ResourceNotFoundException](https://docs.aws.amazon.com/emr-serverless/latest/APIReference/API_CancelJobRun.html#API_CancelJobRun_Errors)

ykmr1224: That should not happen because we retrieve the job info from AsyncQueryJobMetadata before executing cancel.

vamsi-amazon: Is this similar to create job method?

vamsi-amazon (dismissed):

async-query-core/src/main/java/org/opensearch/sql/spark/client/EmrServerlessClientImpl.java
ykmr1224: Yes, I followed the same way as startJobRun method.

penghuo (approved):

vamsi-amazon (approved):

penghuo (approved):
Thx!

PR #2988, opened 2024-09-05
Throw exception for RECOVER INDEX JOB query

ykmr1224:
### Description
- Throw exception for RECOVER INDEX JOB query
- RECOVER INDEX JOB query is intended to be executed directly to spark to recover aborted streaming job, and does not expect execution from SQL plugin/async-query-core.

### Related Issues
n/a

### Check List

### DIFF
--- a/async-query-core/src/main/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcher.java
+++ b/async-query-core/src/main/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcher.java
@@ -119,6 +119,9 @@ public class SparkQueryDispatcher {
     } else if (IndexQueryActionType.REFRESH.equals(indexQueryDetails.getIndexQueryActionType())) {
       // Manual refresh should be handled by batch handler
       return queryHandlerFactory.getRefreshQueryHandler(dispatchQueryRequest.getAccountId());
+    } else if (IndexQueryActionType.RECOVER.equals(indexQueryDetails.getIndexQueryActionType())) {
+      // RECOVER INDEX JOB should not be executed from async-query-core
+      throw new IllegalArgumentException("RECOVER INDEX JOB is not allowed.");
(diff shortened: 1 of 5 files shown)


noCharger (approved):
LGTM but please double check with @dai-chen if there's any concern on disable this path

seankao-az (approved):

async-query-core/src/main/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcher.java
```diff
@@ -119,6 +119,9 @@ private AsyncQueryHandler getQueryHandlerForFlintExtensionQuery(
     } else if (IndexQueryActionType.REFRESH.equals(indexQueryDetails.getIndexQueryActionType())) {
       // Manual refresh should be handled by batch handler
       return queryHandlerFactory.getRefreshQueryHandler(dispatchQueryRequest.getAccountId());
+    } else if (IndexQueryActionType.RECOVER.equals(indexQueryDetails.getIndexQueryActionType())) {
+      // RECOVER INDEX JOB should not be executed from async-query-core
+      throw new IllegalArgumentException("RECOVER INDEX JOB is not allowed.");
```
dai-chen, line 124: throw unsupported operation exception?

dai-chen (approved):

async-query-core/src/main/java/org/opensearch/sql/spark/dispatcher/SparkQueryDispatcher.java
ykmr1224: I checked other codes and `IllegalArgumentException` tend to be used when user specified something unsupported in the parameters. And `UnsupportedOperationException` tend to be used for when the operation (like `create`, `delete`, etc.) itself is not supported. I see mixed cases, though...

PR #2781, opened 2024-06-27
Make aggregation statement compilation robust

LantaoJin:
### Description
This PR is to address the weakness of query compilation for aggregation statement, including query with aggregate function, window function, group-by clause etc.

Here are some examples to demo what this PR fixes. The reproduce details is [here](https://github.com/opensearch-project/sql/issues/2767)
| ID | Query | Description | Before | After (SemanticCheckException) |
| --- | --- | --- | --- | --- |
| 1 | SELECT query_id FROM test_bulk1 GROUP BY name | column not in group-by clause | return `null` | `Field [query_id] must appear in the GROUP BY clause or be used in an aggregate function` |
| 2 | SELECT query_id, max(query_id) FROM test_bulk1 | group-by is missing | `Explicit GROUP BY clause is required because expression [query_id] contains non-aggregated column` | `Explicit GROUP BY clause is required because expression [query_id] contains non-aggregated column` |
| 3 | SELECT query_id FROM test_bulk1 GROUP BY max(query_id) | aggregate functions in group-by | `Server side error during query execution` | `Aggregate function is not allowed in a GROUP BY clause, but found [max]` |
| 4 | SELECT max(query_id) FROM test_bulk1 GROUP BY 1 | group-by position refers to aggregate function | `Server side error during query execution` | `Aggregate function is not allowed in a GROUP BY clause, but found [max]` |
| 5 | SELECT max(query_id) FILTER(WHERE query_id) FROM test_bulk1 | non boolean filter in aggregate | `Client side error during query execution: [filter] must not be null` | `FILTER or HAVING expression must be type boolean, but found [LONG]` |
| 6 | SELECT max(query_id) FILTER(WHERE max(query_id)>10) FROM test_bulk1 | aggregate in aggregate-filter error | `Server side error during query execution` | `Aggregate function is not allowed in a FILTER, but found [max]` |
| 7 | SELECT * FROM test_bulk1 WHERE ROW_NUMBER() OVER(ORDER BY name) > 0  | window function in where-clause | `Server side error during query execution` | `window functions are not allowed in WHERE or HAVING` |

 
### Issues Resolved
Resolves https://github.com/opensearch-project/sql/issues/2767, https://github.com/opensearch-project/sql/issues/910, 
 
### Check List
- [x] New functionality includes testing.
  - [x] All tests pass, including unit test, integration test and doctest
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
  - [x] New functionality has user manual doc added
[...]

PR #3039, opened 2024-09-17
Implement SQL validation based on grammar element

ykmr1224:
### Description
- Implement SQL validation based on grammar element
- This will reject query using deny list based on grammar element
- Deny list can be defined per datasource type

### Related Issues

### Check List


async-query-core/build.gradle
```diff
@@ -130,7 +130,7 @@ jacocoTestCoverageVerification {
             }
             limit {
                 counter = 'BRANCH'
-                minimum = 1.0
+                minimum = 0.9
```
ykmr1224, line 133: Reduced since there was a branch which could not be covered due to Spark SQL grammar.

async-query-core/src/main/java/org/opensearch/sql/spark/validator/SQLQueryValidationVisitor.java
```diff
@@ -0,0 +1,609 @@
...
+        break;
+      case UDF:
+        validateAllowed(GrammarElement.UDF);
+        break;
+    }
+  }
+
+  private void validateAllowed(GrammarElement element) {
+    if (!grammarElementValidator.isValid(element)) {
+      throw new IllegalArgumentException(element + " is not allowed.");
```
penghuo, line 606: add datasource info in exception? and add log.

async-query-core/src/main/java/org/opensearch/sql/spark/validator/FunctionType.java
```diff
@@ -0,0 +1,436 @@
...
+  GENERATOR("Generator"),
+  UDF("User Defined Function");
+
+  private final String name;
+
+  private static final Map<FunctionType, Set<String>> FUNCTION_TYPE_TO_FUNCTION_NAMES_MAP =
+      ImmutableMap.<FunctionType, Set<String>>builder()
+          .put(
+              AGGREGATE,
+              Set.of(
```
penghuo, line 43: this is spark function list? should we limit to Opensearch SQL supported function?

async-query-core/src/main/java/org/opensearch/sql/spark/validator/S3GlueGrammarElementValidator.java
```diff
@@ -0,0 +1,81 @@
...
+import static org.opensearch.sql.spark.validator.GrammarElement.SHOW_VIEWS;
+import static org.opensearch.sql.spark.validator.GrammarElement.TABLESAMPLE;
+import static org.opensearch.sql.spark.validator.GrammarElement.TABLE_VALUED_FUNCTION;
+import static org.opensearch.sql.spark.validator.GrammarElement.TRANSFORM;
+import static org.opensearch.sql.spark.validator.GrammarElement.UDF;
+
+import com.google.common.collect.ImmutableSet;
+import java.util.Set;
+
+public class S3GlueGrammarElementValidator extends DenyListGrammarElementValidator {
```
penghuo, line 42: make it as configure file? the benfit is it could be changed without recompile code.

async-query-core/src/main/java/org/opensearch/sql/spark/validator/S3GlueGrammarElementValidator.java
```diff
@@ -0,0 +1,81 @@
...
+              DROP_FUNCTION,
+              DROP_VIEW,
+              INSERT,
+              LOAD,
+              CLUSTER_BY,
+              DISTRIBUTE_BY,
+              HINTS,
+              INLINE_TABLE,
+              FILE,
+              CROSS_JOIN,
```
penghuo, line 58: disable Join? why?

async-query-core/src/main/java/org/opensearch/sql/spark/validator/S3GlueGrammarElementValidator.java
ykmr1224: That's possible, but I don't see much benefit to make it configuration file unless we make it configurable at runtime.
With class file, we can do compile time validation.

async-query-core/src/main/java/org/opensearch/sql/spark/validator/FunctionType.java
ykmr1224: Yes, it is from Spark function list. Let me check if we want to limit per function...

penghuo: should we align function with OpenSearch SQL Reference?

async-query-core/src/main/java/org/opensearch/sql/spark/validator/S3GlueGrammarElementValidator.java
ykmr1224: It is because those joins are expensive.

async-query-core/src/main/java/org/opensearch/sql/spark/validator/SQLQueryValidationVisitor.java
ykmr1224: Added logging at upper level as this class doesn't know the datasourceType.

async-query-core/src/main/java/org/opensearch/sql/spark/validator/FunctionType.java
ykmr1224: The idea is that we want to allow the functions available and don't handicap the functionality.

async-query-core/src/main/java/org/opensearch/sql/spark/validator/S3GlueGrammarElementValidator.java
penghuo: PPL add JOIN support with ALL join types. https://github.com/opensearch-project/opensearch-spark/blob/main/docs/PPL-Join-command.md

We should align PPL and SQL capability.

ykmr1224:
Filed a issue for update/test spark version: https://github.com/opensearch-project/sql/issues/3041

penghuo (approved):
Thx!

async-query-core/build.gradle
dai-chen: Can we add it to the exclude list above instead of reducing this on entire module?

async-query-core/src/main/java/org/opensearch/sql/spark/validator/S3GlueGrammarElementValidator.java
dai-chen: Np: I don't see any different behavior in each subclass. Probably no need to make DenyListGrammarElementValidator abstract.

dai-chen (dismissed):

async-query-core/src/main/java/org/opensearch/sql/spark/validator/S3GlueGrammarElementValidator.java
ykmr1224: Do you mean we don't need subclass `S3GlueGrammarElementValidator` or parent class `DenyListGrammarElementValidator`?
For S3GlueGrammarElementValidator, we can make it just an instance of DenyListGrammarElementValidator, but that would move the list of denied grammar elements into Module file, and make it harder to have unit tests for this S3Glue specific validation.

dai-chen (approved):

penghuo (approved):

LantaoJin:
This PR reminds me of another PR https://github.com/opensearch-project/sql/pull/2781 which is an implementation the Aggregate validation based on grammar. I haven't go through the whole details of this PR. Does this PR could resolve the problems described in https://github.com/opensearch-project/sql/pull/2781#issue-2378135613? @ykmr1224 @penghuo

ykmr1224:
> This PR reminds me of another PR #2781 which is an implementation the Aggregate validation based on grammar. I haven't go through the whole details of this PR. Does this PR could resolve the problems described in [#2781 (comment)](https://github.com/opensearch-project/sql/pull/2781#issue-2378135613)? @ykmr1224 @penghuo

No, I think #2781 is for query executed in SQL plugin, but this PR is for async-query. And it is rather prohibiting grammar elements available in Spark SQL.

PR #3061, opened 2024-10-08
Fix record skipping when querying paginated data across shards

Swiddis:
### Description
When pulling unordered data from an index with multiple shards, data gets lost if the fetchSize is not a multiple of the shard count, as the persisted cursor position to continue paging is based on the last seen `_doc` which is duplicated when the primary shard count exceeds 1. This PR currently adds a reproducer for the bug -- finding a fix is still in progress.

### Related Issues
#3064 

### Check List


integ-test/src/test/java/org/opensearch/sql/sql/PaginationWindowIT.java
```diff
@@ -92,4 +96,39 @@ public void testFetchSizeLargerThanResultWindowFails() throws IOException {
         () -> executeQueryTemplate("SELECT * FROM %s", TEST_INDEX_PHRASE, window + 1));
     resetMaxResultWindow(TEST_INDEX_PHRASE);
   }
+
+  @Test
+  public void testMultiShardPagesEqualsActualData() throws IOException {
+    // A bug made it so when pulling unordered data from an index with multiple shards, data gets lost if the fetchSize
```
penghuo, line 102: > pulling unordered data from an index
it is not accurate, it is ordered data (sord by _doc)?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
```diff
@@ -189,6 +189,9 @@ public OpenSearchResponse searchWithPIT(Function<SearchRequest, SearchResponse>
       // Set sort field for search_after
       if (this.sourceBuilder.sorts() == null) {
         this.sourceBuilder.sort(DOC_FIELD_NAME, ASC);
+        // Workaround to preserve sort location more exactly,
```
penghuo, line 192: if this is workaround, could u add the long-term solution issue?

penghuo (commented):
Does it impact Join in legacy module also?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
```diff
@@ -189,6 +189,9 @@ public OpenSearchResponse searchWithPIT(Function<SearchRequest, SearchResponse>
...
+        // see https://github.com/opensearch-project/sql/pull/3061
+        this.sourceBuilder.sort("_id", ASC);
```
fddattal, line 194: I see there are 3 places in the code where we sort on doc field.
Can you help me understand why we only need this here?

https://github.com/search?q=repo%3Aopensearch-project%2Fsql%20%22sort(DOC_FIELD_NAME%2C%20ASC)%22&type=code

integ-test/src/test/java/org/opensearch/sql/sql/PaginationWindowIT.java
Swiddis: Unordered meaning that there's no `ORDER BY` clause supplied by the user -- in this case we default to sorting by `_doc` but semantically the actual order doesn't matter. I wasn't there when it was written, but I suspect sorting by `_doc` is meant to line up with other SQL flavors that have a hidden `ROWID` that they use by default.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
Swiddis: #3064

penghuo (approved):
Thx!

penghuo:
what is the plan for Join in legacy module?
it is ok to public separate PR, just want to make sure it is planned.

legacy/src/main/java/org/opensearch/sql/legacy/query/DefaultQueryAction.java
```diff
@@ -110,6 +110,7 @@ public void checkAndSetScroll() {
         boolean ordered = select.isOrderdSelect();
         if (!ordered) {
           request.addSort(FieldSortBuilder.DOC_FIELD_NAME, SortOrder.ASC);
+          request.addSort("_id", SortOrder.ASC);
```
vamsi-amazon, line 113: Nit: Use constants for this string "_id"

Swiddis: I was going to put it next to doc_field_name but that's not part of this repo, so I'd need another place to put it -- where's a good place?

Swiddis: I think ideally we want to push for this [getting fixed upstream](https://github.com/opensearch-project/OpenSearch/blob/d6ea8eb56077de73e01d8a8a4eb11d3d4a52dea8/server/src/main/java/org/opensearch/search/sort/FieldSortBuilder.java#L102), and it will propagate back down to us. That builder class in OS seems to be where this constant is "supposed" to go. If upstream can update to a new field, that would let us remove all these refs and fix #3064 in the same breath.

penghuo (approved):
Thx!

vamsi-amazon (approved):

PR #3063, opened 2024-10-09
Fix: CSV and Raw output, escape quotes

### LINKED ISSUE #3050: [BUG] CSV output doesn't handle entries with quotes but not commas.
**What is the bug?**
CSV output of entries like `"a" b` or `a " b` should, according to RFC 4180, escape them to be `"""a"" b"` and `"a "" b"`, respectively.  See section 2.5, https://www.rfc-editor.org/rfc/rfc4180

1. Add values to a test index.
```
#!/usr/bin/env bash
# Assuming OPENSEARCH_SERVER is the server,
# OPENSEARCH_USER is the host, OPENSEARCH_PASSWORD is the password.
# and the user has the rights to create and query this index.
# Add an entry to the index that has two quotes in it, not at the boundaries.
curl -XPUT "https://${OPENSEARCH_SERVER}:9200/csv_error/_doc/1" \
    -d'{  "field1": "\"a\" b" }' --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
# Add an entry that has a quote in it.
curl -XPUT "https://${OPENSEARCH_SERVER}:9200/csv_error/_doc/2" \
    -d'{ "field1": "\"a\" b" }' --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
# Optional, add an entry that has no quotes or commas.
curl -XPUT "https://${OPENSEARCH_SERVER}:9200/csv_error/_doc/3" \
    -d'{ "field1": "a b" }' --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
# Optional, add an entry that has commas.
curl -XPUT "https://${OPENSEARCH_SERVER}:9200/csv_error/_doc/4" \
    -d'{ "field1": "a, b" }' --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
```
2. Query the index to CSV output.
```
#!/usr/bin/env bash
curl -XPOST "https://${OPENSEARCH_SERVER}:9200/_plugins/_sql?format=csv" \
    -d'{ "query": "SELECT * FROM csv_error LIMIT 50" }' \
     --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
```
3. Response data is:
```
"a" b  # expected to be """a"" b"
a " b  # expected to be "a "" b"
a b     # correct
"a, b" # correct
```

The CSV output escapes double quotes properly.

 - OS: Ubuntu 20.04
 - Version - 2.17.0.0
 - Plugins - default

As far as I can tell, the problem appears to be in function `quoteIfRequired` in `FlatResponseBase.java`, https://github.com/opensearch-project/sql/blob/c1e623d6d3464056a4b16ae7e51dca218d73a5b6/protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java#L85 but it's possible changing it there might affect other output formats.

Michael-S:
Fixes #3050

### Description

### Related Issues

### Check List
- [x ] New functionality includes testing.
- [ ] New functionality has been documented. (N/A, bug fix)
 - [ ] New functionality has javadoc added. (N/A, bug fix)
 - [ ] New functionality has a user manual doc added. (N/A, bug fix)
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md). (N/A, bug fix)
- [x ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java
+++ b/protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java
@@ -84,6 +84,10 @@ public class FlatResponseBase {
 
   protected String quoteIfRequired(String separator, String cell) {
     final String quote = "\"";
-    return cell.contains(separator) ? quote + cell.replaceAll("\"", "\"\"") + quote : cell;
+    if (cell != null && (cell.contains(separator) || cell.contains(quote))) {
+      return quote + cell.replaceAll(quote, quote + quote) + quote;
+    } else {
+      return cell;
+    }
   }
 }
--- a/protocol/src/test/java/org/opensearch/sql/protocol/response/format/CsvResponseFormatterTest.java
+++ b/protocol/src/test/java/org/opensearch/sql/protocol/response/format/CsvResponseFormatterTest.java
@@ -108,8 +108,11 @@ public class CsvResponseFormatterTest {
     QueryResult response =
         new QueryResult(
             schema,
-            Arrays.asList(tupleValue(ImmutableMap.of("na,me", "John,Smith", ",,age", "30,,,"))));
-    String expected = "\"na,me\",\",,age\"%n\"John,Smith\",\"30,,,\"";
+            Arrays.asList(
+                tupleValue(ImmutableMap.of("na,me", "John,Smith", ",,age", "30,,,")),
+                tupleValue(ImmutableMap.of("na,me", "\"Janice Jones", ",,age", "26\""))));
+    String expected =
+        "\"na,me\",\",,age\"%n\"John,Smith\",\"30,,,\"%n\"\"\"Janice Jones\",\"26\"\"\"";
     assertEquals(format(expected), formatter.format(response));
   }
 
--- a/protocol/src/test/java/org/opensearch/sql/protocol/response/format/RawResponseFormatterTest.java
+++ b/protocol/src/test/java/org/opensearch/sql/protocol/response/format/RawResponseFormatterTest.java
@@ -120,10 +120,18 @@ public class RawResponseFormatterTest {
     QueryResult response =
         new QueryResult(
             schema,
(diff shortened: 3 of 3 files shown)


protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java
```diff
@@ -84,6 +84,10 @@ protected List<List<String>> formatData(List<List<String>> lines) {
 
   protected String quoteIfRequired(String separator, String cell) {
     final String quote = "\"";
-    return cell.contains(separator) ? quote + cell.replaceAll("\"", "\"\"") + quote : cell;
+    if (cell != null && (cell.contains(separator) || cell.contains(quote))) {
+      return quote + cell.replaceAll(quote, quote + quote) + quote;
```
dai-chen, line 88: No matter cell contains separator or quote, we always replace quote?

Michael-S: If you read the RFC for the CSV output format, https://www.ietf.org/rfc/rfc4180.txt section 2.5 states:

> Each field may or may not be enclosed in double quotes (however some programs, such as Microsoft Excel, do not use double quotes  at all).  If fields are not enclosed with double quotes, then double quotes may not appear inside the fields.

So my understanding is that we would quote if the field contains separator or quote.

dai-chen (approved):

Swiddis (approved):

MaxKsyunz:
@Swiddis @dai-chen this PR is failing code coverage. 

Was there a plan to add them soon after?

ykmr1224:
> @Swiddis @dai-chen this PR is failing code coverage.
> 
> Was there a plan to add them soon after?

Addressed in #3155

Issue #3011, opened 2024-09-11
[META]PPL new `trendline` command


YANG-DB:
**Is your feature request related to a problem?**

Adding a new PPL `trendline` command to support computing a moving averages of fields.

We would like to support two flavours of moving average:

 ****SMA** : Simple moving average**

- f[i]: The value of field 'f' in the i-th data-point
- n: The number of data-points in the moving window (period)
- t: The current time index

`SMA(t) = (1/n) * Σ(f[i]), where i = t-n+1 to t`

---

****WMA** :  Weighted moving average**

`WMA(t) = Σ(w[i] * f[i]) / Σ(w[i]), where i = t-n+1 to t`
Where w[i] is the weight for the i-th data-point. 

In a typical WMA, the weights are linearly decreasing from the most recent to the oldest data-point:
`w[i] = n - (t - i), where i = t-n+1 to t`

The complete forumlation would be:
`WMA(t) = Σ((n - (t - i)) * f[i]) / Σ(n - (t - i)), where i = t-n+1 to t`

---

**Example**

The next command shows a trendline over a 5 month period events by month
```sql
source=t | stats count(date_month) | trendline sma(5, count) AS trend | fields  trend
```
The next command would compute a 5-point simple moving average of the 'cpu_usage' field and store it in a new field called 'smooth_cpu'.
```sql
source=t| trendline sma(5,cpu_usage) as smooth_cpu
```

Multiple trendlines could be calculated in a single command, such as 
```sql
| trendline sma(10,memory) as mem_trend wma(5,network_traffic) as net_trend.
```
---

Support for PPL `trendline` functionality is required for both:

 _- OpenSearch based PPL engine_ 
- https://github.com/opensearch-project/sql/issues/3013
- https://github.com/opensearch-project/sql/issues/3277

 _- Spark based PPL engine_
- https://github.com/opensearch-project/opensearch-spark/issues/655

**Do you have any additional context?**

LantaoJin:
Hi @YANG-DB , could you provide any background about distinguishing between `trendline` and `stats`? Based on the examples above, I don't get the key of difference between them.

YANG-DB:
Hi
'stats' command only supports standard [average]( https://github.com/opensearch-project/sql/blob/main/docs%2Fuser%2Fppl%2Fcmd%2Fstats.rst#L117) without the sliding average window, we could add that support to 'stats' command but I think that separating the sliding window average from standard stats will simply the actual usage.
[...]

PR #3071, opened 2024-10-12
Add trendline PPL command

### LINKED ISSUE #3013: [FEATURE]PPL new `trendline` command

Adding a new PPL `trendline` command to support computing a moving averages of fields.

We would like to support two flavours of moving average:

 ****SMA** : Simple moving average**

- f[i]: The value of field 'f' in the i-th data-point
- n: The number of data-points in the moving window (period)
- t: The current time index

`SMA(t) = (1/n) * Σ(f[i]), where i = t-n+1 to t`

---

****WMA** :  Weighted moving average**

`WMA(t) = Σ(w[i] * f[i]) / Σ(w[i]), where i = t-n+1 to t`
Where w[i] is the weight for the i-th data-point. 

In a typical WMA, the weights are linearly decreasing from the most recent to the oldest data-point:
`w[i] = n - (t - i), where i = t-n+1 to t`

The complete forumlation would be:
`WMA(t) = Σ((n - (t - i)) * f[i]) / Σ(n - (t - i)), where i = t-n+1 to t`

---

**Example**

The next command shows a trendline over a 5 month period events by month
```sql
source=t | stats count(date_month) | trendline sma(5, count) AS trend | fields  trend
```
The next command would compute a 5-point simple moving average of the 'cpu_usage' field and store it in a new field called 'smooth_cpu'.
```sql
source=t| trendline sma(5,cpu_usage) as smooth_cpu
```

Multiple trendlines could be calculated in a single command, such as 
```sql
| trendline sma(10,memory) as mem_trend wma(5,network_traffic) as net_trend.
```

jduo:
### Description
Adds the trendline command

### Related Issues
Resolves #3013 
#3011 

### Check List


jduo:
I have this almost hooked up. I loaded the students table which has name, gpa, and grad_year fields.
When I issue this PPL query, it seems like it is using the schema from the implied ProjectOperator instead of using the schema from the TRENDLINE command, even though I overrode TrendlineOperator#schema() to just build a schema based on the computations list:
`{
  "query" : "source=students | TRENDLINE SMA(1, gpa) as foo "
}
`

I get the following JSON result of null arrays:
`
{
    "schema": [
        {
            "name": "grad_year",
            "type": "long"
        },
        {
            "name": "name",
            "type": "string"
        },
        {
            "name": "gpa",
            "type": "float"
        }
    ],
    "datarows": [
        [
            null,
            null,
            null
        ],
        [
            null,
            null,
            null
        ],
        [
            null,
            null,
            null
        ]
    ],
    "total": 3,
    "size": 3
}`

However if change the PPL to use an alias that happens to have the same name as the original field:
`{
  "query" : "source=students | TRENDLINE SMA(1, gpa) as gpa "
}
`
I get data back correctly for one of the array elements in each row.

Is it correct that ProjectOperator does not use the schema from its input?

jduo:
> I have this almost hooked up. I loaded the students table which has name, gpa, and grad_year fields. When I issue this PPL query, it seems like it is using the schema from the implied ProjectOperator instead of using the schema from the TRENDLINE command, even though I overrode TrendlineOperator#schema() to just build a schema based on the computations list: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as foo " } `
> 
> I get the following JSON result of null arrays: ` { "schema": [ { "name": "grad_year", "type": "long" }, { "name": "name", "type": "string" }, { "name": "gpa", "type": "float" } ], "datarows": [ [ null, null, null ], [ null, null, null ], [ null, null, null ] ], "total": 3, "size": 3 }`
> 
> However if change the PPL to use an alias that happens to have the same name as the original field: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as gpa " } ` I get data back correctly for one of the array elements in each row.
> 
> Is it correct that ProjectOperator does not use the schema from its input?

I would expect the only field out of this schema to be the one computation in trendline ("foo"), rather than all 3 fields in the real index, but perhaps I'm mistaken here.

YANG-DB:
> I have this almost hooked up. I loaded the students table which has name, gpa, and grad_year fields. When I issue this PPL query, it seems like it is using the schema from the implied ProjectOperator instead of using the schema from the TRENDLINE command, even though I overrode TrendlineOperator#schema() to just build a schema based on the computations list: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as foo " } `
> 
> I get the following JSON result of null arrays: ` { "schema": [ { "name": "grad_year", "type": "long" }, { "name": "name", "type": "string" }, { "name": "gpa", "type": "float" } ], "datarows": [ [ null, null, null ], [ null, null, null ], [ null, null, null ] ], "total": 3, "size": 3 }`
> 
> However if change the PPL to use an alias that happens to have the same name as the original field: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as gpa " } ` I get data back correctly for one of the array elements in each row.
> 
> Is it correct that ProjectOperator does not use the schema from its input?

@vamsi-amazon @penghuo can you please verify ?

jduo:
> > I have this almost hooked up. I loaded the students table which has name, gpa, and grad_year fields. When I issue this PPL query, it seems like it is using the schema from the implied ProjectOperator instead of using the schema from the TRENDLINE command, even though I overrode TrendlineOperator#schema() to just build a schema based on the computations list: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as foo " } `
> > I get the following JSON result of null arrays: ` { "schema": [ { "name": "grad_year", "type": "long" }, { "name": "name", "type": "string" }, { "name": "gpa", "type": "float" } ], "datarows": [ [ null, null, null ], [ null, null, null ], [ null, null, null ] ], "total": 3, "size": 3 }`
> > However if change the PPL to use an alias that happens to have the same name as the original field: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as gpa " } ` I get data back correctly for one of the array elements in each row.
> > Is it correct that ProjectOperator does not use the schema from its input?
> 
> @vamsi-amazon @penghuo can you please verify ?

Testing Splunk's trendline command, they have the behavior where:
1. If the field in the input is not in the trendline computations, it shows up unaltered.
2. If the field is used in trendline and the computation alias is the same as the field name, it gets replaced with the trendline computation.
3. If the field is used in trendline and the computation alias has a different name than the field name, it shows up as a new field in the result.

jduo:
Requesting reviews from @LantaoJin @MaxKsyunz 
Thanks

YANG-DB:
> > > I have this almost hooked up. I loaded the students table which has name, gpa, and grad_year fields. When I issue this PPL query, it seems like it is using the schema from the implied ProjectOperator instead of using the schema from the TRENDLINE command, even though I overrode TrendlineOperator#schema() to just build a schema based on the computations list: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as foo " } `
> > > I get the following JSON result of null arrays: ` { "schema": [ { "name": "grad_year", "type": "long" }, { "name": "name", "type": "string" }, { "name": "gpa", "type": "float" } ], "datarows": [ [ null, null, null ], [ null, null, null ], [ null, null, null ] ], "total": 3, "size": 3 }`
> > > However if change the PPL to use an alias that happens to have the same name as the original field: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as gpa " } ` I get data back correctly for one of the array elements in each row.
> > > Is it correct that ProjectOperator does not use the schema from its input?
> > 
> > 
> > @vamsi-amazon @penghuo can you please verify ?
> 
> Possible design for trendline output schema:
> 
> 1. If the field in the input is not in the trendline computations, it shows up unaltered.
> 2. If the field is used in trendline and the computation alias is the same as the field name, it gets replaced with the trendline computation.
> 3. If the field is used in trendline and the computation alias has a different name than the field name, it shows up as a new field in the result.

@jduo did you manage to review the spark `trendline` [PR](https://github.com/opensearch-project/opensearch-spark/pull/748) ?



> > > I have this almost hooked up. I loaded the students table which has name, gpa, and grad_year fields. When I issue this PPL query, it seems like it is using the schema from the implied ProjectOperator instead of using the schema from the TRENDLINE command, even though I overrode TrendlineOperator#schema() to just build a schema based on the computations list: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as foo " } `
> > > I get the following JSON result of null arrays: ` { "schema": [ { "name": "grad_year", "type": "long" }, { "name": "name", "type": "string" }, { "name": "gpa", "type": "float" } ], "datarows": [ [ null, null, null ], [ null, null, null ], [ null, null, null ] ], "total": 3, "size": 3 }`
> > > However if change the PPL to use an alias that happens to have the same name as the original field: `{ "query" : "source=students | TRENDLINE SMA(1, gpa) as gpa " } ` I get data back correctly for one of the array elements in each row.
> > > Is it correct that ProjectOperator does not use the schema from its input?
> > 
> > 
> > @vamsi-amazon @penghuo can you please verify ?
> 
> Possible design for trendline output schema:
> 
> 1. If the field in the input is not in the trendline computations, it shows up unaltered.
> 2. If the field is used in trendline and the computation alias is the same as the field name, it gets replaced with the trendline computation.
> 3. If the field is used in trendline and the computation alias has a different name than the field name, it shows up as a new field in the result.

@jduo yes I think it make sense...
@penghuo @dai-chen  ??

jduo:
The majority of the implementation is done. There's some more work left to support datetime types. Only simple moving average is implemented, not weighted moving average.

jduo:
Datetime support has been added so this is effectively code complete (only supporting simple moving average for this iteration).

jduo:
@MaxKsyunz @Yury-Fridlyand

docs/user/ppl/cmd/trendline.rst
```diff
@@ -0,0 +1,63 @@
...
+
+
+Description
+============
+| Use the ``trendline`` command to calculate the moving average on one or more fields in a search result.
+
+
+Syntax
+============
+trendline <average-type>(<number-of-samples>, <source-field>) AS <target-field> [" " <average-type>(<number-of-samples>, <source-field>) AS <target-field> ]...
```
Yury-Fridlyand, line 19: Create a code block for syntax please

ppl/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -54,6 +55,10 @@ STR:                                'STR';
 IP:                                 'IP';
 NUM:                                'NUM';
 
+// TRENDLINE KEYWORDS
+SMA:                                'SMA';
+WMA:                                'WMA';
```
Yury-Fridlyand, line 60: Maybe don't parse WMA if it is not supported/implemented yet

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
```diff
@@ -0,0 +1,56 @@
...
+  @Test
+  public void testTrendlineOverwritesExistingField() throws IOException {
+    final JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | where balance > 39000 | sort balance | trendline sma(2, balance) as"
+                    + " age | fields age",
+                TEST_INDEX_BANK));
+    verifyDataRows(result, rows(new Object[] {null}), rows(44313.0), rows(39882.5));
+  }
```
Yury-Fridlyand, line 55: Please test error cases - `wma` and wrong data types.
How does it work if a column contains nulls or only nulls? A missing field/column case?

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
```diff
@@ -316,5 +329,14 @@ public String visitAlias(Alias node, String context) {
       String expr = node.getDelegated().accept(this, context);
       return StringUtils.format("%s", expr);
     }
+
+    @Override
+    public String visitTrendlineComputation(Trendline.TrendlineComputation node, String context) {
+      final String dataField = node.getDataField().accept(this, context);
+      final String aliasClause = " as " + node.getAlias();
+      final String computationType = node.getComputationType().name().toLowerCase(Locale.ROOT);
```
Yury-Fridlyand, line 337: In `Trendline.java` you use uppercase. Just to confirm that there is no conflict.

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
jduo: If a column has a null it doesn't count as a sample (similar to how in SQL if you average a column and there are nulls they don't increment the divisor).

jduo: I removed the WMA entry in the enum for now. Wrong data types are covered in the TrendlineOperatorTest unit test -- is it common to duplicate that type of test in integration?

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
jduo: Yes this is intentional. It's upper-cased there so that it can be passed to Enum.valueOf() correctly, but lower-cased here which seems to be the standard for writing PPL.

ppl/src/main/antlr/OpenSearchPPLLexer.g4
jduo: Done.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -0,0 +1,324 @@
...
+    static ArithmeticEvaluator getEvaluator(ExprCoreType type) {
+      switch (type) {
+        case DOUBLE:
+          return NumericArithmeticEvaluator.INSTANCE;
+        case DATE:
+          return DateArithmeticEvaluator.INSTANCE;
+        case TIME:
+          return TimeArithmeticEvaluator.INSTANCE;
+        case TIMESTAMP:
+          return TimestampArithmeticEvaluator.INSTANCE;
```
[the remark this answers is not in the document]
jduo: I have timestamp converting to epoch millis (and date delegate to timestamp but truncating the time portion). This is based on the behavior of the AvgState class used for stats.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -0,0 +1,324 @@
...
+        return new SimpleMovingAverageAccumulator(computation.getKey(), computation.getValue());
+      case WMA:
+      default:
+        throw new IllegalStateException(
+            "Unexpected value: " + computation.getKey().getComputationType());
+    }
+  }
+
+  /** Maintains stateful information for calculating the trendline. */
+  private interface TrendlineAccumulator {
```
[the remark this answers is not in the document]
jduo: I'm leaving it here since these APIs are tightly-coupled to Trendline and not likely to be used independently. I don't have a strong opinion on this though.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -0,0 +1,324 @@
...
+    return tupleValue;
+  }
+
+  private static TrendlineAccumulator createAccumulator(
+      Pair<Trendline.TrendlineComputation, ExprCoreType> computation) {
+    switch (computation.getKey().getComputationType()) {
+      case SMA:
+        return new SimpleMovingAverageAccumulator(computation.getKey(), computation.getValue());
+      case WMA:
+      default:
```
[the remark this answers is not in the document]
jduo: WMA is not implemented. I've removed WMA now.

core/src/main/java/org/opensearch/sql/ast/tree/Trendline.java
```diff
@@ -0,0 +1,74 @@
...
+    return ImmutableList.of(child);
+  }
+
+  @Override
+  public <T, C> T accept(AbstractNodeVisitor<T, C> visitor, C context) {
+    return visitor.visitTrendline(this, context);
+  }
+
+  @Getter
+  public static class TrendlineComputation extends UnresolvedExpression {
```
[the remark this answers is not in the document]
jduo: This is pulled from the corresponding Spark PR opensearch-spark/#748 though this file has changed in that PR too. I think it makes sense as an inner class here as it is Trendline-specific, though it does get used by both LogicalTrendline and TrendlineOperator.

docs/user/ppl/cmd/trendline.rst
```diff
@@ -0,0 +1,63 @@
+=============
+rename
```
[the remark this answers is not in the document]
jduo: Done.

docs/user/ppl/cmd/trendline.rst
jduo: All other commands are not using a code block, so this is consistent.

core/src/main/java/org/opensearch/sql/ast/tree/Trendline.java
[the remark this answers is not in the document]
Yury-Fridlyand: I just compared list of classes with other features we added in past. I don't have a preference there.

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
Yury-Fridlyand: There are 3 reasons:
1. A lot of things mocked in UTs, so without IT you can't see an accumulated error (swiss cheese scenario).
2. IT can (and should) cover edge cases too.
3. You can easily convert IT to doctests. Once you have a dataset and queries, it is easy to add user docs/guides or record a demo using them.

jduo:
This functionality hasn't been tested with pagination.

jduo:
There are syntax changes in the Spark version of this command that haven't been reflected here (opensearch-spark/#833).
- Alias has been made optional.
- There is now an optional sort specification.

I think the latter we could support here by having Analyzer return a LogicalSort on top of the LogicalTrendline if the sort clause is added.

It's not clear from me in the Spark version what the field name should be if an alias is not specified.

salyh:
> There are syntax changes in the Spark version of this command that haven't been reflected here (opensearch-spark/#833).
> 
>     * Alias has been made optional.
> 
>     * There is now an optional sort specification.
> 
> 
> I think the latter we could support here by having Analyzer return a LogicalSort on top of the LogicalTrendline if the sort clause is added.
> 
> It's not clear from me in the Spark version what the field name should be if an alias is not specified. @YANG-DB @salyh @kt-eliatra

https://github.com/opensearch-project/opensearch-spark/pull/833#discussion_r1823003234

jduo:
@salyh @YANG-DB , I have added the sort functionality and made alias optional.

jduo:
Rebasing on upstream/main to see if the bwc test failures go away.

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
jduo: I'm going to hold off on adding ITs for edge/error cases as I think the unit tests catch these scenarios well enough.

jduo:
@salyh , @YANG-DB , @LantaoJin would you be able to review this PR? As the Spark PR finished it might make sense to do this while it is still fresh.

YANG-DB (commented):
@jduo please update the ANTLR as commented

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -49,6 +49,7 @@ commands
    | kmeansCommand
    | adCommand
    | mlCommand
+   | trendlineCommand
```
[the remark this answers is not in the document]
jduo: I've added TRENDLINE to keywordsCanBeId.

Note that it wasn't added in the Spark implementation - should it?
Should SMA also be added to this list?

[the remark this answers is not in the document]
andy-k-improving: @YANG-DB Do we need to a ticket on the `Flint` side review on keywordsCanBeId?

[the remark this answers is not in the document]
YANG-DB: @jduo  in spark `trendline` is present under [commandName](https://github.com/opensearch-project/opensearch-spark/blob/3ff2ef23739bc900be3ea4fdac5b9b2615f40265/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4#L92) which is added directly to the  [keywordsCanBeId](https://github.com/opensearch-project/opensearch-spark/blob/3ff2ef23739bc900be3ea4fdac5b9b2615f40265/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4#L1092)

YANG-DB (dismissed):

YANG-DB:
@jduo can u plz check why the CI build [failed](https://github.com/opensearch-project/sql/actions/runs/12146465231/job/33870457556?pr=3075#step:6:82654) ?
is it related to this PR or generic ?

jduo:
> @jduo can u plz check why the CI build [failed](https://github.com/opensearch-project/sql/actions/runs/12146465231/job/33870457556?pr=3075#step:6:82654) ? is it related to this PR or generic ?

There's a mistake with the field name in the docs. I've updated now.

YANG-DB (approved):

YANG-DB (dismissed):
Thanks LGTM !!
@jduo can u plz check the failed CI tasks ?

jduo:
> Thanks LGTM !! @jduo can u plz check the failed CI tasks ?

Looks like a merge error in the grammar (lost a semi-colon). I'll post in an update shortly.

core/src/main/java/org/opensearch/sql/ast/tree/Trendline.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public <R, C> R accept(AbstractNodeVisitor<R, C> nodeVisitor, C context) {
+      return nodeVisitor.visitTrendlineComputation(this, context);
+    }
+  }
+
+  public enum TrendlineType {
+    SMA,
+    WMA
```
acarbonetto, line 70: we should remove WMA here too.  We aren't testing this scenario and the grammar doesn't (yet) support it.

docs/user/ppl/cmd/trendline.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Syntax
+============
+`TRENDLINE [sort <[+|-] sort-field>] SMA(number-of-datapoints, field) [AS alias] [SMA(number-of-datapoints, field) [AS alias]]...`
+
+* [+|-]: optional. The plus [+] stands for ascending order and NULL/MISSING first and a minus [-] stands for descending order and NULL/MISSING last. **Default:** ascending order and NULL/MISSING first.
+* sort-field: mandatory when sorting is used. The field used to sort.
+* number-of-datapoints: mandatory. number of datapoints to calculate the moving average (must be greater than zero).
+* field: mandatory. the name of the field the moving average should be calculated for.
+* alias: optional. the name of the resulting column containing the moving average.
```
acarbonetto, line 24: We could add something like: 
```
By default, the column name appends "_trendline" to the field name.
```

YANG-DB (dismissed):

YANG-DB:
@jduo LGTM !
@Yury-Fridlyand @acarbonetto can u plz review ?

acarbonetto (approved):

PR #3092, opened 2024-10-21
Remove scheduler index from SystemIndexDescriptor

noCharger:
### Description
Remove scheduler index from SystemIndexDescriptor

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
@@ -342,9 +342,6 @@ public class SQLPlugin extends Plugin
     systemIndexDescriptors.add(
         new SystemIndexDescriptor(
             SPARK_REQUEST_BUFFER_INDEX_NAME + "*", "SQL Spark Request Buffer index pattern"));
-    systemIndexDescriptors.add(
-        new SystemIndexDescriptor(
-            OpenSearchAsyncQueryScheduler.SCHEDULER_INDEX_NAME, "SQL Scheduler job index"));
     return systemIndexDescriptors;
   }
 }


plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -342,9 +342,6 @@ public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings sett
     systemIndexDescriptors.add(
         new SystemIndexDescriptor(
             SPARK_REQUEST_BUFFER_INDEX_NAME + "*", "SQL Spark Request Buffer index pattern"));
-    systemIndexDescriptors.add(
-        new SystemIndexDescriptor(
-            OpenSearchAsyncQueryScheduler.SCHEDULER_INDEX_NAME, "SQL Scheduler job index"));
```
ykmr1224, line 347: Should we remove the constant (`SCHEDULER_INDEX_NAME`) as well?

noCharger: no, it's still in use https://github.com/opensearch-project/sql/pull/3092/files#diff-b9ca34b799db4aa7b3c7dbccba65de6f539a5d3d6accf8a1c75e4c6893b9a9b8R264

ykmr1224 (approved):

derek-ho (changes requested):
@noCharger can you help me understand why we are now not registering this as system index, would this be backported to 2.x line?

noCharger:
> @noCharger can you help me understand why we are now not registering this as system index, would this be backported to 2.x line?

@derek-ho This index will be accessed from managed flint, which has a different auth model.

derek-ho (commented):
Removing the block

seankao-az (approved):

derek-ho (commented):
.

derek-ho (approved):

derek-ho (commented):
.

PR #3095, opened 2024-10-22
WIP: Array values are preserved (#1300)

### LINKED ISSUE #1300: [DISCUSSION] Properly support array values in new engine

normanj-bitquill:
### Description
Preserve array values so that they are represented correctly in the result set with all nested values.

### Related Issues
Resolves #1300

### Check List


normanj-bitquill:
I am still working on the code for making this configurable. The idea is to have a query parameter named `arrays` that is optional. If it is included, the only valid value is `preserve`.

If `arrays=preserve`, then arrays returned from the OpenSearch engine can flow to the result set and include all of the nested values.

penghuo (commented):
1. could u add setting `plugins.query.field_type_tolerance`. If the plugins.query.field_type_tolerance setting is set to true, the backend will return scalar data types with array datasets, allowing basic queries like SELECT * FROM tbl WHERE condition; however, using multi-value fields in aggregations, ORDER BY clauses, or expressions will result in exceptions. If the setting is false or does not exist, only the first element of an array is returned, maintaining the current behavior.
2. could u add exception if operator/expression does not support array value, for instance groupby.

integ-test/src/test/java/org/opensearch/sql/legacy/ObjectFieldSelectIT.java
```diff
@@ -81,7 +81,7 @@ public void testSelectObjectFieldOfArrayValuesItself() {
     JSONObject response = new JSONObject(query("SELECT accounts FROM %s"));
 
     // Only the first element of the list of is returned.
```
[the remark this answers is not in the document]
normanj-bitquill: Since the default behaviour is to only use the first element of an array, this comment is still valid.

opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
```diff
@@ -229,6 +229,13 @@ public class OpenSearchSettings extends Settings {
           Setting.Property.NodeScope,
           Setting.Property.Dynamic);
 
+  public static final Setting<?> FIELD_TYPE_TOLERANCE_SETTING =
+      Setting.boolSetting(
+          Key.FIELD_TYPE_TOLERANCE.getKeyValue(),
+          false,
```
penghuo, line 235: Set default = true. the reasons is most of user can get correct result ( similar to DSL response ). Even it is breaking current default behavior.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/OpenSearchResponse.java
```diff
@@ -48,27 +48,35 @@ public class OpenSearchResponse implements Iterable<ExprValue> {
...
   /** OpenSearchExprValueFactory used to build ExprValue from search result. */
   @EqualsAndHashCode.Exclude private final OpenSearchExprValueFactory exprValueFactory;
 
   /** Constructor of OpenSearchResponse. */
   public OpenSearchResponse(
       SearchResponse searchResponse,
       OpenSearchExprValueFactory exprValueFactory,
-      List<String> includes) {
+      List<String> includes,
+      boolean fieldTypeTolerance) {
```
penghuo, line 61: fieldTypeTolerance is not releated to OpenSearchResponse, Could we move this logic to OpenSearchExprValueFactory?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
```diff
@@ -101,6 +103,21 @@ public OpenSearchQueryRequest(
     this.sourceBuilder = sourceBuilder;
     this.exprValueFactory = factory;
     this.includes = includes;
+    this.fieldTypeTolerance = false;
```
penghuo, line 106: fieldTypeTolerance should not associate with OpenSearchRequest, It is been used when OpenSearchExprValueFactory parse response.?

integ-test/src/test/resources/expectedOutput/ppl/explain_filter_agg_push.json
```diff
@@ -8,7 +8,7 @@
...
+          "request": "OpenSearchQueryRequest(indexName\u003dopensearch-sql_test_index_account, sourceBuilder\u003d{\"from\":0,\"size\":10000,\"timeout\":\"1m\",\"query\":{\"range\":{\"age\":{\"from\":30,\"to\":null,\"include_lower\":false,\"include_upper\":true,\"boost\":1.0}}},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}],\"aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"sources\":[{\"state\":{\"terms\":{\"field\":\"state.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}},{\"city\":{\"terms\":{\"field\":\"city.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}}]},\"aggregations\":{\"avg_age\":{\"avg\":{\"field\":\"age\"}}}}}}, needClean\u003dtrue, searchDone\u003dfalse, pitId\u003dnull, cursorKeepAlive\u003dnull, searchAfter\u003dnull, searchResponse\u003dnull, fieldTypeTolerance\u003dfalse)"
```
penghuo, line 11: fieldTypeTolerance is not releated to OpenSearchQueryRequest.

penghuo (changes requested):
Please update doc also https://github.com/opensearch-project/sql/blob/main/docs/user/admin/settings.rst

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
normanj-bitquill: It is only here so that it can be passed down to the OpenSearchResponse. It is not used directly in OpenSearchRequest.

integ-test/src/test/resources/expectedOutput/ppl/explain_filter_agg_push.json
normanj-bitquill: This was the only way I could see to pass the fieldTypeTolerance setting to OpenSearchResponse.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/OpenSearchResponse.java
normanj-bitquill: I can look, but making this change could take hours (with updating the affected tests).

opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
normanj-bitquill: OK.

integ-test/src/test/java/org/opensearch/sql/legacy/ObjectFieldSelectIT.java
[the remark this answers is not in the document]
normanj-bitquill: Removed the comment.

opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
normanj-bitquill: Default is now true.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/OpenSearchResponse.java
normanj-bitquill: Now it is passed through OpenSearchExprValueFactory

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
normanj-bitquill: Updated.

integ-test/src/test/resources/expectedOutput/ppl/explain_filter_agg_push.json
normanj-bitquill: Updated.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
```diff
@@ -151,7 +153,8 @@ public OpenSearchQueryRequest(StreamInput in, OpenSearchStorageEngine engine) th
     }
 
     OpenSearchIndex index = (OpenSearchIndex) engine.getTable(null, indexName.toString());
-    exprValueFactory = new OpenSearchExprValueFactory(index.getFieldOpenSearchTypes());
+    exprValueFactory =
+        new OpenSearchExprValueFactory(index.getFieldOpenSearchTypes(), fieldTypeTolerance);
```
penghuo, line 157: Logic should similar to OpenSearchScrollRequest right? If yes, fieldTypeTolerance should not associate with OpenSearchQueryRequest.

```
    OpenSearchIndex index = (OpenSearchIndex) engine.getTable(null, indexName.toString());
    exprValueFactory =
        new OpenSearchExprValueFactory(
            index.getFieldOpenSearchTypes(), index.isFieldTypeTolerance());
```

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -74,6 +74,8 @@ public class OpenSearchExprValueFactory {
   /** The Mapping of Field and ExprType. */
   private final Map<String, OpenSearchDataType> typeMapping;
 
+  private final boolean fieldTypeTolerance;
```
penghuo, line 77: nit: add doc

opensearch/src/test/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactoryTest.java
```diff
@@ -763,20 +824,31 @@ public void constructBinary() {
...
 
   /**
-   * Return the first element if is OpenSearch Array.
+   * Return the all elements if is OpenSearch Array.
    * https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html.
    */
   @Test
-  public void constructFromOpenSearchArrayReturnFirstElement() {
-    assertEquals(integerValue(1), tupleValue("{\"intV\":[1, 2, 3]}").get("intV"));
+  public void constructFromOpenSearchArrayReturnAll() {
```
penghuo, line 831: we should also verify field_type_tolerance=false works.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
```diff
@@ -47,14 +47,18 @@ public class OpenSearchIndexScan extends TableScanOperator implements Serializab
...
-  private Settings pluginSettings;
+  private boolean fieldTypeTolerance;
 
   /** Creates index scan based on a provided OpenSearchRequestBuilder. */
   public OpenSearchIndexScan(
-      OpenSearchClient client, int maxResponseSize, OpenSearchRequest request) {
+      OpenSearchClient client,
+      boolean fieldTypeTolerance,
+      int maxResponseSize,
+      OpenSearchRequest request) {
```
penghuo, line 57: nit: we could keep original constructor which means fieldTypeTolerance=true, it could avoid unnecessary UT chagne.

opensearch/src/test/java/org/opensearch/sql/opensearch/response/OpenSearchResponseTest.java
```diff
@@ -248,20 +251,15 @@ void iterator_metafields_scoreNaN() {
 
   @Test
   void iterator_with_inner_hits() {
+    Map<String, SearchHits> innerHits = new HashMap<>();
+    innerHits.put("a", mock(SearchHits.class));
+    when(searchHit1.getInnerHits()).thenReturn(innerHits);
```
penghuo, line 256: why change the mock?

integ-test/src/test/java/org/opensearch/sql/legacy/ObjectFieldSelectIT.java
```diff
@@ -79,9 +79,7 @@ public void testSelectNestedFieldItself() {
   @Test
   public void testSelectObjectFieldOfArrayValuesItself() {
     JSONObject response = new JSONObject(query("SELECT accounts FROM %s"));
-
-    // Only the first element of the list of is returned.
-    verifyDataRows(response, rows(new JSONObject("{\"id\": 1}")));
+    verifyDataRows(response, rows(new JSONArray("[{\"id\":1},{\"id\":2}]")));
```
penghuo, line 82: could u test of field_type_tolerance=false

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
Swiddis: +1. I'm also curious if it even needs to be a parameter since every refactor hardcodes it to `true` and the `false` case isn't even tested. What's the use case where we don't want this?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
dai-chen: maybe another option is to create `OpenSearchExprValueFactory` in request builder and pass it to request? In case factory needs more settings passed in future?

integ-test/src/test/java/org/opensearch/sql/ppl/StandaloneIT.java
```diff
@@ -153,6 +153,7 @@ private Settings defaultSettings() {
           new ImmutableMap.Builder<Key, Object>()
               .put(Key.QUERY_SIZE_LIMIT, 200)
               .put(Key.SQL_PAGINATION_API_SEARCH_AFTER, true)
+              .put(Key.FIELD_TYPE_TOLERANCE, false)
```
dai-chen, line 156: I can see some IT set it to true and some set it to false? Maybe most IT should stick with default setting and add dedicate IT to test the opposite?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -164,7 +168,7 @@ public ExprValue construct(String jsonString, boolean supportArrays) {
           new OpenSearchJsonContent(OBJECT_MAPPER.readTree(jsonString)),
           TOP_PATH,
           Optional.of(STRUCT),
-          supportArrays);
+          fieldTypeTolerance || supportArrays);
```
dai-chen, line 171: I'm thinking should we use `fieldTypeTolerance` directly in `parseArray` (not sure `parseGeopoint`)? Because I think `supportArrays` has its own meaning? Unless we change it and rename the argument?

dai-chen (commented):
Please update documentation: https://github.com/opensearch-project/sql/blob/main/docs/user/admin/settings.rst

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
normanj-bitquill: Updated.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
normanj-bitquill: Added a comment.

opensearch/src/test/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactoryTest.java
normanj-bitquill: Added some tests for this.

opensearch/src/test/java/org/opensearch/sql/opensearch/response/OpenSearchResponseTest.java
normanj-bitquill: The mock wasn't actually adding any inner hits before.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -110,6 +110,7 @@ private OpenSearchRequest buildRequestWithPit(
     int size = requestedTotalSize;
     FetchSourceContext fetchSource = this.sourceBuilder.fetchSource();
     List<String> includes = fetchSource != null ? Arrays.asList(fetchSource.includes()) : List.of();
+    boolean fieldTypeTolerance = settings.getSettingValue(Settings.Key.FIELD_TYPE_TOLERANCE);
```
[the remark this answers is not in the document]
normanj-bitquill: Removed.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
normanj-bitquill: This also applies to geopoints and anything else that could be in an array. If `fieldTypeTolerance` is not also applied to `parseGeopoint`, then an array of geopoints would get reduced to only the first element.

integ-test/src/test/java/org/opensearch/sql/ppl/StandaloneIT.java
normanj-bitquill: Changed this to `true` which is the default. There is another IT `ObjectFieldSelectIT.testSelectObjectFieldOfArrayValuesItselfNoFieldTypeTolerance` that tests with field type tolerance set to `false`.

penghuo (approved):
Thanks for the change!
Please update doc in following PR. https://github.com/opensearch-project/sql/blob/main/docs/user/admin/settings.rst

dai-chen (approved):

PR #3139, opened 2024-10-29
Call updateState when query is cancelled

ykmr1224:
### Description
- Call updateState when query is cancelled
  - This change is for async-query-core library consumer to update cancel status
  - This does not affect the async-query behavior since AsyncQueryJobMetadata record does not store status yet. (#3140 is an issue for unifying the model to single one, and we need to handle the update status within the fix)

### Related Issues
#3140 (This PR does not solve this issue)

### Check List


seankao-az (approved):

noCharger:
> This does not affect the async-query behavior since AsyncQueryJobMetadata record does not store status yet

It makes me feel like this PR is a short term solution, is it an urgent change or we can go with longer term approach #3140?

ykmr1224:
> > This does not affect the async-query behavior since AsyncQueryJobMetadata record does not store status yet
> 
> It makes me feel like this PR is a short term solution, is it an urgent change or we can go with longer term approach #3140?

Yes, this is short term, but it does not have side effect. #3140 require much more effort and we don't have bandwidth right now.

noCharger (approved):

Issue #1545, opened 2023-04-18
[BUG] Inconsistent behaviour for date field in nested collection queries


parked-toes:
**What is the bug?**
When a query contains [nested collection](https://opensearch.org/docs/latest/search-plugins/sql/sql/partiql/) (self join) it returns different data and datatype, compared to a "simple" query.

**How can one reproduce the bug?**
1. Create mapping:
```json
PUT datetype_test
{
  "mappings": {
    "properties": {
      "dateAsDate": {
        "type": "date"
      },
      "longAsDate": {
        "type": "date"
      },      
      "id": {
        "type": "long"
      },
      "projects": {
        "type": "nested", 
        "properties": {
          "name": {
            "type": "text",
            "fields": {
              "keyword": {
                "type": "keyword"
              }
            }
          }
        }
      }
    }
  }
}
```

Note that both date fields are mapped as `date`, projects is `nested`

2. Index data:
```json
POST datetype_test/_bulk?refresh
{"index":{"_id":"1"}}
{"id":3,"dateAsDate":"2023-01-01T00:00:00", "longAsDate": 1672531200000, "projects":[{"name":"SQL Spectrum querying"},{"name":"SQL security"},{"name":"OpenSearch security"}]}
{"index":{"_id":"2"}}
{"id":4,"dateAsDate":"2024-01-01T00:00:00", "longAsDate": 1704067200000, "projects":[]}
{"index":{"_id":"3"}}
{"id":6,"dateAsDate":"2025-01-01T00:00:00","longAsDate": 1735689600000, "projects":[{"name":"SQL security"},{"name":"Hello security"}]}
```

3. Query date fields using simple query:
```sql
SELECT t.dateAsDate,
       t.longAsDate
FROM datetype_test AS t
```

Result:
```json
{
  "schema": [
    {
      "name": "dateAsDate",
      "type": "timestamp"
    },
    {
      "name": "longAsDate",
      "type": "timestamp"
    }
  ],
  "datarows": [
    [
      "2023-01-01 00:00:00",
      "2023-01-01 00:00:00"
    ],
    [
      "2024-01-01 00:00:00",
      "2024-01-01 00:00:00"
    ],
    [
      "2025-01-01 00:00:00",
      "2025-01-01 00:00:00"
    ]
  ],
  "total": 3,
  "size": 3,
  "status": 200
}
```

Query **the same** date fields using nested collection query:
```sql
SELECT t.dateAsDate,
       t.longAsDate
[...]

PR #3160, opened 2024-11-20
Fix: Date field format parsing for legacy query engine 

andy-k-improving:
### Description

This is a MR, which aim to fix the issue described over https://github.com/opensearch-project/sql/issues/1545 to align the representation of `date` field between the legacy engine and the v2 engine. 

Code changes highlight:
 - Remove the `break` identifier from `DateFieldFormatter`, the current behaviour will only parse the first `date` field from the resultSet, which is not a valid assumption, because OpenSearch as platform allow multiple `date` fields co-exist on an index, also this is not an uncommon use-case for enterprise. Hence all `date` fields present on any given index should be formatted regardless of the order.
 - To detect and attempt parse any unix timestamp (second || milisecond) on best-effort basis, OpenSearch as platform allow unix timestamp value to be inserted despite `epoch_millis` being explicitly specified on field format, in this particular case there is no clear hint from the query indicate this is numeric timestamp from OpenSearch. 

Although issue is being reported https://github.com/opensearch-project/sql/issues/1545 under nested-query, however this bug has wider impact outside of nested-query, and the conditions to trigger are:
- Query that **unsupported** by the v2 engine and being forwarded to legacy engine
- Multiple `date` fields are being specified on the Index schema WITHOUT any explicit `format` configuration. 

When this happen, the current behaviour will only parse the first `date` field from the resultset.

### Reference doc:
 - OpenSearch Product doc mentioning that integer value will also be accepted. 
https://opensearch.org/docs/latest/field-types/supported-field-types/date/

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/1545

### Testing plan
```
# Create table
PUT http://localhost:9200/datetype_test?pretty

{
  "mappings": {
    "properties": {
      "dateAsDate": {"type": "date"},
      "longAsDate": {"type": "date"},      
      "id": { "type": "long" },
      "projects": {
        "type": "nested", 
        "properties": {
          "name": { "type": "text", "fields": {"keyword": {"type": "keyword"   }
            }
          }}}}}}

# Insert data
POST http://localhost:9200/datetype_test/_bulk?refresh

{"index":{"_id":"1"}}
{"id":3,"dateAsDate":"2023-01-01T00:00:00", "longAsDate": 1672531200000, "projects":[{"name":"SQL Spectrum querying"},{"name":"SQL security"},{"name":"OpenSearch security"}]}
{"index":{"_id":"2"}}
{"id":4,"dateAsDate":"2024-01-01T00:00:00", "longAsDate": 1704067200000, "projects":[]}
{"index":{"_id":"3"}}
{"id":6,"dateAsDate":"2025-01-01T00:00:00","longAsDate": 1735689600000, "projects":[{"name":"SQL security"},{"name":"Hello security"}]}

#Qeury, both fields should be in simple date format.
POST http://localhost:9200/_plugins/_sql

{
  "query" : "SELECT t.id, t.dateAsDate, t.longAsDate FROM datetype_test AS t, t.projects AS p"
}


```

### Check List

### DIFF
--- a/legacy/src/main/java/org/opensearch/sql/legacy/executor/format/DateFieldFormatter.java
+++ b/legacy/src/main/java/org/opensearch/sql/legacy/executor/format/DateFieldFormatter.java
@@ -21,6 +21,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.sql.legacy.esdomain.LocalClusterState;
 import org.opensearch.sql.legacy.esdomain.mapping.FieldMappings;
+import org.opensearch.sql.legacy.utils.StringUtils;
 
 /** Formatter to transform date fields into a consistent format for consumption by clients. */
 public class DateFieldFormatter {
@@ -83,7 +84,7 @@ public class DateFieldFormatter {
       Date date = parseDateString(formats, columnOriginalDate.toString());
       if (date != null) {
         rowSource.put(columnName, DateFormat.getFormattedDate(date, FORMAT_JDBC));
-        break;
+//        break;
       } else {
         LOG.warn("Could not parse date value; returning original value");
       }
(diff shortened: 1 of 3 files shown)


andy-k-improving:
@parked-toes, @dai-chen  and @anasalkouz 
Would you mind to have a look on this?
Thanks!

andy-k-improving:
@dai-chen I have fixed the spotlessJavaCheck, and tested it offline.
Would you mind to trigger the build again?
Thanks,

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/DateFieldFormatter.java
```diff
@@ -152,15 +152,27 @@ private Date parseDateString(List<String> formats, String columnOriginalDate) {
...
-                    columnOriginalDate,
-                    FORMAT_DOT_OPENSEARCH_DASHBOARDS_SAMPLE_DATA_LOGS_EXCEPTION,
-                    FORMAT_DOT_OPENSEARCH_DASHBOARDS_SAMPLE_DATA_FLIGHTS_EXCEPTION,
-                    FORMAT_DOT_OPENSEARCH_DASHBOARDS_SAMPLE_DATA_FLIGHTS_EXCEPTION_NO_TIME,
-                    FORMAT_DOT_OPENSEARCH_DASHBOARDS_SAMPLE_DATA_ECOMMERCE_EXCEPTION,
-                    FORMAT_DOT_DATE_AND_TIME,
-                    FORMAT_DOT_DATE);
+            // It's possible to have date stored in second / millisecond form without explicit
+            // format hint.
+            // Parse it on a best-effort basis.
```
dai-chen, line 157: Could you check if anything in OpenSearch's `DateFormatters` can be reused?

dai-chen:
@andy-k-improving It seem CI is failing on IT task. I've triggered the CI again. Please check if there is still failure.

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/DateFieldFormatter.java
andy-k-improving: I dont see there is a suitable method I can call to detect and parse it on best effort basis, if we decide the use `DataFormatters`, the logic will turns into following:

```
              try {
                parsedDate = Date.from(DateFormatters.from(
                                DateFormatter.forPattern("epoch_millis").parse(columnOriginalDate))
                        .toInstant());
              } catch (DateTimeException ex) {
                // Fallback to epoch_second
                try {
                  parsedDate = Date.from(DateFormatters.from(
                                  DateFormatter.forPattern("epoch_second").parse(columnOriginalDate))
                          .toInstant());
                } catch (DateTimeException innerEx) {
                  parsedDate =
                          DateUtils.parseDate(
                                  columnOriginalDate,
                                  FORMAT_DOT_OPENSEARCH_DASHBOARDS_SAMPLE_DATA_LOGS_EXCEPTION,
                                  FORMAT_DOT_OPENSEARCH_DASHBOARDS_SAMPLE_DATA_FLIGHTS_EXCEPTION,
                                  FORMAT_DOT_OPENSEARCH_DASHBOARDS_SAMPLE_DATA_FLIGHTS_EXCEPTION_NO_TIME,
                                  FORMAT_DOT_OPENSEARCH_DASHBOARDS_SAMPLE_DATA_ECOMMERCE_EXCEPTION,
                                  FORMAT_DOT_DATE_AND_TIME,
                                  FORMAT_DOT_DATE);
                }
                
```
Because I need the try-catch to perform the parsing logic fallback from millisecond, second and eventually a normal date String.
Which I don't think that is ideal from the performance perspective, as this will run against every row being resulted.

andy-k-improving:
> @andy-k-improving It seem CI is failing on IT task. I've triggered the CI again. Please check if there is still failure.

Thanks, indeed we have a breaking change on the date field presentation on v1 engine (The old engine).
I have updated the Integration test to reflect the same, WDYT?

andy-k-improving:
I have updated `testIncorrectFormat` the test, but actual should I remove that test-case as well, because indeed this `incorrect` format is accepted by OpenSearch. 
So it's not really `incorrect` from my perspective.

andy-k-improving:
@dai-chen Would you mind to have another look on this?
Thanks,

YANG-DB (approved):
@andy-k-improving looks good !

dai-chen (approved):

PR #3193, opened 2024-12-05
Cleanup Only - Replace calls to `Arrays.asList` with too few arguments.

currantw:
### Description

Cleanup Only - Replace calls to `Arrays.asList` with too few arguments.

From IntelliJ IDEA:
image

### Related Issues

None

### Check List
- N/A - New functionality includes testing.
- N/A - New functionality has been documented.
- N/A - New functionality has javadoc added.
- N/A - New functionality has a user manual doc added.
- N/A - API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- N/A - Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -5,7 +5,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 ## Current Maintainers
 
 | Maintainer        | GitHub ID                                           | Affiliation |
-| ----------------- | -------------------------------------------------   | ----------- |
+| ----------------- |-----------------------------------------------------| ----------- |
 | Eric Wei          | [mengweieric](https://github.com/mengweieric)       | Amazon      |
 | Joshua Li         | [joshuali925](https://github.com/joshuali925)       | Amazon      |
 | Shenoy Pratik     | [ps48](https://github.com/ps48)                     | Amazon      |
@@ -16,6 +16,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 | Peter Fitzgibbons | [pjfitzgibbons](https://github.com/pjfitzgibbons)   | Amazon      |
 | Simeon Widdis     | [swiddis](https://github.com/swiddis)               | Amazon      |
 | Chen Dai          | [dai-chen](https://github.com/dai-chen)             | Amazon      |
+| Vamsi Manohar     | [vmcircuit](https://github.com/vmcircuit)        | Amazon      |
 | Peng Huo          | [penghuo](https://github.com/penghuo)               | Amazon      |
 | Sean Kao          | [seankao-az](https://github.com/seankao-az)         | Amazon      |
 | Anirudha Jadhav   | [anirudha](https://github.com/anirudha)             | Amazon      |
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -5,7 +5,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 ## Current Maintainers
 
 | Maintainer        | GitHub ID                                           | Affiliation |
-| ----------------- |-----------------------------------------------------| ----------- |
+| ----------------- | ------------------------------------------------- | ----------- |
 | Eric Wei          | [mengweieric](https://github.com/mengweieric)       | Amazon      |
 | Joshua Li         | [joshuali925](https://github.com/joshuali925)       | Amazon      |
 | Shenoy Pratik     | [ps48](https://github.com/ps48)                     | Amazon      |
@@ -16,7 +16,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 | Peter Fitzgibbons | [pjfitzgibbons](https://github.com/pjfitzgibbons)   | Amazon      |
 | Simeon Widdis     | [swiddis](https://github.com/swiddis)               | Amazon      |
 | Chen Dai          | [dai-chen](https://github.com/dai-chen)             | Amazon      |
-| Vamsi Manohar     | [vmcircuit](https://github.com/vmcircuit)        | Amazon      |
+| Vamsi Manohar     | [vmcircuit](https://github.com/vmcircuit)           | Amazon      |
 | Peng Huo          | [penghuo](https://github.com/penghuo)               | Amazon      |
 | Sean Kao          | [seankao-az](https://github.com/seankao-az)         | Amazon      |
 | Anirudha Jadhav   | [anirudha](https://github.com/anirudha)             | Amazon      |
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -5,7 +5,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 ## Current Maintainers
 
 | Maintainer        | GitHub ID                                           | Affiliation |
-| ----------------- | ------------------------------------------------- | ----------- |
+| ----------------- |-----------------------------------------------------| ----------- |
 | Eric Wei          | [mengweieric](https://github.com/mengweieric)       | Amazon      |
 | Joshua Li         | [joshuali925](https://github.com/joshuali925)       | Amazon      |
 | Shenoy Pratik     | [ps48](https://github.com/ps48)                     | Amazon      |
@@ -16,6 +16,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 | Peter Fitzgibbons | [pjfitzgibbons](https://github.com/pjfitzgibbons)   | Amazon      |
 | Simeon Widdis     | [swiddis](https://github.com/swiddis)               | Amazon      |
 | Chen Dai          | [dai-chen](https://github.com/dai-chen)             | Amazon      |
+| Vamsi Manohar     | [vamsi-amazon](https://github.com/vmmusings)        | Amazon      |
 | Peng Huo          | [penghuo](https://github.com/penghuo)               | Amazon      |
 | Sean Kao          | [seankao-az](https://github.com/seankao-az)         | Amazon      |
 | Anirudha Jadhav   | [anirudha](https://github.com/anirudha)             | Amazon      |
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -16,7 +16,6 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 | Peter Fitzgibbons | [pjfitzgibbons](https://github.com/pjfitzgibbons)   | Amazon      |
 | Simeon Widdis     | [swiddis](https://github.com/swiddis)               | Amazon      |
 | Chen Dai          | [dai-chen](https://github.com/dai-chen)             | Amazon      |
-| Vamsi Manohar     | [vmcircuit](https://github.com/vmcircuit)           | Amazon      |
 | Peng Huo          | [penghuo](https://github.com/penghuo)               | Amazon      |
 | Sean Kao          | [seankao-az](https://github.com/seankao-az)         | Amazon      |
 | Anirudha Jadhav   | [anirudha](https://github.com/anirudha)             | Amazon      |
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -16,7 +16,6 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 | Peter Fitzgibbons | [pjfitzgibbons](https://github.com/pjfitzgibbons)   | Amazon      |
 | Simeon Widdis     | [swiddis](https://github.com/swiddis)               | Amazon      |
 | Chen Dai          | [dai-chen](https://github.com/dai-chen)             | Amazon      |
-| Vamsi Manohar     | [vamsi-amazon](https://github.com/vamsi-amazon)     | Amazon      |
 | Peng Huo          | [penghuo](https://github.com/penghuo)               | Amazon      |
 | Sean Kao          | [seankao-az](https://github.com/seankao-az)         | Amazon      |
 | Anirudha Jadhav   | [anirudha](https://github.com/anirudha)             | Amazon      |
--- a/async-query/src/main/java/org/opensearch/sql/spark/cluster/ClusterManagerEventListener.java
+++ b/async-query/src/main/java/org/opensearch/sql/spark/cluster/ClusterManagerEventListener.java
@@ -5,7 +5,6 @@
 
 package org.opensearch.sql.spark.cluster;
 
-import static java.util.Collections.singletonList;
 import static org.opensearch.sql.spark.data.constants.SparkConstants.SPARK_REQUEST_BUFFER_INDEX_NAME;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -183,7 +182,7 @@ public class ClusterManagerEventListener implements LocalNodeClusterManagerListe
 
   @VisibleForTesting
   public List<Cancellable> getFlintIndexRetentionCron() {
-    return singletonList(flintIndexRetentionCron);
+    return List.of(flintIndexRetentionCron);
   }
 
   private String executorName() {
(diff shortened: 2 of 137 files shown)


acarbonetto (approved):

async-query/src/main/java/org/opensearch/sql/spark/cluster/ClusterManagerEventListener.java
```diff
@@ -183,7 +183,7 @@ private void cancel(Cancellable cron) {
 
   @VisibleForTesting
   public List<Cancellable> getFlintIndexRetentionCron() {
-    return Arrays.asList(flintIndexRetentionCron);
+    return Collections.singletonList(flintIndexRetentionCron);
```
YANG-DB, line 186: can u use static import for the `Collections.singletonList` method ?

async-query/src/test/java/org/opensearch/sql/spark/asyncquery/AsyncQueryExecutorServiceSpec.java
```diff
@@ -137,7 +136,7 @@ public class AsyncQueryExecutorServiceSpec extends OpenSearchIntegTestCase {
 
   @Override
   protected Collection<Class<? extends Plugin>> nodePlugins() {
-    return Arrays.asList(TestSettingPlugin.class);
+    return List.of(TestSettingPlugin.class);
```
YANG-DB, line 139: can u use static import for the `List.of` method ?

core/src/main/java/org/opensearch/sql/ast/expression/Argument.java
```diff
@@ -25,7 +25,7 @@ public class Argument extends UnresolvedExpression {
   //    private final DataType valueType;
   @Override
   public List<UnresolvedExpression> getChild() {
-    return Arrays.asList(value);
+    return Collections.singletonList(value);
```
YANG-DB, line 28: can u use static import for the `Collections.singletonList` method ?

YANG-DB (commented):
can u use static import for the `Collections.singletonList` & `List.of`  methods ?

async-query/src/main/java/org/opensearch/sql/spark/cluster/ClusterManagerEventListener.java
currantw: Added new commit that replaces this:

```
import java.util.Collections;
...
Collections.singletonList(...);
```

with static imports:
```
import static java.util.Collections.singletonList;
...
singletonList(...);
```

core/src/main/java/org/opensearch/sql/ast/expression/Argument.java
currantw: Yup. Updated. See comment here: https://github.com/opensearch-project/sql/pull/3193#discussion_r1872382283

async-query/src/test/java/org/opensearch/sql/spark/asyncquery/AsyncQueryExecutorServiceSpec.java
currantw: I think in this case, we need to import all of `java.util.List` because `List` is used elsewhere in the class.

There are a few cases in the code bases where we could switch to using a static imports as below. Would you want me to do so? My only concern is that `of` is pretty generic, in terms of readability?

```
import static java.util.List.of;
...
return of(TestSettingPlugin.class);
```

YANG-DB (dismissed):

opensearch/src/test/java/org/opensearch/sql/opensearch/storage/script/aggregation/dsl/MetricAggregationBuilderTest.java
```diff
@@ -5,6 +5,7 @@
 
 package org.opensearch.sql.opensearch.storage.script.aggregation.dsl;
 
+import static java.util.Collections.singletonList;
```
acarbonetto, line 8: I would recommend changing everything to List.of and remove this dependency.

core/src/test/java/org/opensearch/sql/planner/DefaultImplementorTest.java
```diff
@@ -192,9 +191,8 @@ public void visitWindowOperator_should_return_PhysicalWindowOperator() {
     NamedExpression windowFunction = named(new RowNumberFunction());
     WindowDefinition windowDefinition =
         new WindowDefinition(
-            Collections.singletonList(ref("state", STRING)),
-            Collections.singletonList(
-                ImmutablePair.of(Sort.SortOption.DEFAULT_DESC, ref("age", INTEGER))));
+            singletonList(ref("state", STRING)),
```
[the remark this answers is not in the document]
acarbonetto: I would recommend changing everything to List.of and remove this dependency.

acarbonetto (approved):

core/src/test/java/org/opensearch/sql/planner/DefaultImplementorTest.java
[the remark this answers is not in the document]
currantw: Done! Replaced `Collections.singletonList` with `List.of`. ✅

opensearch/src/test/java/org/opensearch/sql/opensearch/storage/script/aggregation/dsl/MetricAggregationBuilderTest.java
currantw: Done. Replaced all occurrences of `Collections.singletonList` with `List.of`.

core/src/main/java/org/opensearch/sql/ast/expression/Argument.java
currantw: As suggested by @acarbonetto, I've updated these to use `List.of` in cases where `List` is already imported; otherwise, uses static import of `singletonList`.

async-query/src/main/java/org/opensearch/sql/spark/cluster/ClusterManagerEventListener.java
currantw: See comment: https://github.com/opensearch-project/sql/pull/3193#discussion_r1876793204

MAINTAINERS.md
```diff
@@ -16,7 +16,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 | Peter Fitzgibbons | [pjfitzgibbons](https://github.com/pjfitzgibbons)   | Amazon      |
 | Simeon Widdis     | [swiddis](https://github.com/swiddis)               | Amazon      |
 | Chen Dai          | [dai-chen](https://github.com/dai-chen)             | Amazon      |
-| Vamsi Manohar     | [vamsi-amazon](https://github.com/vamsi-amazon)     | Amazon      |
```
currantw, line 19: This was causing the `linkchecker` pipeline job to fail...

acarbonetto: https://github.com/vmcircuit also doesn't go anywhere

acarbonetto: Maybe just remove the link?

acarbonetto (dismissed):

MAINTAINERS.md
currantw: Very weird. I swear I went to this address yesterday 🤔 
Anyway, just removed the link 🤦

acarbonetto (dismissed):

MAINTAINERS.md
YANG-DB: @currantw plz use this: https://github.com/vmmusings

YANG-DB (commented):
plz update Vamsi's github: https://github.com/vmmusings

LantaoJin:
Why do some places change `Collections.singletonList` to `singletonList`, and some places change `Collections.singletonList` to `List.of`? These changes do not fit the title.

MAINTAINERS.md
currantw: Thanks! Updated.

currantw:
> Why do some places change `Collections.singletonList` to `singletonList`, and some places change `Collections.singletonList` to `List.of`? These changes do not fit the title.

In response to review comments that suggested:
- Using `List.of` instead of `Collections.singletonList` in cases where `List` was already imported, to avoid additional imports where not needed.
- Using static import of `Collections.singletonList`, instead of importing all of `Collections`.

LantaoJin:
> > Why do some places change `Collections.singletonList` to `singletonList`, and some places change `Collections.singletonList` to `List.of`? These changes do not fit the title.
> 
> In response to review comments that suggested:
> 
> * Using `List.of` instead of `Collections.singletonList` in cases where `List` was already imported, to avoid additional imports where not needed.
> * Using static import of `Collections.singletonList`, instead of importing all of `Collections`.

This is just a one-time thing, we can't prevent the Collections API from being used again through this PR. If only for this reason, I don't recommend converting `Collections.xx` to `List.of`. It's okey to replace calls for `Arrays.asList`.

acarbonetto:
> This is just a one-time thing, we can't prevent the Collections API from being used again through this PR. If only for this reason, I don't recommend converting `Collections.xx` to `List.of`. It's okey to replace calls for `Arrays.asList`.

Removing dependencies will make the code more readable and easier to use - and I would encourage others to avoid adding unnecessary dependencies to files.  I find that the `List.of` method is more modern and versatile than the Collections apis - but they are only subtle differences between the two.  

For the sake of reducing feature changes, we can revert the `Collections.xx` to `List.of`, but I think this is a positive change.

acarbonetto (approved):

LantaoJin:
> > This is just a one-time thing, we can't prevent the Collections API from being used again through this PR. If only for this reason, I don't recommend converting `Collections.xx` to `List.of`. It's okey to replace calls for `Arrays.asList`.
> 
> Removing dependencies will make the code more readable and easier to use - and I would encourage others to avoid adding unnecessary dependencies to files. I find that the `List.of` method is more modern and versatile than the Collections apis - but they are only subtle differences between the two.
> 
> For the sake of reducing feature changes, we can revert the `Collections.xx` to `List.of`, but I think this is a positive change.

For single responsibility, let's focus on replacing calls to `Arrays` in this PR.

Create a new issue for replacing `Collections.xx` including follows items to discuss:
1. Providing similar code changes in other open source projects.
2. Any documents or articles related to replacing `Collections.xx`.
3. As long-term, a checkstyle for Java to emit warning on the usage of `Collections.xx`.

LantaoJin (changes requested):
Let's focus on cleanup calls to `Arrays.asList` in this PR.

currantw:
> Let's focus on cleanup calls to `Arrays.asList` in this PR.

Thanks @LantaoJin. I've gone through and reverted all of the changes that aren't specifically related to `Arrays.asList`. I don't think I have enough experience in Java to do a full comparison of `Collections.xx` and alternatives, nor do I really feel strongly enough about it to pursue it.

Please let me know if you have any further review comments.

YANG-DB (approved):

YANG-DB:
@currantw LGTM

legacy/src/test/java/org/opensearch/sql/legacy/antlr/semantic/types/TypeExpressionTest.java
```diff
@@ -54,20 +55,20 @@ public String getName() {
             return "Temp type expression with empty specification";
           }
         };
-    assertEquals(UNKNOWN, expr.construct(Arrays.asList(NUMBER)));
+    assertEquals(UNKNOWN, expr.construct(List.of(NUMBER)));
     assertEquals(UNKNOWN, expr.construct(Arrays.asList(STRING, BOOLEAN)));
     assertEquals(UNKNOWN, expr.construct(Arrays.asList(INTEGER, DOUBLE, GEO_POINT)));
```
LantaoJin, line 60: These two are missing.
Please search all code base again.

legacy/src/test/java/org/opensearch/sql/legacy/antlr/semantic/types/TypeExpressionTest.java
```diff
@@ -54,20 +55,20 @@ public String getName() {
...
   }
 
   @Test
   public void compatibilityCheckShouldPassIfAnySpecificationCompatible() {
-    assertEquals(DOUBLE, test123.construct(Arrays.asList(DOUBLE)));
+    assertEquals(DOUBLE, test123.construct(List.of(DOUBLE)));
     assertEquals(DATE, test123.construct(Arrays.asList(STRING, BOOLEAN)));
```
LantaoJin, line 66: ditto

LantaoJin (changes requested):
I only reviewed the product code. Looks good. But there are still some `Arrays.asList` calls in test code. Please search all code base and check them. Make sure only immutable list should be replaced.

currantw:
> I only reviewed the product code. Looks good. But there are still some `Arrays.asList` calls in test code. Please search all code base and check them. Make sure only immutable list should be replaced.

Thanks @LantaoJin. As you'll see from the description, the original intent of this cleanup PR was only to replace `Arrays.asList` with `List.of` for cases where the list has zero or one elements only, which trigger the corresponding warning (if enabled) in IntelliJ IDEA.

My understanding is that, for lists with two or fewer elements, `List.of` uses a field-based (rather than array-based) implementation, which is (slightly) more space-efficient and quicker. There are other potential advantages to like immutability, thread-safety, and not prohibiting `null` elements, though I'm not sure if any of these are relevant.

That being said, I am happy to go ahead with switching from using `Arrays.asList` to `List.of`, if you think this is desirable. I just want to make sure: do you want me to replace calls to `Arrays.asList` _only in files where there have already been changes as part of this PR_, or to do this more generally across the code base (except in cases where the list needs to be mutable or support `null` values, of course).

Please let me know how you want to proceed. Thanks!

YANG-DB:
> Actually I don't think we have any necessary reason to refactor the calls in Tests. They are tests, they are not called in product. The main difference between `Arrays.asList()` and `Collections.xx` or `List.of` are the previous API return a mutable list, but the rests return immutable list. Existing `Arrays.asList` in tests won't have memory problem. And this PR is not a long-term solution, you can't prevent new usage added in future. So If you want to complete this contribution. My suggestion is just replace the product code only with the IntelliJ IDEA's suggestion:
> 
> ```
> List<String> empty = Arrays.asList();
> List<String> one = Arrays.asList("one");
> Applied to
> List<String> empty = Collections.emptyList();
> List<String> one = Collections.singletonList("one");
> ```
@LantaoJin thanks for your feedback !
@currantw I think @LantaoJin has a good point here - we need to prioritize changes according to their expected outcome benefits - I appreciate you’ve taken time to address the IDE's suggestions 

For better clarity and planning onward - before starting the actual coding work - lets first create a detailed issue with the following details:
 - issue type (Bug / feature / enhancement ) 
 - in this case the enhancement - explain the need for the change and how it effects 
 - detail the expected work - in case of large changed please split the issue into smaller tasks

Thanks again and for everyone's work and contributions - we are making great progress !!

currantw:
> Actually I don't think we have any necessary reason to refactor the calls in Tests. They are tests, they are not called in product. The main difference between `Arrays.asList()` and `Collections.xx` or `List.of` are the previous API return a mutable list, but the rests return immutable list. Existing `Arrays.asList` in tests won't have memory problem. And this PR is not a long-term solution, you can't prevent new usage added in future. So If you want to complete this contribution. My suggestion is just replace the product code only with the IntelliJ IDEA's suggestion:

Thanks @LantaoJin and @YANG-DB. Based on the feedback, I'm just going to close this PR. Think I've dedicated enough time to this already! Will make sure to raise a dedicated issue in the future.

Issue #3185, opened 2024-12-04
[FEATURE] Improve grammar validation to verify invalid data types and identifiers


ykmr1224:
**Is your feature request related to a problem?**
In SQL grammar validation, raise error when unsupported data type or invalid identifier is specified in the query.

**What solution would you like?**
We should be able to utilize [unsupportedType](https://github.com/opensearch-project/sql/blob/main/async-query-core/src/main/antlr/SqlBaseParser.g4#L1068) under dataType / [unsupportedType](https://github.com/opensearch-project/sql/blob/main/async-query-core/src/main/antlr/SqlBaseParser.g4#L981) under type.
Invalid identifier can be captured by [errorCapturingIdentifier](https://github.com/opensearch-project/sql/blob/main/async-query-core/src/main/antlr/SqlBaseParser.g4#L1192)

**What alternatives have you considered?**
n/a

**Do you have any additional context?**
n/a

PR #3195, opened 2024-12-05
Add validation for unsupported type/identifier/commands

ykmr1224:
### Description
- Add validation for unsupported type/identifier/commands
  - Those grammars are defined but considered unsupported

### Related Issues
https://github.com/opensearch-project/sql/issues/3185

### Check List


dai-chen (approved):
Just curious if it's defined as unsupported in grammar for all data sources, why we still want them in grammar?

YANG-DB (commented):
@ykmr1224 
can we auto-generate the `SQLGrammarElement` & `PPLGrammarElement` automatically using the ANTLR files ?

ykmr1224:
> Just curious if it's defined as unsupported in grammar for all data sources, why we still want them in grammar?

Those are coming from Spark SQL grammar, which we don't want to touch considering when we upgrade Spark version.
I'm not quite sure the original intention in Spark repository, but it might be to raise error as unsupported type instead of parse error.

ykmr1224:
> @ykmr1224 can we auto-generate the `SQLGrammarElement` & `PPLGrammarElement` automatically using the ANTLR files ?

I considered that option initially, but the grammar file contains lots of elements, where same grammar element can be define as multiple way in g4 file. That could lead to very redundant configuration. Also the error message would need to use node name in g4 file, which would become very unclear for users.

YANG-DB (approved):

Issue #2477, opened 2024-01-25
[BUG] Fix DateTimeFunctionTest. testWeekOfYearWithTimeType() YearweekTest. testYearweekWithTimeType() Unit tests.


vamsimanohar:
**What is the bug?**

* ```DateTimeFunctionTest. testWeekOfYearWithTimeType() YearweekTest. testYearweekWithTimeType()``` unit tests are temporarily fixed in this PR here : https://github.com/opensearch-project/sql/pull/2476. 
* The ask from this bug is to come up with a proper fix for it.

Issue details: Jan 2024 starts with Monday and we calculate weekoftheyear using SUNDAY as the firstdayoftheweek this makes 22nd Jan 3rd week of the year rather 4th week. 

We can consider moving to LocalDateTime instead of Calendar.

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

kenrickyap:
am currently looking these test failures

acarbonetto:
Reopened due to current CI test failures: https://github.com/opensearch-project/sql/actions/runs/13124882810/job/36619055365

```
ExtractTest > testExtractDatePartWithTimeType() FAILED
    org.opentest4j.AssertionFailedError: expected: <5> but was: <6>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:635)
        at app//org.opensearch.sql.expression.datetime.ExtractTest.datePartWithTimeArgQuery(ExtractTest.java:90)
        at app//org.opensearch.sql.expression.datetime.ExtractTest.testExtractDatePartWithTimeType(ExtractTest.java:106)
```

PR #3230, opened 2025-01-03
json-valid PPL function

### LINKED ISSUE #3207: [FEATURE] Add `json_valid` as a PPL function
As part of the RFC to add JSON functions, the `json_valid` function would be useful to validate incoming json-encoded strings. 

* META RFC: #3027
* PPL RFC: #3028
* Related Spark PPL PR: https://github.com/opensearch-project/opensearch-spark/pull/780

Note: The Jackson data-bind library is already available in opensearch-sql to validate json-encoded strings. 

```
### `JSON_VALID`

**Description**

`json_valid(jsonStr)` Evaluates whether a json-encoded string contains valid JSON syntax and returns TRUE or FALSE.

**Argument type:** STRING

**Return type:** BOOLEAN

Example:

    os> source=people | eval `valid_json` = json_valid('[1,2,3,4]'), `invalid_json` = json_valid('{"invalid": "json"') | fields `valid_json`, `invalid_json`
    +--------------+----------------+
    | valid_json   | invalid_json   |
    +--------------+----------------+
    | True         | False          |
    +--------------+----------------+

    os> source=accounts | where json_valid('[1,2,3,4]') and isnull(email) | fields account_number, email
    +------------------+---------+
    | account_number   | email   |
    |------------------+---------|
    | 13               | null    |
    +------------------+---------+
```

N/A

https://github.com/opensearch-project/opensearch-spark/pull/780 - PR to add feature to opensearch-spark PPL

kenrickyap:
### Description
Based on this feature request: https://github.com/opensearch-project/sql/issues/3207

Added json_valid request.

```
### `JSON_VALID`

**Description**

`json_valid(jsonStr)` Evaluates whether a json-encoded string contains valid JSON syntax and returns TRUE or FALSE.

**Argument type:** STRING

**Return type:** BOOLEAN

Example:

    os> source=people | eval `valid_json` = json_valid('[1,2,3,4]'), `invalid_json` = json_valid('{"invalid": "json"') | fields `valid_json`, `invalid_json`
    +--------------+----------------+
    | valid_json   | invalid_json   |
    +--------------+----------------+
    | True         | False          |
    +--------------+----------------+

    os> source=accounts | where json_valid('[1,2,3,4]') and isnull(email) | fields account_number, email
    +------------------+---------+
    | account_number   | email   |
    |------------------+---------|
    | 13               | null    |
    +------------------+---------+
```

### Related Issues
Resolves #3207 

### Check List


core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -0,0 +1,46 @@
+package org.opensearch.sql.expression.json;
```
YANG-DB, line 1: missing license header ...

YANG-DB:
@kenrickyap can you please add the relevant documentation for this new function?

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
kenrickyap: addressed

kenrickyap:
> @kenrickyap can you please add the relevant documentation for this new function?

Added doctest, integ-test, and unit tests

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * @param jsonExprValue JSON string (e.g. "198.51.100.14" or "2001:0db8::ff00:42:8329").
+     * @return true if the string can be parsed as valid JSON, else false.
+     */
+    private ExprValue isValidJson(ExprValue jsonExprValue) {
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            objectMapper.readTree(jsonExprValue.stringValue());
+            return ExprValueUtils.LITERAL_TRUE;
+        } catch (Exception e) {
```
acarbonetto, line 47: don't catch everything - better to catch only what you need

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @Mock private Environment<Expression, ExprValue> env;
+
+    @Test
+    public void json_valid_invalid_json_string() {
+        assertEquals(LITERAL_FALSE, execute(JsonInvalidObject));
+        assertEquals(LITERAL_FALSE, execute(JsonInvalidScalar));
+    }
+
+    @Test
+    public void json_valid_valid_json_string() {
```
acarbonetto, line 44: funny test name, maybe: `json_valid_returns_true`?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,59 @@
...
+public class JsonFunctionsTest {
+
+    private static final ExprValue JsonObject = ExprValueUtils.stringValue("{\"a\":\"1\",\"b\":\"2\"}");
+    private static final ExprValue JsonArray = ExprValueUtils.stringValue("[1, 2, 3, 4]");
+    private static final ExprValue JsonScalarString = ExprValueUtils.stringValue("\"abc\"");
+    private static final ExprValue JsonEmptyString = ExprValueUtils.stringValue("");
+    private static final ExprValue JsonInvalidObject = ExprValueUtils.stringValue("{\"invalid\":\"json\", \"string\"}");
+    private static final ExprValue JsonInvalidScalar = ExprValueUtils.stringValue("abc");
+
+    @Mock private Environment<Expression, ExprValue> env;
```
acarbonetto, line 35: you can remove `env`

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,59 @@
...
+        assertEquals(LITERAL_TRUE, JsonArray);
+        assertEquals(LITERAL_TRUE, JsonScalarString);
+        assertEquals(LITERAL_TRUE, JsonEmptyString);
+    }
+
+    private ExprValue execute(ExprValue jsonString) {
+        final String fieldName = "json_string";
+        FunctionExpression exp = DSL.jsonValid(DSL.literal(jsonString));
+
+        when(DSL.ref(fieldName, STRING).valueOf(env)).thenReturn(jsonString);
```
acarbonetto, line 55: does it work if you remove this?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void json_valid_invalid_json_string() {
+        assertEquals(LITERAL_FALSE, execute(JsonInvalidObject));
+        assertEquals(LITERAL_FALSE, execute(JsonInvalidScalar));
+    }
+
+    @Test
+    public void json_valid_valid_json_string() {
+        assertEquals(LITERAL_TRUE, JsonObject);
```
acarbonetto, line 45: I'm not sure how this is working...
```suggestion
        assertEquals(LITERAL_TRUE, execute(JsonObject));
```

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,59 @@
...
+        assertEquals(LITERAL_TRUE, JsonEmptyString);
+    }
+
+    private ExprValue execute(ExprValue jsonString) {
+        final String fieldName = "json_string";
+        FunctionExpression exp = DSL.jsonValid(DSL.literal(jsonString));
+
+        when(DSL.ref(fieldName, STRING).valueOf(env)).thenReturn(jsonString);
+
+        return exp.valueOf(env);
```
acarbonetto, line 57:
```suggestion
        return exp.valueOf();
```

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Usage: `json_valid(json_string)` checks if `json_string` is a valid STRING string.
+
+Argument type: STRING
+
+Return type: BOOLEAN
+
+Example::
+
+    > source=json_test | where json_valid(json_string) | fields test_name, json_string
```
acarbonetto, line 25: to return a false value:  
```suggestion
    > source=json_test | eval valid=json_valid(json_string) | fields test_name, json_string, valid
```

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,34 @@
...
+    fetched rows / total rows = 4/4
+    +--------------------+--------------------+
+    | test_name          | json_string        |
+    |--------------------|--------------------|
+    | json object        | {"a":"1","b":"2"}  |
+    | json array         | [1, 2, 3, 4]       |
+    | json scalar string | [1, 2, 3, 4]       |
```
acarbonetto, line 32: update

integ-test/src/test/resources/indexDefinitions/json_test_index_mappping.json
```diff
@@ -0,0 +1,12 @@
+{
+  "mappings": {
+    "properties": {
+      "test_name": {
+        "type": "text"
```
acarbonetto, line 5: use keyword instead of text

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: addessed

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: addressed

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: yup it works :)

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: removed

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -0,0 +1,50 @@
...
+                impl(nullMissingHandling(JsonFunctions::isValidJson), BOOLEAN, STRING));
+    }
+
+    /**
+     * Checks if given JSON string can be parsed as valid JSON.
+     *
+     * @param jsonExprValue JSON string (e.g. "198.51.100.14" or "2001:0db8::ff00:42:8329").
+     * @return true if the string can be parsed as valid JSON, else false.
+     */
+    private ExprValue isValidJson(ExprValue jsonExprValue) {
```
acarbonetto, line 41: Please move this logic to `JsonUtils.java` under `org.opensearch.sql.utils` (you will need to create this class). 
Standard procedure seems to be to use a utility function for logic like this.

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
acarbonetto: Probably only need to catch `JsonProcessingException`

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
kenrickyap: addressed

kenrickyap:
pipeline is failing as disabled test are affecting code coverage. will first resolve https://github.com/opensearch-project/sql/issues/2477, then merge this ticket

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,26 @@
...
+import lombok.experimental.UtilityClass;
+import org.opensearch.sql.data.model.ExprValue;
+import org.opensearch.sql.data.model.ExprValueUtils;
+
+@UtilityClass
+public class JsonUtils {
+  /**
+   * Checks if given JSON string can be parsed as valid JSON.
+   *
+   * @param jsonExprValue JSON string (e.g. "198.51.100.14" or "2001:0db8::ff00:42:8329").
```
acarbonetto, line 14: JSON string example here is weird?  While it's a valid JSON string, I think this is a copy-paste from one of the IP functions.

kenrickyap: whoops good catch thanks :)

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,35 @@
...
+    > source=json_test | eval is_valid = json_valid(json_string) | fields test_name, json_string, is_valid
+    fetched rows / total rows = 4/4
+    +---------------------+------------------------------+----------+
+    | test_name           | json_string                  | is_valid |
+    |---------------------|------------------------------|----------|
+    | json object         | {"a":"1","b":"2"}            | True     |
+    | json array          | [1, 2, 3, 4]                 | True     |
+    | json scalar string  | "abc"                        | True     |
+    | json empty string   |                              | True     |
+    | json invalid object | {"invalid":"json", "string"} | True     |
```
acarbonetto, line 34:
```suggestion
    | json invalid object | {"invalid":"json", "string"} | False    |
```

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,35 @@
...
+
+    > source=json_test | eval is_valid = json_valid(json_string) | fields test_name, json_string, is_valid
+    fetched rows / total rows = 4/4
+    +---------------------+------------------------------+----------+
+    | test_name           | json_string                  | is_valid |
+    |---------------------|------------------------------|----------|
+    | json object         | {"a":"1","b":"2"}            | True     |
+    | json array          | [1, 2, 3, 4]                 | True     |
+    | json scalar string  | "abc"                        | True     |
+    | json empty string   |                              | True     |
```
acarbonetto, line 33: is this considered valid?

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionIT.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.opensearch.sql.util.MatcherUtils.rows;
+import static org.opensearch.sql.util.MatcherUtils.schema;
+import static org.opensearch.sql.util.MatcherUtils.verifyDataRows;
+import static org.opensearch.sql.util.MatcherUtils.verifySchema;
+
+import java.io.IOException;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+
+public class JsonFunctionIT extends PPLIntegTestCase {
```
acarbonetto, line 18: Original file includes an `s`:
```suggestion
public class JsonFunctionsIT extends PPLIntegTestCase {
```

integ-test/src/test/resources/indexDefinitions/json_test_index_mappping.json
```diff
@@ -0,0 +1,12 @@
+{
```
acarbonetto, line 1: rename to json_test_index_mapping.json
(you have 3 `p`s)

docs/user/ppl/functions/json.rst
kenrickyap: yes this returns true

acarbonetto (dismissed):

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,35 @@
+====================
+IP Address Functions
```
YANG-DB, line 2: why is the title `IP Address Functions` ?

doctest/test_data/json_test.json
```diff
@@ -0,0 +1,5 @@
+{"test_name":"json object", "json_string":"{\"a\":\"1\",\"b\":\"2\"}"}
+{"test_name":"json array", "json_string":"[1, 2, 3, 4]"}
+{"test_name":"json scalar string", "json_string":"\"abc\""}
+{"test_name":"json empty string","json_string":""}
+{"test_name":"json invalid object", "json_string":"{\"invalid\":\"json\", \"string\"}"}
```
YANG-DB, line 5: can you also add complex nested json object sample ?

kenrickyap: added

docs/user/ppl/functions/json.rst
kenrickyap: Ahh sorry, typo >.<

YANG-DB (approved):

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Test
+  public void test_json_valid() throws IOException {
+    JSONObject result;
+
+    result =
+        executeQuery(
+            String.format(
+                "source=%s | where json_valid(json_string) | fields test_name",
```
LantaoJin, line 31: what if the json_string is `null`?

acarbonetto: I'm assuming null should return `false`?

acarbonetto (dismissed):

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
YANG-DB: > I'm assuming null should return `false`?

@acarbonetto yes I think it makes sense

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
YANG-DB, line 1: missing license header

YANG-DB: actually why do we need 2 separate `JsonUtils` & `JsonFunctions` ? 
would it make sense to unify into a single class ? (inside maybe the different namespaces)

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,52 @@
...
+      ExprValueUtils.stringValue("{\"a\":\"1\",\"b\":{\"c\":\"2\",\"d\":\"3\"}}");
+  private static final ExprValue JsonObject =
+      ExprValueUtils.stringValue("{\"a\":\"1\",\"b\":\"2\"}");
+  private static final ExprValue JsonArray = ExprValueUtils.stringValue("[1, 2, 3, 4]");
+  private static final ExprValue JsonScalarString = ExprValueUtils.stringValue("\"abc\"");
+  private static final ExprValue JsonEmptyString = ExprValueUtils.stringValue("");
+  private static final ExprValue JsonInvalidObject =
+      ExprValueUtils.stringValue("{\"invalid\":\"json\", \"string\"}");
+  private static final ExprValue JsonInvalidScalar = ExprValueUtils.stringValue("abc");
+
```
andy-k-improving, line 32: Can we also have a test-case for the below input:
1. NON-String (Ex: numeric || boolean)
2. Null

The expected behaviour should be:
1. Exception due to incompatible type.
2. False

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,36 @@
...
+    +---------------------+---------------------------------+----------+
+    | test_name           | json_string                     | is_valid |
+    |---------------------|---------------------------------|----------|
+    | json nested object  | {"a":"1","b":{"c":"2","d":"3"}} | True     |
+    | json object         | {"a":"1","b":"2"}               | True     |
+    | json array          | [1, 2, 3, 4]                    | True     |
+    | json scalar string  | "abc"                           | True     |
+    | json empty string   |                                 | True     |
+    | json invalid object | {"invalid":"json", "string"}    | False    |
+    +---------------------+---------------------------------+----------+
```
andy-k-improving, line 36: How about `null`?

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
andy-k-improving: If indeed `null` return false, then we should specify this behaviour on documentation to avoid any confusion,

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
```diff
@@ -0,0 +1,55 @@
...
+
+    result =
+        executeQuery(
+            String.format(
+                "source=%s | where not json_valid(json_string) | fields test_name",
+                TEST_INDEX_JSON_TEST));
+    verifySchema(result, schema("test_name", null, "string"));
+    verifyDataRows(result, rows("json invalid object"), rows("json null"));
+  }
+}
```
andy-k-improving, line 55: Since this is integ-test, I believe we can have a test-case with non-string type, then perform assertion on the exception being thrown.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
acarbonetto: Excellent question.  I dislike "utility classes" (what is the responsibility of a utility class?).  But it gives us a central class to put all the `json` business logic.  This seems to be how we do things (date time and IP address business logic also lives in util classes).  

As for the `JsonFunction` class, it provides an integration layer between the language parser's function expressions and the json business logic.  The casting expressions will be another class that will access the json logic. 

This class could very well be named `Json` or `JsonMapper`.  But if we do this, we should rename all the util classes in the `utils` package.

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
acarbonetto: The first case is covered by the function logic. 
The second case is a good test to add.

docs/user/ppl/functions/json.rst
acarbonetto: that returns null. 

We don't need to include such error cases in the doctests.  Only happy-path scenarios.

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
acarbonetto: `null` returns `null`.

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
acarbonetto: I've added these tests to the unit tests.  I don't believe it makes much sense to add the same test here.

YANG-DB (dismissed):

acarbonetto (approved):

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,31 @@
...
+   */
+  public static ExprValue isValidJson(ExprValue jsonExprValue) {
+    ObjectMapper objectMapper = new ObjectMapper();
+
+    if (jsonExprValue.isNull() || jsonExprValue.isMissing()) {
+      return ExprValueUtils.LITERAL_FALSE;
+    }
+
+    try {
+      objectMapper.readTree(jsonExprValue.stringValue());
```
currantw, line 25: Can we extract the ObjectMapper as a static final class member?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,64 @@
...
+  private static final ExprValue JsonNestedObject =
+      ExprValueUtils.stringValue("{\"a\":\"1\",\"b\":{\"c\":\"2\",\"d\":\"3\"}}");
+  private static final ExprValue JsonObject =
+      ExprValueUtils.stringValue("{\"a\":\"1\",\"b\":\"2\"}");
+  private static final ExprValue JsonArray = ExprValueUtils.stringValue("[1, 2, 3, 4]");
+  private static final ExprValue JsonScalarString = ExprValueUtils.stringValue("\"abc\"");
+  private static final ExprValue JsonEmptyString = ExprValueUtils.stringValue("");
+  private static final ExprValue JsonInvalidObject =
+      ExprValueUtils.stringValue("{\"invalid\":\"json\", \"string\"}");
+  private static final ExprValue JsonInvalidScalar = ExprValueUtils.stringValue("abc");
```
currantw, line 35: Is there any value in having these as separate constants, rather than just inlining them within the tests? Seems like they are each only used once? If these _are_ kept as `static final` test constants, should they be remained to all caps, e.g. `JSON_NESTED_OBJECT`?

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,36 @@
...
+   :local:
+   :depth: 1
+
+JSON_VALID
+----------
+
+Description
+>>>>>>>>>>>
+
+Usage: `json_valid(json_string)` checks if `json_string` is a valid STRING string.
```
currantw, line 17:
```suggestion
Usage: `json_valid(json_string)` checks if `json_string` is a valid JSON string.
```

doctest/test_docs.py
```diff
@@ -30,6 +30,7 @@
 NESTED = "nested"
 DATASOURCES = ".ql-datasources"
 WEBLOGS = "weblogs"
+JSON_TEST = "json_test"
```
currantw, line 33: Nit. Could maybe just be named `JSON`; same with test data file? All of the other datasets (e.g. `WEBLOGS`, `NESTED`) are for testing, and this is already in `doctest/test_docs.py`, so I'm not sure is `_TEST` adds any new information?

integ-test/src/test/java/org/opensearch/sql/legacy/SQLIntegTestCase.java
```diff
@@ -745,7 +746,12 @@ public enum Index {
         TestsConstants.TEST_INDEX_GEOPOINT,
         "dates",
         getGeopointIndexMapping(),
-        "src/test/resources/geopoints.json");
+        "src/test/resources/geopoints.json"),
+    JSON_TEST(
+        TestsConstants.TEST_INDEX_JSON_TEST,
+        "json",
+        getJsonTestIndexMapping(),
+        "src/test/resources/json_test.json");
```
currantw, line 754: Nit. As per previous comment.

```suggestion
    JSON(
        TestsConstants.TEST_INDEX_JSON,
        "json",
        getJsonIndexMapping(),
        "src/test/resources/json.json");
```

currantw (commented):
A few minor comments.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Hmmm. What about the [initialization-on-demand holder idiom](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom)? Agreed that it's probably better to avoid using the memory if it's not needed, but probably also good to avoid creating the same object hundreds of times if there are hundreds of calls to any JSON function? No need to address now either way, but maybe something to consider as JSON work progresses?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
currantw: Don't mind at all! Thanks.

acarbonetto (approved):

doctest/test_docs.py
currantw: Thanks. Agreed, very minor and probably not worth the effort to change. 🤷

YANG-DB (approved):

Issue #3069, opened 2024-10-11
[BUG] PPL does not support querying Alias fields


ashwin-pc:
**What is the bug?**
When querying using PPL but there is an alias field in the filter clause. The query fails.

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

penghuo:
could u add more context? sample query and sample data to reporduce the issue.

xinyual:
I can provide a sample
```
PUT /try
{
    "mappings":{
        "properties": {
            "time1": {
                "type": "date"
            },
            "time2": {
                "type": "alias",
                "path": "time1"
            }
        }
    }
}
```

```
POST /_plugins/_ppl/
{
    "query": "source=try | fields `time2`"
}
```
It shows
```
{
  "error": {
    "reason": "Invalid Query",
    "details": "can't resolve Symbol(namespace=FIELD_NAME, name=time2) in type env",
    "type": "SemanticCheckException"
  },
  "status": 400
}
```

PR #3246, opened 2025-01-15
Support for OpenSearch alias type 

qianheng-aws:
### Description
Add Support for OpenSearch alias type, so PPL can query fields of alias type correctly. 

Implementation:
- Add a new ExprType named `OpenSearchAliasType`
- Add special parsing logic for alias type in `OpenSearchDataType::parseMapping`
- Use original path and type when constructing `ReferenceExpression`. For now, only alias type will have different original path and type.
- When constructing `_source` in `OpenSearchRequestBuilder::pushDownProjects`, use path instead of attr since they are different for alias type. The former one is more accurate since OpenSearch don't accept fields of alias type in `_source`.

e.g. After this PR:

```
# Create Index with alias type
PUT /try
{
    "mappings":{
        "properties": {
            "time1": {
                "type": "date"
            },
            "time2": {
                "type": "alias",
                "path": "time1"
            }
        }
    }
}

# Add doc
POST try/_doc
{
  "time1": 1736911491131
}
```

```
# Query with field of alias type
POST /_plugins/_ppl/
{
   "query": "source=try |where `time2` > '2025-01-01' |  fields `time2`"
}

# Get result
{
  "schema": [
    {
      "name": "time2",
      "type": "timestamp"
    }
  ],
  "datarows": [
    [
      "2025-01-15 03:24:51.131"
    ]
  ],
  "total": 1,
  "size": 1
}
```

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3069

#### Option1(The PR for now):
Add compounded datatype named OpenSearchAliasType in TypeEnvironment, and resolve ident of alias field inner ReferenceExpression. OpenSearchAliasType will record both path and original basic datatype inner it.

#### Option2:
Add a new hashmap in IndexMapping to record the alias field mapping to its referred path, it may has structure like:
Map<String, String> aliasMapping
and also add a new hashmap in SymbleTable to record the alias field, it may has structure like:
Map<Namespace, NavigableMap<String, String>> aliasByNamespace
When resolving ident, we need to lookup aliasByNamespace first to identify if this field is an alias, then lookup the original tableByNamespace to reslve type. It makes more sense to not maintaining a wired data type of alias, but will have drawback of making this analyzing process more complex than option1

#### Option3:
Parsing the data type of alias fields to be the data type of its referred field directly.
But in OpenSearchIndexScan, it should be aware of the mapping of opensearch index, and then populating ExprValue with alias fields filled with the value of their referred fields. It aims to resolve the alias fields from the source data.
That’s said, since we’ve resolved the alias fields from the soruce, they are no longer alias fields in the following processing. Assuming we have update on its referred fields after scanning, the alias field won’t be udpated as its path has no relation with the value on updated path. The functionality may not align with the original purpose of opensearch alias type in some cases.

-----
#### Option for moving to Calcite in future
In Calcite, there is a special data type named SingleColumnAliasRelDataType, which may satisfy our requirements. This RelDataType maintains single column but can resolve 2 kinds of idents with name in original or alias. But I’m afraid it cannot handle the corner cases that more than one alias field refer to the same path, which is rare but actually allowed in OpenSearch. Unless we can extend that RelDataType to remove the limitation and handle multiple fields in alias.

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/data/type/ExprType.java
+++ b/core/src/main/java/org/opensearch/sql/data/type/ExprType.java
@@ -9,6 +9,7 @@ import static org.opensearch.sql.data.type.ExprCoreType.UNKNOWN;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import org.opensearch.sql.data.model.ExprValue;
 import org.opensearch.sql.expression.Expression;
 
@@ -53,5 +54,18 @@ public interface ExprType {
   /** Get the legacy type name for old engine. */
   default String legacyTypeName() {
     return typeName();
+  }
+
+  /** Get the original path. Types like alias type will set the actual path in field property. */
+  default Optional<String> getOriginalPath() {
+    return Optional.empty();
+  }
+
+  /**
+   * Get the original path. Types like alias type should be derived from the type of the original
+   * field.
+   */
+  default ExprType getOriginalExprType() {
+    return this;
   }
 }
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchAliasType.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchAliasType.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.data.type;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import org.opensearch.sql.data.type.ExprType;
+
+/**
+ * The type of alias. See <a
+ * href="https://opensearch.org/docs/latest/opensearch/supported-field-types/alias/">doc</a>
+ */
+public class OpenSearchAliasType extends OpenSearchDataType {
+
+  public static final String typeName = "alias";
+  public static final String pathPropertyName = "path";
+  public static final Set<MappingType> objectFieldTypes =
+      Set.of(MappingType.Object, MappingType.Nested);
+  private final String path;
+  private final OpenSearchDataType originalType;
+
+  public OpenSearchAliasType(String path, OpenSearchDataType type) {
+    super(type.getExprCoreType());
+    if (type instanceof OpenSearchAliasType) {
+      throw new IllegalStateException(
+          String.format("Alias field cannot refer to the path [%s] of alias type", path));
+    } else if (objectFieldTypes.contains(type.getMappingType())) {
+      throw new IllegalStateException(
+          String.format("Alias field cannot refer to the path [%s] of object type", path));
+    }
+    this.path = path;
+    this.originalType = type;
+  }
+
+  @Override
+  public Optional<String> getOriginalPath() {
+    return Optional.of(this.path);
+  }
+
+  @Override
+  public ExprType getOriginalExprType() {
+    return originalType.getExprType();
+  }
+
+  @Override
+  public ExprType getExprType() {
+    return this;
+  }
+
+  @Override
+  public OpenSearchDataType cloneEmpty() {
+    return new OpenSearchAliasType(this.path, originalType.cloneEmpty());
+  }
+
+  @Override
+  public boolean isCompatible(ExprType other) {
+    return originalType.isCompatible(other);
+  }
+
+  @Override
+  public List<ExprType> getParent() {
+    return originalType.getParent();
+  }
+
+  @Override
+  public String typeName() {
+    return originalType.typeName();
+  }
+
+  @Override
+  public String legacyTypeName() {
+    return originalType.legacyTypeName();
+  }
+
+  @Override
+  public boolean shouldCast(ExprType other) {
+    return originalType.shouldCast(other);
+  }
+}
--- a/core/src/main/java/org/opensearch/sql/expression/ReferenceExpression.java
+++ b/core/src/main/java/org/opensearch/sql/expression/ReferenceExpression.java
@@ -23,6 +23,8 @@ import org.opensearch.sql.expression.env.Environment;
 public class ReferenceExpression implements Expression {
   @Getter private final String attr;
 
+  @Getter private final String rawPath;
+
   @Getter private final List<String> paths;
 
   private final ExprType type;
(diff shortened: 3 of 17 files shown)


YANG-DB (dismissed):
nice - thanks !

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchAliasType.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import org.opensearch.sql.data.type.ExprType;
+
+/**
+ * The type of alias. See <a
+ * href="https://opensearch.org/docs/latest/opensearch/supported-field-types/alias/">doc</a>
+ */
+public class OpenSearchAliasType extends OpenSearchDataType {
```
penghuo, line 17: Should an alias be treated as a new data type, or should it remain as just an alias for the original field? Additionally, what are the expectations for the output of the DESCRIBE TABLE command?  for example, `describe try`?

qianheng-aws: 1. I've thought of resolving alias type to  `NamedExpression` which is a resolved expression for `Alias`. In that proposal, we still need to maintain this `OpenSearchAliasType ` in `TypeEnvironment`, so as to identify `ident` of alias type, and resolve them to `NamedExpression`(i.e. Alias) with its `delegated` to be a `ReferenceExpression` to the original field.

2. With the current code, `describe try` returns:
```
{
  "schema": [{"name": "TABLE_CAT","type": "string"},{"name": "TABLE_SCHEM","type": "string"},{"name": "TABLE_NAME","type": "string"},{"name": "COLUMN_NAME","type": "string"},{"name": "DATA_TYPE","type": "string"},{"name": "TYPE_NAME","type": "string"},{"name": "COLUMN_SIZE","type": "string"},{"name": "BUFFER_LENGTH","type": "string"},{"name": "DECIMAL_DIGITS","type": "string"},{"name": "NUM_PREC_RADIX","type": "string"},{"name": "NULLABLE","type": "string"},{"name": "REMARKS","type": "string"},{"name": "COLUMN_DEF","type": "string"},{"name": "SQL_DATA_TYPE","type": "string"},{"name": "SQL_DATETIME_SUB","type": "string"},{"name": "CHAR_OCTET_LENGTH","type": "string"},{"name": "ORDINAL_POSITION","type": "string"},{"name": "IS_NULLABLE","type": "string"},{"name": "SCOPE_CATALOG","type": "string"},{"name": "SCOPE_SCHEMA","type": "string"},{"name": "SCOPE_TABLE","type": "string"},{"name": "SOURCE_DATA_TYPE","type": "string"},{"name": "IS_AUTOINCREMENT","type": "string"},{"name": "IS_GENERATEDCOLUMN","type": "string"}
  ],
"datarows": [
	[ "opensearch", null, "try", "time1", null, "timestamp", null, null, null, 10, 2, null, null, null, null, null, 0, "", null, null, null, null, "NO", ""],
	[ "opensearch", null, "try", "time2", null, "timestamp", null, null, null, 10, 2, null, null, null, null, null, 1, "", null, null, null, null, "NO", ""]
  ],
  "total": 2,
  "size": 2
}
```

penghuo: > I've thought of resolving alias type to NamedExpression which is a resolved expression for Alias
+1, agree
instead create new data type, could we add `alias` field when [parseMapping](https://github.com/opensearch-project/sql/blob/main/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java#L118)? for instance, 

```
    Set<Alias> aliases = new HashSet<>()
    indexMapping.forEach(
        (k, v) -> {

          var type = ((String) innerMap.getOrDefault("type", "object")).replace("_", "");
          if (type == "alias") {
             aliases.add(new Alias(aliasField, rawField)
           }
          // current code
         result
        });
   for(alias: aliases) {
     result.put(alias.aliasField, result.get(alias.rawField)
   }
   return result
```

LantaoJin: Adding `alias` field when parse mapping is more simply. But what if the original field definition changed? In DSL, the alias and field definitions are "static". But in PPL, the field definition can be changed by `eval <original_field> = xxx`. Should the alias change its returns correspondingly.

qianheng-aws: @penghuo it seems not work to parse `alias` field like the code, as the information of `path` cannot be recorded anywhere. For the example in the description, in analyzing process, we only know the datatype of `time2` but we still cannot resolve its value because `time2` isn't a actual exist value in the source of OpenSearch index.

penghuo: > in analyzing process, we only know the datatype of time2 but we cannot resolve its value when executing

agree. 
my concern is alias is not a data type even when move to Calcite, right? is it possible to resolve alias field as NamedExpression?

dai-chen: I was also thinking alias field will be "replaced" everywhere in the query plan after analyzing, similar as table alias?

qianheng-aws: I'm afraid we need to add alias type as well when moving to Calcite. But it's a data type only exist in analyzing process and will be resolved before execution.

>  is it possible to resolve alias field as NamedExpression?

Yes if using the way mentioned in the beginning of this thread.  Taking the above example, we can change resolving `time2` from
```
ReferenceExpression(attr = `time2`, path =`time1`, type = `date`)
```
to
```
NamedExpression(name = `time2`, delegated = ReferenceExpression(attr = `time1`, path =`time1`, type = `date`))
```

They two are nearly semantic equal I think.

penghuo: from original alias data type launch blog, https://www.elastic.co/blog/introducing-field-aliases-in-elasticsearch, this should be treated as a RENAME operation rather than introducing a new data type.

core/src/main/java/org/opensearch/sql/data/type/ExprType.java
```diff
@@ -54,4 +55,17 @@ default List<ExprType> getParent() {
   default String legacyTypeName() {
     return typeName();
   }
+
+  /** Get the original path. Types like alias type will set the actual path in field property. */
+  default Optional<String> getOriginalPath() {
+    return Optional.empty();
+  }
```
penghuo, line 62: Can we avoid modifying ExprType? The getOriginalPath approach seems inappropriate. A type should not be associated with a path.?

qianheng-aws: In Calcite analyzing, it uses a compounded data type named RelDataType, this structure is associated with both fieldName(i.e. path) as well as type.  Of course some basic types like string, int won't implement `getFieldNames` API, while other complex types like `RelRecordType`, `SingleColumnAliasRelDataType` will have fieldName inner them. But they all extends from RelDataType which has interface method `getFieldNames`.
So it may be convenient for calcite to analyze such an alias type.

For this, I think it makes sense to associate our ExprType with path, but all type except `OpenSearchAliasType` will return Option.empty.

dai-chen (approved):
Thanks for the changes!

Issue #3257, opened 2025-01-20
[ENHANCEMENT] Gradle build clean up


andy-k-improving:
**Is your feature request related to a problem?**
Throughout the codebase across different sub-module there are multiple to-do in pace for Gradle file, also certain block of code can be refactor to enhance the readability. 
Hence this is a place-holder issue to track all the enhancement being made against the Gradle build, in order to make the build script more concise.
Given the fact that this is refactor change, NO feature should be introduced or dropped.

**What solution would you like?**

- [ ] Fix todo
    - [ ] // TODO: dependency to spark should be eliminated
    - [ ] // TODO: dependency to datasources should be eliminated
    - [ ] // TODO: need to fix all java doc format
    - [ ] // TODO: fix compiler warnings
- [x] Rewrite `return provider(new Callable<RegularFile>() {` wrapper to enhance readability.  https://github.com/opensearch-project/sql/pull/3275
- [x] Centralise dep version to parent Gradle file. https://github.com/opensearch-project/sql/pull/3294
- [ ] Cache prometheus binary download  https://github.com/opensearch-project/sql/pull/3261
- [ ] Read dep versions from OpenSearch build plugin
- [ ] Gradle version catalog enablement. 

**What alternatives have you considered?**

**Do you have any additional context?**

acarbonetto:
@andy-k-improving can you add a link to the TODO that you wanted to fix?

andy-k-improving:
> [@andy-k-improving](https://github.com/andy-k-improving) can you add a link to the TODO that you wanted to fix?

Updated for the specific to-do planned to fix.

andy-k-improving:
Added two more point which aim centralise dependencies || read directly from OpenSearch build Gradle plugin.

PR #3261, opened 2025-01-23
Build: Cache Prometheus binary 

andy-k-improving:
### Description
As per the current design of Gradle test workflow, Gradle task `startPrometheus` is designed to execute upon every integration test and docTest.
However the current implementation of `startPrometheus` and `stopPrometheus` will force and trigger the re-download of `Prometheus` binary, no matter that is a first run or subsequent run.  (~100 Mb)

This result in waste of bandwidth and reduce on developer productivity, as integration test will be executed frequently during the typical development process. 

Hence this is a PR to introduce FileExistCheck to only download Prometheus binary when it's absent. 

### Test plan:
 - To execute multiple `integTest` and make sure `Prometheus File Already Exists` printed on console and no additional files being downloaded on subsequent IT test execution. 

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3257

### Check List

### DIFF
--- a/integ-test/build.gradle
+++ b/integ-test/build.gradle
@@ -284,9 +284,14 @@ task startPrometheus(type: SpawnProcessTask) {
     mustRunAfter ':doctest:doctest'
 
     doFirst {
-        download.run {
-            src getPrometheusBinaryLocation()
-            dest new File("$projectDir/bin", 'prometheus.tar.gz')
+        File downloadedPrometheus = Paths.get("$projectDir/bin", 'prometheus.tar.gz').toFile()
+        if (!downloadedPrometheus.exists()) {
+            download.run {
+                src getPrometheusBinaryLocation()
+                dest new File("$projectDir/bin", 'prometheus.tar.gz')
+            }
+        } else {
+            println "Prometheus File Already Exists"
         }
         copy {
             from tarTree("$projectDir/bin/prometheus.tar.gz")
--- a/integ-test/build.gradle
+++ b/integ-test/build.gradle
@@ -311,7 +311,6 @@ task startPrometheus(type: SpawnProcessTask) {
 task stopPrometheus(type: KillProcessTask) {
     doLast {
         file("$projectDir/bin/prometheus").deleteDir()
-        file("$projectDir/bin/prometheus.tar.gz").delete()
     }
 }
 
--- a/integ-test/build.gradle
+++ b/integ-test/build.gradle
@@ -290,8 +290,6 @@ task startPrometheus(type: SpawnProcessTask) {
                 src getPrometheusBinaryLocation()
                 dest prometheusFilePath
             }
-        } else {
-            println "Prometheus File Already Exists"
         }
         copy {
             from tarTree("$projectDir/bin/prometheus-${prometheus_binary_version}.tar.gz")
--- a/integ-test/build.gradle
+++ b/integ-test/build.gradle
@@ -284,17 +284,17 @@ task startPrometheus(type: SpawnProcessTask) {
     mustRunAfter ':doctest:doctest'
 
     doFirst {
-        File downloadedPrometheus = Paths.get("$projectDir/bin", 'prometheus.tar.gz').toFile()
-        if (!downloadedPrometheus.exists()) {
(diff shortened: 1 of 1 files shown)


integ-test/build.gradle
```diff
@@ -284,9 +284,14 @@ task startPrometheus(type: SpawnProcessTask) {
     mustRunAfter ':doctest:doctest'
 
     doFirst {
-        download.run {
-            src getPrometheusBinaryLocation()
-            dest new File("$projectDir/bin", 'prometheus.tar.gz')
+        File downloadedPrometheus = Paths.get("$projectDir/bin", 'prometheus.tar.gz').toFile()
```
acarbonetto, line 287: The intention of downloading prometheus each and every time is to make sure we have the latest version each and every time. 

We could check the timestamp of the file and if it's a day old*, re-download (then we would get the latest about once a day).  That might be acceptable.

acarbonetto (commented):
Also, we should ensure that the Prometheus download is cleared on a `clean` call - so that the developers don't need to manually delete this to clean up their state.

acarbonetto:
@dai-chen @penghuo do you remember the reasoning behind wanting to grab a new version of Prometheus each time?

penghuo:
> @dai-chen @penghuo do you remember the reasoning behind wanting to grab a new version of Prometheus each time?

I do not remember. 
It is ok to add cache latest Prometheus version for now. revisit it If we are facing issue in future.

penghuo (dismissed):
Thx!

YANG-DB (dismissed):

andy-k-improving:
Thx both, I will update this to cache the last version only.

integ-test/build.gradle
andy-k-improving: Updated to have `${prometheus_binary_version}` as part of the file name, in the case of any Prometheus bump, it will trigger an re-download.

andy-k-improving:
@YANG-DB @acarbonetto @penghuo I have updated the code to only take re-use Prometheus when it matches the version.
Can you have another look?
Thanks

integ-test/build.gradle
```diff
@@ -306,7 +311,6 @@ task startPrometheus(type: SpawnProcessTask) {
 task stopPrometheus(type: KillProcessTask) {
     doLast {
         file("$projectDir/bin/prometheus").deleteDir()
-        file("$projectDir/bin/prometheus.tar.gz").delete()
```
acarbonetto, line 309: we can delete the tar.gz if it exists:
```
"prometheus-${prometheus_binary_version}.tar.gz"
```

integ-test/build.gradle
```diff
@@ -284,12 +284,17 @@ task startPrometheus(type: SpawnProcessTask) {
...
-            src getPrometheusBinaryLocation()
-            dest new File("$projectDir/bin", 'prometheus.tar.gz')
+        File prometheusFilePath = Paths.get("$projectDir/bin", "prometheus-${prometheus_binary_version}.tar.gz").toFile()
+        if (!prometheusFilePath.exists()) {
+            download.run {
+                src getPrometheusBinaryLocation()
+                dest prometheusFilePath
+            }
+        } else {
+            println "Prometheus File Already Exists"
```
acarbonetto, line 294: not sure we need this...

integ-test/build.gradle
andy-k-improving: Hum, I'm thinking this differently, as if we keep this delete call with `prometheus-versions.gz`.
The module wouldn't be able to achieve the cache behaviour, because `prometheus-versions.gz` will always be absent during the next run of `integ-test`.

integ-test/build.gradle
andy-k-improving: Remove.

integ-test/build.gradle
acarbonetto: ...its the tar.gz that contains the version.  Right.  I was thinking we were checking for the binary.  
Problem is that old tar.gz will stick around.  
Can we make sure that all the prometheus files (tar.gz and binary) gets removed on a 'clean' call at least?  We are currently calling stopPrometheus, but we can do something extra.

acarbonetto (approved):

Swiddis:
Going to merge since this is a generally useful thing to have, speeds up startup times a lot. If we want to clear prometheus on `clean` that can be a new PR

Issue #3027, opened 2024-09-16
[META]Add PPL `JSON` extended functions support


YANG-DB:
**Is your feature request related to a problem?**
We need to enhance OpenSearch PPL (Piped Processing Language) by adding support for JSON-based operations. This extension will provide users with powerful tools to manipulate and analyze JSON data within PPL queries.

## Proposed New Functions

1. `json_object()`: Creates a JSON object from existing key/value pairs. 
```sql
   ... | eval person = json_object("name", "John", "age", 30)
```
2. `json()`: Evaluates whether a value can be parsed as JSON. Returns the value if valid, null otherwise.
```sql
   ... | eval result = json('{"name":"John","age":30}')
```
3. `json_append()`: Appends elements to the contents of a valid JSON object.
```sql
   ... | eval person = json_object("name", "John"), extended_person = json_append(person, "city", "New York")
```
4. `json_array()`: Creates a JSON array using a list of values.
```sql
   ... | eval json_list = json_array("apple", "banana", "cherry")
```
5. `json_array_to_mv()`: Maps the elements of a JSON array to a multivalued field.
```sql
   ... | eval json_list = json_array("apple", "banana", "cherry"), mv_field = json_array_to_mv(json_list)
```
6. `json_delete()`: Removes one or more keys and their corresponding values from the specified JSON object.
```sql
   ... | eval person = json_object("name", "John", "age", 30, "city", "New York"), person_no_age = json_delete(person, "age")
```
7. `json_extend()`: Extends the contents of a valid JSON object with the values of an array.
```sql
   ... | eval person = json_object("name", "John", "age", 30), extended_person = json_extend(person, json_array("city", "New York"))
```
8. `json_extract()`: Returns either a JSON array or a native type value from a field and zero or more paths.
```sql
   ... | eval data = json('{"person": {"name": "John", "age": 30}}'), name = json_extract(data, "person.name")
```
10. `json_keys()`: Returns the keys from the key-value pairs in a JSON object as a JSON array.
```sql
   ... | eval person = json_object("name", "John", "age", 30), keys = json_keys(person)
```
11. `json_set()`: Inserts or overwrites values for a JSON node with provided values and returns an updated JSON object.
```sql
   ... | eval person = json_object("name", "John"), updated_person = json_set(person, "age", 30)
```
13. `json_valid()`: Evaluates whether a JSON object uses valid JSON syntax and returns TRUE or FALSE.
```sql
[...]

PR #3243, opened 2025-01-13
PPL: Add `json` function and `cast(x as json)` function

acarbonetto:
### Description
For OpenSearch-PPL, adds the `json()` and `cast(expression as json)` functions.  Both of these functions can be used to convert a json-encoded string into an expression object.  These functions return any expression type that is valid json (object, array, or scalar).  Unfortunately, because the return type is unknown, the expression's return type is `UNDEFINED`.  

The user can use `cast` to convert the types where necessary for scalar types.  For example: 

```
// converts json-encoded integer to an int expression
cast(json(json_string) as int) 
```

The following casts are available to convert json-encoded scalars:  `int`, `long`, `double`, `float`, `boolean`. 

```
JSON
----------

Description
>>>>>>>>>>>

Usage: `json(value)` Evaluates whether a string can be parsed as a json-encoded string and casted as an expression. Returns the JSON value if valid, null otherwise.

Argument type: STRING

Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY

Example::

    > source=json_test | where json_valid(json_string) | eval json=json(json_string) | fields test_name, json_string, json
    fetched rows / total rows = 4/4
    +---------------------+------------------------------+---------------+
    | test_name           | json_string                  | json          |
    |---------------------|------------------------------|---------------|
    | json object         | {"a":"1","b":"2"}            | {a:"1",b:"2"} |
    | json array          | [1, 2, 3, 4]                 | [1,2,3,4]     |
    | json scalar string  | "abc"                        | "abc"         |
    | json empty string   |                              | null          |
    +---------------------+------------------------------+---------------+
```

### Related Issues
Resolves #3209

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
 - [x] New functionality has javadoc added.
 - [x] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3242, opened 2025-01-13
PPL: Add `json_object` command

acarbonetto:
This PR depends on the `json_valid` command: #3207 

### Description

For OpenSearch-PPL, adds the json_object() functions that takes a even number of varargs (key/value pairs). Returns an expression type that is valid json object (STRUCT).

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3208

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3262, opened 2025-01-23
PPL: Add `json_extract` function

### LINKED ISSUE #3211: [FEATURE] Add `json_extract` function to PPL

As part of the RFC to add JSON functions, the `json_extract` function would be useful to extract specific objects and scalars from an existing json_object. 

* META RFC: #3027
* PPL RFC: #3028
* Related Spark PPL PR: https://github.com/opensearch-project/opensearch-spark/pull/780

```
### `JSON_EXTRACT`

**Description**

`json_extract(doc, path [, path]...)` Extracts a json value or scalar from a json document based on the path(s) specified.

**Argument type:** STRING, STRING

**Return type:** STRING | NULL | JSON ARRAY

- Returns a JSON array for multiple paths or if the path leads to an array.  
- Return null if path is not valid.

Example:

    os> source=people | eval `json_extract(json('{"a":"b"}'), '$.a')` = json_extract('{"a":"b"}', '$a')
    +----------------------------------+
    | json_extract('{"a":"b"}', 'a')   |
    +----------------------------------+
    | b                                |
    +----------------------------------+

    os> source=people | eval `json_extract(json('{"a":[{"b":1},{"b":2}]}'), '$.a[1].b')` = json_extract('{"a":[{"b":1},{"b":2}]}', '$.a[1].b')
    +-----------------------------------------------------------+
    | json_extract('{"a":[{"b":1.0},{"b":2.0}]}', '$.a[1].b')   |
    +-----------------------------------------------------------+
    | 2.0                                                       |
    +-----------------------------------------------------------+

    os> source=people | eval `json_extract('{"a":[{"b":1},{"b":2}]}', '$.a[*].b')` = json_extract(json('{"a":[{"b":1},{"b":2}]}'), '$.a[*].b')
    +-----------------------------------------------------------+
    | json_extract('{"a":[{"b":1.0},{"b":2.0}]}', '$.a[*].b')   |
    +-----------------------------------------------------------+
    | [1.0,2.0]                                                 |
    +-----------------------------------------------------------+

    os> source=people | eval `invalid_json` = json_extract(json('{"invalid": "json"'))
    +----------------+
    | invalid_json   |
    +----------------+
    | null           |
    +----------------+
```

N/A

Related spark PPL PR: https://github.com/opensearch-project/opensearch-spark/pull/780

kenrickyap:
### Description
Implements json_extract function that would be useful to extract specific objects and scalars from an existing json_object.

### Related Issues
Resolves #3211 
Follow up issue for https://github.com/opensearch-project/sql/issues/3027

### Check List


kenrickyap:
would be ideal to merge: 
- https://github.com/opensearch-project/sql/pull/3242
- https://github.com/opensearch-project/sql/pull/3243
before merging this PR

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +44,99 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+  public static ExprValue castJson(ExprValue json) {
+    JsonNode jsonNode = jsonToNode(json);
+    return processJsonNode(jsonNode);
+  }
+
+  public static ExprValue extractJson(ExprValue json, ExprValue path) {
+    String jsonString = json.stringValue();
+    String jsonPath = path.stringValue();
+
+    if (jsonString.equals("")) {
```
normanj-bitquill, line 63: Do you need to handle the case when `json` has a null value?

kenrickyap: handled in array test case but will add more explicit test case

currantw, line 63: Do we need to handle this separately? Seems like a sub-case of `PathNotFoundException`? Might be value in earlier exit though 🤷. If so, not sure if it's better practice in Java to use `String.isEmpty`? A touch more explicit, at least

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +44,99 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+      return LITERAL_NULL;
+    }
+
+    try {
+      Configuration config = Configuration.builder().options(Option.AS_PATH_LIST).build();
```
currantw, line 68: Can this be a static configuration that is re-used every time? (If it's needed - see my other comment).

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +44,99 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+    }
+  }
+
+  /** Converts a JSON encoded string to an Expression object. */
+  public static ExprValue castJson(ExprValue json) {
+    JsonNode jsonNode = jsonToNode(json);
+    return processJsonNode(jsonNode);
+  }
+
+  public static ExprValue extractJson(ExprValue json, ExprValue path) {
```
currantw, line 59: Are you planning to update this later to handle multiple paths as part of a separate PR, or in this one?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Would it make sense to just use `nullMissingHandling` in `JsonFunctions`? Returning null for null, and missing for missing, seems reasonable in this case?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +44,99 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+    String jsonString = json.stringValue();
+    String jsonPath = path.stringValue();
+
+    if (jsonString.equals("")) {
+      return LITERAL_NULL;
+    }
+
+    try {
+      Configuration config = Configuration.builder().options(Option.AS_PATH_LIST).build();
+      List<String> resultPaths = JsonPath.using(config).parse(jsonString).read(jsonPath);
```
currantw, line 69: Why do you need to get the paths of the evaluation hits (using `Option.AS_PATH_LIST`), and _then_ get their values after? Could you not just get the values right away?

```java
List<Object> results = JsonPath.parse(jsonString).read(jsonPath);
List<ExprValue> elements = objects.stream().map(ExprValueUtils::fromObjectValue).toList();
```

More generally, could we split this into multiple steps for clarity?

```java
```java
JsonPath jsonPath = JsonPath.compile(path.stringValue());
ReadContext ctx = JsonPath.parse(json.stringValue());
List<Object> results = ctx.read(jsonPath);
...
```

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +44,99 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+
+      List<ExprValue> elements = new LinkedList<>();
+      for (String resultPath : resultPaths) {
+        Object result = JsonPath.parse(jsonString).read(resultPath);
+        elements.add(ExprValueUtils.fromObjectValue(result));
+      }
+
+      if (elements.size() == 1) {
+        return elements.get(0);
```
currantw, line 78: Nit. Use `getFirst()`?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: this should not be the case for jsonValid ad json-valid shoudl return false if null or missing

core/src/main/java/org/opensearch/sql/expression/DSL.java
```diff
@@ -973,6 +981,10 @@ public static FunctionExpression utc_timestamp(
     return compile(functionProperties, BuiltinFunctionName.UTC_TIMESTAMP, args);
   }
 
+  public static FunctionExpression json_function(Expression value) {
+    return compile(FunctionProperties.None, BuiltinFunctionName.JSON, value);
+  }
```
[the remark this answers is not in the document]
kenrickyap: json_function and jsonCast are both forked from https://github.com/opensearch-project/sql/pull/3243

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -20,10 +22,24 @@
...
   private DefaultFunctionResolver jsonValid() {
     return define(
         BuiltinFunctionName.JSON_VALID.getName(), impl(JsonUtils::isValidJson, BOOLEAN, STRING));
   }
+
+  private DefaultFunctionResolver jsonFunction() {
+    return define(
+        BuiltinFunctionName.JSON.getName(),
+        impl(nullMissingHandling(JsonUtils::castJson), UNDEFINED, STRING));
+  }
```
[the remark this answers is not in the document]
kenrickyap: will address when https://github.com/opensearch-project/sql/pull/3243 is merged as this is an artifact of the fork

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: thanks for catching this, will implement this for this pr!

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: Will do, thanks for the heads up will switch to String.isEmpty

kenrickyap: added nullMissingHandling to json_extract

core/src/main/java/org/opensearch/sql/expression/DSL.java
```diff
@@ -687,6 +687,10 @@ public static FunctionExpression jsonValid(Expression... expressions) {
     return compile(FunctionProperties.None, BuiltinFunctionName.JSON_VALID, expressions);
   }
 
+  public static FunctionExpression jsonExtract(Expression... expressions) {
+    return compile(FunctionProperties.None, BuiltinFunctionName.JSON_EXTRACT, expressions);
+  }
```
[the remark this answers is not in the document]
kenrickyap: will probably stick with current pattern since most of the project uses camel case for functions

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: addressed

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: addressed

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: Addressed

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: resolved

core/src/main/java/org/opensearch/sql/expression/DSL.java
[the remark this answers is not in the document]
currantw: Sounds good. Would be nice to have a linter verify this so we could have some consistency. Oh well...

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Great, thank you!

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -61,4 +76,206 @@ private ExprValue execute(ExprValue jsonString) {
...
   }
+
+  @Test
+  void json_returnsJsonObject() {
+    FunctionExpression exp;
+
+    // Setup
+    final String objectJson =
+        "{\"foo\": \"foo\", \"fuzz\": true, \"bar\": 1234, \"bar2\": 12.34, \"baz\": null, "
+            + "\"obj\": {\"internal\": \"value\"}, \"arr\": [\"string\", true, null]}";
```
currantw, line 87: Nit.

Since we already have a multi-line string, maybe we could format it so it's a bit easier to read (all the escapes already make it tough enough!) Also, what about using field names like `"string_field"` to make it a bit more self-documenting?

🤷 

```suggestion
    final String objectJson =
        "{\"string_field\": \"foo\","
          + "\"bool_field\": true,"
          + "\"int_field\": 1234,"
          + "\"float_field\": 12.34,"
          + "\"null_field\": null, "
          + "\"obj_field\": {\"internal\": \"value\"}, \"arr\": [\"string\", true, null]}";
```

currantw: I guess this is just forked from the other branch. Oh well. Feel free to ignore.

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -61,4 +76,206 @@ private ExprValue execute(ExprValue jsonString) {
...
+                List.of(
+                    new ExprIntegerValue(1), new ExprIntegerValue(2), new ExprIntegerValue(3))));
+
+    // extract specific index from JSON list
+    for (int i = 0; i < expectedExprValue.size(); i++) {
+      String path = String.format("$.a[%d]", i);
+      Expression pathExpr = DSL.literal(ExprValueUtils.stringValue(path));
+      FunctionExpression expression = DSL.jsonExtract(jsonArray, pathExpr);
+      assertEquals(expectedExprValue.get(i), expression.valueOf());
+    }
```
currantw, line 192: What about an invalid index? Is that something we want to test? (Or is this covered by existing tests/something we are delegating to `JsonPath`?)

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -61,4 +76,206 @@ private ExprValue execute(ExprValue jsonString) {
...
+    // missing bracket
+    assertThrows(SemanticCheckException.class, () -> DSL.castJson(DSL.literal("{{[}}")).valueOf());
+
+    // missing quote
+    assertThrows(
+        SemanticCheckException.class, () -> DSL.castJson(DSL.literal("\"missing quote")).valueOf());
+  }
+
+  @Test
+  void json_extract_search_arrays() {
```
currantw, line 170: Seems like we are only testing values within an array right now?

Some other cases we might want:
- Basic top-level path like `$.a`? 
- Nesting (e.g. array value inside object, object value inside array, object value inside object, array value inside array).

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,88 @@ Return type: BOOLEAN
...
+    | json empty string   |                                     | null                        |
+    +---------------------+-------------------------------------+-----------------------------+
+
+JSON_EXTRACT
+____________
+
+Description
+>>>>>>>>>>>
+
+Usage: `json_extract(doc, path)` Extracts a json value or scalar from a json document based on the path(s) specified.
```
currantw, line 72: *path (singular), right?

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,88 @@ Return type: BOOLEAN
...
+
+Argument type: STRING, STRING
+
+Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
```
currantw, line 76: Does it return `MISSING` if the JSON string is missing?

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,88 @@ Return type: BOOLEAN
...
+
+- Returns a JSON array for multiple paths or if the path leads to an array.
+- Return null if path is not valid.
```
currantw, line 79: Do we want to specify here how a NULL or MISSING JSON string is handled?

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,88 @@ Return type: BOOLEAN
...
+    +---------------------+-------------------------------------+-------------------+
+    | test_name           | json_string                         | json_extract      |
+    |---------------------|-------------------------------------|-------------------|
+    | json nested object  | {"a":"1","b":{"c":"2","d":"3"}}     | {c:"2",d:"3"}     |
+    | json nested list    | {"a":"1","b":[{"c":"2"},{"c":"3"}]} | [{c:"2"},{c:"3"}] |
+    | json object         | {"a":"1","b":"2"}                   | 2                 |
+    | json array          | [1, 2, 3, 4]                        | null              |
+    | json scalar string  | "abc"                               | null              |
+    | json empty string   |                                     | null              |
+    +---------------------+-------------------------------------+-------------------+
```
currantw, line 94: Should we have a test here for `null` and `missing` values? Or, as suggested in comment above, perhaps just a comment documenting the behaviour in those cases?

currantw (commented):
Reviewed to `f030057`

core/src/main/java/org/opensearch/sql/expression/DSL.java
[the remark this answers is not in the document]
kenrickyap: yeah that would be nice

docs/user/ppl/functions/json.rst
kenrickyap: It will throw an error

docs/user/ppl/functions/json.rst
kenrickyap: thanks fixed :)

docs/user/ppl/functions/json.rst
kenrickyap: Documentation only needs to document common uses for function but will state that error will be thrown if missing or null

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: added and basic and nested case

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: Oh good catch :) added unit test

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -61,4 +76,239 @@ private ExprValue execute(ExprValue jsonString) {
...
+        SemanticCheckException.class, () -> DSL.castJson(DSL.literal("\"missing quote")).valueOf());
+  }
+
+  @Test
+  void json_extract_search() {
+    Expression jsonArray = DSL.literal(ExprValueUtils.stringValue("{\"a\":1}"));
+    ExprValue expectedExprValue = new ExprIntegerValue(1);
+    Expression pathExpr = DSL.literal(ExprValueUtils.stringValue("$.a"));
+    FunctionExpression expression = DSL.jsonExtract(jsonArray, pathExpr);
+    assertEquals(expectedExprValue, expression.valueOf());
```
currantw, line 175: Nits.

Seems pretty inconsistent throughout this file, most of which you didn't write ... but might make sense to be consistent in how we create expression values (here _and_ and the other `json_extract_*` test methods)? Right now, for example, we are using _both_ `ExprValueUtils.stringValue()` and `new ExprStringValue()`. Would it make sense to just use one of these? To my eyes, just using `ExprValueUtils` is probably preferable because it would avoid us having to import every `Expr*Value` class (which is the main point of this utility class, I think).

Also, to make this more readable, might make sense to separate expected and actual cases, and name them consistently?

Lastly, is this a `jsonArray`? Perhaps `jsonExpr` could be more accurate? Or `jsonStringExpr`? 🤷 

```suggestion
    ExprValue expectedExprValue = ExprValueUtils.integerValue(1);
    
    Expression jsonStringExpr = DSL.literal(ExprValueUtils.stringValue("{\"a\":1}"));
    Expression pathStringExpr = DSL.literal(ExprValueUtils.stringValue("$.a"));
    ExprValue actualExprValue = DSL.jsonExtract(jsonStringExpr, pathStringExpr).valueOf();
    
    assertEquals(expectedExprValue, actualExprValue);
```

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -61,4 +76,239 @@ private ExprValue execute(ExprValue jsonString) {
...
+    // index out of bounds
+    assertThrows(
+        SemanticCheckException.class,
+        () -> DSL.jsonExtract(jsonArray, DSL.literal(new ExprStringValue("$.a[3]"))).valueOf());
+
+    // negative index
+    assertThrows(
+        SemanticCheckException.class,
+        () -> DSL.jsonExtract(jsonArray, DSL.literal(new ExprStringValue("$.a[-1]"))).valueOf());
+  }
```
currantw, line 191: Some more nits.
* Is the `jsonArray` string malformed..?
* Would `jsonStringExpr` be more accurate (see previous comment)?

A more important questions:
* Is this exception actually being raised because the index is out of bounds. Based on my reading of the implementation and my assumption about how `JsonPath` works, I assume that these would throw a `InvalidPathException`, and so would return `null`. Could you be getting a `SemanticCheckException` because of the malformed JSON string?
* Moreover, since there are two `SemanticCheckExceptions`, does it make sense to check the error message (using wildcard character `*` for error details that come from the `JsonPath` exception) so that we know that the "right" `SemanticCheckException` is being raised?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
currantw: One more thought: would it make sense to extract a private method like `test_json_extract` that you could use to simplify these tests?

```java
private static void `test_json_extract(String jsonString, String pathString, ExprValue expected) {
    Expression jsonStringExpr = DSL.literal(ExprValueUtils.stringValue(jsonString));
    Expression pathStringExpr = DSL.literal(ExprValueUtils.stringValue(pathString));
    ExprValue actualExprValue = DSL.jsonExtract(jsonStringExpr, pathStringExpr).valueOf();

   assertEquals(expectedExprValue, actualExprValue);
}
```

This could simplify these test methods a lot perhaps? 🤷

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -61,4 +76,239 @@ private ExprValue execute(ExprValue jsonString) {
...
+                .valueOf());
+
+    // invalid json
+    assertThrows(
+        SemanticCheckException.class,
+        () ->
+            DSL.jsonExtract(
+                    DSL.literal(new ExprStringValue("{\"invalid\":\"json\", \"string\"}")),
+                    DSL.literal(new ExprStringValue("$.a")))
+                .valueOf());
```
currantw, line 279: As suggested in another comment, should we check the error message here to make sure we are getting the "correct" exception in both cases? Plus, it tests the error message that _is_ oart of your implementation 🤷

docs/user/ppl/functions/json.rst
currantw: It makes sense to me to throw an error if the `path` is `null` or `missing`, since this doesn't seem like it would be an expected use case. I'm a bit less sure about `doc`.

Is there a case where we might want to use `json_extract` on a JSON string field in a series of records, where it might potentially be `null` or `missing` for some records? In this case, I assume we wouldn't want an error. (We could, of course, add a `where` clause before to eliminate these `null` or `missing` records.)

What do other functions do? I know that the `cidr_match` function that I wrote returns `null`/`missing` if the IP address _or_ the CIDR range is `null`. Don't know if it _should_ do that, though (at least with respect to a `null` or `missing` CIDR range).

@acarbonetto any thoughts?

currantw (commented):
Reviewed to `2b08007`.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +42,94 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+      return ExprValueUtils.fromObjectValue(results);
+    } catch (PathNotFoundException e) {
+      return LITERAL_NULL;
+    } catch (InvalidPathException e) {
+      final String errorFormat = "JSON path '%s' is not valid. Error details: %s";
+      throw new SemanticCheckException(String.format(errorFormat, path, e.getMessage()), e);
+    } catch (InvalidJsonException e) {
+      final String errorFormat = "JSON string '%s' is not valid. Error details: %s";
+      throw new SemanticCheckException(String.format(errorFormat, json, e.getMessage()), e);
     }
```
currantw: I'm not really sure what exception is preferred here... But on my previous PR, I was advised to use `InvalidArgumentException`. Perhaps that is more appropriate here?

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,90 @@ Return type: BOOLEAN
...
+    | json empty string   |                                     | null                        |
+    +---------------------+-------------------------------------+-----------------------------+
+
+JSON_EXTRACT
+____________
+
+Description
+>>>>>>>>>>>
+
+Usage: `json_extract(doc, path)` Extracts a json value or scalar from a json document based on the path specified.
```
currantw, line 72: The documentation for the `json` function about says it returns a "JSON value", which seems to include scalars, arrays, and structs; here, we are using "json value" and "scalar" as separate possible return "types". We should probably be consistent between the two (and any future JSON functions as well)?

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,90 @@ Return type: BOOLEAN
...
+
+Argument type: STRING, STRING
+
+Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
+
+- Returns a JSON array for multiple paths or if the path leads to an array.
```
currantw, line 78: Is there a better way to say this than "multiple paths"? My understanding is that it's not "multiple paths", but rather a single path that points to multiple value (semantics, I know 😆 ). What about something like "Returns a JSON array if the given path corresponds to multiple value, or if it points to a JSON array."

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: sorry this should return null, good catch

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: I am following Andrew's implementation of json and json_cast where malformed json throws SematicCheckError

docs/user/ppl/functions/json.rst
kenrickyap: updated description

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: added helper

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: added message check

docs/user/ppl/functions/json.rst
kenrickyap: added new comment

currantw (commented):
Reviewed to `e57fa21`.

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,90 @@ Return type: BOOLEAN
...
+>>>>>>>>>>>
+
+Usage: `json_extract(doc, path)` Extracts a JSON value from a json document based on the path specified.
+
+Argument type: STRING, STRING
+
+Return type: STRING/BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
+
+- Returns a JSON array if path points to multiple results (e.g. $.a[*]) or if the path points to an array.
+- Return null if path is not valid is MISSING or NULL.
```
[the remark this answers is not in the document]
kenrickyap: fixed

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
currantw: Thanks. Your implementation with `startsWith` is better than what I suggested. 😆

currantw (approved):
Reviewed to `dfe5d3a`. Looks good to me.

currantw (approved):
Reviewed to `dfe5d3a`. Looks good to me.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
YANG-DB, line 1: can u plz add license header ?
thanks

YANG-DB (dismissed):

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: added

YANG-DB (dismissed):

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +47,101 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+    String jsonPath = path.stringValue();
+
+    if (jsonString.isEmpty()) {
+      return LITERAL_NULL;
+    }
+
+    try {
+      Object results = JsonPath.parse(jsonString).read(jsonPath);
+      return ExprValueUtils.fromObjectValue(results);
+    } catch (PathNotFoundException e) {
```
acarbonetto, line 106:
```suggestion
    } catch (PathNotFoundException ignored) {
```

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +47,101 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+      return LITERAL_NULL;
+    } catch (InvalidPathException e) {
```
acarbonetto, line 108: please don't reuse variable names:
```suggestion
    } catch (InvalidPathException invalidPathException) {
```

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +47,101 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+      final String errorFormat = "JSON path '%s' is not valid. Error details: %s";
+      throw new SemanticCheckException(String.format(errorFormat, path, e.getMessage()), e);
+    } catch (InvalidJsonException e) {
```
acarbonetto, line 111: please don't reuse variable names:
```suggestion
    } catch (InvalidJsonException invalidJsonException) {
```

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -7,58 +7,320 @@
...
+
+    // missing quote
+    assertThrows(
+        SemanticCheckException.class, () -> DSL.castJson(DSL.literal("\"missing quote")).valueOf());
+  }
+
+  @Test
+  void json_extract_search() {
+    ExprValue expected = new ExprIntegerValue(1);
+    execute_extract_json(expected, "{\"a\":1}", "$.a");
```
acarbonetto, line 197: let's add a test for each scalar type: boolean, double, string, integer (already have) and null (not exactly the same case as the test case below).

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -7,58 +7,320 @@
...
+  }
+
+  @Test
+  void json_extract_search_arrays_out_of_bound() {
+    execute_extract_json(LITERAL_NULL, "{\"a\":[1,2,3]}", "$.a[4]");
```
acarbonetto, line 202: move to `json_extract_returns_null

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -7,58 +7,320 @@
...
+            "true",
+            "false",
+            "");
+
+    jsonStrings.forEach(str -> execute_extract_json(LITERAL_NULL, str, "$.a.path_not_found_key"));
+
+    // null json
+    assertEquals(
+        LITERAL_NULL,
+        DSL.jsonExtract(DSL.literal(LITERAL_NULL), DSL.literal(new ExprStringValue("$.a")))
```
acarbonetto, line 258:
```suggestion
        DSL.jsonExtract(DSL.literal(LITERAL_NULL), DSL.literal("$.a"))
```

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -7,58 +7,320 @@
...
+            .valueOf());
+
+    // missing json
+    assertEquals(
+        LITERAL_MISSING,
+        DSL.jsonExtract(DSL.literal(LITERAL_MISSING), DSL.literal(new ExprStringValue("$.a")))
```
acarbonetto, line 264:
```suggestion
        DSL.jsonExtract(DSL.literal(LITERAL_MISSING), DSL.literal("$.a"))
```

acarbonetto: same for rest of file...

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -20,10 +22,24 @@
...
+
+  private DefaultFunctionResolver jsonExtract() {
+    return define(
+        BuiltinFunctionName.JSON_EXTRACT.getName(),
+        impl(JsonUtils::extractJson, UNDEFINED, STRING, STRING));
```
andy-k-improving, line 43: Any reason `nullMissingHandling( )` being skipped here?
By having it, you don't have to deal with `null` or missing case at all.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -23,9 +47,101 @@ public static ExprValue isValidJson(ExprValue jsonExprValue) {
...
+
+  /**
+   * Extract value of JSON string at given JSON path.
+   *
+   * @param json JSON string (e.g. "{\"hello\": \"world\"}").
+   * @param path JSON path (e.g. "$.hello")
+   * @return ExprValue of value at given path of json string.
+   */
+  public static ExprValue extractJson(ExprValue json, ExprValue path) {
+    if (json == LITERAL_NULL || json == LITERAL_MISSING) {
```
andy-k-improving, line 92: What's the rational to have this custom null || missing handle here, instead of relying on `nullMissingHandling( )`, as I don't any specific particular `null` handling logic here, compare to what `nullMissingHandling` provided.

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,90 @@ Return type: BOOLEAN
...
+Description
+>>>>>>>>>>>
+
+Usage: `json_extract(doc, path)` Extracts a JSON value from a json document based on the path specified.
+
+Argument type: STRING, STRING
+
+Return type: STRING/BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
+
+- Returns a JSON array if `path` points to multiple results (e.g. $.a[*]) or if the `path` points to an array.
```
andy-k-improving, line 78: JSON array is not a valid type in the context of `OpenSearchSQL`, in the case of the multiple matches, the json library will return a list, which then being translated to `ExprCollectionValue` via `return ExprValueUtils.fromObjectValue(results);` on your implementation, it's probably either STRUCT || ARRAY, but I'm not sure.

docs/user/ppl/functions/json.rst
```diff
@@ -23,14 +23,90 @@ Return type: BOOLEAN
...
+- Return null if `path` is not valid, or if JSON `doc` is MISSING or NULL.
+- Throws SemanticCheckException if `doc` or `path` is malformed.
+- Throws ExpressionEvaluationException if `path` is missing.
+
+Example::
+
+    > source=json_test | where json_valid(json_string) | eval json_extract=json_extract(json_string, '$.b') | fields test_name, json_string, json_extract
```
andy-k-improving, line 85: You will need `os>` prefix to make execute a part of docsTest, or else it will be skipped, see `expressions.rst#49` for usage reference.

andy-k-improving: Same for the other examples within the file.

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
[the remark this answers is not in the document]
kenrickyap: yes we only take single paths similar to spark

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: nullMissingHandling throws an error if input param is null or missing, but we want to return the same value

docs/user/ppl/functions/json.rst
kenrickyap: for lists it will return an ARRAY

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: when do we want to use one other the other?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: should we be consistent with the rest of the repo since it seems catch use e in other cases

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: Right, that make sense.

docs/user/ppl/functions/json.rst
andy-k-improving: Thanks!

docs/user/ppl/functions/json.rst
kenrickyap: added os

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: added

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: changed

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: addressed

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: addressed

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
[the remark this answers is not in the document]
kenrickyap: updated to take multiple paths

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,72 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+
+Argument type: STRING, STRING
+
+Return type: STRING/BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
+
+- Up to 3 paths can be provided, and results of each `path` with be returned in an ARRAY.
+- Returns an ARRAY if `path` points to multiple results (e.g. $.a[*]) or if the `path` points to an array.
+- Return null if `path` is not valid, or if JSON `doc` is MISSING or NULL.
+- Throws SemanticCheckException if `doc` or `path` is malformed.
+- Throws ExpressionEvaluationException if `path` is missing.
```
currantw, line 82: Looks like this documentation might need some updates for multiple paths?
- Usage: `json_extract(doc, path[, path[, path[[)` (or another way to indicate that up to 3 paths can be provided)
- Usage: "based on the path(s) specified"
- Argument type: "STRING, STRING[, STRING[, STRING]] (or another way to indicate that up to 3 paths can be provided)
- What about the case where only one of the paths is invalid? Should this read "Returns null if _any_ `path` is not valid"? Similarly, should the last two bullets be updated to specify that they apply to "any `path`"?

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
```diff
@@ -183,4 +190,31 @@ public void test_cast_json_scalar_to_type() throws IOException {
...
+        rows("json nested array", null),
+        rows("json scalar string", null),
+        rows("json scalar int", null),
+        rows("json scalar float", null),
+        rows("json scalar double", null),
+        rows("json scalar boolean true", null),
+        rows("json scalar boolean false", null),
+        rows("json empty string", null),
+        rows("json nested list", new JSONArray(List.of(Map.of("c", "2"), Map.of("c", "3")))));
+  }
```
currantw, line 219: As discussed during yesterday's review session, we certainly don't need to test all edge cases in IT, but I think we probably want to have more than one test. At least, we should probably have one with two path strings and one with three path strings?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -79,6 +85,58 @@ public static ExprValue castJson(ExprValue json) {
...
+   *
+   * @param json JSON string (e.g. "{\"hello\": \"world\"}").
+   * @param paths list of JSON path (e.g. "$.hello")
+   * @return ExprValue of value at given path of json string.
+   */
+  public static ExprValue extractJson(ExprValue json, ExprValue... paths) {
+    List<ExprValue> resultList = new ArrayList<>();
+
+    for (ExprValue path : paths) {
+      System.out.println("Processing path: " + path);
```
currantw, line 99: Looks like an errant debug statement?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -79,6 +85,58 @@ public static ExprValue castJson(ExprValue json) {
...
+      if (json == LITERAL_NULL || json == LITERAL_MISSING) {
+        return json;
+      }
```
currantw, line 102: Probably don't want to use `==` for comparison (my understanding is that this will compare the memory locations of the objects, and so depends on `ExprNullValue` and `ExprMissingValue` being implemented as singletons.

```suggestion
      if (json.isNull() || json.isMissing()) {
        return json;
      }
```

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -79,6 +85,58 @@ public static ExprValue castJson(ExprValue json) {
...
 
+  /**
+   * Extract value of JSON string at given JSON path.
+   *
+   * @param json JSON string (e.g. "{\"hello\": \"world\"}").
+   * @param paths list of JSON path (e.g. "$.hello")
+   * @return ExprValue of value at given path of json string.
+   */
+  public static ExprValue extractJson(ExprValue json, ExprValue... paths) {
+    List<ExprValue> resultList = new ArrayList<>();
```
currantw, line 96: Since know that the array list is always going to be the same since as `paths`, can we just initialize it with a specified `initialCapacity` to save some memory space?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -79,6 +85,58 @@ public static ExprValue castJson(ExprValue json) {
...
+
+    if (resultList.size() == 1) {
+      return resultList.getFirst();
+    } else {
+      return new ExprCollectionValue(resultList);
+    }
+  }
+
+  private static ExprValue extractJsonPath(String json, String path) {
+    if (json.isEmpty() || json.equals("null")) {
```
currantw, line 118: Do we need a separate case for either of these? Seems like this would just throw a `PathNotFoundException` and returns `LITERAL_NULL` below if not (at least for "null", I'm not sure about the empty string). Probably makes sense to reduce branching here if possible. 🤷 

Moreover, do we actually want to return `LITERAL_NULL` for the empty string? Is that an invalid JSON string? If so, does it make sense to raise the corresponding exception instead?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -216,5 +218,153 @@ void json_returnsSemanticCheckException() {
...
+            List.of(new ExprStringValue("foo"), new ExprFloatValue(12.34), LITERAL_NULL));
+    Expression pathExpr1 = DSL.literal(ExprValueUtils.stringValue("$.foo"));
+    Expression pathExpr2 = DSL.literal(ExprValueUtils.stringValue("$.bar2"));
+    Expression pathExpr3 = DSL.literal(ExprValueUtils.stringValue("$.potato"));
+    Expression jsonExpr = DSL.literal(ExprValueUtils.stringValue(objectJson));
+    ExprValue actual = DSL.jsonExtract(jsonExpr, pathExpr1, pathExpr2, pathExpr3).valueOf();
+    assertEquals(expected, actual);
+  }
+
+  private static void execute_extract_json(ExprValue expected, String json, String path) {
```
currantw, line 364: Can this be generalized to allow multiple paths, so you can use it for `json_extract_search_list_of_paths` as well?

```suggestion
  private static void execute_extract_json(ExprValue expected, String json, String... paths) {
```

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -216,5 +218,153 @@ void json_returnsSemanticCheckException() {
...
+    assertTrue(
+        invalidJsonError
+            .getMessage()
+            .startsWith(
+                "JSON string '{\"invalid\":\"json\", \"string\"}' is not valid. Error"
+                    + " details:"));
+  }
+
+  @Test
+  void json_extract_throws_ExpressionEvaluationException() {
```
currantw, line 335: Here (as well as all other other error tests) can also use `execute_extract_json` as well, right?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -216,5 +218,153 @@ void json_returnsSemanticCheckException() {
...
+  }
+
+  @Test
+  void json_extract_search() {
+    ExprValue expected = new ExprIntegerValue(1);
+    execute_extract_json(expected, "{\"a\":1}", "$.a");
+  }
+
+  @Test
+  void json_extract_search_arrays() {
```
currantw, line 241: Do we have tests for the following:
- Multiple paths where one or more paths match an array
- Multiple paths where one or more paths match more than one value

currantw (commented):
Reviewed to `ec6ff5e`.

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,72 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+Description
+>>>>>>>>>>>
+
+Usage: `json_extract(doc, path[, path])` Extracts a JSON value from a json document based on the path specified.
+
+Argument type: STRING, STRING
+
+Return type: STRING/BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
+
+- Up to 3 paths can be provided, and results of each `path` with be returned in an ARRAY.
```
acarbonetto, line 78: phrasing is a little confusing. 
```suggestion
- Up to 3 paths can be provided, and results of all possible `path`s will be returned in an ARRAY.
```

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,72 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+____________
+
+Description
+>>>>>>>>>>>
+
+Usage: `json_extract(doc, path[, path])` Extracts a JSON value from a json document based on the path specified.
+
+Argument type: STRING, STRING
+
+Return type: STRING/BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
```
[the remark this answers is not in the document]
acarbonetto: Shouldn't we always return an ARRAY?

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,72 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+
+- Up to 3 paths can be provided, and results of each `path` with be returned in an ARRAY.
+- Returns an ARRAY if `path` points to multiple results (e.g. $.a[*]) or if the `path` points to an array.
+- Return null if `path` is not valid, or if JSON `doc` is MISSING or NULL.
```
acarbonetto, line 80: null or empty array?

docs/user/ppl/functions/json.rst
acarbonetto, line 82: why?

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
acarbonetto: ...and maybe test the 4 path case where we get a parsing error

acarbonetto: we want to test a couple of other error scenarios mentioned:  such as null returns and Semantic errors for malformed json.

docs/user/ppl/functions/json.rst
kenrickyap: addressed

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
kenrickyap: added test IT test cases for use of multiple paths and for the cases when semantic errors are thrown. Though I am curious why we need to test for errors in this case as I cannot find any other IT test testing for errors.

In addition by the time the IT test receive errors they are received as ResponseErrors so had to check if error contains reference to semantic error.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: resolved

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
kenrickyap: both "null" and "" are considered valid json strings by json_valid, since we expect that users will use json_valid to first filter for valid json then use json extract to avoid errors, we need these specific cases to be consistent with json_valid

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: resolved

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
kenrickyap: added

docs/user/ppl/functions/json.rst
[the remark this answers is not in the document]
kenrickyap: in the case that only one string is provided, we return the associated type which will not be a json string. e.g. if path "$.a" is provided from json_string "{"a": 1}: then a integer is returned

docs/user/ppl/functions/json.rst
kenrickyap: resolved

docs/user/ppl/functions/json.rst
kenrickyap: in the case of multiple paths an array of nulls would be returned

docs/user/ppl/functions/json.rst
kenrickyap: this is not handled in code in this PR, this is the default error thrown if provided function does not matched defined schema for functions. Thrown by DefaultFunctionResolver.

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
currantw: Thanks!

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,73 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+    |--------------------+--------------------------------------+-------------------------|
+    | json nested object | {"a":"1","b":{"c":"2","d":"3"}}      | {'c': '2', 'd': '3'}    |
+    | json nested list   | {"a":"1","b":[{"c":"2"}, {"c":"3"}]} | [{'c': '2'},{'c': '3'}] |
+    | json object        | {"a":"1","b":"2"}                    | 2                       |
+    | json array         | [1, 2, 3, 4]                         | null                    |
+    | json scalar string | "abc"                                | null                    |
+    | json empty string  |                                      | null                    |
+    +--------------------+--------------------------------------+-------------------------+
+
+    os> source=json_test | where test_name="json nested list" | eval json_extract=json_extract(json_string, '$.b[1].c') | fields test_name, json_string, json_extract
```
Yury-Fridlyand, line 100: It will be kind of you if you post examples and description to the PR

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,73 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+    | json empty string  |                                      | null                                      |
+    +--------------------+--------------------------------------+-------------------------------------------+
+
+JSON_EXTRACT
+____________
+
+Description
+>>>>>>>>>>>
+
+Usage: `json_extract(doc, path[, path[, path])` Extracts a JSON value from a json document based on the path(s) specified.
```
Yury-Fridlyand, line 72: So, it accepts 1-3 paths. Why not array of paths then?

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,73 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+- Up to 3 paths can be provided, and results of all possible `path`s will be returned in an ARRAY.
+- If only one `path` is provided, returns an ARRAY if `path` points to multiple results (e.g. $.a[*]) or if the `path` points to an array.
+- Return null if `path` is not valid, or if JSON `doc` is MISSING or NULL.
+- If multiple paths are provided with paths that are not valid, will return an ARRAY where results of invalid paths are null.
+- Throws SemanticCheckException if `doc` or any `path` is malformed.
+- Throws ExpressionEvaluationException if any `path` is missing.
+
+Example::
+
+    os> source=json_test | where json_valid(json_string) | eval json_extract=json_extract(json_string, '$.b') | fields test_name, json_string, json_extract
```
Yury-Fridlyand, line 87: Does it accept "legacy" path (path which starts with ".")? if yes, this should be tested and documented. If no, this should be documented.

LantaoJin:
Resolved by https://github.com/opensearch-project/sql/pull/3559 with Calcite implementation.

PR #3263, opened 2025-01-24
Improved patterns command with new algorithm

### LINKED ISSUE #3251: [RFC] Improve patterns command with more advanced log pattern algorithms
## Problem Statement
OpenSearch PPL currently supports [patterns command](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/patterns.rst) to extract log patterns by deriving a `patterns_field` from selected log message field. The default way it generates log pattern field is a simple approach that ignores specified chars and treat the remaining chars as the pattern. Users can query top N log patterns by grouping log messages by the new derived log pattern field like this query 
`source=t | patterns message | stats count() as count, take(message, 100) by patterns_field | sort - count`. We will call this current approach the simple log pattern in the whole doc. The problem of this approach is its grouping accuracy is relatively low in most of industrial software logs. To achieve relatively high grouping accuracy, it needs ops engineer to have specific domain knowledge and manually pass regex pattern to this command. Today's industrial log statements evolve quickly, manually parsing log pattern becomes a challenging task. So we want to introduce more advanced log pattern algorithm to automatically extract high grouping accuracy log patterns.

### Current State
Though this patterns command is simple, it has following downsides:

1. The default generated pattern is usually a combination of punctuations by removing alphanumerical chars, which are not human-readable. i.e. `"... - - [--::.] \" / /.\"   \"-\" \"/. (;  _; :.) / /.\"` for original message `117.251.123.89 - - [2018-07-30T09:46:45.214Z] \"GET /apm HTTP/1.1\" 200 8002 \"-\" \"Mozilla/5.0 (X11; Linux x86_64; rv:6.0a1) Gecko/20110421 Firefox/6.0a1\`
2. It doesn't have high grouping accuracy when there are multiple log statements share the same punctuation pattern. Plus, text tokens dominate in industrial software log messages. i.e. These three different log messages will be treated as sharing the same pattern ` , .`. for `Server started, waiting.`, `Server stopped, waiting.`, `Changing view, hdfs.`. 

## Long-Term Goals
* Improve patterns command with automatic log pattern extraction algorithms(or with minimal human intervention)
* Improve log pattern grouping accuracy
* Scale log pattern extraction across millions of log messages

## Proposal
1. Empower patterns command with advanced log pattern algorithms mentioned in [previously proposed OpenSearch RFC](https://github.com/opensearch-project/OpenSearch/issues/16627).
2. Refactor patterns command as a standalone abstraction(currently it's an implementation of ParseExpression), considering there are different methods of log pattern extraction.
3. Push down complex log pattern extraction code to OpenSearch(probably by painless script) to achieve distributed log pattern extraction. _(**Probably out of scope**: We could invest a code generation module in sql plugin that translates physical plan's Volcano model to a single function that fuses multiple operators together.)_

## Approach
As proposed, we will make patterns command a standalone abstraction. Current simple log pattern fuses a PatternsExpression into Project operator. The PatternsExpression is bound to the target log message field and applies regex on the log message field to generate parsed result as new derived pattern field in Project operator.

Simple log pattern's computing paradigm fits very well in Volcano model's iterator pattern, aka processing result from upstream operator row by row. However, the proposed log pattern algorithm called Brain in previous RFC conflicts with it. The computing paradigm of Brain algorithm is to iterate all of input messages firstly to calculate something like token histogram and then it denotes variable tokens row by row for input messages based on that histogram. So it's hard to extend current ParseExpression to support it.

Luckily, we observed that log pattern algorithms can be classified as two types: ONLINE and OFFLINE. ONLINE algorithms are kind of streaming algorithm. Simple log pattern and other ONLINE algorithms fall into this category. OFFLINE algorithms have to iterate all of log messages firstly and then figure out log pattern in the second pass or more passes. Window operator happens to satisfy both computing paradigms. For ONLINE log pattern algorithms, we can create OpenSearch specialized streaming window functions, just like SQL's ROW_NUMBER(), RANK(), etc. For OFFLINE log pattern algorithms, we can create OpenSearch specialized buffering window functions, just like SQL's SUM(), PERCENT_RANK(), etc. The difference is that the log pattern buffering window function has an n-to-n relationship, rather than an n-to-1 relationship.

### First Stage

In first stage, we want to focus on grouping log patterns on OpenSearch coordinator node. I will use simple plan representation in text with explanation to elaborate the implementation draft.

For the following simple PPL query:
`source=t | patterns message`

#### Current Simple Log Pattern Plan
The current final physical plan is like:
```
ProjectOperator[message#1, PatternsExpression(message#1) as patterns_field#2]
+- OpenSearchIndexScan
```
The above plan means patterns_field is a new field with the evaluation result of PatternsExpression(message#1)

#### New Simple Log Pattern Plan
After introducing consolidation pattern window functions, the query could be(For now, I reuse the SQL syntax to represent a window operator in PPL):
`source=t | patterns(SIMPLE, message#1, other algorithm specific arguments...) over()`

The final plan for original Simple log pattern will be changed to:
```
Project[message#1, patterns_field#2]
+- Window[simple(message#1), windowsSpec(partitionBy=null, frame=CurrentRowWindowFrame)]
   +- OpenSearchIndexScan
```
The above plan means while analyzing unresolved plan of patterns command, if the algorithm method falls into ONLINE algorithm, we assign CurrentRowWindowFrame to window specification. CurrentRowWindowFrame only stores current value and previous value of the iterator. It's enough for the row by row iterating paradigm. This operator can be fused with other operators in a single pipeline, that means it's non-blocking.

#### New Brain Log Pattern Plan
The final plan for OFFLINE algorithms like Brain log pattern algorithm will be like:
```
Project[message#1, patterns_field#2]
+- Window[brain(message#1), windowsSpec(partitionBy=null, frame=PeerRowsWindowFrame)]
   +- OpenSearchIndexScan
```
The above plan means while analyzing unresolved plan of patterns command, if the algorithm method falls into OFFLINE algorithm, we assign PeerRowsWindowFrame to window specification. PeerRowsWindowFrame buffers each row value into peers, aka List<ExprValue>. While calling next() function, the algorithm will compute histograms before continuing iteration. This operation will be blocking like aggregation does. 

### Second Stage

The goal of second stage is to scale log pattern extraction by pushing down its execution to OpenSearch. This needs further research and testing. So leaving this part TBD.

### New operator
Instead of extending existing WindowOperator, we can create a brand new operator called LogPatternOperator to accommodate mentioned implementations. But this will have similar abstraction to window functions.

## Implementation Discussion
* If we adopt window function abstraction, do we want to support `partition by` and `sort by` for patterns command? I do see it could be useful but we haven't expect such requirements before.
* Different log pattern algorithms have different grouping accuracy over different logs. Do we want to allow users to override complex input parameters per algorithm?
* If we want to support distributed log pattern extraction, OpenSearch doesn't have shuffle exchange, how do we do the reduce operation for log pattern results from data node?

songkant-aws:
### Description
This PR introduces enhancement on original patterns command. It keeps the patterns command naming and rebuild it with specific window functions to parse log messages with different pattern method(log parser algorithms). See this RFC: https://github.com/opensearch-project/sql/issues/3251

The sample query input and output will be like:

* New brain algorithm log parser:
Screenshot 2025-02-14 at 14 22 01

* Original regex based patterns command:
Screenshot 2025-02-14 at 14 22 21

* After changing default pattern method to brain, the default behavior is changed:
Screenshot 2025-02-14 at 14 23 15

### Related Issues
Resolves #3251 

### Check List


YANG-DB:
Thanks for the initiative!!

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -283,18 +289,35 @@ public UnresolvedPlan visitParseCommand(OpenSearchPPLParser.ParseCommandContext
...
-    return new Parse(ParseMethod.PATTERNS, sourceField, pattern, arguments);
+    Literal alias = arguments.getOrDefault("new_field", AstDSL.stringLiteral("patterns_field"));
+    return new Pattern(
+        new Alias(
+            "patterns_field",
+            new WindowFunction(
+                new Function(
+                    ctx.pattern_method != null
+                        ? ctx.pattern_method.getText().toLowerCase(Locale.ROOT)
+                        : BuiltinFunctionName.BRAIN.name(), // By default, use new algorithm
```
penghuo, line 315: Could we introduce a setting to control the default pattern algorithms?

core/src/main/java/org/opensearch/sql/expression/window/patterns/StreamPatternWindowFunction.java
```diff
@@ -0,0 +1,80 @@
...
+                    expression instanceof NamedArgumentExpression
+                        && ((NamedArgumentExpression) expression)
+                            .getArgName()
+                            .equalsIgnoreCase("pattern"))
+            .map(
+                expression ->
+                    ((NamedArgumentExpression) expression).getValue().valueOf().stringValue())
+            .findFirst()
+            .orElse("");
+    return new StreamPatternRowWindowFrame(
```
penghuo, line 55: do we need StreamPatternRowWindowFrame? could we just use CurrentRowWindowFrame instead? patternExpression could be member of StreamPatternWindowFunction?

core/src/main/java/org/opensearch/sql/expression/window/frame/PeerRowsWindowFrame.java
```diff
@@ -109,20 +118,15 @@ public void load(PeekingIterator<ExprValue> it) {
...
+  protected boolean isPeer(ExprValue next) {
     List<Expression> sortFields =
         windowDefinition.getSortList().stream().map(Pair::getRight).collect(Collectors.toList());
 
     ExprValue last = peers.get(peers.size() - 1);
     return resolve(sortFields, last).equals(resolve(sortFields, next));
   }
 
-  private boolean isSamePartition(ExprValue next) {
+  protected boolean isSamePartition(ExprValue next) {
```
penghuo, line 129: revert unnecessary change.

core/src/main/java/org/opensearch/sql/expression/window/frame/BufferPatternRowsWindowFrame.java
```diff
@@ -0,0 +1,62 @@
...
+
+    List<String> logMessages =
+        peers.stream()
+            .map(
+                exprValue -> {
+                  ExprValue value = sourceField.valueOf(exprValue.bindingTuples());
+                  return value.stringValue();
+                })
+            .toList();
+    this.preprocessedMessages.addAll(logParser.preprocessAllLogs(logMessages));
```
penghuo, line 56: BufferPatternRowsWindowFrame should have it own spec, doet it means over all rows? @dai-chen  WindowFrame definition and How to use WIndowFrame should be seperate, right?

penghuo (commented):
In general, 
* add user facing doc to explain new BRAIN alg with example will be helpfull.
* add UT could help reviwer understand the logic also. @dai-chen did a good example. https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/expression/window/frame/PeerRowsWindowFrame.java
* add benchmark to compare BRAIN and SIMPLE also help a lot.

core/src/main/java/org/opensearch/sql/expression/window/patterns/StreamPatternWindowFunction.java
songkant-aws: Sure, I will remove the need of StreamPatternRowWindowFrame in the next revision.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
songkant-aws: Ok, I see. Like a java properties to specify default algorithm name? I can add that.

core/src/main/java/org/opensearch/sql/expression/window/frame/BufferPatternRowsWindowFrame.java
songkant-aws: For simplicity, I set the spec to empty `partition by` and empty `sort by` during AST tree parsing unresolved WindowFunction, which treats the window frame range is all rows on coordinator node. Because I haven't seen requirements on sorting and partitioning on other columns. 

I think we can add `partition by` and `sort by` syntax if we see values in case users want to specify them. Thoughts?

core/src/main/java/org/opensearch/sql/expression/window/WindowFunctions.java
```diff
@@ -34,6 +38,8 @@ public void register(BuiltinFunctionRepository repository) {
     repository.register(rowNumber());
     repository.register(rank());
     repository.register(denseRank());
+    repository.register(brain());
+    repository.register(simplePattern());
```
dai-chen, line 42: I'm thinking are these new algorithm real window function? If users specify `order by` or `partition by`, does it still generate meaningful result?

core/src/main/java/org/opensearch/sql/ast/dsl/AstDSL.java
```diff
@@ -489,6 +491,22 @@ public static Parse parse(
     return new Parse(parseMethod, sourceField, pattern, arguments, input);
   }
 
+  public static Pattern pattern(
+      UnresolvedPlan input,
+      PatternMethod patternMethod,
+      UnresolvedExpression sourceField,
+      String alias,
+      java.util.Map<String, Literal> arguments) {
+    return new Pattern(
```
dai-chen, line 500: Just wonder if pattern is mostly syntax sugar for pattern window function, is new logical operator still required?

core/src/main/java/org/opensearch/sql/expression/window/patterns/BufferPatternWindowFunction.java
```diff
@@ -0,0 +1,88 @@
...
+import org.opensearch.sql.expression.NamedArgumentExpression;
+import org.opensearch.sql.expression.env.Environment;
+import org.opensearch.sql.expression.function.BuiltinFunctionName;
+import org.opensearch.sql.expression.window.WindowDefinition;
+import org.opensearch.sql.expression.window.WindowFunctionExpression;
+import org.opensearch.sql.expression.window.frame.BufferPatternRowsWindowFrame;
+import org.opensearch.sql.expression.window.frame.WindowFrame;
+
+@EqualsAndHashCode(callSuper = true)
+public class BufferPatternWindowFunction extends FunctionExpression
```
dai-chen, line 29: `order by` expression in window definition decides who're peers right? What's the order by expression for buffer offline pattern function? I didn't find it and may miss it in design doc:

```
Project[message#1, patterns_field#2]
+- Window[brain(message#1), windowsSpec(partitionBy=null, frame=PeerRowsWindowFrame)]
   +- OpenSearchIndexScan
```

dai-chen (commented):
A high-level question about the long-term approach: Currently, the pattern command always generates a new field, which offers flexibility as users can perform aggregation (stats) or other operations on it afterward. However, for large datasets, I assume most users would prefer functionality similar to [CloudWatch Logs Query Syntax - Pattern](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_QuerySyntax-Pattern.html).

Do you see any potential issues with the current implementation, which is based on SQL window functions combined with aggregate functions, as we scale this feature in the future?

core/src/main/java/org/opensearch/sql/ast/dsl/AstDSL.java
songkant-aws: After the integration with LogicalWindow, I see they are quite similar. Yeah, I think Pattern operator is probably not needed.

songkant-aws:
> Do you see any potential issues with the current implementation, which is based on SQL window functions combined with aggregate functions, as we scale this feature in the future?

@dai-chen It's a good question. Actually, it's one of my previous thought to use a new specific operator to directly return grouped logs with sample count or grouped samples.

It's more like a compound aggregation operator with a default `group by`. But current aggregation abstraction only supports row by row iteration. The log pattern algorithms cover both streaming and buffering computing paradigm. Then I figure out this is more close to window operator's abstraction. To achieve what Cloudwatch has done, I think we could extend pattern functions from `AggregateWindowFunction` that has a special aggregator state structure.

As to `partition by` and `sort by` specification, my initial plan is to deliver a simple version without them because that's probably enough.
`partition by` is probably more useful than `sort by`. Imagine users may want to see different log patterns for INFO, WARN, ERROR partitions.
I haven't seen a use case for `sort by` yet. Maybe they need to see latest logs `sort by` timestamp?

songkant-aws:
To scale the pattern function, I think other issues are pending investigation.

1. How to push down operator to OpenSearch data node? Is Painless script one of option?
2. OpenSearch doesn't have shuffle mechanism for now, how to perform large dataset aggregation like operation with simple map reduce? If data volume is quite large, spilling additional data to disk is probably required during execution.

songkant-aws:
AvgTime benchmark:

```
Benchmark                                          (testDataType)  Mode  Cnt  Score    Error  Units
ComparisonOperatorBenchmark.testEqualOperator                 int  avgt    3  0.002 ±  0.001  ms/op
ComparisonOperatorBenchmark.testEqualOperator              string  avgt    3  0.005 ±  0.001  ms/op
ComparisonOperatorBenchmark.testEqualOperator                date  avgt    3  0.001 ±  0.001  ms/op
ComparisonOperatorBenchmark.testGreaterOperator               int  avgt    3  0.002 ±  0.001  ms/op
ComparisonOperatorBenchmark.testGreaterOperator            string  avgt    3  0.005 ±  0.002  ms/op
ComparisonOperatorBenchmark.testGreaterOperator              date  avgt    3  0.001 ±  0.001  ms/op
ComparisonOperatorBenchmark.testLessOperator                  int  avgt    3  0.002 ±  0.001  ms/op
ComparisonOperatorBenchmark.testLessOperator               string  avgt    3  0.006 ±  0.001  ms/op
ComparisonOperatorBenchmark.testLessOperator                 date  avgt    3  0.002 ±  0.001  ms/op
PatternsWindowFunctionBenchmark.testBrain                     N/A  avgt    3  0.131 ±  0.002  ms/op
PatternsWindowFunctionBenchmark.testSimplePattern             N/A  avgt    3  0.001 ±  0.001  ms/op
```

core/src/main/java/org/opensearch/sql/ast/dsl/AstDSL.java
songkant-aws: I don't find the Window unresolvedPlan. So replace Pattern with Window instead.

songkant-aws:
This PR still needs some action items.
1. Pending adding setting to switch default log pattern algorithm
2. Pending discussion whether we want to deliver `partition by` and `order by` given its effect. (Log patterns are probably special window function that is not sensitive to `partition by` and `order by`)

core/src/main/java/org/opensearch/sql/expression/window/frame/BufferPatternRowsWindowFrame.java
LantaoJin: Tests with non-empty `partition by` and non-empty `sort by` are still required in PR.

docs/user/ppl/cmd/patterns.rst
```diff
@@ -12,24 +12,45 @@ patterns
...
 
 
 Syntax
 ============
-patterns [new_field=<new-field-name>] [pattern=<pattern>] <field>
+patterns [new_field=<new-field-name>] (algorithm parameters...) <field> <pattern_method>
+
+Simple Pattern
+============
+patterns [new_field=<new-field-name>] [pattern=<pattern>] <field> SIMPLE_PATTERN
```
LantaoJin, line 25: The seems a breaking change, why not make `SIMPLE_PATTERN` optional for compatibility.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -115,18 +115,25 @@ parseCommand
...
+   | (VARIABLE_COUNT_THRESHOLD EQUAL variable_count_threshold = integerLiteral)
+   | (FREQUENCY_THRESHOLD_PERCENTAGE EQUAL frequency_threshold_percentage = decimalLiteral)
    ;
 
-patternsMethod
-   : PUNCT
-   | REGEX
+patternMethod
+   : SIMPLE_PATTERN
+   | BRAIN
```
LantaoJin, line 136: `patternMethod` should be added to `keywordsCanBeId` as well.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -115,18 +115,25 @@ parseCommand
    : PARSE (source_field = expression) (pattern = stringLiteral)
    ;
 
+patternsMethod
+   : PUNCT
+   | REGEX
```
LantaoJin, line 120: `patternsMethod` should be added to `keywordsCanBeId`

LantaoJin (changes requested):

docs/user/ppl/cmd/patterns.rst
songkant-aws: It's still compatible. If we omit SIMPLE_PATTERN, it will use BRAIN method by default. Peng suggested we can add a configuration to decide which is the default behavior.

LantaoJin: > It's still compatible. If we omit `SIMPLE_PATTERN`, it will use BRAIN method by default. Peng suggested we can add a configuration to decide which is the default behavior.

We don't call it compatible if it brings breaking change with original command syntax. I doubt about using the new BRAIN algorithm as default output before it has been running stably for a while. @penghuo any thoughts here?

penghuo: > Peng suggested we can add a configuration to decide which is the default behavior.
Did we add setting already? 
If not, we should default to SIMPLE_PATTERN.

penghuo:
@songkant-aws could u merge conflict?

core/src/main/java/org/opensearch/sql/expression/window/frame/PeerRowsWindowFrame.java
songkant-aws: Done

core/src/main/java/org/opensearch/sql/expression/window/frame/BufferPatternRowsWindowFrame.java
songkant-aws: Added non-empty `partition by` and `sort by` unit tests.

docs/user/ppl/cmd/patterns.rst
songkant-aws: Add a new setting now. It will default to SIMPLE_PATTERN. Please review the change. Let me know if it makes sense.

ppl/src/main/antlr/OpenSearchPPLParser.g4
songkant-aws: Done

ppl/src/main/antlr/OpenSearchPPLParser.g4
songkant-aws: Done

songkant-aws:
@penghuo @dai-chen @LantaoJin I have addressed several comments and added a new setting `plugins.ppl.default.pattern.method` to allow overriding default pattern method. For now, the default pattern method is still regex based SIMPLE_PATTERN. Please let me know if the current change is reasonable. Thanks!

penghuo:
@songkant-aws please fix failed UT coverage. then we are ready to merge.

songkant-aws:
@penghuo Fixed all of CI checks with more test coverage and cases.

penghuo (approved):
Thx!

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -283,18 +292,35 @@ public UnresolvedPlan visitParseCommand(OpenSearchPPLParser.ParseCommandContext
   @Override
   public UnresolvedPlan visitPatternsCommand(OpenSearchPPLParser.PatternsCommandContext ctx) {
     UnresolvedExpression sourceField = internalVisitExpression(ctx.source_field);
-    ImmutableMap.Builder<String, Literal> builder = ImmutableMap.builder();
+    List<UnresolvedExpression> unresolvedArguments = new ArrayList<>();
+    unresolvedArguments.add(sourceField);
+    AtomicReference<String> alias = new AtomicReference<>("patterns_field");
```
LantaoJin, line 297: `AstBuilder` is created per query. No necessary to wrap with `AtomicReference`

common/src/main/java/org/opensearch/sql/common/patterns/BrainLogParser.java
```diff
@@ -0,0 +1,443 @@
...
+   * @param delimiters a list of delimiters to be replaced with empty string after regex replacement
+   */
+  public BrainLogParser(
+      int variableCountThreshold,
+      float thresholdPercentage,
+      Map<Pattern, String> filterPatternVariableMap,
+      List<String> delimiters) {
+    if (thresholdPercentage < 0.0f || thresholdPercentage > 1.0f) {
+      throw new IllegalArgumentException("Threshold percentage must be between 0.0 and 1.0");
+    }
```
LantaoJin, line 133: What if the `thresholdPercentage = 0` or `thresholdPercentage = 1`

core/src/main/java/org/opensearch/sql/expression/window/patterns/BufferPatternWindowFunction.java
```diff
@@ -0,0 +1,70 @@
...
+import org.opensearch.sql.expression.FunctionExpression;
+import org.opensearch.sql.expression.env.Environment;
+import org.opensearch.sql.expression.function.BuiltinFunctionName;
+import org.opensearch.sql.expression.window.WindowDefinition;
+import org.opensearch.sql.expression.window.WindowFunctionExpression;
+import org.opensearch.sql.expression.window.frame.BufferPatternRowsWindowFrame;
+import org.opensearch.sql.expression.window.frame.WindowFrame;
+import org.opensearch.sql.utils.FunctionUtils;
+
+@EqualsAndHashCode(callSuper = true)
```
LantaoJin, line 28: minor: `@ToString`

core/src/main/java/org/opensearch/sql/expression/window/patterns/StreamPatternWindowFunction.java
```diff
@@ -0,0 +1,70 @@
...
+import org.opensearch.sql.expression.env.Environment;
+import org.opensearch.sql.expression.function.BuiltinFunctionName;
+import org.opensearch.sql.expression.parse.PatternsExpression;
+import org.opensearch.sql.expression.window.WindowDefinition;
+import org.opensearch.sql.expression.window.WindowFunctionExpression;
+import org.opensearch.sql.expression.window.frame.CurrentRowWindowFrame;
+import org.opensearch.sql.expression.window.frame.WindowFrame;
+import org.opensearch.sql.utils.FunctionUtils;
+
+@EqualsAndHashCode(callSuper = true)
```
LantaoJin, line 29: ditto

common/src/main/java/org/opensearch/sql/common/patterns/BrainLogParser.java
songkant-aws: It's valid, which means 0 or the highest frequency.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
songkant-aws: Seems like IDE has compilation error when using a temp String variable in lambda expression. So I use AtomicReference as a workaround

core/src/main/java/org/opensearch/sql/expression/window/patterns/BufferPatternWindowFunction.java
songkant-aws: I override this function's toString specifically.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
LantaoJin: Ok, I see, the `String alias` must be a "final" variable from the enclosing scope.

Issue #3030, opened 2024-09-16
[FEATURE]Add `flatten` Command to PPL


YANG-DB:
**Is your feature request related to a problem?**
OpenSearch Piped Processing Language (PPL) currently lacks a native command to flatten nested objects or arrays in documents. Many datasets, especially those containing JSON objects, have deeply nested fields that are difficult to work with in their raw form. The `flatten` command will simplify these structures and make it easier to analyze and extract data.

**What solution would you like?**
Introduce a flatten command in PPL that can handle arrays or nested fields, producing a `flattened` result that contains all the nested elements at the top level.

**Syntax:**
```sql
source=<data_source> | flatten <nested_field>  | fields <fields_to_select>
```
 - The `flatten` command takes a nested array or object field and returns each element as part of a flat structure.

---

**Example Use Cases**

1. Flattening an Array Field

```sql
source=my-index  | flatten bridges | fields _time, bridges, city, country
```
This query `flattens` the bridges array field.

_Example Input:_

```json
{
  "_time": "2024-09-13T12:00:00",
  "bridges": [
    {"name": "Tower Bridge", "length": 801},
    {"name": "London Bridge", "length": 928}
  ],
  "city": "London",
  "country": "England"
}
```

_Expected Output:_

```json
[
  {
    "_time": "2024-09-13T12:00:00",
    "name": "Tower Bridge",
    "length": 801,
    "city": "London",
    "country": "England"
  },
  {
    "_time": "2024-09-13T12:00:00",
    "name": "London Bridge",
    "length": 928,
    "city": "London",
    "country": "England"
  }
]
```

2. Flattening a Nested Object

```sql
source=my-index | flatten details | fields _time, details 
```

This query `flattens` the details object field.

_Example Input:_

```json
{
  "_time": "2024-09-13T12:00:00",
  "details": {
    "name": "Alice",
    "age": 30,
    "address": {
      "street": "Main St",
      "city": "New York"
    }
  }
}
```

_Expected Output:_

```json
{
  "_time": "2024-09-13T12:00:00",
  "name": "Alice",
  "age": 30,
  "street": "Main St",
  "city": "New York"
}
```

---

**Additional Considerations**

- The flatten command should work efficiently with large arrays or deeply nested structures.
- It must handle complex JSON objects where multiple levels of nesting exist.
 - Consider supporting multi-level flattening for more deeply nested fields (e.g., flatten details.address).

acarbonetto:
[...]

Issue #3145, opened 2024-10-31
[FEATURE] Add Support for IP Data Type


currantw:
**Is your feature request related to a problem?**

OpenSearch SQL plugin does not support the [IP address field type](https://opensearch.org/docs/latest/field-types/supported-field-types/ip/).

🚫 IP address fields are converted to strings:

`search=weblog | fields host`

returns `host` with field type `string`.

🚫 IP address fields cannot be correctly for equality:

`search=weblog | where host = "2001:0db7::ff00:42:8329" | fields host`

will not return the value `2001:0db7:0000:0000:0000:ff00:0042:8329`, even though both strings represent the same IP address.

**What solution would you like?**

Outcomes:
- IP addresses can be retrieved using the OpenSearch SQL plugin without conversion to strings.
- IP addresses supports equality operations (`=` and `!=`).
- IP addresses supports comparison operations (`<`, `<=`, `>`, and `>`) if they are both IPv4 or IPv6.
- IP addresses supports sorting (again, if they are all IPv4 or IPv6).
- IP addresses work with IP-specific functions (currently only `cidrmatch` - see #3110).

Proposed Solution:
- Add an new `IP` type to `ExprCoreType`.
- Replace `OpenSearchExprIpValue` with `ExprIpValue` , and update implementation.
- Update `OpenSearchDataType.MappingType` to map `"ip"` fields to `ExprCoreType.IP`.
- Update `OpenSearchExprValueFactory`.
- Update other code, unit tests, and integration tests as necessary.

**What alternatives have you considered?**

None

**Do you have any additional context?**

This is closely related to https://github.com/opensearch-project/sql/pull/3110. This issue added a new `cidrmatch(ip, cidr)` function that returns whether the given IP address is within the specified CIDR IP address range. As part of this work, the SQL plugin was updated to cast IP addresses to strings - previously, it would raise an exception.

kedbirhan:
we really need this feature, we can't even do basic IP lookups right now!

query
```sql
SELECT * 
FROM logs-vpc
WHERE dstaddr = "10.100.138.82"
LIMIT 10;
```

log

```log
[...]

Issue #3016, opened 2024-09-13
[FEATURE]New `expand_field` PPL Command


YANG-DB:
**Is your feature request related to a problem?**
Adding a PPL new `expand_field` command which adds array and nested object expansion functionality to PPL

**Is your feature request related to a problem? Please describe.**
OpenSearch's Piped Processing Language (PPL) currently lacks an efficient way to expand arrays and nested objects into separate events, similar to SQL's UNNEST or JSON expansion functions. This limitation hinders the analysis of complex data structures, particularly when working with JSON logs or documents containing arrays or nested objects.

**Describe the solution you'd like**
We propose adding a new command to PPL that would allow users to expand arrays and nested objects into separate events, similar to SQL's UNNEST function, but with additional flexibility. 

**The functionality should:**

1. Expand array fields or nested objects into separate events (similar to SQL's UNNEST)
2. Retain all other fields from the original event in each new event (addressing a limitation of SQL UNNEST)
3. Support nested fields and complex JSON structures (going beyond basic SQL capabilities)
4. Allow for subsequent processing of each expanded value in the PPL pipeline
5. Work seamlessly with unstructured or semi-structured data (unlike SQL, which typically requires predefined schemas)

**SQL-like example and comparison:**
Consider this SQL-like syntax:
```sql
SELECT * FROM my_index
CROSS JOIN UNNEST(items) AS expanded_item
WHERE expanded_item.status = 'active'
```

The proposed OpenSearch PPL equivalent might look like:
```sql
source =  my_index 
| parse my_nested_field `?<items>[*]` as items
| expand_field items
| where items.status = "active"
```

_Key differences and advantages:_

1. No need for explicit JOIN syntax, making it more intuitive for log analysis
2. Automatic handling of nested structures without need for complex JSON parsing functions
3. Ability to work with dynamic schemas and unstructured data

**Describe alternatives you've considered**
Current alternatives include:

1. Using complex JSON path queries, which can be cumbersome
2. Processing the data outside of OpenSearch, reducing real-time analysis capabilities

**Additional context**
This feature would bridge the gap between SQL's structured data handling and the need for flexible, real-time analysis of semi-structured log data. It combines the power of SQL's UNNEST with the flexibility required for log and event processing.

**Potential Impact**

* Simplified queries for complex data structures in logs and events
* Enhanced real-time analytics capabilities for nested JSON data
* Improved performance compared to client-side processing of nested structures
[...]

PR #3267, opened 2025-01-27
Add flatten command to ANTLR lexer and parser.

currantw:
Signed-off-by: currantw <taylor.curran@improving.com>

### Description

Implements `flatten` command _for objects only_. Does not implement flattening for arrays or aliasing.

### Related Issues

Part of scope of #3030.

### Check List
- [] New functionality has been documented.


core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +458,69 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+      fieldExpr = (ReferenceExpression) expressionAnalyzer.analyze(node.getField(), context);
+    } catch (SemanticCheckException e) {
+      throw new IllegalArgumentException("Invalid field name for flatten command", e);
+    }
+
+    // Verify that the field type is valid.
+    ExprType fieldType = fieldExpr.type();
+    if (fieldType != STRUCT) {
+      throw new IllegalArgumentException(
+          String.format("Invalid field type '%s' for flatten command", fieldType));
```
acarbonetto, line 482:
```suggestion
          String.format("Invalid field type '%s' for flatten command, expected OBJECT", fieldType));
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +458,69 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    java.util.Map<String, ExprType> addFieldsMap = new HashMap<>();
+    java.util.Map<String, ExprType> removeFieldsMap = new HashMap<>();
+
+    removeFieldsMap.put(fieldName, fieldType);
+
+    for (java.util.Map.Entry<String, ExprType> entry : fieldsMap.entrySet()) {
+      String path = entry.getKey();
+      List<String> pathComponents = Arrays.stream(PATH_SEPARATOR_PATTERN.split(path)).toList();
+
+      // Verify that path starts with the field name.
```
acarbonetto, line 498: is it possible for the path _not_ to start with the field name?

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +458,69 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    java.util.Map<String, ExprType> fieldsMap = env.lookupAllTupleFields(Namespace.FIELD_NAME);
+
+    java.util.Map<String, ExprType> addFieldsMap = new HashMap<>();
+    java.util.Map<String, ExprType> removeFieldsMap = new HashMap<>();
+
+    removeFieldsMap.put(fieldName, fieldType);
+
+    for (java.util.Map.Entry<String, ExprType> entry : fieldsMap.entrySet()) {
+      String path = entry.getKey();
+      List<String> pathComponents = Arrays.stream(PATH_SEPARATOR_PATTERN.split(path)).toList();
```
acarbonetto, line 496: I think turning this into a list, removing the first element, and joining back into a string is a little overkill.  
Consider just using a substring - starting at the index of the first `PATH_SEPARATOR`.  If there's no `PATH_SEPARATOR` then `continue`.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +458,69 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+
+      // Verify that new field does not overwrite an existing field.
+      if (fieldsMap.containsKey(newPath)) {
+        throw new IllegalArgumentException(
+            String.format("Flatten command cannot overwrite field '%s'", newPath));
+      }
+
+      ExprType type = entry.getValue();
+      addFieldsMap.put(newPath, type);
+      removeFieldsMap.put(path, type);
```
acarbonetto, line 514: why do we remove `path`?

core/src/main/java/org/opensearch/sql/ast/tree/Eval.java
```diff
@@ -10,14 +10,12 @@
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
-import lombok.Setter;
```
acarbonetto, line 13: why?

core/src/main/java/org/opensearch/sql/data/model/ExprTupleValue.java
```diff
@@ -69,23 +68,14 @@ public ExprValue keyValue(String key) {
...
-        Entry<String, ExprValue> otherEntry = otherIterator.next();
-        if (!(thisEntry.getKey().equals(otherEntry.getKey())
-            && thisEntry.getValue().equals(otherEntry.getValue()))) {
-          return false;
-        }
-      }
-      return !(thisIterator.hasNext() || otherIterator.hasNext());
     }
+
+    return valueMap.equals(other.tupleValue());
```
acarbonetto, line 78: revert this please

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
```diff
@@ -0,0 +1,74 @@
...
+
+    // Build the flattened field map.
+    String fieldName = field.getAttr();
+    ExprValue exprValue = fieldsMap.get(fieldName);
+
+    Map<String, ExprValue> flattenedFieldsMap = exprValue.tupleValue();
+
+    // Update field map.
+    fieldsMap.putAll(flattenedFieldsMap);
+    fieldsMap.remove(fieldName);
```
acarbonetto, line 61: do we want to remove the original struct from the environment?

core/src/main/java/org/opensearch/sql/data/model/ExprTupleValue.java
currantw: I had a comment explaining this ... but apparently I forgot to _actually post them_! 😆 

https://github.com/opensearch-project/sql/pull/3267#discussion_r1931681630

core/src/main/java/org/opensearch/sql/ast/tree/Eval.java
currantw: I had a comment on this one as well that I never posted 🤦 
https://github.com/opensearch-project/sql/pull/3267#discussion_r1931676132

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +458,72 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    LogicalPlan child = node.getChild().getFirst().accept(this, context);
+    TypeEnvironment env = context.peek();
+
+    // Verify that the field name is valid.
+    ReferenceExpression fieldExpr;
+    try {
+      fieldExpr = (ReferenceExpression) expressionAnalyzer.analyze(node.getField(), context);
+    } catch (SemanticCheckException e) {
+      throw new IllegalArgumentException("Invalid field name for flatten command", e);
+    }
```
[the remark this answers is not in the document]
normanj-bitquill: It looks like `SemanticCheckException` is a `RuntimeException` so you skip catching and throwing a new exception. The message for your `IllegalArgumentException` is helpful. I don't know how useful the message is in `SemanticCheckException`.

ppl/src/test/java/org/opensearch/sql/ppl/parser/AstBuilderTest.java
```diff
@@ -446,6 +447,14 @@ public void testEvalCommand() {
         eval(relation("t"), let(field("r"), function("abs", field("f")))));
   }
 
+  @Test
+  public void testFlattenCommand() {
+    String fieldName = "field_name";
+    assertEqual(
+        String.format("source=t | flatten %s", fieldName),
```
normanj-bitquill, line 454: Should use `StringUtils.format()`, it will handle choosing the Locale to use rather than relying on JVM behaviour.

ppl/src/test/java/org/opensearch/sql/ppl/parser/AstExpressionBuilderTest.java
```diff
@@ -252,6 +253,14 @@ public void testEvalIfFunctionExpr() {
                     intLiteral(0)))));
   }
 
+  @Test
+  public void testFlattenExpr() {
+    String fieldName = "field_name";
+    assertEqual(
+        String.format("source=t | flatten %s", fieldName),
```
normanj-bitquill, line 260: Same as above, `StringUtils.format()` is a better choice.

core/src/main/java/org/opensearch/sql/data/model/ExprTupleValue.java
```diff
@@ -69,23 +68,14 @@ public ExprValue keyValue(String key) {
...
   }
```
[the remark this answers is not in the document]
YANG-DB: Nice catch !!

ppl/src/test/java/org/opensearch/sql/ppl/parser/AstExpressionBuilderTest.java
currantw: Thanks, done! Also updated a few other places in this PR that also used `String.format` instead of `StringUtils.format`.

ppl/src/test/java/org/opensearch/sql/ppl/parser/AstBuilderTest.java
currantw: Thanks, done! Also updated a few other places in this PR that also used `String.format` instead of `StringUtils.format`.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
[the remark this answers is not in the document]
currantw: It's not terrible, but I think this one is a bit more user friendly. That being said, it doesn't actually contain the offending the offending field name, so might not be as helpful for someone using multiple chained `flatten` command 🤷 

> can't resolve Symbol(namespace=FIELD_NAME, name=invalid) in type env

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Thanks. Added supported field type to the error message (though with slightly different wording).

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Good call. Updated to do so.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: The `fieldsMap` includes _all_ the fields in the type environment, so yes, if I understand your question correctly?

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: I guess that's based on whether or not we want to keep the original struct in the environment? I'll address that in more detail on that thread, if that's OK.

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
currantw: I _think_ we do. Here are some of my thoughts.
- That's what Spark already does (it doesn't preserve the original field).
- I can't think of a clear use case for _not_ removing the original struct. Under what circumstances would a user want to flatten a struct, and _also_ preserve the original field?
- It seems more intuitive to me. "Flatten" seems to imply that we are flattening the original struct, not (in essence), copying and _then_ flattening it.
- Not removing the original struct would be even more unintuitive (at least to me) if we are flattening a nested struct (which is supported).
- If a user really _did_ want to preserve a copy of the original struct, it would be easy enough to do so using `eval`.

Thoughts?

core/src/test/java/org/opensearch/sql/expression/ip/IPFunctionsTest.java
```diff
@@ -48,19 +51,27 @@ public class IPFunctionsTest {
...
+    assertTrue(
+        ex.getMessage().matches("^IP address string 'INVALID' is not valid. Error details: .*"));
+
+    ex =
+        assertThrows(
+            SemanticCheckException.class,
+            () -> execute(IPv4AddressWithin, ExprValueUtils.stringValue("INVALID")));
+    assertTrue(
+        ex.getMessage()
+            .matches("^IP address range string 'INVALID' is not valid. Error details: .*"));
```
currantw, line 70: Minor unrelated cleanup to this test class I wrote in #3145. Realized that this test wasn't actually correctly testing the error message 🤦 A couple other IntelliJ code quality suggestions.

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,65 @@
...
+PPL query::
+
+    os> source=cities | flatten location.coordinates | fields name, location
+    fetched rows / total rows = 2/2
+    +-----------+----------------------------------------------------------------------------------------------------+
+    | name      | location                                                                                           |
+    +-----------+----------------------------------------------------------------------------------------------------+
+    | Seattle   | {'state': 'Washington', 'country': 'United States', 'latitude': 47.6061, 'longitude': -122.3328}   |
+    | Vancouver | {'province': 'British Columbia', 'country': 'Canada', 'latitude': 49.2827, 'longitude': -123.1207} |
+    +-----------+----------------------------------------------------------------------------------------------------+
```
currantw, line 64: Also supports flattening nested structs!

integ-test/src/test/java/org/opensearch/sql/ppl/FlattenCommandIT.java
```diff
@@ -0,0 +1,76 @@
...
+        schema("province", "string"),
+        schema("country", "string"),
+        schema("latitude", "float"),
+        schema("longitude", "float"));
+    verifyDataRows(
+        result,
+        rows("Washington", null, "United States", 47.6061, -122.3328),
+        rows(null, "British Columbia", "Canada", 49.2827, -123.1207));
+  }
+}
```
andy-k-improving, line 76: It's possible that `coordinates` itself being null for some row, in this case the commdn should handle it gracefully and rather than `fail-fast`.

Sample data:
```
{"index":{"_id":"5"}}
{"name": "Seattle", "location": { "state": "Washington", "country": "United States", "coordinates": null}}
```

Current behaviour:
 ```
{'reason': 'Invalid Query', 'details': 'invalid to get tupleValue from value of type UNDEFINED', 'type': 'ExpressionEvaluationException'}
opensearchsql> source=cities | flatten location.coordinates; 

```

File: core/src/main/java/org/opensearch/sql/analysis/Analyzer.java [hunk 15]
```diff
@@ -277,7 +284,7 @@ public LogicalPlan visitRename(Rename node, AnalysisContext context) {
     LogicalPlan child = node.getChild().get(0).accept(this, context);
     ImmutableMap.Builder<ReferenceExpression, ReferenceExpression> renameMapBuilder =
         new ImmutableMap.Builder<>();
-    for (Map renameMap : node.getRenameList()) {
+    for (org.opensearch.sql.ast.expression.Map renameMap : node.getRenameList()) {
```
andy-k-improving, line 287: Why are we touching the `visitRenam( )` function, as I don't see how this related to `flatten` functionality.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -127,6 +134,8 @@ public class Analyzer extends AbstractNodeVisitor<LogicalPlan, AnalysisContext>
 
   private final BuiltinFunctionRepository repository;
 
+  private static final String PATH_SEPARATOR = ".";
```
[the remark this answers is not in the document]
currantw: > As I don't see any usage outside of visitFlatten.

Yup, you're right. I assume it will _ultimately_ need to be moved outside of the method (`visitRename` is going to need it, I think, when #3279 is addressed), but you're right that it should be inside `visitFlatten` for the time being. Ultimately, there should probably be some sort of `PathUtils` class that encapsulates logic for splitting and combining path strings, but that's well beyond the scope of this issue...

core/src/test/java/org/opensearch/sql/planner/DefaultImplementorTest.java
```diff
@@ -325,4 +328,22 @@ public void visitTrendline_should_build_TrendlineOperator() {
     assertInstanceOf(TrendlineOperator.class, implemented);
     assertSame(physicalChild, implemented.getChild().get(0));
   }
+
+  @Test
+  void visitFlatten_should_build_FlattenOperator() {
```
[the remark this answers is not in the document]
currantw: IntelliJ is not so fond of this pattern .. but you're probably right that it's better to be consistent 🤷 

image

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: There are two `Map` classes that are imported into here, now that we are using `Map` in `visitFlatten`, so there is a naming conflict. I figured it made sense to use the fully-qualified name for `org.opensearch.sql.ast.expression.Map`rather than for `java.util.Map`, as I think most people, if they see `Map`, are going to assume that it's `java.util.Map`. 🤷

integ-test/src/test/java/org/opensearch/sql/ppl/FlattenCommandIT.java
currantw: Good idea! I will implement this and add corresponding tests.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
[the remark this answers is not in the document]
andy-k-improving: Yes, share the same thought, let's only expand the scope from method private to class private or even util, if such use-case arise in the future.

doctest/test_mapping/cities.json
```diff
@@ -0,0 +1,34 @@
+{
```
acarbonetto, line 1: I know you like city planning, but I'm not certain we need to add a new dataset?  

Note, I'm getting errors when I run against game_of_thrones dataset: 
- `source=game_of_thrones | flatten name`
- `source=game_of_thrones | flatten parents`

```
{
  "error": {
    "reason": "Invalid Query",
    "details": "Cannot invoke \"org.opensearch.sql.data.model.ExprValue.tupleValue()\" because \"exprValue\" is null",
  },
  "status": 400
}
```

doctest/test_data/cities.json
```diff
@@ -0,0 +1,2 @@
+{"name": "Seattle", "location": { "state": "Washington", "country": "United States", "coordinates": {"latitude": 47.6061, "longitude": -122.3328}}}
+{"name": "Vancouver", "location": { "province": "British Columbia", "country": "Canada", "coordinates": {"latitude": 49.2827, "longitude": -123.1207}}}
```
acarbonetto, line 2: test multi-values too.  For example: 
```
{"name": "Vancouver", "location": [{"province": "British Columbia", "country": "Canada", "coordinates": {"latitude": 49.2827, "longitude": -123.1207}}, {"state": "Washington", "country": "United States", "coordinates": {"latitude": 45.6273, "longitude": -122.6662}}]}
```

I'm getting 
```
{
  "error": {
    "reason": "Invalid Query",
    "details": "invalid to get tupleValue from value of type ARRAY",
  },
  "status": 400
}
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,118 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+   * the given flatten node.
+   */
+  @Override
+  public LogicalPlan visitFlatten(Flatten node, AnalysisContext context) {
+    LogicalPlan child = node.getChild().getFirst().accept(this, context);
+
+    // [A] Get field name and type
+    // ---------------------------
+
+    // Verify that the field name is valid.
```
acarbonetto, line 467: What do you mean by "valid" here?

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,118 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+      fieldExpr = (ReferenceExpression) expressionAnalyzer.analyze(node.getField(), context);
+    } catch (SemanticCheckException e) {
+      throw new IllegalArgumentException("Invalid field name for flatten command", e);
+    }
+
+    String fieldName = fieldExpr.getAttr();
+    ExprType fieldType = fieldExpr.type();
+
+    // Verify that the field type is valid.
+    if (fieldType != STRUCT) {
```
acarbonetto, line 479: in the case of ARRAY, you should grab the first value according to: https://github.com/opensearch-project/sql/issues/3030#issuecomment-2625694041
Otherwise, add a TODO to include it when we do ARRAY-side of flatten.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,118 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    //       nested_struct: {
+    //         string: "value" }}}
+    //
+    // Example 1: 'flatten struct'
+    //
+    // { integer: 0,
+    //   nested_struct: {
+    //     string: "value" }}
+    //
+    // Example 2: 'flatten nested_struct'
```
acarbonetto, line 509: did you mean: `flatten struct .nested_struct`?

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,118 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    // { struct: {
+    //     integer: 0,
+    //       nested_struct: {
+    //         string: "value" }}}
+    //
+    // Example 1: 'flatten struct'
+    //
+    // { integer: 0,
+    //   nested_struct: {
+    //     string: "value" }}
```
acarbonetto, line 507: Surround using `[]`: 
```
    // [
    //   integer: 0,
    //   nested_struct: {string: "value" }
    // ]
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,118 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    //
+    // Example 2: 'flatten nested_struct'
+    //
+    // { struct: {
```
acarbonetto, line 511:
```
    // [
    //   struct: {
    //     integer: 0,
    //     string: "value" 
    //   }
    // ]
```

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,65 @@
...
+.. rubric:: Table of contents
+
+.. contents::
+   :local:
+   :depth: 2
+
+Description
+============
+
+Using ``flatten`` command to flatten an `object`. New fields are added to the search results corresponding
```
acarbonetto, line 14:
```suggestion
The ``flatten`` command flattens an `object`'s fields. New fields are added to the search results corresponding
```

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,65 @@
...
+to each of the object's fields, while the object field itself is removed from the search results.
```
acarbonetto, line 15: > the object field itself is removed from the search results

Is this necessary? Is this what the Spark command does?  The SPL2 splunk command does not remove the existing field from the results. 

Furthermore, the `fields` can be used to filter out unwanted (ie the flattened) fields.  Splunk example: https://docs.splunk.com/Documentation/SCS/current/SearchReference/flattencommandexamples#4._Removing_unwanted_fields_in_the_output

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Syntax
+============
+
+`flatten field`
```
acarbonetto, line 20:
```suggestion
`flatten <field>`
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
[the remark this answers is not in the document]
currantw: Moved into `visitFlatten`. Thanks Andy.

core/src/test/java/org/opensearch/sql/planner/DefaultImplementorTest.java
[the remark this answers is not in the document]
currantw: Updated to match pattern.

integ-test/src/test/java/org/opensearch/sql/ppl/FlattenCommandIT.java
currantw: Agreed. Good idea. I have updated `FlattenOperator` to handle the case that the struct to flatten is `null` or `missing`,  _and_ the case where one of the struct _along the path_ to the target struct is `null` or `missing`. Test updates make to `FlattenOperatorsTests`, `flatten.rst` (doc test), and `FlattenCommandIT` to test this case. Let me know if you think I've missed anything!

doctest/test_mapping/cities.json
currantw: > I know you like city planning

Very true! 😆 

I initially tried to use the Game of Thrones dataset, but I ran into a few issues:
* The objects don't have a `struct` data type type (the data type is "undefined")
* They were missing some test cases that I needed (i.e. struct nested inside a struct)

I also considered using the `bridges` data that is part of the issue on GitHub, but it is also missingthe struct-in-struct case. 

I thought it would be easier to create a new data, rather than extend and modify and esisting one, but I'm happy to do that if you think it's preferable to adding a new data set. 🤷

doctest/test_data/cities.json
currantw: I assume the query here was trying to flatten `location`? Is this expected to support multi-values? I thought we weren't going to support arrays here? (Not totally sure what the different is between multi-values and arrays, if any?)

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Yup, thanks. Updated.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Sure. Updated to do so.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Thanks. Updated.

docs/user/ppl/cmd/flatten.rst
currantw: Thanks. That's better. Done.

docs/user/ppl/cmd/flatten.rst
currantw: Done. 👍

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Override
+  public ExprValue next() {
+    return flattenExprValueAtPath(input.next(), field.getAttr());
+  }
+
+  /**
+   * Flattens the {@link ExprTupleValue} at the specified path and returns the update value.. If the
+   * value is null or missing, the unmodified value is returned..
```
acarbonetto:
```suggestion
   * value is null or missing, the unmodified value is returned.
```

doctest/test_data/cities.json
acarbonetto: ok - nevermind.  the behaviour wasn't defined in the proposal, and I think its fine to continue with the exception until someone requests something otherwise.

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
currantw: As discussed, I will keep the original struct, and raise an issue in Spark to implement the same behaviour.

core/src/main/java/org/opensearch/sql/data/model/ExprTupleValue.java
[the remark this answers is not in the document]
currantw: Good call. Done.

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
currantw: Done!

doctest/test_mapping/cities.json
currantw: Discussed. Going to leave the new dataset, since it contains some test cases that are not part of the Game of Thrones dataset (or others).

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Thanks. Updated, and added test case to `FlattenOperatorTest`.

docs/user/ppl/cmd/flatten.rst
YANG-DB, line 1: Please review the ppl [spark flatten command](https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/ppl-flatten-command.md?plain=1) to match syntax and parameters

YANG-DB (changes requested):
see spark flatten command https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/ppl-flatten-command.md?plain=1
and match

docs/user/ppl/cmd/flatten.rst
currantw: Updated so that the original field is not removed. Raised corresponding issue in OpenSearch Spark to make the behaviour consistent: https://github.com/opensearch-project/opensearch-spark/issues/1029

docs/user/ppl/cmd/flatten.rst
currantw: The OpenSearch Spark `flatten` command removes the original field (i.e. the field being flattened) from the output. This isn't consistent with Splunk. I have raised [this issue](https://github.com/opensearch-project/opensearch-spark/issues/1029) on the OpenSearch SQL to make it consistent with Splunk. Please let me know if you think we should do the opposite (i.e. update OpenSearch Spark to remove the field from the output), even if that is not consistent with Splunk.

currantw: Second, OpenSearch Spark supports of "flattening" a struct _or_ an array, which is inconsistent with Splunk - in Splunk, the user needs to use `expand` for an array and `flatten` for a struct/map. @acarbonetto has raised [this issue](https://github.com/opensearch-project/opensearch-spark/issues/1024) in OpenSearch Spark to update that project to be consistent with Splunk. The intent of _this pull request_ was to match Splunk (i.e. do not support expanding arrays) and the intent of that Spark issue, and to only support `flatten` for struct/map objects. Support for arrays would be handled by the `expand` command issue: https://github.com/opensearch-project/sql/issues/3016

currantw: Lastly, my plan was to add support for aliasing in a separate issue.

docs/user/ppl/cmd/flatten.rst
currantw: Replicates discussion in [this thread](https://github.com/opensearch-project/sql/pull/3267#discussion_r1939770658). Resolving thread...

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,100 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    ReferenceExpression fieldExpr =
+        (ReferenceExpression) expressionAnalyzer.analyze(node.getField(), context);
+    String fieldName = fieldExpr.getAttr();
+
+    // [A] Determine fields to add
+    // ---------------------------
+
+    // Iterate over all the fields defined in the type environment. Find all those that are
+    // descended from field that is being flattened. Determine the new path to add and remove the
+    // existing path. When determining the new path, we need to preserve the portion of the
```
acarbonetto, line 473: I don't understand the use case for removing the existing path.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,100 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    // path corresponding to the flattened field's parent, if one exists, in order to support
+    // flattening nested structs - see example below.
+    //
+    // Input Data:
```
acarbonetto, line 477: move the example to the javadoc

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,100 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+
+    Map<String, ExprType> addFieldsMap = new HashMap<>();
+
+    TypeEnvironment env = context.peek();
+    Map<String, ExprType> fieldsMap = env.lookupAllTupleFields(Namespace.FIELD_NAME);
+
+    final String pathSeparator = ".";
+    final String fieldDescendantPath = fieldName + pathSeparator;
+    final Optional<String> fieldParentPath =
+        fieldName.contains(pathSeparator)
```
acarbonetto, line 510: should fieldName always be at the root?  For example, if we `flatten struct.nested_struct` and then there's a struct that contains `anotherobject.struct.nested_struct` this would match.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,100 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
     return new LogicalEval(child, expressionsBuilder.build());
   }
 
+  /**
+   * Builds and returns a {@link org.opensearch.sql.planner.logical.LogicalFlatten} corresponding to
+   * the given flatten node.
```
acarbonetto, line 458: mention: updates current environment

core/src/main/java/org/opensearch/sql/ast/tree/Flatten.java
```diff
@@ -0,0 +1,40 @@
...
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+import org.opensearch.sql.ast.AbstractNodeVisitor;
+import org.opensearch.sql.ast.expression.Field;
+
+/** AST node representing a flatten <field> operation. */
+@Getter
```
acarbonetto, line 17: needed?

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
```diff
@@ -0,0 +1,99 @@
...
+
+    Matcher matcher = PATH_SEPARATOR_PATTERN.matcher(path);
+    Map<String, ExprValue> exprValueMap = ExprValueUtils.getTupleValue(exprValue);
+
+    // [A] Flatten nested struct value
+    // -------------------------------
+
+    if (matcher.find()) {
+      String currentPathComponent = path.substring(0, matcher.start());
+      String remainingPath = path.substring(matcher.end());
```
acarbonetto, line 68: make sure this works with multiple periods in the path: `root.first.second`

core/src/test/java/org/opensearch/sql/planner/logical/LogicalFlattenTest.java
```diff
@@ -0,0 +1,128 @@
...
+    assertTypeDefined("double", DOUBLE);
+    assertTypeDefined("nested_struct", STRUCT);
+    assertTypeDefined("nested_struct.string", STRING);
+  }
+
+  @Test
+  void testStructNested() {
+    LogicalPlan expectedLogicalPlan =
+        LogicalPlanDSL.flatten(
+            LogicalPlanDSL.relation(TABLE_NAME, table), DSL.ref("struct.nested_struct", STRUCT));
```
acarbonetto, line 84: test with multiple periods: `struct.nested.alsonested`

core/src/test/java/org/opensearch/sql/planner/logical/LogicalFlattenTest.java
```diff
@@ -0,0 +1,128 @@
...
+    assertEquals(expectedLogicalPlan, actualLogicalPlan);
+
+    assertTypeDefined("struct", STRUCT);
+    assertTypeDefined("struct.integer", INTEGER);
+    assertTypeDefined("struct.double", DOUBLE);
+    assertTypeDefined("struct.string", STRING);
+  }
+
+  @Test
+  void testInvalidName() {
```
acarbonetto, line 97: test with period at the start and end of the path: 
`.invalid`
`invalid.`

core/src/test/java/org/opensearch/sql/planner/physical/FlattenOperatorTest.java
```diff
@@ -0,0 +1,163 @@
...
+                Map.entry("double", doubleExprValue),
+                Map.entry("struct_nested", structNestedExprValue)));
+
+    assertEquals(expected, actual);
+  }
+
+  @Test
+  void testStructNested() {
+    ExprValue actual =
+        execute(flatten(inputPlan, DSL.ref("struct.struct_nested", STRUCT))).getFirst();
```
acarbonetto, line 115: test with two levels: `struct.nested.leaf`

acarbonetto (commented):
a couple more tests to add

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: As discussed, I forgot to update the documentation in this method. Updated! ✅

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Done! ✅

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Done! ✅

core/src/main/java/org/opensearch/sql/ast/tree/Flatten.java
currantw: Nope, you're right. Only needed for `field`. Updated accordingly.

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
currantw: Added test case `testStructNestedDeep` in `FlattenOperatorTest` (path is `"struct.struct_nested.struct_nested_deep"`) to verify this.

core/src/test/java/org/opensearch/sql/planner/physical/FlattenOperatorTest.java
currantw: Added ✅

core/src/test/java/org/opensearch/sql/planner/logical/LogicalFlattenTest.java
currantw: Added to `testInvalidName` in `LogicalFlattenTest`.

core/src/test/java/org/opensearch/sql/planner/logical/LogicalFlattenTest.java
currantw: Added `testStructNestedDeep` to `LogicalFlattenTest`.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
acarbonetto: oh, I didn't realize that you were doing another check for this later.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +453,112 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    // descended from field that is being flattened, and determine the new paths to add. When
+    // determining the new paths, we need to preserve the portion of the path corresponding to the
+    // flattened field's parent, if one exists, in order to support flattening nested structs.
+
+    Map<String, ExprType> addFieldsMap = new HashMap<>();
+
+    TypeEnvironment env = context.peek();
+    Map<String, ExprType> fieldsMap = env.lookupAllTupleFields(Namespace.FIELD_NAME);
+
+    final String pathSeparator = ".";
```
acarbonetto, line 519: This is defined in our grammar/lexer, but we don't reuse that symbol :( 

Can you copy this?  https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/utils/ExpressionUtils.java#L17
or better yet, import this:
https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/analysis/DataSourceSchemaIdentifierNameResolver.java#L24

currantw: What I ended up doing was extracting a `PathUtils` class with `SEPARATOR` and `SEPARATOR_PATTERN` static fields. This would also help integrate with the changes I want to make as part of the `expand` command. Does that work?

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +454,111 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    TypeEnvironment env = context.peek();
+    Map<String, ExprType> fieldsMap = env.lookupAllTupleFields(Namespace.FIELD_NAME);
+
+    final String fieldDescendantPath = fieldName + PathUtils.SEPARATOR;
+    final String fieldParentPath =
+        fieldName.contains(PathUtils.SEPARATOR)
+            ? fieldName.substring(0, fieldName.lastIndexOf(PathUtils.SEPARATOR))
+            : "";
+
+    for (String path : fieldsMap.keySet()) {
```
acarbonetto, line 526: consider changing this to `fieldsMap.entrySet()`
Then you get a list of entries (pairs) of key-values, and you don't need to perform the `fieldsMap.get` operation on line 539.

acarbonetto: Also consider doing a `stream.filter()` for all paths that start with `fieldDescendantPath` and you can remove the `continue` call on link 530.  
Slightly cleaner to avoid an extra branch and `continue` call.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +454,111 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+      String name = entry.getKey();
+      ExprType type = entry.getValue();
+
+      // Verify that new field does not overwrite an existing field.
+      if (fieldsMap.containsKey(name)) {
+        throw new IllegalArgumentException(
+            StringUtils.format("Flatten command cannot overwrite field '%s'", name));
+      }
+
+      env.define(DSL.ref(name, type));
```
acarbonetto, line 556: you don't need to save entry.getValue() to a variable that is only used once. 
and then you aren't needlessly accessing the variable if it throws
```suggestion
      env.define(DSL.ref(name, entry.getValue()));
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +454,111 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+
+      // Verify that the path is descended from the flattened field.
+      if (!path.startsWith(fieldDescendantPath)) {
+        continue;
+      }
+
+      // Build the new path.
+      String newPath = path.substring(fieldDescendantPath.length());
+      if (!fieldParentPath.isEmpty()) {
+        newPath = String.join(PathUtils.SEPARATOR, fieldParentPath, newPath);
```
acarbonetto, line 536:
```suggestion
        newPath = fieldParentPath + PathUtils.SEPARATOR + newPath;
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +454,111 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+
+    Map<String, ExprType> addFieldsMap = new HashMap<>();
+
+    TypeEnvironment env = context.peek();
+    Map<String, ExprType> fieldsMap = env.lookupAllTupleFields(Namespace.FIELD_NAME);
+
+    final String fieldDescendantPath = fieldName + PathUtils.SEPARATOR;
+    final String fieldParentPath =
+        fieldName.contains(PathUtils.SEPARATOR)
+            ? fieldName.substring(0, fieldName.lastIndexOf(PathUtils.SEPARATOR))
```
acarbonetto, line 523:
```suggestion
            ? fieldName.substring(0, fieldName.lastIndexOf(PathUtils.SEPARATOR)) + PathUtils.SEPARATOR
```

acarbonetto: If you add the SEPERATOR here, on line 534, you can always call 
```
newPath = fieldParentPath + newPath;
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +454,111 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    // --------------------------------------
+
+    for (Map.Entry<String, ExprType> entry : addFieldsMap.entrySet()) {
+      String name = entry.getKey();
+      ExprType type = entry.getValue();
+
+      // Verify that new field does not overwrite an existing field.
+      if (fieldsMap.containsKey(name)) {
+        throw new IllegalArgumentException(
+            StringUtils.format("Flatten command cannot overwrite field '%s'", name));
```
acarbonetto, line 553: this only shows the first field that overlaps, consider doing this first: 
```
String overlappingFieldNames = String.join(",", 
    addFieldsMap.keySet().stream().filter(name -> fieldsMap.containsKey(name)).collect(Collection.toList());
if (!overlappingFieldNames.isEmpty()) {
   throw new SemanticCheckException(..., overlappingFieldNames);
}
```

Then you can process the fields without concern: 
```
addFieldsMap.entrySet().stream().foreach(e -> env.define(DSL.ref(e.getKey(), e.getValue())));
```

core/src/main/java/org/opensearch/sql/analysis/DataSourceSchemaIdentifierNameResolver.java
```diff
@@ -21,8 +22,6 @@ public class DataSourceSchemaIdentifierNameResolver {
   private final String identifierName;
   private final DataSourceService dataSourceService;
 
-  private static final String DOT = ".";
```
acarbonetto, line 24: ❤️

core/src/main/java/org/opensearch/sql/analysis/DataSourceSchemaIdentifierNameResolver.java
```diff
@@ -35,7 +34,7 @@ public DataSourceSchemaIdentifierNameResolver(
       DataSourceService dataSourceService, List<String> parts) {
     this.dataSourceService = dataSourceService;
     List<String> remainingParts = captureSchemaName(captureDataSourceName(parts));
-    identifierName = String.join(DOT, remainingParts);
+    identifierName = String.join(PathUtils.SEPARATOR, remainingParts);
```
acarbonetto, line 37: you can static import the SEPARATOR
```suggestion
    identifierName = String.join(SEPARATOR, remainingParts);
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +454,111 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+    // descended from field that is being flattened, and determine the new paths to add. When
+    // determining the new paths, we need to preserve the portion of the path corresponding to the
+    // flattened field's parent, if one exists, in order to support flattening nested structs.
+
+    Map<String, ExprType> addFieldsMap = new HashMap<>();
+
+    TypeEnvironment env = context.peek();
+    Map<String, ExprType> fieldsMap = env.lookupAllTupleFields(Namespace.FIELD_NAME);
+
+    final String fieldDescendantPath = fieldName + PathUtils.SEPARATOR;
```
acarbonetto, line 520: you should be able to import the static constant
```suggestion
    final String fieldDescendantPath = fieldName + SEPARATOR;
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +454,111 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+
+    // Iterate over all the fields defined in the type environment. Find all those that are
+    // descended from field that is being flattened, and determine the new paths to add. When
+    // determining the new paths, we need to preserve the portion of the path corresponding to the
+    // flattened field's parent, if one exists, in order to support flattening nested structs.
+
+    Map<String, ExprType> addFieldsMap = new HashMap<>();
+
+    TypeEnvironment env = context.peek();
+    Map<String, ExprType> fieldsMap = env.lookupAllTupleFields(Namespace.FIELD_NAME);
```
acarbonetto, line 518: static import for the constant
```suggestion
    Map<String, ExprType> fieldsMap = env.lookupAllTupleFields(FIELD_NAME);
```

core/src/main/java/org/opensearch/sql/expression/ReferenceExpression.java
```diff
@@ -17,6 +15,7 @@
 import org.opensearch.sql.data.type.ExprCoreType;
 import org.opensearch.sql.data.type.ExprType;
 import org.opensearch.sql.expression.env.Environment;
+import org.opensearch.sql.utils.PathUtils;
```
acarbonetto, line 18:
```suggestion
import static org.opensearch.sql.utils.PathUtils.SEPARATOR;
```

core/src/main/java/org/opensearch/sql/expression/ReferenceExpression.java
```diff
@@ -106,7 +105,7 @@ public ExprValue resolve(ExprTupleValue value) {
   }
 
   private ExprValue resolve(ExprValue value, List<String> paths) {
-    ExprValue wholePathValue = value.keyValue(String.join(PATH_SEP, paths));
+    ExprValue wholePathValue = value.keyValue(String.join(PathUtils.SEPARATOR, paths));
```
acarbonetto, line 108:
```suggestion
    ExprValue wholePathValue = value.keyValue(String.join(SEPARATOR, paths));
```

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Flattens the {@link ExprTupleValue} at the specified path and returns the update value. If the
+   * value is null or missing, the unmodified value is returned.
+   */
+  private static ExprValue flattenExprValueAtPath(ExprValue exprValue, String path) {
+
+    Matcher matcher = PathUtils.SEPARATOR_PATTERN.matcher(path);
+    Map<String, ExprValue> exprValueMap = ExprValueUtils.getTupleValue(exprValue);
```
acarbonetto, line 59: I'm a little concerned that exprValue might not be an Object.

acarbonetto: the description says we return exprValue if its null or missing

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
```diff
@@ -0,0 +1,97 @@
...
+
+    // [A] Flatten nested struct value
+    // -------------------------------
+
+    if (matcher.find()) {
```
acarbonetto, line 64: feels like we're duplicating a lot of checks for root-level and nested call.  Can we just update path if the matcher.find() is true?  

```
    if (matcher.find()) {
      path = path.substring(0, matcher.start());
    }
```

core/src/main/java/org/opensearch/sql/utils/ExpressionUtils.java
```diff
@@ -14,8 +14,6 @@
 @UtilityClass
 public class ExpressionUtils {
 
-  public static final String PATH_SEP = ".";
```
acarbonetto, line 17: not used anywhere?

core/src/main/java/org/opensearch/sql/utils/PathUtils.java
```diff
@@ -0,0 +1,19 @@
...
+package org.opensearch.sql.utils;
+
+import java.util.regex.Pattern;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class PathUtils {
+
+  /** Path separator string */
+  public final String SEPARATOR = ".";
```
acarbonetto, line 15: maybe we can add these to the OperatorUtils instead of creating a new class for two constants?

core/src/test/java/org/opensearch/sql/expression/ip/IPFunctionsTest.java
```diff
@@ -24,8 +26,9 @@
 import org.opensearch.sql.expression.FunctionExpression;
 import org.opensearch.sql.expression.env.Environment;
 
+@ToString
```
acarbonetto, line 29: ??

core/src/test/java/org/opensearch/sql/expression/ip/IPFunctionsTest.java
```diff
@@ -48,19 +51,27 @@ public class IPFunctionsTest {
...
+            SemanticCheckException.class,
+            () -> execute(IPv4AddressWithin, ExprValueUtils.stringValue("INVALID")));
+    assertTrue(
+        ex.getMessage()
+            .matches("^IP address range string 'INVALID' is not valid. Error details: .*"));
   }
 
   @Test
-  public void cidrmatch_valid_arguments() {
+  void cidrmatch_valid_arguments() {
```
acarbonetto, line 74: intellij suggests removing the `public`?

core/src/test/java/org/opensearch/sql/planner/logical/LogicalFlattenTest.java
```diff
@@ -0,0 +1,148 @@
...
+    assertTypeDefined("struct.struct_nested.boolean", BOOLEAN);
+  }
+
+  @Test
+  void testInvalidName() {
+    Exception ex;
+
+    ex = assertThrows(SemanticCheckException.class, () -> executeFlatten("invalid"));
+    assertEquals(
+        "can't resolve Symbol(namespace=FIELD_NAME, name=invalid) in type env", ex.getMessage());
```
acarbonetto, line 109: it's not necessary to test the message content of thrown errors - at least not exactly.  Messages change frequently, and we don't want to have to modify a bunch of tests each time.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
acarbonetto: also, I think this should be a `SemanticCheckException`

core/src/test/java/org/opensearch/sql/planner/logical/LogicalFlattenTest.java
```diff
@@ -0,0 +1,148 @@
...
+    assertEquals(
+        "can't resolve Symbol(namespace=FIELD_NAME, name=.invalid) in type env", ex.getMessage());
+
+    ex = assertThrows(SemanticCheckException.class, () -> executeFlatten("invalid."));
+    assertEquals(
+        "can't resolve Symbol(namespace=FIELD_NAME, name=invalid.) in type env", ex.getMessage());
+  }
+
+  @Test
+  void testInvalidDuplicate() {
```
acarbonetto, line 121: test with more than one duplicate

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: > consider changing this to fieldsMap.entrySet()

Sure. Done. ✅

currantw: > Also consider doing a stream.filter() for all paths that start with fieldDescendantPath and you can remove the continue call on link 530.
Slightly cleaner to avoid an extra branch and continue call.

Sure. Done ✅ 

That being said, it's not clear to me why this is preferable - perhaps I'm just not used to Java streams? Is this mostly a preference to avoid `continue`?

We are replacing one loop through the values with two loops through them (although there will obviously be fewer values in the second loop). We are also creating an additional intermediary object (the list of filtered values). Won't this be both slower and less memory efficient?

We are also replacing two lines of code (the `if` and `continue` lines) with (based on how Spotless formats it) four lines of code. The expression also seems much more complex: one method call (to `startsWith`), vs. additional calls to `entrySet`, `stream`, `filter`, and `toList`. 

Perhaps I am just too used to Perl (oh, how I miss prefix `next`!), but this seems more difficult to read, at least to me? Perhaps there is a simpler way to implement than what I've done?

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Done! ✅

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Done ✅

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Sure. Also renamed to `fieldParentPathPrefix` for clarity.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Done. ✅ 

Also added a test to `LogicalFlattenTests` for the case where there are two duplicates.

core/src/main/java/org/opensearch/sql/analysis/DataSourceSchemaIdentifierNameResolver.java
currantw: Done ✅

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Done. ✅

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Done. ✅

core/src/main/java/org/opensearch/sql/expression/ReferenceExpression.java
currantw: Done ✅

core/src/main/java/org/opensearch/sql/expression/ReferenceExpression.java
currantw: Done ✅

core/src/main/java/org/opensearch/sql/utils/ExpressionUtils.java
currantw: It was used elsewhere. Those were updated to use `PathUtils.SEPARATOR`.

core/src/test/java/org/opensearch/sql/expression/ip/IPFunctionsTest.java
currantw: Removed ✅

core/src/test/java/org/opensearch/sql/expression/ip/IPFunctionsTest.java
currantw: Yes. Here is the message:

image

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
acarbonetto: What if it's an array, String, or integer?

core/src/main/java/org/opensearch/sql/utils/PathUtils.java
currantw: My intent was to add more functionality to this class as part of the `expand` command (check whether a path exists in an `ExprValue`, get a value at the specified path within a root `ExprValue).

core/src/test/java/org/opensearch/sql/planner/logical/LogicalFlattenTest.java
currantw: Hmm. I can see that argument when the message is coming from elsewhere (such as here).

What about cases like `testInvalidDuplicate`, where the class we are testing is the one that throws the exception? In that case, is there value in checking that the error message that we've implemented is "correct" (i.e. consistent with our expectation). For example, does it list _all_ of the duplicate field names? If we only test that an error is thrown, not _which_ error message it contains, how would we differentiate between the case where there is one duplicate field and when there are two?

Even when the exception does come from elsewhere, how to we check if the _right_  exception is being thrown? Consider the duplicate field test. If I had written the test incorrectly without defining the `duplicate` field in the type environment, it would throw a `SemanticCheckException` because of an undefined symbol (e.g. "can't resolve Symbol(namespace=FIELD_NAME, name=duplicate) in type env". If I was only checking the exception _type_, this test would pass, even though it wasn't actually testing the duplicate name behaviour.

Or is this just an argument for having more specific `Exception` classes, and the real problem here is that we are using `SemanticCheckException` for pretty much everything?

(Definitely agree that we don't want our tests to depend on Exception messages from _external_ libraries. I had a comment along those lines on one of Andy's PRs recently.)

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: As discussed, I've moved these constants to `ExprValueUtils` and renamed them to `QUALIFIED_NAME_SEPARATOR` and `QUALIFIED_NAME_SEPARATOR_PATTERN`.

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
currantw: Addressed based on discussion. Added to Javadoc to clarify that `rootExprValue` is expected to be an `ExprValueTuple`.

core/src/main/java/org/opensearch/sql/utils/PathUtils.java
currantw: As discussed, moved to `ExprValueUtils`.

core/src/test/java/org/opensearch/sql/planner/logical/LogicalFlattenTest.java
currantw: Discussed. Checking that error messages _from this class_ contain field name, but not testing other parts of the error msg.

core/src/main/java/org/opensearch/sql/planner/physical/FlattenOperator.java
currantw: Updated to remove duplication!

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +449,107 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+   *     integer: 0,
+   *     nested_struct: { string: "value" }
+   *   }
+   * ]
+   * </pre>
+   *
+   * Query 1: <code>flatten struct</code>
+   *
+   * <pre>
+   * [
```
acarbonetto, line 472: Change from array outer structure to object
```suggestion
   * {
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +449,107 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+   *   struct: {
+   *     integer: 0,
+   *     nested_struct: { string: "value" }
+   *   },
+   *   integer: 0,
+   *   nested_struct: { string: "value" }
+   * ]
```
acarbonetto, line 479:
```suggestion
   * }
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +449,107 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+   * </pre>
+   *
+   * Query 2: <code>flatten struct.nested_struct</code>
+   *
+   * <pre>
+   * [
```
acarbonetto, line 485:
```suggestion
   * {
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +449,107 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+   *   struct: {
+   *     integer: 0,
+   *     nested_struct: { string: "value" },
+   *     string: "value"
+   *   }
+   * ]
```
acarbonetto, line 491:
```suggestion
   * }
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +449,107 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+  /**
+   * Builds and returns a {@link org.opensearch.sql.planner.logical.LogicalFlatten} corresponding to
+   * the given flatten node, and adds the new fields to the current type environment.
+   *
+   * <p><b>Example</b>
+   *
+   * <p>Input Data:
+   *
+   * <pre>
+   * [
```
acarbonetto, line 461:
```suggestion
   * {
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -448,6 +449,107 @@ public LogicalPlan visitEval(Eval node, AnalysisContext context) {
...
+   *   struct: {
+   *     integer: 0,
+   *     nested_struct: { string: "value" }
+   *   }
+   * ]
```
acarbonetto, line 466:
```suggestion
   * }
```

acarbonetto (approved):

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Updated!

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Updated!

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Updated!

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Updated!

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
currantw: Updated!

acarbonetto (approved):

LantaoJin:
resolved by https://github.com/opensearch-project/sql/pull/3747

Issue #3214, opened 2024-12-19
[FEATURE] Add `json_set` function to PPL


acarbonetto:
**Is your feature request related to a problem?**

As part of the RFC to add JSON functions, the `json_set` function would be useful to set or update elements of an existing json_object. 

* META RFC: #3027
* PPL RFC: #3028
* Related Spark PPL PR: https://github.com/opensearch-project/opensearch-spark/pull/780

**What solution would you like?**

```
### `JSON_SET`

**Description**

`json_set(doc, path, value [, path, value]...)` Updates existing or adds new json elements to a json document at the specified paths. Returns the updated document. 

**Arguments type:** JSON_OBJECT, STRING, STRING | JSON_OBJECT | JSON_ARRAY, ...

**Return type:** JSON_OBJECT

Example:

    os> source=people | eval updated = json_set('{"a":[{"b":1},{"b":2}]}', '$.a[*].b', 3) | head 1 | fields updated 
    fetched rows / total rows = 1/1
    +-------------------------+
    | updated                 |
    +-------------------------+
    | {"a":[{"b":3},{"b":3}]} |
    +-------------------------+
```

**What alternatives have you considered?**

N/A

**Do you have any additional context?**

LantaoJin:
Resolved by https://github.com/opensearch-project/sql/pull/3559 with Calcite implementation.

Issue #3028, opened 2024-09-16
[FEATURE]Add PPL `JSON` extended functions support


YANG-DB:
**Is your feature request related to a problem?**
We need to enhance OpenSearch PPL (Piped Processing Language) by adding support for JSON-based operations. This extension will provide users with powerful tools to manipulate and analyze JSON data within PPL queries.

## Proposed New Functions

1. `json_object()`: Creates a JSON object from existing key/value pairs. 
```sql
   ... | eval person = json_object("name", "John", "age", 30)
```
2. `json()`: Evaluates whether a value can be parsed as JSON. Returns the value if valid, null otherwise.
```sql
   ... | eval result = json('{"name":"John","age":30}')
```
3. `json_append()`: Appends elements to the contents of a valid JSON object.
```sql
   ... | eval person = json_object("name", "John"), extended_person = json_append(person, "city", "New York")
```
4. `json_array()`: Creates a JSON array using a list of values.
```sql
   ... | eval json_list = json_array("apple", "banana", "cherry")
```
5. `json_array_to_mv()`: Maps the elements of a JSON array to a multivalued field.
```sql
   ... | eval json_list = json_array("apple", "banana", "cherry"), mv_field = json_array_to_mv(json_list)
```
6. `json_delete()`: Removes one or more keys and their corresponding values from the specified JSON object.
```sql
   ... | eval person = json_object("name", "John", "age", 30, "city", "New York"), person_no_age = json_delete(person, "age")
```
7. `json_extend()`: Extends the contents of a valid JSON object with the values of an array.
```sql
   ... | eval person = json_object("name", "John", "age", 30), extended_person = json_extend(person, json_array("city", "New York"))
```
8. `json_extract()`: Returns either a JSON array or a native type value from a field and zero or more paths.
```sql
   ... | eval data = json('{"person": {"name": "John", "age": 30}}'), name = json_extract(data, "person.name")
```
10. `json_keys()`: Returns the keys from the key-value pairs in a JSON object as a JSON array.
```sql
   ... | eval person = json_object("name", "John", "age", 30), keys = json_keys(person)
```
11. `json_set()`: Inserts or overwrites values for a JSON node with provided values and returns an updated JSON object.
```sql
   ... | eval person = json_object("name", "John"), updated_person = json_set(person, "age", 30)
```
13. `json_valid()`: Evaluates whether a JSON object uses valid JSON syntax and returns TRUE or FALSE.
```sql
[...]

PR #3271, opened 2025-01-28
Feature: PPL - json_set function

andy-k-improving:
This PR depends on t he `json_valid`, `json_object` and `json_extract`:
https://github.com/opensearch-project/sql/pull/3262, https://github.com/opensearch-project/sql/pull/3242, https://github.com/opensearch-project/sql/pull/3230

### Description
To introduce Json function `json_set` to perform insert || override with provided JsonPath String.

### Related Issues
Resolves: https://github.com/opensearch-project/sql/issues/3214
Partially resolves: https://github.com/opensearch-project/sql/issues/3028

### Check List


YANG-DB:
Hi @andy-k-improving 
is this still in draft mode ? if not can u plz resolve the conflicts ?
thanks

andy-k-improving:
> Hi @andy-k-improving is this still in draft mode ? if not can u plz resolve the conflicts ? thanks

Functionality wide, this is ready.
however my code is based on some other json function as we share some util and Gradle deps.
https://github.com/opensearch-project/sql/pull/3262, https://github.com/opensearch-project/sql/pull/3242, https://github.com/opensearch-project/sql/pull/3230

async-query-core/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -387,7 +387,7 @@ JSON_VALID:                         'JSON_VALID';
 //JSON_APPEND:                        'JSON_APPEND';
 //JSON_DELETE:                        'JSON_DELETE';
 //JSON_EXTEND:                        'JSON_EXTEND';
-//JSON_SET:                           'JSON_SET';
+JSON_SET:                           'JSON_SET';
```
normanj-bitquill, line 390: Looks like the formatting is off.

async-query-core/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -875,7 +875,7 @@ jsonFunctionName
 //   | JSON_APPEND
 //   | JSON_DELETE
 //   | JSON_EXTEND
-//   | JSON_SET
+   | JSON_SET
```
normanj-bitquill, line 878: Formatting

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -0,0 +1,67 @@
...
+    return define(
+        BuiltinFunctionName.JSON_EXTRACT.getName(),
+        impl(nullMissingHandling(JsonUtils::extractJson), UNDEFINED, STRING, STRING));
+  }
+
+
+  private DefaultFunctionResolver jsonSet() {
+    return define(
+            BuiltinFunctionName.JSON_SET.getName(),
+            impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, STRING),
```
normanj-bitquill, line 57: Some types are missing here, such as Byte, Date, and IP.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,205 @@
...
+                // New element in the array.
+                String updatedJson = docContext.add(pathUnquoted, valueUnquoted).jsonString();
+                return new ExprStringValue(updatedJson);
+
+            }
+            default -> {
+                return LITERAL_NULL;
+            }
+        }
+    } catch (InvalidModificationException | InvalidJsonException | InvalidPathException | IllegalArgumentException | UnsupportedOperationException ex) {
```
normanj-bitquill, line 182: This line is too long and should be split.

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Usage: `json_set(json_string, json_path, value)` Perform value insertion or override with provided Json path and value. Returns the updated JSON object if valid, null otherwise.
+
+Argument type: JSON_STRING, STRING, STRING/BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
+
+Return type: JSON_STRING
+
+Example::
+
+    > source=json_test eval updated=json_set(json_string, "$.c.innerProperty", "test_value") | fields test_name, updated
```
normanj-bitquill, line 78: Why can't this query run in the doctests?

andy-k-improving:
@normanj-bitquill Fixed above comments, would you mind to have another look?

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -0,0 +1,71 @@
...
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, LONG),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, SHORT),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, INTEGER),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, DOUBLE),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, BOOLEAN),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, ARRAY),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, STRUCT),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, BYTE),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, DATE),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, IP));
```
currantw, line 69: A few questions/thoughts:
- Looks like you're missing `FLOAT`?
- You're supporting `DATE`. What about `TIME`, `TIMESTAMP`, or `INTERVAL` types?
- What about `UNDEFINED`? Is there anyway to set a JSON value to `null`?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+
+    // in all other cases, return null
+    return LITERAL_NULL;
+  }
+
+  /** Converts a JSON encoded string to an Expression object. */
+  public static ExprValue setJson(ExprValue json, ExprValue path, ExprValue valueToInsert) {
+
+    String jsonUnquoted = StringUtils.unquoteText(json.stringValue());
+    String pathUnquoted = StringUtils.unquoteText(path.stringValue());
```
currantw, line 149: What does this method need to use `unquoteText` but the other JSON methods above don't?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+        Configuration.defaultConfiguration().addOptions(Option.SUPPRESS_EXCEPTIONS);
+    try {
+      JsonPath jsonPath = JsonPath.compile(pathUnquoted);
+      DocumentContext docContext = JsonPath.using(conf).parse(jsonUnquoted);
+      switch (docContext.read(jsonPath)) {
+        case null -> {
+          // Insert a new property
+          recursiveCreate(docContext, pathUnquoted, valueUnquoted);
+          return new ExprStringValue(docContext.jsonString());
+        }
```
currantw, line 161: Why are you using `SUPPRESS_EXCEPTIONS` here?

From the documentation:
> Suppress all exceptions when evaluating path ... If an exception is thrown and the option [ALWAYS_RETURN_LIST](https://javadoc.io/static/com.jayway.jsonpath/json-path/2.2.0/com/jayway/jsonpath/Option.html#ALWAYS_RETURN_LIST) is not present null is returned.

Won't this mean that the first `switch` case (insert a new property) will be executed whenever an invalid path argument is given (i.e. if it isn't even valid JSON path syntax)? Why are we handling and throwing an exception if the path isn't a valid JSON path?

currantw: @kenrickyap current PR for `json_expand` handles `InvalidPathException`. Probably make sense for these two functions (and all other JSON functions) to handle this the same? Would it be possible to extract this logic to a shared method?

```java
private static JsonPath getJsonPath(String jsonPathString) {

  try {
    return JsonPath.compile(jsonPathString);
  } catch () {
      final String errorMsgFormat = "JSON path '%s' is not valid. Error details: %s";
      throw new SemanticCheckException(String.format(errorMsgFormat, jsonPathString, e.getMessage()), e);
  }
}
```

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+      }
+      return ExprTupleValue.fromExprValueMap(values);
+    }
+
+    // in all other cases, return null
+    return LITERAL_NULL;
+  }
+
+  /** Converts a JSON encoded string to an Expression object. */
+  public static ExprValue setJson(ExprValue json, ExprValue path, ExprValue valueToInsert) {
```
currantw, line 146: What about the case where the JSON path points to multiple places? Does this method support inserting multiples values (it doesn't seem like it)? Would it just insert the first one, would it fail? Is this a case we should consider, and have an appropriate error message for it?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+        case String ignored -> {
+          // Override an existing property
+          String updatedJson = docContext.set(pathUnquoted, valueUnquoted).jsonString();
+          return new ExprStringValue(updatedJson);
+        }
+        case Iterable<?> ignored -> {
+          // New element in the array.
+          String updatedJson = docContext.add(pathUnquoted, valueUnquoted).jsonString();
+          return new ExprStringValue(updatedJson);
+        }
```
currantw, line 171: Am I understanding this correctly? If I call `jsonSet` with a JSON path that points to an array, this will _insert_ the given value into the array, rather than set it equal to the given value?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+      switch (docContext.read(jsonPath)) {
+        case null -> {
+          // Insert a new property
+          recursiveCreate(docContext, pathUnquoted, valueUnquoted);
+          return new ExprStringValue(docContext.jsonString());
+        }
+        case String ignored -> {
+          // Override an existing property
+          String updatedJson = docContext.set(pathUnquoted, valueUnquoted).jsonString();
+          return new ExprStringValue(updatedJson);
```
currantw, line 165: Have you tested `set` for the case where the value does not already exist, and confirmed that it fails? I have seen some implementation of similar JSON libraries where `set` would create the specified path, if it did not already exist (the `JsonPath` documentation is a bit unclear on that front, at least to me).

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: What about the case where the JSON path points to a nested JSON object? Is a JSON object `Iterable`, and so would it fall into this case?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+        case Iterable<?> ignored -> {
+          // New element in the array.
+          String updatedJson = docContext.add(pathUnquoted, valueUnquoted).jsonString();
+          return new ExprStringValue(updatedJson);
+        }
+        default -> {
+          return LITERAL_NULL;
+        }
```
currantw, line 174: What are the expected cases where we could end up here?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+      }
+    } catch (InvalidModificationException
+        | InvalidJsonException
+        | InvalidPathException
+        | IllegalArgumentException
+        | UnsupportedOperationException ex) {
+      return LITERAL_NULL;
```
currantw, line 181: Do we really want to just return `null` in all of these cases? If the JSON path is invalid (i.e. invalid syntax), for example, it seems to me like it might be more appropriate to raise an exception?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+  private static void recursiveCreate(DocumentContext docContext, String path, Object value) {
+    final int pos = path.lastIndexOf('.');
+    final String parent = path.substring(0, pos);
+    final String current = path.substring(pos + 1);
+    // Attempt to read the current path as it is, trigger the recursive in case of deep insert.
+    if (docContext.read(parent) == null) {
+      recursiveCreate(docContext, parent, new LinkedHashMap<>());
+    }
+    docContext.put(parent, current, value);
+  }
```
currantw, line 201: As mentioned in another comment, have you verified that this (recursive insertion) is not handled by `WriteContext.set`?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,290 @@
...
+    assertThrows(
+        SemanticCheckException.class, () -> DSL.castJson(DSL.literal("\"missing quote")).valueOf());
+  }
+
+  @Test
+  void json_set_InsertString() {
+    FunctionExpression functionExpression =
+        DSL.jsonSet(DSL.literal("[]"), DSL.literal("$"), DSL.literal("test_value"));
+    assertEquals("[\"test_value\"]", functionExpression.valueOf().stringValue());
+  }
```
currantw, line 172: Might be worth extracting a helper function? Could help make this a big easier to read?

```java

  void json_set_InsertString() {
    assertEquals("[\"test_value\"]", execute_set_json("[]", "$", "test_value"));
  }

static String execute_set_json(String jsonString, String jsonPathString, Object value) {
  ExprValue jsonExprValue = ExprValueUtils.stringValue(jsonString);
  ExprValue jsonPathExprValue = ExprValueUtils.stringValue(jsonPathString);
  ExprValue setExprValue = ExprValueUtils.fromObjectValue(value);

  FunctionExpression functionExpression = DSL.jsonSet(jsonExprValue, jsonPathExprValue, setExprValue);
  return functionExpression.valueOf().stringValue();
}

```

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,89 @@
...
+
+JSON_SET
+----------
+
+Description
+>>>>>>>>>>>
+
+Usage: `json_set(json_string, json_path, value)` Perform value insertion or override with provided Json path and value. Returns the updated JSON object if valid, null otherwise.
+
+Argument type: JSON_STRING, STRING, STRING/BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
```
currantw, line 72: Looks like you need to update this to include a few more supported types? Also, shouldn't this just be `STRING` instead of `JSON_STRING`?

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Return type: JSON_STRING
```
currantw, line 74: Similarly, should this just be `STRING`?

integ-test/src/test/java/org/opensearch/sql/ppl/JsonFunctionsIT.java
```diff
@@ -0,0 +1,125 @@
...
+  @Test
+  public void test_json_set() throws IOException {
+    JSONObject result;
+
+    result =
+        executeQuery(
+            String.format(
+                "source=%s | eval updated=json_set(json_string, \\\"$.c.innerProperty\\\","
+                    + " \\\"test_value\\\") | fields test_name, updated",
+                TEST_INDEX_JSON_TEST));
```
currantw, line 111: I curious: why do we need so many escape characters here? Why doesn't one suffice?

currantw (commented):
Reviewed to `2509fc8`.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: Thats unnecessary, removed.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: `SUPPRESS_EXCEPTIONS` is used to make sure exception won't be thrown in the case of "valid jsonPath, but element is absent". 
Which exception shouldn't be re-throw in this case, because it's expected to have a valid jsonPath but the element is absent.
In this particular case, `recursiveCreate( )` will be called to perform recursive insert. 

Regarding the route of invalid jsonPath, exception won't be suppressed regardless of `SUPPRESS_EXCEPTIONS` option, see `json_set_insert_invalid_path` test-case for the demonstration.

andy-k-improving: Updated the logic to throw `SemanticCheckException` instead, see test-cases: `json_set_insert_invalid_path` and `json_set_insert_invalid_jsonObject` for demonstration.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: Updated the logic to handle multiple matches, see test-cases:
* json_set_multiMatches_property( )
* json_set_multiMatches_array( )

For specific handling for normal Json Object || array.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: You are right, actually the correct behaviour should be overriding the value, instead of adding a entry to the array.

In the case of nested object,`String` will be returned in the case of a property, and `Java collection || Iterable` will be returned in the case of Json array.

See respective test-cases for detail usage: 
* json_set_insert_new_property_nested
* json_set_singleMatch_property
* json_set_singleMatch_array

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: This is being validated under `json_set_insert_new_property_nested`, do you have any particular use-case in mind which the current logic may have missed?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
andy-k-improving: I don't think so, `DSL.jsonSet( )` requires `Expression` instead of `ExprValue`, the third argument won't cater the use-case of primitive type (long instead of Long).

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,435 @@
...
+        DSL.jsonSet(
+            DSL.literal("[]"),
+            DSL.literal("$"),
+            DSL.literal(
+                ExprTupleValue.fromExprValueMap(Map.of("name", new ExprStringValue("alice")))));
+    assertEquals("[{\"name\":\"alice\"}]", functionExpression.valueOf().stringValue());
+  }
+
+  @Test
+  void json_set_InsertArray() {
```
currantw, line 296: What is the difference between this and `json_set_InsertList`?

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
currantw: Looks good. Thanks for also adding corresponding unit tests. 🎉

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Thanks! Could you also update the corresponding variable names? `jsonUnquoted`, `pathUnquoted`, and `valueUnquoted` don't really make sense without the calls to `StringUtils.unquoteText`...?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: I see. Thanks for the explanation!

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Cool! Nice to see. Thanks for looking into that and adding tests for it too!

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+        recursiveCreate(docContext, pathUnquoted, valueUnquoted).jsonString();
+      } else if (readResult instanceof String) {
+        docContext.set(pathUnquoted, valueUnquoted).jsonString();
+      } else {
+        if (((List<?>) readResult).isEmpty()) {
+          docContext.add(pathUnquoted, valueUnquoted).jsonString();
+        } else {
+          docContext.set(pathUnquoted, valueUnquoted).jsonString();
+        }
+      }
```
currantw, line 166: Can reduce nesting here?

```suggestion
      if (readResult == null) {
        recursiveCreate(docContext, pathUnquoted, valueUnquoted).jsonString();
      } else if (readResult instanceof String) {
        docContext.set(pathUnquoted, valueUnquoted).jsonString();
      } else if (((List<?>) readResult).isEmpty()) {
        docContext.add(pathUnquoted, valueUnquoted).jsonString();
      } else {
        docContext.set(pathUnquoted, valueUnquoted).jsonString();
      }
```

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Thanks! I'm still a bit confused about this case:

```java 
if (((List<?>) readResult).isEmpty()) {
  docContext.add(pathUnquoted, valueUnquoted).jsonString();
} else {
  docContext.set(pathUnquoted, valueUnquoted).jsonString();
}
```

Am I understanding this right: if the path points to an _empty_ array, then we will add the value to the array, while if it points to a _non-empty_ array, then we will set it equal to the given value. Why the difference between these two cases?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Overtaken by changes. Please feel free to resolve.

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
currantw: Sure 🤷

docs/user/ppl/functions/json.rst
currantw: Thanks!

docs/user/ppl/functions/json.rst
currantw: Thanks!

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -0,0 +1,81 @@
...
+    return define(
+        BuiltinFunctionName.JSON_EXTRACT.getName(),
+        impl(nullMissingHandling(JsonUtils::extractJson), UNDEFINED, STRING, STRING));
+  }
+
+  private DefaultFunctionResolver jsonSet() {
+    return define(
+        BuiltinFunctionName.JSON_SET.getName(),
+
+        // Numeric types
```
currantw, line 64: Is this comment is accurate anymore..?

docs/user/ppl/functions/json.rst
```diff
@@ -0,0 +1,89 @@
...
+   :local:
+   :depth: 1
+
+JSON_VALID
+----------
+
+Description
+>>>>>>>>>>>
+
+Usage: `json_valid(json_string)` checks if `json_string` is a valid STRING string.
```
currantw, line 17: Typo

```suggestion
Usage: `json_valid(json_string)` checks if `json_string` is a valid string.
```

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -0,0 +1,202 @@
...
+        values.put(entry.getKey(), processJsonNode(entry.getValue()));
+      }
+      return ExprTupleValue.fromExprValueMap(values);
+    }
+
+    // in all other cases, return null
+    return LITERAL_NULL;
+  }
+
+  /** Converts a JSON encoded string to an Expression object. */
```
currantw, line 144: Looks like this Javadoc needs updating?

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: What about the remaining exceptions (`InvalidModificationException`, `IllegalArgumentException`, `UnsupportedOperationException`). Under what circumstances would we expect these to occur? I wasn't able to find test cases for any of these?

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
```diff
@@ -0,0 +1,435 @@
...
+
+  @Test
+  void json_set_insert_invalid_jsonObject() {
+    FunctionExpression functionExpression =
+        DSL.jsonSet(DSL.literal("[xxxx}}}}}"), DSL.literal("$"), DSL.literal("18"));
+    assertThrows(SemanticCheckException.class, () -> functionExpression.valueOf());
+  }
+
+  @Test
+  void json_set_noMatch_property() {
```
currantw, line 337: Does this duplicate `json_set_insert_new_property_nested` or am I missing a subtle difference between them?

currantw (commented):
Reviewed to `24eec27`. Sorry about all the duplicate comments! GitHub seems to do that for some reason when I comment on an existing thread as part of a review. Guess I'll just stick with single comments from here on... 😬

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: Yes, that is because I have the wrong test-case at first place.
It turns out this piece of information is not related `((List<?>) readResult).isEmpty()` and in the case of readResult being `non-null`.
The code logic should ahead to override || replace the value supplied data , regardless of the type. 

I have simplified the logic to reflect above, can current structure looks like the following:
```
      if (readResult == null) {
        recursiveCreate(docContext, jsonPathString, valueToInsertObj);
      } else {
        docContext.set(jsonPathString, valueToInsertObj);
      }
      return 
```
It either do insert or update, with no alternative.

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
andy-k-improving: Removed, it's duplicated.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: Those are originally setup for use-case to add element into an existing array, and since updating an existing array should never happen, removed. these catch clause also.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: Done.

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
andy-k-improving: Removed.

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
andy-k-improving: Yes, removed `json_set_insert_new_property_nested` instead and keep this.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
andy-k-improving: Done.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Great, looks good! 🎉

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Overtaken by changes. Please feel free to resolve.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Overtaken by changes/other threads. Please feel free to resolve.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Addressed. Thanks. Please resolve.

core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
currantw: Thanks!

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Thanks. Please resolve.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Thanks. Looks good to me. Can resolve.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
currantw: Overtaken by changes. Please feel free to resolve.

core/src/main/java/org/opensearch/sql/ast/expression/Cast.java
```diff
@@ -56,6 +57,7 @@ public class Cast extends UnresolvedExpression {
           .put("timestamp", CAST_TO_TIMESTAMP.getName())
           .put("datetime", CAST_TO_DATETIME.getName())
           .put("ip", CAST_TO_IP.getName())
+          .put("json", CAST_TO_JSON.getName())
```
acarbonetto, line 60: can you rebase please?

core/src/main/java/org/opensearch/sql/expression/json/JsonFunctions.java
```diff
@@ -0,0 +1,79 @@
...
+  private DefaultFunctionResolver jsonExtract() {
+    return define(
+        BuiltinFunctionName.JSON_EXTRACT.getName(),
+        impl(nullMissingHandling(JsonUtils::extractJson), UNDEFINED, STRING, STRING));
+  }
+
+  private DefaultFunctionResolver jsonSet() {
+    return define(
+        BuiltinFunctionName.JSON_SET.getName(),
+        impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, BYTE),
```
acarbonetto, line 63: as we're doing in SPARK, can we change this to a single signature: 
```
impl(nullMissingHandling(JsonUtils::setJson), UNDEFINED, STRING, STRING, ARRAY));
```
the array would receive path-value pairs.

core/src/main/java/org/opensearch/sql/utils/JsonUtils.java
```diff
@@ -109,4 +169,60 @@ private static ExprValue processJsonNode(JsonNode jsonNode) {
...
+   * @param valueToInsert value to be added
+   * @return JsonString after the upsert operation.
+   */
+  public static ExprValue setJson(ExprValue json, ExprValue path, ExprValue valueToInsert) {
+
+    String jsonString = json.stringValue();
+    String jsonPathString = path.stringValue();
+    Object valueToInsertObj = valueToInsert.value();
+    Configuration conf =
+        Configuration.defaultConfiguration().addOptions(Option.SUPPRESS_EXCEPTIONS);
```
Yury-Fridlyand, line 188: Why suppress?

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,101 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+
+
+
+JSON_SET
+----------
+
+Description
+>>>>>>>>>>>
+
+Usage: `json_set(json_string, json_path, value)` Perform value insertion or override with provided Json path and value. Returns the updated JSON object if valid, null otherwise.
```
Yury-Fridlyand, line 131: Same comments I posted in #3262 are also applicable there.

docs/user/ppl/functions/json.rst
```diff
@@ -49,14 +50,101 @@ Return type: BOOLEAN/DOUBLE/INTEGER/NULL/STRUCT/ARRAY
...
+
+Argument type: STRING, STRING, BYTE/SHORT/INTEGER/LONG/FLOAT/DOUBLE/STRING/BOOLEAN/DATE/TIME/TIMESTAMP/INTERVAL/IP/STRUCT/ARRAY
```
Yury-Fridlyand, line 133: This becomes unreadable. maybe something like
```suggestion
Argument type: STRING, STRING, <any>
```

LantaoJin:
Resolved by https://github.com/opensearch-project/sql/pull/3559 with Calcite implementation.

Issue #3277, opened 2025-01-29
[FEATURE] PPL trendline command with WMA support


acarbonetto:
Followup to https://github.com/opensearch-project/sql/issues/3013

**Is your feature request related to a problem?**

Expland the new PPL `trendline` command to support computing a weighted moving averages of fields.

****WMA** :  Weighted moving average**

`WMA(t) = Σ(w[i] * f[i]) / Σ(w[i]), where i = t-n+1 to t`
Where w[i] is the weight for the i-th data-point. 

In a typical WMA, the weights are linearly decreasing from the most recent to the oldest data-point:
`w[i] = n - (t - i), where i = t-n+1 to t`

The complete forumlation would be:
`WMA(t) = Σ((n - (t - i)) * f[i]) / Σ(n - (t - i)), where i = t-n+1 to t`

---

**Example**

Multiple trendlines could be calculated in a single command, such as 
```sql
| trendline sma(10,memory) as mem_trend wma(5,network_traffic) as net_trend.
```

LantaoJin:
Resolved by https://github.com/opensearch-project/sql/pull/3741

PR #3293, opened 2025-02-03
Feature: PPL command - WMA Trendline

andy-k-improving:
### Description
This PR introduce a new varient (Weighted Moving Average - WMA) for existing PPL trendline command, with corresponded test-cases and documentation for usage examples.

Reference:
WMA calculation: https://corporatefinanceinstitute.com/resources/career-map/sell-side/capital-markets/weighted-moving-average-wma/
WMA implementation on Spark: https://github.com/opensearch-project/opensearch-spark/pull/872

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3011, https://github.com/opensearch-project/sql/issues/3277

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
+++ b/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
@@ -23,6 +23,7 @@ import lombok.ToString;
 import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.sql.ast.tree.Trendline;
 import org.opensearch.sql.data.model.ExprDoubleValue;
+import org.opensearch.sql.data.model.ExprLongValue;
 import org.opensearch.sql.data.model.ExprTupleValue;
 import org.opensearch.sql.data.model.ExprValue;
 import org.opensearch.sql.data.model.ExprValueUtils;
@@ -193,28 +194,22 @@ public class TrendlineOperator extends PhysicalPlan {
   private static class WeightedMovingAverageAccumulator implements TrendlineAccumulator {
     private final LiteralExpression dataPointsNeeded;
     private final ArrayList<ExprValue> receivedValues;
-//    private final ArithmeticEvaluator evaluator;
+    private final ExprCoreType type;
+
 
     public WeightedMovingAverageAccumulator(
             Trendline.TrendlineComputation computation, ExprCoreType type) {
-      dataPointsNeeded = DSL.literal(computation.getNumberOfDataPoints().doubleValue());
-      receivedValues = new ArrayList<>(computation.getNumberOfDataPoints()+1);
-//      evaluator = TrendlineAccumulator.getEvaluator(type);
+      this.dataPointsNeeded = DSL.literal(computation.getNumberOfDataPoints().doubleValue());
+      this.receivedValues = new ArrayList<>(computation.getNumberOfDataPoints()+1);
+      this.type = type;
     }
 
     @Override
     public void accumulate(ExprValue value) {
-//      if (dataPointsNeeded.valueOf().integerValue() == 1) {
-//        receivedValues.add(value);
-//        return;
-//      }
-
       receivedValues.add(value);
-
       if (receivedValues.size() > dataPointsNeeded.valueOf().integerValue()) {
         receivedValues.removeFirst();
       }
-
     }
 
     @Override
@@ -226,15 +221,49 @@ public class TrendlineOperator extends PhysicalPlan {
       }
       return computeWma(receivedValues);
     }
+
+    /**
+     * Compute WMA values from provided dataset with in ascending order.
+     * @param receivedValues the dataset for WMA calculation, sorted in ascending order.
+     * @return ExprValue which represent he result onf WMA.
+     */
+    private ExprValue computeWma(ArrayList<ExprValue> receivedValues) {
+      if (type == ExprCoreType.DOUBLE) {
+        return new ExprDoubleValue(calculateWmaInDouble(receivedValues));
+
+      } else if (type == ExprCoreType.DATE) {
+        return ExprValueUtils.dateValue(
+                ExprValueUtils.timestampValue(Instant.ofEpochMilli(
+                        calculateWmaInLong(receivedValues))).dateValue());
+
+      } else if ( type == ExprCoreType.TIME) {
+        return ExprValueUtils.timeValue(
+                LocalTime.MIN.plus(calculateWmaInLong(receivedValues), MILLIS));
+
+      } else if (type == ExprCoreType.TIMESTAMP) {
+        return ExprValueUtils.timestampValue(Instant.ofEpochMilli(
+                calculateWmaInLong(receivedValues)));
+      }
+     return null;
+    }
+
+    private double calculateWmaInDouble (ArrayList<ExprValue> receivedValues) {
+      double sum = 0D;
+      for (int i=0 ; i<receivedValues.size() ; i++) {
+        sum += receivedValues.get(i).doubleValue() / (i+1);
+      }
+      return sum / receivedValues.size();
+    }
+
+    private long calculateWmaInLong (ArrayList<ExprValue> receivedValues) {
+      long sum = 0L;
+      for (int i=0 ; i<receivedValues.size() ; i++) {
+        sum += receivedValues.get(i).longValue() / (i+1);
+      }
+      return sum / receivedValues.size();
+    }
   }
 
-  private static ExprValue computeWma(ArrayList<ExprValue> receivedValues) {
-    double sum = 0D;
-    for (int i=0 ; i<receivedValues.size() ; i++) {
-      sum += receivedValues.get(i).doubleValue() / i+1;
-    }
-    return new ExprDoubleValue(sum / receivedValues.size());
-  }
 
 
   private interface ArithmeticEvaluator {
--- a/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
+++ b/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
@@ -11,6 +11,7 @@ import com.google.common.collect.EvictingQueue;
 import com.google.common.collect.ImmutableMap.Builder;
 import java.time.Instant;
 import java.time.LocalTime;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -21,6 +22,7 @@ import lombok.Getter;
 import lombok.ToString;
 import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.sql.ast.tree.Trendline;
+import org.opensearch.sql.data.model.ExprDoubleValue;
 import org.opensearch.sql.data.model.ExprTupleValue;
 import org.opensearch.sql.data.model.ExprValue;
 import org.opensearch.sql.data.model.ExprValueUtils;
@@ -106,9 +108,10 @@ public class TrendlineOperator extends PhysicalPlan {
 
   private static TrendlineAccumulator createAccumulator(
       Pair<Trendline.TrendlineComputation, ExprCoreType> computation) {
-    // Add a switch statement based on computation type to choose the accumulator when more
-    // types of computations are supported.
-    return new SimpleMovingAverageAccumulator(computation.getKey(), computation.getValue());
+      return switch (computation.getKey().getComputationType()) {
+          case SMA -> new SimpleMovingAverageAccumulator(computation.getKey(), computation.getValue());
+          case WMA -> new WeightedMovingAverageAccumulator(computation.getKey(), computation.getValue());
+      };
   }
 
   /** Maintains stateful information for calculating the trendline. */
@@ -186,6 +189,53 @@ public class TrendlineOperator extends PhysicalPlan {
       return evaluator.evaluate(runningTotal, dataPointsNeeded);
     }
   }
+
+  private static class WeightedMovingAverageAccumulator implements TrendlineAccumulator {
+    private final LiteralExpression dataPointsNeeded;
+    private final ArrayList<ExprValue> receivedValues;
+//    private final ArithmeticEvaluator evaluator;
+
+    public WeightedMovingAverageAccumulator(
+            Trendline.TrendlineComputation computation, ExprCoreType type) {
+      dataPointsNeeded = DSL.literal(computation.getNumberOfDataPoints().doubleValue());
+      receivedValues = new ArrayList<>(computation.getNumberOfDataPoints()+1);
+//      evaluator = TrendlineAccumulator.getEvaluator(type);
+    }
+
+    @Override
+    public void accumulate(ExprValue value) {
+//      if (dataPointsNeeded.valueOf().integerValue() == 1) {
+//        receivedValues.add(value);
+//        return;
+//      }
+
+      receivedValues.add(value);
+
+      if (receivedValues.size() > dataPointsNeeded.valueOf().integerValue()) {
+        receivedValues.removeFirst();
+      }
+
+    }
+
+    @Override
+    public ExprValue calculate() {
+      if (receivedValues.size() < dataPointsNeeded.valueOf().integerValue()) {
+        return null;
+      } else if (dataPointsNeeded.valueOf().integerValue() == 1) {
+        return receivedValues.getFirst();
+      }
+      return computeWma(receivedValues);
+    }
+  }
+
+  private static ExprValue computeWma(ArrayList<ExprValue> receivedValues) {
+    double sum = 0D;
+    for (int i=0 ; i<receivedValues.size() ; i++) {
+      sum += receivedValues.get(i).doubleValue() / i+1;
+    }
+    return new ExprDoubleValue(sum / receivedValues.size());
+  }
+
 
   private interface ArithmeticEvaluator {
     Expression calculateFirstTotal(List<ExprValue> dataPoints);
--- a/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
+++ b/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
@@ -234,34 +234,46 @@ public class TrendlineOperator extends PhysicalPlan {
       } else if (type == ExprCoreType.DATE) {
         return ExprValueUtils.dateValue(
                 ExprValueUtils.timestampValue(Instant.ofEpochMilli(
-                        calculateWmaInLong(receivedValues))).dateValue());
+                        calculateWmaInTs(receivedValues))).dateValue());
 
       } else if ( type == ExprCoreType.TIME) {
         return ExprValueUtils.timeValue(
-                LocalTime.MIN.plus(calculateWmaInLong(receivedValues), MILLIS));
+                LocalTime.MIN.plus(calculateWmaInTime(receivedValues), MILLIS));
 
       } else if (type == ExprCoreType.TIMESTAMP) {
         return ExprValueUtils.timestampValue(Instant.ofEpochMilli(
-                calculateWmaInLong(receivedValues)));
+                calculateWmaInTs(receivedValues)));
       }
      return null;
     }
 
     private double calculateWmaInDouble (ArrayList<ExprValue> receivedValues) {
       double sum = 0D;
+      int totalWeight = (receivedValues.size()*(receivedValues.size()+1)) / 2;
       for (int i=0 ; i<receivedValues.size() ; i++) {
-        sum += receivedValues.get(i).doubleValue() / (i+1);
+        sum += receivedValues.get(i).doubleValue() * ((i + 1D) / totalWeight);
       }
-      return sum / receivedValues.size();
+      return sum;
     }
 
-    private long calculateWmaInLong (ArrayList<ExprValue> receivedValues) {
+    private long calculateWmaInTs (ArrayList<ExprValue> receivedValues) {
       long sum = 0L;
+      int totalWeight = (receivedValues.size()*(receivedValues.size()+1)) / 2;
       for (int i=0 ; i<receivedValues.size() ; i++) {
-        sum += receivedValues.get(i).longValue() / (i+1);
+        sum += (long) (receivedValues.get(i).timestampValue().toEpochMilli() * ((i + 1D) / totalWeight));
       }
-      return sum / receivedValues.size();
+      return sum;
     }
+
+    private long calculateWmaInTime (ArrayList<ExprValue> receivedValues) {
+      long sum = 0L;
+      int totalWeight = (receivedValues.size()*(receivedValues.size()+1)) / 2;
+      for (int i=0 ; i<receivedValues.size() ; i++) {
+        sum += (long) (MILLIS.between(LocalTime.MIN, receivedValues.get(i).timeValue()) * ((i + 1D) / totalWeight));
+      }
+      return sum;
+    }
+
   }
 
 
--- a/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
+++ b/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
@@ -195,13 +195,23 @@ public class TrendlineOperator extends PhysicalPlan {
   private static class WeightedMovingAverageAccumulator implements TrendlineAccumulator {
     private final LiteralExpression dataPointsNeeded;
     private final ArrayList<ExprValue> receivedValues;
-    private final ExprCoreType type;
+    private final WmaTrendlineEvaluator evaluator;
 
     public WeightedMovingAverageAccumulator(
         Trendline.TrendlineComputation computation, ExprCoreType type) {
       this.dataPointsNeeded = DSL.literal(computation.getNumberOfDataPoints().doubleValue());
-      this.receivedValues = new ArrayList<>(computation.getNumberOfDataPoints() + 1);
-      this.type = type;
+      this.receivedValues = new ArrayList<>(computation.getNumberOfDataPoints());
+      this.evaluator = getEvaluator(type);
+    }
+
+    static WmaTrendlineEvaluator getEvaluator(ExprCoreType type) {
+        return switch (type) {
+            case DOUBLE -> NumericWmaEvaluator.INSTANCE;
+            case DATE, TIMESTAMP -> TimeStampWmaEvaluator.INSTANCE;
+            case TIME -> TimeWmaEvaluator.INSTANCE;
+            default -> throw new IllegalArgumentException(
+                    String.format("Invalid type %s used for weighted moving average.", type.typeName()));
+        };
     }
 
     @Override
@@ -219,57 +229,64 @@ public class TrendlineOperator extends PhysicalPlan {
       } else if (dataPointsNeeded.valueOf().integerValue() == 1) {
         return receivedValues.getFirst();
       }
-      return computeWma(receivedValues);
+      return evaluator.evaluate(receivedValues);
     }
 
-    /**
-     * Compute WMA values from provided dataset with in ascending order.
-     *
-     * @param receivedValues the dataset for WMA calculation, sorted in ascending order.
-     * @return ExprValue which represent he result onf WMA.
-     */
-    private ExprValue computeWma(ArrayList<ExprValue> receivedValues) {
-      if (type == ExprCoreType.DOUBLE) {
-        return new ExprDoubleValue(calculateWmaInDouble(receivedValues));
+    private static class NumericWmaEvaluator implements WmaTrendlineEvaluator {
 
-      } else if (type == ExprCoreType.DATE) {
-        return ExprValueUtils.timestampValue(
-            Instant.ofEpochMilli(
-                calculateWmaInLong(receivedValues, i -> i.timestampValue().toEpochMilli())));
-      } else if (type == ExprCoreType.TIME) {
+      private static final NumericWmaEvaluator INSTANCE = new NumericWmaEvaluator();
+
+      @Override
+      public ExprValue evaluate(ArrayList<ExprValue> receivedValues) {
+        double sum = 0D;
+        int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
+        for (int i = 0; i < receivedValues.size(); i++) {
+          sum += receivedValues.get(i).doubleValue() * ((i + 1D) / totalWeight);
+        }
+        return new ExprDoubleValue(sum);
+      }
+    }
+
+    private static class TimeStampWmaEvaluator implements WmaTrendlineEvaluator {
+
+      private static final TimeStampWmaEvaluator INSTANCE = new TimeStampWmaEvaluator();
+
+      @Override
+      public ExprValue evaluate(ArrayList<ExprValue> receivedValues) {
+        long sum = 0L;
+        int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
+        for (int i = 0; i < receivedValues.size(); i++) {
+          sum += (long) (receivedValues.get(i).timestampValue().toEpochMilli() * ((i + 1D) / totalWeight));
+        }
+
+        return ExprValueUtils.timestampValue(Instant.ofEpochMilli((sum)));
+      }
+    }
+
+    private static class TimeWmaEvaluator implements WmaTrendlineEvaluator {
+
+      private static final TimeWmaEvaluator INSTANCE = new TimeWmaEvaluator();
+
+      @Override
+      public ExprValue evaluate(ArrayList<ExprValue> receivedValues) {
+        long sum = 0L;
+        int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
+        for (int i = 0; i < receivedValues.size(); i++) {
+          sum += (long) (MILLIS.between(LocalTime.MIN, receivedValues.get(i).timeValue()) * ((i + 1D) / totalWeight));
+        }
         return ExprValueUtils.timeValue(
-            LocalTime.MIN.plus(
-                calculateWmaInLong(
-                    receivedValues, i -> MILLIS.between(LocalTime.MIN, i.timeValue())),
-                MILLIS));
-
-      } else if (type == ExprCoreType.TIMESTAMP) {
-        return ExprValueUtils.timestampValue(
-            Instant.ofEpochMilli(
-                calculateWmaInLong(receivedValues, i -> i.timestampValue().toEpochMilli())));
+                        LocalTime.MIN.plus(sum, MILLIS));
       }
-      return null;
     }
 
-    private double calculateWmaInDouble(ArrayList<ExprValue> receivedValues) {
-      double sum = 0D;
-      int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
-      for (int i = 0; i < receivedValues.size(); i++) {
-        sum += receivedValues.get(i).doubleValue() * ((i + 1D) / totalWeight);
-      }
-      return sum;
-    }
 
-    private long calculateWmaInLong(
-        ArrayList<ExprValue> receivedValues, Function<ExprValue, Long> convertFunc) {
-      long sum = 0L;
-      int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
-      for (int i = 0; i < receivedValues.size(); i++) {
-        sum += (long) (convertFunc.apply(receivedValues.get(i)) * ((i + 1D) / totalWeight));
-      }
-      return sum;
+    private interface WmaTrendlineEvaluator {
+      ExprValue evaluate(ArrayList<ExprValue> receivedValues);
     }
   }
+
+
+
 
   private interface ArithmeticEvaluator {
     Expression calculateFirstTotal(List<ExprValue> dataPoints);
--- a/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
+++ b/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
@@ -17,7 +17,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Function;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
@@ -205,13 +204,13 @@ public class TrendlineOperator extends PhysicalPlan {
     }
 
     static WmaTrendlineEvaluator getEvaluator(ExprCoreType type) {
-        return switch (type) {
-            case DOUBLE -> NumericWmaEvaluator.INSTANCE;
-            case DATE, TIMESTAMP -> TimeStampWmaEvaluator.INSTANCE;
-            case TIME -> TimeWmaEvaluator.INSTANCE;
-            default -> throw new IllegalArgumentException(
-                    String.format("Invalid type %s used for weighted moving average.", type.typeName()));
-        };
+      return switch (type) {
+        case DOUBLE -> NumericWmaEvaluator.INSTANCE;
+        case DATE, TIMESTAMP -> TimeStampWmaEvaluator.INSTANCE;
+        case TIME -> TimeWmaEvaluator.INSTANCE;
+        default -> throw new IllegalArgumentException(
+            String.format("Invalid type %s used for weighted moving average.", type.typeName()));
+      };
     }
 
     @Override
@@ -256,7 +255,10 @@ public class TrendlineOperator extends PhysicalPlan {
         long sum = 0L;
         int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
         for (int i = 0; i < receivedValues.size(); i++) {
-          sum += (long) (receivedValues.get(i).timestampValue().toEpochMilli() * ((i + 1D) / totalWeight));
+          sum +=
+              (long)
+                  (receivedValues.get(i).timestampValue().toEpochMilli()
+                      * ((i + 1D) / totalWeight));
         }
 
         return ExprValueUtils.timestampValue(Instant.ofEpochMilli((sum)));
@@ -272,21 +274,19 @@ public class TrendlineOperator extends PhysicalPlan {
         long sum = 0L;
         int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
         for (int i = 0; i < receivedValues.size(); i++) {
-          sum += (long) (MILLIS.between(LocalTime.MIN, receivedValues.get(i).timeValue()) * ((i + 1D) / totalWeight));
+          sum +=
+              (long)
+                  (MILLIS.between(LocalTime.MIN, receivedValues.get(i).timeValue())
+                      * ((i + 1D) / totalWeight));
         }
-        return ExprValueUtils.timeValue(
-                        LocalTime.MIN.plus(sum, MILLIS));
+        return ExprValueUtils.timeValue(LocalTime.MIN.plus(sum, MILLIS));
       }
     }
-
 
     private interface WmaTrendlineEvaluator {
       ExprValue evaluate(ArrayList<ExprValue> receivedValues);
     }
   }
-
-
-
 
   private interface ArithmeticEvaluator {
     Expression calculateFirstTotal(List<ExprValue> dataPoints);
--- a/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
+++ b/core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
@@ -17,13 +17,13 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
 import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.sql.ast.tree.Trendline;
 import org.opensearch.sql.data.model.ExprDoubleValue;
-import org.opensearch.sql.data.model.ExprLongValue;
 import org.opensearch.sql.data.model.ExprTupleValue;
 import org.opensearch.sql.data.model.ExprValue;
 import org.opensearch.sql.data.model.ExprValueUtils;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -187,6 +191,103 @@ public ExprValue calculate() {
...
+      @Override
+      public ExprValue evaluate(ArrayList<ExprValue> receivedValues) {
+        double sum = 0D;
+        int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
+        for (int i = 0; i < receivedValues.size(); i++) {
+          sum += receivedValues.get(i).doubleValue() * ((i + 1D) / totalWeight);
+        }
+        return new ExprDoubleValue(sum);
+      }
+    }
```
currantw, line 247: There seems to have been quite an explosion of classes here! Some ideas:
- Do we need to do this with singleton classes? Rather than an `Evaluator` class with a single `evaluate` method, would it make sense to instead store a reference to a `evaluate` _method_ that takes a _collection_ of `ExprValue`s and returns the resulting `ExprValue`? Don't know what is more common, but it seems like a lot to have seven separate `Evaluator` classes; perhaps seven `Evaluator` _method_ would be more manageable?
- As alluded to above, could this take `Collection<ExprValue>` instead of an `ArrayList`?
- As mentioned in a previous comment, I think `LinkedList` might be more efficient for this purpose -- but not if we iterator over it like this! Are we able to use an iterator to do so? Then we would get `O(1)` access for both a `LinkedList` or an `ArrayList` (and, as mentioned in previous comment, I think it might be better to generalize this to a `Collection` if possible anyway). See below for how I think this could look.
- As mentioned in a previous comment, SMA and WMA are pretty much the same ... only the weights change. Would it be possible to change the `evaluate` signature so that it takes a `Collection` of `ExprValues` _and_ a `Collection` of `double` weights? i.e. `evaluiate(Collection<ExprValue> values, Collection<double> weights)`? This would have the added advantage that, in the case of WMA, we wouldn't need to re-calculate the weights every time.

```java
  public ExprValue evaluate(Collection<ExprValue> values) {

    // Calculate weights
    int n = values.size();
    double denominator = (double) (n * (n + 1)) / 2.0;
    List<Double> weights = IntStream.range(n, 0).mapToDouble(i -> (double) i / denominator).boxed().toList().reversed();

    // Calculate weighted average.
    double average = 0.0;
    
    Iterator<ExprValue> valuesIterator = values.iterator();
    Iterator<Double> weightsIterator = weights.iterator();
    
    while(valuesIterator.hasNext() && weightsIterator.hasNext()) {
      average += valuesIterator.next().doubleValue() * weightsIterator.next();
    }

    return new ExprDoubleValue(average);
  }
```

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -187,6 +191,103 @@ public ExprValue calculate() {
...
+        int totalWeight = (receivedValues.size() * (receivedValues.size() + 1)) / 2;
+        for (int i = 0; i < receivedValues.size(); i++) {
+          sum +=
+              (long)
+                  (receivedValues.get(i).timestampValue().toEpochMilli()
+                      * ((i + 1D) / totalWeight));
+        }
+
+        return ExprValueUtils.timestampValue(Instant.ofEpochMilli((sum)));
+      }
```
currantw, line 265: At it's heart, most of the logic between `NumericWmaEvaluator`, `TimeWmaEvaluator`, and `TimeStampWmaEvaluator` seems to be common: all the logic for determining the weights and calculating the weights average is the same, the only differences are (a) the type (double vs. long), and (b) how we get the numerical value from the `ExprValue`, and (c) how we convert the resulting weighted average back into an `ExprValue`.

Would it be possible to extract all of this common logic to `WmaTrendlineEvaluator` in a helper method like `Long evaluateHelper(Collection<Long> values)`, and then have each sub-class only (i) convert a collection of `ExprValue` to a collection of `Long`, (ii) call `evaluateHelper` (pick a better name, obviously) to get the weighted average as a `Long`, and then (iii) convert that `Long` back into an `ExprValue`?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -187,6 +191,103 @@ public ExprValue calculate() {
...
+                  (MILLIS.between(LocalTime.MIN, receivedValues.get(i).timeValue())
+                      * ((i + 1D) / totalWeight));
+        }
+        return ExprValueUtils.timeValue(LocalTime.MIN.plus(sum, MILLIS));
+      }
+    }
+
+    private interface WmaTrendlineEvaluator {
+      ExprValue evaluate(ArrayList<ExprValue> receivedValues);
+    }
```
currantw, line 288: As alluded to in other comments, I'm not sure we need separate `Evaluators` for simple and weights moving averages, if we design the interface right.

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -395,4 +396,318 @@ public void calculates_simple_moving_average_timestamp() {
         plan.next());
     assertFalse(plan.hasNext());
   }
+
+  @Test
+  public void calculates_weighted_moving_average_one_field_one_sample() {
+    when(inputPlan.hasNext()).thenReturn(true, false);
+    when(inputPlan.next())
+        .thenReturn(ExprValueUtils.tupleValue(ImmutableMap.of("distance", 100, "time", 10)));
```
currantw, line 404: Might be more readable and a bit more concise (less duplication) to extract a `mockInputPlan` method?

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -395,4 +396,318 @@ public void calculates_simple_moving_average_timestamp() {
...
+    var plan =
+        new TrendlineOperator(
+            inputPlan,
+            Collections.singletonList(
+                Pair.of(
+                    AstDSL.computation(1, AstDSL.field("distance"), "distance_alias", WMA),
+                    ExprCoreType.DOUBLE)));
+
+    plan.open();
+    assertTrue(plan.hasNext());
```
currantw, line 415: Do we need to check that it `hasNext`? Won't this already be converted by calling `next`? Moreover, would it be better to call `execute(plan)` to get the full list of `ExprValue`s that are returned, since this only actually checks whether there is one - it doesn't also verify that there is _only one_.

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -395,4 +396,318 @@ public void calculates_simple_moving_average_timestamp() {
...
+        IllegalArgumentException.class,
+        () -> {
+          new TrendlineOperator(
+              inputPlan,
+              Collections.singletonList(
+                  Pair.of(
+                      AstDSL.computation(2, AstDSL.field("distance"), "distance_alias", WMA),
+                      ExprCoreType.ARRAY)));
+        });
+  }
```
currantw, line 712: Are there other possible error cases to test? Could the number of rows be zero or negative?

docs/user/ppl/cmd/trendline.rst
```diff
@@ -13,7 +13,7 @@ Description
 ============
 | Using ``trendline`` command to calculate moving averages of fields.
 
-Syntax
+Syntax - SMA (Simple Moving Average)
 ============
 `TRENDLINE [sort <[+|-] sort-field>] SMA(number-of-datapoints, field) [AS alias] [SMA(number-of-datapoints, field) [AS alias]]...`
```
currantw, line 18: Does it make sense to split up the syntax? Everything here _looks_ to be the same, other than SMA vs. WMA. Can we combine the syntax into one line, and _then_ explain the differences between the two? Like this, it (at least for me) creates the impression that the syntax for the two is different, when it is actually essentially the same 🤷 

```suggestion
Syntax
============
`TRENDLINE [sort <[+|-] sort-field>] <trendline-type>(number-of-datapoints, field) [AS alias] [<trendline-type>(number-of-datapoints, field) [AS alias]]...`
```

This also makes the think: is it possible to combine SMA and WMA trendline in the same call? If so, is this something we have documented and/or tested?

docs/user/ppl/cmd/trendline.rst
```diff
@@ -70,7 +68,7 @@ PPL query::
     | 15.5 | 30.5      |
     +------+-----------+
 
-Example 4: Calculate the moving average on one field without specifying an alias.
+Example 3: Calculate the moving average on one field without specifying an alias.
 =================================================================================
 
 The example shows how to calculate the moving average on one field.
```
currantw, line 74: Otherwise it's the same as Example 1 😉 

```suggestion
The example shows how to calculate the moving average on one field without specifying an alias.
```

docs/user/ppl/cmd/trendline.rst
```diff
@@ -88,3 +86,77 @@ PPL query::
...
 
+Syntax - WMA (Weighted Moving Average)
+============
+`TRENDLINE [sort <[+|-] sort-field>] WMA(number-of-datapoints, field) [AS alias] [WMA(number-of-datapoints, field) [AS alias]]...`
+
+* [+|-]: optional. The plus [+] stands for ascending order and NULL/MISSING first and a minus [-] stands for descending order and NULL/MISSING last. **Default:** ascending order and NULL/MISSING first.
+* sort-field: mandatory when sorting is used. The field used to sort.
+* number-of-datapoints: mandatory. The number of datapoints to calculate the moving average (must be greater than zero).
+* field: mandatory. The name of the field the moving average should be calculated for.
+* alias: optional. The name of the resulting column containing the moving average (defaults to the field name with "_trendline").
```
currantw, line 97: To follow on from my previous question about combining WMA and SMA into the same statement ... should the default field add `_wma_trendline` or `sma_trendline` to facilitate this? It seems like a reasonable use case that a user might want to get both the WMA and SMA for a data set at the same time, and perhaps we want to avoid a name clash by default? 🤷

docs/user/ppl/cmd/trendline.rst
```diff
@@ -88,3 +86,77 @@ PPL query::
...
+
+It is calculated like
+
+    f[i]: The value of field 'f' in the i-th data point
+    n: The number of data points in the moving window (period)
+    t: The current time index
+    w[i]: The weight assigned to the i-th data point, typically increasing for more recent points
```
currantw, line 104: Since we are using a fixed formula to calculate the weights, should we include that in this documentation? We also say that it is "_typically_ increasing for more recent points" (emphasis added) - isn't this _always_ true, based on the current implementation?

docs/user/ppl/cmd/trendline.rst
```diff
@@ -88,3 +86,77 @@ PPL query::
...
+    | 4.333333333333333  | 34.0      |
+    | 10.666666666666666 | 32.0      |
+    | 16.333333333333332 | 30.5      |
+    +--------------------+-----------+
+
+
+Example 3: Calculate the weighted moving average on one field without specifying an alias.
+=================================================================================
+
+The example shows how to calculate the weighted moving average on one field.
```
currantw, line 148: As previous comment:

```suggestion
The example shows how to calculate the weighted moving average on one field without specifying an alias.
```

docs/user/ppl/cmd/trendline.rst
```diff
@@ -88,3 +86,77 @@ PPL query::
...
+
+PPL query::
+
+    os> source=accounts | trendline wma(2, account_number)  | fields account_number_trendline;
```
currantw, line 152: Might as well while we're here 🤷 

Same extra space in the SMA example as well.

```suggestion
    os> source=accounts | trendline wma(2, account_number) | fields account_number_trendline;
```

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
```diff
@@ -75,4 +75,71 @@ public void testTrendlineWithSort() throws IOException {
...
+
+  @Test
+  public void testTrendlineOverwritesExistingFieldWma() throws IOException {
+    final JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | where balance > 39000 | sort balance | trendline wma(2, balance) as"
+                    + " age | fields age",
+                TEST_INDEX_BANK));
+    System.out.println("Result (Overwrite) : " + result.toString());
```
currantw, line 116: Looks like this might have got left in from debugging..?

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
```diff
@@ -75,4 +75,71 @@ public void testTrendlineWithSort() throws IOException {
...
+    verifyDataRows(
+        result, rows(new Object[] {null}), rows(40101.666666666664), rows(45570.666666666664));
```
currantw, line 118: Can you just do this? Looks like some other tests have this?

```suggestion
        result, rows(null), rows(40101.666666666664), rows(45570.666666666664));
```

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
```diff
@@ -75,4 +75,71 @@ public void testTrendlineWithSort() throws IOException {
...
+
+  @Test
+  public void testTrendlineWithSortWma() throws IOException {
+    final JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | where balance > 39000 | trendline sort balance wma(2, balance) |"
+                    + " fields balance_trendline",
+                TEST_INDEX_BANK));
+    System.out.println("Result (WithSortWma): " + result.toString());
```
currantw, line 141: Similar to previous comment.

currantw (commented):
Reviewed to `f6c82ae`.

YANG-DB (approved):

YANG-DB:
LGTM - thanks

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
andy-k-improving: This method signature of `rows( )` is defaulted to read an array, and I reckon we should keep that by supplying an array for the method invoke, or explicit to specify multiple `null` like what test-case `testTrendlineMultipleFields` do.

Because `...` is being used, so specifying a single `null` will mis-lead the compiler which `null` will be assigned to `expectedObjects`, in contrast to the case of  `null, null` and `new Object[] {null}` which both hint compiler an valid Object array is being passed in, with content `null`.

 
image

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -106,9 +108,11 @@ private Map<String, ExprValue> consumeInputTuple(ExprValue inputValue) {
...
       Pair<Trendline.TrendlineComputation, ExprCoreType> computation) {
-    // Add a switch statement based on computation type to choose the accumulator when more
-    // types of computations are supported.
-    return new SimpleMovingAverageAccumulator(computation.getKey(), computation.getValue());
+    return switch (computation.getKey().getComputationType()) {
+      case SMA -> new SimpleMovingAverageAccumulator(computation.getKey(), computation.getValue());
+      case WMA -> new WeightedMovingAverageAccumulator(
+          computation.getKey(), computation.getValue());
+    };
   }
```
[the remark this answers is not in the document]
andy-k-improving: That make sense to have default when target Object is not primitive type (Actual object), however in this case `enum` is being used, so the choice is limited and finite, only SMA or WMA, there won't be the case of having another value or null. 

Similar opinion from the referenced site:

> Also, think you are checking for days of the week declared an [Enum, an enumerated type](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum).
> 
> Then obviously, you can have only seven different values. Another example is checking for keystrokes which can have four values; move up, down, left, and right. Therefore, the default case is not required to define in such cases.

Also it's possible for user to supply an foreign value, due to ANTLR grammar check and nature of Java enum type.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -187,6 +191,103 @@ public ExprValue calculate() {
     }
   }
 
+  private static class WeightedMovingAverageAccumulator implements TrendlineAccumulator {
+    private final LiteralExpression dataPointsNeeded;
+    private final ArrayList<ExprValue> receivedValues;
+    private final WmaTrendlineEvaluator evaluator;
```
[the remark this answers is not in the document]
andy-k-improving: See below:

1. Moved both properties to parent class.
2. `LiteralExpression` is being used because the current implementation of  SMA is leveraging the class of `DSL` to perform calculation, see `TimeArithmeticEvaluator.evaluate( )` 

```
      return ExprValueUtils.timeValue(
          LocalTime.MIN.plus(
              DSL.divide(runningTotal, numberOfDataPoints).valueOf().longValue(), MILLIS));
    }
```

Hence data need to be kept in `Expression` form in order to provide backward compatibility to current SMA implementation. 

3. Half-half, but if we standardise this way, this will force SMA value to be re-calculated completely upon every invoke of `evaluate( )`, which in contrast the current SMA is taking an optimised approach to leverage the concept of `running-total` in order to avoid the complete computation. 

Hence in term of Java class structure perspective, that make sense to have a generic parent class to regular a given `Trendline` implementation will:
 - Read `dataPointsNeeded`
 - To have reservoir-ish collection to serve as an internal buffer 
 - Consume new value via `accumulate( )` call 
 - `calculate( )` to output the up-to-date trendline result. 

But we shouldn't enfroce the computation of each implementation, ex: 
 - How to calculate result (Equal weight or not)

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -187,6 +191,103 @@ public ExprValue calculate() {
...
+
+    public WeightedMovingAverageAccumulator(
+        Trendline.TrendlineComputation computation, ExprCoreType type) {
+      this.dataPointsNeeded = DSL.literal(computation.getNumberOfDataPoints().doubleValue());
```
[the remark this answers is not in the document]
andy-k-improving: In the case of WMA, I agree there is no need to have it in `double`, I have updated it to pass `integer` directly to `DSL.literall`.
In the case of SMA, the `double` is require to perform `DSL.divide()`, or else the result will be rounded.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -187,6 +191,103 @@ public ExprValue calculate() {
...
+      this.receivedValues = new ArrayList<>(computation.getNumberOfDataPoints());
```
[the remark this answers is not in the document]
andy-k-improving: Random access is required, as I have to access all items in the collection to calculate WMA result, see `NumericWmaEvaluator#evalute( )` for detail.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
andy-k-improving: I understand the point for too many Evaluator, indeed these are two distinct types of `Evaluator`, in the case of SMA, `Evaluator` not just the new item, but also the `running-total` in order to avoid re-computation, and all evaluator under the SMA umbrella take this into consideration especially on the API signature level, ex:
```
public ExprValue evaluate(Expression runningTotal, LiteralExpression numberOfDataPoints) {
      return DSL.divide(runningTotal, numberOfDataPoints).valueOf();
    }
```
Also the method calculateFirstTotal is unique to SMA:
```
@Override
    public Expression calculateFirstTotal(List<ExprValue> dataPoints) {
      Expression total = DSL.literal(0.0D);
      for (ExprValue dataPoint : dataPoints) {
        total = DSL.add(total, DSL.literal(dataPoint.doubleValue()));
      }
      return DSL.literal(total.valueOf().doubleValue());
    }
```

However in contrast, WMA don't share the same characteristic, which re-computation is required upon every update, also the concept of `running-total` is not appliable here. 

Regarding the concern of too many evaluators, I have updated to move all WMA related evaludator inside of class `WeightedMovingAverageAccumulator` in order to further narrow down the scope.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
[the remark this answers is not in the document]
currantw: Agree that in this case the options are "limited and finite", but I think that it differs from the weekday and keystroke examples because the options are not expected to be _fixed_ - additional options may be added in the future. The days of the week or the four keystroke values are almost certain not to change in the future, so a default case is unnecessary.

But I don't think that same logic necessarily applies to the trendline type enum. I seems possible to me (even likely) that there will be demand to add another trendline type in the future (there are _lots_ of ways to create trendlines!). If that happens, and the implementer forgets to add a new `case` here, that would lead to a `NullValueException` somewhere down the line, rather than a more user friendly exception like "trendline type XYZ is not supported". Probably not a likely scenario, but still.

Anyway, happy to leave it as is. Please feel free to resolve.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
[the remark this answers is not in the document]
currantw: In `NumericWmaEvaluator#evalute`, aren't we accessing the values _sequentially_, rather than _randomly_? I believe that using an iterator to access the values, rather than accessing them by index, would result in O(1) performance for both a linked list and array list, right?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -106,16 +110,28 @@ private Map<String, ExprValue> consumeInputTuple(ExprValue inputValue) {
...
+      case SMA -> new SimpleMovingAverageAccumulator(computation.getKey(), computation.getValue());
+      case WMA -> new WeightedMovingAverageAccumulator(
+          computation.getKey(), computation.getValue());
+    };
   }
 
   /** Maintains stateful information for calculating the trendline. */
-  private interface TrendlineAccumulator {
-    void accumulate(ExprValue value);
+  private abstract static class TrendlineAccumulator<C extends Collection<ExprValue>> {
```
currantw, line 121: Rather than using a template here, could `TrendlineAccumulator`s not just all use a `Queue` to store their values?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -106,16 +110,28 @@ private Map<String, ExprValue> consumeInputTuple(ExprValue inputValue) {
...
+    private TrendlineAccumulator(LiteralExpression dataPointsNeeded, C receivedValues) {
+      this.dataPointsNeeded = dataPointsNeeded;
+      this.receivedValues = receivedValues;
+    }
+
+    abstract void accumulate(ExprValue value);
+
+    abstract ExprValue calculate();
 
     static ArithmeticEvaluator getEvaluator(ExprCoreType type) {
```
currantw, line 136: Should this be moved to `SimpleMovingAverageAccumulator`? I think it's only used there, right?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -187,6 +203,103 @@ public ExprValue calculate() {
...
+        Trendline.TrendlineComputation computation, ExprCoreType type) {
+      super(
+          DSL.literal(computation.getNumberOfDataPoints()),
+          new ArrayList<>(computation.getNumberOfDataPoints()));
+      this.evaluator = getWmaEvaluator(type);
+    }
+
+    static WmaTrendlineEvaluator getWmaEvaluator(ExprCoreType type) {
+      return switch (type) {
+        case DOUBLE -> NumericWmaEvaluator.INSTANCE;
```
currantw, line 220: What about other numerical data types (INTEGER, FLOAT, BYTE). Do we support trendlines for these?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -187,6 +203,103 @@ public ExprValue calculate() {
...
+      receivedValues.add(value);
+      if (receivedValues.size() > dataPointsNeeded.valueOf().integerValue()) {
+        receivedValues.removeFirst();
+      }
+    }
+
+    @Override
+    public ExprValue calculate() {
+      if (receivedValues.size() < dataPointsNeeded.valueOf().integerValue()) {
+        return null;
```
currantw, line 239: Should we return an `ExprNullValue` here instead?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
currantw: Thanks for explanation of this in another comments. Please resolve. ✅

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
currantw: Thanks. Good explanation for why SMA and WMA accumulators should be separate - thanks for that!

For WMA, I still think there is opportunity to combine some common logic. The weights should only need to be calculated once - can we pass them directly to the Evaluator from the Accumulator? Moreover, as I (tried to) describe in [this comment](https://github.com/opensearch-project/sql/pull/3293/files#r1941674993), can we extract the common logic from all the WMA accumulators (related applying the weights to the values), and only have the different parts (mapping each value to a number, mapping the result back to the right `ExprValue`) split out into the different implementations.

As mentioned elsewhere, I think we should also use an iterator for these loops, so that we can get O(1) access for a linked list or queue.

currantw (commented):
Reviewed to `4d95529`.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
andy-k-improving: No really, because in the case of `null`, the field insertion will be skipped completely for that particular data row, instead of honouring the value of `calculateResult`.
Hence there will be no different between `null` and `ExprNullValue`, as it won't be populated to the final result. 
```
apBuilder.putAll(inputStruct);

    // Add calculated trendline values, which might overwrite existing fields from the input.
    for (int i = 0; i < accumulators.size(); ++i) {
      final ExprValue calculateResult = accumulators.get(i).calculate();
      final String field = computations.get(i).getKey().getAlias();
      if (calculateResult != null) {
        mapBuilder.put(field, calculateResult);
      }
    }
```

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
andy-k-improving: Updated the code to support all numeric types, except BYTE.
See `calculates_weighted_moving_average_data_type_support_integer( )`-ish for test on respective datatype.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
andy-k-improving: Yes, moved it to `SimpleMovingAverageAccumulator`

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
andy-k-improving: Removed the generic with `Queue<ExprValue>` instead, with the aim to standardise the contract between SMA and WMA

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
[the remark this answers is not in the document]
andy-k-improving: Yep, updated to use `LinkedList` with iterator for WMA computation.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
andy-k-improving: Above make sense, and I have now using `BiFunction` to replace original usage of custom interface of `wmaEvalulator` along with the static class creation.
Also I have moved out the logic of `totalWeight` calculation out from respective function call, as that is common to all calculation.

docs/user/ppl/cmd/trendline.rst
andy-k-improving: Updated to avoid ambiguous statement.

docs/user/ppl/cmd/trendline.rst
andy-k-improving: I see this differently, as we should stick with the Splunk offering as much as possible:
https://docs.splunk.com/Documentation/Splunk/9.4.0/SearchReference/Trendline
Which has no default option nor support multiple options. 
In the case if user is required to have multiple trendline value side by side, they will be recommended to use `appendCol` instead.
Our position here should be creating primitive command || functionality which do one thing at best, then user is expected to mix and match in order to satisfy their own requirement on the query, rather than having one powerful command which try to cater all requirements at once.

docs/user/ppl/cmd/trendline.rst
```diff
@@ -15,17 +15,16 @@ Description
 
 Syntax
 ============
-`TRENDLINE [sort <[+|-] sort-field>] SMA(number-of-datapoints, field) [AS alias] [SMA(number-of-datapoints, field) [AS alias]]...`
+`TRENDLINE [sort <[+|-] sort-field>] <trendline-type>(number-of-datapoints, field) [AS alias] [<trendline-type>(number-of-datapoints, field) [AS alias]]...`
 
 * [+|-]: optional. The plus [+] stands for ascending order and NULL/MISSING first and a minus [-] stands for descending order and NULL/MISSING last. **Default:** ascending order and NULL/MISSING first.
 * sort-field: mandatory when sorting is used. The field used to sort.
+* trendline-type: mandatory. The type of algorithm being used for the calculation, only SMA || WMA are supported at the moment.
```
currantw, line 22: I think plain English might be a bit more understandable here? 🤷 

```suggestion
* trendline-type: mandatory. The type of algorithm being used for the calculation, only SMA (simple moving average) or WMA (weighted moving average) are supported at the moment.
```

docs/user/ppl/cmd/trendline.rst
```diff
@@ -88,3 +87,69 @@ PPL query::
...
+In the case of Weighted Moving Average - WMA, result will be calculated as per the below formula.
+
+    f[i]: The value of field 'f' in the i-th data point
+    n: The number of data points in the moving window (period)
+    t: The current time index
+    w[i]: The weight of the i-th data point, increasing by one per step to prioritize recent points.
+
+    WMA(t) = ( Σ from i=t−n+1 to t of (w[i] * f[i]) ) / ( Σ from i=t−n+1 to t of w[i] )
+
+Example 1: Calculate the weighted moving average on one field.
```
currantw, line 101: Since the headers for the WMA examples include "weighted moving average:, probably makes sense to update the headers for the SMA examples to include "_simple_ moving average" for consistency?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -106,44 +110,41 @@ private Map<String, ExprValue> consumeInputTuple(ExprValue inputValue) {
...
     private Expression runningTotal = null;
 
     public SimpleMovingAverageAccumulator(
         Trendline.TrendlineComputation computation, ExprCoreType type) {
-      dataPointsNeeded = DSL.literal(computation.getNumberOfDataPoints().doubleValue());
-      receivedValues = EvictingQueue.create(computation.getNumberOfDataPoints());
-      evaluator = TrendlineAccumulator.getEvaluator(type);
+      super(
+          DSL.literal(computation.getNumberOfDataPoints().doubleValue()),
+          EvictingQueue.create(computation.getNumberOfDataPoints()));
```
currantw, line 146: Can this logic be moved up to the parent class? Seems like it is duplicated in both the SMA and WMA cases, with the exception that the SMA sub-class uses an `EvictingQueue` while the WMA sub-class uses a `LinkedList`. Can't they both use the same data structure?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -185,133 +186,226 @@ public ExprValue calculate() {
       }
       return evaluator.evaluate(runningTotal, dataPointsNeeded);
     }
-  }
 
-  private interface ArithmeticEvaluator {
-    Expression calculateFirstTotal(List<ExprValue> dataPoints);
+    static ArithmeticEvaluator getEvaluator(ExprCoreType type) {
+      return switch (type) {
+        case INTEGER, SHORT, LONG, FLOAT, DOUBLE -> NumericArithmeticEvaluator.INSTANCE;
```
currantw, line 192: Thanks for adding support for these other types! 🎉

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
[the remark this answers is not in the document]
currantw: Agree with above. Thanks for explanation. Please feel free to resolve.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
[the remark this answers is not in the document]
currantw: Sure. Sounds good. Please feel free to resolve.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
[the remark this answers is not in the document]
currantw: Thanks for the update! As mentioned in another comment, I think we can move this up to the parent class. I don't see any reason why SMA should use an `EvictingQueue` and WMA should use `LinkedList`?

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
currantw: > Above make sense, and I have now using BiFunction to replace original usage of custom interface of wmaEvalulator along with the static class creation.

Thanks. This looks good to me.

> Also I have moved out the logic of totalWeight calculation out from respective function call, as that is common to all calculation.

I like that you have moved the `totalWeight` (the denominator) so that it doesn't need to be re-calculated each time. However, I think it is possible to move _all_ the weight calculations out of these functions, and just pass a list containing all the weights to the function (i.e. to store `n` "complete" weights values as a `WeightedMovingAverageAccumulator` member).

I also think it is possible to extract a the common logic for determining `sum` into another helper function: as mentioned in a previous comment, that logic is the same, except for mapping the `ExprValue` list to longs.

Let me know if you want to discuss either.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
currantw: As mentioned in another comment, still think there is common logic that can be extracted. Please let me know if you want to discuss.

docs/user/ppl/cmd/trendline.rst
currantw: > Which has no default option

Splunk does have a default field name though, right? From the documentation, it seems like it would be `sma.<period>(<field>)` (e.g. `sma5(foo)`) for SMA and `wma.<period>(<field>)` (e.g. `wma5(foo)` for WMA). Splunk also support using a single `trendline` command to generate multiple trendline fields (e.g. both SMA and EMA). Both of these are illustrated in Example 1 in the link you posted above.

Given that, and given that we already support creating multiple trendlines with a single command, it seems to me that what I've proposed above is consistent with what Splunk already does, unless I'm missing something?

docs/user/ppl/cmd/trendline.rst
currantw: Thanks! 🎉

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
currantw: Thanks. No additional comments from me here. Please feel free to resolve.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
currantw: Thanks. Please resolve.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
currantw: Thanks. Please feel free to resolve.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
currantw: Sure.I guess it's semantic 🤷 Please feel free to resolve.

currantw (commented):
Reviewed to `b8cf496d`.

docs/user/ppl/cmd/trendline.rst
andy-k-improving: Done

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
andy-k-improving: Done.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
andy-k-improving: As discussed offline the methods are not abstracted.

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -102,31 +101,101 @@ public void calculates_simple_moving_average_one_field_two_samples_three_rows()
...
+            tupleValue(ImmutableMap.of("distance", 200, "time", 10)),
+            tupleValue(ImmutableMap.of("distance", 200, "time", 10))));
+
+    var plan =
+        new TrendlineOperator(
+            inputPlan,
+            Collections.singletonList(
+                Pair.of(
+                    AstDSL.computation(2, AstDSL.field("distance"), "distance_alias", SMA),
+                    ExprCoreType.SHORT)));
```
acarbonetto, line 128: nit: seems like all these typed 

use `@ParameterizedTest` as an annotation on the test (instead of @Test)

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -371,28 +407,452 @@ public void calculates_simple_moving_average_timestamp() {
...
+
+    List<ExprValue> result = execute(plan);
+    assertEquals(2, result.size());
+    assertThat(
+        result,
+        containsInAnyOrder(
+            tupleValue(ImmutableMap.of("distance", 100, "time", 10)),
+            tupleValue(
+                ImmutableMap.of(
+                    "distance", 200, "time", 10, "distance_alias", 166.66666666666663))));
```
acarbonetto, line 473: if you can, simplify this magic number. 
its hard to tell if this number is correct.

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -371,28 +407,452 @@ public void calculates_simple_moving_average_timestamp() {
...
+    assertEquals(3, result.size());
+    assertThat(
+        result,
+        containsInAnyOrder(
+            tupleValue(ImmutableMap.of("distance", 100, "time", 10)),
+            tupleValue(
+                ImmutableMap.of("distance", 200, "time", 10, "distance_alias", 166.66666666666663)),
+            tupleValue(
+                ImmutableMap.of(
+                    "distance", 200, "time", 10, "distance_alias", 199.99999999999997))));
```
acarbonetto, line 502: same here

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -371,28 +407,452 @@ public void calculates_simple_moving_average_timestamp() {
...
+            tupleValue(
+                ImmutableMap.of(
+                    "timestamp",
+                    Instant.EPOCH.plusMillis(1500),
+                    "timestamp_alias",
+                    Instant.EPOCH.plusMillis(1333)))));
+  }
+
+  @Test
+  public void use_illegal_core_type_wma() {
```
acarbonetto, line 810:
```suggestion
  public void use_illegal_core_type_wma_throws_IllegalArgumentException() {
```

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -371,28 +407,452 @@ public void calculates_simple_moving_average_timestamp() {
...
+            new TrendlineOperator(
+                inputPlan,
+                Collections.singletonList(
+                    Pair.of(
+                        AstDSL.computation(2, AstDSL.field("distance"), "distance_alias", WMA),
+                        ExprCoreType.ARRAY))));
+  }
+
+  @Test
+  public void use_invalid_dataPoints_zero() {
```
acarbonetto, line 823:
```suggestion
  public void use_invalid_dataPoints_zero_throws_IllegalArgumentException() {
```

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -371,28 +407,452 @@ public void calculates_simple_moving_average_timestamp() {
...
+            new TrendlineOperator(
+                inputPlan,
+                Collections.singletonList(
+                    Pair.of(
+                        AstDSL.computation(0, AstDSL.field("distance"), "distance_alias", WMA),
+                        ExprCoreType.INTEGER))));
+  }
+
+  @Test
+  public void use_invalid_dataPoints_negative() {
```
acarbonetto, line 836:
```suggestion
  public void use_invalid_dataPoints_negative_throws_IllegalArgumentException() {
```

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
```diff
@@ -371,28 +407,452 @@ public void calculates_simple_moving_average_timestamp() {
...
+  public void use_illegal_core_type_wma() {
+    assertThrows(
+        IllegalArgumentException.class,
+        () ->
+            new TrendlineOperator(
+                inputPlan,
+                Collections.singletonList(
+                    Pair.of(
+                        AstDSL.computation(2, AstDSL.field("distance"), "distance_alias", WMA),
+                        ExprCoreType.ARRAY))));
```
acarbonetto, line 819: you could use a @ParameterizedTest to test all invalid type in the enum

integ-test/src/test/java/org/opensearch/sql/ppl/TrendlineCommandIT.java
```diff
@@ -75,4 +75,68 @@ public void testTrendlineWithSort() throws IOException {
...
+  public void testTrendlineWithSortWma() throws IOException {
+    final JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | where balance > 39000 | trendline sort balance wma(2, balance) |"
+                    + " fields balance_trendline",
+                TEST_INDEX_BANK));
+    verifyDataRows(
+        result, rows(new Object[] {null}), rows(40101.666666666664), rows(45570.666666666664));
+  }
```
acarbonetto, line 141: please add test with both wma and sma trendline default name

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
andy-k-improving: This is now part of the parameterizedTest.

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
andy-k-improving: This is now part of the parameterizedTest.

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
andy-k-improving: This is now part of the parameterizedTest.

core/src/test/java/org/opensearch/sql/planner/physical/TrendlineOperatorTest.java
andy-k-improving: done.

core/src/main/java/org/opensearch/sql/planner/physical/TrendlineOperator.java
```diff
@@ -185,133 +191,233 @@ public ExprValue calculate() {
...
+                DSL.divide(runningTotal, numberOfDataPoints).valueOf().longValue(), MILLIS));
+      }
     }
 
-    @Override
-    public ExprValue evaluate(Expression runningTotal, LiteralExpression numberOfDataPoints) {
-      return ExprValueUtils.timeValue(
-          LocalTime.MIN.plus(
-              DSL.divide(runningTotal, numberOfDataPoints).valueOf().longValue(), MILLIS));
+    private static class TimestampArithmeticEvaluator implements ArithmeticEvaluator {
```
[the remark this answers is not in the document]
Yury-Fridlyand: I see you already done this in `WMA_*_EVALUATOR`

LantaoJin:
Resolved by https://github.com/opensearch-project/sql/pull/3741

PR #3318, opened 2025-02-11
Make basic aggregation working (part 1)

### LINKED ISSUE #3308: [FEATURE] Calcite Engine Framework: Basic aggregation integration

PPL basic Aggregation in Calcite should work (at least 1 agg function)

LantaoJin:
### Description
This PR includes 5 changes:
1. Introduce `OpenSearchTypeSystem` (partial code) to address the type mismatch issue in Agg
2. Using AccessController to workaround the CI failures caused by java security.
3. Add parameters to enable calcite codegen and debug, ref https://janino-compiler.github.io/janino/#debugging
4. Refactor Calcite integration tests, now we have both integration test `CalcitePPL*IT` and unit test `CalcitePPL*Test` for each command.
5. Fix bugs in `OpenSearchIndexEnumerator`

### Related Issues
Partially Resolves #3308 

### Check List


core/src/main/java/org/opensearch/sql/executor/OpenSearchTypeSystem.java
```diff
@@ -0,0 +1,29 @@
...
+  private OpenSearchTypeSystem() {}
+
+  public RelDataType deriveAvgAggType(RelDataTypeFactory typeFactory, RelDataType argumentType) {
+    switch (argumentType.getSqlTypeName()) {
+      case INTEGER:
+      case BIGINT:
+        return typeFactory.createSqlType(SqlTypeName.DOUBLE);
+
+      default:
+        return super.deriveSumType(typeFactory, argumentType);
```
penghuo, line 26: why default to deriveSumType instead of deriveAvgAggType?

core/src/main/java/org/opensearch/sql/executor/OpenSearchTypeSystem.java
```diff
@@ -0,0 +1,29 @@
...
+public class OpenSearchTypeSystem extends RelDataTypeSystemImpl {
+  public static final RelDataTypeSystem INSTANCE = new OpenSearchTypeSystem();
+
+  private OpenSearchTypeSystem() {}
+
+  public RelDataType deriveAvgAggType(RelDataTypeFactory typeFactory, RelDataType argumentType) {
+    switch (argumentType.getSqlTypeName()) {
+      case INTEGER:
+      case BIGINT:
+        return typeFactory.createSqlType(SqlTypeName.DOUBLE);
```
penghuo, line 23: is it a bug in Calcite? if the argumentType is INTEGER, deriveAvgAggType is INTEGER?

penghuo (approved):

core/src/main/java/org/opensearch/sql/executor/OpenSearchTypeSystem.java
LantaoJin: I think so,  for example table1:
| id |
|------|
| 1    |
| 2    |
| 3    |
| 4    |

`select avg(id) from table1` **returns 2 instead of 2.5**.

Not sure it is a bug or by designed. From semantic purpose, query avg on an `id` column is semantic wrong. The column which could be applied avg should be designed as decimal number. Will open a Calcite issue for checking.

Issue #3339, opened 2025-02-19
[FEATURE] Unify OpenSearch PPL Data Type and Spark PPL Data Type


penghuo:
**Is your feature request related to a problem?**
Currently, OpenSearch PPL data type and Spark PPL data type are not aligned. For instance,  https://github.com/opensearch-project/opensearch-spark/issues/1057.  

**What solution would you like?**
We proposed a breaking change in OpenSearch PPL Data Type.

OpenSearch PPL Data Type | Spark PPL Data Type | Unified PPL Data Type
-- | -- | --
byte | tinyint | tinyint 
short | smallint | smallint 
integer | int | int
long | bigint | bigint 

It is breaking change, and release in OpenSearch 3.0,  change includes
* ExprCoreType
* Integration Test
* Documentation

**What alternatives have you considered?**

**Do you have any additional context?**

penghuo:
* OSD PR, [[Discover] Add New Mappings for S3 Type](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9430)

LantaoJin:
@penghuo , do you mean in 3.0.0, the PPL data type won't include byte, short, integer and long. instead, they will upgrade to 
byte -> tinyint
short -> smallint
integer -> int
long -> bigint

penghuo:
> @penghuo , do you mean in 3.0.0, the PPL data type won't include byte, short, integer and long. instead, they will upgrade to
byte -> tinyint
short -> smallint
integer -> int
long -> bigint

* change includes
  ** data type in response
  ** update [OpenSearch data type doc](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/general/datatypes.rst), to reflect NEW data type.

* not touched
  ** internal ExprCoreType will not changed.

PR #3345, opened 2025-02-24
Unified OpenSearch PPL Data Type

penghuo:
### Description
This PR introduces a language specification abstraction to support SQL and PPL query processing. The main changes include:
* A new interface, LangSpec, is added with a default SQL implementation and a custom PPLLangSpec that maps specific expression types (e.g., mapping BYTE to “tinyint”).
* Updates in index describe requests and query results ensure that the correct type names are used based on the active language specification (SQL or PPL).
* Updated tests verify the behavior of the language specification mappings and system index utilities. 
* Updated [OpenSearch PPL data type doc](https://github.com/opensearch-project/sql/blob/7e381d54bf3af878d1a6e942b9347d30076b9c66/docs/user/ppl/general/datatypes.rst).

### To Reviewer
* Ideally, Ideally, the query engine should use well-defined data types, with LangSpec serving as the protocol for translating these engine types to language-specific types.  But, currently, the describe implementation relies on OpenSearchDataType (an extension of ExprCoreDataType). After the Calcite migration, all data types should be unified as CalciteDataType. then the specially handing in describe is not necessary.

### Related Issues
https://github.com/opensearch-project/sql/issues/3339

### Check List


core/src/main/java/org/opensearch/sql/lang/PPLLangSpec.java
```diff
@@ -0,0 +1,50 @@
...
+  public static final PPLLangSpec PPL_SPEC = new PPLLangSpec();
+
+  private static Map<ExprType, String> exprTypeToPPLType = new HashMap<>();
+
+  static {
+    exprTypeToPPLType.put(ExprCoreType.BYTE, "tinyint");
+    exprTypeToPPLType.put(ExprCoreType.SHORT, "smallint");
+    exprTypeToPPLType.put(ExprCoreType.INTEGER, "int");
+    exprTypeToPPLType.put(ExprCoreType.LONG, "bigint");
+  }
```
LantaoJin, line 30: Since this is a breaking change, why don't we directly change the old type to the new type, instead of introducing `LangSpec`? Isn't this unnecessarily adding complexity?

penghuo: I agree. The core type was not upgraded because I intended for the data type changes to affect only PPL and not SQL. This PR focuses on unifying PPL data types, while SQL data types can be addressed in a separate issue, as changes there would impact JDBC, ODBC, and CLI.

Ideally, the query engine should use well-defined data types, with LangSpec serving as the protocol for translating these engine types to language-specific types. Once the Calcite implementation is complete, CalciteDataType will translate to ExprDataType, and LangSpec will translate from ExprDataType to the PPL response data type.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/system/OpenSearchDescribeIndexRequest.java
```diff
@@ -64,7 +79,10 @@ public List<ExprValue> search() {
       results.add(
           row(
               entry.getKey(),
-              entry.getValue().legacyTypeName().toLowerCase(),
+              (langSpec.language() == LangSpec.LangType.PPL
+                      ? langSpec.typeName(entry.getValue().getExprType())
+                      : entry.getValue().legacyTypeName())
```
dai-chen, line 84: So for SQL, we also have a legacy spec?

dai-chen (dismissed):

core/src/main/java/org/opensearch/sql/lang/LangSpec.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Represents a language specification for query processing.
+ *
+ * <p>This interface defines basic methods for language-specific behaviors, such as determining the
+ * language type and mapping expression types to type names. Two language specifications are
+ * provided: SQL and PPL.
+ */
+public interface LangSpec {
+  /** Enumerates the supported language types. */
+  enum LangType {
```
LantaoJin, line 21: minor: a little duplicated with https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/executor/QueryType.java

LantaoJin (commented):
Seems the `SystemFunctionIT` should be updated correspondingly?

core/src/main/java/org/opensearch/sql/lang/LangSpec.java
penghuo: done.

LantaoJin (approved):

dai-chen (approved):

PR #3350, opened 2025-02-26
Fix execution errors caused by plan gap

qianheng-aws:
### Description
This PR fixes several errors caused by Calcite optimization. These issues arise because there are several gaps between our OpenSearch Plan and the Calcite Plan, and the Calcite optimization process is specific to SQL and its own semantic requirements.

This PR includes changes:
1. Add OpenSearch settings for calcite plan push down and make it disabled by default. Will set default to enabled once all tests works with push down.

2. Fix error for single column row. Calcite has optimization that transform such row into scalar object and every enumerable operators all follow that convention. In order to combine our scan operators with them, we should adjust to follow it as well.

3. Fix error for sort or collation behavior of the final results. Calcite doesn't ensure collation of a query if its plan doesn't end with sort operator. For example, it has optimization that will remove the sort inner a subquery since it doesn't ensure the collation outside of that subquery. While for the semantic of our PPL, we expect to ensure collation even a query has other operators(i.e. project) after sort command. 
However, we don't plan to hack the prepare process of Calcite,  the only thing we can do is transform our plan to conform to Calcite conventions. For this case, before optimization, we should add a sort operator as the root of our plan if the original root has collation for its output data. And in optimization, we rely on calcite optimizer to remove redundant sort operators.

4. Fix error for pushdown by lazily construct OpenSearchRequestBuilder. The previous implementation has a problem in optimization process that scan operators in different equivalent plans may share the same objects in  OpenSearchRequestBuilder, since OpenSearchRequestBuilder cannot do deep copy. It has issue that transformation on one plan may affect other plans while it's never allowed.
To fix this, we should lazily construct OpenSearchRequestBuilder and maintain these push down actions in a queue, until the final implementation stage. Since queue supports deep copy and each push down actions are all enclosure, we can avoid the above issue. 

### Related Issues

### Check List


core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -132,7 +137,31 @@ public void executePlanByCalcite(
...
+    /* Calcite only ensures collation of the final result produced from the root sort operator.
+     * While we expect that the collation can be preserved through the pipes over PPL, we need to
+     * explicitly add a sort operator on top of the original plan
+     * to ensure the correct collation of the final result.
+     * See logic in ${@link CalcitePrepareImpl}
+     * For the redundant sort, we rely on Calcite optimizer to eliminate
+     */
+    RelCollation collation = osPlan.getTraitSet().getCollation();
+    if (!(osPlan instanceof Sort) && collation != RelCollations.EMPTY) {
+      calcitePlan = LogicalSort.create(osPlan, collation, null, null);
```
LantaoJin, line 161: Any reason for adding this NotNull annotation? Would be better to add this?
```
if (calcitePlan = null) return osPlan;
```

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchFilterIndexScanRule.java
```diff
@@ -42,7 +42,8 @@ public void onMatch(RelOptRuleCall call) {
   }
 
   protected void apply(RelOptRuleCall call, Filter filter, CalciteOpenSearchIndexScan scan) {
-    if (scan.pushDownFilter(filter)) {
+    CalciteOpenSearchIndexScan newScan = scan.pushDownFilter(filter);
+    if (newScan != null) {
```
LantaoJin, line 46: `newScan` won't be used after this check. the name `newScan` confused me.
could you change it to
```
if (scan.pushDownFilter(filter) != null) {
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOpenSearchIndexScan.java
```diff
@@ -56,24 +59,31 @@ public class CalciteOpenSearchIndexScan extends OpenSearchTableScan {
    */
   public CalciteOpenSearchIndexScan(
       RelOptCluster cluster, RelOptTable table, OpenSearchIndex index) {
-    this(cluster, table, index, index.createRequestBuilder(), table.getRowType());
+    this(cluster, table, index, table.getRowType(), null);
```
LantaoJin, line 62: `this(cluster, table, index, table.getRowType(), new PushDownContext());`
and you can remove L74

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOpenSearchIndexScan.java
```diff
@@ -118,17 +138,18 @@ public Enumerator<Object> enumerator() {
     };
   }
 
-  public boolean pushDownFilter(Filter filter) {
+  public CalciteOpenSearchIndexScan pushDownFilter(Filter filter) {
```
LantaoJin, line 141: why change the return type? seems the return value is never in use besides the null-checker in `OpenSearchFilterIndexScanRule`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOpenSearchIndexScan.java
```diff
@@ -143,7 +164,19 @@ public CalciteOpenSearchIndexScan pushDownProject(List<Integer> selectedColumns)
     }
     RelDataType newSchema = builder.build();
     CalciteOpenSearchIndexScan newScan = this.copyWithNewSchema(newSchema);
-    newScan.requestBuilder.pushDownProjectStream(newSchema.getFieldNames().stream());
+    newScan.pushDownContext.add(
+        requestBuilder -> requestBuilder.pushDownProjectStream(newSchema.getFieldNames().stream()));
     return newScan;
   }
+
+  static class PushDownContext extends ArrayDeque<PushDownAction> {
```
LantaoJin, line 172: Could you add some comments about this context. such as purpose and usage

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
```diff
@@ -61,8 +61,13 @@ private void fetchNextBatch() {
...
   public Object current() {
-    Object[] p = fields.stream().map(k -> current.tupleValue().get(k).valueForCalcite()).toArray();
-    return p;
+    /* In Calcite enumerable operators, row of single column will be optimized to a scalar value.
+     * See {@link PhysTypeImpl}
+     */
+    if (fields.size() == 1) {
+      return current.tupleValue().get(fields.getFirst()).valueForCalcite();
+    }
+    return fields.stream().map(k -> current.tupleValue().get(k).valueForCalcite()).toArray();
```
LantaoJin, line 70: The number of `fields` here has been reduced to the actual number of outputs, right? So `get(k)` won't return `null` anymore

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchFilterIndexScanRule.java
qianheng-aws: Sorry, it's typo. It happens the current logic works as well.

```
call.transformTo(scan);
  
->

call.transformTo(newScan);
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
qianheng-aws: yeah

core/src/main/java/org/opensearch/sql/executor/QueryService.java
penghuo: > While we expect that the collation can be preserved through the pipes over PPL

could u elberate more on this? i do not think PPL command should preserved order.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOpenSearchIndexScan.java
```diff
@@ -85,8 +95,10 @@ public RelNode copy(RelTraitSet traitSet, List<RelNode> inputs) {
   @Override
   public void register(RelOptPlanner planner) {
     super.register(planner);
-    for (RelOptRule rule : OpenSearchIndexRules.OPEN_SEARCH_INDEX_SCAN_RULES) {
-      planner.addRule(rule);
+    if (osIndex.getSettings().getSettingValue(Settings.Key.CALCITE_PUSHDOWN_ENABLED)) {
```
penghuo, line 98: is it for test purpose, right? I do not think end-user has any reasons to disable optimization.

qianheng-aws: 1. For example, PPL `source=table | sort a | field a` will have plan `scan -> sort -> project`, we expect the final result should have collation on column `a`.
However, calcite won't ensure that. If we translate the above PPL or plan into SQL, it should be `select a from (select * from table order by a)`. In Calcite, it doesn't ensure collation inner a subquery and will have several optimization to remove the sort operator.
- [remove sort in subquery when converting to RelNode] https://github.com/apache/calcite/blob/a94927e9b80f9f5bf639e31c2636536cb6aebc1a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java#L6454C36-L6454C58
- [set root collation to be empty if top operator is not sort, then the optimizer won't choose a plan having sort operator since it will cost more while we don't have collation requirement for root] https://github.com/apache/calcite/blob/a94927e9b80f9f5bf639e31c2636536cb6aebc1a/core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java#L1055

If we do want to ensure collation of the final result, the PPL should be `source=table | field a | sort a`, which can be translated to SQL `select a from table order by a`. That's why we need add sort operator to ensure collation and rely on optimizer to eliminate the redundant sort for the consideration of simplicity.


2. Yeah, I think so. It's per @LantaoJin's suggestion.

qianheng-aws: 2. And it allows us to disable the push down feature quickly without rollback if there are some unexpected issues.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
LantaoJin: I've pushed the [workaround code](https://github.com/opensearch-project/sql/blob/c2fe0ad4ffa0b5de4fcbbc7c585501241934470e/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java#L67) to dev branch mistakenly. could you merge with the latest branch code to resolve conflicts @qianheng-aws

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOpenSearchIndexScan.java
LantaoJin: @penghuo Adding this configuration is by my suggestion. This is an advanced configuration aimed at developers. Similar to how databases typically have switches for optimization rules. Moreover, this configuration is currently very useful for our development - in some scenarios, enabling push-down might cause issues, and having this switch helps us determine whether problems are caused by push-down or other factors.

LantaoJin: The optimal solution would be to provide a method that allows free selection of optimization rules. However, considering there aren't many custom optimization rules in the short term, adding a push-down config seems to be the most cost-effective approach.

qianheng-aws:
`./gradlew :integ-test:integTest --tests '*Calcite*IT'` successfully on my local. Push down feature has some issues, disabled it to unblock our tests. Will enable once all tests work with it.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOpenSearchIndexScan.java
penghuo: got it, make sense, if Settings.Key.CALCITE_PUSHDOWN_ENABLED default value is enabled, should be fine. let's consider remove it when we are confident.

core/src/main/java/org/opensearch/sql/executor/QueryService.java
penghuo: @qianheng-aws @LantaoJin 
I seem, some PPL command should preserved order, for instance, `fields`, `take`, but others not, `stats`.
In ANSI SQL, `select a from tbl order by b`, order is perserved becuase of logically processing order of select statement is acutall `from -> select -> order`.

PR #3367, opened 2025-02-28
[v3.0.0] Deprecate OpenSearch DSL format

### LINKED ISSUE #3280: Deprecated OpenSearch DSL format.

seankao-az:
### Description
Removes the OpenSearch DSL response format.

This PR focus only on deprecating the format, and will silent the affected legacy test cases. #3372 is created to track the task to decide which of the legacy tests are worth keeping and what to do with them.

Documentation website changes required:
* response_formats.md
* limitation.md
* sql-ppl-api.md

possibly more

### Related Issues
Resolves #3280

### Check List
~~- [ ] New functionality includes testing.~~

~~- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).~~


legacy/src/main/java/org/opensearch/sql/legacy/query/OpenSearchActionFactory.java
```diff
@@ -116,7 +126,7 @@ public static QueryAction create(Client client, QueryActionRequest request)
         } else {
           sqlExpr.accept(new TermFieldRewriter());
           // migrate aggregation to query planner framework.
-          if (shouldMigrateToQueryPlan(sqlExpr, request.getFormat())) {
+          if (!bypassMigrateToQueryPlan && shouldMigrateToQueryPlan(sqlExpr)) {
```
seankao-az, line 129: Previously some test cases which uses Format.JSON to bypasses this check and gets QueryAction for SELECT (either AggregationQueryAction or DefaultQueryAction). After removing Format.JSON, those test case wouldn't work without this flag.

legacy/src/main/java/org/opensearch/sql/legacy/query/OpenSearchActionFactory.java
```diff
@@ -191,12 +201,7 @@ private static boolean isJoin(SQLQueryExpr sqlExpr, String sql) {
   }
 
   @VisibleForTesting
-  public static boolean shouldMigrateToQueryPlan(SQLQueryExpr expr, Format format) {
-    // The JSON format will return the OpenSearch aggregation result, which is not supported by the
-    // QueryPlanner.
-    if (format == Format.JSON) {
-      return false;
-    }
```
dai-chen, line 199: Now there is no JSON format, I'm thinking should we ignore/fix the failed IT instead of adding a flag `bypassMigrateToQueryPlan`?

seankao-az: main problem trying to solve with this flag is `CheckScriptContents.getScriptFieldFromQuery` needs access to a `SearchRequestBuilder` in `QueryAction`, which I think `QueryPlanQueryAction` doesn't expose. Will double check if there's another way

seankao-az: I think for the tests that used this method, it's a simple fix to change the query to not have any aggregation. Doing so will avoid migrating to QueryPlanner as well.

Those are UTs testing the string operators, and the GROUP BY and ORDER BY aren't used anyway, so it should be safe to remove them.

legacy/src/main/java/org/opensearch/sql/legacy/query/OpenSearchActionFactory.java
```diff
@@ -63,16 +63,26 @@ public class OpenSearchActionFactory {
 
   public static QueryAction create(Client client, String sql)
       throws SqlParseException, SQLFeatureNotSupportedException, SQLFeatureDisabledException {
-    return create(client, new QueryActionRequest(sql, new ColumnTypeProvider(), Format.JSON));
+    return create(client, sql, false);
+  }
+
+  public static QueryAction create(Client client, String sql, boolean bypassMigrateToQueryPlan)
```
penghuo: This function only used in UT? if so, add a tag/java-doc on this function.

seankao-az: i removed this in latest commit. It was an attempt to fix some test cases, but it turns out changing the test cases is easier.

penghuo (approved):
Thx!

noCharger (approved):

Issue #3334, opened 2025-02-19
[BUG][Calcite Engine] Incorrect Pushdown of Equality/Inequality (=, !=) on text Fields


qianheng-aws:
**What is the bug?**
PredicateAnalyzer may have potential bug when handling pushdown of Equality/Inequality on text Fields.

**Do you have any additional context?**
Similar issue in opensearch-spark, https://github.com/opensearch-project/opensearch-spark/issues/1056

LantaoJin:
how about predicates expression with `OR`

qianheng-aws:
Since it needs to identify the text type and its subfields, it may has dependency on this issue: https://github.com/opensearch-project/sql/issues/3333

Issue #3373, opened 2025-03-04
[BUG] Some Subquery query return incorrect result when enable pushdown


LantaoJin:
**What is the bug?**

CalcitePPLInSubqueryIT. testInSubqueryWithTableAlias
CalcitePPLInSubqueryIT. testSelfInSubquery

failed when enable pushdown.

qianheng-aws:
tests failed because of `text` field push down, the same issue as https://github.com/opensearch-project/sql/issues/3334, could be addressed by https://github.com/opensearch-project/sql/pull/3376

PR #3376, opened 2025-03-04
[BugFix] support push down text field correctly.

qianheng-aws:
### Description
When pushing filter with text field, I should construct QueryBuilder by using its subfield of keyword type; Otherwise, we cannot push down this field.

This PR includes change:
1. In PredicateAnalyzer, using keyword subfield as the reference for NamedFieldExpression if it's type is text type. It will return null if no such keyword subfield and throw PredicateAnalyzerException, and then scan won't really do filter push down later.
2. remove the @Ignore annotation for `CalciteSortCommandIT` since we already addressed the issue of order perseverance for PPL in this PR: https://github.com/opensearch-project/sql/pull/3350.

`./gradlew :integ-test:integTest --tests '*Calcite*IT'` succeed locally.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3334, https://github.com/opensearch-project/sql/issues/3373

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
@@ -48,6 +48,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexInputRef;
@@ -64,6 +65,9 @@ import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.index.query.RangeQueryBuilder;
 import org.opensearch.sql.calcite.plan.OpenSearchConstants;
+import org.opensearch.sql.opensearch.data.type.OpenSearchDataType;
+import org.opensearch.sql.opensearch.data.type.OpenSearchDataType.MappingType;
+import org.opensearch.sql.opensearch.data.type.OpenSearchTextType;
 
 /**
  * Query predicate analyzer. Uses visitor pattern to traverse existing expression and convert it to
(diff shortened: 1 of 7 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -962,29 +969,58 @@ static boolean isCastExpression(Expression exp) {
...
       this.name = null;
+      this.type = null;
     }
 
-    private NamedFieldExpression(RexInputRef ref, List<String> schema) {
+    private NamedFieldExpression(
+        RexInputRef ref, List<String> schema, Map<String, OpenSearchDataType> typeMapping) {
       this.name =
           (ref == null || ref.getIndex() >= schema.size()) ? null : schema.get(ref.getIndex());
+      this.type = typeMapping.get(name);
```
LantaoJin, line 987: NPE if typeMapping is null

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -962,29 +969,58 @@ static boolean isCastExpression(Expression exp) {
...
 
     String getRootName() {
       return name;
     }
 
+    OpenSearchDataType getOpenSearchDataType() {
+      return type;
+    }
+
+    boolean isTextType() {
```
LantaoJin, line 1003: how about change to `notKeyword()`, `isTextType` looks specific purpose

qianheng-aws: I don't know if there is other case like Text type where we need to find keyword subfield. 

There is similar specific method name in flint, https://github.com/opensearch-project/opensearch-spark/blob/c0c315f010fc1ef4606964e6b34a8ba6fb79949e/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/storage/FlintQueryCompiler.scala#L208, maybe just follow it.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
qianheng-aws: It won't return null for this map from the source, see https://github.com/opensearch-project/sql/blob/482c681ca813e7511654e1f6fe7bf05cbeb759ab/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java#L139.

Issue #3332, opened 2025-02-19
[FEATURE] Calcite Engine Framework: Enable Aggregation Operator Pushdown


qianheng-aws:
Is your feature request related to a problem?
Framework work of the development feature of Calcite Engine.

What solution would you like?
Enable Aggregation Operator Pushdown (pushdown plan to DSL API)

PR #3389, opened 2025-03-06
Calcite enable pushdown aggregation

qianheng-aws:
### Description
This PR include changes:
1. Migrate pushdown process from physical index scan operator to logical index scan operator. The reason is that, in Calcite, some functions(i.e. AVG) only have logical definition while without physical definition or implementation, so we cannot push down such functions for a physical plan.
2. Enable  aggregation pushdown for the general cases
3. Add `VARSAMP` and `VARPOP` implementation. 

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3332

`./gradlew :integ-test:integTest --tests '*Calcite*IT'` succeed locally.

### Check List


LantaoJin:
Please update the latest code and rerun all ITs and UTs locally. [PR](https://github.com/opensearch-project/sql/pull/3388) added and refactored many ITs and UTs.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalOSIndexScan.java
```diff
@@ -0,0 +1,166 @@
...
+import org.opensearch.sql.opensearch.data.type.OpenSearchDataType;
+import org.opensearch.sql.opensearch.planner.physical.EnumerableIndexScanRule;
+import org.opensearch.sql.opensearch.planner.physical.OpenSearchIndexRules;
+import org.opensearch.sql.opensearch.request.AggregateAnalyzer;
+import org.opensearch.sql.opensearch.request.PredicateAnalyzer;
+import org.opensearch.sql.opensearch.response.agg.OpenSearchAggregationResponseParser;
+import org.opensearch.sql.opensearch.storage.OpenSearchIndex;
+
+@Getter
+public class CalciteLogicalOSIndexScan extends CalciteOSIndexScan {
```
LantaoJin, line 41: minor: "OS" is not very convenient for searching classes. (It's fine to use it as a variable name). I think we can remove the "OS" in class name, so does in its parent class. We don't have other type of Index. Or keeping the full name "OpenSearch" also acceptable.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOSIndexScan.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.calcite.plan.RelTraitSet;
+import org.apache.calcite.rel.RelWriter;
+import org.apache.calcite.rel.core.TableScan;
+import org.apache.calcite.rel.hint.RelHint;
+import org.apache.calcite.rel.type.RelDataType;
+import org.opensearch.sql.opensearch.request.OpenSearchRequestBuilder;
+import org.opensearch.sql.opensearch.storage.OpenSearchIndex;
+
+/** Relational expression representing a scan of an OpenSearchIndex type. */
+public abstract class CalciteOSIndexScan extends TableScan {
```
LantaoJin, line 24: ditto. rename to `CalciteOpenSearchIndexScan` or `CalciteIndexScan`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOSIndexScan.java
```diff
@@ -0,0 +1,115 @@
...
+    private boolean isAggregatePushed = false;
+
+    @Override
+    public PushDownContext clone() {
+      return (PushDownContext) super.clone();
+    }
+
+    @Override
+    public boolean add(PushDownAction pushDownAction) {
+      // Defense check. I never do push down to this context after aggregate push-down.
```
LantaoJin, line 73: remove "I"

LantaoJin (commented):
I think we still need UT for pushdown features, it shouldn't be blocked by explain feature. 
We can add a `CalcitePushdownTest` with following code for checking.
```
RelNode finalPlan = Frameworks.getPlanner(getConfig().build()).transform(0, RelTraitSet.createEmpty(), root);
String finalPlanString = RelOptUtil.dumpPlan("", finalPlan, SqlExplainFormat.TEXT, SqlExplainLevel.ALL_ATTRIBUTES);
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOSIndexScan.java
```diff
@@ -0,0 +1,115 @@
...
+      RelOptCluster cluster,
+      RelTraitSet traitSet,
+      List<RelHint> hints,
+      RelOptTable table,
+      OpenSearchIndex osIndex,
+      RelDataType schema,
+      PushDownContext pushDownContext) {
+    super(cluster, traitSet, hints, table);
+    this.osIndex = requireNonNull(osIndex, "OpenSearch index");
+    ;
```
penghuo, line 46: remove it? format checker does not detect it?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalOSIndexScan.java
```diff
@@ -0,0 +1,166 @@
...
+              PushDownType.AGGREGATION,
+              aggregate,
+              requestBuilder -> {
+                requestBuilder.pushDownAggregation(aggregationBuilder);
+                requestBuilder.pushTypeMapping(extendedTypeMapping);
+              }));
+      return newScan;
+    } catch (Exception e) {
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Cannot pushdown the aggregate {}", aggregate, e);
```
penghuo, line 159: any reasons only add stacktrace in debug log?

qianheng-aws: There are too many cases where there is "pushdown failure" and it will report such issue many times since Calcite will hit this rule many times in its optimizing processes.

Hence, we need to avoid log explosion by removing trackstrace in no debugging mode.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteOSIndexScan.java
qianheng-aws: Seems "./gradlew spotlessApply" doesn't fix it.

qianheng-aws:
> I think we still need UT for pushdown features, it shouldn't be blocked by explain feature. We can add a `CalcitePushdownTest` with following code for checking.
> 
> ```
> RelNode finalPlan = Frameworks.getPlanner(getConfig().build()).transform(0, RelTraitSet.createEmpty(), root);
> String finalPlanString = RelOptUtil.dumpPlan("", finalPlan, SqlExplainFormat.TEXT, SqlExplainLevel.ALL_ATTRIBUTES);
> ```

`Frameworks.getPlanner` is not suitable for this purpose. It has state inner and is designed for planning a query from query string. We cannot directly use its `transform` API.

May need further investigation or implementation on getting the final optimized plan.

LantaoJin (approved):

Issue #3407, opened 2025-03-11
[FEATURE] Core Syntax: Implement ppl dedup command with Calcite


LantaoJin:
**Is your feature request related to a problem?**
Core Syntax: Implement ppl dedup command with Calcite

**Do you have any additional context?**
https://github.com/opensearch-project/opensearch-spark/issues/421

Issue #3415, opened 2025-03-13
[FEATURE] Support consecutive `dedup` command with Calcite


LantaoJin:
**Is your feature request related to a problem?**
Currently, the `dedup` command in Calcite is leveraged WindowFunction. But in a Window, the dedup keys are partitioned, so it cannot apply consecutive deduplication.

**What solution would you like?**
If no sort field specified, the OpenSearch response will sort by `_doc`.
To apply consecutive deduplication, we may need a table function to do it.  

**What alternatives have you considered?**

**Do you have any additional context?**

PR #3416, opened 2025-03-13
Implement ppl `dedup` command with Calcite

LantaoJin:
### Description
Implement ppl `dedup` command with Calcite.
Tests under `ppl/src/test/java/org/opensearch/sql/ppl/calcite` passed.
`./gradlew :integ-test:integTest --tests '*Calcite*IT'` passed locally.

This PR leverage Calcite WindowFunction to implement `dedup`. In Window, we partitioned by the dedupe columns, for example, `col1` is a dedupe column, Leveraging window function row_number, we get such a sequence to dedupe.
| col1 | row_number |
| ----- | ----- | 
| A | 1 |
| A | 2 |
| A | 3 |
| B | 1 |
| B | 2 |

But the V2 `dedup` command support "consecutive" argument (If true, remove only consecutive events with duplicate combinations of values). This is out of scope for current solution (which is not supported in PPL Spark either).
Open https://github.com/opensearch-project/sql/issues/3415 to track it.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3407

### Check List


LantaoJin:
@penghuo @qianheng-aws any feedback on this PR?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -447,4 +450,77 @@ public RelNode visitLookup(Lookup node, CalcitePlanContext context) {
...
+            .as("_row_number_");
+    context.relBuilder.projectPlus(rowNumber);
+    RexNode _row_number_ = context.relBuilder.field("_row_number_");
+    // Filter ('_row_number_ <= n)
+    context.relBuilder.filter(
+        context.relBuilder.lessThanOrEqual(
+            _row_number_, context.relBuilder.literal(allowedDuplication)));
+    // DropColumns('_row_number_)
+    context.relBuilder.projectExcept(_row_number_);
+    if (keepEmpty) {
```
qianheng-aws, line 504: Just discussion, how about building a plan like instead of using union? Current implementation will make the original root plan execute twice and make the whole process unefficient.
```
 * | dedup 2 a, b keepempty=false
     * DropColumns('_row_number_)
     * +- Filter ('_row_number_ <= n or a is null or b is null)
     *    +- Window [row_number() windowspecdefinition('a, 'b, 'a ASC NULLS FIRST, 'b ASC NULLS FIRST, specifiedwindowoundedpreceding$(), currentrow$())) AS _row_number_], ['a, 'b], ['a ASC NULLS FIRST, 'b ASC NULLS FIRST]
     *          +- ...
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -460,6 +463,89 @@ public RelNode visitLookup(Lookup node, CalcitePlanContext context) {
...
+       * | dedup 2 a, b keepempty=false
+       * DropColumns('_row_number_)
+       * +- Filter ('_row_number_ <= n OR isnull('a) OR isnull('b))
+       *    +- Window [row_number() windowspecdefinition('a, 'b, 'a ASC NULLS FIRST, 'b ASC NULLS FIRST, specifiedwindowoundedpreceding$(), currentrow$())) AS _row_number_], ['a, 'b], ['a ASC NULLS FIRST, 'b ASC NULLS FIRST]
+       *        +- ...
+       */
+      // Window [row_number() windowspecdefinition('a, 'b, 'a ASC NULLS FIRST, 'b ASC NULLS FIRST,
+      // specifiedwindowoundedpreceding$(), currentrow$())) AS _row_number_], ['a, 'b], ['a ASC
+      // NULLS FIRST, 'b ASC NULLS FIRST]
+      RexNode rowNumber =
```
qianheng-aws, line 493: Many code duplication.

For keepEmpty is `true`, it only needs extra condition for the filter after window

For keepEmpty `false`, it only needs extra filter before window. 

While other part should be the same.

Issue #3420, opened 2025-03-13
[FEATURE] Support IN expression


LantaoJin:
**Is your feature request related to a problem?**
comparisonExpression
   : left = valueExpression comparisonOperator right = valueExpression  # compareExpr
   | valueExpression IN valueList                                       # inExpr
   ;

qianheng-aws:
working on this issue

Issue #3428, opened 2025-03-14
[BUG] calcite like function didn't work as expected


qianheng-aws:
**What is the bug?**
`CalciteWhereCommandIT::testIsNotNullFunction` running failed.

LantaoJin:
@xinyual

penghuo:
> the v2 engine will return true while Calcite return false.

It is a bug in v2 engine. test in spark-sql

```
spark-sql (default)> select "Amber JOHnny" LIKE "Ambe_";
Amber JOHnny LIKE Ambe_
false


spark-sql (default)> select "Amber JOHnny" LIKE "Ambe_%";
Amber JOHnny LIKE Ambe_%
true
```

LantaoJin:
Similar case as `WHERE Like(TextBody, 'test*')` should return false

PR #3429, opened 2025-03-14
[Calcite Engine] Support In expression

qianheng-aws:
### Description
Support In expression, also add data anonymizer for it.

`./gradlew :integ-test:integTest --tests '*Calcite*IT` succeed locally.

`../../../ppl/src/test/java/org/opensearch/sql/ppl/calcite/*` UT succeed locally.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3420

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -147,11 +147,22 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
   @Override
   public RexNode visitIn(In node, CalcitePlanContext context) {
     final RexNode field = analyze(node.getField(), context);
-    return context.rexBuilder.makeIn(
-        field,
-        node.getValueList().stream()
-            .map(value -> analyze(value, context))
-            .collect(Collectors.toList()));
+    final List<RexNode> valueList =
+        node.getValueList().stream().map(value -> analyze(value, context)).toList();
+    final List<RelDataType> dataTypes =
+        new java.util.ArrayList<>(valueList.stream().map(RexNode::getType).toList());
+    dataTypes.add(field.getType());
+    RelDataType commonType = context.rexBuilder.getTypeFactory().leastRestrictive(dataTypes);
+    if (commonType != null) {
+      List<RexNode> newValueList =
+          valueList.stream().map(value -> context.rexBuilder.makeCast(commonType, value)).toList();
+      return context.rexBuilder.makeIn(field, newValueList);
+    } else {
+      throw new SemanticCheckException(
+          StringUtils.format(
+              "In expression types are incompatible: fields type %s, values type %s",
+              dataTypes.getLast(), dataTypes.subList(0, dataTypes.size() - 1)));
+    }
   }
 
   @Override
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -52,7 +52,9 @@ import org.opensearch.sql.ast.expression.subquery.InSubquery;
 import org.opensearch.sql.ast.expression.subquery.ScalarSubquery;
 import org.opensearch.sql.ast.tree.UnresolvedPlan;
 import org.opensearch.sql.calcite.utils.BuiltinFunctionUtils;
+import org.opensearch.sql.calcite.utils.OpenSearchTypeFactory;
 import org.opensearch.sql.common.utils.StringUtils;
+import org.opensearch.sql.data.type.ExprType;
 import org.opensearch.sql.exception.CalciteUnsupportedException;
 import org.opensearch.sql.exception.SemanticCheckException;
 
@@ -158,10 +160,12 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
           valueList.stream().map(value -> context.rexBuilder.makeCast(commonType, value)).toList();
       return context.rexBuilder.makeIn(field, newValueList);
     } else {
+      List<ExprType> exprTypes =
+          dataTypes.stream().map(OpenSearchTypeFactory::convertRelDataTypeToExprType).toList();
       throw new SemanticCheckException(
           StringUtils.format(
               "In expression types are incompatible: fields type %s, values type %s",
-              dataTypes.getLast(), dataTypes.subList(0, dataTypes.size() - 1)));
+              exprTypes.getLast(), exprTypes.subList(0, exprTypes.size() - 1)));
     }
   }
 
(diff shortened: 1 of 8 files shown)


ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
```diff
@@ -386,6 +387,12 @@ public String visitCompare(Compare node, String context) {
       return StringUtils.format("%s %s %s", left, node.getOperator(), right);
     }
 
+    @Override
+    public String visitIn(In node, String context) {
+      String field = analyze(node.getField(), context);
```
qianheng-aws, line 392: Calcite will optimize In expression to equal comparison if there is only 1 element in value list. We don't need to do such optimization ourself here like v2.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -139,6 +140,16 @@ public RexNode visitNot(Not node, CalcitePlanContext context) {
     return context.relBuilder.not(expr);
   }
 
+  @Override
+  public RexNode visitIn(In node, CalcitePlanContext context) {
```
qianheng-aws, line 144: Calcite will optimize In expression to equal comparison if there is only 1 element in value list. We don't need to do such optimization ourself here like v2.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
```diff
@@ -9,12 +9,21 @@
...
   public void init() throws IOException {
     enableCalcite();
     disallowCalciteFallback();
     super.init();
   }
+
+  @Ignore("https://github.com/opensearch-project/sql/issues/3428")
+  @Override
+  public void testIsNotNullFunction() throws IOException {}
+  ;
```
LantaoJin, line 23: remove this `;`

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
```diff
@@ -154,6 +154,11 @@ public void testNotExpression() {
     assertEquals("source=t | where not a = ***", anonymize("source=t | where not a=1 "));
   }
 
+  @Test
+  public void testInExpression() {
+    assertEquals("source=t | where a in (***)", anonymize("source=t | where a in (1, 2, 3) "));
```
LantaoJin, line 159: add a test for
```
where a not in (1, 2, 3)
```

integ-test/src/test/java/org/opensearch/sql/ppl/WhereCommandIT.java
```diff
@@ -106,4 +106,20 @@ public void testWhereWithMetadataFields() throws IOException {
             String.format("source=%s | where _id='1' | fields firstname", TEST_INDEX_ACCOUNT));
     verifyDataRows(result, rows("Amber"));
   }
+
+  @Test
+  public void testWhereWithIn() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | where firstname in ('Amber') | fields firstname", TEST_INDEX_ACCOUNT));
```
LantaoJin, line 115: need a test for
```
where age in (36, 36.0, '36')
```

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
qianheng-aws: "not in" is not supported in our g4, which mean v2 engine don't support it either.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
```diff
@@ -9,12 +9,25 @@
...
-@Ignore("Not all boolean functions are supported in Calcite now")
 public class CalciteWhereCommandIT extends WhereCommandIT {
   @Override
   public void init() throws IOException {
     enableCalcite();
     disallowCalciteFallback();
     super.init();
   }
+
+  @Ignore("https://github.com/opensearch-project/sql/issues/3428")
```
penghuo, line 20: https://github.com/opensearch-project/sql/issues/3428 is LIKE function, why releated to testIsNotNull?

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
penghuo: > We don't need to do such optimization ourself here like v2.

Do u mean we push down IN to OpenSearch as terms query, so Calcite optimized is not needed?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
```diff
@@ -9,12 +9,25 @@
...
+  public void testIsNotNullFunction() throws IOException {}
+
+  @Ignore("https://github.com/opensearch-project/sql/issues/3333")
+  @Override
+  public void testWhereWithMetadataFields() throws IOException {}
+
+  @Override
+  protected String getIncompatibleTypeErrMsg() {
+    return "In expression types are incompatible: fields type BIGINT, values type [INTEGER,"
+        + " INTEGER, CHAR(4)]";
```
penghuo, line 31: CHAR(4) is not a valid OpenSearch PPL datatype.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
```diff
@@ -9,12 +9,25 @@
 import org.junit.Ignore;
 import org.opensearch.sql.ppl.WhereCommandIT;
 
-@Ignore("Not all boolean functions are supported in Calcite now")
 public class CalciteWhereCommandIT extends WhereCommandIT {
   @Override
   public void init() throws IOException {
     enableCalcite();
```
penghuo, line 15: should we disableCalcite also? otherwise it may impact other tests?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
qianheng-aws: this test case contains a like function as well.

`"source=%s | where isnotnull(age) and like(firstname, 'Ambe_') | fields firstname"`

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
qianheng-aws: 1. No, I mean v2 has optimization to transform `column in (a, ,b ,c)` to `column = a or column = b or column = c` when converting. Since calcite will do such optimization in its optimization phase, we just need to convert it to the former expression directly without too much consideration.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
qianheng-aws: We have `exclude 'org/opensearch/sql/calcite/remote/**' in build.gradle, it won't really run in CI.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
LantaoJin: `WhereCommandIT` is an existed ppl IT which is extended by `CalciteWhereCommandIT`

LantaoJin (approved):

PR #3439, opened 2025-03-17
Support ppl CAST function with Calcite

### LINKED ISSUE #3417: [FEATURE] Migrate existing CAST function with Calcite engine
```
// cast function
dataTypeFunctionCall
   : CAST LT_PRTHS expression AS convertedDataType RT_PRTHS
   ;
```

LantaoJin:
### Description
Support ppl CAST function with Calcite

### Related Issues
Resolves #3417 

### Check List


core/src/main/java/org/opensearch/sql/ast/expression/DataType.java
```diff
@@ -26,7 +26,9 @@ public enum DataType {
   DATE(ExprCoreType.DATE),
   TIME(ExprCoreType.TIME),
   TIMESTAMP(ExprCoreType.TIMESTAMP),
-  INTERVAL(ExprCoreType.INTERVAL);
+  INTERVAL(ExprCoreType.INTERVAL),
+
+  IP(ExprCoreType.IP);
```
LantaoJin, line 31: Cast_to_IP unsupported yet since IP type is not supported with Calcite. But this is a missing code in DataType class.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -358,14 +360,21 @@ private RelNode resolveSubqueryPlan(UnresolvedPlan subquery, CalcitePlanContext
...
   public RexNode visitCast(Cast node, CalcitePlanContext context) {
-    throw new CalciteUnsupportedException("CastWhen function is unsupported in Calcite");
+    RexNode expr = analyze(node.getExpression(), context);
+    SqlTypeName sqlTypeName =
+        OpenSearchTypeFactory.convertExprTypeToRelDataType(node.getDataType().getCoreType())
+            .getSqlTypeName();
+    RelDataType type = context.rexBuilder.getTypeFactory().createSqlType(sqlTypeName);
+    RelDataType nullableType =
+        context.rexBuilder.getTypeFactory().createTypeWithNullability(type, true);
+    return context.rexBuilder.makeCast(SqlParserPos.ZERO, nullableType, expr, false, true);
```
LantaoJin, line 372: use this 4 lines instead of `context.relbuilder.cast()` because we need the last parameter to `true`

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLCastFunctionIT.java
```diff
@@ -0,0 +1,151 @@
...
+                TEST_INDEX_STATE_COUNTRY));
+
+    verifySchema(actual, schema("age", "string"));
+
+    verifyDataRows(actual, rows("70"), rows("30"), rows("25"), rows("20"));
+  }
+
+  @Test
+  public void testCastToUnknownType() {
+    assertThrows(
```
penghuo, line 55: what is error message?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLCastFunctionIT.java
```diff
@@ -0,0 +1,151 @@
...
+
+  @Test
+  public void testCastNullValues() {
+    String actual =
+        execute(
+            String.format(
+                "source=%s | eval a = cast(state as string) | fields a",
+                TEST_INDEX_STATE_COUNTRY_WITH_NULL));
+
+    assertEquals(
```
penghuo, line 86: why not verifyDataRows?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLCastFunctionIT.java
```diff
@@ -0,0 +1,151 @@
...
+            + "}",
+        actual);
+  }
+
+  @Test
+  public void testCastToUnsupportedType() {
+    String actual =
+        execute(
+            String.format(
+                "source=%s | eval a = cast(name as boolean) | fields a", TEST_INDEX_STATE_COUNTRY));
```
penghuo, line 122: cast("1" as boolean) should return True?

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
penghuo: does Calcite has same [Data Type Conversion rule](https://github.com/opensearch-project/sql/blob/main/docs/user/general/datatypes.rst#id4)? anything we want to call-out and update doc?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLCastFunctionIT.java
LantaoJin: in V2, the cast function has a `convertedType` list.
https://github.com/opensearch-project/sql/blob/f805df0ab8796de5aefc076308dccf0a808244c4/ppl/src/main/antlr/OpenSearchPPLParser.g4#L407-L420

Any unsupported type will throw SyntaxCheckException, no matter what the message is.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLCastFunctionIT.java
LantaoJin: I don't know how to check null row with `verifyDataRows`
The method `rows()` cannot work with null.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLCastFunctionIT.java
LantaoJin: no, only 'true', 'True', 'TRUE' etc could return `True` in cast_to_boolean.
cast("1" as boolean), cast(1 as boolean), cast(2.5 as boolean) are all return `null`.

This seems a breaking change comparing to V2:
cast("1" as boolean) returns false
cast(1 as boolean), cast(2.5 as boolean) return true
cast(0 as boolean) returns false

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
LantaoJin: I will add more test for this table

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLCastFunctionIT.java
LantaoJin: Em, I think this is a bug of Calcite, we can fix it in our side.

LantaoJin: https://github.com/opensearch-project/sql/issues/3443

LantaoJin:
> does Calcite has same [Data Type Conversion rule](https://github.com/opensearch-project/sql/blob/main/docs/user/general/datatypes.rst#id4)? anything we want to call-out and update doc?

Call-out:
1. cast_to_boolean behaviour is different with V2. But we can discuss which would we prefer to, see #3443 
2. safe-cast behaviour is different with V2.  Cast will return null instead of throwing exception. But we can discuss which would we prefer to. (Spark/postgre use safe-cast)
3. cast boolean to float/double trigger a Calcite bug. See the new tests added in `CalcitePPLCastFunctionIT`

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLCastFunctionIT.java
LantaoJin: The latest commit follows the behaviour of SparkSQL. Check above

LantaoJin:
Failure caused by a flaky test `CalcitePPLStringBuiltinFunctionIT > testPosition FAILED`, Merging to dev branch.

PR #3440, opened 2025-03-17
Add direct-query-core module for prometheus integration

joshuali925:
### Description
This PR adds direct-query-core module to SQL. It implements the following changes
- move some shared classes from prometheus module to direct-query-core
   - prometheus client
   - client exception
   - metric metadata model
   - test fixtures
- add direct-query-core module to support `ExecuteDirectQuery` and `GetDirectQueryResources` actions
- implement query handler for executing the actions. currently the only supported data source type is `prometheus`

This PR does not expose the actions in SQL plugin at runtime. A follow up PR on `direct-query` module will add the rest and transport actions to SQL

It is currently in draft since some tests are not fixed yet

### Related Issues

### Check List


direct-query-core/src/main/java/org/opensearch/sql/directquery/DirectQueryExecutorServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+    String dataSourceType = null;
+    String result;
+
+    try {
+      // Get data source type before creating the client
+      dataSourceType =
+          dataSourceClientFactory.getDataSourceType(dataSourceName).name().toLowerCase();
+
+      // Let Java infer the type
+      var client = dataSourceClientFactory.createClient(dataSourceName);
```
joshuali925, line 46: i don't think java will infer the type, remove the unnecessary comments?

direct-query-core/src/main/java/org/opensearch/sql/directquery/DirectQueryExecutorServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+      // No need to specify Object type parameter, let the registry infer it
+      result =
+          queryHandlerRegistry
+              .getQueryHandler(client)
+              .map(
+                  handler -> {
+                    try {
+                      return handler.executeQuery(client, request);
+                    } catch (IOException e) {
+                      return "{\"error\": \"Error executing query: " + e.getMessage() + "\"}";
```
joshuali925, line 57: if prometheus returns an error, should the SQL request fail with non-200 response?

direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    String queryString =
+        params.entrySet().stream()
+            .map(
+                entry ->
+                    URLEncoder.encode(entry.getKey(), StandardCharsets.UTF_8)
+                        + "="
+                        + URLEncoder.encode(entry.getValue(), StandardCharsets.UTF_8))
+            .collect(Collectors.joining("&"));
```
joshuali925, line 101:
```suggestion
    String queryString = this.paramsToQueryString(params);
```

direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
```diff
@@ -0,0 +1,287 @@
...
+        String.format("%s/api/v1/query?%s", uri.toString().replaceAll("/$", ""), queryString);
+
+    logger.info("Making Prometheus instant query request: {}", queryUrl);
+    Request request = new Request.Builder().url(queryUrl).build();
+
+    logger.info("Executing Prometheus request with headers: {}", request.headers().toString());
+    Response response = this.okHttpClient.newCall(request).execute();
+
+    logger.info("Received Prometheus response for instant query: code={}", response);
+    // Return the full response object, not just the data field
```
joshuali925, line 113: what does the comment mean?

direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
```diff
@@ -0,0 +1,287 @@
...
+      params.put("timeout", timeout.toString());
+    }
+
+    return params.entrySet().stream()
+        .map(
+            entry ->
+                URLEncoder.encode(entry.getKey(), StandardCharsets.UTF_8)
+                    + "="
+                    + URLEncoder.encode(entry.getValue(), StandardCharsets.UTF_8))
+        .collect(Collectors.joining("&"));
```
joshuali925, line 285: also reuse the `paramsToQueryString` here

direct-query-core/src/main/java/org/opensearch/sql/prometheus/exception/PrometheusClientException.java
```diff
@@ -1,11 +1,9 @@
...
- *  * SPDX-License-Identifier: Apache-2.0
- *
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
  */
 
-package org.opensearch.sql.prometheus.exceptions;
+package org.opensearch.sql.prometheus.exception;
 
 import org.opensearch.sql.datasources.exceptions.DataSourceClientException;
```
joshuali925, line 8: this should be `package org.opensearch.sql.datasource.client.exceptions.DataSourceClientException;`? i think we should remove one of the two DataSourceClientException

RyanL1997:
Hi @joshuali925 , do we still need this?

joshuali925:
we do, but currently no bandwidth

goyamegh:
> we do, but currently no bandwidth

@lezzago and I will take over this and take it further. Thanks.

direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/ExecuteDirectQueryRequest.java
```diff
@@ -0,0 +1,55 @@
...
+  // Required fields
+  private String dataSources; // Required: From URI path parameter or request body
+  private String query; // Required: String for Prometheus, object for CloudWatch
+  @Setter private LangType language; // Required: SQL, PPL, or PROMQL
+  private String sourceVersion; // Required: API version
+
+  // Optional fields
+  private Integer maxResults; // Optional: limit for Prometheus, maxDataPoints for CW
+  private Integer timeout; // Optional: number of seconds
+  private DataSourceOptions options; // Optional: Source specific arguments
```
lezzago, line 28: Would we want a separate queryOptions or is the idea to have this contain query options as those are in a way specific to a datasource.

direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/ExecuteDirectQueryResponse.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class ExecuteDirectQueryResponse {
+  private String queryId;
+  private String result;
+  private String sessionId;
```
lezzago, line 18: Is this needed if we have the queryId? What benefit is there to add another tracker if we can just use queryId everywhere instead.

joshuali925: i remember session id is for EMR job control, and different queries might reuse the same EMR job. I might be wrong but i agree direct (sync) query shouldn't have session concept.

will leave it to query API owner @goyamegh

direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/ExecuteDirectQueryRequest.java
joshuali925: can't remember the details, will leave it to query API owner @goyamegh

currently this is the only usage https://github.com/opensearch-project/sql/blob/ac6059fb760a9383e0ce1057dfad0cc752c27f70/direct-query-core/src/main/java/org/opensearch/sql/prometheus/model/PrometheusOptions.java#L15-L21
does look like query options

PR #3456, opened 2025-03-20
Fix flaky tests

yuancu:
### Description
-  Keep generated code from spotless check (inside benchmark & ppl)
- Fix flaky tests: `testSubstring`, `testPosition`, `testLike`

### Check List

### DIFF
--- a/benchmarks/build.gradle
+++ b/benchmarks/build.gradle
@@ -20,4 +20,13 @@ dependencies {
     annotationProcessor group: 'org.openjdk.jmh', name: 'jmh-generator-annprocess', version: '1.36'
 }
 
+spotless {
+    java {
+        target fileTree('.') {
+            include '**/*.java'
+            exclude '**/jmh_generated/**'
+        }
+    }
+}
+
 compileJava.options.compilerArgs.addAll(["-processor", "org.openjdk.jmh.generators.BenchmarkProcessor"])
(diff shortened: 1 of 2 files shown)


benchmarks/build.gradle
```diff
@@ -20,4 +20,13 @@ dependencies {
     annotationProcessor group: 'org.openjdk.jmh', name: 'jmh-generator-annprocess', version: '1.36'
 }
 
+spotless {
```
LantaoJin, line 23: is this change required for this issue?

ppl/build.gradle
```diff
@@ -66,7 +66,7 @@ spotless {
     java {
         target fileTree('.') {
             include '**/*.java'
-            exclude '**/build/**', '**/build-*/**'
+            exclude '**/build/**', '**/build-*/**', 'src/main/gen/**'
```
LantaoJin, line 69: ditto

benchmarks/build.gradle
yuancu: No. They led to build failures, so I made the changes. Should I revert them?

LantaoJin (approved):

benchmarks/build.gradle
LantaoJin: It’s ok

Issue #3452, opened 2025-03-20
[FEATURE] Support nested (object) fields with Calcite


LantaoJin:
**Is your feature request related to a problem?**
`org.opensearch.sql.ppl.ObjectFieldOperateIT` should work with Calcite

Issue #3459, opened 2025-03-21
[FEATURE][Calcite Engine] Support access to nested fields after command `FIELD`


qianheng-aws:
**Is your feature request related to a problem?**
Need to support access nested fields after command `FIELD`.

PPL like below, although it has projection only includes field `city`, we still expect to use nested field `city.name ` since it's nested in field `city`.

`source=sql_test_index__deep_nested | fields city | sort city.name`

**Background:**
PPL will expand and flatten all nested fields in it's table schema, so we can access them easily.

PR #3471, opened 2025-03-24
Add missing text function

xinyual:
### Description
Add missing text function including
ascii/left/ locate/strcmp, substr

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3467

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3476, opened 2025-03-25
[Calcite engine] Support nested field

qianheng-aws:
### Description
Support nested field

### Implementation options
- option1(This PR): For opensearch index, since we already have nested fields flattened and accessible in the mapping information,  we just have to fill the flatten nested fields with its related value, then it could support access to the nested fields as well as get the correct value.

- option2: https://github.com/opensearch-project/sql/issues/3459 

### Related Issues
Resolves  https://github.com/opensearch-project/sql/issues/3452

### Check List


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -160,6 +163,23 @@ public RelNode visitProject(Project node, CalcitePlanContext context) {
     return context.relBuilder.peek();
   }
 
+  /** See logic in {@link org.opensearch.sql.analysis.symbol.SymbolTable#lookupAllFields} */
```
LantaoJin, line 166: there is a bug in V2. could you check is it existing in this pr? https://github.com/opensearch-project/sql/issues/3477

qianheng-aws: Has a new discovery for that issue: https://github.com/opensearch-project/sql/issues/3477#issuecomment-2750194736

LantaoJin: Thanks for correcting. Anyway, my point is for the field name with dot. But seems the field with dot in OpenSearch are all treated as nested type.

LantaoJin (approved):

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
penghuo: Add comments, we should handle this case in future. https://github.com/opensearch-project/sql/issues/3477#issuecomment-2752605994

integ-test/src/test/java/org/opensearch/sql/ppl/ObjectFieldOperateIT.java
```diff
@@ -52,7 +52,10 @@ public void group_object_field_in_stats() throws IOException {
     JSONObject result =
         executeQuery(
             String.format("source=%s | stats count() by city.name", TEST_INDEX_DEEP_NESTED));
-    verifySchema(result, schema("count()", "integer"), schema("city.name", "string"));
+    verifySchema(
+        result,
+        schema("count()", isCalciteEnabled() ? "long" : "integer"),
```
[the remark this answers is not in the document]
qianheng-aws: resolved

penghuo (approved):

LantaoJin (approved):
failures not related, due to #3471 .merging to main.

Issue #3478, opened 2025-03-25
[FEATURE] Calcite Engine: Support timestamp type


qianheng-aws:
**Is your feature request related to a problem?**
OpenSearch timestamp type is different from other databases including Calcite.

In Calcite, timestamp is calculated as long while don't support nanos precision.

In OpenSearch, timestamp is calculated as string and support nanos precision. In V2, we support comparison between string and columns of timestamp type, and timestamp related functions also support pass in string as its arguments.

**What solution would you like?**
Convert opensearch timestamp as UDT instead of calcite's timestamp

**What alternatives have you considered?**

**Do you have any additional context?**

PR #3483, opened 2025-03-26
[Calcite Engine]Support UDT for date, time, timestamp

qianheng-aws:
### Description
Support UDT for date, time, timestamp, this PR also includes change for refactoring span function since we change these types into UDT.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3478

### Check List


ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAggregationTest.java
```diff
@@ -229,6 +229,7 @@ public void testAvgBySpanAndFields() {
     verifyPPLToSparkSQL(root, expectedSparkSql);
   }
 
+  @Ignore("Span doesn't support type of calcite date")
```
LantaoJin, line 232: Does UDT break the principle of "All PPL could be translated to Spark SQL via Calcite"?

core/src/main/java/org/opensearch/sql/calcite/type/ExprUserDefinedType.java
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.type;
+
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFamily;
+import org.opensearch.sql.data.type.ExprType;
+
+public interface ExprUserDefinedType extends RelDataType, RelDataTypeFamily {
```
LantaoJin, line 12: Can you rename it to `ExprRelDataType` to keep the RelDataType suffix?

core/src/main/java/org/opensearch/sql/calcite/type/ExprBasicSqlUDT.java
```diff
@@ -0,0 +1,362 @@
...
+ * ExprBasicSqlUDT represents a standard atomic SQL type (excluding interval types) for a {@link
+ * ExprType}.
+ *
+ * <p>Instances of this class are immutable.
+ *
+ * <p>Compared to Calcite's BasicSqlType, this class supports flexible inheritance of UDTs.
+ *
+ * <p>TODO: may no need to be abstract if it can cover all functions of subclass.
+ */
+public abstract class ExprBasicSqlUDT extends AbstractSqlType implements ExprUserDefinedType {
```
LantaoJin, line 54: `ExprBasicSqlUDT` rename to `ExprSqlType`

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -134,49 +187,34 @@ public static RelDataType convertExprTypeToRelDataType(ExprType fieldType, boole
...
+            INTERVAL_HOUR,
+            INTERVAL_MINUTE,
+            INTERVAL_SECOND -> INTERVAL;
+        case ARRAY -> ARRAY;
+        case MAP -> STRUCT;
+        case NULL -> UNDEFINED;
+        default -> throw new IllegalArgumentException(
             "Unsupported conversion for Relational Data type: " + type.getSqlTypeName());
-    }
+      };
```
LantaoJin, line 217: remove `;`

integ-test/src/test/java/org/opensearch/sql/calcite/remote/fallback/CalciteTextFunctionIT.java
```diff
@@ -13,4 +15,14 @@ public void init() throws Exception {
...
+
+  @Ignore("https://github.com/opensearch-project/sql/issues/3481")
+  @Override
+  public void testStrcmp() throws IOException {}
+  ;
+
+  @Ignore("https://github.com/opensearch-project/sql/issues/3481")
+  @Override
+  public void testLocate() throws IOException {}
+  ;
```
LantaoJin, line 27: can remove now

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
qianheng-aws: This symbol is required

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAggregationTest.java
qianheng-aws: It doesn't break it. This test didn't work because we expect span function only support our UDT of date, time, timestamp but not support calcite's date.

It keeps the same behavior with v2 engine since it doesn't have a type of calcite's date. It we want to extend span function to support as well, it could be done in the future since we don't have some requirement for now because the only datasource is opensearch index.

penghuo (approved):

LantaoJin (approved):

PR #3390, opened 2025-03-06
Add math udf

xinyual:
### Description
This PR add math built-in functions. 
And also add corresponding IT for them. 
Due to the lack of IT in spark PPL, we also implement some by ourselves.
Functions | Covered by [Spark IT](https://github.com/opensearch-project/opensearch-spark/blob/main/integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLBuiltinFunctionITSuite.scala) | Covered by IT
-- | -- | --
abs | Y | Y
acos |   | Y
asin |   | Y
atan |   | Y
atan2 |   | Y
cbrt |   | Y
ceiling | Y | Y
conv |   | Y
cos |   | Y
cot |   | Y
crc32 |   | Y
degrees |   | Y
e |   | Y
exp |   | Y
floor | Y | Y (Modified)
ln | Y | Y
log |   | Y
log2 |   | Y
log10 |   | Y
mod | Y | Y
pi |   | Y
pow | Y | Y
radians |   | Y
rand |   | Y
round |   | Y
sign |   | Y
sin |   | Y
sqrt | Y | Y

The integration tests covered by  Spark are copied to this IT (Apart from `floor`). The rest is implemented by ourselves.
./gradlew :integ-test:integTest --tests '*Calcite*IT' succeed in local

### Related Issues
Resolves #3398 

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3513, opened 2025-04-03
Fix divide function with a UDF

yuancu:
### Description
Fix `/` function to make it return double values.

### Related Issues
Relevant to #3390 

### Check List
 - [] New functionality has javadoc added.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/udf/mathUDF/DivideFunction.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/udf/mathUDF/DivideFunction.java
@@ -1,0 +1,18 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.udf.mathUDF;
+
+import org.opensearch.sql.calcite.udf.UserDefinedFunction;
+
+public class DivideFunction implements UserDefinedFunction {
+
+  @Override
+  public Object eval(Object... args) {
+    double dividend = ((Number) args[0]).doubleValue();
+    double divisor = ((Number) args[1]).doubleValue();
+    return dividend / divisor;
+  }
+}
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLBuiltinFunctionIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLBuiltinFunctionIT.java
@@ -377,4 +377,14 @@ public class CalcitePPLBuiltinFunctionIT extends CalcitePPLIntegTestCase {
         actual, schema("name", "string"), schema("age", "integer"), schema("thirty_one", "double"));
     verifyDataRows(actual, rows("Hello", 30, 31));
   }
+
+  @Test
+  public void testDivide() {
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | where name = 'Hello' | eval res1 = 22 / 7, res2 = age / 11 | fields res1", TEST_INDEX_STATE_COUNTRY));
+    verifySchema(actual, schema("res1", "double"));
+    verifyDataRows(actual, closeTo(3.142857142857143, 2.727272727272727));
+  }
 }
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/BuiltinFunctionUtils.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/BuiltinFunctionUtils.java
@@ -90,6 +90,7 @@ import org.opensearch.sql.calcite.udf.datetimeUDF.YearFunction;
 import org.opensearch.sql.calcite.udf.datetimeUDF.YearWeekFunction;
 import org.opensearch.sql.calcite.udf.mathUDF.CRC32Function;
 import org.opensearch.sql.calcite.udf.mathUDF.ConvFunction;
+import org.opensearch.sql.calcite.udf.mathUDF.DivideFunction;
 import org.opensearch.sql.calcite.udf.mathUDF.EulerFunction;
 import org.opensearch.sql.calcite.udf.mathUDF.ModFunction;
 import org.opensearch.sql.calcite.udf.mathUDF.SqrtFunction;
@@ -138,7 +139,7 @@ public interface BuiltinFunctionUtils {
       case "*":
         return SqlStdOperatorTable.MULTIPLY;
       case "/":
-        return SqlStdOperatorTable.DIVIDE;
+        return TransferUserDefinedFunction(DivideFunction.class, "/", ReturnTypes.DOUBLE);
         // Built-in String Functions
       case "ASCII":
         return SqlStdOperatorTable.ASCII;


core/src/main/java/org/opensearch/sql/calcite/utils/MathUtils.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.utils;
+
+public class MathUtils {
+  public static final double EPSILON = 0.0000001;
```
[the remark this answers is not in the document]
yuancu: > Just wonder where does this magic number come from initially?

It's a little arbitrary: I played with the MOD function in V2, checking until which floating point it will return a null, e.g. 0.00001, 0.000001. Then I take the threshold as this epsilon value.

But I haven't tested whether DIVIDE also follows the same value. I'll double check.

yuancu:
> I think not all changes are not covered in the PR description?

I updated the description to explain what else I modified. Thanks for reminding me!

core/src/main/java/org/opensearch/sql/calcite/udf/mathUDF/ModFunction.java
```diff
@@ -32,33 +33,22 @@ public Object eval(Object... args) {
...
       long l0 = num0.longValue();
       long l1 = num1.longValue();
       // It returns negative values when l0 is negative
       long result = l0 % l1;
       // Return the wider type between l0 and l1
-      if (num0 instanceof Long || num1 instanceof Long) {
-        return result;
-      }
-      return (int) result;
+      return MathUtils.coerceToWidestIntegralType(num0, num1, result);
```
Swiddis, line 46: question: is this correct?

I'd expect we coerce it to the smallest type (e.g. short % byte == byte, byte % short == byte) due to the nature of mod. I guess it's up to how strongly-typed PPL is. I'm not sure where the type difference matters in practice for PPL, but in other languages I dislike needing to write things along the lines of `byte modded = (byte) (intVal % 64)` if the cast is mathematically redundant.

core/src/main/java/org/opensearch/sql/calcite/udf/mathUDF/DivideFunction.java
```diff
@@ -0,0 +1,33 @@
...
+    if (Math.abs(divisor.doubleValue()) < MathUtils.EPSILON) {
+      return null;
+    }
+
+    double result = dividend.doubleValue() / divisor.doubleValue();
+    if (MathUtils.isIntegral(dividend) && MathUtils.isIntegral(divisor)) {
+      return MathUtils.coerceToWidestIntegralType(dividend, divisor, (long) result);
+    }
+    return MathUtils.coerceToWidestFloatingType(dividend, divisor, result);
+  }
```
Swiddis, line 32: question: Why compute with double first?

Wouldn't it be more accurate (avoiding potential FPA) to do a different computation for each branch, instead of coercing to double and then back to long?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLBuiltinFunctionIT.java
```diff
@@ -377,4 +380,62 @@ public void testSignAndRound() {
...
+            0.8225806704669051));
+  }
+
+  @Test
+  public void testDivideShouldReturnNull() {
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | where key = 'null' | head 1 | eval r2 = 4 / dbl, r3 = `int` / 5, r4 ="
+                    + " 22 / 0, r5 = 22.0 / 0, r6 = 22.0 / 0.0 | fields r2, r3, r4, r5, r6",
```
Swiddis, line 430: suggestion: Shouldn't we be following [IEEE 754's convention for division by 0?](https://en.wikipedia.org/wiki/IEEE_754#Exception_handling)

If not, we should clarify in [the docs](https://github.com/opensearch-project/sql/blob/920f7115f3322d78bcff58491406e83af809227d/docs/user/ppl/functions/expressions.rst#operators).

Swiddis (commented):
Some thoughts, one suggestion I'd like to see either revised or documented

core/src/main/java/org/opensearch/sql/calcite/udf/mathUDF/DivideFunction.java
yuancu: My thought was that double operations will return an equally accurate value as integral operations and have a wider representation range than integral numbers (e.g. `Double.MAX_VALUE > Long.MAX_VALUE`)

But you are right. I can not rule out the possibility where a double representing an integral number is stored slighted smaller than its true integral value, and then truncated to another integral number 1 digit smaller. I'll fix it.

LantaoJin:
> CalcitePPLDateTimeBuiltinFunctionIT > testDateFormatAndDatetimeAndFromDays FAILED

Could you check this?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLBuiltinFunctionIT.java
```diff
@@ -377,4 +380,62 @@ public void testSignAndRound() {
...
+            2,
+            3,
+            1,
+            1.1774194,
+            2.0666666,
+            3.142857142857143,
+            3.142857142857143,
+            3.0,
+            2.4333334,
+            0.8225806704669051));
```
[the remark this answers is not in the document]
yuancu: It turns out that the returned representation from calcite is already truncated. See [OpenSearchExecutionEngine.java#L147](https://github.com/opensearch-project/sql/blob/d9ef4fa070aa240b2b103e5758fbfad77067199b/opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java#L147) and [JdbcOpenSearchDataTypeConvertor.java#L67](https://github.com/opensearch-project/sql/blob/d9ef4fa070aa240b2b103e5758fbfad77067199b/opensearch/src/main/java/org/opensearch/sql/opensearch/util/JdbcOpenSearchDataTypeConvertor.java#L67).

In the line 67
```
Object value = rs.getObject(i);
 ```
The value for e.g. `7.3 / 6.2` is returned as `Float(1.1774194)`. Therefore, it will fail tests if the expected value is set to the precise computed results (around `1.1774193548387095` in this case).

That was why I used the literal values which exactly match its actual result in tests.

This is also how previous versions before introducing calcite behaves (opensearch-sql 2.19.1.0). 

IMO, it might not to the user's interests whether they are shown a more precise value with an error smaller to 1e-10, or a truncated value with an error around 1e-5.

Admittedly, the current implementation is not resilient to changes in underlying implementations. But I can not come up with a better solution than using literal values in tests. Please feel free to leave your thoughts!

**Comment by yuancu** [comment]
> > CalcitePPLDateTimeBuiltinFunctionIT > testDateFormatAndDatetimeAndFromDays FAILED
> 
> Could you check this? @yuancu

I think it has been resolved by @xinyual in issue #3516 . I'll double check.

File: integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLBuiltinFunctionIT.java (hunk 4 above)
Comment (yuancu) [reply integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLBuiltinFunctionIT.java:430]: > suggestion: Shouldn't we be following [IEEE 754's convention for division by 0?](https://en.wikipedia.org/wiki/IEEE_754#Exception_handling)
> 
> If not, we should clarify in [the docs](https://github.com/opensearch-project/sql/blob/920f7115f3322d78bcff58491406e83af809227d/docs/user/ppl/functions/expressions.rst#operators).

The existing implementation of PPL does not support Infinity yet -- It will return null when the divisor is 0, and will throw the following error when the result overflows what the corresponding numeric type can represent:

```
{
  "error": {
    "reason": "Invalid Query",
    "details": "Infinity is not a valid double value as per JSON specification. To override this behavior, use GsonBuilder.serializeSpecialFloatingPointValues() method.",
  },
  "status": 400
}
```

For the time being, I'll clarify this in the  document. Should we create a new issue on whether to support +- INFINITY in PPL?

core/src/main/java/org/opensearch/sql/calcite/udf/mathUDF/ModFunction.java
yuancu: Intuitively, it is indeed efficient to coerce the return type to the smallest one, since the result won't be wider than any of the operands. But I did found any database or programming language coerce MOD's return type this way.

For example
- MYSQL and JAVA prompts return types to INTEGER when operands are byte / short, and coerce return types to BIGINT when any of the operands is BIGINT.
- PostgreSQL and opensearch-sql v2.x adopt the common super type of the operands, similar to our implementations.

I assume it might because that even though the final result of MOD fits in a smaller type, the intermediate calculation could overflow the narrow type.

LantaoJin (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLBuiltinFunctionIT.java
Swiddis: If it's already the existing behavior I think it's fine without an issue, just documenting is sufficient. Just noticed it since division was on-topic and I know FPA has some weird behaviors.

Swiddis (approved):

PR #3515, opened 2025-04-03
Fix: CSV handling of embedded crlf

### LINKED ISSUE #3514: [BUG] CSV output does not handle entries with CRLF
**What is the bug?**
CSV output of entries like a\nb or a\r\nb should, according to RFC 4180, escape them to be "a\nb" and "a\r\nb", respectively, respectively. See section 2.5, https://www.rfc-editor.org/rfc/rfc4180

1. Add values to a test index
```
#!/usr/bin/env bash

# Assuming OPENSEARCH_SERVER is the server,
# OPENSEARCH_USER is the host, OPENSEARCH_PASSWORD is the password.
# and the user has the rights to create and query this index.

# Add an entry to the index that has a newline in it.
curl -XPUT "https://${OPENSEARCH_SERVER}:9200/csv_error/_doc/1" \
    --insecure \
    -d'{  "field1": "a\nb" }' --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
# Optional, add an entry that has a quote in it.
curl -XPUT "https://${OPENSEARCH_SERVER}:9200/csv_error/_doc/2" \
    --insecure \
    -d'{ "field1": "\"a\" b" }' --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
# Optional, add an entry that has no quotes, newlines, or commas.
curl -XPUT "https://${OPENSEARCH_SERVER}:9200/csv_error/_doc/3" \
    --insecure \
    -d'{ "field1": "a b" }' --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
```
2. Query the index to CSV output or RAW output
```
#!/usr/bin/env bash

# CSV
curl -XPOST "https://${OPENSEARCH_SERVER}:9200/_plugins/_sql?format=csv" \
    --insecure \
    -d'{ "query": "SELECT * FROM csv_error LIMIT 50" }' \
    --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"

# Raw
curl -XPOST "https://${OPENSEARCH_SERVER}:9200/_plugins/_sql?format=raw" \
    --insecure \
    -d'{ "query": "SELECT * FROM csv_error LIMIT 50" }' \
    --user "${OPENSEARCH_USER}":"${OPENSEARCH_PASSWORD}"
```
Output:
```
field1
a                      <-- no start quote
b                      <-- no end quote
"""a"" b"
```

The entries with `\r` and `\n` in any combination will be quoted.

 - OS: Ubuntu 20.04
 - Version 2.19.1
 - Plugins - default

I should have caught this when I worked on https://github.com/opensearch-project/sql/issues/3050

Michael-S:
### Description

According to RFC 4180 for the CSV file format,
section 2, item 6, if a CSV cell contains a
carriage return ('\r') or line feed ('\n') it
must be quoted.

### Related Issues
Resolves #3514

### Check List
- [x ] New functionality includes testing.
- [ ] New functionality has been documented.  (N/A, bug fix)
 - [ ] New functionality has javadoc added.  (N/A, bug fix)
 - [ ] New functionality has a user manual doc added.  (N/A, bug fix)
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md). (N/A, bug fix)
- [x ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose). (N/A, bug fix)

### DIFF
--- a/protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java
+++ b/protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java
@@ -84,7 +84,10 @@ public class FlatResponseBase {
 
   protected String quoteIfRequired(String separator, String cell) {
     final String quote = "\"";
-    if (cell.contains(separator) || cell.contains(quote)) {
+    if (cell.contains(separator)
+        || cell.contains(quote)
+        || cell.contains("\r")
+        || cell.contains("\n")) {
       return quote + cell.replaceAll(quote, quote + quote) + quote;
     } else {
       return cell;
--- a/protocol/src/test/java/org/opensearch/sql/protocol/response/format/RawResponseFormatterTest.java
+++ b/protocol/src/test/java/org/opensearch/sql/protocol/response/format/RawResponseFormatterTest.java
@@ -136,6 +136,42 @@ public class RawResponseFormatterTest {
   }
 
   @Test
+  void quoteIfRequiredCrlf() {
+    ExecutionEngine.Schema schema =
+        new ExecutionEngine.Schema(
+            ImmutableList.of(
+                new ExecutionEngine.Schema.Column("field1", "field1", STRING),
+                new ExecutionEngine.Schema.Column("num", "num", INTEGER)));
+    QueryResult response =
+        new QueryResult(
+            schema,
+            Arrays.asList(
+                tupleValue(ImmutableMap.of("field1", "abc", "num", "5")),
+                tupleValue(ImmutableMap.of("field1", "def\r", "num", "\n6")),
+                tupleValue(ImmutableMap.of("field1", "gh\ri", "num", "7\n")),
+                tupleValue(ImmutableMap.of("field1", "jk\r\nl", "num", "\n\r8")),
+                tupleValue(ImmutableMap.of("field1", "\r\nmno\r\n", "num", "9\n\r0"))));
+    String expected =
+        "field1|num%n"
+            + "abc|5%n"
+            + "\"def\r\"|\"\n6\"%n"
+            + "\"gh\ri\"|\"7\n\"%n"
+            + "\"jk\r\nl\"|\"\n\r8\"%n"
+            + "\"\r\nmno\r\n\"|\"9\n\r0\"";
+    assertEquals(format(expected), getRawFormatter().format(response));
+    // Pretty formatting raw or CSV data with embedded newlines and carriage
+    // returns will still look awful on output.
+    String expectedPretty =
+        "field1   |num   %n"
+            + "abc      |5     %n"
+            + "\"def\r\"   |\"\n6\"  %n"
+            + "\"gh\ri\"   |\"7\n\"  %n"
+            + "\"jk\r\nl\"  |\"\n\r8\" %n"
+            + "\"\r\nmno\r\n\"|\"9\n\r0\"";
+    assertEquals(format(expectedPretty), getRawFormatterPretty().format(response));
+  }
+
+  @Test
   void formatError() {
     Throwable t = new RuntimeException("This is an exception");
     String expected =
--- a/protocol/src/test/java/org/opensearch/sql/protocol/response/format/CsvResponseFormatterTest.java
+++ b/protocol/src/test/java/org/opensearch/sql/protocol/response/format/CsvResponseFormatterTest.java
@@ -110,9 +110,11 @@ public class CsvResponseFormatterTest {
             schema,
             Arrays.asList(
                 tupleValue(ImmutableMap.of("na,me", "John,Smith", ",,age", "30,,,")),
+                tupleValue(ImmutableMap.of("na,me", "Line\nBreak", ",,age", "28,,,")),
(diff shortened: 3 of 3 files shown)


protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java
```diff
@@ -84,7 +84,10 @@ protected List<List<String>> formatData(List<List<String>> lines) {
 
   protected String quoteIfRequired(String separator, String cell) {
     final String quote = "\"";
-    if (cell.contains(separator) || cell.contains(quote)) {
+    if (cell.contains(separator)
+        || cell.contains(quote)
+        || cell.contains("\r")
+        || cell.contains("\n")) {
```
Swiddis, line 90: thought: We're now traversing the string 4 times instead of 2.

It might not have much of an impact in practice (benchmark?). But for formatting a large amount of data, I could see all these redundant checks adding up (imagining something like 1M rows of small cells with 0 special characters). Is there any built-in that can do this better?

Swiddis (approved):
Thanks for the contribution!

protocol/src/test/java/org/opensearch/sql/protocol/response/format/RawResponseFormatterTest.java
```diff
@@ -135,6 +135,42 @@ void quoteIfRequired() {
...
+    String expected =
+        "field1|num%n"
+            + "abc|5%n"
+            + "\"def\r\"|\"\n6\"%n"
+            + "\"gh\ri\"|\"7\n\"%n"
+            + "\"jk\r\nl\"|\"\n\r8\"%n"
+            + "\"\r\nmno\r\n\"|\"9\n\r0\"";
+    assertEquals(format(expected), getRawFormatter().format(response));
+    // Pretty formatting raw or CSV data with embedded newlines and carriage
+    // returns will still look awful on output.
```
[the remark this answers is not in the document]
Michael-S: I'm not sure what would look better for pretty-printing with newlines in fields.

protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java
Michael-S: I think to see if something matters, we would have to benchmark it.  My first thought would be to use a `java.lang.String::matches` with a regex like `"[\",\r\n]"` (but the first two entries would need to be dynamically added based on the input quote and separator characters).  My second would be to do a single traversal like:

```
for (int i = 0; i < cell.length(); i++) {
   int c = cell.codePointAt(c);
   if (c == ... || c == ... || c == ... || c == ) {
       return quote + cell.replaceAll(quote, quote + quote) + quote;
    }
    replace c;
```

For right now, if I understand OpenSearch properly the plugin is restricted to 10,000 rows.  I'm guessing on that scale, the extra string traversals are harmless.   But that's a completely wild guess, I won't argue with anyone who suggests otherwise.

Swiddis: Yeah, checks out. Unless someone is outputting [very large text fields](https://github.com/opensearch-project/opensearch-spark/issues/625) the traversals shouldn't be noticeable. I think you could theoretically nitpick and say the "c == ... || c == ..." approach is better for CPU cache locality, but I don't think there's a realistic situation where that makes a real difference. Thanks!

Swiddis:
Failing test tracked in #3516

protocol/src/test/java/org/opensearch/sql/protocol/response/format/RawResponseFormatterTest.java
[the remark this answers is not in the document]
Swiddis: IMO escaped literals like "\\n" would be better, but it's no big deal. Can let someone open a feature request down the road if they have a specific solution they like.

protocol/src/main/java/org/opensearch/sql/protocol/response/format/FlatResponseBase.java
Michael-S: > thought: We're now traversing the string 4 times instead of 2.
> 
> It might not have much of an impact in practice (benchmark?). But for formatting a large amount of data, I could see all these redundant checks adding up (imagining something like 1M rows of small cells with 0 special characters). Is there any built-in that can do this better?

You piqued my interest, so I wrote a simplistic benchmark.  https://github.com/Michael-S/silly_java_string_search
Ten million random unicode strings, length 20-520, with the quotable characters sprinkled throughout.  On my mid-range Intel laptop, in a single-threaded search: using the four `contains` calls took about 2.5 seconds; a manual loop to look at each character with four equality checks took about 3.8 seconds; and a regex took about 6.1 second.  (And the regex didn't find the same number of matches as the others, which confused me because I thought the regex was simple.)

So it looks like `contains` is a decent default.  Maybe some JVM wizard has a better solution.

seankao-az (approved):

Issue #3516, opened 2025-04-04
[BUG] Flaky test: `CalcitePPLDateTimeBuiltinFunctionIT > testDateFormatAndDatetimeAndFromDays`


Swiddis:
**What is the bug?**

`CalcitePPLDateTimeBuiltinFunctionIT > testDateFormatAndDatetimeAndFromDays` has failed multiple times for unrelated code changes:
- In #3513 -- https://github.com/opensearch-project/sql/actions/runs/14240445113/job/39908933119#step:4:1091
- In #3515 -- https://github.com/opensearch-project/sql/actions/runs/14250153623/job/39940768892?pr=3515#step:6:1094

Merged 2 days ago at the time of writing: 920f711

**How can one reproduce the bug?**
By accident.

**What is the expected behavior?**
The test should behave consistently.

**What is your host/environment?**
GitHub actions. Happens for at least Linux and Mac with both Java 21 and 23. Haven't seen it for Windows yet.

**Do you have any screenshots?**
N/A

**Do you have any additional context?**
N/A

Swiddis:
Digging through [the failing test](https://github.com/opensearch-project/sql/blob/920f7115f3322d78bcff58491406e83af809227d/integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLDateTimeBuiltinFunctionIT.java#L950-L1001) and tracing several dependencies:

It looks like `DATETIME('1984-04-12T09:07:42.000Z', '+08:00')` and `DATETIME('1984-04-12 09:07:42', '+00:00')` inconsistently return null with Calcite (which causes two more nulls later). I can't pinpoint why at a glance, these should ideally be pure functions. Will leave it to the assignee.

xinyual:
Already raise the pr for it. https://github.com/opensearch-project/sql/pull/3517

PR #3517, opened 2025-04-05
use correct timezone name

xinyual:
### Description
Previously, we use getID() and it will return some three-letter time zone IDs like "PLT" which is already deprecated (see [here](https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html)). But toZoneId().toString() will return the full name like 'Asia/Karachi'.

### Related Issues
#3516 

### Check List


Swiddis (approved):
Nice! Is the reason for the breakage that, depending on the actual timezone string, it may or may not be convertible back into a zone? So then it was failing tests based on where the GHA runner was

LantaoJin (approved):

Issue #3493, opened 2025-03-31
[RFC][Enhancement][Calcite Engine] UDF utils refactoring


qianheng-aws:
**Is your feature request related to a problem?**
The current implementation lacks readability as we split a function's implementation into several parts (operator, arguments, return type), especially when we also have post-processing on the operator's return and pre-processing on the arguments. We may also need to implement argument checking in the future. Therefore, it would be better to organize the implementation of a function cohesively.

Another disadvantage is that we always return a RexCall when visitFunction is called, but some functions(like DATE, TYPEOF) can be evaluated during the building phase, allowing us to return the final results directly. In these cases, returning a RexLiteral directly would be more efficient than returning a RexCall.

PR #3512, opened 2025-04-03
Validate arguments for datetime, string, and math UDFs

yuancu:
### Description

Created a custom validator to validate the arguments of UDFs

TODOs:
- [ ] Migrate the parameter validation based on #3522 

#### String

- REPLACE
- ASCII
- LENGTH
- LOWER
- LTRIM
- REVERSE
- RTRIM
- UPPER
- TRIM
- CONCAT_WS
- SUBSTRING
- SUBSTR
- LOCATE
- POSITION
- STRCMP
- LEFT
- RIGHT
- CONCAT

#### Math

- ABS
- ACOS
- ASIN
- COS
- COT
- DEGREES
- EXP
- FLOOR
- LN
- LOG2
- LOG10
- RADIANS
- SIGN
- SIN
- SQRT
- CBRT
- ATAN
- LOG
- ATAN2
- MOD
- POW
- POWER
- CEIL
- CEILING
- CONV
- CRC32
- E
- PI
- RAND
- ROUND

#### Date / time

- ADDDATE
- SUBDATE
- DATE_ADD
- DATE_SUB
- ADDTIME
- SUBTIME
- DATEDIFF
- CONVERT_TZ
- CURDATE
- CURRENT_DATE
- CURRENT_TIME
- LOCALTIMESTAMP
- LOCALTIME
- CURRENT_TIMESTAMP
- NOW
- CURTIME
- UTC_DATE
- UTC_TIME
- UTC_TIMESTAMP
- DAY
- DAY_OF_WEEK
- DAYOFWEEK
- DAY_OF_YEAR
- DAYOFYEAR
- DAYNAME
- DAYOFMONTH
- DAY_OF_MONTH
- MONTHNAME
- QUARTER
- TO_DAYS
- YEAR
- DATE
- HOUR
- HOUR_OF_DAY
- LAST_DAY
- MICROSECOND
- MINUTE
- MINUTE_OF_DAY
- MINUTE_OF_HOUR
- MONTH
- MONTH_OF_YEAR
- SECOND
- SECOND_OF_MINUTE
- WEEKDAY
- TIME
- TIME_TO_SEC
- MAKEDATE
- MAKETIME
- EXTRACT
- FROM_DAYS
- SEC_TO_TIME
- FROM_UNIXTIME
- GET_FORMAT
- STR_TO_DATE
- DATETIME
- SYSDATE
- DATE_FORMAT
- TIME_FORMAT
- PERIOD_ADD
- PERIOD_DIFF
- TIMESTAMP
[...]

PR #3522, opened 2025-04-07
[Calcite Engine] Function framework refactoring

qianheng-aws:
### Description
Do refactoring for the framework of function.

It includes change or enhancement:
1. Add `PPLFuncImpTable` for registering logical-level implementation for resolving a BuiltinFunctionName. It also includes resolving for functions with different parameters by using CalciteFuncSignature.
2. Add `PPLBuiltinOperators` for all self-implemented UDFs. This class is a physical-level implementation of an operator.
3. Migrate all `SqlStdOperatorTable` and `SqlLibraryOperators` to the new framework, and also migrate the span function and typeof function.
4. Make the UDF static instead of creating a new one when calling visitFunction.
5. Replace function name(STRING) with BuiltinFunctionName(ENUM) to strengthen code robustness

The new framework aims to hide all implementation at the physical level, so the plan will be more concise and clear.

Take `span` for example: for a time-related field, span will use a ScalarFunctionImpl (physical-level implementation, can only see `span(\$1, \$2, \$3)` in the plan); while using logical implementation like `FLOOR($1 / $2) * $2` for a non-time field. This makes the plan weird and confusing, and it also blocks push-down since the push-down rule takes effect by identifying the `span` function operator. After migrating, the `SPAN` operator is able to handle both cases internally.

`PPLFuncImpTable` will provide more flexible implementation of a function than only using SqlOperator. 

Take `typeof` for example, after migrating, the new implementation will replace using ScalarFunction `TYPEOF(\$1)` with returning a `RexLiteral` directly, which avoids row-by-row evaluation.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3493

### Check List


core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -0,0 +1,345 @@
...
+
+  public interface FunctionImp1 extends FunctionImp {
+    List<RelDataType> ANY_TYPE_1 = List.of(ANY_TYPE);
+
+    RexNode apply(RexBuilder builder, RexNode arg1);
+
+    @Override
+    default RexNode apply(RexBuilder builder, RexNode... args) {
+      if (args.length != 1) {
+        throw new IllegalArgumentException("This function requires exactly 1 arguments");
```
penghuo, line 114: use function name in exception instead of "This"

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -0,0 +1,345 @@
...
+        mapBuilder = ImmutableMap.builder();
+    builder.map.forEach((k, v) -> mapBuilder.put(k, v.immutable()));
+    this.map = ImmutableMap.copyOf(mapBuilder.build());
+  }
+
+  public @Nullable FunctionImp get(final BuiltinFunctionName functionName) {
+    final PairList<CalciteFuncSignature, FunctionImp> implementList =
+        requireNonNull(map.get(functionName));
+    if (implementList.isEmpty()) {
+      throw new NullPointerException();
```
penghuo, line 166: response is nullable, but throw NullPointerException?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -0,0 +1,345 @@
...
+    }
+    return implementList.getFirst().getValue();
```
penghuo, line 168: always return first? if it is by design, add an UT.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -0,0 +1,345 @@
...
+    final PairList<CalciteFuncSignature, FunctionImp> implementList = map.get(functionName);
+    if (implementList == null || implementList.isEmpty()) {
+      throw new IllegalStateException(String.format("Cannot resolve function: %s", functionName));
+    }
+    List<RelDataType> argTypes = args.stream().map(RexNode::getType).toList();
+    for (Map.Entry<CalciteFuncSignature, FunctionImp> implement : implementList) {
+      if (implement.getKey().match(functionName.getName(), argTypes)) {
+        return implement.getValue();
+      }
+    }
```
penghuo, line 198: In worst case, what is the maximum number of signatures that would need to be checked?

core/src/main/java/org/opensearch/sql/expression/function/UserDefinedFunctionHelper.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.calcite.schema.ImplementableFunction;
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.InferTypes;
+import org.apache.calcite.sql.type.SqlOperandMetadata;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.validate.SqlUserDefinedFunction;
+
+public interface UserDefinedFunctionHelper {
```
penghuo, line 18: Add doc

core/src/main/java/org/opensearch/sql/expression/function/UserDefinedFunctionHelper.java
```diff
@@ -0,0 +1,40 @@
...
+
+  ImplementableFunction getFunction();
+
+  SqlReturnTypeInference getReturnTypeInference();
+
+  default SqlOperandMetadata getOperandMetadata() {
+    return null;
+  }
+  ;
```
penghuo, line 27: useless?

core/src/main/java/org/opensearch/sql/expression/function/SpanFunctionImpl.java
```diff
@@ -0,0 +1,116 @@
...
+        };
+      } else if (fieldType instanceof ExprSqlType exprSqlType) {
+        // TODO: pass in constant arguments when constructing
+        String methodName =
+            switch (exprSqlType.getUdt()) {
+              case EXPR_DATE -> "evalDate";
+              case EXPR_TIME -> "evalTime";
+              case EXPR_TIMESTAMP -> "evalTimestamp";
+              default -> throw new IllegalArgumentException(
+                  String.format("Unsupported expr udt: %s", exprSqlType.getUdt()));
```
penghuo, line 71: UDT is internal implementation, we should hide internal error message from user.

core/src/main/java/org/opensearch/sql/expression/function/SpanFunctionImpl.java
```diff
@@ -0,0 +1,116 @@
...
+            };
+        ScalarFunctionImpl function =
+            (ScalarFunctionImpl)
+                ScalarFunctionImpl.create(
+                    Types.lookupMethod(
+                        SpanFunctionImpl.class, methodName, String.class, int.class, String.class));
+        return function.getImplementor().implement(translator, call, RexImpTable.NullAs.NULL);
+      }
+      throw new IllegalArgumentException(String.format("Unsupported RelDataType: %s", fieldType));
```
penghuo, line 80: RelDataType is internal implementation, we should hide internal error message from user.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
qianheng-aws: Removed, it's replaced with resolve and never used for now.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
qianheng-aws: ditto

core/src/main/java/org/opensearch/sql/expression/function/CalciteFuncSignature.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.opensearch.sql.expression.function;
+
+import static org.opensearch.sql.expression.function.PPLFuncImpTable.FunctionImp.ANY_TYPE;
+
+import java.util.List;
+import org.apache.calcite.rel.type.RelDataType;
+
+/** Function signature is composed by function name and arguments list. */
+public record CalciteFuncSignature(FunctionName functionName, List<RelDataType> funcArgTypes) {
```
xinyual, line 14: Do we really need need this signature? If we follow the SpanFunctionImpl, we could put all real implement function like `evalDate` into same class. Also, the argument validation check would be in another [PR](https://github.com/opensearch-project/sql/pull/3512). I think the only reason here is about implicit conversion. But I think it's already disabled in calcite now?

core/src/main/java/org/opensearch/sql/expression/function/ScalarUDF.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package org.opensearch.sql.expression.function;
+
+import org.apache.calcite.linq4j.tree.Types;
+import org.apache.calcite.schema.ImplementableFunction;
+import org.apache.calcite.schema.impl.ScalarFunctionImpl;
+
+/** The UDF which implements its functionality by invoking the `eval` method of a class */
+public abstract class ScalarUDF implements UserDefinedFunctionHelper {
```
xinyual, line 13: Almost all UDF functions has different implementations for different arguments. So I think this class is useless?

core/src/main/java/org/opensearch/sql/expression/function/CalciteFuncSignature.java
qianheng-aws: I'm wondering if there is chance that:

1. an function may has different operator depends on params' type or size and wants this information exposed on our plan. Although, by most cases,  we expect the one-to-one relationship between function and operator and hide different processing logic in one operator.

2. different functions map to one operator while each of them has their specific limitation on the params' size or type. E.g. We use operator `NOT_EQUAL` for both function `!=` and `XOR`, while `XOR` has its limitation that only accepting boolean, so we could put this limitation check on its signature. Of course, we can implement `XOR` by an UDF with more efforts. It depends on whether we want to expose this implementation details to customer in the plan or not.

It stills remains discussion on the above 2 cases. We can keep this redundant level here since it doesn't harm the functionality or performance, until the final decision in the end.

core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
```diff
@@ -0,0 +1,42 @@
...
+public class PPLBuiltinOperators extends ReflectiveSqlOperatorTable {
+
+  public static final SqlOperator SPAN = new SpanFunctionImpl().toUDF("SPAN");
+
+  /**
+   * Invoking an implementor registered in {@link RexImpTable}, need to use reflection since they're
+   * all private Use method directly in {@link BuiltInMethod} if possible, most operators'
+   * implementor could be substituted by a single method.
+   */
+  private static Expression invokeCalciteImplementor(
```
LantaoJin, line 30: How does this method invoke with `private static`

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -195,7 +197,9 @@ public RexNode visitCompare(Compare node, CalcitePlanContext context) {
         transferCompareForDateRelated(leftCandidate, context, whetherCompareByTime);
     final RexNode right =
         transferCompareForDateRelated(rightCandidate, context, whetherCompareByTime);
-    return context.relBuilder.call(op, left, right);
+    return PPLFuncImpTable.INSTANCE
+        .resolve(node.getOperator(), List.of(left, right))
+        .apply(context.rexBuilder, left, right);
```
LantaoJin, line 202: Method name confusion: how about change to following? 
```
.apply(node.getOperator(), List.of(left, right))
.resolve(context.rexBuilder, left, right);
```

`apply()` in `FunctionImpX` normally should return an instance of `FunctionImpX`

core/src/main/java/org/opensearch/sql/expression/function/ScalarUDF.java
qianheng-aws: Yeah, useless for now. I'm not sure if it could be useful in the future.

This is for the case where one reflective method could cover the functionality of a operator, like the BuiltInMethod in calcite. For more complex implementation like handling different arguments needs different logic, should use `ImplementorUDF`

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -0,0 +1,345 @@
...
+  public @Nullable FunctionImp get(final BuiltinFunctionName functionName) {
+    final PairList<CalciteFuncSignature, FunctionImp> implementList =
+        requireNonNull(map.get(functionName));
+    if (implementList.isEmpty()) {
+      throw new NullPointerException();
+    }
+    return implementList.getFirst().getValue();
+  }
+
+  public @Nullable FunctionImp resolveSafe(final String functionName, List<RexNode> args) {
```
LantaoJin, line 171: `applyNullable`

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -0,0 +1,345 @@
...
+    try {
+      return resolve(functionName, args);
+    } catch (IllegalStateException e) {
```
LantaoJin, line 174: Exception e

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -0,0 +1,345 @@
...
+      throw new IllegalArgumentException(String.format("Unsupported function: %s", functionName));
+    }
+    BuiltinFunctionName.of(functionName);
+    return resolve(funcNameOpt.get(), args);
+  }
+
+  public FunctionImp resolve(final BuiltinFunctionName functionName, List<RexNode> args) {
+    final PairList<CalciteFuncSignature, FunctionImp> implementList = map.get(functionName);
+    if (implementList == null || implementList.isEmpty()) {
+      throw new IllegalStateException(String.format("Cannot resolve function: %s", functionName));
```
LantaoJin, line 191: why `IllegalStateException`? do you mean `IllegalArgumentException`? Or `SemanticCheckException`

core/src/main/java/org/opensearch/sql/expression/function/SpanFunctionImpl.java
```diff
@@ -0,0 +1,116 @@
...
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return ReturnTypes.ARG0;
+  }
+
+  public static class SpanImplementor implements NotNullImplementor {
+    @Override
+    public Expression implement(
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+      assert call.getOperands().size() == 3;
+      assert translatedOperands.size() == 3;
```
LantaoJin, line 48: can you add a message for these assert?

core/src/main/java/org/opensearch/sql/expression/function/SpanFunctionImpl.java
```diff
@@ -0,0 +1,116 @@
...
+                ScalarFunctionImpl.create(
+                    Types.lookupMethod(
+                        SpanFunctionImpl.class, methodName, String.class, int.class, String.class));
+        return function.getImplementor().implement(translator, call, RexImpTable.NullAs.NULL);
+      }
+      throw new IllegalArgumentException(String.format("Unsupported RelDataType: %s", fieldType));
+    }
+  }
+
+  @Strict
```
LantaoJin, line 84: This annotation applied to a user-defined function that indicates that the function returns null if and only if one or more of its arguments are null. please confirm

core/src/main/java/org/opensearch/sql/expression/function/UserDefinedFunctionHelper.java
LantaoJin: Why a helper an interface? Can be `UserDefinedFunctionFactory` or `UserDefinedFunctionTemplate?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
qianheng-aws: It depends on how many FunctionImp one function may have. But, in fact, it's always only 1 for now. 

And we expect different implementation of one function should all be included in the implementor of its related operator. i.e. implementation details should be hidden in the physical-level instead of logical-level, unless we implicitly want to expose the logical detail of the implementation.(like if want integral division for integer, we may want cast invoking is exposed in the plan instead of being hidden in the implementor). 

In the visible future, the maximum number should be only 1 or very small. And we can deprecate this extra level if we actually doesn't have such requirement for it.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
qianheng-aws: Addressed by re-catching this exception and add information of function name in the error message.

core/src/main/java/org/opensearch/sql/expression/function/SpanFunctionImpl.java
qianheng-aws: Yes, I think that satisfy our expectation.

penghuo (approved):

Issue #3397, opened 2025-03-07
[FEATURE] Calcite Engine: Support Alias type


LantaoJin:
**Is your feature request related to a problem?**
Support Alias type with Calcite

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

PR #3538, opened 2025-04-11
[Calcite Engine] Support alias type field

qianheng-aws:
### Description
Support alias type. 

Since Calcite's logical fields is strictly binding to an offset of its physical array, we have to use an approach different from v2(https://github.com/opensearch-project/sql/pull/3246) but more like the implementation in opensearch-spark(https://github.com/opensearch-project/opensearch-spark/pull/1032).

### Related Issues
Resolves  https://github.com/opensearch-project/sql/issues/3397

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
@@ -11,7 +11,9 @@ import static org.opensearch.sql.legacy.SQLIntegTestCase.Index.DATA_TYPE_NUMERIC
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_ALIAS;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_DATATYPE_NONNUMERIC;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_DATATYPE_NUMERIC;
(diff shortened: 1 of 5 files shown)


integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
```diff
@@ -85,8 +87,9 @@ public void test_alias_data_type() throws IOException {
     JSONObject result =
         executeQuery(
             String.format(
-                "source=%s | where alias_col > 1 " + "| fields original_col, alias_col ",
+                "source=%s | eval a = 1 | where alias_col > 1 | fields original_col, alias_col ",
```
penghuo, line 90: why need a=1?

penghuo (dismissed):

integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
qianheng-aws: > why need a=1?

Avoid the filter push down to ensure the PR really works.

LantaoJin: Please add a test in our standalone IT which could ensure the pushdown w/wo always works. The `eval a=1` in future will trigger push down as well.

PR #3542, opened 2025-04-14
fix timestamp bug

xinyual:
### Description
This pr fix the timestamp bug when take time argument. It uses the current function properties. We now follow the same logic as https://github.com/opensearch-project/sql/blob/e23a61dcd9f15864a6bf4a5daa7ac4263a890716/core/src/main/java/org/opensearch/sql/expression/datetime/DateTimeFunctions.java#L880 to firstly transfer it to timestamp.

### Related Issues
https://github.com/opensearch-project/sql/issues/3545

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/udf/datetimeUDF/TimestampFunction.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/udf/datetimeUDF/TimestampFunction.java
@@ -7,7 +7,6 @@ package org.opensearch.sql.calcite.udf.datetimeUDF;
 
 import static org.opensearch.sql.calcite.utils.UserDefinedFunctionUtils.restoreFunctionProperties;
 import static org.opensearch.sql.calcite.utils.datetime.DateTimeApplyUtils.transferInputToExprTimestampValue;
-import static org.opensearch.sql.calcite.utils.datetime.DateTimeApplyUtils.transferInputToExprValue;
 import static org.opensearch.sql.expression.datetime.DateTimeFunctions.exprAddTime;
 
 import java.util.Objects;
@@ -20,7 +19,8 @@ import org.opensearch.sql.expression.function.FunctionProperties;
 /**
  * We need to write our own since we are actually implement timestamp add here
  * (STRING/DATE/TIME/DATETIME/TIMESTAMP) -> TIMESTAMP (STRING/DATE/TIME/DATETIME/TIMESTAMP,
- * STRING/DATE/TIME/DATETIME/TIMESTAMP) -> TIMESTAMP
+ * STRING/DATE/TIME/DATETIME/TIMESTAMP) -> TIMESTAMP It's implicitly transferred into timestamp, so
+ * we need to do the same thing
  */
 public class TimestampFunction implements UserDefinedFunction {
   @Override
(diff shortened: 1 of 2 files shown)


integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLDateTimeBuiltinFunctionIT.java
```diff
@@ -101,6 +101,20 @@ public void testDate() {
             Date.valueOf("1984-04-12")));
   }
 
+  @Test
+  public void testTimestampWithTimeInput() {
+    String utcTomorrow = LocalDate.now().plusDays(1).toString();
```
qianheng-aws, line 106: There may be an edge case this test should be flakey when it was ran at the time point crossing a day.

I think it should use a fixed time via Hook.CURRENT_TIMESTAMP once changing to use the time properties in Calcite.

qianheng-aws (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLDateTimeBuiltinFunctionIT.java
xinyual: Currently we don't use the latest refactor framework so Hook wouldn't work. But after changing it, I will use Hook in this IT.

qianheng-aws:
CI failed for java21, is this a known failure?

qianheng-aws:
> CI failed for java21, is this a known failure?

Should be addressed by this fix: https://github.com/opensearch-project/sql/pull/3543

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLDateTimeBuiltinFunctionIT.java
qianheng-aws: Please add a TODO here.

LantaoJin:
Could you open an issue?

core/src/main/java/org/opensearch/sql/calcite/udf/datetimeUDF/TimestampFunction.java
```diff
@@ -20,7 +19,8 @@
 /**
  * We need to write our own since we are actually implement timestamp add here
  * (STRING/DATE/TIME/DATETIME/TIMESTAMP) -> TIMESTAMP (STRING/DATE/TIME/DATETIME/TIMESTAMP,
- * STRING/DATE/TIME/DATETIME/TIMESTAMP) -> TIMESTAMP
+ * STRING/DATE/TIME/DATETIME/TIMESTAMP) -> TIMESTAMP It's implicitly transferred into timestamp, so
```
LantaoJin, line 22: please format this java doc for more readable

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLDateTimeBuiltinFunctionIT.java
penghuo: > `timestamp(time('12:00:00'), time('12:00:00'))`

what is meaning of time+time? should we limit the first argument be date only?

core/src/main/java/org/opensearch/sql/calcite/udf/datetimeUDF/TimestampFunction.java
xinyual: Done.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLDateTimeBuiltinFunctionIT.java
xinyual: In v2, we will implicitly convert the time into timestamp. The logic of v2 timestamp is transfer two arguments into timestamp (if we have two.), then extract time part from the second timestamp and then add it to the first timestamp.

xinyual:
> Could you open an issue?

Done. https://github.com/opensearch-project/sql/issues/3545

LantaoJin (approved):

qianheng-aws (approved):

PR #3544, opened 2025-04-14
[Calcite Engine] Only enable fallback for tests that need to fall back

qianheng-aws:
### Description
This PR aims to refactoring the current Calcite remote IT. It has drawbacks:
1. Although enabling fallback by default, Calcite\*IT will run most tests without fallback in fact, it has duplication process with NonFallbackCalcite\*IT which extends from it. This will increase our IT time cost a lot with unnecessary tests.
2. For some tests we originally expect them throw exception(because of incorrect PPL), we don't want to test with fallback to v2 because v2 will throw exception as well. E.g. WhereCommandIT::testInWithIncompatibleType. Testing the error message in separate suite(calcite enable/disable) is enough.

This PR unify Calcite\*IT with NonFallbackCalcite\*IT together, and override tests with fallback for those we've previous ignored in NonFallbackCalcite\*IT

And this PR will also fix the cases where some test/suite have been ignored incorrectly.

Resolves https://github.com/opensearch-project/sql/issues/3546

### Check List


integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
```diff
@@ -143,4 +143,21 @@ public static void disallowCalciteFallback() throws IOException {
             "persistent", Settings.Key.CALCITE_FALLBACK_ALLOWED.getKeyValue(), "false"));
     System.out.println(Settings.Key.CALCITE_FALLBACK_ALLOWED.name() + " disabled");
   }
+
+  public static void withFallbackEnabled(Runnable f, String msg) throws IOException {
+    System.out.printf("Need fallback to v2 due to %s%n", msg);
```
dai-chen, line 148: Is sysout the preferred way to log in test?

penghuo (approved):

integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
qianheng-aws: LOG should be better, will change to use LOG.info.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/fallback/CalciteWhereCommandIT.java
```diff
@@ -1,16 +0,0 @@
-/*
- * Copyright OpenSearch Contributors
- * SPDX-License-Identifier: Apache-2.0
- */
-
-package org.opensearch.sql.calcite.remote.fallback;
-
-import org.opensearch.sql.ppl.WhereCommandIT;
-
-public class CalciteWhereCommandIT extends WhereCommandIT {
```
LantaoJin, line 10: > This PR unify Calcite*IT with NonFallbackCalcite*IT together, and override tests with fallback for those we've previous ignored in NonFallbackCalcite*IT

Why does this PR only delete one IT file?

qianheng-aws: Will migrate others if this approach is acceptable

LantaoJin (approved):

penghuo (approved):

Issue #3548, opened 2025-04-15
[FEATURE] Calcite Engine: Support BINARY type


qianheng-aws:
**Is your feature request related to a problem?**
V2 PPL uses String as BINARY while Calcite use ByteString

**What solution would you like?**
Use UDT

PR #3549, opened 2025-04-15
Support UDT for BINARY

qianheng-aws:
### Description
We previous transform OpenSearch binary type to Calcite's SqlTypeName.BINARY. However, OpenSearch stores binary value as string while Calcite stores it as ByteString, the gap will lead to execution error if transforming directly.

This PR fixes it by viewing our OpenSearch binary type as a new UDT, which maps to SqlTypeName.VARCHAR

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3548

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/data/type/ExprCoreType.java
+++ b/core/src/main/java/org/opensearch/sql/data/type/ExprCoreType.java
@@ -51,6 +51,8 @@ public enum ExprCoreType implements ExprType {
   /** Geometry. Only support point now. */
   GEO_POINT(UNDEFINED),
 
+  BINARY(UNDEFINED),
+
   /** Struct. */
   STRUCT(UNDEFINED),
 
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
@@ -17,6 +17,7 @@ import static org.opensearch.sql.util.MatcherUtils.verifyDataRows;
 import static org.opensearch.sql.util.MatcherUtils.verifySchema;
 
 import java.io.IOException;
+import org.json.JSONArray;
 import org.json.JSONObject;
 import org.junit.Test;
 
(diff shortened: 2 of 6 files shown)


integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
```diff
@@ -60,6 +61,21 @@ public void test_nonnumeric_data_types() throws IOException {
         schema("object_value", "struct"),
         schema("nested_value", "array"),
         schema("geo_point_value", "geo_point"));
+    verifyDataRows(
```
LantaoJin, line 64: Could you change `verifySchema` to `verifySchemaInOrder` and `verifyDataRows` to `verifyDataRowsInOrder` for this test?

core/src/main/java/org/opensearch/sql/data/type/ExprCoreType.java
```diff
@@ -51,6 +51,8 @@ public enum ExprCoreType implements ExprType {
   /** Geometry. Only support point now. */
   GEO_POINT(UNDEFINED),
 
+  BINARY(UNDEFINED),
```
penghuo, line 54: Does Calcite impl required ExprCoreType mapping? Should we consider deprecated ExprCoreType in future?

integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
```diff
@@ -60,6 +61,21 @@ public void test_nonnumeric_data_types() throws IOException {
...
+            "2020-10-13 13:00:00",
+            true,
+            "127.0.0.1",
+            new JSONArray(
+                "[{\"last\": \"Smith\", \"first\": \"John\"}, {\"last\": \"White\", \"first\":"
+                    + " \"Alice\"}]"),
+            new JSONObject("{\"last\": \"Dale\", \"first\": \"Dale\"}"),
+            "keyword",
+            new JSONObject("{\"lon\": 74, \"lat\": 40.71}"),
+            "U29tZSBiaW5hcnkgYmxvYg=="));
```
penghuo, line 78: what is current value without this PR?

qianheng-aws: This is the value V2 returned.

For Calcite, it will throw exception. But this test was ignored previously so CI can pass.

core/src/main/java/org/opensearch/sql/data/type/ExprCoreType.java
qianheng-aws: For the current implementation it requires, because we will always populate the final results with ExprValue transformed from Calcite's results.

https://github.com/opensearch-project/sql/blob/3c1591acc99d637b88a749be7090b47d96adca09/opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java#L215

integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
qianheng-aws: The fallback reason in Calcite's IT is outdated actually, we already have supported IP type. It's true reason is the incompatible BIANRY type.

integ-test/src/test/java/org/opensearch/sql/ppl/DataTypeIT.java
qianheng-aws: Done

LantaoJin (approved):

penghuo (approved):

PR #3550, opened 2025-04-15
Speed up aggregation pushdown for single group-by expression

### LINKED ISSUE #3528: [BUG]Span query in PPL is slower than date histogram aggregation in query DSL
**What is the bug?**

In Discover page of OSD, when executing a PPL query without any filter against an index pattern containing very huge data(about 750B documents), the query latency of PPL is much more than the latency of the similar query in DQL, which is one hundred seconds vs. seconds. 

The PPL is: `source = index* | where `@timestamp` >= '2025-03-25 03:31:32' and `@timestamp` <= '2025-04-09 03:31:32' | stats count() by span(@timestamp, 12h) `,

and the similar query DSL is:
```
{
  "query": {
    "bool": {
      "must": [],
      "filter": [
       
        {
          "range": {
            "@timestamp": {
              "gte": "2025-03-25T03:31:32.935Z",
              "lte": "2025-04-09T03:31:32.935Z",
            }
          }
        }
      ]
    }
  },
  "size":500,
  "aggs": {
    "2": {
      "date_histogram": {
        "field": "@timestamp",
        "fixed_interval": "12h",
        "time_zone": "+00:00",
        "min_doc_count": 1
      }
    }
  }
}
```
.

I see PPL will convert the span query to a composite aggregation, like this:
```
"aggregations": {
    "composite_buckets": {
      "composite": {
        "size": 1000,
        "sources": [
          {
            "span(@timestamp,12h)": {
              "date_histogram": {
                "field": "@timestamp",
                "missing_bucket": true,
                "missing_order": "first",
                "order": "asc",
                "fixed_interval": "12h"
              }
            }
          }
        ]
      },
      "aggregations": {
        "count()": {
          "value_count": {
            "field": "_index"
          }
        }
      }
    }
  }
```
, seems composite aggregation is slower than the date histogram aggregation.

1. Find a big dataset which contains billions of documents
2. Execute both the PPL and query DSL above to compare the latency

PPL should improve the performance.

OpenSearch 3.0.0

LantaoJin:
### Description
Currently, the aggregation pushdown implementation build a composite aggregation builder which is [expensive](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-composite-aggregation.html).

This PR is refactoring the implementation of **single group-by** aggregation pushdown which could get 100x faster than current implementation. See the usage case in #3528 

### Related Issues
Resolves #3528 

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
@@ -49,12 +49,14 @@ import org.opensearch.search.aggregations.AggregatorFactories.Builder;
 import org.opensearch.search.aggregations.bucket.composite.CompositeValuesSourceBuilder;
 import org.opensearch.search.aggregations.bucket.composite.TermsValuesSourceBuilder;
 import org.opensearch.search.aggregations.bucket.missing.MissingOrder;
+import org.opensearch.search.aggregations.bucket.terms.TermsAggregationBuilder;
 import org.opensearch.search.aggregations.metrics.ExtendedStats;
 import org.opensearch.search.aggregations.support.ValueType;
 import org.opensearch.search.aggregations.support.ValuesSourceAggregationBuilder;
 import org.opensearch.search.sort.SortOrder;
 import org.opensearch.sql.opensearch.data.type.OpenSearchDataType;
 import org.opensearch.sql.opensearch.request.PredicateAnalyzer.NamedFieldExpression;
+import org.opensearch.sql.opensearch.response.agg.BucketAggregationParser;
 import org.opensearch.sql.opensearch.response.agg.CompositeAggregationParser;
 import org.opensearch.sql.opensearch.response.agg.MetricParser;
 import org.opensearch.sql.opensearch.response.agg.NoBucketAggregationParser;
@@ -126,6 +128,12 @@ public class AggregateAnalyzer {
         return Pair.of(
             ImmutableList.copyOf(metricBuilder.getAggregatorFactories()),
             new NoBucketAggregationParser(metricParserList));
+      } else if (aggregate.getGroupSet().length() == 1) {
+        ValuesSourceAggregationBuilder<?> bucketBuilder =
+            createBucketAggregation(groupList.getFirst(), fieldExpressionCreator);
+        return Pair.of(
+            Collections.singletonList(bucketBuilder.subAggregations(metricBuilder)),
+            new BucketAggregationParser(metricParserList));
       } else {
         List<CompositeValuesSourceBuilder<?>> buckets =
             createCompositeBuckets(groupList, fieldExpressionCreator);
@@ -262,6 +270,28 @@ public class AggregateAnalyzer {
     if (List.of(TIMESTAMP, TIME, DATE)
         .contains(groupExpr.getOpenSearchDataType().getExprCoreType())) {
       sourceBuilder.userValuetypeHint(ValueType.LONG);
+    }
+
+    return sourceBuilder;
+  }
+
+  private static ValuesSourceAggregationBuilder<?> createBucketAggregation(
+      Integer group, FieldExpressionCreator fieldExpressionCreator) {
+    NamedFieldExpression groupExpr = fieldExpressionCreator.create(group);
+    // TODO: support histogram bucket(i.e. PPL span expression)
+    // https://github.com/opensearch-project/sql/issues/3384
+    return createTermsAggregationBuilder(groupExpr);
+  }
+
+  private static TermsAggregationBuilder createTermsAggregationBuilder(
+      NamedFieldExpression groupExpr) {
+
+    TermsAggregationBuilder sourceBuilder =
+        new TermsAggregationBuilder(groupExpr.getRootName()).size(AGGREGATION_BUCKET_SIZE);
+    // Time types values are converted to LONG in ExpressionAggregationScript::execute
+    if (List.of(TIMESTAMP, TIME, DATE)
+        .contains(groupExpr.getOpenSearchDataType().getExprCoreType())) {
+      sourceBuilder.userValueTypeHint(ValueType.LONG);
     }
 
     return sourceBuilder;
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/aggregation/AggregationQueryBuilder.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/aggregation/AggregationQueryBuilder.java
@@ -29,11 +29,13 @@ import org.opensearch.sql.expression.NamedExpression;
 import org.opensearch.sql.expression.ReferenceExpression;
 import org.opensearch.sql.expression.aggregation.NamedAggregator;
 import org.opensearch.sql.opensearch.data.type.OpenSearchDataType;
+import org.opensearch.sql.opensearch.response.agg.BucketAggregationParser;
 import org.opensearch.sql.opensearch.response.agg.CompositeAggregationParser;
 import org.opensearch.sql.opensearch.response.agg.MetricParser;
 import org.opensearch.sql.opensearch.response.agg.NoBucketAggregationParser;
 import org.opensearch.sql.opensearch.response.agg.OpenSearchAggregationResponseParser;
 import org.opensearch.sql.opensearch.storage.script.aggregation.dsl.BucketAggregationBuilder;
+import org.opensearch.sql.opensearch.storage.script.aggregation.dsl.CompositeAggregationBuilder;
 import org.opensearch.sql.opensearch.storage.script.aggregation.dsl.MetricAggregationBuilder;
 import org.opensearch.sql.opensearch.storage.serialization.ExpressionSerializer;
 
@@ -50,12 +52,16 @@ public class AggregationQueryBuilder extends ExpressionNodeVisitor<AggregationBu
   /** Bucket Aggregation builder. */
   private final BucketAggregationBuilder bucketBuilder;
 
+  /** Composite Aggregation builder for multiple buckets. */
+  private final CompositeAggregationBuilder compositeBuilder;
+
   /** Metric Aggregation builder. */
   private final MetricAggregationBuilder metricBuilder;
 
   /** Aggregation Query Builder Constructor. */
   public AggregationQueryBuilder(ExpressionSerializer serializer) {
     this.bucketBuilder = new BucketAggregationBuilder(serializer);
+    this.compositeBuilder = new CompositeAggregationBuilder(serializer);
     this.metricBuilder = new MetricAggregationBuilder(serializer);
   }
 
@@ -74,13 +80,20 @@ public class AggregationQueryBuilder extends ExpressionNodeVisitor<AggregationBu
       return Pair.of(
           ImmutableList.copyOf(metrics.getLeft().getAggregatorFactories()),
           new NoBucketAggregationParser(metrics.getRight()));
+    } else if (groupByList.size() == 1) {
+      // one bucket, use values source bucket builder for getting better performance
+      return Pair.of(
+          Collections.singletonList(
+              bucketBuilder.build(groupByList.getFirst()).subAggregations(metrics.getLeft())),
+          new BucketAggregationParser(metrics.getRight()));
     } else {
+      // multiple bucket, use composite builder
       GroupSortOrder groupSortOrder = new GroupSortOrder(sortList);
       return Pair.of(
           Collections.singletonList(
               AggregationBuilders.composite(
                       "composite_buckets",
-                      bucketBuilder.build(
+                      compositeBuilder.build(
                           groupByList.stream()
                               .sorted(groupSortOrder)
                               .map(
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/BucketAggregationParser.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/BucketAggregationParser.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.response.agg;
+
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import lombok.EqualsAndHashCode;
+import org.opensearch.search.aggregations.Aggregation;
+import org.opensearch.search.aggregations.Aggregations;
+import org.opensearch.search.aggregations.bucket.MultiBucketsAggregation;
+
+/**
+ * Use BucketAggregationParser only when there is a single group-by key, it returns multiple
+ * buckets. {@link CompositeAggregationParser} is used for multiple group by keys
+ */
+@EqualsAndHashCode
+public class BucketAggregationParser implements OpenSearchAggregationResponseParser {
+  private final MetricParserHelper metricsParser;
+
+  public BucketAggregationParser(MetricParser... metricParserList) {
+    metricsParser = new MetricParserHelper(Arrays.asList(metricParserList));
+  }
+
+  public BucketAggregationParser(List<MetricParser> metricParserList) {
+    metricsParser = new MetricParserHelper(metricParserList);
+  }
+
+  @Override
+  public List<Map<String, Object>> parse(Aggregations aggregations) {
+    Aggregation agg = aggregations.asList().getFirst();
+    return ((MultiBucketsAggregation) agg)
+        .getBuckets().stream().map(b -> parse(b, agg.getName())).collect(Collectors.toList());
+  }
+
+  private Map<String, Object> parse(MultiBucketsAggregation.Bucket bucket, String keyName) {
+    Map<String, Object> resultMap = new LinkedHashMap<>();
+    resultMap.put(keyName, bucket.getKey());
+    resultMap.putAll(metricsParser.parse(bucket.getAggregations()));
+    return resultMap;
+  }
+}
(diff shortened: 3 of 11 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/aggregation/AggregationQueryBuilder.java
```diff
@@ -74,13 +80,20 @@ public AggregationQueryBuilder(ExpressionSerializer serializer) {
       return Pair.of(
           ImmutableList.copyOf(metrics.getLeft().getAggregatorFactories()),
           new NoBucketAggregationParser(metrics.getRight()));
+    } else if (groupByList.size() == 1) {
+      // one bucket, use values source bucket builder for getting better performance
```
LantaoJin, line 84: typo here, should be `one group-by key`, not `one bucket`

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -126,6 +128,12 @@ public static Pair<List<AggregationBuilder>, OpenSearchAggregationResponseParser
         return Pair.of(
             ImmutableList.copyOf(metricBuilder.getAggregatorFactories()),
             new NoBucketAggregationParser(metricParserList));
+      } else if (aggregate.getGroupSet().length() == 1) {
```
penghuo, line 131: CompositeAgg is slow only for single group by case?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -266,4 +274,26 @@ private static CompositeValuesSourceBuilder<?> createTermsSourceBuilder(
...
+    // TODO: support histogram bucket(i.e. PPL span expression)
+    // https://github.com/opensearch-project/sql/issues/3384
+    return createTermsAggregationBuilder(groupExpr);
+  }
+
+  private static TermsAggregationBuilder createTermsAggregationBuilder(
+      NamedFieldExpression groupExpr) {
+
+    TermsAggregationBuilder sourceBuilder =
+        new TermsAggregationBuilder(groupExpr.getRootName()).size(AGGREGATION_BUCKET_SIZE);
```
penghuo, line 290: what if buckets more than AGGREGATION_BUCKET_SIZE?

LantaoJin: Only 1000 buckets return. The default size is 10. This setting is aligned with previous implementation of composite agg which set 1000 buckets.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
LantaoJin: No, composite aggregation is slow for all cases. If there are multiple group-by keys, we have to use composite aggregation with various sources, isn't it?

LantaoJin:
The IT failures are caused by https://github.com/opensearch-project/OpenSearch/issues/17959. It blocks our optimization work.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: Shall we also report this issue to core for optimizing CompositeAgg with single group expression?

I think, since CompositeAgg with single group is logically equivalent to BucketAgg and user is able to write such DSL, it will be a good chance to improve the previous execution efficiency.

penghuo: > If there are multiple group-by keys, we have to use composite aggregation with various sources, isn't it?

DSL support multi-level terms aggregtion. for instance
```
GET /_search
{
  "aggs": {
    "countries": {
      "terms": {
        "field": "artist.country",
      },
      "aggs": {
        "rock": {
          "filter": { "term": { "genre": "rock" } },
          "aggs": {
            "playback_stats": { "stats": { "field": "play_count" } }
          }
        }
      }
    }
  }
}
```

The reasons V2 use composite aggregation is full support pagenation use case.

LantaoJin:
> Hi @LantaoJin , is there any update for this PR?

No, this fixing is blocked by https://github.com/opensearch-project/OpenSearch/issues/17959, I will convert it to draft until any update from that issue.

LantaoJin:
> > Hi @LantaoJin , is there any update for this PR?
> 
> No, this fixing is blocked by [opensearch-project/OpenSearch#17959](https://github.com/opensearch-project/OpenSearch/issues/17959), I will convert it to draft until any update from that issue.

The optimal fixing is still depended on [opensearch-project/OpenSearch#17959](https://github.com/opensearch-project/OpenSearch/issues/17959), but we can provide a command parameter to ignore null bucket values as a workaround, because there are many scenarios that the bucket values cannot be null, e.g grouping by timestamp.

Here is a real-world example, the Discover UI displays the "log count" by building a PPL in frontend
`... | stats count() by span(timestamp, 12h)` (we can get the generated PPL in the request payload)
Screenshot 2025-08-20 at 16 28 39

This PPL can be benefited by current PR since the `timestamp` would not be null.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
LantaoJin: Do you mean multi-terms https://docs.opensearch.org/latest/aggregations/bucket/multi-terms/?
```
GET /_search
{
  "size": 0, 
  "aggs": {
    "hot": {
      "multi_terms": {
        "terms": [{
          "field": "region" 
        },{
          "field": "host" 
        }],
        "order": [{
          "max-cpu": "desc"
        },{
          "max-memory": "desc"
        }]
      },
      "aggs": {
        "max-cpu": { "max": { "field": "cpu" } },
        "max-memory": { "max": { "field": "memory" } }
      }      
    }
  }
}
```
I am not sure the difference, anyway, let's optimize the multiple group-by keys aggregation in a followup PR to simplify this PR.

LantaoJin:
ping @penghuo @qianheng-aws @dai-chen

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -82,7 +82,7 @@ public void supportPushDownSortMergeJoin() throws IOException {
   @Ignore("We've supported script push down on text field")
   @Test
   public void supportPartialPushDown() throws IOException {
-    Assume.assumeTrue("This test is only for push down enabled", isPushdownEnabled());
+    Assume.assumeTrue("This test is only for push down enabled", !isPushdownDisabled());
```
LantaoJin: The reason is [here](https://github.com/opensearch-project/sql/pull/3550/files#diff-994881edfa74bfe25930ed2bb2a8fd40e90ac4e23fc83833eca168e8c72d6172R259-R263)

integ-test/build.gradle
```diff
@@ -451,7 +451,7 @@ integTest {
 
     dependsOn ':opensearch-sql-plugin:bundlePlugin'
     if(getOSFamilyType() != "windows") {
-        dependsOn startPrometheus
+//        dependsOn startPrometheus
```
penghuo: revert it

integ-test/src/test/java/org/opensearch/sql/ppl/StatsCommandIT.java
```diff
@@ -207,6 +208,23 @@ public void testAvgGroupByNullValue() throws IOException {
...
+                "source=%s | stats bucket_nullable=false avg(balance) as a by age",
+                TEST_INDEX_BANK_WITH_NULL_VALUES));
+    verifySchema(response, schema("a", null, "double"), schema("age", null, "int"));
+    verifyDataRows(
+        response,
+        rows(32838D, 28),
+        rows(39225D, 32),
+        rows(4180D, 33),
+        rows(48086D, 34),
+        rows(null, 36));
```
penghuo, line 225: null bucket should be ignored?

LantaoJin: this is the bucket 36 with null value

integ-test/src/test/java/org/opensearch/sql/ppl/StatsCommandIT.java
```diff
@@ -585,4 +722,31 @@ public void testStatsPercentileBySpan() throws IOException {
...
+            throw new RuntimeException(e);
+          }
+          verifySchema(response, schema("a", null, "double"), schema("age", null, "int"));
+          verifyDataRows(
+              response,
+              rows(32838D, 28),
+              rows(39225D, 32),
+              rows(4180D, 33),
+              rows(48086D, 34),
+              rows(null, 36));
```
LantaoJin, line 749: `36` is the bucket, `null` is metric

integ-test/build.gradle
LantaoJin: done

integ-test/src/test/resources/expectedOutput/calcite/explain_limit_agg_pushdown_bucket_nullable1.json
```diff
@@ -0,0 +1,6 @@
...
+    "physical": "EnumerableLimit(fetch=[10000])\n  EnumerableCalc(expr#0..1=[{inputs}], count()=[$t1], state=[$t0])\n    EnumerableLimit(offset=[10], fetch=[10])\n      CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[FILTER->IS NOT NULL($7), AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0},count()=COUNT()), LIMIT->100, LIMIT->[10 from 10]], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"size\":0,\"timeout\":\"1m\",\"query\":{\"exists\":{\"field\":\"state\",\"boost\":1.0}},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}],\"aggregations\":{\"state\":{\"terms\":{\"field\":\"state.keyword\",\"size\":20,\"min_doc_count\":1,\"shard_min_doc_count\":0,\"show_term_doc_count_error\":false,\"order\":{\"_key\":\"asc\"}},\"aggregations\":{\"count()\":{\"value_count\":{\"field\":\"_index\"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
```
penghuo: `\"sort\":[{\"_doc\":{\"order\":\"asc\"}}]` should be removed if it is aggregation

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -547,6 +547,20 @@ public void testPushdownLimitIntoAggregation() throws IOException {
...
+        explainQueryToString(
+            "source=opensearch-sql_test_index_account | stats bucket_nullable=false count() by"
+                + " state | head 100 | head 10 from 10 "));
+
+    expected = loadExpectedPlan("explain_limit_agg_pushdown_bucket_nullable2.json");
+    assertJsonEqualsIgnoreId(
+        expected,
+        explainQueryToString(
+            "source=opensearch-sql_test_index_account | stats bucket_nullable=false count() by"
+                + " state | sort state | head 100 | head 10 from 10 "));
```
penghuo, line 600: change to `sort - state` to verfy sort is pushed down correctlly, otherwise nullable1 and nullable2 execution plan is exactally same.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -547,6 +547,20 @@ public void testPushdownLimitIntoAggregation() throws IOException {
...
+
```
penghuo, line 601: Add a sort by metrics case
`source=opensearch-sql_test_index_account | stats bucket_nullable=false count() as c by state | sort - c`

integ-test/src/test/resources/expectedOutput/calcite/explain_limit_agg_pushdown_bucket_nullable1.json
LantaoJin: This is added by filter pushdown in https://github.com/opensearch-project/sql/blob/7de854542dd16ec9695162533673db6710cbde74/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java#L183-L185

And we don't have opportunity to remove it as long as it was added during filter pushdown. Not sure why we have this logic.

LantaoJin: The filter is pushed down first (add `\"sort\":[{\"_doc\":{\"order\":\"asc\"}}]` in above code), then agg is pushed down (no API to remove from source builder).

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
LantaoJin: No, the nullable1 and nullable2 are copied from pushdown3 and pushdown4 whose the differences are with/without a sort

LantaoJin: `desc` or `asc` impact nothing that a pushdown could be applied or not.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
LantaoJin: Ok, but I think without `multi-terms` bucket pushdown, there is no significant difference with/without this PR.

LantaoJin: added

integ-test/src/test/resources/expectedOutput/calcite/explain_limit_agg_pushdown_bucket_nullable1.json
LantaoJin: Oh, I found a way to empty it. Will update it again

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -594,6 +608,25 @@ public void testPushdownLimitIntoAggregation() throws IOException {
...
+  @Test
+  public void testExplainSortOnMetricsNoBucketNullable() throws IOException {
+    // TODO enhancement later: https://github.com/opensearch-project/sql/issues/4282
+    enabledOnlyWhenPushdownIsEnabled();
+    String expected = loadExpectedPlan("explain_agg_sort_on_metrics1.json");
+    assertJsonEqualsIgnoreId(
+        expected,
+        explainQueryToString(
+            "source=opensearch-sql_test_index_account | stats bucket_nullable=false count() by"
+                + " state | sort `count()`"));
```
penghuo, line 620: It is ok to track in sepereate issue, just want to clarify this case does not require multi-term aggregation.

integ-test/src/test/resources/expectedOutput/calcite/explain_limit_agg_pushdown_bucket_nullable1.json
LantaoJin: Fix it in https://github.com/opensearch-project/sql/pull/3550/commits/ac72fc36ba739502c90766d748cb8677d474a131

integ-test/src/test/resources/expectedOutput/calcite/explain_agg_sort_on_metrics1.json
```diff
@@ -0,0 +1,6 @@
...
+    "physical": "EnumerableLimit(fetch=[10000])\n  EnumerableSort(sort0=[$0], dir0=[ASC-nulls-first])\n    EnumerableCalc(expr#0..1=[{inputs}], count()=[$t1], state=[$t0])\n      CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[FILTER->IS NOT NULL($7), AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0},count()=COUNT())], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"size\":0,\"timeout\":\"1m\",\"query\":{\"exists\":{\"field\":\"state\",\"boost\":1.0}},\"sort\":[],\"aggregations\":{\"state\":{\"terms\":{\"field\":\"state.keyword\",\"size\":1000,\"min_doc_count\":1,\"shard_min_doc_count\":0,\"show_term_doc_count_error\":false,\"order\":{\"_key\":\"asc\"}},\"aggregations\":{\"count()\":{\"value_count\":{\"field\":\"_index\"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
```
penghuo, line 4: `size:1000`, without sort pushdown, it is a correctness issue. @LantaoJin  Should we add limitation for aggregation, max_aggregation_buckets = 1000?
Related to this https://github.com/opensearch-project/sql/issues/4278

LantaoJin: We don't know the cardinality of one bucket key, how to set a limit in SQL plugin? I think what we can do is change the current bucket size configurable rather than hard code (1000).

LantaoJin: But we could open an issue in OS core, the limitation should be enabled in OS core IMO.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
LantaoJin: I see, comment in https://github.com/opensearch-project/sql/issues/4282#issuecomment-3282761135

PR #3560, opened 2025-04-17
Support `explain` PPL command

### LINKED ISSUE #3328: [FEATURE] Support `explain` command with Calcite

When using calcite engine, it should support explain command to show the final calcite plan.

LantaoJin:
### Description
PPL `explain` command is an existing command in PPL-Spark, [PR](https://github.com/opensearch-project/opensearch-spark/pull/687). This PR implement this command both in V2 and V3(Calcite).

#### Syntax
`explain [mode] queryStatement`

#### mode:
- standard (default if not specified): display logical and physical plan
- simple: display logical plan without attributes
- extended: display logical and physical plan and generated code
- cost: display logical and physical plan and its cost

if calcite is disabled, the explain command returns standard output in V2 `_explain` endpoint.

### Related Issues
Resolves #3328 

### Check List


core/src/main/java/org/opensearch/sql/executor/execution/QueryPlanFactory.java
```diff
@@ -121,36 +114,29 @@ public AbstractPlan visitQuery(
...
+              ResponseListener<ExecutionEngine.ExplainResponse>>
           context) {
-    Preconditions.checkArgument(
-        context.getRight().isPresent(), "[BUG] explain listener must be not null");
-
     return new ExplainPlan(
         QueryId.queryId(),
         node.getQueryType(),
-        create(node.getStatement(), Optional.of(NO_CONSUMER_RESPONSE_LISTENER), Optional.empty()),
+        create(node.getStatement(), NO_CONSUMER_RESPONSE_LISTENER, context.getRight()),
```
qianheng-aws, line 138: Since we never have explain command inner a explain command, it actually doesn't need to pass in a explain listener here.

qianheng-aws (commented):
If the explain process on Calcite throws any exception? Will it fallback to v2 engine? 

And please add test cases of failure for explain.

LantaoJin:
> * If the explain process on Calcite throws any exception? Will it fallback to v2 engine?

Yes.

> Please add test cases of failure for explain.

It's hard to make that test case

> * What's the expected response for v2 and sql for these new added explain format?

New added explain format will ignore in v2 and SQL

core/src/main/java/org/opensearch/sql/executor/execution/QueryPlanFactory.java
LantaoJin: We'd better not pass a null as explain listener. I will add a `requireNonNull` here as well.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -19,12 +19,24 @@ pplStatement
...
 dmlStatement
    : queryStatement
+   | explainStatement
    ;
 
 queryStatement
    : pplCommands (PIPE commands)*
    ;
 
+explainStatement
```
penghuo, line 29: OSD discovery assumes that PPL commands can be chained. However, the explain command is standalone and cannot be combined with other commands. @ashwin-pc could we do speicial handling for explain command? or we only allow search command in discovery?

penghuo (commented):
Please update explain command doc.

ppl/src/main/antlr/OpenSearchPPLParser.g4
ashwin-pc: @penghuo can we extend the grammar file to make it a part of the grammar itself?

penghuo: explain command will in g4 file.
my question is: "Is it suitable for the Discovery page?"

ananzh: OSD can update src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLLexer.g4
src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4

These updates ensure that the explain command is recognized as valid PPL syntax in OpenSearch Dashboards (OSD). 

Besides of this, current  discovery UI in OSD is built with the assumption that PPL commands can be chained with pipes (|). We could add special handling.

ananzh: We could modify the query editor to detect when an explain command is used and disable the ability to chain other commands. Then add validation to show an error if a user tries to chain commands after an explain command.

I mean if we want to support. Otherwise, we could restrict the usage to be chained.

core/src/main/java/org/opensearch/sql/executor/execution/QueryPlanFactory.java
qianheng-aws: It won't do check for explain listener actually if node.getStatement() is a `Query`

qianheng-aws:
> > Please add test cases of failure for explain.
> 
> It's hard to make that test case
Why? Can we just add tests on incorrect PPL, like unsupported function or columns cannot be resolved?

> > * What's the expected response for v2 and sql for these new added explain format?
> 
> New added explain format will ignore in v2 and SQL

Will it be better showing error to tell user that it's not a  supported format in v2 or SQL? Otherwise customer need to detect that information by comparing plans we returns, and then finally gets that it's actually meaningless to put such format in their PPL if using v2.

ppl/src/main/antlr/OpenSearchPPLParser.g4
LantaoJin: > Then add validation to show an error if a user tries to chain commands after an explain command.

Clarify:
```
explain source=t | where country = 'a' | stats count() by country
```
✅ 
```
source=t | where country = 'a' | stats count() by country | explain
```
❌ 

`explain` command could be only used as the first command in the PPL query. It is similar to `describe` command.

If  `describe`  is support in discovery, nothing should be changed in discovery? @penghuo @ananzh

LantaoJin: With above explanation, does Discovery need any change? @LantaoJin @ananzh

LantaoJin:
> > > Please add test cases of failure for explain.
> > 
> > 
> > It's hard to make that test case
> 
> Why? Can we just add tests on incorrect PPL, like unsupported function or columns cannot be resolved?

unsupported function or columns cannot be resolved are all throw V2 exceptions before https://github.com/opensearch-project/sql/pull/3571 being merged.

> > > * What's the expected response for v2 and sql for these new added explain format?
> > 
> > 
> > New added explain format will ignore in v2 and SQL
> 
> Will it be better showing error to tell user that it's not a supported format in v2 or SQL? Otherwise customer need to detect that information by comparing plans we returns, and then finally gets that it's actually meaningless to put such format in their PPL if using v2.

I don't think throwing error is a good idea when Calcite disabled. It somewhat break the fallback mechanism. We can document some modes (simple, cost, extended) now is only for Calcite. Maybe in future we can implement these modes for V2 if anyone is interested.

LantaoJin:
> Please update explain command doc.

added

docs/user/ppl/cmd/explain.rst
```diff
@@ -0,0 +1,192 @@
+=============
+explain
```
LantaoJin, line 2: FYI: this doc is not executable since partial example requires calcite enabled.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
```diff
@@ -100,10 +100,11 @@ public void onResponse(TransportPPLQueryResponse response) {
...
                   LOG.error("Error happened during explain", e);
-                  sendResponse(
-                      channel,
-                      INTERNAL_SERVER_ERROR,
-                      "Failed to explain the query due to error: " + e.getMessage());
+                  if (isClientError(e)) {
+                    reportError(channel, e, BAD_REQUEST);
+                  } else {
+                    reportError(channel, e, INTERNAL_SERVER_ERROR);
+                  }
```
LantaoJin, line 107: FYI: This change is fixing an existing bug which is unable to display the explain failure.

penghuo (approved):

qianheng-aws (approved):

PR #3562, opened 2025-04-18
[POC]Support type checker For PPL functions

qianheng-aws:
### Description
Calcite has its type checker mechanism to do validation for its function calls. However, this process happens in SqlNode level and its API only accepts SqlNode level objects like `SqlCallBinding`, `SqlNode` and etc. Since we use RelNode directly, this validation step has been ignored by our current process.

In order to reuse the definition of Calcite's builtin operators, we have to do validation by ourself with an approach to reuse the definition of Calcite's operators.

Luckily, in Calcite, almost all operators use `FamilyOperandTypeChecker` as their type checker with API `getOperandSqlTypeFamily` exposed to us to make use of. We can use this API to do validation conveniently.

 ### Note
This is a draft PR with only one test cases to verify the functionality. 

Before this PR, the test `CalcitePPLFunctionTypeTest::testLowerWithIntegerType` will return a incorrect plan for its PPL:

```
source=EMP | eval lower_name = lower(EMPNO) | fields lower_name


LogicalProject.NONE.[](input=LogicalTableScan#0,exprs=[LOWER($0)])
     LogicalTableScan.NONE.[0](table=[scott, EMP])
```
The above plan is incorrect because $0(i.e. EMPNO) has type of integer, so it shouldn't be passed into function `LOWER` which only allows `SqlTypeFamily.CHARACTOR`.  The error will only be thrown until execution step.
```
java.lang.RuntimeException: java.sql.SQLException: Error while preparing plan [LogicalProject(lower_name=[LOWER($0)])
  BindableTableScan(table=[[scott, EMP]])
]
```

With this PR, the ppl will throw exception in building logical plan because of unsupported function.

### Related Issues

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/CalciteFuncSignature.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/CalciteFuncSignature.java
@@ -5,26 +5,15 @@
 
 package org.opensearch.sql.expression.function;
 
-import static org.opensearch.sql.expression.function.PPLFuncImpTable.FunctionImp.ANY_TYPE;
-
 import java.util.List;
 import org.apache.calcite.rel.type.RelDataType;
 
 /** Function signature is composed by function name and arguments list. */
-public record CalciteFuncSignature(FunctionName functionName, List<RelDataType> funcArgTypes) {
+public record CalciteFuncSignature(FunctionName functionName, PPLTypeChecker typeChecker) {
 
   public boolean match(FunctionName functionName, List<RelDataType> paramTypeList) {
-    if (funcArgTypes == null) return true;
-    if (!functionName.equals(this.functionName()) || paramTypeList.size() != funcArgTypes.size()) {
-      return false;
-    }
-    for (int i = 0; i < paramTypeList.size(); i++) {
-      RelDataType paramType = paramTypeList.get(i);
-      RelDataType funcType = funcArgTypes.get(i);
-      if (ANY_TYPE != funcType && paramType.getFamily() != funcType.getFamily()) {
-        return false;
-      }
-    }
-    return true;
+    if (!functionName.equals(this.functionName())) return false;
+    if (typeChecker == null) return true;
+    return typeChecker.checkOperandTypes(paramTypeList);
   }
 }
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
@@ -6,7 +6,7 @@
 package org.opensearch.sql.expression.function;
 
 import static java.lang.Math.E;
-import static org.opensearch.sql.calcite.utils.OpenSearchTypeFactory.TYPE_FACTORY;
+import static org.apache.calcite.sql.type.SqlTypeFamily.IGNORE;
 import static org.opensearch.sql.calcite.utils.OpenSearchTypeFactory.getLegacyTypeName;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.ABS;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.ACOS;
@@ -67,6 +67,8 @@ import static org.opensearch.sql.expression.function.BuiltinFunctionName.TRIM;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.TYPEOF;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.UPPER;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.XOR;
+import static org.opensearch.sql.expression.function.PPLTypeChecker.family;
+import static org.opensearch.sql.expression.function.PPLTypeChecker.familyWrapper;
 
 import com.google.common.collect.ImmutableMap;
 import java.math.BigDecimal;
@@ -83,6 +85,9 @@ import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.fun.SqlTrimFunction.Flag;
+import org.apache.calcite.sql.type.ImplicitCastOperandTypeChecker;
+import org.apache.calcite.sql.type.SqlOperandTypeChecker;
+import org.apache.calcite.sql.type.SqlTypeFamily;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.opensearch.sql.executor.QueryType;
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/opensearch/sql/expression/function/CalciteFuncSignature.java
```diff
@@ -5,26 +5,15 @@
...
 package org.opensearch.sql.expression.function;
 
-import static org.opensearch.sql.expression.function.PPLFuncImpTable.FunctionImp.ANY_TYPE;
-
 import java.util.List;
 import org.apache.calcite.rel.type.RelDataType;
 
 /** Function signature is composed by function name and arguments list. */
-public record CalciteFuncSignature(FunctionName functionName, List<RelDataType> funcArgTypes) {
+public record CalciteFuncSignature(FunctionName functionName, PPLTypeChecker typeChecker) {
```
LantaoJin, line 12: `PPLTypeChecker` -> `FunctionParamsTypeChecker` could be more readable.

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.calcite.sql.type.SqlTypeFamily;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+public interface PPLTypeChecker {
+  boolean checkOperandTypes(List<RelDataType> types);
+
+  private static boolean validateOperands(
+      List<SqlTypeFamily> funcTypeFamilies, List<RelDataType> operandTypes) {
+    if (funcTypeFamilies.size() != operandTypes.size()) {
+      return true; // Skip checking if sizes do not match because some arguments may be optional
```
LantaoJin, line 23: how to handle the mismatch case with optional argument

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
```diff
@@ -0,0 +1,78 @@
...
+      this.families = List.of(families);
+    }
+
+    public boolean checkOperandTypes(List<RelDataType> types) {
+      if (families.size() != types.size()) return false;
+      return validateOperands(families, types);
+    }
+  }
+
+  class PPLFamilyTypeCheckerWrapper implements PPLTypeChecker {
```
LantaoJin, line 51: What is this `PPLFamilyTypeCheckerWrapper` used for?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -117,13 +120,13 @@ default RexNode resolve(RexBuilder builder, RexNode... args) {
...
-    default List<RelDataType> getParams() {
-      return ANY_TYPE_1;
+    default PPLTypeChecker getTypeChecker() {
+      return IGNORE_1;
     }
   }
 
   public interface FunctionImp2 extends FunctionImp {
-    List<RelDataType> ANY_TYPE_2 = List.of(ANY_TYPE, ANY_TYPE);
+    PPLTypeChecker IGNORE_2 = family(IGNORE, IGNORE);
```
LantaoJin, line 129: Why does the checker name to IGNORE_N?

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
qianheng-aws: It won't do check for the case that has optional arguments, which has the same behavior as the type checker in Calcite. It should throw exception until execution as before.

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
qianheng-aws: It supports build a `PPLTypeChecker` from a `PPLFamilyTypeCheckerWrapper` of Calcite. The latter one only supports SqlNode level check, and we use this wrapper to make it support RelNode level checking per our requirements

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
qianheng-aws: In Calcite, IGNORE is a like like ANY but don't even validate the operand, while ANY accepts all types except CURSOR.

So IGNORE is a more suitable type here than ANY.

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
```diff
@@ -0,0 +1,78 @@
...
+public interface PPLTypeChecker {
+  boolean checkOperandTypes(List<RelDataType> types);
+
+  private static boolean validateOperands(
+      List<SqlTypeFamily> funcTypeFamilies, List<RelDataType> operandTypes) {
+    if (funcTypeFamilies.size() != operandTypes.size()) {
+      return true; // Skip checking if sizes do not match because some arguments may be optional
+    }
+    for (int i = 0; i < operandTypes.size(); i++) {
+      SqlTypeName paramType = operandTypes.get(i).getSqlTypeName();
```
yuancu, line 26: `relDataType.getSqlTypeName()` does not handle UDT. A function like the following may help:

```
public static SqlTypeName convertRelDataTypeToSqlTypeName(RelDataType type) {
    if (type instanceof ExprSqlType exprSqlType) {
      return switch (exprSqlType.getUdt()) {
        case EXPR_DATE -> SqlTypeName.DATE;
        case EXPR_TIME -> SqlTypeName.TIME;
        case EXPR_TIMESTAMP -> SqlTypeName.TIMESTAMP;
        default -> type.getSqlTypeName();
      };
    }
    return type.getSqlTypeName();
  }
  ```

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
yuancu: Maybe `sqlOperandTypeChecker.isOptional(int i)` combined with `sqlOperandTypeChecker.or(SqlOperandTypeChecker checker)` can be used to check functions with multiple signatures?

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
qianheng-aws: These UDT should maps to SqlTypeName.VARCHAR actually. And we have the API to return the related SqlTypeName of our UDT here:

https://github.com/opensearch-project/sql/blob/8134c4e4fdb637ff68624684b7bed3c938f329fa/core/src/main/java/org/opensearch/sql/calcite/type/AbstractExprRelDataType.java#L101

LantaoJin:
resolved by https://github.com/opensearch-project/sql/pull/3626

Issue #3567, opened 2025-04-21
[BUG] Runtime exception in OpenSearchExecutionEngine should be thrown


LantaoJin:
**What is the bug?**
Query
```
source = http_logs | parse request 'GET /(?<domain>[a-zA-Z]+)/(?<picName>[a-zA-Z]+)/.*' | where domain = 'english' | sort - picName | fields domain, picName | head 10
```
will throw a known expection: `Multiple capturing groups (count=2) not allowed in regex input for REGEXP_EXTRACT`
But it was caught by `OpenSearchExecutionEngine`.
```
[2025-04-21T18:44:34,953][ERROR][o.o.s.p.r.RestPPLQueryAction] [integTest-0] Error happened during query handling
java.sql.SQLException: exception while executing query: Multiple capturing groups (count=2) not allowed in regex input for REGEXP_EXTRACT
        at org.apache.calcite.avatica.Helper.createException(Helper.java:56) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.apache.calcite.avatica.Helper.createException(Helper.java:41) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.apache.calcite.avatica.AvaticaConnection.executeQueryInternal(AvaticaConnection.java:579) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.apache.calcite.avatica.AvaticaPreparedStatement.executeQuery(AvaticaPreparedStatement.java:137) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.opensearch.sql.opensearch.executor.OpenSearchExecutionEngine.lambda$execute$6(OpenSearchExecutionEngine.java:177) [opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at java.base/java.security.AccessController.doPrivileged(AccessController.java:319) [?:?]
[...]

PR #3445, opened 2025-03-19
[Calcite Engine]Support metadata field

qianheng-aws:
### Description
Support metadata field in Calcite. This PR also include changes:
1. Fix the bug on v2 engine that cannot evaluate metadata field if it's not pushdown
2. Add limitation of using metadata field on eval, alias and rename

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3333

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3571, opened 2025-04-22
New added commands should throw exception when calcite disabled

### LINKED ISSUE #3568: [BUG] Join/Lookup/Subsearch should throw exception in Analyzer (V2) when calcite disabled.
**What is the bug?**
If Calcite disabled, Join/Lookup/Subsearch commands return incorrect (right) output instead of throwing exception. We should update Analyzer (V2) to throw exception.

LantaoJin:
### Description
New added commands should throw exception when calcite disabled

### Related Issues
Resolves #3568 and #3567

### Check List


qianheng-aws:
If customer set calcite and fallback both enabled, and submit an incorrect query with new commands, it seems will return error "xxx is supported only when plugins.calcite.enabled is enabled". 

But the expected error message should be the exception message thrown by calcite.

integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
```diff
@@ -147,27 +147,39 @@ public static void disallowCalciteFallback() throws IOException {
     LOG.info("{} disabled", Settings.Key.CALCITE_FALLBACK_ALLOWED.name());
   }
 
-  protected static boolean isFallbackEnabled() throws IOException {
-    return Boolean.parseBoolean(
-        getClusterSetting(Settings.Key.CALCITE_FALLBACK_ALLOWED.getKeyValue(), "persistent"));
+  protected static boolean isFallbackEnabled() {
```
qianheng-aws, line 150: Just curious, what's the concern of changing IOException to RuntimeException.

qianheng-aws:
> If customer set calcite and fallback both enabled, and submit an incorrect query with new commands, it seems will return error "xxx is supported only when plugins.calcite.enabled=true" as it will always fallback to v2.
> 
> But the expected error message should be the exception message thrown by calcite.

Maybe we should show Calcite's error message, if fallback to v2 also ends with failure.

integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
LantaoJin: revert

LantaoJin:
> > If customer set calcite and fallback both enabled, and submit an incorrect query with new commands, it seems will return error "xxx is supported only when plugins.calcite.enabled=true" as it will always fallback to v2.
> > But the expected error message should be the exception message thrown by calcite.
> 
> Maybe we should show Calcite's error message, if fallback to v2 also ends with failure.

addressed.

LantaoJin:
The last second commit fails due to https://github.com/opensearch-project/sql/pull/3445#issuecomment-2821414946

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -685,6 +688,18 @@ public LogicalPlan visitCloseCursor(CloseCursor closeCursor, AnalysisContext con
     return new LogicalCloseCursor(closeCursor.getChild().get(0).accept(this, context));
   }
 
+  @Override
```
penghuo, line 691: all following new command and new functions need to follow this pattern to correctly show error messages? is there a way to enfore it?

LantaoJin: >all following new command and new functions need to follow this pattern to correctly show error messages? 

Only new commands need to follow this. For new functions, it will throw `unsupported function name` when calcite disabled.

> is there a way to enforce it?

Yes. When adding a new `visit` method to AbstractNodeVisitor, we should enforce to implement it in its derived classes. This can be done by making the `visit` method of the new command to abstract.
For example:
```
public abstract T visitScalarSubquery(ScalarSubquery node, C context);
```
instead of
```
  public T visitScalarSubquery(ScalarSubquery node, C context) {
  }
```
However, the downside is that all derived classes must implement it, and currently we have too many derived classes. Any thoughts?
Screenshot 2025-04-23 at 10 24 07

penghuo (dismissed):

integ-test/src/test/java/org/opensearch/sql/ppl/NewAddedCommandsIT.java
```diff
@@ -0,0 +1,118 @@
...
+    super.init();
+    loadIndex(Index.BANK);
+    loadIndex(Index.DOG);
+  }
+
+  @Test
+  public void testJoin() throws IOException {
+    JSONObject result;
+    try {
+      result =
```
qianheng-aws, line 32: Why not using assertThrow? This kind of tests will pass as well if no exception is thrown, but that's not the expected behavior actually.

LantaoJin: The CalciteNewAddedCommandsIT expects no exception. If no exception thrown in this class, the assertThat in L111 will fail.

core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -163,16 +172,26 @@ public void executeWithLegacy(
...
+      ResponseListener<ExecutionEngine.ExplainResponse> listener,
+      Optional<Throwable> calciteFailure) {
     try {
       executionEngine.explain(plan(analyze(plan, queryType)), listener);
     } catch (Exception e) {
-      listener.onFailure(e);
+      if (shouldUseCalcite(queryType) && isCalciteFallbackAllowed()) {
+        // if there is a failure thrown from Calcite and execution after fallback V2
+        // keeps failure, we should throw the failure from Calcite.
+        calciteFailure.ifPresentOrElse(
```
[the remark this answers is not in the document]
LantaoJin: done

integ-test/src/test/java/org/opensearch/sql/ppl/NewAddedCommandsIT.java
qianheng-aws: Then it should override to has its own verification logic instead of reusing its parent's. 

Otherwise, these tests verify nothing. Without the change you've made in the production code, these IT can pass as well, right?

And I wonder if we really need `CalciteNewAddedCommandsIT`. Seems it verify nothing but only that the new command can be executed successfully on Calcite. However, we don't even check the results of execution.

qianheng-aws (approved):

integ-test/src/test/java/org/opensearch/sql/ppl/NewAddedCommandsIT.java
LantaoJin: Sync offline.
`CalciteNewAddedCommandsIT` is just for the IT framework style which each remote IT has a peer Calcite IT. We just to assert no failure thrown in `CalciteNewAddedCommandsIT` instead of asserting result.

PR #3504, opened 2025-03-31
[Calcite Engine] Support UDT for IP

qianheng-aws:
### Description
Support UDT for IP, also do refactoring for ExprBasicSqlType.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3395

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3576, opened 2025-04-23
Migrate existing UDFs to PPLFuncImpTable

### LINKED ISSUE #3524: [Umbrella] Migrate existing functions to PPLFuncImpTable
We have new framework for resolving functions. Need to migrate all existing functions to that framework.

yuancu:
### Description
Refactor existing math, datetime, text, condition UDFs to PPLFuncImpTable.

**Why self-implementing instead of leveraging Calcite builtin functions** (WIP)

- **Not implemented by calcite**
  - CONV
  - CRC32 (implemented in a newer version of calcite)
  - E
  - LOCATE

- **Different definition**
  - MOD & DIVIDE: return types (ours return wider type; calcite implementation return divisor type); handling of divisor 0 (ours return null)

- **Datetime related**: we use UDT and accepts STRING most of the time.

Note:
- UDAFs are kept untouched.

TODOs:
- [x] Fix filled date problems related to function properties.

### Related Issues
Resolves #3524 

### Check List


LantaoJin:
> I have a high level question: Previously, the UDF function implementation is built upon Java type system. As we consider introducing the same PPL-Calcite library to Flint Spark, the UDF is possible to be reused to avoid duplicate code.
> 
> However, the new implementation seems tightly coupled with Calcite type system and our ExprValue. Are these standard functions or UDF actually? I'm not sure whether they have the same semantic in Spark or not.

Link to the framework refactor PR https://github.com/opensearch-project/sql/pull/3522

qianheng-aws:
> I have a high level question: Previously, the UDF function implementation is built upon Java type system. As we consider introducing the same PPL-Calcite library to Flint Spark, the UDF is possible to be reused to avoid duplicate code.
> 
> However, the new implementation seems tightly coupled with Calcite type system and our ExprValue. Are these standard functions or UDF actually? I'm not sure whether they have the same semantic in Spark or not.

Let's split the functions into 2 parts:
- For these standard functions, since Calcite has its own implementation already, I think it should use Calcite's implementation directly. In Spark, I think it should has its own implementation as well and we can translate them like what we do on Calcite.
- For the UDF specific for PPL, the new implementation will keep the java-level implementation as well, which should have the same code as the previous version. The difference is that we need to wrap the implementation into a static Calcite's SqlOperator. When migrating to Spark, we can still reuse the java-level implementation as well. For example, see the code difference of `CRC32Function.java`, `Crc32Implementor::crc32` should cover the same functionality as [`CRC32Function::eval`](https://github.com/opensearch-project/sql/blob/47e6a2a619c194d159b6505b3d89364416045094/core/src/main/java/org/opensearch/sql/calcite/udf/mathUDF/CRC32Function.java#L18). If the implementation of previous version are easy to migrate, so does the new one.

For the concern of using ExprValue, it's because we want to reuse the existing logic of v2 engine without extra developing effort. If we want to deprecate ExprValue in the future, we must migrate logic of the reusing part into a code implementation without ExprValue.

qianheng-aws:
The biggest concern, I think, is the gap of type system between opensearch-sql and opensearch-spark.

For example, we use String to store value of  time related type like TIMESTAMP, TIME, DATE because we must keep the nanos precision as v2 engine. However, we translate these types to Spark's DataType like TimeStampType, and that will be transformed into `java.sql.TimeStamp` or `Instant` by catalyst. I'm incline to unify them to use Spark's catalyst type.

For `IP` type, we change from `String` to use `ExprIPValue` which wraps a IPAddress internally. We make that [change](https://github.com/opensearch-project/sql/pull/3504) because the latter one has its own specific comapring logic and we want to align with v2. However, in opensearch-spark, it keeps using Spark's `StringType`, i.e. java string in its catalyst.

qianheng-aws:
> IF: SqlLibraryOperators.IF throws not a rex operator error when IF is translated to a RexCall. (See [here](https://github.com/apache/calcite/blob/36f6dddd894b8b79edeb568982251dec8c9d2fc0/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java#L278))

It's because Calcite convert IF to CASE when converting SqlNode to RelNode. So IF is an operator only exist in SqlNode and doesn't have its physical implementation registered in RexImpTable. That's why it didn't accept RexNode as its oprand.

How about following Calcite's convention and using CASE to implement IF as well in our implementation? Then there's no need to implement IF.

core/src/main/java/org/opensearch/sql/calcite/utils/UserDefinedFunctionUtils.java
```diff
@@ -233,25 +121,120 @@ public static String formatTimestampWithoutUnnecessaryNanos(LocalDateTime localD
...
+        Instant.ofEpochSecond(
+            currentTimeInNanos / 1_000_000_000, currentTimeInNanos % 1_000_000_000);
+    TimeZone timeZone = TimeZone.getDefault();
+    ZoneId zoneId = timeZone.toZoneId();
+    return new FunctionProperties(instant, zoneId, QueryType.PPL);
+  }
+
+  public static List<Expression> addTypeAndContext(
+      List<Expression> operands, RexCall rexCall, Expression root) {
+    // Box primitive values so that they can match signatures with boxed types
```
qianheng-aws, line 140: Curious what's the issue if not boxing the original expression.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -354,6 +352,15 @@ void populate() {
...
                   builder.makeLiteral(getLegacyTypeName(arg.getType(), QueryType.PPL)));
       register(XOR, new XOR_FUNC());
+      register(
+          NULLIF,
+          (FunctionImp2)
+              (builder, arg1, arg2) ->
+                  builder.makeCall(
+                      SqlStdOperatorTable.CASE,
+                      builder.makeCall(SqlStdOperatorTable.EQUALS, arg1, arg2),
+                      builder.makeLiteral(null, TYPE_FACTORY.createSqlType(SqlTypeName.NULL)),
```
qianheng-aws, line 361: Please use `builder.makeNullLiteral(arg1.getType())`.

There is also `constantNull()` in RexBuilder which is equal to the object you create here, but it said to be deprecated in 2.0.

qianheng-aws:
Could you please add simple examples as well as functionality description for each UDF? Something like Spark does: https://github.com/apache/spark/blob/676cac092540755e04a5ba7eaaf1e5aea8138930/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala#L306C29-L306C30

It would help improve the code readability. Thanks!

core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
```diff
@@ -8,19 +8,266 @@
...
+              "exprMinuteOfDay",
+              UserDefinedFunctionUtils.INTEGER_FORCE_NULLABLE,
+              NullPolicy.ARG0)
+          .toUDF("MINUTE_OF_DAY");
+  public static final SqlOperator SECOND = new DatePartFunction(TimeUnit.SECOND).toUDF("SECOND");
+  public static final SqlOperator SECOND_OF_MINUTE =
+      new DatePartFunction(TimeUnit.SECOND).toUDF("SECOND_OF_MINUTE");
+  public static final SqlOperator MICROSECOND =
+      new DatePartFunction(TimeUnit.MICROSECOND).toUDF("MICROSECOND");
+  public static final SqlOperator CURRENT_TIMESTAMP =
```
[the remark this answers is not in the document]
yuancu: All of them are synonyms for [NOW()](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/datetime.rst#now). My concern is that customers may be confused if they call `LOCALTIME` in PPL but read `NOW` in the logical plan. But this should be a minor problem as users can make sense of it. After all, we also registered e.g. `IF` to `CASE`.

If we register the above now-like functions to the same operator, we may need to change corresponding logical plan tests, and we shall better register other synonyms functions to the same operators.

core/src/main/java/org/opensearch/sql/calcite/utils/UserDefinedFunctionUtils.java
yuancu: If I don't box it, a `long` datetime argument  can not match a signature like `toSeconds(Object datetime, SqlTypeName datetimeType, DataContext propertyContext)` or `eval(Object ...args)`, where the `datetime` may be long  or string.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
yuancu: done

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/DateAddSubFunction.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return opBinding ->
+        DateAddSubImplementor.deriveReturnType(
+            opBinding.getOperandType(0), opBinding.getOperandType(1), alwaysReturnTimestamp);
+  }
+
+  public static class DateAddSubImplementor implements NotNullImplementor {
```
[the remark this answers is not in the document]
yuancu: Now I split them into two classes: `DateAddSubFunction` and `AddSubFunction`. But I find it hard to reuse some repetitive code with extensions, e.g. [here](https://github.com/yuancu/sql-plugin/blob/adfc2311bb7a48f882afdb52a0d53d35204172e6/core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/AddSubDateFunction.java#L80) and [here](https://github.com/yuancu/sql-plugin/blob/adfc2311bb7a48f882afdb52a0d53d35204172e6/core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/DateAddSubFunction.java#L65)

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/DateAddSubFunction.java
```diff
@@ -0,0 +1,145 @@
...
+      Expression temporal = translatedOperands.get(0);
+      Expression temporalDelta = translatedOperands.get(1);
+      RelDataType temporalType = call.getOperands().get(0).getType();
+      RelDataType temporalDeltaType = call.getOperands().get(1).getType();
+      SqlTypeName returnType =
+          OpenSearchTypeFactory.convertRelDataTypeToSqlTypeName(
+              deriveReturnType(temporalType, temporalDeltaType, alwaysReturnTimestamp));
+
+      if (SqlTypeFamily.NUMERIC.contains(temporalDeltaType)) {
+        return Expressions.call(
```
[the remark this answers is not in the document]
yuancu: Resolved

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/CurrentFunction.java
```diff
@@ -0,0 +1,88 @@
...
+ * Implementation of the now-like functions: CURRENT_TIMESTAMP, NOW, LOCALTIMESTAMP, LOCALTIME,
+ * CURTIME, CURRENT_TIME, CURRENT_DATE, CUR_DATE
+ *
+ * <p>It returns the current date, time, or timestamp based on the specified return type.
+ */
+public class CurrentFunction extends ImplementorUDF {
+  private final SqlTypeName returnType;
+
+  public CurrentFunction(SqlTypeName returnType) {
+    super(new CurrentFunctionImplementor(returnType), NullPolicy.ANY);
```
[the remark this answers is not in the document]
yuancu: Yes, that makes more sense.

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/CurrentFunction.java
```diff
@@ -0,0 +1,88 @@
...
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+
+      return Expressions.call(
+          CurrentFunctionImplementor.class,
+          "current",
+          Expressions.constant(returnType),
+          translator.getRoot());
+    }
+
+    public static Object current(SqlTypeName returnType, DataContext propertyContext) {
```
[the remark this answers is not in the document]
yuancu: Separating them are indeed more efficient in execution. Resolved.

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/CurrentFunction.java
```diff
@@ -0,0 +1,88 @@
...
+public class CurrentFunction extends ImplementorUDF {
+  private final SqlTypeName returnType;
+
+  public CurrentFunction(SqlTypeName returnType) {
+    super(new CurrentFunctionImplementor(returnType), NullPolicy.ANY);
+    this.returnType = returnType;
+  }
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
```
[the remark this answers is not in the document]
yuancu: I created [PPLReturnTypes](https://github.com/yuancu/sql-plugin/blob/adfc2311bb7a48f882afdb52a0d53d35204172e6/core/src/main/java/org/opensearch/sql/calcite/utils/PPLReturnTypes.java#L16) to hold common `SqlReturnTypeInference` that are not covered by `org.apache.calcite.sql.type.ReturnTypes`. Yet return type inferences that are not shared among UDFs are still placed in individual UDF implementations.

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/CurrentFunction.java
```diff
@@ -0,0 +1,88 @@
...
+import org.opensearch.sql.expression.function.ImplementorUDF;
+
+/**
+ * Implementation of the now-like functions: CURRENT_TIMESTAMP, NOW, LOCALTIMESTAMP, LOCALTIME,
+ * CURTIME, CURRENT_TIME, CURRENT_DATE, CUR_DATE
+ *
+ * <p>It returns the current date, time, or timestamp based on the specified return type.
+ */
+public class CurrentFunction extends ImplementorUDF {
+  private final SqlTypeName returnType;
```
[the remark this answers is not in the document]
yuancu: My main concern is that `RelDataType` is not canonical: `SqlTypeName` is enum. Therefore a date will always have `SqlTypeName.DATE` as its sql type. However, [`RelDataType`](https://calcite.apache.org/javadocAggregate/org/apache/calcite/rel/type/RelDataType.html) represents the type of a scalar expression or entire row returned from a relational expression. It contains other information like precision, scale, nullability, etc. When such **objects** are used in comparison, this may lead to potential bugs. E.g. an operand `RelDataType` with digest `EXPR_TIMESTAMP VARCHAR` is not equal to another commonly used RelDataType `NULLABLE_TIMESTAMP_UDT`.

dai-chen:
@qianheng-aws @LantaoJin Thanks for the context! I need to do more deep dive as I begin the proof of concept.

core/src/main/java/org/opensearch/sql/expression/datetime/DateTimeFunctions.java
```diff
@@ -1211,7 +1211,7 @@ private ExprValue exprAddDateDays(
    * @param isAdd A flag: true to add, false to subtract.
    * @return Timestamp calculated.
    */
-  private ExprValue exprDateApplyDays(
+  public static ExprValue exprDateApplyDays(
```
qianheng-aws, line 1214: Ditto the comment https://github.com/opensearch-project/sql/pull/3576#discussion_r2059766881 here,  it would  impeding performance to add `if` in the implementation for attrs whose value we actually already know in the plan.

qianheng-aws: @yuancu  Please split Add and Sub into 2 funcs as well. Java ternary operator is equal to `if` actually.

yuancu: Done. Split add/sub functions are now implemented inside `AddSubDateFunction.AddSubDateImplementor`. Static functions in `DateTimeFunctions.java` are kept untouched as v2 implementations rely on them.

core/src/main/java/org/opensearch/sql/calcite/utils/PPLReturnTypes.java
```diff
@@ -26,4 +28,13 @@ private PPLReturnTypes() {}
       ReturnTypes.INTEGER.andThen(SqlTypeTransforms.FORCE_NULLABLE);
   public static SqlReturnTypeInference STRING_FORCE_NULLABLE =
       ReturnTypes.VARCHAR.andThen(SqlTypeTransforms.FORCE_NULLABLE);
+  public static SqlReturnTypeInference TIME_APPLY_RETURN_TYPE =
+      opBinding -> {
+        RelDataType temporalType = opBinding.getOperandType(0);
+        if (OpenSearchTypeFactory.convertRelDataTypeToSqlTypeName(temporalType)
```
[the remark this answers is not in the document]
yuancu: I removed the `convertRelDataTypeToSqlTypeName` API and adopted `convertRelDataTypeToExprType` to compare types, so that I can use a more unified interface to compare not only UDT but also builtin types like String. The converted `ExprType` is also widely used in the conversion to `ExprValue` (to reuse v2's implementation).

core/src/main/java/org/opensearch/sql/calcite/utils/datetime/DateTimeApplyUtils.java
```diff
@@ -21,8 +22,10 @@
 import org.opensearch.sql.exception.SemanticCheckException;
 import org.opensearch.sql.expression.function.FunctionProperties;
 
-public interface DateTimeApplyUtils {
-  static Instant applyInterval(Instant base, Duration interval, boolean isAdd) {
+public final class DateTimeApplyUtils {
+  private DateTimeApplyUtils() {}
+
+  public static Instant applyInterval(Instant base, Duration interval, boolean isAdd) {
```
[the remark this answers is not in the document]
yuancu: Removed. It was only used in previous implementations.

core/src/main/java/org/opensearch/sql/expression/function/udf/math/SqrtFunction.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.calcite.adapter.enumerable.RexToLixTranslator;
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.linq4j.tree.Expressions;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+
+/** <code>SQRT(value)</code> returns the square root of a number. */
+public class SqrtFunction extends ImplementorUDF {
```
penghuo, line 20: calcite has sqrt function?

```
0: jdbc:calcite:model=src/test/resources/mode> select sqrt(9);
+--------+
| EXPR$0 |
+--------+
| 3.0    |
+--------+
1 row selected (0.588 seconds)
```

yuancu: `SQRT` is only declared in `SqlStdOperatorTable`, but not implemented in `RexImpTable`. I.e. there isn't a statement like `defineMethod(SQRT, BuiltInMethod.SQRT.method, NullPolicy.STRICT)` in `RexImpTable`.

If I register `sqrt` to the calcite SQRT operator, it will throw an error at runtime:

```
...
java.lang.RuntimeException: java.sql.SQLException: Error while preparing plan [LogicalProject(name=[$0], age=[$5], month=[$3])
              LogicalFilter(condition=[AND(=(SQRT(POWER($5, 2)), 30.0E0), =(CBRT(POWER($3, 3)), 4))])
    ...
    java.lang.IllegalStateException: Unable to implement EnumerableCalc(expr#0..11=[{inputs}], expr#12=[2], expr#13=[POWER($t5, $t12)], expr#14=[SQRT($t13)], expr#15=[30.0E0:DOUBLE], expr#16=[=($t14, $t15)], expr#17=[3], expr#18=[POWER($t3, $t17)], expr#19=[CBRT($t18)], expr#20=[4], expr#21=[=($t19, $t20)], expr#22=[AND($t16, $t21)], name=[$t0], age=[$t5], month=[$t3], $condition=[$t22]): rowcount = 2.25, cumulative cost = {102.25 rows, 2801.0 cpu, 0.0 io}, id = 1388
     CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_state_country]]): rowcount = 100.0, cumulative cost = {100.0 rows, 101.0 cpu, 0.0 io}, id = 1353
```

yuancu: @qianheng-aws pointed out that `SQRT` is converted to `POWER(*, 0.5)`. I've removed the sqrt udf and register the operator to `power` with 0.5 as the second parameter instead.

LantaoJin: > @qianheng-aws pointed out that `SQRT` is converted to `POWER(*, 0.5)`. I've removed the sqrt udf and register the operator to `power` with 0.5 as the second parameter instead.

Since Calcite already has the `SQRT` function definition in `SqlStdOperatorTable` and internal implementation (converting to `POWER(*, 0.5)` is also a kind of implementation), why do we still need to explicitly add this conversion in our side? @yuancu

yuancu: In our implementation, when using operators from Calcite, their implementations have to exist in `RexImpTable`. Otherwise, it will throw an error like `cannot translate call SQRT`. See [RexToLixTranslater.java#L1428](https://github.com/apache/calcite/blob/fcbbea2676d485c370255325cdfcb8a93694f06e/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java#L1428).

In Calcite SQL, the conversion happens in SqlToRelConverter. See [SqlNodeToRexConverterImpl.java#L59](https://github.com/apache/calcite/blob/fcbbea2676d485c370255325cdfcb8a93694f06e/core/src/main/java/org/apache/calcite/sql2rel/SqlNodeToRexConverterImpl.java#L59). However, we do not have access to `SqlCall` in our implementation. Therefore, we cannot directly reuse Calcite's conversion of `SQRT` function.

core/build.gradle
```diff
@@ -57,10 +57,10 @@ dependencies {
     api group: 'com.google.code.gson', name: 'gson', version: '2.8.9'
     api group: 'com.tdunning', name: 't-digest', version: '3.3'
     api "net.minidev:json-smart:${versions.json_smart}"
-    api('org.apache.calcite:calcite-core:1.38.0') {
+    api('org.apache.calcite:calcite-core:1.39.0') {
```
LantaoJin, line 60: Do not upgrade core dependency in this PRs.

LantaoJin (changes requested):

core/src/main/java/org/opensearch/sql/expression/function/udf/math/SqrtFunction.java
penghuo: @yuancu We can similar convert in PPLFuncImplTable.
```
      register(SQRT,
          ((FunctionImp1)
          (builder, arg1) -> builder.makeCall(SqlStdOperatorTable.POWER, arg1,
              builder.makeApproxLiteral(BigDecimal.valueOf(0.5)))));
```
@qianheng-aws IMO, PPLFuncImplTable support 2 ways to define function. 1) Use existing SqlFunctions. 2) UDF. Is it correct?

core/build.gradle
yuancu: Changes reverted.

core/src/main/java/org/opensearch/sql/expression/function/udf/math/SqrtFunction.java
yuancu: Yes. I have register `SQRT` in this fashion.

qianheng-aws (approved):

LantaoJin (approved):

penghuo (approved):

Issue #3477, opened 2025-03-25
[BUG] field name with `dot` is missing in output without `fields` command


LantaoJin:
**What is the bug?**
field name with dot (such as `test.occupation`) is missing in query output without fields command

**How can one reproduce the bug?**
```
PUT /ltjin-test1
{
  "mappings": {
    "properties": {
      "name": {
        "type": "keyword"
      },
      "test.occupation": {
        "type": "text"
      },
      "country": {
        "type": "text"
      },
      "salary": {
        "type": "integer"
      },
      "year": {
        "type": "integer"
      },
      "month": {
        "type": "integer"
      }
    }
  }
}
POST _bulk
{"index":{"_index" : "ltjin-test1","_id":"1"}}
{"name":"Jake","test.occupation":"Engineer","country":"England","salary":100000,"year":2023,"month":4}
```
the `test.occupation` field is missing in following query.
```
POST /_plugins/_ppl
{
  "query" : """
  source = ltjin-test1
  """
}
```
but we can output it by following query.
```
POST /_plugins/_ppl
{
  "query" : """
  source = ltjin-test1 | fields test.occupation
  """
}
```
**What is the expected behavior?**
the `test.occupation` field should output in query
```
POST /_plugins/_ppl
{
  "query" : """
  source = ltjin-test1
  """
}
```
should 

**What is your host/environment?**
 - Version 2.19

qianheng-aws:
It's hard to define this is a bug or `feature` of OpenSearch.

The root cause is the document you put into this index mismatches its mapping.

In the mapping of index, `test.occupation` is a nested field of `test` while it is viewed as a root field in the document.
```
GET /ltjin-test1/_mapping
{
  "ltjin-test1": {
    "mappings": {
      "properties": {
        "country": {
          "type": "text"
        },
        "month": {
          "type": "integer"
        },
        "name": {
          "type": "keyword"
        },
        "salary": {
          "type": "integer"
        },
        "test": {
          "properties": {
            "occupation": {
              "type": "text"
            }
          }
        },
        "year": {
          "type": "integer"
        }
      }
    }
  }
}
```

The correct document content should be:
```
POST _bulk
{"index":{"_index" : "ltjin-test1","_id":"1"}}
{"name":"Jake","test": {"occupation": "Engineer"},"country":"England","salary":100000,"year":2023,"month":4}
```

qianheng-aws:
[...]

PR #3559, opened 2025-04-17
add json functions

xinyual:
### Description
Here we add json related functions to align with spark

Here is the argument and description of functions
Functions | argument | description | return type | implementation
-- | -- | -- | -- | --
json | json(value: ANY) | return value if value is a json, otherwise, return null, e.g. json("123") = "123", json(json_object("1", "2")) = {"1": "2"}, json(1) = null | the first argument type | implement by ourselves
json_valid | json_valid(value: ANY) | return true if value is a json, otherwise, return false, e.g. json("123") = true, json(json_object("1", "2")) = true, json(1) = false | boolean | use calcite  SqlStdOperatorTable.IS_JSON_VALUE
json_object | json_object(key1: String, value1: ANY, key2: string, value2 ....)  | create an object using key value pairs, the key must be string | string | use calcite SqlStdOperatorTable.JSON_OBJECT
json_array | json_array(value1: ANY, value2: ANY, ...)  | create an array using values | string | use SqlStdOperatorTable.JSON_ARRAY
json_array_length (this align with spark since splunk doesn't have this function) | json_array_length(value: STRING) | parse the string to json array and return size, if can't be transferred, return null | integer | implement by ourselves
json_extract | json_extract(target: ANY, path1: STRING,  path2:STRING...)  | it first transfer target to json, then extract value using paths. If only one path, return the value, otherwise, return the list of values. If one path cannot find value, return null as the result for this path. The path use "{<index>}" to represent index for array, for e.g., demo = {"a": [{"b": 1}, {"b": 2}]}, json_extract(demo, "a{0}.b") = 1. "{}" means "{*}" | if one path, we return string, otherwise, return list of string | use calcite JsonFunctions::json_value, modify path to align with splunk
json_delete | json_delete(target: ANY, path1: STRING,  path2:STRING...)  | it first transfer target to json, then delete value using paths. return the object after deleting. If a path cannot find any value, skip it. | string. align with splunk | use calcite JsonFunctions::json_remove, modify path to align with splunk
[...]

PR #3577, opened 2025-04-23
Support parsing documents with flattened value

qianheng-aws:
### Description
It supports parsing flattened document value by populating the ExprTupleValue recursively, so we can access the nested fields as well as the root fields as before.

### Related Issues
Resolves  https://github.com/opensearch-project/sql/issues/3477

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/data/model/ExprTupleValue.java
+++ b/core/src/main/java/org/opensearch/sql/data/model/ExprTupleValue.java
@@ -28,6 +28,11 @@ public class ExprTupleValue extends AbstractExprValue {
     return new ExprTupleValue(linkedHashMap);
   }
 
+  public static ExprTupleValue empty() {
+    LinkedHashMap<String, ExprValue> linkedHashMap = new LinkedHashMap<>();
+    return new ExprTupleValue(linkedHashMap);
+  }
+
   @Override
   public Object value() {
     LinkedHashMap<String, Object> resultMap = new LinkedHashMap<>();
@@ -106,5 +111,18 @@ public class ExprTupleValue extends AbstractExprValue {
   @Override
   public int hashCode() {
     return Objects.hashCode(valueMap);
+  }
+
+  /** Implements mergeTo by merging deeply */
+  @Override
+  public ExprTupleValue mergeTo(ExprValue base) {
+    if (base instanceof ExprTupleValue) {
+      base.tupleValue()
+          .forEach((key, value) -> this.tupleValue().merge(key, value, ExprValue::mergeTo));
+    } else {
+      throw new IllegalArgumentException(
+          String.format("Cannot merge ExprTupleValue to %s", base.getClass().getSimpleName()));
+    }
+    return this;
   }
 }
--- a/core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
+++ b/core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
@@ -170,6 +170,9 @@ public interface ExprValue extends Serializable, Comparable<ExprValue> {
   /**
    * Merge the value to the base value. By default, it overrides the base value with the current
    *
+   * <p>This method will be called when key conflict happens in the process of populating
+   * ExprTupleValue See {@link OpenSearchExprValueFactory::populateValueRecursive}.
+   *
    * @param base the target value to merge
    * @return The merged value
    */
--- a/core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
+++ b/core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
@@ -166,4 +166,14 @@ public interface ExprValue extends Serializable, Comparable<ExprValue> {
   default ExprValue keyValue(String key) {
     return ExprMissingValue.of();
   }
+
+  /**
+   * Merge the value to the base value. By default, it overrides the base value with the current
+   *
+   * @param base the target value to merge
+   * @return The merged value
+   */
+  default ExprValue mergeTo(ExprValue base) {
+    return this;
+  }
 }
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/FlattenDocValueIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/FlattenDocValueIT.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.sql.ppl;
 
+import static org.opensearch.sql.legacy.SQLIntegTestCase.Index.FLATTENED_VALUE;
+import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_FLATTENED_VALUE;
 import static org.opensearch.sql.util.MatcherUtils.rows;
 import static org.opensearch.sql.util.MatcherUtils.schema;
 import static org.opensearch.sql.util.MatcherUtils.verifyDataRows;
@@ -20,29 +22,12 @@ public class FlattenDocValueIT extends PPLIntegTestCase {
   @Override
   public void init() throws Exception {
     super.init();
-    putDocument(
-        "test",
-        1,
-        "{\"log\": { \"json\" : { \"status\": \"SUCCESS\", \"time\": 100} } }"); // non-flatten
-    putDocument(
-        "test", 2, "{\"log.json\": { \"status\": \"SUCCESS\", \"time\": 100} }"); // partly-flatten
-    putDocument(
-        "test", 3, "{\"log.json.status\":  \"SUCCESS\", \"log.json.time\": 100 }"); // fully-flatten
-    putDocument(
-        "test",
-        4,
-        "{\"log.json\": { \"status\": \"SUCCESS\" }, \"log.json.time\": 100 }"); // partly-flatten +
-    // fully-flatten
-    putDocument(
-        "test",
-        7,
-        "{\"log\": { \"json\" : {} }, \"log.json\": { \"status\": \"SUCCESS\" }, \"log.json.time\":"
-            + " 100 }"); // all modes in one
+    loadIndex(FLATTENED_VALUE);
   }
 
   @Test
   public void testFlattenDocValue() throws IOException {
-    JSONObject result = executeQuery("source=test");
+    JSONObject result = executeQuery(String.format("source=%s", TEST_INDEX_FLATTENED_VALUE));
     verifySchema(result, schema("log", "struct"));
     TypeSafeMatcher<JSONArray> expectedRow =
         rows(new JSONObject("{ \"json\" : { \"status\": \"SUCCESS\", \"time\": 100} }"));
@@ -52,7 +37,10 @@ public class FlattenDocValueIT extends PPLIntegTestCase {
   @Test
   public void testFlattenDocValueWithFields() throws IOException {
     JSONObject result =
-        executeQuery("source=test | fields log, log.json, log.json.status, log.json.time");
+        executeQuery(
+            String.format(
+                "source=%s | fields log, log.json, log.json.status, log.json.time",
+                TEST_INDEX_FLATTENED_VALUE));
     verifySchema(
         result,
         schema("log", "struct"),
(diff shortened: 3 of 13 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -317,19 +316,63 @@ private static ExprValue createOpenSearchDateType(Content value, ExprType type)
...
+
+  /**
+   * Populate the current ExprTupleValue recursively.
+   *
+   * <p>If JsonPath is not a root path(i.e. has dot in its raw path), it needs update to its
+   * children recursively until the leaf node.
+   *
+   * <p>If there is existing vale for the JsonPath, we need to merge the new value to the old.
+   */
+  private void populateValueRecursive(ExprTupleValue result, JsonPath path, ExprValue value) {
```
penghuo, line 343: add UT

integ-test/src/test/java/org/opensearch/sql/ppl/FlattenDocValueIT.java
```diff
@@ -0,0 +1,70 @@
...
+import org.hamcrest.TypeSafeMatcher;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+
+public class FlattenDocValueIT extends PPLIntegTestCase {
+  @Override
+  public void init() throws Exception {
+    super.init();
+    putDocument(
```
penghuo, line 23: yamlTest is more readable?

integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
```diff
@@ -172,4 +172,10 @@ public static void withFallbackEnabled(Runnable f, String msg) throws IOExceptio
       }
     }
   }
+
+  protected void putDocument(String index, int id, String json) throws IOException {
```
penghuo, line 176: yamlTest support index any doc.

core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
```diff
@@ -166,4 +166,14 @@ default List<ExprValue> collectionValue() {
...
     return ExprMissingValue.of();
   }
+
+  /**
+   * Merge the value to the base value. By default, it overrides the base value with the current
+   *
+   * @param base the target value to merge
+   * @return The merged value
+   */
+  default ExprValue mergeTo(ExprValue base) {
```
penghuo, line 176: what if ExprIpValue mergeTo ExprIntValue?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -317,19 +316,63 @@ private static ExprValue createOpenSearchDateType(Content value, ExprType type)
...
+    if (path.getPaths().size() == 1) {
+      // Update the current ExprValue by using mergeTo if exists
+      result
+          .tupleValue()
+          .computeIfPresent(path.getRootPath(), (key, oldValue) -> value.mergeTo(oldValue));
```
penghuo, line 348: what case is coverred by mergeTo, same field with different value?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -317,19 +316,63 @@ private static ExprValue createOpenSearchDateType(Content value, ExprType type)
...
+      result.tupleValue().putIfAbsent(path.getRootPath(), value);
+    } else {
+      result.tupleValue().putIfAbsent(path.getRootPath(), ExprTupleValue.empty());
+      populateValueRecursive(
+          (ExprTupleValue) result.tupleValue().get(path.getRootPath()), path.getChildPath(), value);
```
penghuo, line 353: result.tupleValue().get(path.getRootPath()) could be null value?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -317,19 +316,63 @@ private static ExprValue createOpenSearchDateType(Content value, ExprType type)
...
+    }
+  }
+
+  @Getter
+  private static class JsonPath {
```
LantaoJin, line 358: can this be used by [json-path](https://github.com/opensearch-project/sql/pull/3559/files#diff-8661f4ba56dd608a0331a1b951fcc910c80ddd53de7dec30cdf6c0e773cab6d0R66) lib?

integ-test/src/test/java/org/opensearch/sql/ppl/FlattenDocValueIT.java
qianheng-aws: @penghuo If we want to test this case w/wo Calcite enabled, in yamlTest, is there a convenient way to implement this with less code duplication? 

In Java code, we can use inheritance to make it simple like the current code. But in yaml, we may need to duplicate the process with extra settings.

 I've thought of using yamlTest but didn't have much information whether it can do such thing conveniently, so I keep using IT to do the verification.

core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
qianheng-aws: It should override with the current implementation. Such case will only happen when there is duplication value for a field.

1. In normal cases, only ExprTupleValue::mergeTo should be invoked. Cases like:
```
{"log.json.time": 100, "log.json": { "status": "SUCCESS" }}
```
For `"log.json.time": 100`, it will be parsed into `path: log, value: ExprTupleValue(Map.of({\"json\": ExprTupleValue(Map.of({\"time\": 100}))}))`

For `"log.json": { "status": "SUCCESS" }`, it will be parsed into `path: log, value: ExprTupleValue(Map.of({\"json\": ExprTupleValue(Map.of({\"status\": SUCCESS}))}))`

So after putting the first field, there is key conflict when putting the second field. In this case, we need to invoke ExprTupleValue::mergeTo` to merge 2 ExprTupleValue deeply.

2. In some edge cases where there is conflict value in customer's document, like:
```
{"log.json.status": "FAILED", "log.json": { "status": "SUCCESS" }}
```
There is value conflict for the field `log.json.status`, so I want the value added most recently to override any previously added values, and `ExprValue::megeTo` will be invoked. 

But, in json, since the sequence of the fields in the same level are unstable actually, we cannot assert which value will be put firstly. Maybe throw exception to tell customer there is incompatible value conflict will also be an option. What's your opinion on this edge case? @LantaoJin @penghuo

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
qianheng-aws: No, it won't. I have add putIfAbsent operation in the code 1 line above.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
qianheng-aws: See the comment [[here](https://github.com/opensearch-project/sql/pull/3577#discussion_r2057374631)](https://github.com/opensearch-project/sql/pull/3577#discussion_r2057374631)

core/src/main/java/org/opensearch/sql/data/model/ExprTupleValue.java
```diff
@@ -107,4 +112,17 @@ public int compare(ExprValue other) {
   public int hashCode() {
     return Objects.hashCode(valueMap);
   }
+
+  /** Implements mergeTo by merging deeply */
+  @Override
+  public ExprTupleValue mergeTo(ExprValue base) {
```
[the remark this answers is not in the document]
qianheng-aws: Done

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
qianheng-aws: Done

integ-test/src/test/java/org/opensearch/sql/ppl/FlattenDocValueIT.java
qianheng-aws: Added a yaml test with only Calcite enabled.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
qianheng-aws: Not really. That lib requires input json path string starts with `$.` and it's not convenient as this self-defined class because it's more strict defined for a real json path. And its class `Path` doesn't have API to get it's child's path.

core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
LantaoJin: > Maybe throw exception to tell customer there is incompatible value conflict will also be an option. 

Do you mean throw exception in ppl querying or data ingesting? Data ingesting seems out of the scope here.

LantaoJin: IMO, if there are conflict fields. Display the any of them would be all fine as long as the we always populating out one solid field. For example:
```
{"id": 1, "log.json.status": "FAILED", "log.json": { "status": "SUCCESS" }}
{"id": 2, "log.json.status": "FAILED", "log.json": { "status": "SUCCESS" }}
{"id": 3, "log.json.status": "FAILED", "log.json": { "status": "SUCCESS" }}
```
PPL `source=t | fields id, log.json.status` return 
```
1, FAILED
2, FAILED
3, FAILED
```
or 
```
1, SUCCESS
2, SUCCESS
3, SUCCESS
```
are all acceptable.

LantaoJin: Could you add a test for key conflict case?

integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
LantaoJin: can this be deleted now?

core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
qianheng-aws: The problem is key conflict case will flakey because fields order are not ensured in json. We cannot assert which value will be put in the end.

qianheng-aws: Added in the UT of `testPopulateValueRecursive`, where I can control the populate order

LantaoJin (approved):

LantaoJin:
@qianheng-aws could you change this log level from `warn` to `debug` in this PR? This PR will be merge to 3.0.0 GA
This warning log will write out thousands of unnecessary logs when the data is a nested type.

LantaoJin (approved):

LantaoJin:
ping @penghuo @dai-chen

core/src/main/java/org/opensearch/sql/data/model/ExprValue.java
penghuo: > Maybe throw exception to tell customer there is incompatible value conflict will also be an option. What's your opinion on this edge case?

do not throw exception, current approach looks good.  OpenSearch also return the LATEST field.

```
###
POST {{baseUrl}}/test00001/_doc

{
    "log.json.status": "failed1",
    "log.json": {
      "status": "failed2"
    },
    "log": {
      "json": {
        "status": "failed3"
      }
    },
}

###
GET {{baseUrl}}/test00001/_search

{
    "fields": [
        "log.json.status"
    ],
    "_source": false
}

### Return 
        "fields": {
          "log.json.status": [
            "failed3"
          ]
        }
```

penghuo (approved):

Issue #3575, opened 2025-04-23
[FEATURE] Support Array related function with lambda expression


xinyual:
**Is your feature request related to a problem?**
We need to align with Spark on array related functions, and support lambda expression in PPL
https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/functions/ppl-collection.md

PR #3584, opened 2025-04-27
Add lambda function and array related functions

xinyual:
### Description
This pr adds lambda function and array related functions. Calcite don't have array related functions so we need to implement by ourselves.
Now the logic for lambda is:
We will consider lambda function as a new PPL expression and parse it regularly to construct rexnode. To get return type for lambda expression, we need to firstly map the argument type in the calciteContext. For example, forall(array(1, 2, 3), x -> x > 0), then x -> INTEGER. 
We also have an exception for reduce because the acc is the dynamic type.
The calcite/lin4j generate code according to the input type. For example, reduce(array(1.0,2.0 ,3.0), 0, (acc, x) -> acc + x). Ideally, we should map acc -> INTEGER, x -> DOUBLE. But if we map through this, the code of + would be `plus(INTERGER acc, DOUBLE x)`, then after first apply, the acc would be double, then it will throw exception. Thus, we apply ANY to the acc and infer the return type in `getReturnTypeInference`

The function is aligned with https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/functions/ppl-collection.md

TODO: nested object is not supported in lambda currently. It will be automatically supported when we support this. E.g. x -> x.a > 0

For detailed implementation and description:

Functions | argument | description | return type | implementation
-- | -- | -- | -- | --
ARRAY | ARRAY(value1: ANY, value2:ANY, ...) | create an array with input values. Currently we don't allow mixture types. We will infer a least restricted type, for example array(1, "demo") -> ["1", "demo"] | ARRAY | wrap `SqlLibraryOperators.ARRAY`
ARRAY_LENGTH | ARRAY_LENGTH(value: ARRAY) | return array length | integer | `SqlLibraryOperators.ARRAY_LENGTH`
FORALL | forall(value:ARRAY, function: LAMBDA) | check whether all element inside array can meet the lambda function. The function should also return boolean. | boolean | implement by ourselves since we cannot find matched built-in calcite one.
EXISTS | exists(value:ARRAY, function: LAMBDA) | check whether existing one of element inside array can meet the lambda function. The function should also return boolean. | boolean | implement by ourselves since we cannot find matched built-in calcite one.
FILTER | filter(value:ARRAY, function: LAMBDA) | filter the element in the array by the lambda function. The function should return boolean  | array | implement by ourselves since we cannot find matched built-in calcite one.
TRANSFORM | transform(value:ARRAY, function: LAMBDA) | transform the element of array one by one using lambda. Transform can accept one more argument like (x, i) -> x + i, where i is the index of element in array.  | array | implement by ourselves since we cannot find matched built-in calcite one.
REDUCE | reduce(value:ARRAY, base_value:ANY, acc_function: LAMBDA)/reduce(value:ARRAY, base_value:ANY, acc_function: LAMBDA, reduce_function:LAMBDA) | The function will first use acc_function to go through all element and return value to the acc. Then apply reduce function to the acc if exists. The acc_function's lambda format is (acc,x) -> ..., the reduce_function format is (acc) -> ...   | ANY, according to the lambda function | implement by ourselves since we cannot find matched built-in calcite one.

### Related Issues
https://github.com/opensearch-project/sql/issues/3575

### Check List


core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
```diff
@@ -58,6 +58,14 @@ public enum BuiltinFunctionName {
   TAN(FunctionName.of("tan")),
   SPAN(FunctionName.of("span")),
 
+  /** Collection functions */
+  ARRAY(FunctionName.of("array")),
```
LantaoJin, line 62: do you miss `ARRAY_LENGHT`? https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/functions/ppl-collection.md#array

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
```diff
@@ -0,0 +1,81 @@
...
+                .mapToObj(i -> ((Number) args[i]).doubleValue())
+                .collect(Collectors.toList());
+
+        return unboxed;
+      case FLOAT:
+        List<Object> unboxedFloat =
+            IntStream.range(0, args.length - 1)
+                .mapToObj(i -> ((Number) args[i]).floatValue())
+                .collect(Collectors.toList());
+        return unboxedFloat;
```
LantaoJin, line 76: could you explain why this special logic needed?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.calcite.linq4j.tree.Expressions;
+import org.apache.calcite.linq4j.tree.Types;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+
+public class ArrayFunctionImpl extends ImplementorUDF {
```
LantaoJin, line 28: can't we reuse `SqlLibraryOperators.ARRAY`? Again, please add a reason in PR description for any new added function why it must implement by ourselves.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ExistsFunctionImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.calcite.adapter.enumerable.RexToLixTranslator;
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.linq4j.tree.Types;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.schema.impl.ScalarFunctionImpl;
+import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+
+public class ExistsFunctionImpl extends ImplementorUDF {
```
LantaoJin, line 21: can't we reuse `SqlLibraryOperators.ARRAY_CONTAINS`? please check all `SqlLibraryOperators.ARRAY_*` first.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
xinyual: We need to internally convert it. Otherwise, the calcite will directly cast like DOUBLE to INTEGER, which will raise exception.

core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
xinyual: Already add it.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -331,16 +344,132 @@ private boolean isTimeBased(SpanUnit unit) {
...
+        RexNode lambdaNode = analyze(arg, lambdaContext);
+        if (node.getFuncName().equalsIgnoreCase("reduce")) { // analyze again with calculate type
+          lambdaContext =
+              prepareLambdaContext(
+                  context,
+                  (LambdaFunction) arg,
+                  arguments,
+                  node.getFuncName(),
+                  lambdaNode.getType());
+          lambdaNode = analyze(arg, lambdaContext);
```
LantaoJin, line 477: why analyze `reduce` twice?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
```diff
@@ -0,0 +1,99 @@
...
+
+// TODO: Support array of mixture types.
+public class ArrayFunctionImpl extends ImplementorUDF {
+  public ArrayFunctionImpl() {
+    super(new ArrayImplementor(), NullPolicy.ANY);
+  }
+
+  /**
+   * @return We wrap it here to accept null since the original return type inference will generate
+   *     non-nullable type
```
LantaoJin, line 43: Does Spark array accept null either? Why we do this wrap?

docs/user/ppl/functions/collection.rst
```diff
@@ -0,0 +1,201 @@
...
+
+ARRAY
+-----
+
+Description
+>>>>>>>>>>>
+
+Version: 3.1.0
+
+Usage: ``array(value1, value2, value3...)`` create an array with input values. Currently we don't allow mixture types. We will infer a least restricted type, for example ``array(1, "demo")`` -> ["1", "demo"]
```
LantaoJin, line 19: Question: what is the reason to support `infer a least restricted type` instead of throwing exception?

LantaoJin: please add a `Limitation: ` after each `Usage: ` to explain these functions only work with `plugins.calcite.enabled=true`

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
xinyual: Reduce is very special case since it will sometimes change the type of accumulator. For example, reduce([1.0, 2.0], 0, (acc, x) -> acc + x, acc -> acc * 10). Here the lambda (acc, x) -> acc + x firstly will find (integer, double) and then during the iteration, find (double, double). Current solution is we will first analyze once and find the return type is double, then use double as the expected input and cast the initial value of acc to the expected type.

docs/user/ppl/functions/collection.rst
xinyual: > Question: what is the reason to support `infer a least restricted type` instead of throwing exception?

This is aligned with SPARK.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
xinyual: Yes. Spark array accept null.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -337,16 +349,112 @@ private boolean isTimeBased(SpanUnit unit) {
...
     RexNode expr = analyze(node.getExpression(), context);
     return context.relBuilder.alias(expr, node.getVar().getField().toString());
   }
 
+  /**
+   * The function will clone a context for lambda function. For lambda like (x, y, z) -> ..., we
+   * will map type for each lambda argument by the order of previous argument. Also, the function
+   * will add these variables to the context so they can pass visitQualifiedName
+   */
+  private CalcitePlanContext prepareLambdaContext(
```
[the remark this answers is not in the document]
xinyual: Done.

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
```diff
@@ -398,7 +398,9 @@ private static List<ExprType> getExprTypes(SqlTypeFamily family) {
             // Integer is mapped to BIGINT in family.getDefaultConcreteType
           case INTEGER -> List.of(
               OpenSearchTypeFactory.TYPE_FACTORY.createSqlType(SqlTypeName.INTEGER));
-          case ANY, IGNORE -> List.of(
+            // case DATETIME_INTERVAL ->
+            // SqlTypeName.INTERVAL_TYPES.stream().map(OpenSearchTypeFactory.TYPE_FACTORY::createSqlIntervalType).toList();
```
LantaoJin: how DATETIME_INTERVAL impact `reduce`?

LantaoJin (dismissed):

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
xinyual: It's a useless change, I will revert it.

LantaoJin (dismissed):

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
qianheng-aws: Does it necessary to detect a non-any type in analyzing phase? What if the input list has type of ARRAY<ANY>?

qianheng-aws: And does it make sense to infer the return type by using `leastRestrictive(arg0. getComponentType(), arg1.getType()` instead of analyzing twice.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.sql.fun.SqlLibraryOperators;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+import org.opensearch.sql.expression.function.UDFOperandMetadata;
+
+// TODO: Support array of mixture types.
+public class ArrayFunctionImpl extends ImplementorUDF {
```
qianheng-aws, line 36: I would be great to add description and a simple example for each function. It should only show the functionality of this function and as simple as possible. For example, `array(1, 2, 3) -> [1, 2, 3]` would be enough for array function.

It will improve the code readability for developer,  different from the doc for customer. You can do it later in another PR

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
xinyual: > Does it necessary to detect a non-any type in analyzing phase? What if the input list has type of ARRAY?

ANY would block the implementation in two parts: 1. The UDF part sometimes needs type to choose implementation 2. any would also be blocker for type checker.  For example, we use calcite multiply, which only  support numeric/interval *  numeric/interval, any would throw exception when check the type.

> And does it make sense to infer the return type by using `leastRestrictive(arg0. getComponentType(), arg1.getType()` instead of analyzing twice.

leastRestrictive(arg0. getComponentType(), arg1.getType() doesn't work here. For example, acc=0,  (acc, x) -> acc + length(x) * 1.0 would return double, which means we need to cast acc base value to double. But leastRestrictive(integer, string) wouldn't be double.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
xinyual: Already add for each. Please check.

qianheng-aws (approved):

LantaoJin (approved):

PR #3110, opened 2024-10-23
Add CIDR function to PPL (#3036)

currantw:
### Description

Adds the `cidr` function to PPL, which returns whether an IP address is within an IP address range specified in CIDR notation.

### Related Issues

Resolves #3036.

### Check List
- [X] New functionality includes testing.
- [X] New functionality has been documented.
- [X] New functionality has javadoc added.
- [X] New functionality has a user manual doc added.
- [N/A] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [X] Commits are signed per the DCO using `--signoff`.
- [?] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3603, opened 2025-05-06
Implement `cidrmatch` udf with Calcite

### LINKED ISSUE #3505: [FEATURE][Calcite engine] Support function `cidrmatch`
- cidrmatch

yuancu:
### Description
Implement [`CIDRMATCH`](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/ip.rst#cidrmatch) on Calcite engine with a UDF
 
### Issues Resolved
Resolves #3505 
 
### Check List


core/src/main/java/org/opensearch/sql/expression/function/udf/ip/CidrMatchFunction.java
```diff
@@ -0,0 +1,59 @@
...
+import org.opensearch.sql.expression.function.ImplementorUDF;
+import org.opensearch.sql.expression.ip.IPFunctions;
+
+/**
+ * {@code cidrmatch(ip, cidr)} checks if ip is within the specified cidr range.
+ *
+ * <p>Signature:
+ *
+ * <ul>
+ *   <li>(STRING, STRING) -> BOOLEAN
```
dai-chen, line 29: Is function signature (String,String) or (IP, String)?

yuancu: It accepts both. The document is updated to reflect this. Thank you for pointing this out.

- In the documentation of [ppl-on-spark](https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/functions/ppl-ip.md#cidrmatch), the function signature is (string, string). Since we have IP as an UDT, I added (ip, string) as another valid signature.

qianheng-aws (approved):

docs/user/ppl/functions/ip.rst
```diff
@@ -16,7 +16,7 @@ Description
 
 Usage: `cidrmatch(ip, cidr)` checks if `ip` is within the specified `cidr` range.
 
-Argument type: STRING, STRING
+Argument type: STRING/IP, STRING
```
LantaoJin, line 19: Is the IP argument supported in v2?

yuancu: I think so. `cidrmatch` is implemented in #3110; IP data type is added in #3145 . It seems they were introduced in v3, but are already supported before introducing calcite. 

According to the implementation at [IPFunctions.java#L36](https://github.com/opensearch-project/sql/blob/f48095849ed9c6b4b3dc92945921475adb12b393/core/src/main/java/org/opensearch/sql/expression/ip/IPFunctions.java#L36), IP is the only supported argument type. Strings may be cast to IP. The documentation may be out of date since the IP's PR comes later that `cidrmatch`'s PR.

LantaoJin (approved):

PR #3228, opened 2025-01-01
PPL command expression implementation for `geoip`

andy-k-improving:
### Description
Introduce a new PPL command expression `geoip`, to perform geo-spatial information lookup with the provided IPv4 || IPv6 addresses, result of the lookup is formatted into a tuple with attribute as key and location detail as value. 

In this particular setting, SQL plugin will act as a thin client, by relaying the IPEnrichment request to OpenSearch Geo-Spatial plugin, WITHIN the same cluster. 
Detail implementation and interface that exposed on Geo-Spatial side can be found:
https://github.com/opensearch-project/geospatial/pull/700

Internally this functionality is achieved by:
 - Adding an no-op `OpenSearchFunctionExpression` marker to identify this is an expression has no default implement on other runtime (Ex: Prometheus)
 - Update `OpenSearchIndex` in order to provide an OpenSearch specific handler for `eval` operator and its expressions, when OS being used as the storage engine. 

During runtime, all eval expressions, will being passed to `OpenSearchIndex.visitEval( )`, then `OpenSearchEvalOperator` class will pick up the call, by evaluating all `eval` expression as it is, and then handle all occasion of `OpenSearchFunctionExpression` separately, by reading the function name and argument, and execute the appropriate business logic. 

Marker class `OpenSearchFunctionExpression` is being used in this case because the actual implementation require runtime OpenSearch client connectivity, however `core` module is mean to be generic, hence this workaround is being deployed, by tagging it as `OpenSearchFunctionExpression` on `core` and only handle it on the `opensearch` Cradle module .

### Related Issues
Resolves: https://github.com/opensearch-project/sql/issues/3037

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
 - [x] New functionality has javadoc added.
 - [x] New functionality has a user manual doc added.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3604, opened 2025-05-06
Implement `geoip` UDF

### LINKED ISSUE #3506: [FEATURE][Calcite engine] Support function geoip
Support function geoip

yuancu:
### Description
Implement [`GEOIP`](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/ip.rst#geoip) on Calcite engine with a UDF.

### Related Issues
Resolves #3506 

### Check List

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -63,6 +63,8 @@ dependencies {
     api 'org.apache.calcite:calcite-linq4j:1.38.0'
     api project(':common')
     implementation "com.github.seancfoley:ipaddress:5.4.2"
+    implementation group: 'org.opensearch', name:'geospatial-client', version: "${opensearch_build}"
+    implementation group: 'org.opensearch.client', name: 'opensearch-rest-high-level-client', version: "${opensearch_version}"
 
     annotationProcessor('org.immutables:value:2.8.8')
     compileOnly('org.immutables:value-annotations:2.8.8')
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -65,6 +65,7 @@ import org.opensearch.sql.exception.CalciteUnsupportedException;
 import org.opensearch.sql.exception.SemanticCheckException;
 import org.opensearch.sql.expression.function.BuiltinFunctionName;
 import org.opensearch.sql.expression.function.PPLFuncImpTable;
+import org.opensearch.sql.expression.function.udf.ip.GeoIpFunction;
 
 @RequiredArgsConstructor
 public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalcitePlanContext> {
@@ -347,6 +348,14 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
   public RexNode visitFunction(Function node, CalcitePlanContext context) {
     List<RexNode> arguments =
         node.getFuncArgs().stream().map(arg -> analyze(arg, context)).collect(Collectors.toList());
+
+    // GEOIP needs NodeClient to perform RPC calls. Therefore, we handle it separately.
+    if (BuiltinFunctionName.GEOIP.equals(BuiltinFunctionName.of(node.getFuncName()).orElse(null))) {
+      PPLFuncImpTable.FunctionImp geoIpImpl =
+          (builder, args) ->
+              builder.makeCall(new GeoIpFunction(context.getClient()).toUDF("GEOIP"), args);
+      return geoIpImpl.resolve(context.rexBuilder, arguments.toArray(new RexNode[0]));
+    }
     RexNode resolvedNode =
         PPLFuncImpTable.INSTANCE.resolveSafe(
             context.rexBuilder, node.getFuncName(), arguments.toArray(new RexNode[0]));
--- a/core/src/main/java/org/opensearch/sql/datasource/DataSourceService.java
+++ b/core/src/main/java/org/opensearch/sql/datasource/DataSourceService.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import java.util.Set;
 import org.opensearch.sql.datasource.model.DataSource;
 import org.opensearch.sql.datasource.model.DataSourceMetadata;
+import org.opensearch.transport.client.node.NodeClient;
 
 /** DataSource Service manage {@link DataSource}. */
 public interface DataSourceService {
@@ -87,4 +88,13 @@ public interface DataSourceService {
    */
   DataSourceMetadata verifyDataSourceAccessAndGetRawMetadata(
       String dataSourceName, RequestContext context);
+
+  /**
+   * Returns a NodeClient instance for executing requests against the OpenSearch cluster.
+   *
+   * @return NodeClient instance or throw an exception if not implemented.
+   */
+  default NodeClient getNodeClient() {
+    throw new UnsupportedOperationException("Not implemented");
+  }
 }
--- a/core/src/main/java/org/opensearch/sql/expression/function/udf/ip/GeoIpFunction.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/udf/ip/GeoIpFunction.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.expression.function.udf.ip;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import lombok.Getter;
+import org.apache.calcite.adapter.enumerable.NotNullImplementor;
+import org.apache.calcite.adapter.enumerable.NullPolicy;
+import org.apache.calcite.adapter.enumerable.RexToLixTranslator;
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.linq4j.tree.Expressions;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.opensearch.geospatial.action.IpEnrichmentActionClient;
+import org.opensearch.sql.common.utils.StringUtils;
+import org.opensearch.sql.data.model.ExprStringValue;
+import org.opensearch.sql.data.model.ExprTupleValue;
+import org.opensearch.sql.data.model.ExprValue;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+import org.opensearch.transport.client.node.NodeClient;
+
+public class GeoIpFunction extends ImplementorUDF {
+  public GeoIpFunction(NodeClient nodeClient) {
+    super(new GeoIPImplementor(nodeClient), NullPolicy.ANY);
+  }
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return op -> {
+      RelDataTypeFactory typeFactory = op.getTypeFactory();
+      RelDataType varcharType = typeFactory.createSqlType(SqlTypeName.VARCHAR);
+      RelDataType anyType = typeFactory.createSqlType(SqlTypeName.ANY);
+      return typeFactory.createMapType(varcharType, anyType);
+    };
+  }
+
+  public static class GeoIPImplementor implements NotNullImplementor {
+    @Getter private static NodeClient nodeClient;
+
+    public GeoIPImplementor(NodeClient nodeClient) {
+      GeoIPImplementor.nodeClient = nodeClient;
+    }
+
+    @Override
+    public Expression implement(
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+      if (getNodeClient() == null) {
+        throw new IllegalStateException("nodeClient is null.");
+      }
+      List<Expression> operandsWithClient = new ArrayList<>(translatedOperands);
+      // Since a NodeClient cannot be passed as a parameter using Expressions.constant,
+      // it is instead provided through a function call.
+      operandsWithClient.add(Expressions.call(GeoIPImplementor.class, "getNodeClient"));
+      return Expressions.call(GeoIPImplementor.class, "fetchIpEnrichment", operandsWithClient);
+    }
+
+    public static Object fetchIpEnrichment(
+        String dataSource, String ipAddress, NodeClient nodeClient) {
+      return fetchIpEnrichment(dataSource, ipAddress, "", nodeClient);
+    }
+
+    public static Object fetchIpEnrichment(
+        String dataSource, String ipAddress, String commaSeparatedOptions, NodeClient nodeClient) {
+      IpEnrichmentActionClient ipClient = new IpEnrichmentActionClient(nodeClient);
+      dataSource = StringUtils.unquoteText(dataSource);
+      String option = StringUtils.unquoteText(commaSeparatedOptions);
+      // Convert the option into a set.
+      final Set<String> options = new HashSet<>();
+      if (!commaSeparatedOptions.isEmpty()) {
+        options.addAll(
+            Arrays.stream(option.split(",")).map(String::trim).collect(Collectors.toSet()));
+      }
+
+      try {
+        Map<String, Object> geoLocationData = ipClient.getGeoLocationData(ipAddress, dataSource);
+        Map<String, ExprValue> enrichmentResult =
+            geoLocationData.entrySet().stream()
+                .filter(entry -> options.isEmpty() || options.contains(entry.getKey()))
+                .collect(
+                    Collectors.toMap(
+                        Map.Entry::getKey, v -> new ExprStringValue(v.getValue().toString())));
+        return ExprTupleValue.fromExprValueMap(enrichmentResult).valueForCalcite();
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+    }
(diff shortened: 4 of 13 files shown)


core/src/main/java/org/opensearch/sql/expression/function/udf/ip/GeoIpFunction.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public Expression implement(
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+      if (getNodeClient() == null) {
+        throw new IllegalStateException("nodeClient is null.");
+      }
+      List<Expression> operandsWithClient = new ArrayList<>(translatedOperands);
+      // Since a NodeClient cannot be passed as a parameter using Expressions.constant,
+      // it is instead provided through a function call.
+      operandsWithClient.add(Expressions.call(GeoIPImplementor.class, "getNodeClient"));
```
yuancu, line 72: I cannot find a way to pass a `NodeClient` instance to `fetchIpEnrichment` as an `Expression`. Therefore, I created a static member variable `nodeClient` and pass it to `fetchIpEnrichment` with a static function.

This is not a standard way to pass variables. Please feel free to suggest if you have a better solution.

ppl/build.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
     api project(':common')
     api project(':core')
     api project(':protocol')
+    implementation group: 'org.opensearch.client', name: 'opensearch-rest-high-level-client', version: "${opensearch_version}"
```
LantaoJin, line 57: can this change be removed?

core/src/main/java/org/opensearch/sql/datasource/DataSourceService.java
```diff
@@ -87,4 +88,13 @@ public interface DataSourceService {
...
       String dataSourceName, RequestContext context);
+
+  /**
+   * Returns a NodeClient instance for executing requests against the OpenSearch cluster.
+   *
+   * @return NodeClient instance or throw an exception if not implemented.
+   */
+  default NodeClient getNodeClient() {
+    throw new UnsupportedOperationException("Not implemented");
+  }
```
LantaoJin, line 99: I think we'd better not to add client to DataSourceSerivce. Instead, pass `NodeClient` to `queryPlanFactory()` as an parameter.

core/build.gradle
```diff
@@ -63,6 +63,8 @@ dependencies {
     api 'org.apache.calcite:calcite-linq4j:1.38.0'
     api project(':common')
     implementation "com.github.seancfoley:ipaddress:5.4.2"
+    implementation group: 'org.opensearch', name:'geospatial-client', version: "${opensearch_build}"
+    implementation group: 'org.opensearch.client', name: 'opensearch-rest-high-level-client', version: "${opensearch_version}"
```
dai-chen, line 67: Could you elaborate this? Although UDT/UDF regarding IP/geopoint can be considered as part of PPL spec, it doesn't mean all engines need to support it, right? With this change, I think `core` module starts coupling with specific storage which is against initial design.

core/src/main/java/org/opensearch/sql/datasource/DataSourceService.java
yuancu: I added this dependency since `CalcitePlanContext.create(FrameworkConfig, QueryType, NodeClient)` in `ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAbstractTest.java` needs to know `NodeClient`.

I removed the dependency by creating a overriding of create: `create(FrameworkConfig config, QueryType queryType)`. In this way, `CalcitePPLAbstractTest` no longer needs to know `NodeClient`.

core/build.gradle
yuancu: I agree that this change in dependency is worrying.

I added the dependency since the implementation of `geoip` is coupled with the `geospatial-client` and opensearch node client: `IpEnrichmentActionClient` and `NodeClient` are necessary to communicate with the geospatial plugin. This is also the same in `geoip`'s original implementation in #3228.

I have not figured out a way to workaround this. Please feel free to comment if you have any suggestions.

core/src/main/java/org/opensearch/sql/datasource/DataSourceService.java
yuancu: Done

ppl/build.gradle
yuancu: I removed this by creating an overriding of `CalcitePlanContext.create`, so that `CalcitePPLAbstractTest` does not need to know `NodeClient` when creating a `CalcitePlanContext` instance.

core/build.gradle
dai-chen: I see. I feel probably we should only maintain function spec in PPL-Calcite (as part of PPL spec) and each adaptor registers function implementation if they can support?

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/GeoIpFunction.java
qianheng-aws: Will `Expressions.constant` work?

qianheng-aws: Will `Expressions.constant` work?

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/GeoIpFunction.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public static Object fetchIpEnrichment(
+        String dataSource, String ipAddress, String commaSeparatedOptions, NodeClient nodeClient) {
+      IpEnrichmentActionClient ipClient = new IpEnrichmentActionClient(nodeClient);
+      dataSource = StringUtils.unquoteText(dataSource);
+      String option = StringUtils.unquoteText(commaSeparatedOptions);
+      // Convert the option into a set.
+      final Set<String> options = new HashSet<>();
+      if (!commaSeparatedOptions.isEmpty()) {
```
qianheng-aws, line 88: To avoid add if on a constant value in plan, you could extract the main logic into a private function, and 2 functions of different signatures has its own handling logic on `option`.

core/build.gradle
yuancu: Indeed, if we are still going to support different data source engines, it would be better to keep core from coupling with opensearch-specific implementations. @LantaoJin @penghuo What are your opinions on this?

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/GeoIpFunction.java
yuancu: @qianheng-aws Neither `Expressions.constant` nor `Expressions.variable` work in this case.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -347,6 +348,14 @@ public RexNode visitLet(Let node, CalcitePlanContext context) {
   public RexNode visitFunction(Function node, CalcitePlanContext context) {
     List<RexNode> arguments =
         node.getFuncArgs().stream().map(arg -> analyze(arg, context)).collect(Collectors.toList());
+
+    // GEOIP needs NodeClient to perform RPC calls. Therefore, we handle it separately.
+    if (BuiltinFunctionName.GEOIP.equals(BuiltinFunctionName.of(node.getFuncName()).orElse(null))) {
```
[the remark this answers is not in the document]
yuancu: I've only come across one function, `geoip`, that needs `NodeClient`.

Before enhancing `PPLFuncImpTable` with dynamic function registration, I can make GeoIpFunction UDF singleton, only assigning nodeClient to it each time `visitFunction` is invoked.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/GeoIpFunction.java
yuancu: Done

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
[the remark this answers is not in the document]
yuancu: > I've only come across one function, `geoip`, that needs `NodeClient`.
> 
> Before enhancing `PPLFuncImpTable` with dynamic function registration, I can make GeoIpFunction UDF singleton, only assigning nodeClient to it each time `visitFunction` is invoked.

I find it hard to implement this -- I can no longer get the `NotNullImplementor` instance once I create a`SqlUserDefinedFunction` with `implementorUDF.toUDF`.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/GeoIpFunction.java
qianheng-aws: @yuancu What exception will it throw if using `Expressions.constant`?

yuancu:
```java
java.sql.SQLException: Error while preparing plan [LogicalProject(name=[$0], ip=[$1], enrichmentResult=[GEOIP('dummycityindex':VARCHAR, $1, 'city':VARCHAR)])
  CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_geoip]])...
...
Caused by: java.lang.RuntimeException: Error while compiling generated Java code:

public org.apache.calcite.linq4j.Enumerable bind(final org.apache.calcite.DataContext root) {
  final org.opensearch.sql.opensearch.storage.scan.CalciteEnumerableIndexScan v1stashed = (org.opensearch.sql.opensearch.storage.scan.CalciteEnumerableIndexScan) root.get("v1stashed");
  final org.apache.calcite.linq4j.Enumerable _inputEnumerable = v1stashed.scan();
  return new org.apache.calcite.linq4j.AbstractEnumerable(){
      public org.apache.calcite.linq4j.Enumerator enumerator() {
        return new org.apache.calcite.linq4j.Enumerator(){
            public final org.apache.calcite.linq4j.Enumerator inputEnumerator = _inputEnumerable.enumerator();
            public void reset() {
              inputEnumerator.reset();
            }

            public boolean moveNext() {
              return inputEnumerator.moveNext();
            }

            public void close() {
              inputEnumerator.close();
            }

            public Object current() {
              final Object[] current = (Object[]) inputEnumerator.current();
              final String input_value1 = current[1] == null ? null : current[1].toString();
              return new Object[] {
                  current[0],
                  current[1],
                  input_value1 == null ? null : org.opensearch.sql.expression.function.udf.ip.GeoIpFunction.GeoIPImplementor.fetchIpEnrichment("dummycityindex", input_value1, "city", org.opensearch.transport.client.node.NodeClient@52804e28)};
            }

          };
      }

    };
}


public Class getElementType() {
  return java.lang.Object[].class;
}
```

The expression `org.opensearch.sql.expression.function.udf.ip.GeoIpFunction.GeoIPImplementor.fetchIpEnrichment("dummycityindex", input_value1, "city", org.opensearch.transport.client.node.NodeClient@52804e28)};` can not be correctly compiled.

This is because in [`ConstantExpression.write`](https://github.com/apache/calcite/blob/412ac801ebc9484521ec469661c71011ad0ef9b9/linq4j/src/main/java/org/apache/calcite/linq4j/tree/ConstantExpression.java#L86), there isn't a corresponding handling `if` for the `NodeClient` class. A constant of `nodeClient` falls to the default case `writer.append(value)`, which invokes:
```java
    @Override
    public StringBuilder append(Object obj) {
        return append(String.valueOf(obj));
    }
```
Therefore, Expressions.constant(nodeClient) doesn't work.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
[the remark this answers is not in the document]
yuancu: I have only came across with one function, `geoip` that requires NodeClient.

@qianheng-aws I have moved `geoip` implementation to the opensearch package. The function is registered inside `OpenSearchExecutionEngine` during its construction.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -91,13 +91,29 @@ default List<RelDataType> getParams() {
     INSTANCE = new PPLFuncImpTable(builder);
   }
 
-  private final ImmutableMap<BuiltinFunctionName, PairList<CalciteFuncSignature, FunctionImp>> map;
+  private final ImmutableMap<BuiltinFunctionName, PairList<CalciteFuncSignature, FunctionImp>>
```
LantaoJin, line 94: Can we replace `PairList` to `List<Pair>`?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -91,13 +91,29 @@ default List<RelDataType> getParams() {
...
+      functionRegistry;
+  private final Map<BuiltinFunctionName, PairList<CalciteFuncSignature, FunctionImp>>
```
LantaoJin, line 96: can you add some javadocs on these two `functionRegistry`s to explain how to choose them in developing udf.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
yuancu: Technically yes. But I don't see an obvious gain in doing so. Besides, we may have to introduce `javafx` as a new dependency to the core module since `Pair` resides in `javafx.util`. (Although we can also use the alternative in apache commons-lang3)

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
yuancu: Added

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
yuancu: Done

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -91,13 +92,45 @@ default List<RelDataType> getParams() {
     INSTANCE = new PPLFuncImpTable(builder);
   }
 
-  private final ImmutableMap<BuiltinFunctionName, PairList<CalciteFuncSignature, FunctionImp>> map;
```
qianheng-aws, line 94: [non-blocking] what's the concern of changing PairList to List\<Pair\>?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -119,7 +152,12 @@ public RexNode resolve(final RexBuilder builder, final String functionName, RexN
 
   public RexNode resolve(
       final RexBuilder builder, final BuiltinFunctionName functionName, RexNode... args) {
-    final PairList<CalciteFuncSignature, FunctionImp> implementList = map.get(functionName);
+    List<Pair<CalciteFuncSignature, FunctionImp>> implementList =
```
qianheng-aws, line 155: [discussion] It comes to me that it may make more sense to use external function registry prior to core function registry. Then it provides the functionality of overriding core implementation with module's implementation, although we don't have such requirements for now. Or we'd better do this change until we really need it.

What's your opinion? @yuancu @LantaoJin @penghuo

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
yuancu: This is because we may move away from Calcite in the future. We remove dependency on the Calcite library for such general data types.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -91,13 +92,45 @@ default List<RelDataType> getParams() {
...
+        new CalciteFuncSignature(functionName.getName(), functionImp.getParams());
+    if (externalFunctionRegistry.containsKey(functionName)) {
+      externalFunctionRegistry.get(functionName).add(Pair.of(signature, functionImp));
+    } else {
+      externalFunctionRegistry.put(
+          functionName, new ArrayList<>(List.of(Pair.of(signature, functionImp))));
+    }
   }
 
   public @Nullable RexNode resolveSafe(
```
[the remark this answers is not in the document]
qianheng-aws: Emm, I'm guess it's caused by some charset is missing in that machine. The function name in the log is weird. Should be `i` instead `ı`.

```
java.lang.IllegalArgumentException: Unsupported operator: cıdrmatch
```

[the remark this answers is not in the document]
yuancu: Fixed. It was because `I` was converted to `i` in [AstExpressionBuilder.java](https://github.com/opensearch-project/sql/blob/main/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java) when system locale uses Turkic languages (e.g. `az-Latn-AZ`). Fixed by replacing `toLowercase()` with `toLowercase(Locale.ROOT)`.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
yuancu: Update: implemented by retrieving implementations from the external function registry before checking the internal one.

qianheng-aws (approved):

Issue #3557, opened 2025-04-16
[FEATURE] Aggregation is push down but not picked as the final plan in Calcite


qianheng-aws:
### Is your feature request related to a problem?
For test `testFilterAndAggPushDownExplain` in `ExplainIT`, we expect the aggregation being push down to scan in the final plan, while got a plan without being push down.

By debugging, we detect that there is a plan with aggregation being push down in the `RelSubset` but it's not pick as the final plan because it's cost is not the cheapest.

We need to fix this by introducing a new cost computing mechanism for `OpenSearchIndexScan` operator

Issue #3556, opened 2025-04-16
[BUG] Project is not push down correctly in Calcite


qianheng-aws:
**What is the bug?**
For the tests `testFilterPushDownExplain` in `ExplainIT`.

We expect the project being push down to scan operator and got plan like:
```
CalciteEnumerableIndexScan(
    table=[[OpenSearch, opensearch-sql_test_index_account]],
    PushDownContext=[
      [FILTER->>(\$8, 30), FILTER->AND(<(\$8, 40), >(\$3, 10000)), PROJECT->[8]],
      OpenSearchRequestBuilder(....
```

but got a plan without project being push down:
```
EnumerableCalc(expr#0..10=[{inputs}], age=[$t8])
  CalciteEnumerableIndexScan(
    table=[[OpenSearch, opensearch-sql_test_index_account]],
    PushDownContext=[
      [FILTER->>(\$8, 30), FILTER->AND(<(\$8, 40), >(\$3, 10000))],
      OpenSearchRequestBuilder(....
```

PR #3605, opened 2025-05-06
Add a new row count estimation mechanism for CalciteIndexScan

qianheng-aws:
### Description
Add a new row count mechanism for `CalciteIndexScan`, which is a new TableScan implemented by us for opensearch index.

Before this PR, this operator reused the same cost estimation logic as its parent class TableScan, which provided a constant value for its cost (i.e., 100 rows). However, since we also support pushing down operators like Project, Filter, Aggregation, etc. into this scan operator, and some operators actually affect the output rows of scan, we need a new implementation for this scan operator. Otherwise, in some cases, the Calcite optimizer will pick incorrect plans that don't actually have the lowest cost.

To estimate the cost of `CalciteIndexScan` more precisely, we need to imitate the row estimation logic implemented in `RelMdRowCount`. To make the Calcite optimizer prefer plans with push down rather than without, we add a factor (0.9) multiplied to the original row count of scan.

### Background
- In Calcite, by default, the costs of an operator are estimated completely based on their row count, and it will never let the result of row count goes below 1.

### Related Issues
Resolves 
https://github.com/opensearch-project/sql/issues/3557
https://github.com/opensearch-project/sql/issues/3556

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteIndexScan.java
```diff
@@ -34,6 +39,15 @@ public abstract class CalciteIndexScan extends TableScan {
   // so we cannot apply these actions right away.
   protected final PushDownContext pushDownContext;
 
+  /**
+   * The estimated row count of this index scan operator. Initialed with the value of setting {@link
+   * QUERY_SIZE_LIMIT}, and will be updated by the push down actions when estimating.
+   */
+  private Double rowCount;
```
penghuo, line 46: For scan operator, It should be index docs count? the query.size_limit configuration sets the maximum number of rows that a query can return, but not the index scan.

qianheng-aws: Yeah, to be more precisely, it should be docs count.

But it still works fine here to give a constant value here, because:
1. By default behavior without this PR, Calcite will also provide a constant value 100
2. We only need a relative value instead of a precise value for the cost computing, especially for the single table scanning. In the cases of multi-table joining, it should make difference if we introduce optimization like join-reorder, map-join, but seems those optimization only make sense in distribution environment.
3. Using the real doc count will introduce one more request but won't have much worth as said in 2.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteIndexScan.java
```diff
@@ -34,6 +39,15 @@ public abstract class CalciteIndexScan extends TableScan {
...
+
+  /** The status to indicate whether the row count has been estimated. */
+  private boolean rowCountEstimated = false;
```
LantaoJin, line 49: [question] why need this flag? what case will the scan be estimated multiple times? and how many times? Is it possible to remove the member variable `rowCount` in class? The method `estimateRowCount` should work independently.

qianheng-aws: The `estimateRowCount` will be called when:
1. Computing the cost of this operator itself
2. Computing the cost of its parent operator and the computing logic of that is based on the input row count. The times depends on how many parent operators will be generated during the optimization process. For example, when having plan `project-agg(avg)->scan`, the parent operators of `scan` will at least includes `agg(avg)`, `agg(sum/count)`, `EnumAgg` and may be `project` if have optimization of pushing project through agg.

qianheng-aws: `rowCount` and `rowCountEstimated` is viewed as kind of cache for the final estimated row count.

It's OK to remove this cache since other operators don't have such cache.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteIndexScan.java
penghuo: Can we use a fixed value (e.g., 10K) or a different setting? Using query.size_limit, which is unrelated to cost estimation, can be misleading. We don't expect changes to query.size_limit to affect the query pushdown logic, correct?

LantaoJin (approved):

LantaoJin (dismissed):

penghuo (commented):
Update doc. https://github.com/opensearch-project/sql/blob/main/docs/user/admin/settings.rst

docs/user/admin/settings.rst
```diff
@@ -757,3 +757,39 @@ This setting is present from 3.0.0-beta. You can enable Calcite as new query opt
...
+2. This setting is node scope.
+3. This setting can be updated dynamically.
+
+plugins.calcite.pushdown.rowcount.estimation.factor
+=======================
+
+Description
+-----------
+
+This setting is present from 3.0.0-beta. If Calcite pushdown optimization is enabled, this setting is used to estimate the row count of the query plan. The value is a factor to multiply the row count of the table scan to get the estimated row count.
```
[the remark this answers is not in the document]
qianheng-aws: Done

LantaoJin (approved):

penghuo (approved):

PR #3613, opened 2025-05-08
Bugfix: Timestamps sometimes converted to Dates in legacy engine

### LINKED ISSUE #1545: [BUG] Inconsistent behaviour for date field in nested collection queries

### LINKED ISSUE #3159: [BUG] timestamp data type handling is inconsistent
**What is the bug?**

When using SQL query, a timestamp field is handled differently if the SQL query has  a join. 

Data:
POST _bulk
{ "index": { "_index": "index_one"} }
{"stanza":4,"line_number":19,"line_text":"Presently my soul grew stronger; hesitating then no longer,"}
{ "index": { "_index": "index_one"} }
{"stanza":4,"line_number":20,"line_text":"\"Sir,\" said I, \"or Madam, truly your forgiveness I implore;"}
{ "index": { "_index": "index_one"} }
{"stanza":11,"line_number":64,"line_text":"Followed fast and followed faster till his songs one burden bore—"}
{ "index": { "_index": "index_one"} }
{"stanza":11,"line_number":65,"line_text":"Till the dirges of his Hope that melancholy burden bore"}

POST _bulk
{ "index": { "_index": "index_two"} }
{  "log_timestamp": "2024-11-13T09:11:01.668-06:00","stanza": 11,"some_number": "some data 11"}
{ "index": { "_index": "index_two"} }
{  "log_timestamp": "2024-11-13T09:12:01.668-06:00","stanza": 4,"some_number": "some data 4 u"}

1. Query index_two
        POST _plugins/_sql/
        {
          "query": "SELECT  * from index_two"
        }
      The field log_timestamp schema is 'timestamp' but the timezone information is not displayed.  (Image 1)

3. run a join on the 2 indexes.
        POST _plugins/_sql/
        {
          "query": "SELECT  two.log_timestamp, two.stanza, one.line_number, one.line_text FROM index_two two JOIN index_one one ON one.stanza = two.stanza"
        } 
 The field log_timestamp schema is 'date' now and the returned value has the timezone information.    (Image 2)

A TIMESTAMP data type should not change to date unless explicitly cast to one.
Also, the timezone information should always be included in the data. 

 - OS: Ubuntu 24.0.1 LTS
 - Version 2.1s
 - Plugins: SQL 

Image 1.

image

Image 2

image

Issue was found using the JDBC driver (1.4.0.1).  Issue was replicated using the Devtools on the dashboard.

Swiddis:
### Description
Another Legacy-V2 mismatch issue, something in Legacy sometimes converts `timestamp`s to `date`s, causing prod breakage for some JDBC users. I couldn't find the root cause of the inconsistent conversion, so I just threw a hack on `Protocol` instead. (By my understanding, Protocol is only called in Legacy.)

We should be able to remove all this when Calcite happens.

I invite reviewers to find a better way to do this.

### Related Issues
Resolves #1545
Resolves #3159

### Check List


integ-test/src/test/resources/datetime.json
```diff
@@ -3,6 +3,6 @@
 {"index":{"_id":"2"}}
 {"login_time":"2015-01-01T12:10:30Z"}
 {"index":{"_id":"3"}}
-{"login_time":"1585882955"}
+{"login_time":"1585882955000"}
```
Swiddis, line 6: I found while writing the reproducers that this was getting parsed as milliseconds instead of seconds, despite the magnitude. That gets sent to January 19, 1970 instead of (probably intended) April 3, 2020. I decided to update the row.

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/Protocol.java
```diff
@@ -214,7 +214,16 @@ private JSONArray getSchemaAsJson() {
     JSONArray schemaJson = new JSONArray();
 
     for (Column column : schema) {
-      schemaJson.put(schemaEntry(column.getName(), column.getAlias(), column.getType()));
+      // Hacky workaround for #3159, #1545: Legacy sometimes falsely converts `timestamp`s to
+      // `date`s, which causes
+      // breakage for consumers like JDBC. Until Calcite's done and we can delete legacy, just reset
+      // the type.
+      String t = column.getType();
+      if (t.equals("date")) {
```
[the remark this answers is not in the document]
Swiddis: Probably not, but I haven't figured out the 'smart' way to do this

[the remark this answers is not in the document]
Swiddis: Passes all the tests, so if there's a case where overconverting causes issues, it isn't tracked

integ-test/src/test/resources/datetime_nested.json
```diff
@@ -0,0 +1,8 @@
+{"index":{"_id":"1"}}
```
penghuo, line 1: leverage [YamlRestTest](https://github.com/opensearch-project/sql/blob/main/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/3477.yml) in future, avoid unncessary mapping and data files.

integ-test/src/test/java/org/opensearch/sql/sql/ComplexTimestampQueryIT.java
```diff
@@ -0,0 +1,136 @@
...
+      JSONObject column = schema.getJSONObject(i);
+      Assert.assertEquals("timestamp", column.getString("type"));
+    }
+  }
+
+  /** See: <a href="https://github.com/opensearch-project/sql/issues/3204">3204</a> */
+  // TODO currently out of scope due to V1/V2 engine feature mismatch. Should be fixed with Calcite.
+  @Test
+  @Ignore
+  public void joinTimestampComparison() throws IOException {
```
penghuo, line 70: have u try enable calcite, issue still exist?

dai-chen (commented):
Does this previous PR linked in the issue indicate any root cause or better fix? https://github.com/opensearch-project/sql/pull/3160

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/Protocol.java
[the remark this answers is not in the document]
dai-chen: So the hack is to replace the column type name only, but the value is not touched?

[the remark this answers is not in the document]
Swiddis: You can see in the bug reports that even when the column was `date`, the value still holds time information, so it's sufficient to just rename the column.

Swiddis:
Going to retry a proper fix later this week since this has been stuck for a while

Swiddis:
Update: So I spent a while trying to debug what was going on with the type mapping, and found that OpenSearch itself was returning `date`. It turns out that by following the mapping at https://docs.opensearch.org/docs/latest/search-plugins/sql/datatypes/ I was led astray, the real type mapping lives at https://github.com/opensearch-project/sql/blob/main/docs/user/general/datatypes.rst#data-types-mapping. It seems like this mapping was only done in V2, and never ported to Legacy.

I added a comment to this effect, but I think doing this at the serialization stage is correct, since otherwise Legacy seems to depend pretty highly on working with OS types directly and refactoring everything to apply this mapping internally would be a mess.

integ-test/src/test/resources/datetime_nested.json
Swiddis: Per side-channel discussion, leaving the tests as-is for now since we'll be backporting pretty far back

integ-test/src/test/java/org/opensearch/sql/sql/ComplexTimestampQueryIT.java
Swiddis: I tried porting the test to YamlRest w/ Calcite, and got this
```
expected [2xx] status code but api [sql] returned [400 Bad Request] [{
  "error": {
    "details": "Cannot invoke \"String.startsWith(String)\" because \"fieldName\" is null",
  },
  "status": 400
}]
```

Doesn't seem like it's ready yet, will leave it as ignored/todo

integ-test/src/test/java/org/opensearch/sql/sql/ComplexTimestampQueryIT.java
```diff
@@ -0,0 +1,136 @@
...
+  @Test
+  public void selectDatetimeWithoutNested() throws IOException {
+    String query =
+        String.format(
+            Locale.ROOT, "SELECT tab.login_time " + "FROM %s AS tab", TEST_INDEX_DATE_TIME_NESTED);
+
+    JSONObject result = executeQuery(query);
+    JSONArray schema = result.getJSONArray("schema");
+
+    Assert.assertFalse(schema.isEmpty());
```
RyanL1997, line 130: nit - Consider parameterizing the schema type check logic to reduce duplication:
```java
private void assertSchemaTypeIsTimestamp(JSONArray schema) {
    Assert.assertFalse(schema.isEmpty());
    for (int i = 0; i < schema.length(); i++) {
        Assert.assertEquals("timestamp", schema.getJSONObject(i).getString("type"));
    }
}
```

Swiddis: Probably not enough duplication here to warrant (this is a bit of a special case for type bugs), but the robust solution would be to make a custom schema assertion -- `assertSchemaMatches(schema, ["timestamp", "timestamp"])`. Then we throw it behind a custom assertion error that indicates the mismatching data types

penghuo (approved):

RyanL1997 (approved):

RyanL1997 (approved):

PR #3615, opened 2025-05-12
Support Limit pushdown

### LINKED ISSUE #3381: [FEATURE] Calcite Engine Framework: pushdown limit
Calcite Engine Framework: pushdown limit

yuancu:
### Description
Implement LIMIT pushdown.

### Related Issues
Resolves #3381

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
```diff
@@ -0,0 +1,93 @@
+package org.opensearch.sql.opensearch.planner.physical;
```
LantaoJin, line 1: copyright

yuancu: Done

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    final LogicalSort sort = call.rel(0);
+    final CalciteLogicalIndexScan scan = call.rel(1);
+
+    // The LogicalSort is a LIMIT that should be pushed down when its fetch field is not null and
+    // its collation is empty.
+    // For example: `sort name | head 5` should not be pushed down because it has a field collation.
+    if (sort.fetch != null && sort.getCollation().getFieldCollations().isEmpty()) {
```
dai-chen, line 36: np: Is it possible both `sort.fetch` and `sort.getCollation().getFieldCollations()` both null/empty? Maybe clearer to extract this conditional check along with all comments to a method, e.g. `isLimitOffsetOnly`?

dai-chen (dismissed):

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
```diff
@@ -0,0 +1,98 @@
...
+public class OpenSearchLimitIndexScanRule extends RelRule<OpenSearchLimitIndexScanRule.Config> {
+
+  protected OpenSearchLimitIndexScanRule(Config config) {
+    super(config);
+  }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    final LogicalSort sort = call.rel(0);
+    final CalciteLogicalIndexScan scan = call.rel(1);
```
dai-chen, line 31: Does Calcite take care of the case that the child of sort is not index scan?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
yuancu: Thanks for the suggestions! 

Pagination operations will specify `sort.fetch` and optionally `sort.offset`, while sorting operations will specify `sort.collation`. 

`sort.fetch` and `sort.getCollation().getFieldCollations()` may both be null/empty when only offset is specified: e.g. `source country | from 5`. But this is not a valid PPL. Therefore, I think it's not likely in PPL that both of them are null/empty. Please correct me if I'm wrong.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
yuancu: This rule only matches `LogicalSort -> CalciteLogicalIndexScan`, which is specified in `OpenSearchLimitIndexScanRule.Config`. When the child of sort is not index scan, the query will not be caught and pushed-down via this rule.

Do you mean "is it possible that a `limit` does not fall into the `LogicalSort -> CalciteLogicalIndexScan` pattern"?

dai-chen (dismissed):
Thanks for the changes!

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
LantaoJin: I think the implementation here may have a bug:
```
source=t | where a=0 | head 10;
source=t | head 10 | where a=0;
```
Above two PPL will all push down the filter and limit to DSL. Please check are the DSLs generated same?
If they are same, the results are incorrect since the two PPL queries should have different behaviors.

yuancu: Yes, the two queries return the same physical plan with this pushdown:

```java
CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[FILTER->=($8, 20), LIMIT->10, PROJECT->[account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":10,"timeout":"1m","query":{"term":{"age":{"value":20,"boost":1.0}}},"_source":{"includes":["account_number","firstname","address","balance","gender","city","employer","state","age","email","lastname"],"excludes":[]},"sort":[{"_doc":{"order":"asc"}}]}, requestedTotalSize=10, pageSize=null, startFrom=0)])\n
```

Also, in V2, explaining both of them returns the following request:
```python
OpenSearchQueryRequest(indexName=country, sourceBuilder={"from":0,"size":10,"timeout":"1m","query":{"term":{"age":{"value":20,"boost":1.0}}},"_source":{"includes":["state","name","year","country","age","month"],"excludes":[]},"sort":[{"_doc":{"order":"asc"}}]}, needClean=true, searchDone=false, pitId=null, cursorKeepAlive=null, searchAfter=null, searchResponse=null)
```

LantaoJin: The DSL means `sourceBuilder={...}`. They are same. Please fix it by check the size in optimize rule

yuancu: @LantaoJin Fixed

The size in the request builder is set to a default size at initialization (See [OpenSearchIndex.java#L285](https://github.com/opensearch-project/sql/blob/94fb171f81e61745f2ee5264f01984524e716aa3/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java#L285)). Therefore, we can not determine whether if there is already a limit by simply check whether size is set.

I created a flag in `PushDownContext` to log whether there was a limit pushdown.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchFilterIndexScanRule.java
```diff
@@ -55,7 +56,14 @@ public interface Config extends RelRule.Config {
...
                                 b1.operand(CalciteLogicalIndexScan.class)
-                                    .predicate(OpenSearchIndexScanRule::test)
+                                    .predicate(
+                                        // Filter pushdown is skipped if a limit has already been
+                                        // pushed down because the current DSL cannot correctly
+                                        // handle filter pushdown after limit. Both "limit after
+                                        // filter" and "filter after limit" result in the same
+                                        // limit-after-filter DSL.
+                                        Predicate.not(OpenSearchIndexScanRule::isLimitPushed)
+                                            .and(OpenSearchIndexScanRule::test))
```
LantaoJin, line 66: Would `OpenSearchAggregateIndexScanRule` be applied with the same fixing?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -197,7 +197,11 @@ public void pushDownSort(List<SortBuilder<?>> sortBuilders) {
...
   /** Pushdown size (limit) and from (offset) to DSL request. */
   public void pushDownLimit(Integer limit, Integer offset) {
-    requestedTotalSize = limit;
+    // If there are multiple limit, we take the minimum among them
+    // E.g. for `source=t | head 10 | head 5`, we take 5
+    // This also ensures that the limit won't exceed the initial default value. (set to
+    // Settings.Key.QUERY_SIZE_LIMIT in OpenSearchIndex)
+    requestedTotalSize = Math.min(limit, requestedTotalSize);
     startFrom = offset;
     sourceBuilder.from(offset).size(limit);
```
LantaoJin, line 206: shouldn't be `sourceBuilder.from(offset).size(requestedTotalSize)`? 

Could you add 2 tests for this case  in explain IT:
Should pushdown size = 5
```
source=t | head 10 | head 5
source=t | head 5 | head 10
```

Should pushdown size=10
```
source=t | head 10 | filter a = 0 | head 5
```

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
```diff
@@ -109,6 +109,36 @@ public void testLimitPushDownExplain() throws Exception {
                 + "| fields ageMinus"));
   }
 
+  @Test
+  public void testLimitWithFilterPushdownExplain() throws Exception {
+    // TODO: Fix limit-then-filter pushdown without Calcite
```
LantaoJin, line 114: how's the effort to fix the v2 bug together?

LantaoJin (changes requested):

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchFilterIndexScanRule.java
yuancu: Yes, it should. The results are incorrect without it.

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
yuancu: Fixed with minor effort.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -197,9 +197,13 @@ public void pushDownSort(List<SortBuilder<?>> sortBuilders) {
 
   /** Pushdown size (limit) and from (offset) to DSL request. */
   public void pushDownLimit(Integer limit, Integer offset) {
-    requestedTotalSize = limit;
+    // If there are multiple limit, we take the minimum among them
+    // E.g. for `source=t | head 10 | head 5`, we take 5
+    // This also ensures that the limit won't exceed the initial default value. (set to
+    // Settings.Key.QUERY_SIZE_LIMIT in OpenSearchIndex)
+    requestedTotalSize = Math.min(limit, requestedTotalSize);
     startFrom = offset;
```
qianheng-aws, line 205: I think it's not correct to override with the latest offset directly.

Let's say SQL like `select * from (select * from table limit 10 offset 1) limit 5 limit 2`, it's not correct to use `2` as the offset. The right offset should be `1 + 2` I think. Please add a test for this case.

yuancu: You are right. Accumulating them should be the right behavior. Fixed.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * The LogicalSort is a LIMIT that should be pushed down when its fetch field is not null and its
+   * collation is empty. For example: <code>sort name | head 5</code> should not be pushed down
+   * because it has a field collation.
+   *
+   * @param sort The LogicalSort to check.
+   * @return True if the LogicalSort is a LIMIT that can be pushed down, false otherwise.
+   */
+  private static boolean isLogicalSortLimit(LogicalSort sort) {
```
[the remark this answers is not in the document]
yuancu: Fixed

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
yuancu: Yes, it was a mistake. Fixed.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -164,4 +164,23 @@ public CalciteLogicalIndexScan pushDownAggregate(Aggregate aggregate) {
...
+          PushDownAction.of(
+              PushDownType.LIMIT,
+              limit,
+              requestBuilder -> requestBuilder.pushDownLimit(limit, offset)));
+      return newScan;
+    } catch (Exception e) {
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Cannot pushdown the limit {}", limit, e);
+      } else {
+        LOG.warn("Cannot pushdown the limit {}, ", limit);
```
LantaoJin, line 181: `Cannot pushdown the limit <Integer>` is a meaningless message and `warn` level is too severe.

yuancu: Changed the error message to `LOG.debug("Cannot pushdown limit {} with offset {}. Exception: {}", limit, offset, e);`

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
```diff
@@ -0,0 +1,94 @@
...
+        call.transformTo(newScan);
+      }
+    }
+  }
+
+  private static Integer extractLimitValue(RexNode fetch) {
+    if (fetch instanceof RexLiteral) {
+      return ((RexLiteral) fetch).getValueAs(Integer.class);
+    }
+    return null;
```
LantaoJin, line 47: would this fetch always a `RexLiteral`? how about `head 1+1`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
LantaoJin: Checked with @qianheng-aws offline. Now I think the log won't be generated too many. So please added this `else` back and change it to `info`. Can you fix all the these `warn` logs in this class?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
yuancu: Yes, as specified in [OpenSearchPPLParser.g4#L149](https://github.com/opensearch-project/sql/blob/94fb171f81e61745f2ee5264f01984524e716aa3/ppl/src/main/antlr/OpenSearchPPLParser.g4#L149) and in [OpenSearchSQLParser.g4#L144](https://github.com/opensearch-project/sql/blob/d767868f42f012595003eb40be05056c1d5b4804/sql/src/main/antlr/OpenSearchSQLParser.g4#L144).

LantaoJin (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
qianheng-aws: It's not correct to set `requestedTotalSize`  to the min here, I think.

`source=t | head 10 from 1 | head 10 from 2` should return 8 rows, but it actually will return 10 rows under the current implementation.

qianheng-aws (approved):

LantaoJin (approved):

Issue #3566, opened 2025-04-21
[BUG] Correlated exist subquery fail with NPE in RelDecorrelator


LantaoJin:
**What is the bug?**
```
  @Test
  public void testTpchQuery() {
    String actual =
        execute(
            String.format(
                "source = %s | where exists [ source = %s | where l_orderkey = o_orderkey] | stats count() by o_orderpriority",
                TEST_INDEX_TPCH_ORDERS, TEST_INDEX_TPCH_LINEITEM));
    assertEquals("", actual);
  }
```
Fail with 
```
Caused by: java.lang.NullPointerException
        at java.base/java.util.Objects.requireNonNull(Objects.java:233) ~[?:?]
        at org.apache.calcite.sql2rel.RelDecorrelator.getNewForOldInputRef(RelDecorrelator.java:1383) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.apache.calcite.sql2rel.RelDecorrelator.access$400(RelDecorrelator.java:147) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.apache.calcite.sql2rel.RelDecorrelator$DecorrelateRexShuttle.visitInputRef(RelDecorrelator.java:1665) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.apache.calcite.sql2rel.RelDecorrelator$DecorrelateRexShuttle.visitInputRef(RelDecorrelator.java:1623) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.apache.calcite.rex.RexInputRef.accept(RexInputRef.java:125) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
        at org.apache.calcite.sql2rel.RelDecorrelator.decorrelateExpr(RelDecorrelator.java:385) ~[opensearch-sql-3.0.0.0-beta1-SNAPSHOT.jar:3.0.0.0-beta1-SNAPSHOT]
[...]

PR #3616, opened 2025-05-12
All TPC-H PPL queries pass 

### LINKED ISSUE #3178: [FEATURE]Add a TPC-H PPL query suite

The TPC-H (Transaction Processing Performance Council Benchmark H) test suite is a widely recognized benchmark for evaluating the performance and efficiency of decision support systems, particularly those handling complex SQL queries on large datasets.

Incorporating the TPC-H test suite into the OpenSearch SQL repository is crucial for both PPL & SQL for ensuring the robustness, scalability, and optimization of SQL/PPL query processing in OpenSearch.
OpenSearch SQL engine can be tested against real-world analytical workloads, providing insights into query execution performance, join and aggregation handling. 
This test would eventually become a benchmark for identification of bottlenecks, benchmarking against industry standards, and adding confidence among users and developers that OpenSearch SQL/PPL repository can handle sophisticated analytical use cases effectively.

Add TPC-H PPL query suite for:

- ensure all TPC-H SQL can be rewritten by PPL
- ensure all 22 PPL queries can be executed in Spark
- ensure all physical plans generated matches the expected output

- [Spark PPL TPCG](https://github.com/opensearch-project/opensearch-spark/pull/830)
- [Spark TPCH IT folder](https://github.com/opensearch-project/opensearch-spark/tree/main/integ-test/src/integration/resources/tpch)

LantaoJin:
### Description
1. Support #3178 
1.1 Add TPC-H data, mappings and queries rewritten by PPL
1.2 Add a `CalcitePPLTpchIT` to run all 22 queries
1.3 Verified in IT that all query results are as same as SparkSQL's.
2. Bugfix related to TPC-H queries.
2.1 https://github.com/opensearch-project/sql/issues/3566
2.2 https://github.com/opensearch-project/sql/issues/3617

### Related Issues
Resolves #3178, #3566 and #3617

### Check List


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -437,8 +473,47 @@ public RelNode visitJoin(Join node, CalcitePlanContext context) {
...
+          .accept(
+              new RelShuttleImpl() {
+                @Override
+                public RelNode visit(TableScan scan) {
+                  final RelOptTable table = scan.getTable();
+                  rightTableName.get().addAll(table.getQualifiedName());
+                  return super.visit(scan);
+                }
+              });
+      String rightTableQualifiedName = String.join(".", rightTableName.get());
```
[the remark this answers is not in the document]
penghuo: Agree. OpenSearch should be removed. 
Why result have r.id? Should only return id, right?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLJoinIT.java
```diff
@@ -68,7 +69,7 @@ public void testJoinWithCondition() {
         schema("state", "string"),
         schema("country", "string"),
         schema("occupation", "string"),
-        schema("country0", "string"),
+        schema("b.country", "string"),
```
penghuo, line 72: IMO, we should have unified rules to handle duplicated columns. 2 options in my mind
1. similar as lookup command, using APPEND and REPLACE options to handle duplicate columns. 
2. Or follow SQL standard way, then column name of  a.country should be a.country instead of country.

penghuo: what if user does not specify fields command, what is expected output?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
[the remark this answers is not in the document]
LantaoJin: No matter the fields command apply or not, the output is shown above with this patch.
Without this patch, the output is
```
{
  "schema": [
    {
      "name": "id",
      "type": "string"
    },
    {
      "name": "id0",
      "type": "string"
    }
  ],
  "datarows": [
    ...
    ]
  ],
  "total": 4,
  "size": 4
}
```

Without `eval a=1` commands, `| fields table2.id` (in the final project) can be accessed by Calcite's internal alias mapping, then the `table2.id` rename to `id0` and added to its fields list of stack top. That's why duplicated column in schema is `id0`.
Unfortunately, PPL is not SQL, `eval` will invoke RelBuilder.projectPlus(), so if we add a `fields` command after `eval` command, the plan will be
```
project
  \-project <-- this project rename table2.id to id0 and add it to the fields list of stack top
      \-..
          \-join
              \-scan
              \-scan
```

In Spark (or most of other databases), they keep the original name in outputs, for example
```
spark-sql (default)> create table left (id int, col1 string, col2 string) using parquet;
spark-sql (default)> create table right (id int, col1 string, col3 string) using parquet;

spark-sql (default)> select * from left join right on left.id = right.id;
```
outputs
```
id	col1	col2	id	col1	col3
```
But user cannot access ambiguous column without alias or table name.
```
spark-sql (default)> select id from left join right on left.id = right.id;
[AMBIGUOUS_REFERENCE] Reference `id` is ambiguous, could be: [`spark_catalog`.`default`.`left`.`id`, `spark_catalog`.`default`.`right`.`id`].; line 1 pos 7
```
Spark uses `exprId` to identify the duplicated column names. But this is no `exprId` in Calcite, it uses field name to distinguish duplicated column names (id, id0, id1...):
```
0: jdbc:calcite:model=src/test/resources/mode> select t.age,t.age0 from (select * from emps as e1 join emps as e2 on e1.DEPTNO=e2.DEPTNO) t;
+------+------+
| AGE  | AGE0 |
+------+------+
| 25   | 25   |
| 80   | 80   |
| 80   | 5    |
| 5    | 80   |
| 5    | 5    |
| null | null |
| null | null |
| null | null |
| null | null |
+------+------+
```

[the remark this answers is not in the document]
LantaoJin: > Agree. OpenSearch should be removed.

If we remove the `OpenSearch` (database information), reusing this code for Spark in future will try to search the table in its current namespace. For example, there are `database1.table1` and `database2.table1`. The resolving plan only works in current namespace `database1`.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLJoinIT.java
LantaoJin: > we should have unified rules to handle duplicated columns. IMO, 2 options
> 
> 1. similar as lookup command, using APPEND and REPLACE options to handle duplicate columns.
> 2. Or follow SQL standard,  user must explicitly qualify or alias the columns, then, the column name of  a.country should be a.country instead of country.

1. The outputs of `lookup` doesn't contain duplicated fields. I think `lookup` and `join` are different.
2. `user must explicitly qualify or alias the columns` do you mean access them? Yes for now, user must explicitly specific the `id` field by `alias.id` or `tablename.id`, else an ambiguous exception shows. Without this patch, `alias.id` or `tablename.id` cannot work since the `id` renamed to `id0`. It just impacts the output schema.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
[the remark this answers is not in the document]
LantaoJin: > Agree. OpenSearch should be removed.

done

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -280,7 +281,24 @@ public RexNode visitQualifiedName(QualifiedName node, CalcitePlanContext context
...
       } else if (parts.size() == 2) {
         // 1.2 Handle the case of `t1.id = t2.id` or `alias1.id = alias2.id`
-        return context.relBuilder.field(2, parts.get(0), parts.get(1));
+        try {
+          return context.relBuilder.field(2, parts.get(0), parts.get(1));
+        } catch (IllegalArgumentException e) {
+          // Similar to the step 2.3.
+          List<String> candidates =
+              context.relBuilder.peek(1).getRowType().getFieldNames().stream()
+                  .filter(col -> substringAfterLast(col, ".").equals(parts.getLast()))
```
qianheng-aws, line 290: It comes to me that we also have to verify the substring before last `.` is an alias.

For example: if table `t1` with alias `n1` has schema:
```
a: Struct(b: Int), b: Int
```
When having conflict on column `b`, its row type becomes to:
```
a: Struct, a.b: Int, n1.b: Int
```
Then, it may resolve `t1.b` incorrectly by using `t1.a.b` while we actually want `t1.n1.b`.

qianheng-aws:
```
Expression result  = context.relBuilder.field(2, parts.get(0), candidate);
Expression result1 = context.relBuilder.field(2, substringBeforeLast(candidate, "."), candidate);
if (result == result1) return result;
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -437,8 +473,53 @@ public RelNode visitJoin(Join node, CalcitePlanContext context) {
...
+      List<String> leftColumns = context.relBuilder.peek(1).getRowType().getFieldNames();
+      List<String> rightColumns = context.relBuilder.peek().getRowType().getFieldNames();
+      AtomicReference<List<String>> rightTableName = new AtomicReference<>(new ArrayList<>());
+      context
+          .relBuilder
+          .peek()
+          .accept(
+              new RelShuttleImpl() {
+                @Override
+                public RelNode visit(TableScan scan) {
```
qianheng-aws, line 494: What if the column comes from a Subquery instead of TableScan?

LantaoJin: > What if the column comes from a Subquery instead of TableScan?

The subquery alias will be used as field prefix rather than table name. I will add a test for this case

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
LantaoJin: Nested fields resolving is out of the scope, let's fix it in https://github.com/opensearch-project/sql/issues/3459

qianheng-aws: Although #3459 is resolved, we still have this bad case.

LantaoJin:
The fix for #3617  is more complex than expected. To avoid blocking bugfix of #3566 and TPCH related work, I'm splitting this PR into 3 PRs and closing this PR.

PR #3620, opened 2025-05-13
Support Sort pushdown

### LINKED ISSUE #3380: [FEATURE] Calcite Engine Framework: pushdown sort
Calcite Engine Framework: pushdown sort

yuancu:
### Description
Support Sort pushdown

### Related Issues
Resolves #3380 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -120,13 +132,30 @@ public CalciteLogicalIndexScan pushDownProject(List<Integer> selectedColumns) {
       builder.add(fieldList.get(project));
     }
     RelDataType newSchema = builder.build();
-    CalciteLogicalIndexScan newScan = this.copyWithNewSchema(newSchema);
+
+    // Projection may alter the index of the collations.
+    // E.g. For sort age
+    // `Sort($1)\n TableScan(name, age)` may become
+    // `Sort($0)\n Project(age)\n  TableScan(name, age)` after projection.
```
penghuo: Why `Sort($1)\n TableScan(name, age)` may become `Sort($0)\n Project(age)\n TableScan(name, age)`? SortRule should pushdown sort already?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +236,124 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
+
+  /**
+   * Check if the sort by collations contains any aggregators that are pushed down. E.g. In `stats
+   * avg(age) as avg_age by state | sort avg_age`, the sort clause has `avg_age` which is an
+   * aggregator. The function will return true in this case.
+   *
+   * @param collations List of collation names to check against aggregators.
+   * @return True if any collation name matches an aggregator output, false otherwise.
+   */
+  private boolean hasAggregatorInSortBy(List<String> collations) {
```
penghuo, line 331: is there an IT/UT to cover this?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +236,124 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
   }
+
+  public CalciteLogicalIndexScan pushDownSort(List<RelFieldCollation> collations) {
+    try {
+      List<String> collationNames =
+          collations.stream()
+              .map(RelFieldCollation::getFieldIndex)
+              .map(index -> this.getRowType().getFieldNames().get(index))
+              .collect(Collectors.toList());
+      if (getPushDownContext().isAggregatePushed() && hasAggregatorInSortBy(collationNames)) {
```
penghuo, line 252: what is DSL query if  `... | sort age | stats avg(age) as avg_age by state`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +236,124 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
+              PushDownType.SORT,
+              builders.toString(),
+              requestBuilder -> requestBuilder.pushDownSort(builders)));
+      return newScan;
+    } catch (Exception e) {
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Cannot pushdown the sort {}", collations, e);
+      } else {
+        LOG.warn("Cannot pushdown the sort {}, ", collations);
+      }
```
penghuo: santize collations? does it include field name or value?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +236,124 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
+        int index = collation.getFieldIndex();
+        String fieldName = this.getRowType().getFieldNames().get(index);
+        RelFieldCollation.Direction direction = collation.getDirection();
+        // Default sort order is ASCENDING
+        SortOrder order =
+            RelFieldCollation.Direction.DESCENDING.equals(direction)
+                ? SortOrder.DESC
+                : SortOrder.ASC;
+        // TODO: support script sort and distance sort
+        SortBuilder<?> sortBuilder;
```
penghuo, line 283: is there IT/UT cover this?

integ-test/src/test/resources/expectedOutput/calcite/explain_sort_push.json
```diff
@@ -1,6 +1,6 @@
...
-    "physical": "EnumerableCalc(expr#0=[{inputs}], expr#1=[30], expr#2=[>($t0, $t1)], age=[$t0], $condition=[$t2])\n  EnumerableSort(sort0=[$0], dir0=[ASC])\n    CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[age]], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"_source\":{\"includes\":[\"age\"],\"excludes\":[]}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
+    "physical": "CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[age], SORT->[{\n  \"age\" : {\n    \"order\" : \"asc\"\n  }\n}], FILTER->>($0, 30)], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"query\":{\"range\":{\"age\":{\"from\":30,\"to\":null,\"include_lower\":false,\"include_upper\":true,\"boost\":1.0}}},\"_source\":{\"includes\":[\"age\"],\"excludes\":[]},\"sort\":[{\"age\":{\"order\":\"asc\"}}]}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
```
penghuo: Add another IT on timestamp field sort? `source=xxx | sort @timestamp`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +236,124 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
+          existingCollation == null ? List.of() : existingCollation.getFieldCollations();
+      List<RelFieldCollation> mergedCollations =
+          mergeCollations(existingFieldCollations, collations);
+
+      // Propagate the sort to the new scan
+      RelTraitSet traitsWithCollations = getTraitSet().plus(RelCollations.of(mergedCollations));
+      CalciteLogicalIndexScan newScan =
+          new CalciteLogicalIndexScan(
+              getCluster(),
+              traitsWithCollations,
```
qianheng-aws, line 263: After adding collation by pushing down sort, we should remove them if pushing down aggregation afterwards, as aggregation will eliminate any collation.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
yuancu: Yes, `testSortWithAggregationExplain` covers this:
```
source=opensearch-sql_test_index_account | stats avg(age) AS avg_age by state, city | sort avg_age
```

should not be pushed down

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
yuancu: In the latest implementation, I remove sort if an aggregation follows. `source=opensearch-sql_test_index_account | sort balance, age | stats avg(age)` results in the following DSL:
```json
{
  "from": 0,
  "size": 0,
  "timeout": "1m",
  "_source": {
    "includes": [
      "balance",
      "age"
    ],
    "excludes": []
  },
  "aggregations": {
    "avg(age)": {
      "avg": {
        "field": "age"
      }
    }
  }
}
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
yuancu: During optimization, the index in collations should point to its location in the current row instead of its location in the original table. As a result, even if sort is already pushed down, the trait RelCollation, which represents the sort status of the relation being optimized, should be updated to reflect the latest order status when there is a projection after sort.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
yuancu: Fixed. Replaced with their field names.

integ-test/src/test/resources/expectedOutput/calcite/explain_sort_push.json
yuancu: Added.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
yuancu: Fixed

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
yuancu: No, I haven't found any. Do you know the proper syntax for sorting geopoints or with script? I tried sorting on geopoints, it reports `can't sort on geo_point field without using specific sorting feature, like geo_distance`. But I haven't figured out the correct syntax for sorting geopoints (it's not in the g4 file or the documentation).

yuancu:
I found a corner case bug when sorting nested fields: sorting nested text fields does not leverage its keyword.

For example ([mapping](https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/resources/indexDefinitions/nested_simple_index_mapping.json)):
For query: `source=opensearch-sql_test_index_nested_simple | sort address.state`, its DSL is as follows
```json
{
  "from": 0,
  "timeout": "1m",
  "_source": {
    "includes": [
      "name",
      "address",
      "id",
      "age"
    ],
    "excludes": []
  },
  "sort": [
    {
      "address.state": {
        "order": "asc"
      }
    }
  ]
}
```

But when executing the plan, I get the following response:
```json
{
  "error": {
    "details": "Shard[0]: java.lang.IllegalArgumentException: Text fields are not optimised for operations that require per-document field data like aggregations and sorting, so these operations are disabled by default. Please use a keyword field instead. Alternatively, set fielddata=true on [address.state] in order to load field data by uninverting the inverted index. Note that this can use significant memory.\n\nFor more details, please send request for Json format to see the raw response from OpenSearch engine.",
  },
  "status": 400
}
```

In contrast, the DSL in v2 uses the **keyword** field instead of the text field itself.
```json
{
  "from": 0,
  "size": 10000,
  "timeout": "1m",
  "_source": {
    "includes": [
      "name",
      "address",
      "id",
      "age"
    ],
    "excludes": []
  },
  "sort": [
    {
      "address.state.keyword": {
        "order": "asc",
        "missing": "_first"
      }
    }
  ]
}
```

The plan in v2 can be executed with the following result. Although it's not quite correct. The engine seems to pick a random `address.state` from the nested list and displays another one when project with fields.
```json
{
    "schema": [
        {
            "name": "address.state",
            "type": "string"
        }
    ],
    "datarows": [
        [
            "Florida"
        ],
        [
            "CA"
        ],
        [
            "NY"
        ],
        [
            "NC"
        ],
        [
            "TX"
        ]
    ],
    "total": 5,
    "size": 5
}
```

**Update**
Fixed by pointing to its keyword field when sorting a text field (if possible).

yuancu:
Another corner case to be fixed: sort is not pushed down when when there is a filter before it.

The physical plan for the query `source=bank | where birthdate < '2023-01-01 23:59:58.000000000' | sort birthdate | fields  birthdate` is:
```
EnumerableSort(sort0=[$0], dir0=[ASC])
  EnumerableCalc(expr#0=[{inputs}], expr#1=[TIMESTAMP($t0)], expr#2=['2023-01-01 23:59:58':EXPR_TIMESTAMP VARCHAR], expr#3=[<($t1, $t2)], birthdate=[$t0], $condition=[$t3])
    CalciteEnumerableIndexScan(table=[[OpenSearch, bank]], PushDownContext=[[PROJECT->[birthdate]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"timeout":"1m","_source":{"includes":["birthdate"],"excludes":[]}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```

However, in v2, the plan for it is :
```
{
    "root": {
        "name": "ProjectOperator",
        "description": {
            "fields": "[birthdate]"
        },
        "children": [
            {
                "description": {
                    "request": "OpenSearchQueryRequest(indexName=bank, sourceBuilder={\"from\":0,\"size\":10000,\"timeout\":\"1m\",\"query\":{\"range\":{\"birthdate\":{\"from\":null,\"to\":1672617598000,\"include_lower\":true,\"include_upper\":false,\"boost\":1.0}}},\"_source\":{\"includes\":[\"birthdate\"],\"excludes\":[]},\"sort\":[{\"birthdate\":{\"order\":\"asc\",\"missing\":\"_first\"}}]}, needClean=true, searchDone=false, pitId=*, cursorKeepAlive=1m, searchAfter=null, searchResponse=null)"
                },
                "children": []
            }
        ]
    }
}
```

From the plan, `where ... | sort ...` is pushed down in v2 but not in Calcite.

**Update**
- The reason turns out to be related to timestamp comparison. `where account_number > 10 | sort account_number` can be properly pushed down.
- Will be fixed by #3798

penghuo (dismissed):

integ-test/src/test/resources/expectedOutput/calcite/explain_sort_push.json
```diff
@@ -1,6 +1,6 @@
...
-    "physical": "EnumerableCalc(expr#0=[{inputs}], expr#1=[30], expr#2=[>($t0, $t1)], age=[$t0], $condition=[$t2])\n  EnumerableSort(sort0=[$0], dir0=[ASC])\n    CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[age]], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"_source\":{\"includes\":[\"age\"],\"excludes\":[]}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
+    "physical": "CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[age], SORT->[{\n  \"age\" : {\n    \"order\" : \"asc\",\n    \"missing\" : \"_last\"\n  }\n}], FILTER->>($0, 30)], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"query\":{\"range\":{\"age\":{\"from\":30,\"to\":null,\"include_lower\":false,\"include_upper\":true,\"boost\":1.0}}},\"_source\":{\"includes\":[\"age\"],\"excludes\":[]},\"sort\":[{\"age\":{\"order\":\"asc\",\"missing\":\"_last\"}}]}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
```
LantaoJin, line 4: > `\"sort\":[{\"age\":{\"order\":\"asc\",\"missing\":\"_last\"}}]` 

Here `_last` is different with v2's behaviour, in v2, default null direction (missing) for ASC is `first`. But in Calcite, the ASC maps to `last`:
https://github.com/apache/calcite/blob/2fb437b93774d3d69df66547715b374d9281fdbb/core/src/main/java/org/apache/calcite/rel/RelFieldCollation.java#L134-L136

We can add a TODO to align with v2, but my question is that is the behaviour of v2 correct? @dai-chen @penghuo any thoughts about it ?

this issue is not a blocker for this PR.

integ-test/src/test/resources/expectedOutput/ppl/explain_sort_rename_push.json
```diff
@@ -0,0 +1,50 @@
...
+                "description": {
+                  "mapping": {
+                    "firstname": "name"
+                  }
+                },
+                "children": [
+                  {
+                    "name": "OpenSearchIndexScan",
+                    "description": {
+                      "request": "OpenSearchQueryRequest(indexName=opensearch-sql_test_index_account, sourceBuilder={\"from\":0,\"size\":10000,\"timeout\":\"1m\"}, needClean=true, searchDone=false, pitId=s9y3QQEhb3BlbnNlYXJjaC1zcWxfdGVzdF9pbmRleF9hY2NvdW50Fndla1VpMi1kVHh5Qi1lYnhPQnlSbXcAFkU4Qm9UVURIUWI2a3pjNkhmQkxvc2cAAAAAAAAAAAsWcnUyVHZfNk1SeC1DMFNFdGtPcDN4QQEWd2VrVWkyLWRUeHlCLWVieE9CeVJtdwAA, cursorKeepAlive=1m, searchAfter=null, searchResponse=null)"
```
LantaoJin, line 38: Good catching, the v2 can not pushdown with `rename`.

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
```diff
@@ -89,6 +88,77 @@ public void testSortPushDownExplain() throws IOException {
...
+            "source=opensearch-sql_test_index_account "
+                + "| sort account_number, firstname, address, balance "
+                + "| sort - balance, - gender, address "
+                + "| fields account_number, firstname, address, balance, gender"));
+  }
+
+  @Test
+  public void testSortThenAggregatePushDownExplain() throws IOException {
+    // TODO: Remove pushed-down sort in DSL in expectedOutput/ppl/explain_sort_then_agg_push.json
+    //  existing collations should be eliminated when pushing down aggregations
```
LantaoJin: is this TODO for v2?

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
```diff
@@ -89,6 +88,77 @@ public void testSortPushDownExplain() throws IOException {
...
+                + "| sort avg_age"));
+
+    // sorts whose by fields are not aggregators can be pushed down.
+    // This test is covered in testExplain
+  }
+
+  @Test
+  public void testMultiSortPushDownExplain() throws IOException {
+    // TODO: Fix the expected output in expectedOutput/ppl/explain_multi_sort_push.json
+    //  balance and gender should take precedence over account_number and firstname
```
LantaoJin: Seems another bug of v2 sort. 
```
source=opensearch_dashboards_sample_data_logs  | head 10 | sort bytes | sort - bytes
```
with DSL pushdown
```
{
  "from": 0,
  "size": 10,
  "timeout": "1m",
      "sort": [{
        "bytes": {
            "order": "asc",
            "missing": "_first"
        }
    }, {
        "bytes": {
            "order": "desc",
            "missing": "_last"
        }
    }]
}
```
the final result is sorted ASC.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +246,155 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
+                : SortOrder.ASC;
+        // TODO: support script sort and distance sort
+        SortBuilder<?> sortBuilder;
+        if (ScoreSortBuilder.NAME.equals(fieldName)) {
+          sortBuilder = SortBuilders.scoreSort();
+        } else {
+          String missing =
+              switch (nullDirection) {
+                case FIRST -> "_first";
+                case LAST -> "_last";
```
LantaoJin, line 290: a workaround to align with v2 is changing to
```
case FIRST -> "_last";
case LAST -> "_first";
```
But not recommended.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +246,155 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
+          PushDownAction.of(
+              PushDownType.SORT,
+              builders.toString(),
+              requestBuilder -> requestBuilder.pushDownSort(builders)));
+      return newScan;
+    } catch (Exception e) {
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Cannot pushdown the sort {}", getCollationNames(collations), e);
+      } else {
+        LOG.warn("Cannot pushdown the sort {}, ", getCollationNames(collations));
```
LantaoJin: change to `info`.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +246,155 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
+
+  private List<String> getCollationNames(List<RelFieldCollation> collations) {
+    return collations.stream()
+        .map(collation -> getRowType().getFieldNames().get(collation.getFieldIndex()))
+        .toList();
+  }
+
+  /**
+   * Check if the sort by collations contains any aggregators that are pushed down. E.g. In `stats
+   * avg(age) as avg_age by state | sort avg_age`, the sort clause has `avg_age` which is an
```
LantaoJin, line 325: it may have a more complex case:
```
avg(age) as avg_age by state | rename avg_age as new_avg_age | sort new_avg_age
```
I don't think the current `hasAggregatorInSortBy` works for that.

LantaoJin:
@yuancu do we have follow cases in ExplainIT
```
source=t | head 10 | sort field // pushdown limit only
source=t | sort field | head 10 // pushdown sort only or pushdown sort + limit (depends on the execution order of DSL)
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
yuancu: It works. This case is covered by `testSortWithAggregationExplain`. The query `source=opensearch-sql_test_index_account| stats avg(age) AS avg_age by state, city | rename avg_age as new_avg_age | sort new_avg_age` won't be pushed down:

```
EnumerableSort(sort0=[$0], dir0=[ASC])
  EnumerableCalc(expr#0..2=[{inputs}], new_avg_age=[$t2], state=[$t0], city=[$t1])
    CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[state, city, age], AGGREGATION->rel#427:LogicalAggregate.NONE.[](input=RelSubset#426,group={0, 1},avg_age=AVG($2))], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","_source":{"includes":["state","city","age"],"excludes":[]},"aggregations":{"composite_buckets":{"composite":{"size":1000,"sources":[{"state":{"terms":{"field":"state.keyword","missing_bucket":true,"missing_order":"first","order":"asc"}}},{"city":{"terms":{"field":"city.keyword","missing_bucket":true,"missing_order":"first","order":"asc"}}}]},"aggregations":{"avg_age":{"avg":{"field":"age"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```

When sorting `avg_age`, the field names by `getRowType().getFieldNames()` is already the renamed ones (`state`, `city`, `avg_age`).

In comparison: `source=opensearch-sql_test_index_account| stats avg(age) AS avg_age by state, city | rename avg_age as new_avg_age | sort city` will push down sort.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
yuancu: Fixed

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
yuancu: Yes, in v2 it merges collations by simply concatenating them without de-duplication and re-order.

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
yuancu: Yes. Added a `(v2)` in comment.

yuancu:
> @yuancu do we have follow cases in ExplainIT
> 
> ```
> source=t | head 10 | sort field // pushdown limit only
> source=t | sort field | head 10 // pushdown sort only or pushdown sort + limit (depends on the execution order of DSL)
> ```

Added the test cases & fixed the behavior.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
qianheng-aws: The description may be a little confusing. Please change to like:

`Project(age) - TableScan(schema=[name, age], collation=[$1 ASC]` should became `TableScan(schema=[age], collation=[$0 ASC]` after project push down.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -189,4 +246,155 @@ public CalciteLogicalIndexScan pushDownLimit(Integer limit, Integer offset) {
...
+
+  /**
+   * Merges existing and new collations, ensuring that the last occurrence of each field index takes
+   * precedence.
+   *
+   * @param existingCollations Existing collation list.
+   * @param newCollations New collation list to be merged.
+   * @return Merged list of collations.
+   */
+  private static List<RelFieldCollation> mergeCollations(
```
qianheng-aws: Is it correct to merge 2 sorts? Thinking about this case `...| sort c | sort a, b |...`, I think the correct collation should be `[a ASC, b ASC]` and the collation for `c` is not guaranteed. So I think the right behavior should `override` instead of `merge`.

yuancu: That makes sense. I checked with popular databases like MySql and Postgres, the oder of previous sorts are not guaranteed.

E.g. in postgres
```
> SELECT * FROM (SELECT * FROM numbers ORDER BY double_col) AS sub ORDER BY integer_col; 

 double_col | integer_col 
------------+-------------
        0.5 |           1
        1.2 |           1
        5.5 |           1
        3.3 |           2
        0.8 |           2
        1.1 |           2
        1.9 |           3
        2.8 |           3
```

3.3. comes above 0.8 and 1.1

But I'm not sure what customers expect. @LantaoJin @penghuo What's your opinion?

LantaoJin: +1 for overriding IMO.

LantaoJin: It brings me what if `...| sort c | head 100 | sort a, b |...`?
in this case the `sort c` should be pushed down instead of `sort a, b`. But the pushdown of `...| sort c | fields a,b,c | sort a, b |...` requires overriding.

yuancu: The current implementation handles this correctly:

```
> source=bank | sort firstname | head 10 | sort gender, account_number | fields firstname, gender, account_number

EnumerableCalc(expr#0..2=[{inputs}], firstname=[$t1], gender=[$t2], account_number=[$t0])
  EnumerableSort(sort0=[$2], sort1=[$0], dir0=[ASC], dir1=[ASC])
    CalciteEnumerableIndexScan(table=[[OpenSearch, bank]], PushDownContext=[[SORT->[{
  "firstname" : {
    "order" : "asc",
    "missing" : "_last"
  }
}], LIMIT->10, PROJECT->[account_number, firstname, gender]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":10,"timeout":"1m","_source":{"includes":["account_number","firstname","gender"],"excludes":[]},"sort":[{"firstname":{"order":"asc","missing":"_last"}}]}, requestedTotalSize=10, pageSize=null, startFrom=0)])
```

Sorting `firstname` is pushed down, while `gender` and `account_number` are not.

yuancu: Fixed with overriding.

LantaoJin (approved):

LantaoJin:
@yuancu could you manually backport it to 2.19-dev?

yuancu:
@LantaoJin No problem

One issue bothers me is the handling of null values.

The doc states explicity: *Default: ascending order and NULL/MISSING first*. There's also a doctest for this case:
```
os> source=accounts | sort employer | fields employer;
fetched rows / total rows = 4/4
+----------+
| employer |
|----------|
| null     |
| Netagy   |
| Pyrami   |
| Quility  |
+----------+
```

But if I only modify it with swap calcite's default by doing the following during pushdown:
```
case FIRST -> "_last";
case LAST -> "_first";
```

I'm worried that it will still be *asc + null last* without push-down.

Btw, ChatGPT told me the behaviors of other databases (I didn't validate it myself):

| Database   | ASC Default   | DESC Default  |
|------------|---------------|---------------|
| Postgres   | NULLs FIRST   | NULLs LAST    |
| MySQL      | NULLs FIRST   | NULLs LAST    |
| SQL Server | NULLs FIRST   | NULLs LAST    |
| Oracle     | NULLs LAST    | NULLs FIRST   |

Maybe the existing behavior of defaulting to ascending with null first makes sense:  considering null as less than any actual value.

**PR #3835 is raised to fix this**

Issue #3595, opened 2025-04-30
[BUG] PPL Lookup return incorrect join result


penghuo:
**What is the bug?**
Change `query.size_limit` to 1, PPL lookup command return incorrect result.

**How can one reproduce the bug?**
* Prepare data
```
###
POST {{baseUrl}}/_bulk/
Content-Type: application/x-ndjson

{"index": {"_index": "logs"}}
{"regionId": "1", "action": "login_attempt", "timestamp": "2024-04-29T10:00:00Z"}
{"index": {"_index": "logs"}}
{"regionId": "2", "action": "file_upload", "timestamp": "2024-04-29T10:05:00Z"}

###
POST {{baseUrl}}/_bulk/
Content-Type: application/x-ndjson

{"index": {"_index": "region_info"}}
{"regionId": "0", "regionName": "eu"}
{"index": {"_index": "region_info"}}
{"regionId": "0", "regionName": "de"}
{"index": {"_index": "region_info"}}
{"regionId": "1", "regionName": "us-east-1"}
{"index": {"_index": "region_info"}}
{"regionId": "2", "regionName": "us-west-2"}

```
* Settings
```
###
PUT {{baseUrl}}/_plugins/_query/settings
Content-Type: application/x-ndjson

{"persistent" : {"plugins.query.size_limit" : "1"}}
```

* Query
```
###
POST {{baseUrl}}/_plugins/_ppl/
Content-Type: application/x-ndjson

{
    "query": "source=logs | lookup region_info regionId"
}
```
* Result
```
  "datarows": [
    [
      "login_attempt",
      "1",
      "2024-04-29 10:00:00",
      null
    ]
  ]
```

**What is the expected behavior?**
```
  "datarows": [
    [
      "login_attempt",
      "1",
      "2024-04-29 10:00:00",
      "us-east-1"                          <----- expect us-east-1
    ]
  ]
```

qianheng-aws:
When setting `query.size_limit` to 1, it only retrieves the first document for each table, and so the above PPL equals to 

`{"regionId": "1", "action": "login_attempt", "timestamp": "2024-04-29T10:00:00Z"}` 
join 
`{"regionId": "0", "regionName": "eu"}` 
on `regionId`

which cannot match actually. That's why the `regionName` is null.

penghuo:
Releated to https://github.com/opensearch-project/sql/issues/703

[...]

Issue #703, opened 2022-07-20
Extend size_limit setting in query engine to support unlimited index query.


penghuo:
## Problem statements
Currently, the query.size_limit [setting](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/admin/settings.rst#plugins-query-size-limit) configure the maximum amount of documents to be pull from OpenSearch. The default value is: 200. for example, Let's say size_limit = 200, and index has 10K docs.
* expected, return 200 rows.
`source=index`
* expected return 1 docs
`source=index | head 1`
* return 200 rows, it is not user expected.
`source=index | head 11000`

## Proposal
The query.size_limit configure the maximum amount of rows returned by query. The default value is: 200. size_limit must larger than 0. If the query has head(PPL) or limit(SQL). it will override the query.size_limit setting.

## Expectation of search query .
* should return 200 rows.
`source=index`
* should return 1 rows.
`source=index | head 1`
* should return 11000 rows.
`source=index | head 11000`

## Expectation of aggregation query.
* should return 200 rows.
`source=index | stats request, count(*) by request`
* should return 11000 rows.              
`source=index | stats request, count(*) by request | head 11000`

dai-chen:
Probably another way of thinking about this: the `size_limit` setting is just for default behavior. If users specify a larger number by `head` or `LIMIT`, that means they're aware of what they're doing and just want to override the default limit value. This may be safer than setting size limit to -1 and user run a query without head command later?

penghuo:
> Probably another way of thinking about this: the `size_limit` setting is just for default behavior. If users specify a larger number by `head` or `LIMIT`, that means they're aware of what they're doing and just want to override the default limit value. This may be safer than setting size limit to -1 and user run a query without head command later?

Update the proposal as discussed.

seankao-az:
## Design
### Physical Plan Generation
![Extend query size](https://user-images.githubusercontent.com/107227027/182191048-746b0959-8973-4153-a8e9-aeb0671fbc3a.jpg)
1. Analyzer
(1) When visiting the `Relation` node, fetch `index.max_result_window` and store it in [PlanContext](https://github.com/opensearch-project/sql/issues/656#issuecomment-1178062686)
(2) When visiting the `Limit` (SQL) or `Head` (PPL) node, decide whether scrolling is needed, and set `IndexScanType` in `PlanContext` accordingly
2. Optimizer
[...]

Issue #3607, opened 2025-05-06
[BUG] PIT/Scroll request incorrectly triggered for aggregation queries when size limit exceeds `max_result_window`


noCharger:
**What is the bug?**
When executing aggregation queries with `plugins.query.size_limit` set higher than `max_result_window` (default 10,000), the query planner incorrectly triggers a PIT/scroll request. This results in unpredictable results for aggregation queries.

**Root Cause**: The size limit check in the query planner doesn't distinguish between aggregation and non-aggregation queries, contrary to the pagination design doc (https://github.com/opensearch-project/sql/blob/main/docs/dev/Pagination-v2.md#scope).

**Code Reference**: https://github.com/opensearch-project/sql/pull/1666
```
      if (startFrom + size >= maxResultWindow) {
        sourceBuilder.size(maxResultWindow - startFrom);
        return new OpenSearchScrollRequest(
            indexName, cursorKeepAlive, sourceBuilder, exprValueFactory, includes);
```

**Impact**: Unpredictable results for aggregation queries when plugins.query.size_limit exceeds max_result_window.

**How can one reproduce the bug?**

Set plugins.query.size_limit to 20,000:
```
{
    "persistent": {
        "plugins.query.size_limit": "20000"
    }
}
```

Execute an aggregation query:
```
source=fluentd.services | where attrs.SERVICE_NAME='ratelimit' | stats count() as cnt
```

**What is the expected behavior?**
* PIT/Scroll request should not be triggered for aggregation queries
* Accurate number of datarows with the correct count should be returned

**Do you have any screenshots?**

Scroll query returns infinite loop of nulls
<img width="273" alt="Image" src="https://github.com/user-attachments/assets/663f4b05-e508-4fbe-be8b-a77c750f6df1" />

PIT query returns infinite loop of counts
<img width="1245" alt="Image" src="https://github.com/user-attachments/assets/d34450a5-8f5e-43c9-b6ae-4d72b4750a35" />

**Do you have any additional context?**
* Scroll is by default in 2.17
* PIT is by default in 2.19
* Scroll has been removed in 3.x https://github.com/opensearch-project/sql/pull/3346

PR #3623, opened 2025-05-14
Make query.size_limit only affect the final results

qianheng-aws:
### Description
Make query.size_limit only affect the final results

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3595, https://github.com/opensearch-project/sql/issues/703

### Check List

### DIFF
--- a/core/src/testFixtures/java/org/opensearch/sql/executor/DefaultExecutionEngine.java
+++ b/core/src/testFixtures/java/org/opensearch/sql/executor/DefaultExecutionEngine.java
@@ -28,7 +28,8 @@ public class DefaultExecutionEngine implements ExecutionEngine {
       context.getSplit().ifPresent(plan::add);
       plan.open();
 
-      while (plan.hasNext()) {
+      Integer querySizeLimit = context.getQuerySizeLimit();
+      while (plan.hasNext() && (querySizeLimit == null || result.size() < querySizeLimit)) {
         result.add(plan.next());
       }
       QueryResponse response =
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
@@ -203,10 +203,8 @@ public class OpenSearchIndex extends OpenSearchTable {
 
   @Override
   public TableScanBuilder createScanBuilder() {
-    final int querySizeLimit = settings.getSettingValue(Settings.Key.QUERY_SIZE_LIMIT);
-
     final TimeValue cursorKeepAlive = settings.getSettingValue(Settings.Key.SQL_CURSOR_KEEP_ALIVE);
-    var builder = new OpenSearchRequestBuilder(querySizeLimit, createExprValueFactory(), settings);
+    var builder = createRequestBuilder();
     Function<OpenSearchRequestBuilder, OpenSearchIndexScan> createScanOperator =
         requestBuilder ->
             new OpenSearchIndexScan(
@@ -265,12 +263,9 @@ public class OpenSearchIndex extends OpenSearchTable {
     return new AbstractEnumerable<>() {
       @Override
       public Enumerator<Object> enumerator() {
-        final int querySizeLimit = settings.getSettingValue(Settings.Key.QUERY_SIZE_LIMIT);
-
         final TimeValue cursorKeepAlive =
             settings.getSettingValue(Settings.Key.SQL_CURSOR_KEEP_ALIVE);
-        var builder =
-            new OpenSearchRequestBuilder(querySizeLimit, createExprValueFactory(), settings);
+        var builder = createRequestBuilder();
         return new OpenSearchIndexEnumerator(
             client,
             List.copyOf(getFieldTypes().keySet()),
@@ -281,10 +276,7 @@ public class OpenSearchIndex extends OpenSearchTable {
   }
 
   public OpenSearchRequestBuilder createRequestBuilder() {
-    return new OpenSearchRequestBuilder(
-        settings.getSettingValue(Settings.Key.QUERY_SIZE_LIMIT),
-        this.createExprValueFactory(),
-        settings);
+    return new OpenSearchRequestBuilder(getMaxResultWindow(), createExprValueFactory(), settings);
   }
 
   public OpenSearchRequest buildRequest(OpenSearchRequestBuilder requestBuilder) {
--- a/core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
@@ -31,6 +31,7 @@ public class CalcitePlanContext {
   public final ExtendedRexBuilder rexBuilder;
   public final FunctionProperties functionProperties;
   public final QueryType queryType;
+  public final Integer querySizeLimit;
 
   @Getter @Setter private boolean isResolvingJoinCondition = false;
   @Getter @Setter private boolean isResolvingSubquery = false;
@@ -46,8 +47,9 @@ public class CalcitePlanContext {
   private final Stack<RexCorrelVariable> correlVar = new Stack<>();
   private final Stack<List<RexNode>> windowPartitions = new Stack<>();
 
-  private CalcitePlanContext(FrameworkConfig config, QueryType queryType) {
+  private CalcitePlanContext(FrameworkConfig config, Integer querySizeLimit, QueryType queryType) {
     this.config = config;
+    this.querySizeLimit = querySizeLimit;
     this.queryType = queryType;
     this.connection = CalciteToolsHelper.connect(config, TYPE_FACTORY);
     this.relBuilder = CalciteToolsHelper.create(config, TYPE_FACTORY, connection);
@@ -84,7 +86,8 @@ public class CalcitePlanContext {
     }
   }
 
-  public static CalcitePlanContext create(FrameworkConfig config, QueryType queryType) {
-    return new CalcitePlanContext(config, queryType);
+  public static CalcitePlanContext create(
+      FrameworkConfig config, Integer querySizeLimit, QueryType queryType) {
+    return new CalcitePlanContext(config, querySizeLimit, queryType);
   }
 }
--- a/core/src/main/java/org/opensearch/sql/executor/ExecutionContext.java
+++ b/core/src/main/java/org/opensearch/sql/executor/ExecutionContext.java
@@ -12,16 +12,23 @@ import org.opensearch.sql.storage.split.Split;
 /** Execution context hold planning related information. */
 public class ExecutionContext {
   @Getter private final Optional<Split> split;
+  @Getter private final Integer querySizeLimit;
 
   public ExecutionContext(Split split) {
     this.split = Optional.of(split);
+    this.querySizeLimit = null;
   }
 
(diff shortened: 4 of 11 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
```diff
@@ -281,10 +276,7 @@ public Enumerator<Object> enumerator() {
   }
 
   public OpenSearchRequestBuilder createRequestBuilder() {
-    return new OpenSearchRequestBuilder(
-        settings.getSettingValue(Settings.Key.QUERY_SIZE_LIMIT),
-        this.createExprValueFactory(),
-        settings);
+    return new OpenSearchRequestBuilder(getMaxResultWindow(), createExprValueFactory(), settings);
```
penghuo, line 279: OpenSearchRequestBuilder.build() and buildRequestWithPit() already has maxResultWindow as paramater, do we still need in constructor? or consider remove it from build and buildRequestWithPit?

core/src/testFixtures/java/org/opensearch/sql/executor/DefaultExecutionEngine.java
```diff
@@ -28,7 +28,8 @@ public void execute(
       context.getSplit().ifPresent(plan::add);
       plan.open();
 
-      while (plan.hasNext()) {
+      Integer querySizeLimit = context.getQuerySizeLimit();
+      while (plan.hasNext() && (querySizeLimit == null || result.size() < querySizeLimit)) {
```
penghuo, line 32: should we add querySizeLimit in as limit RelNode, and leverage optimization and pushdown?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
noCharger: There's another bug around https://github.com/opensearch-project/sql/issues/3607. Seems can be fixed in this PR.

core/src/testFixtures/java/org/opensearch/sql/executor/DefaultExecutionEngine.java
qianheng-aws: That could be an option but it will change the plan. I've though of implementing it in that way, which should be simpler for production code and will introduce much related change to our test code.

And I found in Spark or Hive, their implement for fetch size have similar logic as this PR. It only affects the size of elements we take from the final results or iterator, instead of changing plan directly. Though I cannot tell what's other difference between these 2 options.

spark: https://github.com/apache/spark/blob/87d518680abbb45aa9ec3f21b5d1fd45ffc07e05/sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/SparkExecuteStatementOperation.scala#L117 

hive: https://github.com/apache/hive/blob/88cc6fddab918ecb188082ca30b6bada517d0d92/jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java#L392

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
qianheng-aws: > OpenSearchRequestBuilder.build() and buildRequestWithPit() already has maxResultWindow as paramater, do we still need in constructor? or consider remove it from build and buildRequestWithPit?

Both is required in the API, and they play as different roles.

The parameter in constructor is stored as field `requestedTotalSize` which decides `how many rows will be returned`, and it will be updated by the limit value in `LIMIT(limit, offset)`. In initialization, we expect `requestedTotalSize` should be set by `maxResultWindow`.

The parameter in build or buildRequestWithPit is actually the real place used to restrict the max window of index.
So it has logic `size = (startFrom + requestedTotalSize > maxResultWindow) ? maxResultWindow - startFrom) : requestedTotalSize`. https://github.com/opensearch-project/sql/blob/1329e0297a72c7ae0d679a2b5feddf5b53b65dde/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java#L111

qianheng-aws: Correct it, after looking into the issue https://github.com/opensearch-project/sql/issues/3607. Seems we should never pass in that parameter in the constructor to initialize `requestedTotalSize`. Without limit pushdown, we actually never know how many rows should be returned. It should be initialized with `-1`, which means no restriction on the total size.

core/src/testFixtures/java/org/opensearch/sql/executor/DefaultExecutionEngine.java
qianheng-aws: Although, in the current implementation, we cannot push down this query_size.limit into the scan, it won't retrieve all documents from the source actually. The real doc count should be `CEIL(query_size.limit/max_result_window)*max_result_window` which should be close to query_size.limit somehow, because when the result iterator reaches that count, it won't trigger the next fetch from source.

opensearch/src/test/java/org/opensearch/sql/opensearch/executor/protector/OpenSearchExecutionProtectorTest.java
```diff
@@ -99,7 +99,7 @@ public void setup() {
   void test_protect_indexScan() {
     String indexName = "test";
     final int maxResultWindow = 10000;
-    final int querySizeLimit = 200;
+    final int requestTotalSizeLimit = 200;
```
LantaoJin, line 102: remove this

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -59,7 +59,7 @@ public class OpenSearchRequestBuilder {
   private final SearchSourceBuilder sourceBuilder;
 
   /** Query size of the request -- how many rows will be returned. */
-  private int requestedTotalSize;
+  private int requestedTotalSize = -1;
```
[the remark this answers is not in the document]
LantaoJin: And let's add an error for query `select ... limit -1` or `search ... | head -1` to avoid bad case

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexScan.java
```diff
@@ -45,14 +45,18 @@ public class OpenSearchIndexScan extends TableScanOperator implements Serializab
...
       OpenSearchClient client, int maxResponseSize, OpenSearchRequest request) {
-    this.client = client;
     this.maxResponseSize = maxResponseSize;
+    this.client = client;
     this.request = request;
   }
 
+  /** Creates index scan based on a provided OpenSearchRequestBuilder. */
+  public OpenSearchIndexScan(OpenSearchClient client, OpenSearchRequest request) {
+    this(client, -1, request);
```
LantaoJin, line 57: just change `private int maxResponseSize;` to `private int maxResponseSize = Int.Max`; than we can delete setting `-1` and special check `maxResponseSize > 0`.

docs/user/optimization/optimization.rst
```diff
@@ -280,7 +280,7 @@ The Aggregation operator will merge into OpenSearch Aggregation::
           {
             "name": "OpenSearchIndexScan",
             "description": {
-              "request": "OpenSearchQueryRequest(indexName=accounts, sourceBuilder={\"from\":0,\"size\":10000,\"timeout\":\"1m\",\"aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"sources\":[{\"gender\":{\"terms\":{\"field\":\"gender.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}}]},\"aggregations\":{\"avg(age)\":{\"avg\":{\"field\":\"age\"}}}}}}, searchDone=false)"
+              "request": "OpenSearchQueryRequest(indexName=accounts, sourceBuilder={\"from\":0,\"size\":0,\"timeout\":\"1m\",\"aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"sources\":[{\"gender\":{\"terms\":{\"field\":\"gender.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}}]},\"aggregations\":{\"avg(age)\":{\"avg\":{\"field\":\"age\"}}}}}}, searchDone=false)"
```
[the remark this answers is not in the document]
qianheng-aws: For aggregation, size is meaningless here, so I don't set a new value to the size of sourceBuilder, then it keeps using the value when pushing down aggregation, which is 0.

https://github.com/opensearch-project/sql/blob/730f0665d970b2b5e40d3ba6ac51b17bf3174dd6/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java#L178

LantaoJin (dismissed):

docs/user/optimization/optimization.rst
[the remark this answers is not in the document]
penghuo: aggregation return both search hits and aggregation results. set size to 0, only return aggregation results.

penghuo (dismissed):

docs/user/optimization/optimization.rst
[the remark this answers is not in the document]
LantaoJin: @penghuo no, the benchmark was already set size=0. The performance issue is not related to this. But yes for a general improving for aggregation.

qianheng-aws:
> @qianheng-aws please update the doc https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/admin/settings.rst#plugins-query-size-limit
> 
> > The size configure the maximum amount of documents to be pull from OpenSearch. The default value is: 10000
> > Notes: This setting will impact the correctness of the aggregation operation, for example, there are 1000 docs in the index, if you change the value to 200, only 200 docs will be extract from index and do aggregation.
> 
> Sounds out of date.

done

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/3655.yml
```diff
@@ -86,21 +86,22 @@ teardown:
...
-
-  - do:
-      headers:
-        Content-Type: 'application/json'
-      ppl.explain:
-        body:
-          query: "source=test | where like(body, '%field 2%') | fields body"
-  - match: { root.children.0.description.request: "OpenSearchQueryRequest(indexName=test, sourceBuilder={\"from\":0,\"size\":10000,\"timeout\":\"1m\",\"query\":{\"wildcard\":{\"body\":{\"wildcard\":\"*field 2*\",\"case_insensitive\":true,\"boost\":1.0}}},\"_source\":{\"includes\":[\"body\"],\"excludes\":[]},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}]}, needClean=true, searchDone=false, pitId=null, cursorKeepAlive=null, searchAfter=null, searchResponse=null)"}
+# ---
+# TODO: pitId is dynamic, try to match the request body more precisely by using regex match
```
LantaoJin, line 90: could you open an issue for this temporary workaround?

LantaoJin (approved):

penghuo (approved):

PR #3628, opened 2025-05-16
Support function `coalesce` with Calcite

### LINKED ISSUE #3610: [FEATURE] Support function `coalesce`

Function | Result type | Description
-- | -- | --
coalesce(expression-list) | Expression | Returns the first non-null value from the list

https://github.com/opensearch-project/opensearch-spark/issues/519

yuancu:
### Description
Support functions coalesce with Calcite, which returns the first non-null value in its parameter list.
 
### Issues Resolved
Resolves #3610
 
### Check List


integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLConditionBuiltinFunctionIT.java
```diff
@@ -108,6 +108,21 @@ public void testIfNull() {
         rows("Hello", 30));
   }
 
+  @Test
+  public void testCoalesce() {
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | where age = 10 | eval new_country = coalesce(name, state, country) | fields name, state, country, new_country",
```
LantaoJin, line 116: could you add a case if the values of `name`, `state`, `country` are all null

docs/user/ppl/functions/condition.rst
```diff
@@ -238,3 +238,29 @@ Example::
...
+
+Usage: coalesce(field1, field2, ...) return the first non-null value in the argument list.
+
+Argument type: all the supported data type, (NOTE : if parameters have different type, you will fail semantic check)
+
+Return type: any
+
+Example::
+
+    os> source=accounts | eval result = coalesce(employer, firstname, lastname) | fields result, firstname, lastname, employer
```
LantaoJin, line 257: Change the `os>` to `PPL` to disable it in current doctest since this is only working with Calcite enabled.

docs/user/ppl/functions/condition.rst
```diff
@@ -238,3 +238,29 @@ Example::
...
+--------
+
+Description
+>>>>>>>>>>>
+
+Version: 3.1.0
+
+Usage: coalesce(field1, field2, ...) return the first non-null value in the argument list.
+
+Argument type: all the supported data type, (NOTE : if parameters have different type, you will fail semantic check)
```
LantaoJin, line 251: `(NOTE : if parameters have different type, you will fail semantic check)`
Change to the `The data types of all fields must be same.`

yuancu: Done

docs/user/ppl/functions/condition.rst
yuancu: Done

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLConditionBuiltinFunctionIT.java
yuancu: I added a cased of `coalesce(name, state, name) = null` to avoid injecting another row. Does this suffice?

docs/user/ppl/functions/condition.rst
```diff
@@ -238,3 +238,29 @@ Example::
...
+--------
+
+Description
+>>>>>>>>>>>
+
+Version: 3.1.0
+
+Usage: coalesce(field1, field2, ...) return the first non-null value in the argument list.
+
+Argument type: all the supported data type. The data types of all fields must be same.
```
dai-chen, line 251: Is all argument type same enforced by our language or Calcite?

yuancu: By both.
- In Calcite,  qSqlStdOperatorTable.COALESCE` uses the operand type checker [SameOperandTypeChecker](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/type/SameOperandTypeChecker.java), which effective requires all the values to be of the same or comparable types.
- In PPL-on-OpenSearch v2, we already have a similar function `IFNULL`, which returns the first non-null value between the two operands. If I try to feed 2 operands with different types, I will get the following error:
```
> source=country | eval a = ifnull(name, age) | fields a
```
```json
{
  "error": {
    "reason": "Invalid Query",
    "details": "ifnull function expected {[BYTE,BYTE],[SHORT,SHORT],[INTEGER,INTEGER],[LONG,LONG],[FLOAT,FLOAT],[DOUBLE,DOUBLE],[STRING,STRING],[BOOLEAN,BOOLEAN],[DATE,DATE],[TIME,TIME],[DATETIME,DATETIME],[TIMESTAMP,TIMESTAMP],[INTERVAL,INTERVAL],[STRUCT,STRUCT],[ARRAY,ARRAY]}, but get [STRING,LONG]",
  },
  "status": 400
}
```

Therefore, I think we also expect `COALESCE` to receive arguments of the same types. The limitation might come from the underlying implementation: the return type has to be inferrable from the operand types.

dai-chen (approved):

LantaoJin (approved):

anasalkouz:
Is this merged to 3.1?

yuancu:
> Is this merged to 3.1?
> 

Yes!

PR #3636, opened 2025-05-19
Mirroring all the content from /plugin/opensearch-sql-plugin/ to /direct-query

ahkcs:
Mirroring all the content from /plugin/opensearch-sql-plugin/ to /direct-query

### DIFF
--- a/.github/workflows/maven-publish.yml
+++ b/.github/workflows/maven-publish.yml
@@ -7,12 +7,13 @@ on:
       - main
       - 1.*
       - 2.*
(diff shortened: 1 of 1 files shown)


noCharger (commented):
Let's only publish async-query-core jar, not everything from sql. Also let's create poc version.

dai-chen (commented):
Could you elaborate what does this PR do? Especially, the "Mirroring" approach in PR description.

Swiddis (approved):

PR #3660, opened 2025-05-23
Fix: Long IN-lists causes crash

### LINKED ISSUE #1469: [BUG] Long IN value list cause stack overflow exception
**What is the bug?**

`StackOverflowError` occurs and OpenSearch JVM exits if the value list in IN expression is very long.

```
$ head big-in-list.txt
{ "query": "select * from opensearch_dashboards_sample_data_flights where FlightDelayMin in (1, 2, 3 ...)" }

$ curl -X POST "localhost:9200/_plugins/_sql" -H 'Content-Type: application/json' --data "@big-in-list.txt"
curl: (52) Empty reply from server
```

OpenSearch log:

```
java.lang.StackOverflowError
        ...
	at org.opensearch.sql.data.type.WideningTypeRule.distance(WideningTypeRule.java:51)
	at org.opensearch.sql.data.type.WideningTypeRule.lambda$distance$0(WideningTypeRule.java:50)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.reduce(ReferencePipeline.java:558)
	at org.opensearch.sql.data.type.WideningTypeRule.distance(WideningTypeRule.java:51)
	at org.opensearch.sql.data.type.WideningTypeRule.distance(WideningTypeRule.java:40)
	at org.opensearch.sql.expression.function.FunctionSignature.match(FunctionSignature.java:53)
	at org.opensearch.sql.expression.function.DefaultFunctionResolver.resolve(DefaultFunctionResolver.java:49)
	at org.opensearch.sql.expression.function.BuiltinFunctionRepository.getFunctionBuilder(BuiltinFunctionRepository.java:177)
	at org.opensearch.sql.expression.function.BuiltinFunctionRepository.resolve(BuiltinFunctionRepository.java:159)
	at org.opensearch.sql.expression.function.BuiltinFunctionRepository.compile(BuiltinFunctionRepository.java:135)
	at org.opensearch.sql.expression.function.BuiltinFunctionRepository.compile(BuiltinFunctionRepository.java:119)
	at org.opensearch.sql.analysis.ExpressionAnalyzer.visitCompare(ExpressionAnalyzer.java:234)
	at org.opensearch.sql.analysis.ExpressionAnalyzer.visitIn(ExpressionAnalyzer.java:221)
	at org.opensearch.sql.analysis.ExpressionAnalyzer.visitIn(ExpressionAnalyzer.java:222)
	at org.opensearch.sql.analysis.ExpressionAnalyzer.visitIn(ExpressionAnalyzer.java:222)
	at org.opensearch.sql.analysis.ExpressionAnalyzer.visitIn(ExpressionAnalyzer.java:222)
	at org.opensearch.sql.analysis.ExpressionAnalyzer.visitIn(ExpressionAnalyzer.java:222)
	at org.opensearch.sql.analysis.ExpressionAnalyzer.visitIn(ExpressionAnalyzer.java:222)
```

The query can return correct results as expected or terminate properly at compile time.

 - Plugins: latest code in `main` branch

### Root Cause

The root cause I found is IN expression is compiled recursively into comparisons connected by OR. Each value in list cause the recursive call and stack go deeper. Code: https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java#L222

```
  private Expression visitIn(
      UnresolvedExpression field, List<UnresolvedExpression> valueList, AnalysisContext context) {
    if (valueList.size() == 1) {
      return visitCompare(new Compare("=", field, valueList.get(0)), context);
    } else if (valueList.size() > 1) {
      return DSL.or(
          visitCompare(new Compare("=", field, valueList.get(0)), context),
          visitIn(field, valueList.subList(1, valueList.size()), context));
    } else {
      throw new SemanticCheckException("Values in In clause should not be empty");
    }
  }
```

### Solutions

Options to solve this problem include:

1. Add value list length check and document the limitation
2. Rewrite the recursion in `visitIn` function above to for loop
3. Compile IN expression to `field = ANY(ARRAY(values))` which maybe optimal at both compile and runtime

Swiddis:
### Description
Per #1469, very large lists of expressions in IN clauses cause `StackOverflowError`s. Avoiding all recursion here is nontrivial because `DSL.or` can only take two arguments. This PR instead rewrites the method to have the recursion amount grow logarithmically, by building the `or` tree as a balanced tree instead. This avoids hitting any recursion limits unless there's well over 2²⁵⁶ items in the list.

Implementing something like `DSL.any` is on the table for future work if we need to optimize further, but it's more complex than this fix and it's not clear that this is a hot path.

### Related Issues
Resolves #1469

### Check List
- [X] New functionality has been documented.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java
+++ b/core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java
@@ -292,15 +292,38 @@ public class ExpressionAnalyzer extends AbstractNodeVisitor<Expression, Analysis
 
   private Expression visitIn(
       UnresolvedExpression field, List<UnresolvedExpression> valueList, AnalysisContext context) {
-    if (valueList.size() == 1) {
-      return visitCompare(new Compare("=", field, valueList.get(0)), context);
-    } else if (valueList.size() > 1) {
-      return DSL.or(
-          visitCompare(new Compare("=", field, valueList.get(0)), context),
-          visitIn(field, valueList.subList(1, valueList.size()), context));
-    } else {
+    if (valueList.isEmpty()) {
       throw new SemanticCheckException("Values in In clause should not be empty");
     }
+
+    Expression[] expressions = new Expression[valueList.size()];
+
+    for (int i = 0; i < expressions.length; i++) {
+      expressions[i] = visitCompare(new Compare("=", field, valueList.get(i)), context);
+    }
+
+    return buildOrTree(expressions, 0, expressions.length);
+  }
+
+  /**
+   * `DSL.or` can only take two arguments. To represent large lists without massive recursion, we want to represent the
+   * expression as a balanced tree. This builds that tree from a node list.
+   *
+   *
+   * @param children The list of expressions to merge.
+   * @param start The starting position (inclusive) for the current combination step.
+   * @param end The ending position (exclusive) for the current combination step. MUST be greater-than start.
+   * @return The final `DSL.or` expression.
+   */
+  private Expression buildOrTree(Expression[] children, int start, int end) {
+    if (end - start == 1) {
+      return children[start];
+    }
+    if (end - start == 2) {
+      return DSL.or(children[start], children[end - 1]);
+    }
+    int split = start + (end - start) / 2;
+    return DSL.or(buildOrTree(children, start, split), buildOrTree(children, split, end));
   }
 
   @Override
--- a/core/src/test/java/org/opensearch/sql/analysis/ExpressionAnalyzerTest.java
+++ b/core/src/test/java/org/opensearch/sql/analysis/ExpressionAnalyzerTest.java
@@ -27,6 +27,8 @@ import static org.opensearch.sql.data.type.ExprCoreType.STRUCT;
 import static org.opensearch.sql.expression.DSL.ref;
 
 import com.google.common.collect.ImmutableMap;
+
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


dai-chen (commented):
Thanks for the fix. In original issue, one of the possible solution is

> "Compile IN expression to field = ANY(ARRAY(values)) which maybe optimal at both compile and runtime"

Could you compare it with current approach?

LantaoJin:
Will this improving be a throw away work in future? This issue only happens in v2. And with calcite, the query `| where age in (10, 40.0, 100)` will pushdown as
```
  "query": {
    "terms": {
      "age": [
        10,
        40,
        100
      ],
      "boost": 1
    }
  }
```
instead of following in v2:
```
  "query": {
    "bool": {
      "should": [
        {
          "term": {
            "age": {
              "value": 10,
              "boost": 1
            }
          }
        },
        {
          "bool": {
            "should": [
              {
                "term": {
                  "age": {
                    "value": 40,
                    "boost": 1
                  }
                }
              },
              {
                "term": {
                  "age": {
                    "value": 100,
                    "boost": 1
                  }
                }
              }
            ],
            "adjust_pure_negative": true,
            "boost": 1
          }
        }
      ],
      "adjust_pure_negative": true,
      "boost": 1
    }
  }
```

dai-chen (approved):

Issue #3619, opened 2025-05-13
[RFC] Support DECIMAL data type in SQL/PPL


LantaoJin:
**Problem Statement**
Since OpenSearch Core doesn't have a `Decimal` type, ref [link](https://docs.opensearch.org/docs/latest/field-types/supported-field-types/numeric/), SQL and PPL correspondingly don't have `Decimal` types either. 

**Current State**
For decimal number processing, `Double` or `Float` types are used. Both `Double` and `Float` have precision loss issues. For example, https://github.com/opensearch-project/sql/issues/3614 mentioned an issue about precision loss in `Double`

**Long-Term Goals**
* Support `Decimal` type in SQL/PPL
* Support `DecimalLiteral` and set numeric literal to decimal literal as default, e.g. 0.06 is decimal literal.
* Support type suffix to specify numeric literal, e.g. 0.06D is double literal, 0.06F is float literal, 0.06db is decimal literal.
* Support auto implicitly cast (least restrictive type) for operations, e.g. decimal(3.2) + decimal(3.3) = decimal(3.3), double + decimal = double, etc.

**Proposal**
As PPL/SQL serves as a unified and general-purpose query language, it shouldn't be completely bound to the underlying engine and should maintain certain cross-platform compatibility, as we've already supported PPL queries on Spark. To better support cross-engine query requirements and avoid `Double` precision loss issues"

**Approach**
- Add a new Data Type `Decimal`
- Update g4 files to support type suffix literal
- Support auto implicitly cast (least restrictive type) for operations in Calcite

**Alternative**
This is two alternative:
1. Continue as-is.
2. To fix the literal issue only https://github.com/opensearch-project/sql/issues/3614, we can just treat double literal as decimal literal.

**Implementation Discussion**
Here is some reference about how Spark handle decimal:
```
spark-sql (default)> select 0.06 - 0.01; -- default numeric literal is decimal literal
(0.06 - 0.01)
0.05
spark-sql (default)> select typeof(0.06 - 0.01);
typeof((0.06 - 0.01))
decimal(3,2)

spark-sql (default)> select 0.06D - 0.01D; -- support type suffix
(0.06 - 0.01)
0.049999999999999996
spark-sql (default)> select typeof(0.06D - 0.01D);
typeof((0.06 - 0.01))
double

spark-sql (default)> select 0.06 - 0.01D; -- support implicitly casting
(0.06 - 0.01)
0.04999999999999999
```

LantaoJin:
[...]

PR #3673, opened 2025-05-26
Support decimal literal with Calcite

### LINKED ISSUE #3614: [BUG]  Numeric literal precision is inconsistent with Spark/MySQL
**What is the bug?**
In SQL or PPL (both v2 and v3)
```
POST /_plugins/_ppl
{
  "query" : """
source = people | eval a = 0.06 -0.01 | head 1
  """
}
POST /_plugins/_sql
{
  "query" : """
select 0.06 - 0.01;
  """
}
```
returns 
```
{
  "schema": [
    {
      "name": "0.06 - 0.01",
      "type": "double"
    }
  ],
  "datarows": [
    [
      0.049999999999999996
    ]
  ],
  "total": 1,
  "size": 1,
  "status": 200
}
```

But Spark/MySQL/Progres return **0.05** since they use decimal literal instead of double literal.

It leads to inconsistency results between OpenSearch PPL and Spark PPL (also OpenSearch SQL and Spark SQL)

```
spark-sql (default)> select 0.06 - 0.01;
(0.06 - 0.01)
0.05
spark-sql (default)> select cast(0.06 as double) - cast(0.01 as double);
(CAST(0.06 AS DOUBLE) - CAST(0.01 AS DOUBLE))
0.049999999999999996
spark-sql (default)> select typeof(0.06 - 0.01);
typeof((0.06 - 0.01))
decimal(3,2)
```

Should return 0.05

LantaoJin:
### Description
Support decimal literal with Calcite

- 

### Related Issues
Resolves #3614 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/ast/expression/Literal.java
+++ b/core/src/main/java/org/opensearch/sql/ast/expression/Literal.java
@@ -6,10 +6,10 @@
 package org.opensearch.sql.ast.expression;
 
 import com.google.common.collect.ImmutableList;
+import java.math.BigDecimal;
 import java.util.List;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
-import lombok.RequiredArgsConstructor;
 import org.opensearch.sql.ast.AbstractNodeVisitor;
 
 /**
@@ -18,11 +18,19 @@ import org.opensearch.sql.ast.AbstractNodeVisitor;
  */
 @Getter
 @EqualsAndHashCode(callSuper = false)
-@RequiredArgsConstructor
 public class Literal extends UnresolvedExpression {
 
   private final Object value;
   private final DataType type;
+
+  public Literal(Object value, DataType dataType) {
+    if (dataType == DataType.DECIMAL && value instanceof Double) {
+      this.value = BigDecimal.valueOf((Double) value);
+    } else {
+      this.value = value;
+    }
+    this.type = dataType;
+  }
 
   @Override
   public List<UnresolvedExpression> getChild() {
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/MathematicalFunctionIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/MathematicalFunctionIT.java
@@ -40,7 +40,11 @@ public class MathematicalFunctionIT extends PPLIntegTestCase {
   public void testCeil() throws IOException {
     JSONObject result =
         executeQuery(String.format("source=%s | eval f = ceil(age) | fields f", TEST_INDEX_BANK));
-    verifySchema(result, schema("f", null, "bigint"));
+    if (isCalciteEnabled()) {
+      verifySchema(result, schema("f", null, "int"));
+    } else {
+      verifySchema(result, schema("f", null, "bigint"));
+    }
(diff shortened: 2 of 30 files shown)


core/src/main/java/org/opensearch/sql/ast/expression/Literal.java
```diff
@@ -18,12 +18,20 @@
...
 @Getter
 @EqualsAndHashCode(callSuper = false)
-@RequiredArgsConstructor
 public class Literal extends UnresolvedExpression {
 
   private final Object value;
   private final DataType type;
 
+  public Literal(Object value, DataType dataType) {
+    if (dataType == DataType.DECIMAL && value instanceof Double) {
```
qianheng-aws, line 27: When will we create a decimal literal with double value?

penghuo (approved):
Thanks @LantaoJin for this PR! 

This looks like a solid approach to fixing the decimal precision issue. I like how you've implemented decimal literals by having them represented as BigDecimal in Calcite while maintaining the current runtime type system (mapping to DOUBLE). This cleanly addresses the inconsistency between the behavior in OpenSearch SQL/PPL and other systems like Spark/MySQL for arithmetic operations like "0.06 - 0.01".

The test cases confirm it produces the expected precise results now, and the implementation is well-structured with a clear path forward for full decimal type support in the future (via issue #3619).

One minor suggestion would be to add a small note in the documentation about this change to help users understand the precision behavior, but that could also be addressed separately.

penghuo (approved):
Thanks for the change! One minor suggestion would be to add a small note in the documentation about this change to help users understand the precision behavior, but that could also be addressed separately.

core/src/main/java/org/opensearch/sql/ast/expression/Literal.java
LantaoJin: This is to be compatible with calls like `new Literal(2.0, Decimal)` , otherwise we can only use `new Literal(new BigDecimal("2.0"), Decimal)`

integ-test/src/test/java/org/opensearch/sql/ppl/MathematicalFunctionIT.java
```diff
@@ -40,7 +40,11 @@ public void testAbs() throws IOException {
   public void testCeil() throws IOException {
     JSONObject result =
         executeQuery(String.format("source=%s | eval f = ceil(age) | fields f", TEST_INDEX_BANK));
-    verifySchema(result, schema("f", null, "bigint"));
+    if (isCalciteEnabled()) {
+      verifySchema(result, schema("f", null, "int"));
+    } else {
+      verifySchema(result, schema("f", null, "bigint"));
+    }
```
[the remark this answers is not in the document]
LantaoJin: > Why dose this IT could pass before?

`CalciteMathematicalFunctionIT` was missing before this PR.

LantaoJin:
> Thanks for the change! One minor suggestion would be to add a small note in the documentation about this change to help users understand the precision behavior, but that could also be addressed separately.

Sure, I will summary the break changes for 3.1.0 release in a new PR.

LantaoJin:
> Thanks for the changes! Should we also update user manualhttps://github.com/opensearch-project/sql/blob/main/docs/user/ppl/general/datatypes.rst?

Decimal is not a new supported datatype, we'd better update that doc after #3619 addressed. I will highlight this breaking changes in release notes.

PR #3678, opened 2025-05-27
Support Grok command in Calcite engine

### LINKED ISSUE #3677: [FEATURE] Support Grok command for Calcite engine
Support Grok command for Calcite engine. Currently, we only support single parse method for Parse unresolvedPlan.

songkant-aws:
### Description
Support Grok command in Calcite engine

### Related Issues
Resolves #3677 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
@@ -232,6 +232,7 @@ public class PPLFuncImpTable {
       registerOperator(DIVIDE, PPLBuiltinOperators.DIVIDE);
       registerOperator(SHA2, PPLBuiltinOperators.SHA2);
       registerOperator(CIDRMATCH, PPLBuiltinOperators.CIDRMATCH);
+      registerOperator(INTERNAL_GROK, PPLBuiltinOperators.GROK);
 
       // Register PPL Datetime UDF operator
       registerOperator(TIMESTAMP, PPLBuiltinOperators.TIMESTAMP);
--- a/core/src/main/java/org/opensearch/sql/expression/function/udf/GrokFunction.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/udf/GrokFunction.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.expression.function.udf;
+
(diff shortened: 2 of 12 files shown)


core/src/main/java/org/opensearch/sql/expression/function/udf/GrokFunction.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  public static class GrokImplementor implements NotNullImplementor {
+    @Override
+    public Expression implement(
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+      return Expressions.call(GrokFunction.GrokImplementor.class, "grok", translatedOperands);
+    }
+
+    public static Map<String, String> grok(String input, String regex) {
```
dai-chen, line 54: Could you elaborate what's internal functions? I think we're considering this kind of static method as contract between Calcite and Spark and register it as Spark's UDF. Anything different for grok to get it work in Spark SQL?

songkant-aws: The underlying function grokCompiler and parsing logic is the same. So yes, it's true that they will share the same contract by leveraging this static method as UDF. 

The reason why it's treated as internal function is that PPL syntax didn't expose `grok` as user-facing function signature but as a standalone command. We discussed this `INTERNAL_${FUNC_NAME}` naming with @LantaoJin . For those parse related commands, I think they are just syntax sugar for functions which could be called inside Project.

LantaoJin (approved):

core/src/main/java/org/opensearch/sql/expression/function/udf/GrokFunction.java
dai-chen: Got it. Thanks!

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -235,6 +235,7 @@ void populate() {
       registerOperator(IF, SqlStdOperatorTable.CASE);
       registerOperator(IFNULL, SqlStdOperatorTable.COALESCE);
       registerOperator(COALESCE, SqlStdOperatorTable.COALESCE);
+      registerOperator(INTERNAL_ITEM, SqlStdOperatorTable.ITEM);
```
dai-chen, line 238: Calcite ITEM function is public right? Why we need another internal one?

dai-chen (approved):
Thanks for the changes!

PR #3679, opened 2025-05-27
Return double with correct precision for `UNIX_TIMESTAMP`

### LINKED ISSUE #3611: [BUG] Return type of `UNIX_TIMESTAMP` is double but still precision missing
Whatever in v2 and v3, the function `UNIX_TIMESTAMP` return Double.
But the precision is still missing for micro seconds
```
POST /_plugins/_sql
{
  "query" : """
select unix_timestamp('2008-09-01 06:12:27.123456')
  """ 
}
```
return 1220227200.

In MySQL, it return Double with scalar
```
mysql> SELECT UNIX_TIMESTAMP('2015-11-13 10:20:19.012');
        -> 1447431619.012
```

This behavior is different with SparkSQL.

Change the return type to Long

Support precision 

https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_unix-timestamp
https://spark.apache.org/docs/latest/api/sql/#unix_timestamp

yuancu:
### Description
When `UNIX_TIMESTAMP` is fed with a timestamp string, it could not return a correct double value since the string is implicitly cast to date before feeding to the function.

> Deprecated solution:
> I corrected this behavior by adding STRING as an acceptable type to `UNIX_TIMESTAMP` and adding corresponding handlers for strings. It is worth noting that timestamp string is mentioned as an acceptable argument in [the documentation](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/datetime.rst#unix_timestamp):
> The date argument may be a DATE, or TIMESTAMP string
> Therefore, The change to the function signature should be appropriate.

I reordered the signatures of `UNIX_TIMESTAMP` so that `unix_timestamp(timestamp)` will be prioritized when the input is a string. In such case, the string will be cast to a timestamp. The reordering is valid since this coercion should also be compatible with date and time types and will result in correct values for these types.

### Related Issues
Resolves #3611 

### Check List


LantaoJin:
> Just wondering how did we decide to fix these cases one by one instead of adding an implicit conversion from String to Timestamp?

+1. @yuancu I have mentioned it multiple times, could you make it happen before 3.1.0? We'd better not to extend original function definitions.

LantaoJin:
> > > Just wondering how did we decide to fix these cases one by one instead of adding an implicit conversion from String to Timestamp?
> > 
> > 
> > +1. @yuancu I have mentioned it multiple times, could you make it happen before 3.1.0? We'd better not to extend original function definitions.
> 
> Do you mean when we find a type mismatch, we coerce it to timestamp instead of date?

It depends on what the format of string is. `2020-01-01` or `2020-01-01 00:00:00`

yuancu:
Ok I get it. I'll try to fix it.

Update: The functions are resolved based on the distance between input argument types and expected argument types.
`unix_timestamp` expects date / timestamp / double or no input.
When the input is of string type, its distance to date / timestamp are the same. It picks the definition that is defined earlier. For now, we prioritize `unix_timestamp(timestamp)` over `unix_timestamp(date)` as timestamp allows a wider range of inputs -- date / time / timestamp string can be cast to timestamp.

TODO:
- [ ] Implement literal cast: e.g. literal "2001-10-10" can be implicitly cast to a date, "2001-10-10 13:29:09" can be implicitly cast to a timestamp.
- [ ] Allow casting date/time string to timestamp  #3728

LantaoJin (dismissed):

LantaoJin:
@yuancu could your resolve the conflicts?

yuancu:
> @yuancu could your resolve the conflicts?

No problem.

LantaoJin (dismissed):
LGTM, waiting CI pass

LantaoJin (approved):

qianheng-aws (approved):

qianheng-aws:
@yuancu  Do we have issue to track supporting implicit type coercion in v3? If not, please create a one and attach the ignored test there.

yuancu:
> @yuancu Do we have issue to track supporting implicit type coercion in v3? If not, please create a one and attach the ignored test there.

#3761 is created.

Issue #2813, opened 2024-07-10
[WIP][FEATURE] Support nested aggregation


LantaoJin:
**Is your feature request related to a problem?**
A clear and concise description of what the problem is, e.g. _I'm always frustrated when [...]_

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

alexey-temnikov:
## Additional Bug Report: Aggregation Operations Return Null

Created #4564 to track a specific manifestation of this issue where PPL aggregation operations (`stats`, `top`) return `null` for nested field values.

**Key findings from #4564:**
- `fields` command works correctly on nested fields 
- `stats COUNT() by nested.field` returns null (incorrect)
- `top nested.field` returns null (incorrect)

LantaoJin:
Nested aggregation for SQL is not in plan

PR #2814, opened 2024-07-10
Support nested aggregation

LantaoJin:
### Description
Support nested aggregation. With this PR, follow SQL are able to exectue a nested aggregation query.
```SQL
SELECT min(nested(pages.load_time, pages)) FROM logs WHERE response = 200;
```
And it equals the DSL
```
GET logs/_search
{
  "query": {
    "match": { "response": "200" }
  },
  "aggs": {
    "pages": {
      "nested": {
        "path": "pages"
      },
      "aggs": {
        "min_load_time": { "min": { "field": "pages.load_time" } }
      }
    }
  }
}
```
 
### Issues Resolved
PR resolves https://github.com/opensearch-project/sql/issues/2813 and https://github.com/opensearch-project/sql/issues/2529, and partly resolves https://github.com/opensearch-project/sql/issues/2739 since nested function hasn't supported in PPL yet.
 
### Check List
- [x] New functionality includes testing.
  - [x] All tests pass, including unit test, integration test and doctest
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
  - [x] New functionality has user manual doc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #2739, opened 2024-06-11
[BUG] `fields` is changing the outcome of the following commands that depend on selected fields


TackAdam:
**What is the bug?**
When running `stats count()` operation on a nested parameter the result seems to just return 0. If the field is specified before running the count, it will show the correct value. 

**How can one reproduce the bug?**
1. Go to 'OpenSearch Playground:Query Workbench https://playground.opensearch.org/app/opensearch-query-workbench#/
2. Click on PPL and enter the PPL "source = jaeger-span-* |stats count (references.refType)" and hit Run
<img width="809" alt="Screenshot 2024-06-11 at 4 34 36 PM" src="https://github.com/opensearch-project/sql/assets/105462877/e56514dd-b910-4fba-9760-0db85db56b8f">
-Output will be 0 

Enter the PPL "source = jaeger-span-* | fields references.refType|stats count (references.refType)" and hit Run
<img width="992" alt="Screenshot 2024-06-11 at 4 36 22 PM" src="https://github.com/opensearch-project/sql/assets/105462877/d71e5e88-94e6-4ca9-b37a-0a1b1f67cab0">
-Output will be 193

**What is the expected behavior?**
The `stats count()` should return a consistent value regardless of any preceeding field selection.

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

LantaoJin:
This may be not a bug of SQL plugin. Regarding to nested field, OpenSearch DSL query will return `0` value too. Try below DSL
```
GET jaeger-span-*/_search
{
  "size": 0,
  "aggs": {
    "count_refType": {
      "value_count": {
        "field": "references.refType"
      }
    }
  }
}
```
PPL `source = jaeger-span-* | stats count (references.refType)` equals to above DSL query. But `source = jaeger-span-* | fields references.refType | stats count (references.refType)` equals searching keyword `patients.name` first then aggregate.

LantaoJin:
Open https://github.com/opensearch-project/OpenSearch/issues/14347 to track

dblock:
Is this fixed with the related OpenSearch issue @LantaoJin?

[Catch All Triage - Attendees [1](https://github.com/dblock/), [2](https://github.com/krisfreedain), [3](https://github.com/andrross), [4](https://github.com/gkamat), [5](https://github.com/getsaurabh02)]

LantaoJin:
[...]

Issue #2529, opened 2024-02-24
[BUG] Incorrect value of quotted nested fields aggregations in JSON export


smortex:
**What is the bug?**

When quoting the field passed to an aggregation and when this field is a nested field, a SQL query with the JSON format return an incorrect value (0 or null) for the aggregation instead of the actual result.

**How can one reproduce the bug?**

Given an index with documents with a nested (`field_nested.c2`) and a non-nested field (`field_float`), using the following two queries to perform some basic aggregations:

* ```SELECT sum(`field_float`) AS `SUM(field_float)`, sum(`field_nested.c2`) AS `SUM(field_nested.c2)` FROM data1 LIMIT 5000;```
* ```SELECT sum(field_float) AS `SUM(field_float)`, sum(field_nested.c2) AS `SUM(field_nested.c2)` FROM data1 LIMIT 5000;```

We expect the same result.  But when requesting JSON documents, the first query produce wrong values:

First query result:
```
{    
    "took": 1,    
    "timed_out": false,    
    "_shards": {    
        "total": 1,    
        "successful": 1,    
        "skipped": 0,    
        "failed": 0    
    },    
    "hits": {    
        "total": {    
            "value": 3,    
            "relation": "eq"    
        },    
        "max_score": null,    
        "hits": []    
    },    
    "aggregations": {    
        "SUM(field_nested.c2)": {    
            "value": 0    
        },    
        "SUM(field_float)": {    
            "value": 0.6000000163912773    
        }    
    }    
} 
```

Second query result:
```
{    
    "took": 1,    
    "timed_out": false,    
    "_shards": {    
        "total": 1,    
        "successful": 1,    
        "skipped": 0,    
        "failed": 0    
    },    
    "hits": {    
        "total": {    
            "value": 3,    
            "relation": "eq"    
        },    
        "max_score": null,    
        "hits": []    
    },    
    "aggregations": {    
        "SUM(field_nested.c2)": {    
            "value": 6    
        },    
        "SUM(field_float)": {    
            "value": 0.6000000163912773    
        }    
    }    
} 
```

Note how `aggregations."SUM(field_nested.c2)".value` is `0` in the first example and `6` it the second.

**What is the expected behavior?**

Quoted and non-quoted nested fields in aggregations should return the same (correct) value.

**What is your host/environment?**

 - OS: Debian
 - Version 11
 - OpenSearch 2.12.0

**Do you have any additional context?**

* For the bug to happen, the field must be nested *and* quoted;
[...]

PR #3696, opened 2025-05-29
Support nested aggregation when calcite enabled

LantaoJin:
### Description

Support [nested aggregation](https://docs.opensearch.org/docs/latest/aggregations/bucket/nested/) in PPL only when calcite enabled.

With this PR, follow PPL query is able to execute a nested aggregation query.
```SQL
source=logs | where response = 200 | stats min(pages.load_time)
```
And it equals the DSL
```
GET logs/_search
{
  "query": {
    "match": { "response": "200" }
  },
  "aggs": {
    "pages": {
      "nested": {
        "path": "pages"
      },
      "aggs": {
        "min_load_time": { "min": { "field": "pages.load_time" } }
      }
    }
  }
}
```

Deprecated implementation for SQL: https://github.com/opensearch-project/sql/pull/2814

**Limitation:**
- PPL only
- Calcite should be enabled
- Aggregate pushdown happens

### Related Issues
PR partially resolves https://github.com/opensearch-project/sql/issues/2813 and https://github.com/opensearch-project/sql/issues/2529, and https://github.com/opensearch-project/sql/issues/2739

### Check List

### DIFF
--- a/.github/get-sonatype-credentials.sh
+++ b/.github/get-sonatype-credentials.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+# Script to retrieve Sonatype credentials from AWS Secrets Manager
+
+SONATYPE_USERNAME=$(aws secretsmanager get-secret-value --secret-id maven-snapshots-username --query SecretString --output text)
+SONATYPE_PASSWORD=$(aws secretsmanager get-secret-value --secret-id maven-snapshots-password --query SecretString --output text)
+echo "::add-mask::$SONATYPE_USERNAME"
+echo "::add-mask::$SONATYPE_PASSWORD"
+echo "SONATYPE_USERNAME=$SONATYPE_USERNAME" >> $GITHUB_ENV
+echo "SONATYPE_PASSWORD=$SONATYPE_PASSWORD" >> $GITHUB_ENV
--- a/.github/workflows/maven-publish.yml
+++ b/.github/workflows/maven-publish.yml
@@ -8,6 +8,9 @@ on:
       - 1.*
       - 2.*
 
+env:
+  SNAPSHOT_REPO_URL: https://aws.oss.sonatype.org/content/repositories/snapshots/
+
 jobs:
   build-and-publish-snapshots:
     strategy:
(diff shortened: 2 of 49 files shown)


LantaoJin:
> QQ: I recall we didn't support nested field type in V2 engine, right? Did we add it to Calcite or this is only for pushdown? Ref: #52

Nested field maps to `ARRAY `type in OpenSearch Core (you can get it by creating an index with nested field and checking its mapping via DSL), so IMO v2 supports nested field with the `ExprCoreType.ARRAY` type.

I believe we also can support nested aggregation in v2 but the fixing will be complex by my trying. Now I have no idea how to pass the `fieldTypes` map to Aggregator logic. We need it to identify whether or not the `root` in aggregation NamedExpression `root.subfield` is a ARRAY type.

For pushdown disabled, I think we need additional effort to correct the aggregation behavior. So I add a TODO in code. Or find a way to throw an exception.

qianheng-aws:
The value of nested field in PPL is incorrect. 

Seems we will only show the first element of that nested fields while it's actually an inner field of an array of object. That's why it will calculate wrong results if no push down.
https://github.com/opensearch-project/sql/blob/df9f5dd117690cb7b83bada1ba504a1d556ec57d/core/src/main/java/org/opensearch/sql/data/model/ExprValueUtils.java#L225


DSL:
```
POST /test1/_search
{
  "_source":{"includes":["address.area"]}
}

{
  "took": 11,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 1,
      "relation": "eq"
    },
    "max_score": 1.0,
    "hits": [
      {
        "_index": "test1",
        "_id": "u6OQIJcBpwe8-qLMf_ZG",
        "_score": 1.0,
        "_source": {
          "address": [
            {
              "area": 300.13
            },
            {
              "area": 400.99
            },
            {
              "area": 127.4
            },
            {
              "area": 10.24
            }
          ]
        }
      }
    ]
  }
}
```

PPL:
```
POST /_plugins/_ppl
source = test1 | fields address.area

{
  "schema": [
    {
      "name": "address.area",
      "type": "double"
    }
  ],
  "datarows": [
    [
      300.13
    ]
  ],
  "total": 1,
  "size": 1
}
```

LantaoJin:
> Seems we will only show the first element of that nested fields while it's actually an inner field of an array of object. That's why it will calculate wrong results if no push down.
> 
> https://github.com/opensearch-project/sql/blob/df9f5dd117690cb7b83bada1ba504a1d556ec57d/core/src/main/java/org/opensearch/sql/data/model/ExprValueUtils.java#L225
> 

Thanks, please let me know why we just return first element? Let me try to change this logic first.

LantaoJin:
> > Seems we will only show the first element of that nested fields while it's actually an inner field of an array of object. That's why it will calculate wrong results if no push down.
> > https://github.com/opensearch-project/sql/blob/df9f5dd117690cb7b83bada1ba504a1d556ec57d/core/src/main/java/org/opensearch/sql/data/model/ExprValueUtils.java#L225
> 
> Thanks, please let me know why we just return first element? Let me try to change this logic first.

Perhaps I know the reason now. I update above code to
```
wholePathValue = new ExprCollectionValue(value.collectionValue().stream().map(e -> e.keyValue(paths.getFirst())).toList());
```
but it doesn't work since we flatten the schema (mapping): https://github.com/opensearch-project/sql/blob/5eb2e5ba9bb3cea40beacd2fae4d180480c9ed93/opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchDataType.java#L294-L296

For example `address.area` is mapping to `ExprCoreType.Double` and we don't have an `Array[Double]` type in ExprCoreType. To support reading a sub-field, we need support `Array` type of ExprCoreType.

Swiddis:
@LantaoJin what's the status of this?

LantaoJin:
> @LantaoJin what's the status of this?

Deprioritized it since we created a META ticket #4625  to discuss all cases for nested type as there is no user cases for customer F. I hope we can have a roadmap instead of keep the nested type as a Schrödinger's cat.

LantaoJin:
Close this as the refactor #4979 opened.

Issue #3102, opened 2024-10-23
[BUG] PPL head limit and SQL limit offset throw exception


penghuo:
**What is the bug?**
```
###
POST {{baseUrl}}/_plugins/_ppl
Content-Type: application/x-ndjson

{
  "query": "source = people"
}

###
POST {{baseUrl}}/_plugins/_ppl/_explain
Content-Type: application/x-ndjson

{
  "query": "source = people | head 1 from 1"
}
```
* error message
```
{
  "error": {
    "reason": "Invalid SQL query",
    "details": "Validation Failed: 1: using [from] is not allowed in a scroll context;2: [size] cannot be [0] in a scroll context;",
    "type": "ActionRequestValidationException"
  },
  "status": 400
}
```

**How can one reproduce the bug?**
```
###
PUT {{baseUrl}}/_plugins/_query/settings
Content-Type: application/x-ndjson

{
  "transient" : {
    "plugins.query.size_limit" : 1
  }
}

###
PUT {{baseUrl}}/people/_settings
Content-Type: application/x-ndjson

{
  "index": {
    "max_result_window": 1
  }
}

```

**What is the expected behavior?**
No exception.

dblock:
[[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/dblock), [2](https://github.com/reta), [3](https://github.com/andrross/), [4](https://github.com/dolla24)]

PR #3713, opened 2025-05-30
[BugFix] Prevent push down limit with offset reach maxResultWindow

qianheng-aws:
### Description
Prevent push down limit with offset reach maxResultWindow

The index setting of `max_result_window` is the maximum restriction of `from + size`.
https://docs.opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index-settings/

Thus in our code of building OpenSearch request, we have logic of using `size` equals to `maxResultWindow -statFrom` if `startFrom + size >  maxResultWindow`. It will have problem if we got new `size` equal to 0 or less than 0.

We could address this issue by preventing such limit from push down if it would cause the new `startFrom` to reach the maxResultWindow.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3102

### Check List

### DIFF
--- a/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/3102.yml
+++ b/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/3102.yml
@@ -20,18 +20,8 @@ setup:
           - '{"index": {}}'
           - '{"id": 3}'
 
-
 ---
-teardown:
-  - do:
-      query.settings:
-        body:
-          transient:
-            plugins.calcite.enabled : false
-            plugins.calcite.fallback.allowed : true
-
----
-"Prevent push down limit if the offset exceeds max_result_window":
+"Prevent push down limit if the offset reach max_result_window":
   - do:
       headers:
         Content-Type: 'application/json'


integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/3102.yml
```diff
@@ -0,0 +1,53 @@
...
+
+
+---
+teardown:
+  - do:
+      query.settings:
+        body:
+          transient:
+            plugins.calcite.enabled : false
+            plugins.calcite.fallback.allowed : true
```
penghuo: nit, no need revert setting, by default is disabled

LantaoJin: why we need this teardown step? it seems not enable calcite in setup step. 

@penghuo is each xyx.yml independent?

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/3102.yml
```diff
@@ -0,0 +1,53 @@
...
+
+  - do:
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: 'source=test | head 2 | head 1 from 1 '
+  - match: { "total": 1 }
+  - match: { "schema": [ { "name": "id", "type": "bigint" } ] }
+  - match: { "datarows": [ [ 2 ] ] }
```
LantaoJin, line 43: what's the result without this fixing?

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/3102.yml
penghuo: no need revert setting, by default is disabled

PR #3729, opened 2025-06-03
Support `appendcol` command with Calcite

### LINKED ISSUE #3172: [FEATURE]add `appendcol` ppl command
This feature request proposes the addition of an `appendcol`- command to OpenSearch’s Piped Processing Language (PPL). The `appendcol` command (modeled after Splunk’s command of the same name), allows users to append the result of one or more searches as additional columns to the existing search results, offering a powerful mechanism to enrich data by combining results horizontally.

The following query:
```sql
source=logs | stats count by status | appendcol [ stats avg(response_time) by status ]
```

In this example the `appendcol` command adds a new column containing the average response time for each status alongside the original column that shows the count by status. The original row structure is preserved.

_Implementation could probably use  rewriting into a join query._

**Add Proposal Document**

In many use cases, users need to compare or merge multiple result sets side-by-side without altering the existing row structure. The `appendcol` command will become a used feature for horizontally merging datasets, and bringing this feature to OpenSearch PPL would greatly enhance data processing flexibility.

By adding `appendcol` to PPL, users can:

* Combine different queries into a unified result set without duplicating rows.
* Enrich an existing dataset with additional metrics or fields from other searches.
* Improve the efficiency and readability of complex queries.

### **Technical Details**

**Syntax**
 The `appendcol` command would accept a query inside square brackets `[ ]`, representing the additional pipeline that produces the column(s) to append to the original result set.

**Behavior**
 The new column(s) would be aligned with the rows of the original dataset based on their order of appearance. Each appended column must produce the same number of rows as the base dataset to ensure proper alignment. Any discrepancies in row counts could result in null values for mismatched rows.

LantaoJin:
### Description
Support `appendcol` command with Calcite

### Related Issues
Resolves #3172 

### Check List

### DIFF
--- a/docs/user/ppl/cmd/appendcol.rst
+++ b/docs/user/ppl/cmd/appendcol.rst
@@ -1,0 +1,138 @@
+=========
+appendcol
+=========
+
+.. rubric:: Table of contents
+
+.. contents::
+   :local:
+   :depth: 2
+
+
+Description
+============
+| (Experimental)
+| (From 3.1.0)
+| Using ``appendcol`` command to append the result of a sub-search and attach it alongside with the input search results (The main search).
+
+Version
+=======
+3.1.0
+
+Syntax
+============
+appendcol [override=<boolean>] <sub-search>
+
+* override=<boolean>: optional. Boolean field to specify should result from main-result be overwritten in the case of column name conflict.
+* sub-search: mandatory. Executes PPL commands as a secondary search. The sub-search uses the same data specified in the source clause of the main search results as its input.
+
+Configuration
+=============
+This command requires Calcite enabled.
+
+Enable Calcite::
+
+	>> curl -H 'Content-Type: application/json' -X PUT localhost:9200/_plugins/_query/settings -d '{
+	  "transient" : {
+	    "plugins.calcite.enabled" : true
+	  }
+	}'
+
+Result set::
+
+    {
+      "acknowledged": true,
+      "persistent": {
+        "plugins": {
+          "calcite": {
+            "enabled": "true"
+          }
+        }
+      },
+      "transient": {}
+    }
+
+Example 1: Append a count aggregation to existing search result
+===============================================================
+
+This example appends "count by gender" to "sum by gender, state".
+
+PPL query::
+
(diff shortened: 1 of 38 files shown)


docs/user/ppl/cmd/appendcol.rst
```diff
@@ -0,0 +1,138 @@
...
+    +--------+-------+----------+------------+
+
+Example 2: Append a count aggregation to existing search result with override option
+====================================================================================
+
+This example appends "count by gender" to "sum by gender, state" with override option.
+
+PPL query::
+
+    PPL> source=accounts | stats sum(age) by gender, state | appendcol override=true [ stats count(age) by gender ] | head 10;
```
penghuo, line 86: query result is same as override = false?, we should demo the usage of override?

docs/user/ppl/cmd/appendcol.rst
```diff
@@ -0,0 +1,138 @@
...
+
+Version
+=======
+3.1.0
+
+Syntax
+============
+appendcol [override=<boolean>] <sub-search>
+
+* override=<boolean>: optional. Boolean field to specify should result from main-result be overwritten in the case of column name conflict.
```
penghuo, line 26: in case appendcol is implemented as full join, we should consider enforce limitations?

LantaoJin: I will create a new issue about operator limitation for all high cost operators.

docs/user/ppl/cmd/appendcol.rst
LantaoJin: The result is different with override=false. check the (second row, first column)

LantaoJin:
> Thanks for the changes!
> 
> I wonder in the long term whether `appendcols` should be a new relational operator that extends the SQL standard? Semantically, it's closer to a vertical version of `UNION ALL`, aligning rows by position rather than key. While a `JOIN` using `ROW_NUMBER()` can reproduce the behavior, does it introduces unnecessary overhead?

AFAIK, only `lateral` join (ANSI 2003) is close to a vertical version of `UNION ALL`. But for the case of aligning one-by-one rows by position, `lateral` join is not a proper solution.

PR #3736, opened 2025-06-04
Add Commit-Based Artifact Download Support

ahkcs:
### Description

This PR modified workflow to support commit-based downloads using a commit-version-map.json file. This allows users to download specific artifact versions that correspond to a particular Git commit without needing to know the precise version number.

```Problem Addressed```

Previously, our download scripts required specifying an exact version (e.g., 0.1.0-SNAPSHOT), which made it difficult to retrieve artifacts that were built from a specific commit.

```Solution```

We've implemented a commit-to-version mapping system that:

Creates and maintains a mapping file (commit-version-map.json) during snapshot publishing
Enables artifact download scripts to look up specific versions by commit ID
Works for both language-grammar and async-query-core artifacts

```Key Changes```

In the Publishing Workflow

Added code to generate/update the commit-version-map.json file
Captures detailed version information including both base version and timestamped artifact version

### DIFF
--- a/.github/workflows/publish-snapshots-and-grammar-files.yml
+++ b/.github/workflows/publish-snapshots-and-grammar-files.yml
@@ -10,6 +10,7 @@ on:
 
 env:
   SNAPSHOT_REPO_URL: https://aws.oss.sonatype.org/content/repositories/snapshots/
+  COMMIT_MAP_FILENAME: commit-version-map.json
 
 jobs:
   publish-grammar-files:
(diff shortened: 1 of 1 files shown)


noCharger (commented):
Are we creating a new file instead of reusing maven-metadata.xml?

ahkcs:
> Are we creating a new file instead of reusing maven-metadata.xml?

Yes, we are creating a new file commit-history json to manage commit-version mapping

Swiddis (commented):
What happens if two jobs run at once and pull the previous mapping at the same time?

Ideally, we should include the commit ID as part of the artifact key itself, and then in the script pull by the commit directly without needing to parse a separate mapping.

Swiddis (approved):

noCharger:
> What happens if two jobs run at once and pull the previous mapping at the same time?
> 
> Ideally, we should include the commit ID as part of the artifact key itself, and then in the script pull by the commit directly without needing to parse a separate mapping.

+1 on this.

Swiddis (approved):

noCharger (approved):

PR #3747, opened 2025-06-08
Support `flatten` command with Calcite

### LINKED ISSUE #3712: [FEATURE] Support `flatten` command with Calcite
Support `flatten` command with Calcite.
Ref:
https://github.com/opensearch-project/opensearch-spark/issues/669
https://github.com/opensearch-project/sql/issues/3029

yuancu:
### Description
Implement `flatten` command that flattens a nested struct / object field into separate fields in a document.

Flatten extracts the fields in a struct into separate other fields in the document:

E.g. For
```
{
  "name": "Jack",
  "address": {
    "state": "Oregon",
    "city": "Portland"
  }
}
```
Flattening `address` will result in
```
{
  "name": "Jack",
  "address": {
    "state": "Oregon",
    "city": "Portland"
  },
  "state": "Oregon",
  "city": "Portland"
}
```

### Related Issues
Resolves #3712

### Check List


ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLFlattenTest.java
```diff
@@ -0,0 +1,135 @@
...
+        .traitDefs((List<RelTraitDef>) null)
+        .programs(Programs.heuristicJoinOrder(Programs.RULE_SET, true, 2));
+  }
+
+  @Test
+  public void testFlatten() {
+    String ppl = "source=DEPT | flatten EMP";
+    RelNode root = getRelNode(ppl);
+    // Regarded as an identity scan. See RelBuilder#L2801
+    String expectedLogical = "LogicalTableScan(table=[[scott, DEPT]])\n";
```
yuancu: The logical plans of `flatten` is suspicious.

The logical plan right after `visitFlatten` in `CalciteRelNodeVisitor` is:
```
LogicalProject(DEPTNO=[$0], EMP=[$1], EMP.EMPNO=[$2], EMP.EMPNAME=[$3], EMPNO=[$2], EMPNAME=[$3])
  LogicalTableScan(table=[[scott, DEPT]])
```

But it is later transformed to a simple logical table scan.

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,100 @@
...
+
+    PPL> source=nested | flatten message as (creator, dow, info);
+    fetched rows / total rows = 2/2
+    +-----------------------------------------+--------+---------+-----+------+
+    | message                                 | myNum  | creator | dow | info |
+    |-----------------------------------------|--------|---------|-----|------|
+    | {"info":"a","author":"e","dayOfWeek":1} | 1      | e       | 1   | a    |
+    | {"info":"b","author":"f","dayOfWeek":2} | 2      | f       | 2   | b    |
+    +-----------------------------------------+--------+---------+-----+------+
+
```
LantaoJin, line 106: could you add a limitation section to describe that it only works when calcite enabled.

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,100 @@
...
+The flattened fields will be ordered **lexicographically** by their original
+key names in the struct. I.e. if the struct has keys ``b``, ``c`` and ``Z``,
+the flattened fields will be ordered as ``Z``, ``b``, ``c``.
+
+Note that ``flatten`` does not work on arrays. Please use ``expand`` command
+to expand an array field into multiple rows instead. If the field is an nested
+array of structs, only the first element of the array will be flattened.
+
+Syntax
```
LantaoJin, line 31: missing version section before syntax
```
Version
=======
3.1.0
```

docs/user/ppl/cmd/appendcol.rst
```diff
@@ -11,9 +11,10 @@ appendcol
 
 Description
 ============
-| (Experimental)
-| (From 3.1.0)
-| Using ``appendcol`` command to append the result of a sub-search and attach it alongside with the input search results (The main search).
+    (Experimental)
+    (From 3.1.0)
+
```
LantaoJin: I think we can remove these highlights, we already have a Version section after Description

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLFlattenTest.java
LantaoJin: seem `relBuilder.projectPlus(expandedFields)` doesn't work in this case, please check the result by adding `verifyResult(root, expectedResult);`

yuancu: Fixed the problem and corrected unit tests.

docs/user/ppl/cmd/flatten.rst
yuancu: Added. I think this means the same as "Since 3.1.0"? Maybe we could make it explicit somewhere that these features tagged as *since 3.1.0* only works with calcite enabled.

docs/user/ppl/cmd/flatten.rst
yuancu: Added

docs/user/ppl/cmd/flatten.rst
LantaoJin: 3.1.0 != calcite enabled

docs/user/ppl/cmd/appendcol.rst
yuancu: Yep. Removed.

docs/user/ppl/cmd/flatten.rst
yuancu: Indeed, I also see 3.0.0. Thanks for explanation.

docs/user/ppl/cmd/ad.rst
```diff
@@ -11,7 +11,8 @@ ad (deprecated by ml command)
 
 Description
 ============
-| The ``ad`` command applies Random Cut Forest (RCF) algorithm in the ml-commons plugin on the search result returned by a PPL command. Based on the input, the command uses two types of RCF algorithms: fixed in time RCF for processing time-series data, batch RCF for processing non-time-series data.
+
+The ``ad`` command applies Random Cut Forest (RCF) algorithm in the ml-commons plugin on the search result returned by a PPL command. Based on the input, the command uses two types of RCF algorithms: fixed in time RCF for processing time-series data, batch RCF for processing non-time-series data.
```
LantaoJin: why made these lots of code changes? I am afraid it may easy to introduce PR conflicts.

yuancu: Since I already removed tags like the following from these docs, I find it may be appropriate to correct / unify some language use in these docs alongside. After all, the docs are rarely updated once the corresponding PRs of their functionality are merged. 

```
    (Experimental)
    (From 3.1.0)
```

LantaoJin: Still confuse me what is the relationship between removing these two lines and changes from other doc files.

docs/user/ppl/cmd/join.rst
```diff
@@ -11,13 +11,12 @@ join
...
-| (Experimental)
-| (From 3.0.0)
-| Using ``join`` command to combines two datasets together. The left side could be an index or results from a piped commands, the right side could be either an index or a subquery.
+
+Use ``join`` command to combines two datasets together. The left side could be an index or results from a piped commands, the right side could be either an index or a subquery.
 
 Version
 =======
-3.0.0
+Since 3.0.0
```
LantaoJin: do not add `Since`

docs/user/ppl/cmd/join.rst
```diff
@@ -11,13 +11,12 @@ join
 
 Description
 ============
-| (Experimental)
-| (From 3.0.0)
```
LantaoJin: just remove these two lines, do not change rests

docs/user/ppl/cmd/join.rst
yuancu: Without since, it sounds like the command is at its 3.0.0 version (indicating e.g. it might not be compatible with 4.0.0). I think using since makes it less confusing and is a common practice.

I'll revert it in this PR. We may raise another issue to fix this in the future.

LantaoJin: No, this version contains `from` meaning. it is the min version to support. if one command/function retired from 4.0.0, we can just remove it from doc. the doc is a historical doc. if you are a 4.0.0 user, read the related doc from 4.0.0 tag. else, if you are a 2.15 user, read the 2.15 doc

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,116 @@
...
+    {"message":{"info":"b","author":"f","dayOfWeek":2},"myNum":2}
+
+with the following mapping:
+
+.. code-block:: json
+
+    {
+      "mappings": {
+        "properties": {
+          "message": {
```
penghuo, line 63: `message` is Json Object, why index as nested field. Usually, [Json Array is indexed as nested field](https://docs.opensearch.org/docs/latest/field-types/supported-field-types/nested/).

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,116 @@
...
+Use ``flatten`` command to flatten a nested struct / object field into separate
+fields in a document.
+
+The flattened fields will be ordered **lexicographically** by their original
+key names in the struct. I.e. if the struct has keys ``b``, ``c`` and ``Z``,
+the flattened fields will be ordered as ``Z``, ``b``, ``c``.
+
+Note that ``flatten`` does not work on arrays. Please use ``expand`` command
+to expand an array field into multiple rows instead. If the field is an nested
+array of structs, only the first element of the array will be flattened.
```
penghuo: > If the field is an nested array of structs,

Why work on Array of structs if flatten does not work on Array?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -973,6 +1006,63 @@ public RelNode visitTableFunction(TableFunction node, CalcitePlanContext context
     throw new CalciteUnsupportedException("Table function is unsupported in Calcite");
   }
 
+  /**
+   * Visit flatten command.
+   *
+   * <p>The flatten command is used to flatten a struct field into multiple fields. This
+   * implementation simply projects the flattened fields and renames them according to the provided
+   * aliases or the field names in the struct. This is possible because the struct / object field
+   * are always read in a flattened manner in OpenSearch.
```
penghuo, line 1020: Could u explain flattened manner in OpenSearch? what is expected result from OpenSearch

docs/user/ppl/cmd/flatten.rst
yuancu: Yes, using object here is more clear. Fixed.

However, the boundary between array and object is blurry in OpenSearch. Both object and array fields can store a struct or an array of structs. Both nested and object works in this case.

docs/user/ppl/cmd/flatten.rst
yuancu: It's because although not expected, an object/nested field can store arrays. I have updated the doc to make it easier to understand.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: Flatten extracts the fields in a struct into separate other fields in the document:

E.g. For
```
{"name": "Jack", "address": {"state": "Oregon", "city": "Portland"}}
```
Flattening `address` will result in
```
{"name": "Jack", "address": {"state": "Oregon", "city": "Portland"}, "state": "Oregon", "city": "Portland"}
```

(Also copied to PR description)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -978,6 +1011,63 @@ public RelNode visitTableFunction(TableFunction node, CalcitePlanContext context
...
+   *
+   * @param node Flatten command node
+   * @param context CalcitePlanContext
+   * @return RelNode representing the visited logical plan
+   */
+  @Override
+  public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
```
qianheng-aws, line 1027: Seems the implementation relies on the flattened nested fields in the RelRowType, but it may not work for PPL:

```
source = index | fields message| flatten message
```

qianheng-aws (changes requested):
Need test for this case:

```
source = index | fields message| flatten message
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: It fails this test case indeed. Trying to fix it.

yuancu:
> Need test for this case:
> 
> ```
> source = index | fields message| flatten message
> ```

We need complete support for typed arrays and structs in order to make it work in such scenarios.

I added an ignored test case and left a TODO item, saying that we'll correct it after fixing #3459 and #3751

docs/user/ppl/cmd/flatten.rst
```diff
@@ -0,0 +1,125 @@
...
+    +-----------------------------------------+--------+---------+-----+------+
+    | message                                 | myNum  | creator | dow | info |
+    |-----------------------------------------|--------|---------|-----|------|
+    | {"info":"a","author":"e","dayOfWeek":1} | 1      | e       | 1   | a    |
+    | {"info":"b","author":"f","dayOfWeek":2} | 2      | f       | 2   | b    |
+    +-----------------------------------------+--------+---------+-----+------+
+
+Limitations
+===========
+* ``flatten`` after ``fields`` is not supported under the current
```
LantaoJin: Change to:

`flatten` command may not work as expected when its flattened fields are invisible.
For example in query `source=my-index | fields message | flatten message`, 
the `flatten message` command doesn't work since some flattened fields such as `message.info` and `message.author` after command `fields message` are invisible.
As an alternative, you can change to `source=my-index | flatten message`

yuancu: Done.

LantaoJin (approved):

PR #3748, opened 2025-06-09
Fix: correct ATAN(x, y) and CONV(x, a, b) functions bug

### LINKED ISSUE #3672: [BUG] `ATAN(x, y)` and `CONV(x, a, b)` functions not work with Calcite
**What is the bug?**
`ATAN(x, y)` and `CONV(x, a, b)` functions not work with Calcite

```
... | eval f = atan(2, 3) | fields f
... | eval f = conv(age, 10, 16) | fields f
```

 - Plugins

ishaoxy:
### Description
Solved the bug that **ATAN(x, y)** function does not support two parameters in v3 version. Also fixed a bug in the **CONV(x, a, b)** function where incorrect type conversion caused unexpected results.

### Related Issues
Resolves #3672

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMathematicalFunctionIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMathematicalFunctionIT.java
@@ -5,8 +5,6 @@
 
 package org.opensearch.sql.calcite.remote;
 
-import java.io.IOException;
-import org.junit.Ignore;
 import org.opensearch.sql.ppl.MathematicalFunctionIT;
 
 public class CalciteMathematicalFunctionIT extends MathematicalFunctionIT {


LantaoJin:
@ishaoxy Remove these rows https://github.com/opensearch-project/sql/blob/270aa0dc3349e3a3c3b4e8e0e19c772f81c17781/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMathematicalFunctionIT.java#L20-L29
to enable the tests with your patch.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -531,6 +531,11 @@ void populate() {
                       builder.makeLiteral(" "),
                       arg),
               PPLTypeChecker.family(SqlTypeFamily.STRING)));
+      register(
+          ATAN,
```
dai-chen, line 535: Could you elaborate why we register ATAN2 as ATAN? I see both Calcite and SparkSQL has ATAN and ATAN2 separately: https://calcite.apache.org/docs/reference.html, https://spark.apache.org/docs/latest/api/sql/#atan

LantaoJin: ATAN can accept 2 arguments in v2 which is not existing in standard SQL, this PR is to align with v2 to avoid breaking changes. If you think it's unnecessary, we can delete it in v2 and comment ATAN will 2 arguments will not accept since 3.1.0.

LantaoJin (approved):

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
dai-chen: I see. Sure, we can go ahead if we follow aligning with V2 criteria for all functions. Just feel this may cause unnecessary work in both Calcite and Spark. We can discuss later. Thanks!

dai-chen (approved):

PR #3749, opened 2025-06-09
Pushdown system limit automatically for data-intensive operations

### LINKED ISSUE #3731: [ENHANCEMENT] Set operator limitation for high cost operators
From 3.0.0, calcite engine provides more power to support complex operators such as join. But SQL plugin still leverage coordinator node to run all operators in memory if pushdown doesn't work. So configurable operator limitations are required for that.

Add configurable limitations for each high cost operators.

Keep as-is and limit by memory monitor.

LantaoJin:
### Description
From v3.0.0, PPL introduces commands that may increase data volume. To prevent out-of-memory problem, the system automatically enforces this **system limit** in the pushdown context of the physical index scan operator for such commands.

Now, only join physical operator in PPL query will be affected by this PR. In future, we can specify more sub-configs to control detailed operations, such UDTF etc.

`plugins.query.system_limit`: The size configures the maximum amount of documents to be pull from OpenSearch for data-intensive operations (e.g. join, lookup). The default value is: 50000.

Note, only apply system limit automatically to data-intensive (data-bloat) operations. For non-data-intensive operations, the maximum amount of documents to be pull from OpenSearch equals to value of `plugins.query.size_limit`. For example, even `SORT` is a high cost physical operator, the amount of documents is sorted cannot greater than value of `plugins.query.size_limit`.

### Related Issues
Resolves #3731 

### Check List


docs/user/ppl/admin/settings.rst
```diff
@@ -159,3 +159,48 @@ Rollback to default value::
...
+
+plugins.query.system_limit
+==========================
+
+Description
+-----------
+
+The size configures the maximum amount of documents to be pull from OpenSearch for data-intensive operations (e.g. join, lookup). The default value is: 50000.
+
+From v3.0.0, PPL introduces commands that may increase data volume. To prevent out-of-memory problem, the system automatically enforces this ``system limit`` in the pushdown context of the physical index scan operator for such commands.
```
penghuo: > such commands

list commands affect by `plugins.query.system_limit.xxx` in [limitation docs](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/limitations/limitations.rst).

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
```diff
@@ -40,6 +40,10 @@ public enum Key {
     /** Common Settings for SQL and PPL. */
     QUERY_MEMORY_LIMIT("plugins.query.memory_limit"),
     QUERY_SIZE_LIMIT("plugins.query.size_limit"),
+    QUERY_SYSTEM_LIMIT("plugins.query.system_limit"),
```
penghuo, line 46: does `plugins.query.system_limit` needed? does it equal to `plugins.query.system_limit.scan_index`?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -135,6 +137,7 @@ protected enum PushDownType {
     AGGREGATION,
     // SORT,
     LIMIT,
+    SYSTEM_LIMIT, // not user specified limit, it's the limit pushed for high-cost operators
```
penghuo, line 140: Why differentiate SYSTEM_LIMT and USER_LIMIT. Can USER_LIMIT support both use case?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -81,7 +84,34 @@ public void register(RelOptPlanner planner) {
       for (RelOptRule rule : OpenSearchIndexRules.OPEN_SEARCH_INDEX_SCAN_RULES) {
         planner.addRule(rule);
       }
+      Integer systemLimit = osIndex.getSettings().getSettingValue(Settings.Key.QUERY_SYSTEM_LIMIT);
+      if (systemLimit >= 0) {
```
penghuo: negative value means unlimited? if so, add to doc.

docs/user/admin/settings.rst
```diff
@@ -188,6 +188,51 @@ Result set::
       }
     }
 
+plugins.query.system_limit
+==========================
+
+Description
+-----------
+
+The size configures the maximum amount of documents to be pull from OpenSearch for data-intensive operations (e.g. join, lookup). The default value is: 50000.
```
penghuo: The PR description provides a clearer definition. The size parameter specifies the maximum allowable data bloat from input documents during operations.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/SystemLimitRuleConfig.java
```diff
@@ -0,0 +1,77 @@
...
+    return this.operandSupplier;
+  }
+
+  @Override
+  public String description() {
+    return this.description;
+  }
+
+  /** Operand supplier to specify physical join operator */
+  public static RelRule.OperandTransform SYSTEM_LIMIT_JOIN_TRANSFORM =
```
penghuo: One concern with the rule-based approach is that it lacks awareness of the original command context. Consequently, it cannot manage specific commands directly. For example, an expand command translates to a correlated join internally, but since the rule doesn't recognize this translation, users cannot effectively control parameters such as system_limit.expand_size.

Is it possible apply limit when composing the RelNode? for instance Join(left, right) can be compose as Join(Limit(left, 50000), Limit(right, 50000))

LantaoJin: > For example, an expand command translates to a correlated join internally, but since the rule doesn't recognize this translation, users cannot effectively control parameters such as system_limit.expand_size.

The system limit rule only impacts physical plans, so whatever operators which will converted to hash join or sort merge join will be controlled. Check in-subquery in CalciteExplainIT.

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
LantaoJin: IMO, `plugins.query.system_limit` is a base config, if we have more specific config such as `plugins.query.system_limit.A` and `plugins.query.system_limit.B`, user can configure `plugins.query.system_limit` to impact both limitation of A and B.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
LantaoJin: No. Image the follow PPLs:
```
source=t | head 10 | where a = 1
```
We cannot pushdown filter `a = 1` for above PPL, but we can pushdown both filter `a = 1` and limit `10` for following PPL.
```
source=t | where a = 1 | head 10
```
This `limit 10` is USER LIMIT.

But for PPL
```
source=t | where a=1 | join <join conditions> [ source=s | where b=1 ]
```
To pushdown filter `a=1` and `b=1` to scan, it will check `isLimitPushedDown()`, SYSTEM LIMIT won't be impacted.

That's why pushdown system limit is hard to implement in logical plan level. We have to distinguish how to convert different logical limits, or all logical limit will convert to the same `EnumerableSort`. Current implement just add SYSTEM LIMIT pushdown, no logical and physical operator introduced in a query plan.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
LantaoJin: Em, actually, the config requires min value 0, so I think we can delete it checking. I will update the doc about config ranges. similar to Query.size_limit. If user want unlimited, set it to Int.MaxValue.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
LantaoJin: Check the `CalciteExplainIT`, only pushdown context impacted, no logical or physical operators changes.

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
LantaoJin: How about remove `plugins.query.system_limit.join` in this PR? We can add specific configs in feature. That's why I didn't mention `plugins.query.system_limit.join` in user doc.

docs/user/admin/settings.rst
LantaoJin: I think the `maximum amount of documents to be pull from OpenSearch` is a precision description comparing to `maximum allowable data bloat`.

qianheng-aws:
1. Will this change leads to incorrect results?

3. To avoid data bloating, why not adding a limit operator for each child of join operator? It should have similar effect.

docs/user/ppl/limitations/limitations.rst
```diff
@@ -84,3 +84,15 @@ plugins.query.field_type_tolerance setting is enabled, the SQL/PPL plugin will h
...
+===============================================
+
+Since v3.0.0, PPL introduces commands that may increase data volume. To prevent out-of-memory problem, the system
+automatically enforces this system limit in the pushdown context of the physical index scan operator for such commands.
+``plugins.query.system_limit``: The size configures the maximum amount of documents to be pull from OpenSearch for
+data-intensive operations (e.g. join, lookup). The default value is: 50000.
+Note, only apply system limit automatically to data-intensive (data-bloat) operations. For non-data-intensive operations,
+the maximum amount of documents to be pull from OpenSearch equals to value of plugins.query.size_limit.
+For example, even SORT is a high cost physical operator, the amount of documents to sorted cannot greater than value of
+``plugins.query.size_limit``.
```
penghuo: > For example, even SORT is a high cost physical operator, the amount of documents to sorted cannot greater than value of ``plugins.query.size_limit``.

Not accurate, sort will work on entire dataset, plugins.query.size_limit: configures the maximum amount of rows to be fetched from query execution results.

docs/user/admin/settings.rst
penghuo: Agree, it is a short-term workaround,  but it leading to incorrectness results. 

For instance, `source=t1 | eval a=b+1 | sort a | join b [source=t2]`,  
* with current impl, the query rewrite as `source=t1 | head 50000 | eval a=b+1 | sort a | join b [source=t2 | head 50000]`
* the expected is `source=t1 | eval a=b+1 | sort a | head 50000 | join b [source=t2 | head 50000]`

docs/user/ppl/limitations/limitations.rst
LantaoJin: Yes, you are right. We need a specific `system_limit.sort`.

docs/user/admin/settings.rst
LantaoJin: > Agree, it is a short-term workaround, but it leading to incorrectness results.
> 
> For instance, `source=t1 | eval a=b+1 | sort a | join b [source=t2]`,
> 
> * with current impl, the query rewrite as `source=t1 | head 50000 | eval a=b+1 | sort a | join b [source=t2 | head 50000]`
> * the expected is `source=t1 | eval a=b+1 | sort a | head 50000 | join b [source=t2 | head 50000]`

Do you mean to mitigate the incorrectness results, we shouldn't add limit to high cost operators, such as sort (Sort is a high cost operators which will read all data from index). We just limit the data-intensive operators such as join. If yes, I think we can just add limit to right side of join (for inner and left join).

LantaoJin:
@penghuo @qianheng-aws @dai-chen I have updated the description with new code refactor, and docs. please take another look.

RyanL1997:
Hi @LantaoJin , is this still requred?

LantaoJin:
Close as there is a new alternative PR https://github.com/opensearch-project/sql/pull/4501

PR #3760, opened 2025-06-10
Fix field not found issue in join output when column names are ambiguous

LantaoJin:
### Description
```
source=table1 | join left=t1 right=t2 on t1.id=t2.id table2 | eval a = 1 | fields t1.id, t2.id
```
Fail with {alias=t2, fieldName=id} field not found

#### Explanation
Join outputs could contain duplicated column name, typically the join criteria `ON t1.id = t2.id`. The `id` in output of join operator is ambiguous. Calcite will rename the ambiguous column name with numeric suffix, the output generates `id` and `id0` when a new project add to stack.

#### Solution
To avoid `id0`, this PR renames the `id0` to `<alias>.id`, or `<tableName>.id` if no alias existing.

#### Example
table1:
| id   |
| --- |
| 1 |
| 2 |

table2
| id   |
| --- |
| 1 |
| 3 |
```
source=table1 | join left=t1 right=t2 on t1.id=t2.id table2 | eval a = 1
```
outputs `t1.id`, `t2.id` and `a`

```
source=table1 | join on table1.id=table2.id table2 | eval a = 1
```
outputs `table1.id`, `table2.id` and `a`

```
source=table1 | join on table1.id=t2.id table2 as t2 | eval a = 1
```
outputs `table1.id`, `t2.id` and `a`

```
source=table1 | join right=tt on table1.id=t2.id [ source=table2 as t2 | eval b = id ] | eval a = 1
```
outputs `table1.id`, `tt.id`, `tt.b` and `a`

### Related Issues
Resolves #3717 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -6,6 +6,8 @@
 package org.opensearch.sql.calcite;
 
 import static org.apache.calcite.sql.SqlKind.AS;
+import static org.opensearch.sql.ast.tree.Join.JoinType.ANTI;
+import static org.opensearch.sql.ast.tree.Join.JoinType.SEMI;
 import static org.opensearch.sql.ast.tree.Sort.NullOrder.NULL_FIRST;
 import static org.opensearch.sql.ast.tree.Sort.NullOrder.NULL_LAST;
 import static org.opensearch.sql.ast.tree.Sort.SortOption.DEFAULT_DESC;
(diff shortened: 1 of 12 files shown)


LantaoJin:
ping @dai-chen @penghuo for additional review.

docs/user/ppl/cmd/join.rst
```diff
@@ -80,117 +80,61 @@ Example 1: Two indices join
...
+
+   * - Query
+     - Output
+   * - source=table1 | join left=t1 right=t2 on t1.id=t2.id table2 | eval a = 1
+     - t1.id, t2.id, a
+   * - source=table1 | join on table1.id=table2.id table2 | eval a = 1
+     - table1.id, table2.id, a
+   * - source=table1 | join on table1.id=t2.id table2 as t2 | eval a = 1
+     - table1.id, t2.id, a
+   * - source=table1 | join right=tt on table1.id=t2.id [ source=table2 as t2 | eval b = id ] | eval a = 1
```
penghuo, line 139: does `table1.id=tt.id` works?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -531,8 +533,42 @@ public RelNode visitJoin(Join node, CalcitePlanContext context) {
...
+      List<String> leftColumns = context.relBuilder.peek(1).getRowType().getFieldNames();
+      List<String> rightColumns = context.relBuilder.peek().getRowType().getFieldNames();
+      List<String> rightTableName =
+          PlanUtils.findTable(context.relBuilder.peek()).getQualifiedName();
+      // Using `table.column` instead of `catalog.database.table.column` as column prefix because
+      // the schema for OpenSearch index is always `OpenSearch`. But if we reuse this logic in other
+      // query engines, the column can only be searched in current schema namespace. For example,
+      // If the plan convert to Spark plan, and there are two table1: database1.table1 and
+      // database2.table1. The query with column `table1.id` can only be resolved in the namespace
+      // of "database1". User should run `using database1` before the query which access `table1.id`
```
penghuo, line 554: Does it restrict joins between tables in the same database?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -531,8 +533,42 @@ public RelNode visitJoin(Join node, CalcitePlanContext context) {
...
+      context.relBuilder.join(
+          JoinAndLookupUtils.translateJoinType(node.getJoinType()), joinCondition);
+    } else {
+      // Join condition could contain duplicated column name, Calcite will rename the duplicated
+      // column name with numeric suffix, e.g. ON t1.id = t2.id, the output contains `id` and `id0`
+      // when a new project add to stack. To avoid `id0`, we will rename the `id0` to `alias.id`
+      // or `tableIdentifier.id`:
+      List<String> leftColumns = context.relBuilder.peek(1).getRowType().getFieldNames();
+      List<String> rightColumns = context.relBuilder.peek().getRowType().getFieldNames();
+      List<String> rightTableName =
```
dai-chen, line 547: Just to confirm: does these logic below work for multi-join query plan tree recursively?

docs/user/ppl/cmd/join.rst
LantaoJin: it works

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: I don't think so, the table name is just an `aliasField` to find the fieldName in RelBuilder. Ref https://github.com/apache/calcite/blob/15e0a3b56f7e48858b19127120b84987040c43dd/core/src/main/java/org/apache/calcite/tools/RelBuilder.java#L620-L622

But it brings me that one corner case cannot work:
If the join between tables in different databases contains same names. For example `db1.tableA join db2.tableA`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: Yes.

penghuo (approved):

PR #3766, opened 2025-06-12
Trigger GC when catch MemoryUsageExceedException on the second retry

### LINKED ISSUE #3750: [FEATURE] Memory limitation when merging thousands of complex indices
In #3653 we merge objects type when fetching schemas from several indices using index schema. But in the testing, when the cluster meets higher concurrency (e.g. 64) and complex scenario (e.g. 1000 indices with 15 depth), it will throw out of memory.  We need to figure out what is the bottleneck and see how to resolve it.

Figure out what is the bottleneck of out of memory problem.

LantaoJin:
### Description
When we do the benchmark of merging thousands of complex indices: when the cluster meets higher concurrency (e.g. 64) and complex scenario (e.g. 1000 indices with 15 depth), it will throw out of memory.

In fact, this OOM issue is not solely caused by insufficient heap memory. The primary reason is "lazy" heap memory GC under high concurrency, leading to numerous query failures because OpenSearchResourceMonitor#isHealthy() returned true.

#### Changes in this PR:​​
1. Replaces the memory-intensive schema caches with `SoftReference`, allowing the cache to be prioritized for garbage collection when memory is low.​​
2. Makes `OpenSearchResourceMonitor` retry upon `MemoryUsageExceedException`, triggering `System.gc()` on the second retry attempt.​

#### Benchmarking:
before: total 100 queries in 64 concurrent threads, 72 fails, 0 FullGC

Screenshot 2025-06-12 at 15 38 17

after: total 100 queries in 64 concurrent threads, 0 ~ 5 fails, 23 FullGC (Trigger GC on the first retry)

Screenshot 2025-06-12 at 18 59 26

after: total 100 queries in 64 concurrent threads, 0 ~ 3 fails, 7 FullGC (Trigger GC on the second retry)

Screenshot 2025-06-17 at 17 28 10

### Related Issues
Resolves #3750 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchMemoryHealthy.java
```diff
@@ -44,9 +44,9 @@ public boolean isMemoryHealthy(long limitBytes) {
     }
   }
 
-  static class RandomFail {
+  static class FastFail {
     public boolean shouldFail() {
-      return ThreadLocalRandom.current().nextBoolean();
+      return false;
```
penghuo, line 49: If always return false, we should remove FastFail class and MemoryUsageExceedFastFailureException.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchResourceMonitor.java
```diff
@@ -36,6 +36,15 @@ public OpenSearchResourceMonitor(Settings settings, OpenSearchMemoryHealthy memo
             .ignoreExceptions(OpenSearchMemoryHealthy.MemoryUsageExceedFastFailureException.class)
             .build();
     retry = Retry.of("mem", config);
+    retry
+        .getEventPublisher()
+        .onRetry(
+            event -> {
+              if (event.getNumberOfRetryAttempts() == 1) {
+                System.gc();
```
penghuo, line 49: My concern is that triggering System.gc() may impact other OpenSearch operations, such as indexing.

Swiddis: Would the fix still work if we did gc on the second retry?

Since we have exponential backoff, we could probably mitigate a lot of the load by letting some retries play out, and then we won't see 20+ GCs for this query load.

LantaoJin: > Would the fix still work if we did gc on the second retry?
> 
> Since we have exponential backoff, we could probably mitigate a lot of the load by letting some retries play out, and then we won't see 20+ GCs for this query load.

Seems better, reduced to 7 GCs.

LantaoJin:
in the commit dac4878, I pass the `ClusterService` to `ResourceMonitor` and trigger `System.gc()` only in the dedicated coordinator node. At the same time, change the trigger time to second retry. The GC reduced to 7 from 20+. @penghuo @Swiddis

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchResourceMonitor.java
songkant-aws: I remember SoftReference has default GC behavior in JVM when JVM is close to OOM. Just curious what's the behavior of not explicitly calling System.gc()?

Swiddis (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchResourceMonitor.java
Swiddis: > Just curious what's the behavior of not explicitly calling System.gc()?

The default GC runs async in the background, if you're still initializing a lot of objects on a smaller-resource cluster you can end up running out of memory before GC finishes (especially if running over older GC generations). Synchronously calling it has the benefit that you actually wait for the GC to finish before the retry.

Issue #3734, opened 2025-06-04
[FEATURE] Export PPL-Calcite engine as reusable library


dai-chen:
**Is your feature request related to a problem?**

Currently, the Calcite-based PPL engine lives within the OpenSearch SQL/PPL plugin and cannot be reused outside of that context. This limits its applicability in other environments, where the same parsing and logical planning capabilities would be valuable.

Use cases:

1. https://github.com/opensearch-project/opensearch-spark/issues/1136
2. https://github.com/opensearch-project/sql/issues/627
3. PPL CLI

**What solution would you like?**

### Granularity: Modular Publishing vs. Fat JAR

We considered two approaches for packaging and publishing the Calcite-based PPL engine for external use:

- **Option A: Modular Publishing**  
  - **Description:** Publish each internal module (`ppl`, `core`, `opensearch`, etc.) as an independent Maven artifact.  
  - **Pros:** Enables better reuse and flexibility—consumers (e.g., Spark) can depend only on the components they need.  
  - **Cons:** Requires publishing all relevant internal modules.

- **Option B: Fat JAR**  
  - **Description:** Bundle all internal modules and dependencies into a single artifact.  
  - **Pros:** Simplifies consumption (e.g., by the PPL CLI) with a single plug-and-play artifact.  
  - **Cons:** Tightly couples all components, increases artifact size, and reduces modularity.

<img width="1314" alt="Image" src="https://github.com/user-attachments/assets/87b01231-b5ae-4d0e-8b13-2d786dc7a794" />

### The New API Module

We propose introducing a new `api` module as a high-level integration layer for the Calcite-based engine and as primary entry point for external consumers.

- **Expose a unified API**: The current Calcite engine is tightly coupled with OpenSearch internals (e.g., `DataSourceService`, `Settings`), making reuse difficult. The new module will provide a clean, reusable interface for external consumers without exposing low-level implementation details.

- **Guaranteed Interoperability**: Integration tests will validate the contract between the Calcite engine and external consumers (e.g., Spark, CLI), ensuring correctness and long-term compatibility.

Tasks:

- [x] https://github.com/opensearch-project/sql/pull/3763
    - [x] **Add a Gradle task** to build a shaded (fat) JAR that bundles all required transitive dependencies for easy external consumption.
    - [x] **Publish the packaged library** to a Maven repository so it can be consumed by Spark or other tools.
- [x] https://github.com/opensearch-project/sql/pull/3783
[...]

PR #3783, opened 2025-06-16
Add unified query API module for external integration

dai-chen:
### Description

This PR introduces a new api module containing the `UnifiedQueryPlanner` class, which provides a high-level interface for parsing and planning PPL queries. This module is designed to support external consumers such as Spark and CLI without exposing Calcite or OpenSearch internals. Unit tests are included to document usage and verify correctness.

TODO: update issue with long term. add README. address query limit.

### Related Issues

Resolves https://github.com/opensearch-project/sql/issues/3734

### Check List

### DIFF
--- a/api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
+++ b/api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.api;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import org.antlr.v4.runtime.tree.ParseTree;
+import org.apache.calcite.jdbc.CalciteSchema;
+import org.apache.calcite.plan.RelTraitDef;
+import org.apache.calcite.rel.RelCollation;
+import org.apache.calcite.rel.RelCollations;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Sort;
+import org.apache.calcite.rel.logical.LogicalSort;
+import org.apache.calcite.rel.metadata.DefaultRelMetadataProvider;
+import org.apache.calcite.schema.Schema;
+import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.schema.impl.AbstractSchema;
+import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.tools.FrameworkConfig;
+import org.apache.calcite.tools.Frameworks;
+import org.apache.calcite.tools.Programs;
+import org.opensearch.sql.ast.statement.Query;
+import org.opensearch.sql.ast.statement.Statement;
+import org.opensearch.sql.ast.tree.UnresolvedPlan;
+import org.opensearch.sql.calcite.CalcitePlanContext;
+import org.opensearch.sql.calcite.CalciteRelNodeVisitor;
+import org.opensearch.sql.common.antlr.Parser;
+import org.opensearch.sql.executor.OpenSearchTypeSystem;
+import org.opensearch.sql.executor.QueryType;
+import org.opensearch.sql.ppl.antlr.PPLSyntaxParser;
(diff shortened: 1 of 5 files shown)


settings.gradle
```diff
@@ -8,6 +8,7 @@ rootProject.name = 'opensearch-sql'
 
 include 'opensearch-sql-plugin'
 project(':opensearch-sql-plugin').projectDir = file('plugin')
+include 'api'
```
Swiddis, line 11: ~~Should we include it as part of the plugin snapshot and keep the plugin as a multipurposed jar like this? Seems like something where we should make the core SQL functionality a library, the opensearch plugin is one target that we can build, and the programmatic API is a different target.~~ Ok if I'm understanding right this is being shipped as its own library already as part of the module list earlier, it just seems odd to also put it in the OS plugin if the OS plugin isn't using it

api/README.md
```diff
@@ -0,0 +1,74 @@
...
+Use the declarative, fluent builder API to initialize the `UnifiedQueryPlanner`.
+
+```java
+UnifiedQueryPlanner planner = UnifiedQueryPlanner.builder()
+    .language(QueryType.PPL)
+    .catalog("opensearch", schema)
+    .cacheSchema(true)
+    .build();
+
+RelNode plan = planner.plan("source = opensearch.test");
```
Swiddis, line 21: How does one execute the plan after they receive it?

api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
```diff
@@ -0,0 +1,183 @@
...
+   */
+  public RelNode plan(String query) {
+    return preserveCollation(analyze(parse(query)));
+  }
+
+  private Parser buildQueryParser(QueryType queryType) {
+    if (queryType == QueryType.PPL) {
+      return new PPLSyntaxParser();
+    }
+    throw new IllegalArgumentException("Unsupported query type: " + queryType);
```
Swiddis, line 92: Is there a specific technical reason we can't include SQL yet? This is just using the ANTLR parser, right?

dai-chen: Because we haven't decided how to unify SQL yet. For example, we can enable SQL by Calcite parser directly or by our own ANTLR parser and AstBuilder.

settings.gradle
dai-chen: Right, for now it is for external consumer. Later we can refactor `plugin` module to use this unified API as well.

api/README.md
dai-chen: Good question. Currently, the plan isn’t directly executable. As noted in the README, the planner is designed to eventually return an executable plan—either a Calcite physical plan for immediate execution in the current JVM (useful for the OpenSearch plugin and CLI), or a SparkSQL plan for distributed execution by Spark (useful for PPL in Spark).

I initially considered designing the API this way, but haven’t yet found a clean way to model everything within Calcite’s optimizer. I plan to work on this later, especially since PPL in Spark Phase 2 may require it.

api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
Swiddis: My understanding was that Calcite came after parsing, does Calcite also have a parser? Good to know

dai-chen: Yes, Calcite itself has its own parser. You can find more details in their documentation and paper.

dai-chen:
@LantaoJin @penghuo Please have a look when you have a moment. This is currently only for initial phase in https://github.com/opensearch-project/opensearch-spark/issues/1136 so we can begin publishing PRs on Spark side. Thanks!

LantaoJin (approved):

dai-chen:
There seems flaky test.

```
2025-06-24T17:39:21.3968640Z 3577 tests completed, 1 failed, 540 skipped
2025-06-24T17:39:21.3969870Z Tests with failures:
2025-06-24T17:39:21.4097140Z  - org.opensearch.sql.calcite.tpch.CalcitePPLTpchIT.testQ19
```

PR #2130, opened 2023-09-22
Removed Domain Validation

vamsi-amazon:
### Description
* Currently in datasource uri validation, we validate domain which is excessively restrictive and hampering ease of use in case of docker and using ips in prometheus URL. This PR addresses this problem by removing the restriction. 

* We can always avoid ssrf issue by specifying the host denylist `plugins.query.datasources.uri.hosts.denylist` 

Manual Testing
```
curl --location 'localhost:9200/_plugins/_query/_datasources/' \
--header 'Content-Type: application/json' \
--data '{
  "name": "my_prometheus3",
  "connector": "PROMETHEUS",
  "allowedRoles": [],
  "properties": {
    "prometheus.uri": "http://10.0.0.1:9090"
  }
}'
```
This call worked which used to not function earlier.
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3797, opened 2025-06-18
Remove unneeded dependency on commons-validator

Swiddis:
### Description

Well, that was fun.

After #3782 didn't actually update the beanutils dep, I did a bunch of tracing how we depend on it, such that the resolution isn't actually changing anything. It turns out we import this dependency via `apache.commons.validator`, which [has an open issue to update beanutils as of a couple weeks ago](https://issues.apache.org/jira/browse/VALIDATOR-500). So I went to figure out what we depend on the validator for.

Answer: we don't. The last ref was removed in #2130 ([diff](https://github.com/opensearch-project/sql/pull/2130/files#diff-b9c313c1ef71065a4f4a166763b18b244b1c6d5908cf4ed3512337ce6a4ca98eL11)).

So removing it will remove pointless CVE chasing, and make our bundle smaller.

### Related Issues
Same CVE as the last one.

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -149,7 +149,6 @@ allprojects {
         resolutionStrategy.force 'org.apache.commons:commons-text:1.11.0'
         resolutionStrategy.force 'commons-io:commons-io:2.15.0'
         resolutionStrategy.force 'org.yaml:snakeyaml:2.2'
-        resolutionStrategy.force 'org.apache.commons:commons-beanutils:2.0.0'
     }
 }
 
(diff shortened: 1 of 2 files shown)


dai-chen (approved):

brianf-aws:
Hey @Swiddis I'm curious to know how you were able to trace it? Looking to get some insight as our plugin does use org.apache.commons:commons-beanutils. I forced it 1.10 but when I check its gradle cache I can see its pulling in 1.9.4 via a transitive dependency

Swiddis:
In our case we had to get it to show dependencies just for our plugin's build task, `gradle :opensearch-sql:dependencies` instead of `gradle dependencies`. I'm not sure what the equivalent task group is for ml-commons.

RyanL1997 (approved):

noCharger (approved):

Issue #3379, opened 2025-03-05
[FEATURE] Calcite Engine Framework: Pushdown scripts


LantaoJin:
**Is your feature request related to a problem?**
Framework work of the development feature of Calcite Engine.

**What solution would you like?**
1. Enable Script Pushdown
2. Eval expression could be pushdown by script pushdown feature.

qianheng-aws:
Solution we have currently, need POC to verify their feasibility.

----
Option1:
Client: RexNode -> Linq4j Expression -> Compile to Java Code -> Submit Java Code String to Script API
Server: Receive Java Code -> Java Code execution

RPOS: faster when handling big data
CONS: the step of compiling and execution requires more developing effort

---
Option2:
Client: RexNode -> Linq4j Expression -> Kyro serialization -> Submit byte string to Script API
Server: Receive byte string -> deserialize to Linq4j Expression -> invoke the evaluate API of Linq4j Expression

PROS: faster when handling small data, since saving the step of compile
CONS: introduce the dependency of Kyro both in opensearch core and sql plugin.

model-collapse:
I am voting to a ultimate solution which can adaptively select between option1 and option2 according to the estimation of processing time. For short term, I think we might ASAP adding Kyro to core.

dai-chen:
Hi team, I’d like to share some findings from my exploration to provide you more ideas as we evaluate different strategies. cc: @penghuo

### Summary

- **Option 3: Reusing the V2 Script Engine** – Leverages the existing `opensearch_query_expression` engine by converting Calcite expressions into the V2 expression tree.
- **Option 4: Registering a New Script Language** – Introduces a expression script language that safely executes allowlisted expressions, fully abstracting the execution details from users.

In the near term, Option 4 offers a practical path to unblock integration by avoiding the security concerns of Options 1 and 2 and minimizing effort. It also decouples Calcite’s centralized planning from execution—important given ongoing debate over whether Calcite’s physical plan should be a universal model and intrude into the execution layer. Looking ahead, Option 5 could evolve into a more formal, engine-agnostic expression layer that bridges planning and execution without tight dependence on Calcite.

### Option 3: Reusing the V2 Script Engine

Instead of introducing a new script engine, we can reuse the existing one (`opensearch_query_expression`) from the V2 engine. In this approach, the script engine is treated as a standalone script language (similar to Painless), independent of both the V2 engine which is being deprecated and Calcite engine which is focused on logical optimization.

[...]

PR #3812, opened 2025-06-23
Support filter script push down for calcite

qianheng-aws:
### Description
Implement filter script push down for calcite by registering a new script language for java code string. This new script language can only be invoked by Node client since it requires a option of `engine_type` which is not allowed in Rest client.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3379

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
@@ -1233,8 +1233,6 @@ public class PredicateAnalyzer {
         return booleanValue();
       } else if (isString()) {
         return RexLiteral.stringValue(literal);
-      } else if (isDecimal()) {
-        return ((BigDecimal) literal.getValue()).doubleValue();
       } else {
         return rawValue();
       }
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
@@ -32,11 +32,6 @@ import static org.opensearch.sql.data.type.ExprCoreType.INTEGER;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import java.io.IOException;
-import java.io.Serializable;
-import java.io.StringReader;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Type;
 import java.time.chrono.ChronoZonedDateTime;
 import java.util.List;
@@ -69,7 +64,6 @@ import org.apache.calcite.rex.RexExecutable;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexProgram;
 import org.apache.calcite.rex.RexProgramBuilder;
-import org.apache.calcite.runtime.Utilities;
 import org.apache.calcite.schema.SchemaPlus;
 import org.apache.calcite.sql.validate.SqlConformance;
 import org.apache.calcite.sql.validate.SqlConformanceEnum;
@@ -77,9 +71,6 @@ import org.apache.calcite.util.BuiltInMethod;
 import org.apache.calcite.util.Util;
 import org.apache.commons.lang3.StringUtils;
 import org.checkerframework.checker.nullness.qual.Nullable;
-import org.codehaus.commons.compiler.CompileException;
-import org.codehaus.janino.ClassBodyEvaluator;
-import org.codehaus.janino.Scanner;
 import org.opensearch.index.fielddata.ScriptDocValues;
 import org.opensearch.script.FilterScript;
 import org.opensearch.script.ScriptContext;
@@ -116,7 +107,8 @@ public class CalciteScriptEngine implements ScriptEngine {
   @Override
   public <T> T compile(
       String scriptName, String scriptCode, ScriptContext<T> context, Map<String, String> options) {
-    Function1<DataContext, Object[]> function = compile(scriptCode, "generated Rex code");
+    Function1<DataContext, Object[]> function =
+        new RexExecutable(scriptCode, "generated Rex code").getFunction();
 
     if (CONTEXTS.containsKey(context)) {
       return context.factoryClazz.cast(CONTEXTS.get(context).apply(function));
@@ -304,31 +296,5 @@ public class CalciteScriptEngine implements ScriptEngine {
     }
 
     return code;
-  }
-
-  /**
-   * This function is copied from Calcite RexExecutable It's used to compile java code string to
-   * java function.
-   */
-  public static Function1<DataContext, Object[]> compile(String code, Object reason) {
-    try {
-      ClassBodyEvaluator cbe = new ClassBodyEvaluator();
-      cbe.setClassName("Reducer");
-      cbe.setExtendedClass(Utilities.class);
-      cbe.setImplementedInterfaces(new Class[] {Function1.class, Serializable.class});
-      cbe.setParentClassLoader(RexExecutable.class.getClassLoader());
-      cbe.cook(new Scanner((String) null, new StringReader(code)));
-      Class c = cbe.getClazz();
-      Constructor<Function1<DataContext, Object[]>> constructor = c.getConstructor();
-      return (Function1) constructor.newInstance();
-    } catch (IOException
-        | InstantiationException
-        | IllegalAccessException
-        | InvocationTargetException
-        | NoSuchMethodException
-        | CompileException var5) {
-      Exception e = var5;
-      throw new RuntimeException("While compiling " + reason, e);
-    }
   }
 }
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
@@ -24,6 +24,12 @@ public class CalciteExplainIT extends ExplainIT {
 
   // Only for Calcite, as v2 gets unstable serialized string for function
   @Test
+  public void testFilterScriptPushDownExplain() throws Exception {
+    super.testFilterScriptPushDownExplain();
+  }
+
+  // Only for Calcite, as v2 gets unstable serialized string for function
+  @Test
   public void testFilterFunctionScriptPushDownExplain() throws Exception {
     super.testFilterFunctionScriptPushDownExplain();
   }


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/PPLCompoundedScriptEngine.java
```diff
@@ -0,0 +1,57 @@
...
+import org.opensearch.script.ScriptContext;
+import org.opensearch.script.ScriptEngine;
+import org.opensearch.sql.opensearch.storage.serialization.DefaultExpressionSerializer;
+
+/**
+ * Custom expression script engine that supports using core engine expression code in DSL as a new
+ * script language just like built-in Painless language.
+ */
+@RequiredArgsConstructor
+public class PPLCompoundedScriptEngine implements ScriptEngine {
```
penghuo: Why PPL specific? SQL on Calcite will reuse this engine?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -922,6 +960,132 @@ public QueryExpression notIn(LiteralExpression literal) {
...
+      JavaTypeFactoryImpl typeFactory =
+          new JavaTypeFactoryImpl(rexBuilder.getTypeFactory().getTypeSystem());
+      RexToLixTranslator.InputGetter getter =
+          new ScriptInputGetter(typeFactory, rowType, fieldTypes);
+      this.code = CalciteScriptEngine.translate(rexBuilder, List.of(rexNode), getter, rowType);
+    }
+
+    @Override
+    public QueryBuilder builder() {
+      long currentTime = Hook.CURRENT_TIME.get(-1L);
```
penghuo, line 981: 1. why need current_time?
2. does these error message will expose to user?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -922,6 +960,132 @@ public QueryExpression notIn(LiteralExpression literal) {
...
+        RexNode rexNode,
+        RexBuilder rexBuilder,
+        RelDataType rowType,
+        Map<String, ExprType> fieldTypes) {
+      // Compile code when creating to detect exception as early as possible
+      JavaTypeFactoryImpl typeFactory =
+          new JavaTypeFactoryImpl(rexBuilder.getTypeFactory().getTypeSystem());
+      RexToLixTranslator.InputGetter getter =
+          new ScriptInputGetter(typeFactory, rowType, fieldTypes);
+      this.code = CalciteScriptEngine.translate(rexBuilder, List.of(rexNode), getter, rowType);
```
penghuo, line 976: Why not generate code in buider() method?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,334 @@
...
+      }
+      return value;
+    }
+  }
+
+  /**
+   * This function is copied from Calcite RexExecutorImpl It's used to compile RexNode expression to
+   * java code string.
+   */
+  public static String translate(
```
penghuo, line 247: Could call RexExecutorImpl.getExecutable() without copy code?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,334 @@
...
+      if (Variable.UTC_TIMESTAMP.camelName.equals(name))
+        return params.get(Variable.UTC_TIMESTAMP.camelName);
+
+      ScriptDocValues<?> docValue = docProvider.get().get(name);
+      if (docValue == null || docValue.isEmpty()) {
+        return null; // No way to differentiate null and missing from doc value
+      }
+
+      Object value = docValue.get(0);
+      if (value instanceof ChronoZonedDateTime) {
```
penghuo, line 233: in which case value is ChronoZonedDateTime?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,334 @@
...
+        // We store timestamp as string in the current implementation with Calcite.
+        // And the string should have the format defined in ExprTimestampValue
+        // TODO: should we change to store timestamp as Instant in the future.
```
penghuo, line 236: did not get it. we should store timestamp as Instant inside query engine.

penghuo:
>  This new script language can only be invoked by Node client since it requires a option of engine_type which is not allowed in Rest client.

This is a side project that runs the plugin as a standalone CLI and leverages DSL queries to interact with the OpenSearch cluster. 
Is there any workaround, can we encode the engine_type with script_code?

LantaoJin:
> >  This new script language can only be invoked by Node client since it requires a option of engine_type which is not allowed in Rest client.
> 
> 
> 
> This is a side project that runs the plugin as a standalone CLI and leverages DSL queries to interact with the OpenSearch cluster. 
> 
> Is there any workaround, can we encode the engine_type with script_code?

Yes, we can check the "calcite" related package information in script code to determine the engine type, rather than encode additional information.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/PPLCompoundedScriptEngine.java
qianheng-aws: Yes, remove `PPL` in the name may be more appropriate

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
qianheng-aws: Some implementation of date-related functions rely on the `current_time`, so we need to pass its value to this script to be used in execution later

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
qianheng-aws: In order to detect compiling error as early as possible to prevent useless visiting for the other expressions. The most common exception is that it will throw exception if the field is text type without keyword subfield or struct type, see the code in CalciteScriptEngine::ScriptInputGetter::field.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: No, that method uses ` RexToLixTranslator.InputGetter` while our logic needs a specific implemented `CalciteScriptEngine::ScriptInputGetter`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: Seems `date` will be if I remember. It just copy the same logic as v2.

qianheng-aws:
> > This new script language can only be invoked by Node client since it requires a option of engine_type which is not allowed in Rest client.
> 
> This is a side project that runs the plugin as a standalone CLI and leverages DSL queries to interact with the OpenSearch cluster. Is there any workaround, can we encode the engine_type with script_code?

Add option incompatible with rest client is also for security consideration. Otherwise, users can use Rest client to submit java code script without limitation.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: But our date related functions accept fields of date type as string. See `DatetimeFunction` or others.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
qianheng-aws: > does these error message will expose to user?

No, it will be caught by the filter push down logic in `CalciteLogicalIndexScan`, and then it won't choose push down.

Actually this is defensive check. We will always set this hook value in `OpenSearchDriver::connect`.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,334 @@
...
+    } catch (IOException
+        | InstantiationException
+        | IllegalAccessException
+        | InvocationTargetException
+        | NoSuchMethodException
+        | CompileException var5) {
+      Exception e = var5;
+      throw new RuntimeException("While compiling " + reason, e);
+    }
+  }
```
songkant-aws: Minor: Since we already get the code string, we can new RexExecutable and call getFunction() to avoid this copied method?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
penghuo: could u double confirm, date value in OpenSearch is epoch instant without timezone.

penghuo:
> Add option incompatible with rest client is also for security consideration. Otherwise, users can use Rest client to submit java code script without limitation.

Current implementation depend on Script.java in Core deny any options,  if it is not true in future, user can run ANY java code. 
```
if (options.size() > 1 || options.size() == 1 && options.get(CONTENT_TYPE_OPTION) == null) {
    options.remove(CONTENT_TYPE_OPTION);

    throw new IllegalArgumentException("illegal compiler options [" + options + "] specified");
}
```

Are there any workarounds we can consider? Instead of using raw Java code, can we leverage LINQ4j Enumerable or V2 Expression to limit user input to only allowlisted expressions and reduce the risk?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -923,6 +960,132 @@ public QueryExpression notIn(LiteralExpression literal) {
...
```
dai-chen, line 976: I'm thinking is it possible we delay this compilation to Java until `compile()` in ScriptEngine? In this way, we serialize and deserialize `RexCall` instead, similar as V2 expression pushdown? Just some thoughts.

qianheng-aws: RexNode is unserializable.  We've though of similar method as mentioned in the option2 here: https://github.com/opensearch-project/sql/issues/3379#issuecomment-2893090703.

@songkant-aws is working on option2 but seems it's not feasible.

songkant-aws: Leveraging Kryo serialization for classes without zero-argument constructor requires explicitly registering specific class serializer. That means we may need to spend more efforts on creating separate customized Kryo serializers for Calcite Expression classes/Enumerable classes. 

If we go this way, I'm not confident yet to ensure all kinds of expressions can be correctly serialized.

dai-chen: > RexNode is unserializable.

You mean it's not by JDK or Kryo right? Does other serializer work, or even Json's?

Do we know how many classes related to `RexCall` we need to serialize? Just thinking if that's not many and if we can ignore fields not needed by code-gen, can we have customize serializer?

dai-chen: @qianheng-aws I checked the option 2 you posted. I feel it's different?

What I was thinking is:

- Coordinator: RexNode -> serialized code
- Worker: Code -> RexNode -> Linq4j expression -> Interpret or compile to Java

In this way, the script engine can be bounded to allowlisted RexNode.

dai-chen: @songkant-aws Not sure if you've explored the idea above. If so, could you share your PoC branch if any? I can do more test from my side. Thanks!

songkant-aws: @dai-chen  For the sake of convenience, I pulled Heng's code in my local and add my Kryo serialization prototype code on top of it. And then pushed the code to my personal branch: https://github.com/songkant-aws/sql/tree/kryo-serialization

Basically, the logic is similar. I just feel adding additional serialization logic for third party classes requires too much effort.

qianheng-aws: > Do we know how many classes related to RexCall we need to serialize?

The common RexNode we need may not be too many. Like RexCall, RexLiteral, RexInputRef, RexLambda, RexLambdaRef... I think the biggest blocker is the inner fields of them. Like RexCall, it has plenty of SqlOperators implemented by Calcite or us.

dai-chen: @songkant-aws @qianheng-aws Got it. Will have a look and also see if other options on my side. Thanks!

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -424,35 +454,42 @@ private QueryExpression binary(RexCall call) {
...
-      PredicateAnalyzerException firstError = null;
       boolean partial = false;
+      // For function isEmpty and isBlank, we implement them via expression `isNull or {@function}`,
+      // Unlike `OR` in Java, `SHOULD` in DSL will evaluate both branches and lead to NPE.
+      if (call.getKind() == SqlKind.OR
+          && call.getOperands().size() == 2
+          && (call.getOperands().get(0).getKind() == SqlKind.IS_NULL
+              || call.getOperands().get(1).getKind() == SqlKind.IS_NULL)) {
+        throw new UnsupportedScriptException(
+            "DSL will evaluate both branches of OR with isNUll, prevent push-down to avoid NPE");
```
LantaoJin, line 465: is it an existed issue in v2?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -923,6 +960,132 @@ public QueryExpression notIn(LiteralExpression literal) {
...
+          new ScriptInputGetter(typeFactory, rowType, fieldTypes);
+      this.code = CalciteScriptEngine.translate(rexBuilder, List.of(rexNode), getter, rowType);
+    }
+
+    @Override
+    public QueryBuilder builder() {
+      long currentTime = Hook.CURRENT_TIME.get(-1L);
+      if (currentTime < 0) {
+        throw new UnsupportedScriptException(
+            "ScriptQueryExpression requires a valid current time from hook, but it is not set");
```
LantaoJin, line 984: non-blocker: can we add a UT for this branch?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    /**
+     * DocValue only support long and double for integer and float, cast to the related type first
+     */
+    private Expression tryConvertDocValue(Expression docValueExpr, ExprType exprType) {
+      return switch (exprType) {
+        case INTEGER -> EnumUtils.convert(docValueExpr, Long.class);
+        case FLOAT -> EnumUtils.convert(docValueExpr, Double.class);
+        default -> docValueExpr;
```
LantaoJin, line 190: can the `exprType` be a `SHORT` or `BINARY`? for doc value. Seems all numeric types will impact.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,300 @@
...
+        // And the string should have the format defined in ExprTimestampValue
+        // TODO: should we change to store timestamp as Instant in the future.
+        return new ExprTimestampValue(((ChronoZonedDateTime<?>) value).toInstant()).value();
+      }
+      return value;
+    }
+  }
+
+  /**
+   * This function is copied from Calcite RexExecutorImpl It's used to compile RexNode expression to
```
LantaoJin, line 244: Seems you copied the decompiled code instead source code. Can you check it?:
`java.util.Iterator var5`, `Expressions.declare(16`, `Expressions.methodDecl(1`, etc.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/CalciteFilterScript.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.lucene.index.LeafReaderContext;
+import org.opensearch.script.FilterScript;
+import org.opensearch.search.lookup.SearchLookup;
+import org.opensearch.sql.opensearch.storage.script.core.CalciteScript;
+
+/**
+ * Calcite script executor that executes the generated code on each document and determine if the
+ * document is supposed to be filtered out or not.
+ */
+@EqualsAndHashCode(callSuper = false)
```
LantaoJin, line 21: should we use `true` here?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
qianheng-aws: I think V2 won't have this issue. Firstly, it seems doesn't support function `isEmpty` and `isBlank`. And secondly, it won't have such issue if it includes the null handling logic in one expression.

We have such issue because we break `isEmpty` into 2 expressions(isNull or isEmpty, all from Calcite) and Calcite's `isEmpty` doesn't have null handling in its implementation.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: It imitates the same logic as v2. As said in the description, it only needs specific handling for Integer and Long.

https://github.com/opensearch-project/sql/blob/870b82c137fcd640e45783792865ab523a1561ad/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/core/ExpressionScript.java#L139-L143

qianheng-aws:
Mark draft currently for security consideration.

LantaoJin:
Close due to #3859 merged.

PR #3821, opened 2025-06-25
Add compare_ip operator udfs

### LINKED ISSUE #3776: [FEATURE] Support comparing IP values with Calcite
Current comparison operators are mapped to Calcite's built-in comparisons, but the IP types are implemented as user-defined types (UDT). As a result, comparison operators does not support comparison between IP types.

This is not a problem when pushdown is enabled, since such comparison will be pushed down and can be properly executed. For example, `source=weblogs | where host > '1.2.3.4' | fields host` will be executed in the following way ([weblogs dataset](https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/resources/weblogs.json)):

```json
{
    "calcite": {
        "logical": "LogicalProject(host=[$0])\n  LogicalFilter(condition=[>($0, '1.2.3.4':VARCHAR)])\n    CalciteLogicalIndexScan(table=[[OpenSearch, weblogs]])\n",
        "physical": "CalciteEnumerableIndexScan(table=[[OpenSearch, weblogs]], PushDownContext=[[PROJECT->[host], FILTER->>($0, '1.2.3.4':VARCHAR)], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"query\":{\"range\":{\"host\":{\"from\":\"1.2.3.4\",\"to\":null,\"include_lower\":false,\"include_upper\":true,\"boost\":1.0}}},\"_source\":{\"includes\":[\"host\"],\"excludes\":[]},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}]}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
    }
}
```

However, when pushdown is disabled, it raises the following error:
```
{
  "error": {
    "reason": "Invalid Query",
    "details": "No assign rules for OTHER defined",
    "type": "CalciteUnsupportedException"
  },
  "status": 400
}
```
The mysterious error arises from the process of coercing operand types of the built-in greater than operator of Calcite, since it does not handle UDTs properly.

Implement user-defined function for IP comparison operators.

Convert IP UDT to a Calcite's built-in type, so that the comparison can be handled properly. However, there isn't a direct correspondence of IP.

I'm also thinking of converting it to a tuple of numbers, maybe Calcite comparison operators can recognize them.

https://github.com/opensearch-project/sql/issues/3229

ishaoxy:
### Description

- Introduced a new UDF `CompareIpFunction` for IP address comparisons.
- Since Calcite does not provide a native SQL type for IP addresses, EXPR_IP is mapped to `SqlTypeName.OTHER`.
- Enhanced the `registerOperator` method to support registering one or multiple operators under a single function name, enabling function overloading based on operand types.
- Registered the newly implemented IP comparison operators under the existing comparison function names(e.g., =, !=, <, >, etc.).
- Added new and specific type checkers for IP comparison and CIDR functions to ensure accurate operand validation and operator resolution.

### Related Issues
Resolves #3776

### Check List


core/src/main/java/org/opensearch/sql/expression/function/udf/ip/EqualsIpFunction.java
```diff
@@ -0,0 +1,87 @@
...
+import org.opensearch.sql.utils.IPUtils;
+
+/**
+ * {@code Equals(ip1, ip2)} checks if two IP addresses are equal.
+ *
+ * <p>Signature:
+ *
+ * <ul>
+ *   <li>(STRING, STRING) -> BOOLEAN
+ *   <li>(IP, STRING) -> BOOLEAN
```
yuancu: I think (STRING, STRING) should not be compared as IP?

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/EqualsIpFunction.java
```diff
@@ -0,0 +1,87 @@
...
+      try {
+        IPAddress ipAddress1 = IPUtils.toAddress(ip1);
+        IPAddress ipAddress2 = IPUtils.toAddress(ip2);
+        return IPUtils.compare(ipAddress1, ipAddress2) == 0;
+      } catch (SemanticCheckException e) {
+        return false;
+      }
+    }
+
+    public static boolean Equals(String ip1, ExprIpValue ip2) {
```
yuancu: lowercase `equals` may be more appropriate.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/EqualsIpFunction.java
```diff
@@ -0,0 +1,87 @@
...
+ * <p>Signature:
+ *
+ * <ul>
+ *   <li>(STRING, STRING) -> BOOLEAN
+ *   <li>(IP, STRING) -> BOOLEAN
+ *   <li>(STRING, IP) -> BOOLEAN
+ *   <li>(IP, IP) -> BOOLEAN
+ * </ul>
+ */
+public class EqualsIpFunction extends ImplementorUDF {
```
yuancu: I'm thinking whether there is a better way to implement the comparing classes without the need to declare 6 classes doing virtually the same thing.

E.g.

```java
class CompareIPFunction {
  private BiPredicate biPredicate;

  private CompareIPFunction(BiPredicate biPredicate){
    this.biPredicate = biPredicate;
  }

  public static CompareIPFunction less() {
    return new CompareIPFunction((a, b) -> a < b);
  }

...

  public boolean compare(IPAddress ip1, IPAddress ip2){
    return biPredicate.test(IPUtils.compare(ip1, ip2), 0)
  }

...

}
```

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/CidrMatchFunction.java
```diff
@@ -12,9 +12,7 @@
 import org.apache.calcite.linq4j.tree.Expression;
 import org.apache.calcite.linq4j.tree.Expressions;
 import org.apache.calcite.rex.RexCall;
-import org.apache.calcite.sql.type.OperandTypes;
-import org.apache.calcite.sql.type.ReturnTypes;
-import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.type.*;
```
yuancu: The project avoids using `import *`. You can configure the IDE to not convert multiple imports to `import *`.

qianheng-aws:
I think we should implement `implicit conversion` before implementing comparison for non-primitive types.

After that, for Ip comparison, we can convert string to ExprIP. And then we can implement some generic comparison functions for types of comparable, via calling their `compareTo` method.

But current approach is still acceptable if we won't have other UDT in the future.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/CidrMatchFunction.java
```diff
@@ -44,9 +42,12 @@ public SqlReturnTypeInference getReturnTypeInference() {
 
   @Override
   public UDFOperandMetadata getOperandMetadata() {
-    // EXPR_IP is mapped to SqlTypeFamily.VARCHAR in
+    // EXPR_IP is mapped to SqlTypeFamily.NULL in
     // UserDefinedFunctionUtils.convertRelDataTypeToSqlTypeName
-    return UDFOperandMetadata.wrap(OperandTypes.STRING_STRING);
+    return UDFOperandMetadata.wrap(
```
qianheng-aws: Maybe we should use new `PPLTypeChecker` here and override its `checkOperandTypes` method to only allow accept `string` or `ip`.

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
```diff
@@ -400,6 +400,10 @@ private static List<ExprType> getExprTypes(SqlTypeFamily family) {
               OpenSearchTypeFactory.TYPE_FACTORY.createSqlType(SqlTypeName.INTEGER));
           case ANY, IGNORE -> List.of(
               OpenSearchTypeFactory.TYPE_FACTORY.createSqlType(SqlTypeName.ANY));
+            // We borrow SqlTypeFamily.NULL to represent EXPR_IP. This is a workaround
```
qianheng-aws: If we stop use `CompositeOperandTypeChecker` and create a new TypeChecker ourself, maybe we can avoid using `SqlTypeFamily.NULL` to represent EXPR_IP and check operands on RelDataType level.

yuancu: @qianheng-aws How about this implementation https://github.com/ishaoxy/sql/pull/1
It doesn't use NULL for IP, but created two other classes for IP type checking.

yuancu: It just came to me that there's not only compare that checks IP type. CIDR_MATCH also has to validate IP types. If we make a specific IP type checker for compare operators, we may also have to create one for each other functions like cidrmatch and geoip.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -614,6 +614,14 @@ public PPLTypeChecker getTypeChecker() {
     }
 
     void populate() {
+      // register operators for IP comparing
+      registerOperator(NOTEQUAL, PPLBuiltinOperators.NOT_EQUALS_IP);
```
penghuo: Questions, Do we need to support IP as UDT in PPL engine?


In OpenSearch PPL, IP handling depends on *index field type*. CIDR-based filtering should use ip_range queries for ip fields, script pushdown / in-memory processing for keyword, text, and runtime string fields. 

Field Type | Use Case | Expectation
-- | -- | --
IP Field | search index=log ip="192.168.0.0/16" | Rewrite as term query
  | search index \| where cidrmatch("192.168.0.0/16", ip) | Rewrite as term query
Keyword Field | search index=log ip="192.168.0.0/16" | Rewrite as term query, extactally keyword match
  | search index \| where cidrmatch("192.168.0.0/16", ip) | Script pushdown — ip field is a string, not rewrite as term query.
Text Field | search index=log ip="192.168.0.0/16" | Rewrite as query_string query, full text search
  | search index \| where cidrmatch("192.168.0.0/16", ip) | Script pushdown — ip field is a string, not rewrite as term query.
Runtime Field | search index=log \| parse ip=regex(...)\| where ip="192.168.0.0/16" | Script pushdown — ip field is a string, it is a string comparsion query
  | search index=log \| parse ip=regex(...)\| where cidrmatch("192.168.0.0/16", ip) | Script pushdown — ip field is a string, not rewrite as term query.

ishaoxy:
> Please add IT for ip comparison.

Handled.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/EqualsIpFunction.java
ishaoxy: Typing error, fixed.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/EqualsIpFunction.java
ishaoxy: Thanks for reminding me this rule.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/CidrMatchFunction.java
ishaoxy: Got it.

qianheng-aws:
> @qianheng-aws How about this implementation https://github.com/ishaoxy/sql/pull/1
It doesn't use NULL for IP, but created two other classes for IP type checking.

Yes, I think that's what we need. In the future when we have type conversion, that IPTypeChecker should only accept type of IP, and type conversion should convert `String` to `ExprIpType`

> It just came to me that there's not only compare that checks IP type. CIDR_MATCH also has to validate IP types. If we make a specific IP type checker for compare operators, we may also have to create one for each other functions like cidrmatch and geoip.

Yes since we treat `ExprIPType` as a specific type instead of `String`

@yuancu

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/EqualsIpFunction.java
ishaoxy: Handled.

yuancu: Tweaked the code a little bit: https://github.com/ishaoxy/sql/pull/2

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -614,6 +614,14 @@ public PPLTypeChecker getTypeChecker() {
...
+      registerOperator(EQUAL, PPLBuiltinOperators.EQUALS_IP);
+      registerOperator(GREATER, PPLBuiltinOperators.GREATER_IP);
+      registerOperator(GTE, PPLBuiltinOperators.GTE_IP);
+      registerOperator(LESS, PPLBuiltinOperators.LESS_IP);
+      registerOperator(LTE, PPLBuiltinOperators.LTE_IP);
```
LantaoJin: SqlStdOperatorTable.EQUALS` is a very basic std operator which widely used in calcite internal, I am worry that it may introduce potential bugs and performance regression for pushdown.

yuancu: But Calcite's built-in operators like `SqlStdOperatorTable.EQUALS` does not handle our UDT like IP. These new operators are only effective to IP comparison, controlled via type checkers. Comparison between other types will still falls to Calcite's built-in comparison operators.

There was two solutions:
- one is to convert IP UDT to a type that is comparable by calcite's comparators.
- another is to add new operators exclusively for IP comparision

We opted the latter.

ishaoxy: In order to keep the comparison logic the same as v2, I chose to add these specific operator udfs for IP comparison instead of converting it to string type that calcite can compare.

LantaoJin: > These new operators are only effective to IP comparison, controlled via type checkers. Comparison between other types will still falls to Calcite's built-in comparison operators.

Can we move above logic to a specific method `registerOverrideOperator`, it quite confused me.

LantaoJin: Or modify to
```
registerOperator(EQUAL, check(PPLBuiltinOperators.EQUALS_IP，SqlStdOperatorTable.EQUALS));
```
and leverage `check()` to manage the specific register logic via typeChecker.

`check()` can name to `case()` or `set()` or `queue()`...

ishaoxy: > Or modify to
> 
> ```
> registerOperator(EQUAL, check(PPLBuiltinOperators.EQUALS_IP，SqlStdOperatorTable.EQUALS));
> ```
> 
> and leverage `check()` to manage the specific register logic via typeChecker.
> 
> `check()` can name to `case()` or `set()` or `queue()`...

Modified,  improving codes readability.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/CidrMatchFunction.java
ishaoxy: Handled.

core/src/main/java/org/opensearch/sql/calcite/utils/UserDefinedFunctionUtils.java
```diff
@@ -115,7 +115,9 @@ public static SqlTypeName convertRelDataTypeToSqlTypeName(RelDataType type) {
         case EXPR_DATE -> SqlTypeName.DATE;
         case EXPR_TIME -> SqlTypeName.TIME;
         case EXPR_TIMESTAMP -> SqlTypeName.TIMESTAMP;
-        case EXPR_IP -> SqlTypeName.VARCHAR;
+          // EXPR_IP is mapped to SqlTypeName.OTHER since there is no
+          // corresponding SqlTypeName in Calcite.
+        case EXPR_IP -> SqlTypeName.OTHER;
```
LantaoJin, line 126: please update the PR description to align changes.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -473,40 +476,58 @@ private abstract static class AbstractBuilder {
...
     abstract void register(BuiltinFunctionName functionName, FunctionImp functionImp);
 
-    void registerOperator(BuiltinFunctionName functionName, SqlOperator operator) {
-      SqlOperandTypeChecker typeChecker;
-      if (operator instanceof SqlUserDefinedFunction udfOperator) {
-        typeChecker = extractTypeCheckerFromUDF(udfOperator);
-      } else {
-        typeChecker = operator.getOperandTypeChecker();
-      }
+    public void registerOperator(BuiltinFunctionName functionName, SqlOperator... operators) {
```
LantaoJin, line 499: please add a javadoc for this method to explain what are the operators for

ishaoxy: Added.

core/src/main/java/org/opensearch/sql/calcite/utils/UserDefinedFunctionUtils.java
ishaoxy: Thanks for reminding me.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/CompareIpFunction.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    public static boolean compare(Object obj1, Object obj2, ComparisonType comparisonType) {
+      try {
+        String ip1 = extractIpString(obj1);
+        String ip2 = extractIpString(obj2);
+        IPAddress addr1 = IPUtils.toAddress(ip1);
+        IPAddress addr2 = IPUtils.toAddress(ip2);
+        int result = IPUtils.compare(addr1, addr2);
+        return switch (comparisonType) {
```
qianheng-aws: For efficiency consideration, we'd better move this switch case to the `implement` method

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/CompareIpFunction.java
```diff
@@ -0,0 +1,127 @@
...
+          CompareImplementor.class,
+          "compare",
+          translatedOperands.get(0),
+          translatedOperands.get(1),
+          Expressions.constant(comparisonType));
+    }
+
+    public static boolean compare(Object obj1, Object obj2, ComparisonType comparisonType) {
+      try {
+        String ip1 = extractIpString(obj1);
```
qianheng-aws: Why not leverage `ExprIpValue::compare`? Then we don't need to do extra transformation for `ExprIPValue`

ishaoxy: Sorry for only focusing on IPUtils before and ignoring the functions encapsulated in ExprIpValue. Now fixed.

core/src/main/java/org/opensearch/sql/expression/function/udf/ip/CompareIpFunction.java
ishaoxy: Fixed.

LantaoJin (approved):

qianheng-aws (approved):

LantaoJin:
@ishaoxy please follow the instructions to backport manually

Issue #3384, opened 2025-03-06
[FEATURE] Calcite Engine Framework: pushdown aggregation with histogram bucket


qianheng-aws:
**Is your feature request related to a problem?**
Calcite Engine Framework: pushdown aggregation with histogram bucket

Issue #3386, opened 2025-03-06
[FEATURE] Calcite Engine Framework: pushdown aggregation with non-reference expression


qianheng-aws:
**Is your feature request related to a problem?**
Calcite Engine Framework: pushdown aggregation with non-reference expression

**What solution would you like?**
Use `script` API to do push down. And we also need to match `aggregate-project-scan`  pattern to do this optimization since Calcite aggregate only supports reference expression as its arguments.

PR #3823, opened 2025-06-26
Support span push down

qianheng-aws:
### Description
Support span push down

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3384

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -170,6 +184,12 @@ private static Pair<Builder, List<MetricParser>> processAggregateCalls(
     return Pair.of(metricBuilder, metricParserList);
   }
 
+  private static RexInputRef convertAggArgThroughProject(AggregateCall aggCall, Project project) {
+    RexNode argRex = project.getProjects().get(aggCall.getArgList().getFirst());
+    if (argRex instanceof RexInputRef rexInputRef) return rexInputRef;
+    else throw new IllegalArgumentException("Unsupported aggregate argument: " + argRex);
```
LantaoJin, line 190: the case like `avg(a+1)`? do we have an issue to track?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -231,21 +251,39 @@ private static Pair<ValuesSourceAggregationBuilder<?>, MetricParser> createRegul
...
+    if (rex instanceof RexInputRef rexInputRef) {
+      NamedFieldExpression groupExpr = fieldExpressionCreator.create(rexInputRef.getIndex());
+      return createTermsSourceBuilder(groupExpr);
+    } else if (rex instanceof RexCall rexCall
+        && rexCall.getKind() == SqlKind.OTHER_FUNCTION
+        && rexCall.getOperator().getName().equalsIgnoreCase(BuiltinFunctionName.SPAN.name())
+        && rexCall.getOperands().size() == 3
+        && rexCall.getOperands().getFirst() instanceof RexInputRef rexInputRef
+        && rexCall.getOperands().get(1) instanceof RexLiteral valueLiteral
+        && rexCall.getOperands().get(2) instanceof RexLiteral unitLiteral) {
```
LantaoJin, line 275: Can we create a static method in PlanUtil for this logic?

integ-test/src/test/resources/expectedOutput/calcite/explain_filter_agg_push.json
```diff
@@ -1,6 +1,6 @@
...
+    "physical": "EnumerableCalc(expr#0..2=[{inputs}], avg_age=[$t2], state=[$t0], city=[$t1])\n  CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[city, state, age], FILTER->>($2, 30), AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0, 1},avg_age=AVG($2))], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"size\":0,\"timeout\":\"1m\",\"query\":{\"range\":{\"age\":{\"from\":30,\"to\":null,\"include_lower\":false,\"include_upper\":true,\"boost\":1.0}}},\"_source\":{\"includes\":[\"city\",\"state\",\"age\"],\"excludes\":[]},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}],\"aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"sources\":[{\"state\":{\"terms\":{\"field\":\"state.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}},{\"city\":{\"terms\":{\"field\":\"city.keyword\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}}]},\"aggregations\":{\"avg_age\":{\"avg\":{\"field\":\"age\"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
```
LantaoJin, line 4: we need more test cases for:

- span(field, integer)
- span(field, decimal)
- span(timestamp, integer unit)
- span(date, integer unit)

The unit should contain at least "day, m, M, QUARTER" and "MS" (DSL unsupported)

qianheng-aws:
IT failed because we have different behavior for span(“1961-04-12 09:10:00”) between push down and non push down.
- Push down: it’s rounded to “1961-04-12 09:00:00"
- Non push down: it’s rounded to “1961-04-12 10:00:00” because the timestamp is a negative value as it’s before 1971

v2 has the same problem since the Calcite reuses the same logic as v2. Maybe we should fix this in another PR. For this PR, we will allow different results for this minor case and let IT pass.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: Yes, this one https://github.com/opensearch-project/sql/issues/3386

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: Just want to make use of the auto conversion in java21. If we extract these into a separate method, may need to do explicit conversion.

integ-test/src/test/resources/expectedOutput/calcite/explain_filter_agg_push.json
qianheng-aws: - span(timestamp, integer unit)
- span(date, integer unit)
will both hit the same code coverage.

So as to `day, m, QUARTER, MS`

qianheng-aws: I've tested with `MS`, its supported in DSL

```
{
    "from": 0,
    "size": 0,
    "timeout": "1m",
    "aggregations": {
        "composite_buckets": {
            "composite": {
                "size": 1000,
                "sources": [
                    {
                        "span(birthdate,1ms)": {
                            "date_histogram": {
                                "field": "l_receiptdate",
                                "missing_bucket": true,
                                "missing_order": "first",
                                "order": "asc",
                                "fixed_interval": "1ms"
                            }
                        }
                    }
                ]
            },
            "aggregations": {
                "count()": {
                    "value_count": {
                        "field": "_index"
                    }
                }
            }
        }
    }
}
```

qianheng-aws: - span(field, decimal)

Our SpanFunction doesn't support decimal currently, since linq4j divide will throw exception `Binary numeric promotion not possible on types "int" and "java.math.BigDecimal"`, and we implement it by using `field / interval * interval`.

Maybe support it in another PR later.

LantaoJin (dismissed):

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchAggregateIndexScanRule.java
```diff
@@ -56,11 +59,18 @@ public interface Config extends RelRule.Config {
                     b0.operand(LogicalAggregate.class)
                         .oneInput(
                             b1 ->
-                                b1.operand(CalciteLogicalIndexScan.class)
-                                    .predicate(
-                                        Predicate.not(OpenSearchIndexScanRule::isLimitPushed)
-                                            .and(OpenSearchIndexScanRule::noAggregatePushed))
-                                    .noInputs()));
+                                b1.operand(LogicalProject.class)
```
penghuo, line 62: Question: OpenSearchProjectIndexScanRule should push down project?

qianheng-aws: Yes, both rules will take effect. 

But because Calcite's optimizer is not doing transformation on one plan, but keep generating new plans, putting them in the RelSet, and it will choose the cheapest plan in the end, so OpenSearchProjectIndexScanRule pushing down project won't prevent this OpenSearchAggregateIndexScanRule.

For example, let's say we have original plan `Agg - Project - Scan`, it will generate a new plan `Agg - Scan` by OpenSearchProjectIndexScanRule and a new plan `Scan` by  OpenSearchAggregateIndexScanRule. So we have 3 plans in the RelSet [`Agg - Project - Scan`, `Agg - Scan`, `Scan`], and in the end the optimizer will choose `Scan` as the final plan since its cost is the cheapest.

penghuo (dismissed):

LantaoJin (approved):

PR #3826, opened 2025-06-26
Fix JOIN_TIME_OUT_HINT

ahkcs:
Fix JOIN_TIME_OUT_HINT


legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -105,6 +127,158 @@ public void run() throws IOException, SqlParseException {
...
   }
 
+  /**
+   * Extract JOIN_TIME_OUT hint value from the request builder
+   *
+   * @return TimeValue for custom keepalive, or null if no hint found
+   */
+  protected TimeValue extractJoinTimeoutFromHints() {
+    try {
+      LOG.info("🔍 DEBUG: Starting hint extraction");
```
Swiddis: suggestion: `LOG.debug`, no?

Although in this case [I'd use trace](https://stackoverflow.com/a/64806781/7543069)

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
```diff
@@ -19,10 +28,128 @@
...
+
+      // Access the private 'request' field from HashJoinQueryPlanRequestBuilder
+      java.lang.reflect.Field requestField =
+          planRequestBuilder.getClass().getDeclaredField("request");
+      requestField.setAccessible(true);
+      SqlRequest sqlRequest = (SqlRequest) requestField.get(planRequestBuilder);
+
+      if (sqlRequest != null) {
+        String originalSql = sqlRequest.getSql();
+        LOG.info("🔍 Original SQL: {}", originalSql);
```
Swiddis: todo (security): don't log the query directly, might contain confidential info.

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
```diff
@@ -19,10 +28,128 @@
...
+          return timeout;
+        } else {
+          LOG.info("⚠️ No JOIN_TIME_OUT hint found in SQL string");
+        }
+      } else {
+        LOG.warn("⚠️ SqlRequest is null");
+      }
+
+    } catch (Exception e) {
+      LOG.warn("⚠️ Error accessing original SQL request", e);
```
Swiddis: question: When does this ever happen? Can we do any better than catching `Exception`?

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
```diff
@@ -19,10 +28,128 @@
...
+      if (matcher.find()) {
+        String timeoutStr = matcher.group(1);
+        int timeoutSeconds = Integer.parseInt(timeoutStr);
+
+        LOG.info("✅ Parsed JOIN_TIME_OUT hint: {} seconds", timeoutSeconds);
+        return TimeValue.timeValueSeconds(timeoutSeconds);
+      }
+
+    } catch (Exception e) {
+      LOG.warn("Error parsing JOIN_TIME_OUT from SQL: {}", sql, e);
```
Swiddis: Same as above, when can this ever fail? And can we be more specific in our exceptions?

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
```diff
@@ -19,13 +19,15 @@
...
 import org.opensearch.common.action.ActionFuture;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.sql.legacy.esdomain.LocalClusterState;
 import org.opensearch.transport.client.Client;
 
 /** Handler for Point In Time */
 public class PointInTimeHandlerImpl implements PointInTimeHandler {
   private final Client client;
   private String[] indices;
+  private final TimeValue customKeepAlive;
```
Swiddis: suggestion: For type safety and implementation clarity, I think this should be an `Optional<TimeValue>`.

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
```diff
@@ -49,22 +65,26 @@ public PointInTimeHandlerImpl(Client client, String[] indices) {
...
+    LOG.info("Creating PIT with keepalive: {} ({}ms)", keepAlive, keepAlive.getMillis());
+
+    CreatePitRequest createPitRequest = new CreatePitRequest(keepAlive, false, indices);
     ActionFuture<CreatePitResponse> execute =
         client.execute(CreatePitAction.INSTANCE, createPitRequest);
     try {
       CreatePitResponse pitResponse = execute.get();
       pitId = pitResponse.getId();
-      LOG.info("Created Point In Time {} successfully.", pitId);
+      LOG.info("Created Point In Time {} with keepalive {} successfully.", pitId, keepAlive);
```
Swiddis, line 86: issue: Since the PIT ID is long, this is going to flood the logs for other oncalls.

At least we should make this `debug`, I wonder if we could go further and truncate the PIT id?

opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
```diff
@@ -6,8 +6,7 @@
 package org.opensearch.sql.opensearch.setting;
 
 import static org.opensearch.common.settings.Settings.EMPTY;
-import static org.opensearch.common.unit.TimeValue.timeValueDays;
-import static org.opensearch.common.unit.TimeValue.timeValueMinutes;
+import static org.opensearch.common.unit.TimeValue.*;
```
Swiddis: nit: rest of the team doesn't like wildcard imports, this was probably done automatically by the IDE. [You can disable it.]

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -87,7 +93,23 @@ public void sendResponse(RestChannel channel) throws IOException {
...
     try {
       long timeBefore = System.currentTimeMillis();
-      pit = new PointInTimeHandlerImpl(client, indices);
+
+      LOG.info("🔍 Starting join execution, checking for JOIN_TIME_OUT hints...");
+
+      // ✅ Extract JOIN_TIME_OUT hint and create PIT with custom keepalive
+      TimeValue customKeepAlive = extractJoinTimeoutFromHints();
+
+      if (customKeepAlive != null) {
```
Swiddis: thought: since `run` is already a long function, we should probably extract the PIT construction to its own method for clarity.

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -105,6 +127,158 @@ public void run() throws IOException, SqlParseException {
...
+            }
+          }
+        }
+      }
+
+      // Debug second table
+      TableInJoinRequestBuilder secondTable = requestBuilder.getSecondTable();
+      LOG.info("🔍 DEBUG: secondTable = {}", secondTable != null ? "exists" : "null");
+
+      if (secondTable != null) {
```
Swiddis: suggestion: duplicate of the above logic, we should break this into a function.

Swiddis (approved):

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
ahkcs: changed to debug

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
ahkcs: deleted

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
ahkcs: Added throw

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
ahkcs: Added throw

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
ahkcs: truncated

opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
ahkcs: fixed

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
ahkcs: Fixed

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
ahkcs: extracted

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
ahkcs: Changed

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -87,22 +94,149 @@ public void sendResponse(RestChannel channel) throws IOException {
...
+      throw e;
+    } catch (SqlParseException e) {
+      LOG.error("SQL parsing error during join query execution", e);
+      throw e;
+    } catch (RuntimeException e) {
+      LOG.error("Runtime error during join query execution", e);
       throw new IllegalStateException("Error occurred during join query run", e);
+    } catch (Exception e) {
+      LOG.error("Unexpected error during join query execution", e);
+      throw new IllegalStateException("Unexpected error occurred during join query run", e);
```
Swiddis: Log-rethrow is usually a bad pattern, obfuscates the original error and pollutes logs.

The rule of thumb I learned is "log the error where you handle it," since the outermost scope has the most context to make a useful error message, and you don't end up with a two-page stack trace filled with "Caused-by:". Try documenting the error contract of the function in a javadoc: what exceptions can it throw and what do they mean? Then in the callers of the function, review if it makes sense to handle the error there, or if it should propagate further up. (For best results, it may be best to define a custom exception type)

Eventually something should know how to handle the issue and either a) stop it from propagating to the user or b) format as in a human-readible user error that explains the problem and how to fix it.

ahkcs: removed Log-rethrow

Swiddis (approved):

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -87,24 +94,143 @@ public void sendResponse(RestChannel channel) throws IOException {
   public void run() throws IOException, SqlParseException {
     try {
       long timeBefore = System.currentTimeMillis();
-      pit = new PointInTimeHandlerImpl(client, indices);
-      pit.create();
+
+      LOG.debug("🔍 Starting join execution, checking for JOIN_TIME_OUT hints...");
+
+      // ✅ Create PIT with appropriate timeout
```
RyanL1997: we should remove these comments

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -87,24 +94,143 @@ public void sendResponse(RestChannel channel) throws IOException {
...
+      createPointInTimeWithCustomTimeout();
+
+      // Execute the query
       results = innerRun();
+
+      // Record execution time
```
RyanL1997: same here

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -87,24 +94,143 @@ public void sendResponse(RestChannel channel) throws IOException {
...
+  }
+
+  /** Create Point-in-Time with custom timeout if JOIN_TIME_OUT hint is present */
+  private void createPointInTimeWithCustomTimeout() {
+    Optional<TimeValue> customKeepAlive = extractJoinTimeoutFromHints();
+
+    if (customKeepAlive.isPresent()) {
+      TimeValue keepAlive = customKeepAlive.get();
+      LOG.info(
+          "✅ Using custom PIT keepalive from JOIN_TIME_OUT hint: {} seconds ({}ms)",
```
RyanL1997: nit: remove these emojis since it my cause some display problem for some terminals

Swiddis (dismissed):

Swiddis:
`PointInTimeHandlerImplTest` is failing in CI (along with those datetime tests that have been flaky since forever)

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
ahkcs: removed

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
ahkcs: removed

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
ahkcs: done

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
```diff
@@ -49,20 +67,26 @@ public PointInTimeHandlerImpl(Client client, String[] indices) {
...
+    LOG.info("Creating PIT with keepalive: {} ({}ms)", keepAlive, keepAlive.getMillis());
+
+    CreatePitRequest createPitRequest = new CreatePitRequest(keepAlive, false, indices);
     ActionFuture<CreatePitResponse> execute =
         client.execute(CreatePitAction.INSTANCE, createPitRequest);
     try {
       CreatePitResponse pitResponse = execute.get();
       pitId = pitResponse.getId();
-      LOG.info("Created Point In Time {} successfully.", pitId);
+      LOG.debug(
```
RyanL1997: why we switch to debug here instead?

ahkcs:
> `PointInTimeHandlerImplTest` is failing in CI (along with those datetime tests that have been flaky since forever)

Fixed

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
ahkcs: switched back

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
```diff
@@ -18,13 +27,109 @@
...
+      } else {
+        LOG.debug("SqlRequest is null");
+      }
+
+    } catch (Exception e) {
+      LOG.error("Error accessing original SQL request: {}", e.getMessage());
+    }
+
+    return Optional.empty();
+  }
```
dai-chen: `Config` is already in HashJoinQueryPlanRequestBuilder right? Why we need to do this? Is reflection really required?

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/logical/node/TableScan.java
```diff
@@ -21,9 +23,19 @@ public class TableScan implements LogicalOperator {
...
+  /** Configuration object for accessing custom settings */
+  private final Optional<Config> config;
+
   public TableScan(TableInJoinRequestBuilder request, int pageSize) {
     this.request = request;
     this.pageSize = pageSize;
+    this.config = Optional.empty();
+  }
+
+  public TableScan(TableInJoinRequestBuilder request, int pageSize, Config config) {
```
dai-chen: I see there is `hintLimit` in `TableInJoinRequestBuilder`. We can avoid all these changes by adding more hint values or `Config` to request builder?

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
ahkcs: I've refactored to parse the JOIN_TIME_OUT hint directly in the HashJoinQueryPlanRequestBuilder constructor. Reflection is now removed

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/logical/node/TableScan.java
ahkcs: I've updated the approach to follow the existing hintLimit pattern. Added hintConfig field to TableInJoinRequestBuilder (similar to hintLimit) and set it in the plan() method in HashJoinQueryPlanRequestBuilder.

Swiddis (dismissed):

dai-chen (commented):
Thanks for the changes! Just one last comment is UT / IT are both missing? If so, I think we should add both if possible.

noCharger (commented):
Please remove internal link in PR description.

noCharger:
> Thanks for the changes! Just one last comment is UT / IT are both missing? If so, I think we should add both if possible.

+1 on test coverage

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/logical/node/TableScan.java
```diff
@@ -2,7 +2,6 @@
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
  */
-
```
noCharger: remove lint changes if it's irrelevant

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/core/Config.java
```diff
@@ -133,4 +148,25 @@ public void configureTimeOut(Object[] params) {
...
+  }
+
+  /**
+   * Check if custom PIT keepalive is set
+   *
+   * @return true if custom timeout is configured
+   */
+  public boolean hasCustomPitKeepAlive() {
+    return customPitKeepAlive.isPresent();
+  }
```
noCharger: 1. can we remove log on getter/setter?
2. can we name methods consistently with the rest in this class?

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/HashJoinQueryPlanRequestBuilder.java
```diff
@@ -59,4 +84,51 @@ public QueryPlanner plan() {
...
+      LOG.debug("SQL string is null or empty, no JOIN_TIME_OUT hint to parse");
+      return Optional.empty();
+    }
+
+    try {
+      // Regex pattern to match /*! JOIN_TIME_OUT(number) */
+      Pattern pattern =
+          Pattern.compile(
+              "/\\*!\\s*JOIN_TIME_OUT\\s*\\(\\s*(\\d+)\\s*\\)\\s*\\*/", Pattern.CASE_INSENSITIVE);
+      Matcher matcher = pattern.matcher(sql);
```
noCharger: Using regex is concerning on correctness and performance; can we reuse `HintFactory` / `SqlParser`?

legacy/src/main/java/org/opensearch/sql/legacy/query/join/TableInJoinRequestBuilder.java
```diff
@@ -17,6 +20,7 @@ public class TableInJoinRequestBuilder {
   private List<Field> returnedFields;
   private Select originalSelect;
   private Integer hintLimit;
+  private Config hintConfig;
```
noCharger: Config has more fields than just hint. At this level this should be the same as `hintLimit `

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
```diff
@@ -19,13 +20,16 @@
...
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.sql.legacy.esdomain.LocalClusterState;
+import org.opensearch.sql.legacy.query.planner.core.Config;
 import org.opensearch.transport.client.Client;
 
 /** Handler for Point In Time */
 public class PointInTimeHandlerImpl implements PointInTimeHandler {
   private final Client client;
   private String[] indices;
+  private final Optional<Config> config;
```
noCharger: Config has more fields than just hint. We should reduce the scope

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -47,6 +49,7 @@ public abstract class ElasticJoinExecutor extends ElasticHitsExecutor {
   private final Set<String> aliasesOnReturn;
   private final boolean allFieldsReturn;
   protected final String[] indices;
+  protected final JoinRequestBuilder requestBuilder;
```
noCharger: If we just need the config, why need the whole request builder?

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
```diff
@@ -18,13 +22,50 @@
...
+    // Log the timeout configuration that was already parsed in the request builder
+    if (request.getConfig().hasCustomPitKeepAlive()) {
+      LOG.info(
+          "QueryPlanElasticExecutor: Using custom PIT keepalive from JOIN_TIME_OUT hint: {}"
+              + " seconds",
+          request
+              .getConfig()
+              .getCustomPitKeepAlive()
+              .map(timeout -> String.valueOf(timeout.getSeconds()))
+              .orElse("unknown"));
```
noCharger: why need this transformation here?

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
```diff
@@ -18,13 +22,50 @@
...
+      LOG.info(
+          "QueryPlanElasticExecutor: No JOIN_TIME_OUT hint found, using default PIT keepalive");
+    }
+
+    // Create the query planner (this will also set hint configs on table builders)
     this.queryPlanner = request.plan();
   }
 
+  @Override
+  public void run() throws IOException, SqlParseException {
```
noCharger: why need to override the execution here?

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/logical/node/TableScan.java
ahkcs: Removed

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/core/Config.java
ahkcs: Removed methods

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/HashJoinQueryPlanRequestBuilder.java
ahkcs: Removed the use of regex and reused HintFactory/SqlParser

legacy/src/main/java/org/opensearch/sql/legacy/query/join/TableInJoinRequestBuilder.java
ahkcs: Removed hintConfig, using ```TimeValue hintJoinTimeout```

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
ahkcs: Reduced scope to only TimeValue

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
ahkcs: Removed

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
ahkcs: Removed the run() method

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/QueryPlanElasticExecutor.java
ahkcs: Removed transformation

integ-test/src/test/java/org/opensearch/sql/legacy/JoinTimeoutHintIT.java
```diff
@@ -0,0 +1,340 @@
...
+                + "FROM %s a JOIN %s d ON d.holdersName = a.firstname LIMIT 3",
+            TEST_INDEX_PEOPLE,
+            TEST_INDEX_DOG);
+
+    try {
+      JSONObject result = executeQuery(query);
+      assertNotNull("Basic join without hints should work", result);
+
+      if (result.has("error")) {
+        System.err.println("Basic join failed: " + result.getJSONObject("error"));
```
dai-chen: Please remove all system print. Meanwhile I assume this base case is covered in other tests. Probably you can focus on normal and edge cases for timeout hint in this new test class.

integ-test/src/test/java/org/opensearch/sql/legacy/JoinTimeoutHintIT.java
```diff
@@ -0,0 +1,340 @@
...
+      System.err.println("Basic join failed: " + e.getMessage());
+      e.printStackTrace();
+      // Don't throw here - we want to understand what's happening
+    }
+  }
+
+  /** Test that different JOIN_TIME_OUT values are parsed and applied correctly */
+  @Test
+  public void testJoinTimeoutHintWithDifferentValues() throws IOException {
+    int[] timeoutValues = {30, 60, 120, 300}; // Reduced set for faster testing
```
dai-chen: If you want, you can check JUnit parameterized test. I recall we have such test for your reference, but they're mostly UT. If you've covered this in UT, I think you can simplify this IT to avoid overlap.

integ-test/src/test/java/org/opensearch/sql/legacy/JoinTimeoutHintIT.java
ahkcs: Removed, trimmed IT test

integ-test/src/test/java/org/opensearch/sql/legacy/JoinTimeoutHintIT.java
ahkcs: Removed, trimmed IT test

integ-test/src/test/java/org/opensearch/sql/legacy/JoinTimeoutHintIT.java
```diff
@@ -0,0 +1,229 @@
...
+    assertNotNull("Query with large timeout should execute", result);
+
+    int resultsCount = getResultsCount(result);
+    assertTrue("Should handle large timeout value", resultsCount >= 0);
+    assertTrue("Should respect LIMIT", resultsCount <= 5);
+  }
+
+  /** Consistency test: Verify repeated queries with JOIN_TIME_OUT produce consistent results */
+  @Test
+  public void testJoinTimeoutHintConsistency() throws IOException {
```
dai-chen: This may become flaky because no deterministic guarantee either in index scan or limit operator?

ahkcs: Removed test case, trimmed further

legacy/src/main/java/org/opensearch/sql/legacy/query/join/TableInJoinRequestBuilder.java
```diff
@@ -59,4 +61,16 @@ public Integer getHintLimit() {
...
+    return hintJoinTimeout;
+  }
+
+  public void setHintJoinTimeout(TimeValue hintJoinTimeout) {
+    this.hintJoinTimeout = hintJoinTimeout;
+  }
+
+  public boolean hasJoinTimeoutHint() {
+    return hintJoinTimeout != null;
+  }
```
noCharger: do we really need this method since getHintJoinTimeout can be used

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/physical/node/pointInTime/PointInTime.java
```diff
@@ -70,4 +89,10 @@ protected void loadNextBatch() {
               .get();
     }
   }
+
+  private String truncatePitId(String pitId) {
+    if (pitId == null) return "null";
+    if (pitId.length() <= 20) return pitId;
+    return pitId.substring(0, 20);
+  }
```
noCharger: why truncate it?

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
```diff
@@ -87,21 +84,12 @@ public void sendResponse(RestChannel channel) throws IOException {
...
     } catch (Exception e) {
       LOG.error("Failed during join query run.", e);
       throw new IllegalStateException("Error occurred during join query run", e);
-    } finally {
-      try {
-        pit.delete();
-      } catch (RuntimeException e) {
-        Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
-        LOG.info("Error deleting point in time {} ", pit);
-      }
```
noCharger: why this finally block removed?

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
```diff
@@ -19,13 +20,15 @@
...
 import org.opensearch.common.action.ActionFuture;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.sql.legacy.esdomain.LocalClusterState;
 import org.opensearch.transport.client.Client;
 
 /** Handler for Point In Time */
 public class PointInTimeHandlerImpl implements PointInTimeHandler {
   private final Client client;
   private String[] indices;
+  private final Optional<TimeValue> customKeepAlive;
```
noCharger, line 32: is it necessary to use optional?

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/physical/node/pointInTime/PointInTime.java
```diff
@@ -35,8 +42,20 @@ public void close() {
...
+    if (request.hasJoinTimeoutHint()) {
+      TimeValue customTimeout = request.getHintJoinTimeout();
+      LOG.info(
+          "PointInTime: Creating PIT with JOIN_TIME_OUT hint: {} seconds",
+          customTimeout.getSeconds());
+      pit =
+          new PointInTimeHandlerImpl(
+              client, request.getOriginalSelect().getIndexArr(), customTimeout);
+    } else {
+      LOG.info("PointInTime: Creating PIT with default settings");
```
noCharger:
```suggestion
      LOG.info("PointInTime: Creating PIT with default timeout value: {}");
```

noCharger:
Let's resolve the conflicts before merging

legacy/src/main/java/org/opensearch/sql/legacy/executor/join/ElasticJoinExecutor.java
ahkcs: Restored

legacy/src/main/java/org/opensearch/sql/legacy/pit/PointInTimeHandlerImpl.java
ahkcs: It’s not strictly necessary, but using Optional<TimeValue> makes it explicit that the value may be absent, which improves code readability and helps avoid null checks, it was added due to @Swiddis's comment

legacy/src/main/java/org/opensearch/sql/legacy/query/planner/physical/node/pointInTime/PointInTime.java
ahkcs: Removed

legacy/src/main/java/org/opensearch/sql/legacy/query/join/TableInJoinRequestBuilder.java
ahkcs: Removed

noCharger (approved):
Thanks for the change!

Issue #720, opened 2022-08-02
[BUG] Expressions in PPL require whitespace around * characters


kroepke:
**What is the bug?**
If a multiplication expression is written without whitespace around the operator, the lexer mistakenly tokenizes it as `ID_LITERAL`.
This leads to unexpected behavior (and IMHO wrong parsing) wherever multiplication is allowed.

**How can one reproduce the bug?**
1. Execute the PPL query `search source=opensearch_dashboards_sample_data_logs | stats avg(bytes*1024)`
2. The query fails with `org.opensearch.sql.exception.SemanticCheckException: can't resolve Symbol(namespace=FIELD_NAME, name=bytes*1024) in type env`
3. Change the query to `search source=opensearch_dashboards_sample_data_logs | stats avg(bytes * 1024)` (note the surrounding whitespace)
4. Run it again, the query succeeds

**What is the expected behavior?**
An expression should parse successfully with or without optional whitespace present, the queries above should behave identically.
If certain identifiers can contain wildcards those should be a different token, as they are not allowed everywhere where a plain ID is allowed at.

**What is your host/environment?**
 - Version 1.3, 2.1
 - Plugins SQL since 1.13.2.0 up until current

**Do you have any additional context?**
The change was introduced in https://github.com/opensearch-project/sql/commit/7dc8699d62180c803fd5147d03331b787360d9a0#diff-fbfc17858912dc1b9fbe756efdc25267d6eb4f0f6e4b9d4fc1500d072f526892 and [tests were deliberately changed](https://github.com/opensearch-project/sql/commit/7dc8699d62180c803fd5147d03331b787360d9a0#diff-30c2b9f0820c1e4d867c7c2b14c35bad2db5fa7395b81f59a9130b9f98180594R82) to include whitespace to work around the lexer issue.
I believe this is wrong behavior, which is very difficult to understand for users.

As far as I can tell from the code, the intention was to allow wildcards and other non-numeric and non-alpha characters for index names. If that's the case the grammar should reflect the two different uses and not treat them as the same token type.

Unfortunately with the way things are now it becomes essentially impossible to disambiguate in the lexer.

kroepke:
Failing test in `org.opensearch.sql.ppl.antlr.PPLSyntaxParserTest`

```Java
  @Test
  public void testLiteralSwallowsExpression() {
    ParseTree tree = new PPLSyntaxParser().parse("source=foo | stats avg(bytes*1024)");
    assertNotEquals(null, tree);
[...]

PR #3830, opened 2025-06-30
Support multiply without blanks

qianheng-aws:
### Description
Support multiply without blanks.

Implements it by making `ID` not allow containing `*`. 

But since we should support index patterns which contains `*`, we have to make compromise that wrapping it with backticks. However, for the most common case where the index pattern ends with `*`, it doesn't need wrapping with backticks as we add this specific pattern in the `tableIdent`. 

For example:

```
source=index* | eval  a = b*10
```

Or 

```
source=`*index*2020*` | eval a = b*10
```

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/720

### Check List

### DIFF
--- a/ppl/src/test/java/org/opensearch/sql/ppl/antlr/PPLSyntaxParserTest.java
+++ b/ppl/src/test/java/org/opensearch/sql/ppl/antlr/PPLSyntaxParserTest.java
@@ -341,7 +341,7 @@ public class PPLSyntaxParserTest {
     exceptionRule.expectMessage(
         StringContainsInOrder.stringContainsInOrder(
             "[=] is not a valid term at this part of the query: 'describe source=' <-- HERE.",
-            "Expecting tokens:"));
+            "Expecting one of"));
 
     new PPLSyntaxParser().parse("describe source=t");
   }


ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -1099,7 +1099,8 @@ ident
    ;
 
 tableIdent
-   : (CLUSTER)? ident
+   // specially allow index name ends with * without wrapping it with backticks
+   : (CLUSTER)? (ident | ID STAR)
```
LantaoJin: I am not sure the _table-ident-end-with-star_ is the only one _table-ident-with-star_ pattern we can ignore backticks.
In another word, this makes a breaking change that blocks the use case of
```
source=table*.2025*
```

qianheng-aws: This PR will no doubt introduce breaking change. 

I just want to mitigate the gap for the most common case where only the table idents end ends with star. If we think cluster ident ends with * is also widely used, I may need to find a way to support this case without backticks as well.

qianheng-aws: If the breaking change for the case where star in the middle of table ident is also unacceptable, the current approach won't make it. Then we need to find another way.

penghuo: We should support `source=table*.2025*`.
And we should differenate 
1. wildcard-field vs field. currently, PPL do not support wildcard-field, but it is needed https://github.com/opensearch-project/sql/pull/3830/files
2. wildcard-string, e.g. `source=table*.2025*`

qianheng-aws: 1. Do we have plan to support wildcard-field in the future? It's also supported by current g4 with adding backticks
2. It can work around by adding backticks 
```
source=`table*.2025*`
```

LantaoJin: > It can work around by adding backticks

I don't think the current implementation is a optimal fixing to merge, actually the current issue also contains a workaround `a = b * 10`. The patching doesn't make thing better IMO.

qianheng-aws: Find a way to support source=table*.2025*, without adding backticks. It will eliminate the breaking change with the latest code. @LantaoJin @penghuo

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/720.yml
```diff
@@ -32,7 +32,7 @@ teardown:
         Content-Type: 'application/json'
       ppl:
         body:
-          query: 'source=test* | eval a = balance*10 | fields a'
+          query: 'source= *te*st* | eval a = balance*10 | fields a'
```
LantaoJin, line 35: why add a SPACE after `=`? can it be `source=*te*st*`?

LantaoJin: please add a case:
```
source=*te*st* | eval a = balance*10 | fields a
source=`*te*st*` | eval a = balance *10 | fields a
source=*te*st*:`*te*st*` | eval a = balance* 10 | fields a
```

ppl/src/main/antlr/OpenSearchPPLParser.g4
LantaoJin: > wildcard-field vs field. currently, PPL do not support wildcard-field, but it is needed

Seems the `*` can be used in field name in OpenSearch:
```
PUT test002
{
  "mappings": {
    "properties": {
      "gender": {
        "type": "text",
        "fielddata": true
      },
      "updated_at": {
        "type": "date",
        "format": "strict_date_time"
      },
      "*id": {
        "type": "integer"
      }
    }
  }
}

POST test002/_bulk
{"create": {}}
{"gender":"M","updated_at":"2013-07-15T10:41:34Z", "*id": 1}
{"create": {}}
{"gender":"M","updated_at":"2013-07-15T10:42:34Z", "*id": 1}
{"create": {}}
{"gender":"F","updated_at":"2013-07-15T10:43:34Z", "*id": 1}
{"create": {}}
{"gender":"F","updated_at":"2013-07-15T10:44:34Z", "*id": 1}

POST /_plugins/_ppl/
{
  "query": """
  source = test002 | eval `10 * *id` = 10 * *id | fields `10 * *id`
  """
}

{
  "schema": [
    {
      "name": "10 * *id",
      "type": "int"
    }
  ],
  "datarows": [
    [
      10
    ],
    [
      10
    ],
    [
      10
    ],
    [
      10
    ]
  ],
  "total": 4,
  "size": 4
}
```

I think we should just call out in doc that the blanks are required for multiply operator in PPL/SQL syntax.

qianheng-aws:
> wildcard-field vs field. currently, PPL do not support wildcard-field, but it is needed

Since we need to support `wildcard-field` as well, it's not easy to fix this issue without introducing breaking change. 

Will close this PR

ppl/src/main/antlr/OpenSearchPPLParser.g4
qianheng-aws: Since we need to support wildcard-field as well, it's not easy to fix this issue without introducing breaking change.

Will close this PR

Issue #3839, opened 2025-07-02
[FEATURE][Calcite] Support filter push down for search Sarg


qianheng-aws:
**Is your feature request related to a problem?**
When Calcite enabled, it cannot push down filter like:

```
... | where a >= 1 and a <= 10 |...
```

Calcite use Sarg value(i.e. Range[1, 10]) to express `a >= 1 and a <= 10`, and it doesn't support translate such value in its `PredicateAnalyzer`, so it doesn't support push down.

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

PR #3840, opened 2025-07-02
Support filter push down for Sarg value

qianheng-aws:
### Description
Support filter push down for Sarg value

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3839

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -416,8 +405,26 @@ private QueryExpression binary(RexCall call) {
         case SEARCH:
           if (isSearchWithComplementedPoints(call)) {
             return QueryExpression.create(pair.getKey()).notIn(pair.getValue());
-          } else {
+          } else if (isSearchWithPoints(call)) {
             return QueryExpression.create(pair.getKey()).in(pair.getValue());
+          } else {
+            Sarg<?> sarg = pair.getValue().literal.getValueAs(Sarg.class);
+            Set<? extends Range<?>> rangeSet = requireNonNull(sarg).rangeSet.asRanges();
```
LantaoJin, line 517: After fallback being disabled in future, `requireNonNull` means user will meet NPE which is unrecoverable case when the `sarg` is null here. For pushdown, as an enhancement, maybe just throw `PredicateAnalyzerException`.

qianheng-aws: We have `catch (Throwable e)` in the `analyze` method which is the outermost shell for PredicateAnalyzer. So it will always catch all exception including NPE and re-throw `ExpressionNotAnalyzableException`.

And I think it's just a defensive check, `isSearchWithPoints`, `isSearchWithComplementedPoints` both has similar code.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -416,8 +405,26 @@ private QueryExpression binary(RexCall call) {
...
+            boolean isTimeStamp =
+                (pair.getKey() instanceof NamedFieldExpression namedField)
+                    && ExprCoreType.TIMESTAMP.equals(
+                        namedField.getExprType() instanceof OpenSearchDataType osType
+                            ? osType.getExprCoreType()
+                            : namedField.getExprType());
```
penghuo: add isTimeStamp() function in NamedFieldExpression, similar to isTextType()

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -925,6 +942,37 @@ public QueryExpression notIn(LiteralExpression literal) {
       builder = boolQuery().mustNot(termsQuery(getFieldReferenceForTermQuery(), collection));
       return this;
     }
+
+    @Override
+    public QueryExpression between(Range<?> range, boolean isTimeStamp) {
```
penghuo, line 1162: what if the Sarg include Range.singleton, between logic works?
for instance, `... | where (age>20 and age< 30) or age=35  ==> Sarg(Range.closed(20, 30), Range.singleton(35))`

qianheng-aws: It still use RangeQueryBuilder for point 35(or Range[35, 35]), currently and will be `Range: {from: 35, to: 35}`.

Maybe using TermQuery will be more efficiency. Will do that change.

PR #3834, opened 2025-07-01
Support relevance query functions pushdown implementation in Calcite

songkant-aws:
### Description
Support relevance query functions pushdown implementation in Calcite

### Related Issues
Resolves #3462 

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3851, opened 2025-07-07
[BugFix] Fix relevance query function over optimization issue in ReduceExpressionsRule

songkant-aws:
### Description
Fix relevance query function over optimization issue in ReduceExpressionsRule

### Related Issues
Related to https://github.com/opensearch-project/sql/pull/3834

### Check List


qianheng-aws (approved):

core/src/main/java/org/opensearch/sql/expression/function/UserDefinedFunctionBuilder.java
```diff
@@ -43,4 +43,30 @@ default SqlUserDefinedFunction toUDF(String functionName) {
...
+  /**
+   * In some rare cases, we need to call out the UDF to be not deterministic to avoid Volcano
+   * planner over-optimization. For example, we don't need ReduceExpressionsRule to optimize
+   * relevance query UDF.
+   *
+   * @param functionName UDF name to be registered
+   * @param isDeterministic Specified isDeterministic flag
+   * @return Calcite SqlUserDefinedFunction
+   */
+  default SqlUserDefinedFunction toUDF(String functionName, boolean isDeterministic) {
```
LantaoJin, line 48: can you combine the methods toUDF in L35? We don't want to keep diff implementations.

integ-test/src/test/java/org/opensearch/sql/ppl/RelevanceFunctionIT.java
```diff
@@ -144,13 +144,38 @@ public void test_mixed_relevance_function_and_normal_filter() throws IOException
...
+  }
+
   @Test
   public void not_pushdown_throws_exception() throws IOException {
     String query1 =
         "SOURCE="
             + TEST_INDEX_BEER
-            + " | STATS count(AcceptedAnswerId) as idCount"
-            + " | WHERE simple_query_string(['Tags'], 'taste') and idCount > 200";
+            + " | EVAL answerId = AcceptedAnswerId + 1"
```
LantaoJin, line 177: why we change this IT case?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1068,4 +1073,14 @@ void populate() {
                   ctx.relBuilder));
     }
   }
+
+  private static RexNode derefMapCall(RexNode rexNode) {
```
LantaoJin: can you move this method to `PlanUtil`?

core/src/main/java/org/opensearch/sql/expression/function/UserDefinedFunctionBuilder.java
songkant-aws: Done

integ-test/src/test/java/org/opensearch/sql/ppl/RelevanceFunctionIT.java
songkant-aws: Because I think the previous test case semantic is ambiguous `simple_query_string` can't query on target field right after an aggregation. Use eval operator to simulate a filter that can't be pushed down makes a lot sense.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
songkant-aws: Done

LantaoJin (approved):

qianheng-aws (approved):

PR #3854, opened 2025-07-07
Default to UTC for date/time functions across PPL and SQL

### LINKED ISSUE #3725: [BUG] PPL date modifier function does not work as expected
**What is the bug?**
PPL now() with date modifier does not work as expected.  
* Run PPL and DSL queries at PST 2025-06-02 13:19:00. JVM timezone is America/Los_Angeles.
* PPL query return empty results
```
###

{
    "query": "source=t0001 | where @timestamp >= date_sub(now(), INTERVAL 1 DAY) and @timestamp <= now()"
}

"datarows": [],
```

* DSL query return expected result
```
###
POST {{baseUrl}}/t0001/_search

{
  "query": {
      "range" : {
          "@timestamp" : {
              "gte" : "now-1d",
              "lte" : "now"
          }
      }
  }
}

"hits": {
    "hits": [
      {
        "_source": {
          "@timestamp": "2025-06-02T14:00:00"
        }
      },
      {
        "_source": {
          "@timestamp": "2025-06-02T15:00:00"
        }
      },
      {
        "_source": {
          "@timestamp": "2025-06-02T16:00:00"
        }
      }
    ]
  }
```

```
###
PUT {{baseUrl}}/t0001

{
  "mappings": {
    "properties": {
      "@timestamp": { "type": "date" }
    }
  }
}

###
POST {{baseUrl}}/t0001/_bulk

{"index": {}}
{"@timestamp": "2025-06-02T14:00:00"}
{"index": {}}
{"@timestamp": "2025-06-02T15:00:00"}
{"index": {}}
{"@timestamp": "2025-06-02T16:00:00"}
```

yuancu:
### Description

Datetime functions in PPL/SQL has been using cluster's JVM time zone as the default timezone. For example, if the cluster resides in PST and it's 13:00:00 there, `SELECT CURRENT_TIME()` will return the the current timestamp as `13:00:00`.

However, this creates a mismatch with the data stored in OpenSearch, as `[dates are internally stored in the UTC format](https://docs.opensearch.org/docs/latest/field-types/supported-field-types/date/)`. Therefore, when comparing dates generated with functions with those stored in OpenSearch, there will be a mismatch in their timezone. Examples are detailed in #3725.

This PR unifies all date / time representations in PPL / SQL to UTC. As a result, `SELECT CURRENT_TIME()` will return the current time in UTC timezone; in query: `... | eval t = timestamp('2025-07-07 13:00:00') | where date > t`, `t` will be regarded as `2025-07-07T13:00:00.000Z` in UTC.

This will fix existing problems of comparing with now, where the current zoned timestamp in the cluster's time zone will be directly compared with an UTC timestamp from the index.

As the next step, we should allow users to specify timezone per query or cluster-wise.

### Alternative solutions that I have considered

We can still use zoned time in PPL / SQL, but only make the conversion when interacting with the data from the index. This includes:
- Convert the UTC date to cluster timezone when read from OpenSearch
- When comparing with `now()` or literal, pushing down the time zone as a parameter as well (or convert the now / literal to UTC before pushing down).
- ...

I did not implement this solution since users will read a data that's different from what they stored, and they have no control over it.

### Comparison with other databases / languages

- Most SQL databases have session controls. Date/time are returned in the session's time zone. If time zone not specified, dates will be returned in UTC time zone.
- Elasticsearch supports `time_zone` argument in some DSL queries like range query, which will convert the date/time in the query to UTC before comparing with dates in the index. If not specified, all dates are regarded as in UTC time zone. E.g. `SELECT NOW()` will return a UTC time.
- [Splunk treats date/time literals that did not specify a time zone as in cluster's time zone](https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-manual/dates-and-time/time-zones). Date / time are converted to UNIX time for processing.

### Related Issues
Resolves #3725

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/datetime/DateTimeFunctions.java
+++ b/core/src/main/java/org/opensearch/sql/expression/datetime/DateTimeFunctions.java
@@ -81,6 +81,7 @@ import org.opensearch.sql.expression.function.FunctionResolver;
 import org.opensearch.sql.expression.function.FunctionSignature;
 import org.opensearch.sql.expression.function.SerializableFunction;
 import org.opensearch.sql.expression.function.SerializableTriFunction;
+import org.opensearch.sql.utils.DateTimeFormatters;
 import org.opensearch.sql.utils.DateTimeUtils;
 
 /**
@@ -1280,8 +1281,13 @@ public class DateTimeFunctions {
       ExprValue startingDateTime, ExprValue fromTz, ExprValue toTz) {
     if (startingDateTime.type() == ExprCoreType.STRING) {
       try {
-        startingDateTime = new ExprTimestampValue(startingDateTime.stringValue());
-      } catch (ExpressionEvaluationException e) {
+        // CONVERT_TZ only expects a timestamp in the format "yyyy-MM-dd HH:mm:ss[.SSSSSSSSS]".
+        startingDateTime =
+            new ExprTimestampValue(
+                LocalDateTime.parse(
+                    startingDateTime.stringValue(),
+                    DateTimeFormatters.DATE_TIME_FORMATTER_VARIABLE_NANOS));
+      } catch (DateTimeParseException e) {
         return ExprNullValue.of();
       }
     }


docs/user/ppl/functions/datetime.rst
```diff
@@ -1303,7 +1303,7 @@ NOW
 Description
 >>>>>>>>>>>
 
-Returns the current date and time as a value in 'YYYY-MM-DD hh:mm:ss' format. The value is expressed in the cluster time zone.
+Returns the current date and time as a value in 'YYYY-MM-DD hh:mm:ss' format. The value is expressed in the UTC time zone.
```
LantaoJin, line 1313: I don't think this changes should impact any SYS* datetime functions. SYSDATE() should be expressed in the same timezone as the server is on.

We want to change the current_timestamp/now to UTC since we don't have a session in OS. But would not change the SYS* functions. ref https://database-heartbeat.com/2021/09/28/sysdate-vs-current_date-in-oracle-database/

docs/user/ppl/functions/datetime.rst
```diff
@@ -1303,7 +1303,7 @@ NOW
...
 `NOW()` returns a constant time that indicates the time at which the statement began to execute. This differs from the behavior for `SYSDATE() <#sysdate>`_, which returns the exact time at which it executes.
```
LantaoJin, line 1314: we need update user doc to highlight the new behaviours for any impacted functions. for example, `now`

docs/user/ppl/functions/datetime.rst
yuancu: Set `sysdate` to use cluster time zone.

Originally, `sysdate` only differs with `now` in that it returns execution timestamp instead of query start timestamp. Now it differs with `now` in one more way (time zone). I've made it clear in the doc. Is this change expected?

Two more functions that worry me are `localtime` and `localtimestamp`. They'll return UTC timestamp instead of *local* timestamp. This may be confusing for users. I think it's because we followed [the function definition of MySql](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_localtimestamp), where it says *[LOCALTIME](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_localtime) and [LOCALTIME()](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_localtime) are synonyms for [NOW()](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_now)*. However, since we have altered the definition of `now`, the `local*` functions no longer points to the *local* time zone.

docs/user/ppl/functions/datetime.rst
yuancu: Fixed. Explicitly mentioned UTC time-zone for all now-like functions.

docs/user/ppl/functions/datetime.rst
LantaoJin: It's okey as long as keep the behaviours of localtime() as same as now(). maybe a better way is marking local*  functions deprecated since now we don't have any session management.

LantaoJin (approved):

penghuo:
> This will be a break change, and may be counter-intuitive for some since users will get a timestamp that's different from their local timestamp.

Could we avoid breaking change by providing timezone setting? e.g. default value is jvm timezone. 

> As the next step, we should allow users to specify timezone per query or per session.

Timezone query paramater will overide timezone setting, right?

> Default to UTC for date/time functions across PPL and SQL

I agree, timestamp Intenal representation should always be epoch value. 
my proposal is apply timezone for display and parsing timestamp string literal only

yuancu:
> Timezone query paramater will overide timezone setting, right?

Yes

> My proposal is apply timezone for display and parsing timestamp string literal only

In this case, if a user feed in a timestamp string, should we regard it as a UTC timestamp or a timestamp at user's time zone?

I feel it will be more intuitive to treat it as  in user's time zone. Since a user will get what they stores when they read it back from the index.

```
             -------convert---> |￣￣￣￣￣￣￣￣|
user (local)                    | index (UTC)  |
             <----convert back- |______________|
```

The downside: It makes the user interface different from what's stored in the index. This may be a little weird for a query language.

> Could we avoid breaking change by providing timezone setting? e.g. default value is jvm timezone.

I am considering adding such a setting. There may be three ways to achieve so:
- setting per query
- setting per session
- a cluster-wide setting

They all have their merits and disadvantages. It may be too verbose or troublesome if a user has to specify a timezone per query. Enabling cluster-wise setting allows users to specify timezone once and use it forever, although clients from different time zones have to use the same time-zone set by the server. Using session-timezone / profile-timezone may be a great middle ground. However, introducing such an involved concept purely for timezone settings may be an overkill.

Of course, we can use a combination of them -- e.g. enable cluster-wide timezone setting, but also support overriding it with query parameters.

docs/user/ppl/functions/datetime.rst
```diff
@@ -244,8 +244,9 @@ CURDATE
 Description
 >>>>>>>>>>>
 
-Returns the current time as a value in 'YYYY-MM-DD'.
-CURDATE() returns the time at which it executes as `SYSDATE() <#sysdate>`_ does.
+Returns the current date as a value in 'YYYY-MM-DD' format.
+CURDATE() returns the current UTC date at the time the statement is executed.
```
penghuo: rephrase it, --> CURDATE() returns the current date in UTC at the time the statement is executed.

docs/user/ppl/functions/datetime.rst
```diff
@@ -244,8 +244,9 @@ CURDATE
 Description
```
penghuo, line 250: We should add a Note section at Top of datatime to describe UTC behaviour.

docs/user/ppl/functions/datetime.rst
```diff
@@ -1594,7 +1595,7 @@ Description
 >>>>>>>>>>>
 
 Returns the current date and time as a value in 'YYYY-MM-DD hh:mm:ss[.nnnnnn]'.
-SYSDATE() returns the time at which it executes. This differs from the behavior for `NOW() <#now>`_, which returns a constant time that indicates the time at which the statement began to execute.
+SYSDATE() returns the time at which it executes in the cluster's time zone. This differs from the behavior for `NOW() <#now>`_, which returns a constant time that indicates the time at which the statement began to execute in the UTC time zone.
```
penghuo: > SYSDATE() returns the time at which it executes in the cluster's time zone.

It should returns the current date in UTC.

core/src/test/java/org/opensearch/sql/expression/datetime/DateTimeTest.java
```diff
@@ -53,9 +53,7 @@ public void localDateTimeConversion() {
     String timeZone = "America/Los_Angeles";
     LocalDateTime timeConverted = LocalDateTime.parse(dt, formatter);
     ZonedDateTime timeZoneLocal =
-        timeConverted
-            .atZone(ZoneId.of(TimeZone.getDefault().getID()))
-            .withZoneSameInstant(ZoneId.of(timeZone));
+        timeConverted.atZone(ZoneOffset.UTC).withZoneSameInstant(ZoneId.of(timeZone));
     FunctionExpression expr = DSL.datetime(DSL.literal(dt), DSL.literal(timeZone));
```
penghuo, line 57: Timezone "America/Los_Angeles" is only for testing right? By default DSL.datetime always use UTC timezone?

yuancu: Yes, "America/Los_Angeles" is only for testing. `DSL.datetime(datetime, timezone)` converts the datetime (assumed to be in UTC) to the specified timezone. If there isn't a timezone specified -- `DSL.datetime(datetime)`, it uses UTC timezone.

docs/user/ppl/functions/datetime.rst
yuancu: @LantaoJin [argued](https://github.com/opensearch-project/sql/pull/3854#discussion_r2191394393) that `SYSDATE` should return the current date at server's time zone, as how [Oracle's SYSDATE ](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/SYSDATE.html)does.

In [MySql](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_sysdate), SYSDATE does not differ with NOW w.r.t. timezone.

docs/user/ppl/functions/datetime.rst
yuancu: Added a block with the node directive. Do you think it is detailed enough?

docs/user/ppl/functions/datetime.rst
yuancu: Fixed

core/src/test/java/org/opensearch/sql/expression/datetime/DateTimeTest.java
yuancu: One proposal related to this test: I think `ExprTimestampValue` should hold a `LocalTimestamp` instead of an `Instant`, since it just represents a timestamp without zone information.

For example, the function `datetime(datetime, timezone)` converts a timestamp to another timezone. But if representing the input and output timestamp with an `Instant`, it will remain the same. In the code, there are also many spoiler plates to add a timezone to a LocalDateTime in order to work with ExprTimestampValue.

docs/user/ppl/functions/datetime.rst
LantaoJin: I have checked the behaviours in Oracle Autonomous Database. The default time zone is UTC and by default calls to SYSDATE and SYSTIMESTAMP return the date and time in UTC. [ref](https://docs.oracle.com/en/cloud/paas/autonomous-database/serverless/adbsb/autonomous-initialization-parameters.html#GUID-1D5E830F-2986-4E6C-AF8F-899AC3C85D07). let's update it to UTC. @yuancu

yuancu: Updated the implementation.

Besides, I found a bug which may be annoying for `SYSDATE` users -- multiple calls to `SYSDATE` with the same parameter within a query will return the same value.

`SYSDATE` is expected to return different values when it is called within the same query. 
E.g. `sd1` and `sd2` are different in this query: `source=dates | eval sd1 = sysdate(5) | sort date | eval sd2 = sysdate(6) | fields sd1, sd2`

| sd1 | sd2  |
| ----| ---- |
| 2025-07-22 06:16:33.13648 | 2025-07-22 06:16:33.136559 |

However, this query gives the same result: `source=dates | eval sd1 = sysdate(6) | sort date | eval sd2 = sysdate(6) | fields sd1, sd2`

| sd1 | sd2  |
| ----| ---- |
| 2025-07-22 06:19:07.006268 | 2025-07-22 06:19:07.006268 |

The physical plan gives a hint to the problem:
```
EnumerableCalc(expr#0=[{inputs}], expr#1=[6], expr#2=[SYSDATE($t1)], sd1=[$t2], sd2=[$t2])
  CalciteEnumerableIndexScan(table=[[OpenSearch, dates]], PushDownContext=[[PROJECT->[date], SORT->[{
  "date" : {
    "order" : "asc",
    "missing" : "_first"
  }
}]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"timeout":"1m","_source":{"includes":["date"],"excludes":[]},"sort":[{"date":{"order":"asc","missing":"_first"}}]}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```

Both `sd1` and `sd2` point to `$t2`.

Will raise another issue to fix it.

docs/user/ppl/functions/datetime.rst
```diff
@@ -8,6 +8,12 @@ Date and Time Functions
    :local:
    :depth: 1
 
+.. note::
+
+    All PPL date and time functions use the UTC time zone by default, unless otherwise specified. Both input and output
```
penghuo: > unless otherwise specified? 

In which case?

yuancu: Thanks for reminding. `SYSDATE` was an exception. Now we changed it to UTC as well. I'll rephrase the note to "All PPL date and time functions use the UTC time zone"

penghuo (dismissed):

PR #3850, opened 2025-07-04
Support partial filter push down

qianheng-aws:
### Description
Support partial filter push down

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3470

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3849, opened 2025-07-04
Support full expression in WHERE clauses

LantaoJin:
### Description
PPL where clause throws syntax error for:

- SOURCE=test | WHERE x
- SOURCE=test | WHERE x OR y
- SOURCE=test | WHERE true
- SOURCE=test | WHERE (x < 1) = (y > 1)

### Related Issues
Resolves #3273, #3272 and #3317

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3859, opened 2025-07-09
Filter script pushdown with RelJson serialization in Calcite

### LINKED ISSUE #3379: [FEATURE] Calcite Engine Framework: Pushdown scripts

songkant-aws:
### Description
The work on top of preivous Calcite filter script pushdown: https://github.com/opensearch-project/sql/pull/3812.
The issue of previous PR is it sends unsafe arbitrary Java code string to OpenSearch DataNode to compile it, leading to security concern where malicious user could use this exposed script field to run any code to attack cluster.

In this PR, it leverages Calcite existing RexNode serialization mechanism to mitigate security issue. We're now transporting logical expression representation, aka RexNode that will not allow user to customize arbitrary programming logic. The registered function calls are limited to PPLBuiltin UDFs and Calcite standard UDFs. The pushdown now undergoes this way: 

|-----------------Coordinator Node -----------------|
optimized RexNode -> encoded RexNode Json String 

---transport call-->

|--------------------------------------------------------Data Node -----------------------------------------------------|
decoded RexNode Json String -> deserialize as RexNode -> translate to Linq4j expressions -> code compile expressions and run

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3379

### Check List


LantaoJin:
@songkant-aws I remember you mentioned "Calcite has the RexNode serialization related implementation", can you point it out here?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
```diff
@@ -0,0 +1,106 @@
...
+
+  public RelJsonSerializer(RelOptCluster cluster) {
+    this.cluster = cluster;
+  }
+
+  public String serialize(
+      RexNode rexNode, RelDataType relDataType, Map<String, ExprType> fieldTypes) {
+    try {
+      // Serialize RexNode and RelDataType by JSON
+      JsonBuilder jsonBuilder = new JsonBuilder();
```
songkant-aws: Calcite related RexNode and RelDataType serialization

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
```diff
@@ -0,0 +1,106 @@
...
+      // PPL Expr types are all serializable
+      Map<String, ExprType> fieldTypes = (Map<String, ExprType>) objectMap.get(FIELD_TYPES);
+      // Deserialize RelDataType and RexNode by JSON
+      RelJson relJson = RelJson.create();
+      Map<String, Object> rowTypeMap = mapper.readValue((String) objectMap.get(ROW_TYPE), TYPE_REF);
+      RelDataType rowType = relJson.toType(cluster.getTypeFactory(), rowTypeMap);
+      OpenSearchRelInputTranslator inputTranslator = new OpenSearchRelInputTranslator(rowType);
+      relJson = relJson.withInputTranslator(inputTranslator).withOperatorTable(pplSqlOperatorTable);
+      Map<String, Object> exprMap = mapper.readValue((String) objectMap.get(EXPR), TYPE_REF);
+      RexNode rexNode = relJson.toRex(cluster, exprMap);
```
songkant-aws: Calcite RexNode and RelDataType deserialization related logic

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/OpenSearchRelInputTranslator.java
```diff
@@ -0,0 +1,47 @@
...
+public class OpenSearchRelInputTranslator implements RelJson.InputTranslator {
+
+  private final RelDataType rowType;
+
+  public OpenSearchRelInputTranslator(RelDataType rowType) {
+    this.rowType = rowType;
+  }
+
+  @Override
+  public RexNode translateInput(
```
songkant-aws: Customized handler to handle RexNode's RelInput

songkant-aws:
@LantaoJin See comments inline for Calcite RexNode serialization usage by JSON.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
```diff
@@ -0,0 +1,106 @@
...
+
+@Getter
+public class RelJsonSerializer {
+
+  private final RelOptCluster cluster;
+
+  public static final String EXPR = "expr";
+  public static final String FIELD_TYPES = "fieldTypes";
+  public static final String ROW_TYPE = "rowType";
+  private static final ObjectMapper mapper = new ObjectMapper();
```
LantaoJin: It brings me that we can leverage jackson-dataformat-msgpack which is a Jackson extension library to serialize the json to binary. (could be faster and memory saving)
```
private static final ObjectMapper mapper = new ObjectMapper(new MessagePackFactory());
```

Ref https://github.com/msgpack/msgpack-java/blob/main/msgpack-jackson/README.md

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/OpenSearchRelInputTranslator.java
```diff
@@ -0,0 +1,47 @@
...
+      RelJson relJson, int input, Map<String, @Nullable Object> map, RelInput relInput) {
+    final RelOptCluster cluster = relInput.getCluster();
+    final RexBuilder rexBuilder = cluster.getRexBuilder();
+
+    // Check if it is a local ref.
+    if (map.containsKey("type")) {
```
qianheng-aws: We can remove this handling logic or throw exception if we won't really have local reference.

It's inappropriate to put input's RowType in local ref

songkant-aws: Yes, we can remove it. I don't see we have such LocalRef case.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
songkant-aws: I checked the Calcite RelJson code. Unfortunately, it doesn't rely on ObjectMapper. It implements its own JsonBuilder to write RexNode to json string. So we cannot leverage this msgpack extension.

integ-test/src/test/java/org/opensearch/sql/ppl/RelevanceFunctionIT.java
```diff
@@ -174,8 +174,9 @@ public void not_pushdown_throws_exception() throws IOException {
     String query1 =
         "SOURCE="
             + TEST_INDEX_BEER
-            + " | EVAL answerId = AcceptedAnswerId + 1"
-            + " | WHERE simple_query_string(['Tags'], 'taste') and answerId > 200";
+            + " | STATS count(AcceptedAnswerId) as count"
+            + " | EVAL dateStr = makedate(2025, count)"
+            + " | WHERE simple_query_string(['dateStr'], 'taste')";
```
songkant-aws, line 179: Change to more complex ppl query because now most of EVAL filters can be pushed down as script expression. Adding agg before filters will prevent them pushdown.

songkant-aws:
There is no much difference for local Big5IT benchmark report because Big5 ITs don't have script pushdown case.

CalciteBig5IT:

Summary:
query_string_on_message_filtered: 153 ms
sort_numeric_asc_with_match: 42 ms
terms_significant_2: 96 ms
desc_sort_timestamp_can_match_shortcut: 22 ms
query_string_on_message_filtered_sorted_num: 43 ms
terms_significant_1: 23 ms
sort_numeric_desc: 15 ms
range_numeric: 16 ms
sort_keyword_can_match_shortcut: 16 ms
desc_sort_with_after_timestamp: 12 ms
range_field_disjunction_big_range_small_term_query: 18 ms
composite_terms_keyword: 23 ms
keyword_in_range: 28 ms
asc_sort_timestamp_no_can_match_shortcut: 15 ms
asc_sort_with_after_timestamp: 11 ms
composite_terms: 19 ms
desc_sort_timestamp: 10 ms
range_with_desc_sort: 15 ms
composite_date_histogram_daily: 30 ms
range: 13 ms
sort_numeric_asc: 11 ms
query_string_on_message: 12 ms
default: 8 ms
date_histogram_hourly_agg: 12 ms
range_with_asc_sort: 16 ms
asc_sort_timestamp_can_match_shortcut: 14 ms
term: 9 ms
multi_terms_keyword: 23 ms
sort_keyword_no_can_match_shortcut: 12 ms
desc_sort_timestamp_no_can_match_shortcut: 12 ms
range_field_conjunction_big_range_big_term_query: 13 ms
date_histogram_minute_agg: 16 ms
sort_numeric_desc_with_match: 13 ms
range_field_conjunction_small_range_big_term_query: 9 ms
asc_sort_timestamp: 9 ms
keyword_terms: 12 ms
scroll: 7 ms
range_field_conjunction_small_range_small_term_query: 10 ms
keyword_terms_low_cardinality: 12 ms
Total 39 queries succeed. Average duration: 21 ms

PPLBig5IT:

Summary:
asc_sort_timestamp: 25 ms
range_field_conjunction_small_range_small_term_query: 69 ms
term: 9 ms
composite_terms_keyword: 60 ms
sort_numeric_desc_with_match: 20 ms
range_numeric: 8 ms
terms_significant_2: 11 ms
desc_sort_timestamp_no_can_match_shortcut: 6 ms
sort_numeric_asc: 6 ms
composite_date_histogram_daily: 20 ms
desc_sort_with_after_timestamp: 6 ms
keyword_terms_low_cardinality: 11 ms
keyword_in_range: 8 ms
range_field_disjunction_big_range_small_term_query: 8 ms
range_with_asc_sort: 8 ms
scroll: 7 ms
asc_sort_timestamp_can_match_shortcut: 5 ms
query_string_on_message_filtered: 30 ms
default: 5 ms
date_histogram_hourly_agg: 6 ms
range_field_conjunction_big_range_big_term_query: 6 ms
query_string_on_message: 6 ms
range: 6 ms
keyword_terms: 6 ms
range_field_conjunction_small_range_big_term_query: 6 ms
composite_terms: 13 ms
multi_terms_keyword: 8 ms
desc_sort_timestamp: 5 ms
asc_sort_timestamp_no_can_match_shortcut: 4 ms
sort_numeric_asc_with_match: 3 ms
terms_significant_1: 6 ms
range_with_desc_sort: 7 ms
sort_keyword_no_can_match_shortcut: 4 ms
query_string_on_message_filtered_sorted_num: 6 ms
desc_sort_timestamp_can_match_shortcut: 5 ms
sort_keyword_can_match_shortcut: 4 ms
sort_numeric_desc: 5 ms
asc_sort_with_after_timestamp: 5 ms
date_histogram_minute_agg: 7 ms
Total 39 queries succeed. Average duration: 11 ms

LantaoJin:
> There is no much difference for local Big5IT benchmark report because Big5 ITs don't have script pushdown case.

Maybe the optimization could impact clickbench queries (#3860)

LantaoJin:
High-level question: have you solved how to balance the cost computing for both script and filter pushdown case?
For example, the [q11](https://github.com/opensearch-project/sql/pull/3860/files#diff-d5d82b49db77d27d281186e1368dac6cdee45b2ac74c0da236ed2f186f00b696) can both be effect by script and filter pushdown.

penghuo:
> There is no much difference for local Big5IT benchmark report because Big5 ITs don't have script pushdown case.

Do we have benchmark results for queries with script pushdown? If so, what are the results?"

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
```diff
@@ -525,4 +526,34 @@ public void testMultiFieldsRelevanceQueryFunctionExplain() throws IOException {
...
+
+  @Ignore("The serialized string is unstable because of function properties")
+  @Test
+  public void testFilterScriptPushDownExplain() throws Exception {
+    String expected =
+        isCalciteEnabled()
+            ? loadFromFile("expectedOutput/calcite/explain_filter_script_push.json")
+            : loadFromFile("expectedOutput/ppl/explain_filter_script_push.json");
+
+    assertJsonEqualsIgnoreId(
```
penghuo, line 442: can we avoid compare all function properties?

songkant-aws:
@LantaoJin @penghuo Here is the ClickBenchIT result.

**Four queries in ClickBenchIT hit Calcite script pushdown**

**Five rounds query before script pushdown:**

q21: 19ms 18ms 18ms 19ms 18ms
q22: 25ms 26ms 25ms 25ms 25ms
q23: 34ms 34ms 33ms 35ms 35ms
q24: 25ms 25ms 25ms 25ms 25ms

**Five rounds query after script pushdown:**

q21: 14ms 14ms 14ms 14ms 14ms
q22: 24ms 23ms 22ms 22ms 25ms
q23: 27ms 28ms 27ms 28ms 26ms
q24: 23ms 20ms 23ms 22ms 21ms

```
| Avg diff               | Percentage change |
| q21: 18.4ms -> 14ms    | -23.9%            |
| q22: 25.2ms -> 23.2ms  | -7.9%             |
| q23: 34.2ms -> 27.2ms  | -20.5%            |
| q24: 25ms -> 21.8ms    | -12.8%            |
```

**The overall ClickBenchIT Summary**

```
CalcitePPLClickBenchIT Summary:

q1: 21 ms
q10: 36 ms
q11: 21 ms
q12: 39 ms
q13: 17 ms
q14: 17 ms
q15: 18 ms
q16: 15 ms
q17: 16 ms
q18: 12 ms
q19: 21 ms
q2: 23 ms
q20: 9 ms
q21: 14 ms
q22: 25 ms
q23: 26 ms
q24: 21 ms
q25: 13 ms
q26: 11 ms
q27: 12 ms
q28: 27 ms
q3: 20 ms
q30: 103 ms
q31: 23 ms
q32: 23 ms
q33: 18 ms
q34: 11 ms
q35: 18 ms
q36: 17 ms
q37: 23 ms
q38: 22 ms
q39: 22 ms
q4: 16 ms
q40: 27 ms
q41: 37 ms
q42: 22 ms
q43: 25 ms
q5: 13 ms
q6: 11 ms
q7: 11 ms
q8: 19 ms
q9: 19 ms
Total 42 queries succeed. Average duration: 21 ms
```

```
PPLClickBenchIT Summary:

q1: 13 ms
q10: 5 ms
q11: 8 ms
q12: 6 ms
q13: 7 ms
q14: 7 ms
q15: 6 ms
q16: 4 ms
q17: 5 ms
q18: 4 ms
q19: 8 ms
q2: 8 ms
q20: 5 ms
q21: 5 ms
q22: 8 ms
q23: 10 ms
q24: 6 ms
q25: 6 ms
q26: 6 ms
q27: 6 ms
q28: 8 ms
q3: 7 ms
q30: 46 ms
q31: 5 ms
q32: 5 ms
q33: 5 ms
q34: 4 ms
q35: 6 ms
q36: 7 ms
q37: 17 ms
q38: 9 ms
q39: 11 ms
q4: 6 ms
q40: 8 ms
q41: 8 ms
q42: 8 ms
q43: 7 ms
q5: 5 ms
q6: 7 ms
q7: 5 ms
q8: 7 ms
q9: 7 ms
Total 42 queries succeed. Average duration: 7 ms
```

songkant-aws:
> High-level question: have you solved how to balance the cost computing for both script and filter pushdown case?
> For example, the [q11](https://github.com/opensearch-project/sql/pull/3860/files#diff-d5d82b49db77d27d281186e1368dac6cdee45b2ac74c0da236ed2f186f00b696) can both be effect by script and filter pushdown.

Not yet. Maybe we can do a followup to understand which case should prefer filter to script. For now, if filters can be translated to term query, they are preferred anyway. But I'm not sure if there is a case of translatable filter fallback to script.

qianheng-aws:
LGTM for this PR.

And there are 2 followups in my mind:
1. Cost computing for script push down and primitive push down, see more info in this comment: https://github.com/opensearch-project/sql/issues/3379#issuecomment-3051689591
2. Combing multiple script query expressions into one, if there are. I think it should be inefficiency to compile and execute them separately.

LantaoJin:
> 1. Cost computing for script push down and primitive push down, see more info in this comment: [[FEATURE] Calcite Engine Framework: Pushdown scripts #3379 (comment)](https://github.com/opensearch-project/sql/issues/3379#issuecomment-3051689591)

Before going to details, I have mentioned in https://github.com/opensearch-project/sql/pull/3859#issuecomment-3056832284, I am not sure we could separate the cost computing as a followup, without it, the feature could cause performance regression.

integ-test/src/test/resources/expectedOutput/calcite/explain_filter_script_push.json
```diff
@@ -0,0 +1,6 @@
...
...2hhZGVkLmNvbS5nb29nbGUuY29tbW9uLmNvbGxlY3QuSW1tdXRhYmxlTWFwJFNlcmlhbGl6ZWRGb3JtAAAAAAAAAAACAAJMAARrZXlzdAASTGphdmEvbGFuZy9PYmplY3Q7TAAGdmFsdWVzcQB+ABx4cHVyABNbTGphdmEubGFuZy5PYmplY3Q7kM5YnxBzKWwCAAB4cAAAAAB1cQB+AB4AAAAAc3EAfgAAAAAAA3cEAAAAAnQAB2tleXdvcmRzcQB+ABJ+cQB+AAt0AAZTVFJJTkd+cQB+ABh0AAdLZXl3b3JkcQB+AB14dAAHYWRkcmVzc3NxAH4AEHEAfgAWcQB+ABlxAH4AHXNxAH4AAAAAAAN3BAAAAAB4dAAHYmFsYW5jZXEAfgANdAAGZ2VuZGVyc3EAfgAQcQB+ABZxAH4AGXEAfgAdcQB+ACp0AARjaXR5c3EAfgAQcQB+ABZxAH4AGXEAfgAdc3EAfgAAAAAAA3cEAAAAAnEAfgAicQB+ACN4dAAIZW1wbG95ZXJzcQB+ABBxAH4AFnEAfgAZcQB+AB1zcQB+AAAAAAADdwQAAAACcQB+ACJxAH4AI3h0AAVzdGF0ZXNxAH4AEHEAfgAWcQB+ABlxAH4AHXNxAH4AAAAAAAN3BAAAAAJxAH4AInEAfgAjeHQAA2FnZXEAfgANdAAFZW1haWxzcQB+ABBxAH4AFnEAfgAZcQB+AB1zcQB+AAAAAAADdwQAAAACcQB+ACJxAH4AI3h0AAhsYXN0bmFtZXNxAH4AEHEAfgAWcQB+ABlxAH4AHXNxAH4AAAAAAAN3BAAAAAJxAH4AInEAfgAjeHgAeA==\",\"lang\":\"opensearch_compounded_script\",\"options\":{\"engine_type\":\"calcite\"},\"params\":{\"utcTimestamp\":*}},\"boost\":1.0}}],\"adjust_pure_negative\":true,\"boost\":1.0}},\"_source\":{\"includes\":[\"firstname\",\"age\"],\"excludes\":[]},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}]}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
```
LantaoJin: `SCRIPT->AND(=($0, 'Amber'), =(-($1, 2), 30))` seems not the final plan, we need https://github.com/opensearch-project/sql/pull/3850 first.

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
```diff
@@ -426,6 +427,28 @@ public void testMultiFieldsRelevanceQueryFunctionExplain() throws IOException {
                 + " default_operator='or', analyzer=english)"));
   }
 
+  @Ignore("The serialized string is unstable because of function properties")
```
LantaoJin, line 438: please add the test:
`source=opensearch-sql_test_index_account | where firstname != ''`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CompoundedScriptEngine.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.calcite.plan.RelOptCluster;
+import org.opensearch.script.AggregationScript;
+import org.opensearch.script.FilterScript;
+import org.opensearch.script.ScriptContext;
+import org.opensearch.script.ScriptEngine;
+import org.opensearch.sql.opensearch.storage.serialization.DefaultExpressionSerializer;
+
+/**
+ * Custom expression script engine that supports using core engine expression code in DSL as a new
+ * script language just like built-in Painless language.
```
penghuo, line 27: > just like build-in painless language

what is user interface looks like in rest api?

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -293,7 +300,9 @@ public List<Setting<?>> getSettings() {
 
   @Override
   public ScriptEngine getScriptEngine(Settings settings, Collection<ScriptContext<?>> contexts) {
-    return new ExpressionScriptEngine(new DefaultExpressionSerializer());
+    RexBuilder rexBuilder = new RexBuilder(OpenSearchTypeFactory.TYPE_FACTORY);
+    RelOptCluster cluster = RelOptCluster.create(new VolcanoPlanner(), rexBuilder);
```
penghuo: RexOptCluster can be consturcted in CompoundedScriptEngine?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CompoundedScriptEngine.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Override
+  public String getType() {
+    return COMPOUNDED_LANG_NAME;
+  }
+
+  @Override
+  public <T> T compile(
+      String scriptName, String scriptCode, ScriptContext<T> context, Map<String, String> options) {
+    return switch (options.getOrDefault(ENGINE_TYPE, V2_ENGINE_TYPE)) {
```
penghuo: Can we get rid of options? The concern is we can not run PPL library as a client out of cluster.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.calcite.sql.fun.SqlLibrary;
+import org.apache.calcite.sql.fun.SqlLibraryOperatorTableFactory;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.util.SqlOperatorTables;
+import org.apache.calcite.util.JsonBuilder;
+import org.opensearch.sql.data.type.ExprType;
+import org.opensearch.sql.expression.function.PPLBuiltinOperators;
+
+@Getter
+public class RelJsonSerializer {
```
penghuo: Does this class copy from Calcite? If not, add doc to explain class and method, add UT to test key features.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.storage.serialization;
```
penghuo: serialization -> serde

opensearch/build.gradle
```diff
@@ -41,6 +41,7 @@ dependencies {
     compileOnly group: 'org.opensearch.client', name: 'opensearch-rest-high-level-client', version: "${opensearch_version}"
     implementation group: 'org.opensearch', name:'opensearch-ml-client', version: "${opensearch_build}"
     implementation group: 'org.opensearch', name:'geospatial-client', version: "${opensearch_build}"
+    implementation group: 'org.codehaus.janino', name: 'janino', version: '3.1.11'
```
penghuo: Calcite already depend on janino https://mvnrepository.com/artifact/org.apache.calcite/calcite-core/1.40.0

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,350 @@
...
+
+    public ReferenceFieldVisitor(
+        RelDataType rowType, Map<String, ExprType> fieldTypes, boolean deep) {
+      super(deep);
+      this.rowType = rowType;
+      this.fieldTypes = fieldTypes;
+    }
+
+    @Override
+    public Pair<String, ExprType> visitInputRef(RexInputRef inputRef) {
```
penghuo, line 223: if getValidatedReferenceNameAndType throw UnsupportedScriptException, fallback to no-pushdown logic?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -542,35 +582,41 @@ private QueryExpression binary(RexCall call) {
...
       boolean partial = false;
+      // For function isEmpty and isBlank, we implement them via expression `isNull or {@function}`,
+      // Unlike `OR` in Java, `SHOULD` in DSL will evaluate both branches and lead to NPE.
+      if (call.getKind() == SqlKind.OR
+          && call.getOperands().size() == 2
+          && (call.getOperands().get(0).getKind() == SqlKind.IS_NULL
+              || call.getOperands().get(1).getKind() == SqlKind.IS_NULL)) {
+        throw new UnsupportedScriptException(
+            "DSL will evaluate both branches of OR with isNUll, prevent push-down to avoid NPE");
+      }
```
penghuo, line 621: Add UT.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
```diff
@@ -0,0 +1,106 @@
...
+  private static final ObjectMapper mapper = new ObjectMapper();
+  private static final TypeReference<LinkedHashMap<String, Object>> TYPE_REF =
+      new TypeReference<>() {};
+  private static final SqlOperatorTable pplSqlOperatorTable =
+      SqlOperatorTables.chain(
+          PPLBuiltinOperators.instance(),
+          SqlStdOperatorTable.instance(),
+          // Add a list of necessary SqlLibrary if needed
+          SqlLibraryOperatorTableFactory.INSTANCE.getOperatorTable(
+              SqlLibrary.MYSQL, SqlLibrary.BIG_QUERY, SqlLibrary.SPARK, SqlLibrary.POSTGRESQL));
```
penghuo: Does it means user can any valid function in this List?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -137,45 +149,70 @@ private PredicateAnalyzer() {}
...
+      RexNode expression,
+      List<String> schema,
+      Map<String, ExprType> fieldTypes,
+      RelDataType rowType,
+      RelOptCluster cluster)
+      throws ExpressionNotAnalyzableException {
+    return analyze_(expression, schema, fieldTypes, rowType, cluster).builder();
+  }
+
+  public static QueryExpression analyze_(
```
penghuo: why use analyze_ as function name?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
```diff
@@ -0,0 +1,106 @@
...
+  static {
+    mapper.configure(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS, true);
+  }
+
+  public RelJsonSerializer(RelOptCluster cluster) {
+    this.cluster = cluster;
+  }
+
+  public String serialize(
+      RexNode rexNode, RelDataType relDataType, Map<String, ExprType> fieldTypes) {
```
penghuo: fieldTypes should only include filed used in RexNode, right?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
```diff
@@ -0,0 +1,106 @@
...
+      // Construct envelope of serializable objects
+      Map<String, Object> envelope =
+          Map.of(EXPR, rexNodeJson, FIELD_TYPES, fieldTypes, ROW_TYPE, rowTypeJson);
+
+      // Write bytes of all serializable contents
+      ByteArrayOutputStream output = new ByteArrayOutputStream();
+      ObjectOutputStream objectOutput = new ObjectOutputStream(output);
+      objectOutput.writeObject(envelope);
+      objectOutput.flush();
+      return Base64.getEncoder().encodeToString(output.toByteArray());
```
penghuo: Is Base64 encoding necessary? Did we actually implement a separate script execution based on RexNode?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,350 @@
...
+
+    JavaTypeFactoryImpl typeFactory =
+        new JavaTypeFactoryImpl(relJsonSerializer.getCluster().getTypeFactory().getTypeSystem());
+    RexToLixTranslator.InputGetter getter = new ScriptInputGetter(typeFactory, rowType, fieldTypes);
+    String code =
+        CalciteScriptEngine.translate(
+            relJsonSerializer.getCluster().getRexBuilder(), List.of(rexNode), getter, rowType);
+
+    Function1<DataContext, Object[]> function =
+        new RexExecutable(code, "generated Rex code").getFunction();
```
penghuo, line 135: How are RexExecutable and RexExecutorImpl used in Calcite? Do they follow the same execution path as code generation?

songkant-aws:
Added jmh benchmark for expression serialization

```
# JMH version: 1.35
# VM version: JDK 21.0.5, OpenJDK 64-Bit Server VM, 21.0.5+11-LTS
# VM invoker: /Library/Java/JavaVirtualMachines/amazon-corretto-21.jdk/Contents/Home/bin/java
# VM options: -Dfile.encoding=UTF-8 -Djava.io.tmpdir=/Users/songkant/workspace/opensearch-project-dev/sql/benchmarks/build/tmp/jmh -Duser.country=US -Duser.language=en -Duser.variant
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 1 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.opensearch.sql.expression.operator.predicate.ExpressionScriptSerdeBenchmark.testRexNodeJsonSerde

# Run progress: 74.14% complete, ETA 00:05:01
# Fork: 1 of 1
# Warmup Iteration   1: SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.
18048.349 ns/op
Iteration   1: 15960.460 ns/op
Iteration   2: 15886.672 ns/op[14m 57s]
Iteration   3: 15931.590 ns/op[15m 7s]
Iteration   4: 15969.726 ns/op[15m 17s]
Iteration   5: 16194.336 ns/op[15m 27s]
Iteration   6: 16280.464 ns/op[15m 37s]
Iteration   7: 15936.191 ns/op[15m 47s]
Iteration   8: 16003.951 ns/op[15m 57s]
Iteration   9: 15977.911 ns/op[16m 7s]
Iteration  10: 15953.869 ns/op[16m 17s]


Result "org.opensearch.sql.expression.operator.predicate.ExpressionScriptSerdeBenchmark.testRexNodeJsonSerde":
  16009.517 ±(99.9%) 190.056 ns/op [Average]
  (min, avg, max) = (15886.672, 16009.517, 16280.464), stdev = 125.710
  CI (99.9%): [15819.462, 16199.573] (assumes normal distribution)


# JMH version: 1.35
# VM version: JDK 21.0.5, OpenJDK 64-Bit Server VM, 21.0.5+11-LTS
# VM invoker: /Library/Java/JavaVirtualMachines/amazon-corretto-21.jdk/Contents/Home/bin/java
# VM options: -Dfile.encoding=UTF-8 -Djava.io.tmpdir=/Users/songkant/workspace/opensearch-project-dev/sql/benchmarks/build/tmp/jmh -Duser.country=US -Duser.language=en -Duser.variant
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 1 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.opensearch.sql.expression.operator.predicate.ExpressionScriptSerdeBenchmark.testV2ExpressionSerde

# Run progress: 83.62% complete, ETA 00:03:10
# Fork: 1 of 1
# Warmup Iteration   1: 76917.475 ns/op
Iteration   1: 73595.154 ns/op[16m 37s]
Iteration   2: 73156.246 ns/op[16m 47s]
Iteration   3: 73528.585 ns/op[16m 57s]
Iteration   4: 73522.270 ns/op[17m 7s]
Iteration   5: 73298.841 ns/op[17m 17s]
Iteration   6: 73564.956 ns/op[17m 27s]
Iteration   7: 75099.827 ns/op[17m 37s]
Iteration   8: 73680.223 ns/op[17m 47s]
Iteration   9: 73437.472 ns/op[17m 57s]
Iteration  10: 75394.735 ns/op[18m 7s]


Result "org.opensearch.sql.expression.operator.predicate.ExpressionScriptSerdeBenchmark.testV2ExpressionSerde":
  73827.831 ±(99.9%) 1158.471 ns/op [Average]
  (min, avg, max) = (73156.246, 73827.831, 75394.735), stdev = 766.257
  CI (99.9%): [72669.360, 74986.302] (assumes normal distribution)
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CompoundedScriptEngine.java
qianheng-aws: We can make the Calcite script engine as the default engine type. Then rest client can invoke it as well without passing any options

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CompoundedScriptEngine.java
songkant-aws: rest api doesn't expose ENGINE_TYPE option yet. rest api hardcodes default language as painless script. If user specifies ENGINE_TYPE, he will get exception now.

This class's option selection is for internal V2 fallback in case of unsupported queries in Calcite for NodeClient.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: They are used to reduce constant expressions in Calcite.

If we are talking about the whole plan code-gen, that is more complicated since it includes code-gen for RelNode. But the path of code-gen for RexNode are the same by using `RexToLixTranslator`

opensearch/build.gradle
qianheng-aws: Yeah, it's not necessary anymore after this change: https://github.com/opensearch-project/sql/pull/3859/commits/bc6ecac06599b1646a460d0846900cdcb953b777#diff-59ef715f591a9e9937469905395d8ef89b941c5098176d0906148f608dc81743L80

It's necessary before since I want to import the class of janino in our implementation. This change makes janino only necessary in runtime.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
songkant-aws: Yes, for simplicity. If we want least restrictive function access, we can manually pick needed functions from those libraries to a separate operator table.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
songkant-aws: Yes, it will fallback to no-pushdown logic.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
songkant-aws: I think it's still necessary. The OpenSearch core Script object expects text string of script code. Base64 encoding provides stable and safe text data transmission and avoids special character escaping issue. We want to pass both JSON strings and a map of OpenSearch field ExprType mapping to resolve input type for RexNode script execution. So eventually, we're still transporting a MAP object to data node.

The another option is we can serialize ExprType by JSON as well. For now, it has some circular reference issue that needs some workaround.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1190,6 +1259,171 @@ private static String timestampValueForPushDown(String value) {
...
+        RexNode rexNode,
+        RelDataType rowType,
+        Map<String, ExprType> fieldTypes,
+        RelOptCluster cluster) {
+      ReferenceFieldVisitor validator = new ReferenceFieldVisitor(rowType, fieldTypes, true);
+      // Dry run visitInputRef to make sure the input reference ExprType is valid for script
+      // pushdown
+      validator.visitEach(List.of(rexNode));
+      RelJsonSerializer serializer = new RelJsonSerializer(cluster);
+      this.code = serializer.serialize(rexNode, rowType, fieldTypes);
```
penghuo: ScriptQueryExpression() is invoked multiple times during the optimization stage. Can we defer code generation until the implementation stage?

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
LantaoJin: @songkant-aws can you add above test case in explain IT?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -0,0 +1,350 @@
...
+public class CalciteScriptEngine implements ScriptEngine {
+
+  private final RelJsonSerializer relJsonSerializer;
+
+  public CalciteScriptEngine(RelOptCluster relOptCluster) {
+    this.relJsonSerializer = new RelJsonSerializer(relOptCluster);
+  }
+
+  /** Expression script language name. */
+  public static final String CALCITE_LANG_NAME = "opensearch_calcite";
```
LantaoJin: for more readable, change to:
```
public static final String EXPRESSION_LANG_NAME = "opensearch_calcite_expression";
```

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
songkant-aws: Added the above test. But the test itself is optimized as bool filter query before hitting script pushdown codepath. Ideally we should find a query that is chosen by optimizer. I'm checking if there is one.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
songkant-aws: Done

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
songkant-aws: It's a member of FunctionExpression in V2. It has a reference of current Instant object. So the generated bytes string is not stable. It can't be avoided now.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
songkant-aws: Done

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
songkant-aws: I checked that Calcite will pick the cheapest cost of equivalent plan during optimization among a collection of optimized plans. The pushed down RexNode in those plans could be different. So caching the same canonicalized expression is not the option.

Well, it's not easy to change the current code of lazy building OpenSearch query request. Serializing simple expression is fast for now. We can create an issue to enhance this experience later.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CompoundedScriptEngine.java
songkant-aws: There is some issue of making Calcite as default engine option because legacy SQL relies on V2 Expression as well and it's not reading option map. I'm thinking if we can add a fallback as well. Like try compile calcite first and then fallback to compile v2 expression in case of failure.

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
songkant-aws: Done

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
songkant-aws: Done. Added javadoc and UT

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
songkant-aws: Done

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
songkant-aws: Renamed to analyzeExpression

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serialization/RelJsonSerializer.java
songkant-aws: Yes, it's not necessary for index field mapping. I use filter's rowType instead now.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CompoundedScriptEngine.java
penghuo: Can we encode langType in scriptCode?

songkant-aws: I wrap them into a JSON blob. {"langType": type, "script": ...} It should be lightweight and extensible.

In the long term, I think we should replace serialization mechanism with JSON/Protobuf because they're safer. Java serialization still has risk of loading malicious constructor/function to run unexpected code.

LantaoJin:
@songkant-aws can you fix the conflicts, I think the enhancement has been blocked for too long. @penghuo @qianheng-aws any other concerns? can we merge it first?

penghuo (dismissed):

LantaoJin (dismissed):

qianheng-aws (dismissed):

qianheng-aws:
LGTM. Please resolve the conflict @songkant-aws

penghuo (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -659,6 +668,30 @@ private QueryExpression andOr(RexCall call) {
       }
     }
 
+    private Expression tryAnalyzeOperand(RexNode node) {
+      try {
+        Expression expr = node.accept(this);
+        if (expr instanceof NamedFieldExpression) {
```
qianheng-aws, line 674: [question] In which case will we go into this?

songkant-aws: No, we don't have such case. So it's noop when we encounter it.

qianheng-aws: Seems will throw exception if invoking `CompoundQueryExpression ::and` later
 
https://github.com/opensearch-project/sql/blob/37536b8e7d39d1d5a5ae301c342e1d01c59107a5/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java#L927

qianheng-aws: I'm thinking we do will meet this after merging this PR: https://github.com/opensearch-project/sql/pull/3849, since it support using a singleton field as a condition in filer.  https://github.com/opensearch-project/sql/pull/3849#pullrequestreview-2996708207

But it needs more change to support push down such `NamedFieldExpression`, better in a separate PR.

songkant-aws: It should never happen. We could just throw exception here.

qianheng-aws: It's OK to me leaving this code here. Current behavior of throwing exception in `CompoundQueryExpression ::and` makes more sense to me comparing to ignoring it. Ignoring NamedFieldExpression seems will induce to correctness issue.

qianheng-aws (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
qianheng-aws: @LantaoJin  Do you have any other concern on this?

LantaoJin: > @LantaoJin Do you have any other concern on this?

Looks good to me as long as the case `where field1=10 or field2` can apply partial pushdown `field1=10`.

LantaoJin (approved):

LantaoJin:
@songkant-aws please manually backport it to 2.19-dev

PR #3864, opened 2025-07-10
Support pushdown physical sort operator to speedup SortMergeJoin

### LINKED ISSUE #3863: [FEATURE] Pushdown the Sort operators of SortMergeJoin
In Calcite + OpenSearch, the default physical Join operator is `EnumerableMergeJoin`:
```
EnumerableMergeJoin
\-EnumerableSort
  \-CalciteEnumerableIndexScan
\-EnumerableSort
  \-CalciteEnumerableIndexScan
```
After this patching, he `EnumerableSort` will be pushed down to DSL:
```
EnumerableMergeJoin
\-CalciteEnumerableIndexScan(PushDownContext=[SORT->"join_key"])
\-CalciteEnumerableIndexScan(PushDownContext=[SORT->"join_key"])
```

LantaoJin:
### Description
In Calcite + OpenSearch, the default physical Join operator is `EnumerableMergeJoin`:
```
EnumerableMergeJoin
\-EnumerableSort
  \-CalciteEnumerableIndexScan
\-EnumerableSort
  \-CalciteEnumerableIndexScan
```
After this patching, he `EnumerableSort` will be pushed down to DSL:
```
EnumerableMergeJoin
\-CalciteEnumerableIndexScan(PushDownContext=[SORT->"join_key"])
\-CalciteEnumerableIndexScan(PushDownContext=[SORT->"join_key"])
```

### Related Issues
Resolves #3863 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -123,6 +139,149 @@ public boolean isAggregatePushed() {
...
+          String fieldNameKeyword;
+          if (fieldType instanceof OpenSearchTextType textType) {
+            fieldNameKeyword = OpenSearchTextType.convertTextToKeyword(fieldName, fieldType);
+            if (fieldName.equals(fieldNameKeyword) && !textType.isFieldData()) {
+              // can not convert text to keyword, skip pushdown
+              return null;
+            }
+          } else {
+            fieldNameKeyword = fieldName;
+          }
```
LantaoJin: this method is moved from `CalciteLogicalIndexScan` except these lines. @qianheng-aws

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchTextType.java
```diff
@@ -25,24 +25,32 @@ public class OpenSearchTextType extends OpenSearchDataType {
   // text could have fields
   // a read-only collection
   @EqualsAndHashCode.Exclude Map<String, OpenSearchDataType> fields = ImmutableMap.of();
+  @EqualsAndHashCode.Exclude private boolean fielddata = false;
```
qianheng-aws: Could you elaborate more on this new field?

Looks like it's for identify whether this `text` type has raw data. If that's the case, we should update this place as well: https://github.com/opensearch-project/sql/blob/a3cd42e71d9e6b3df1e38651398ad5850d767db6/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java#L1288-L1302

qianheng-aws: Looks like we'd better detect whether `fielddata` firstly. And if not try to whether there is subfield of keyword type.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
qianheng-aws: Would be better to detect whether the `text` type `isFieldData()` first before finding keyword subfield, and wrapping all this logic `convertTextToKeyword`.

And I think there is a bug in `OpenSearchTextType.convertTextToKeyword`, it doesn't make sense to always convert a field to keyword by appending string `.keyword`. It should append subfield name of keyword type, like what we implement in `PredicateAnalyzer`

LantaoJin: > Would be better to detect whether the text type isFieldData() first before finding keyword subfield

Why? A text type could be both isFieldData and contain keyword subfield.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/type/OpenSearchTextType.java
LantaoJin: > Could you elaborate more on this new field?

pushdown sort to a text field neither contains keyword subfield nor fielddata=false (default value) will throw exception.

The new field is extracting the `fielddata` flag from mapping, for example:
```
      "gender": {
        "type": "text"
      },
      "address": {
        "type": "text",
        "fielddata": true
      },
      "firstname": {
        "type": "text",
        "fielddata": true,
        "fields": {
          "keyword": {
            "type": "keyword",
            "ignore_above": 256
          }
        }
      }
```
we cannot pushdown sort `gender` to DSL.

benchmarks/src/jmh/java/org/opensearch/sql/expression/operator/predicate/MergeArrayAndObjectMapBenchmark.java
```diff
@@ -21,7 +21,7 @@ public class MergeArrayAndObjectMapBenchmark {
   public void testMerge() {
     Map<String, OpenSearchDataType> finalResult = new HashMap<>();
     for (Map<String, OpenSearchDataType> map : candidateMaps) {
-      OpenSearchDescribeIndexRequest.mergeObjectAndArrayInsideMap(finalResult, map);
+      MergeRuleHelper.merge(finalResult, map);
```
LantaoJin, line 24: not related. just fix a compile error in benchmarks module.

integ-test/src/test/resources/indexDefinitions/account_index_mapping.json
```diff
@@ -3,7 +3,13 @@
...
       "gender": {
         "type": "text",
-        "fielddata": true
+        "fielddata": true,
+        "fields": {
+          "keyword": {
+            "type": "keyword",
+            "ignore_above": 256
+          }
+        }
```
LantaoJin, line 12: `gender` in test sample data should contain keyword subfield. If else, we cannot sorting, aggregate on it.  Sorting on a `fielddata` field would get unstable result. ref [link](https://github.com/opensearch-project/sql/pull/3864#discussion_r2206876085)

integ-test/src/test/java/org/opensearch/sql/ppl/StatsCommandIT.java
```diff
@@ -358,27 +357,15 @@ public void testStatsBySpanAndMultipleFields() throws IOException {
...
-    } else {
-      verifyDataRowsInOrder(
-          response,
-          rows(1, 20, "f", "VA"),
-          rows(1, 30, "f", "IN"),
-          rows(1, 30, "f", "PA"),
-          rows(1, 30, "m", "IL"),
-          rows(1, 30, "m", "MD"),
-          rows(1, 30, "m", "TN"),
-          rows(1, 30, "m", "WA"));
```
LantaoJin, line 380: this lower case values in v2 were caused by grouping by a `fielddata` field.

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
```diff
@@ -134,7 +134,7 @@ public void testMultiSortPushDownExplain() throws IOException {
         explainQueryToString(
             "source=opensearch-sql_test_index_account "
                 + "| sort account_number, firstname, address, balance "
-                + "| sort - balance, - gender, address "
```
LantaoJin, line 137: We do not sort on text type without keyword subfield in v3. Even sorting on text type with fielddata=true will get unexpected order:
For example,  if `gender` is a text type field with fielddata=true.

Check following queries and their outputs in v2.

index=test1:
| gender |
| ------ |
| M a 1 |
| M b 1 |
| F c 2 |
| F d d 1 |

`source = test1 | sort gender` returns
| gender |
| ------ |
| M a 1 |
| M b 1 |
| F d d 1 |
| F c 2 |

`source = test1 | sort - gender` returns
| gender |
| ------ |
| M a 1 |
| M b 1 |
| F c 2 |
| F d d 1 |


index=test2:
| gender |
| ------ |
| F d d 1 |
| M a 1 |
| F c 2 |
| M b 1 |

`source=test2 | sort gender ` returns
| gender |
| ------ |
| F d d 1 |
| M a 1 |
| M b 1 |
| F c 2 |

`source = test2 | sort - gender` returns
| gender |
| ------ |
| M a 1 |
| M b 1 |
| F d d 1 |
| F c 2 |

The logic of sorting a fielddata is 1) tokenization 2) internal sort by token 3) sort by first token
ASC: "F d d 1" =tokenization=> "f", "d", "d", "1" =internal sort by token=> "1", "f", "d", "d" =sort by first token=> "1"
DESC: "F d d 1" =tokenization=> "f", "d", "d", "1" =internal sort by token=> "f", "d", "d", "1" =sort by first token=> "f"

LantaoJin:
Could you review this again?@qianheng-aws @penghuo

integ-test/src/test/java/org/opensearch/sql/ppl/RareCommandIT.java
```diff
@@ -59,7 +59,7 @@ public void testRareWithGroup() throws IOException {
           rows("F", "OK", 7),
           rows("F", "KS", 7),
           rows("F", "CO", 7),
-          rows("F", "NV", 8),
+          isPushdownEnabled() ? rows("F", "AR", 8) : rows("F", "NV", 8),
```
qianheng-aws, line 62: How do we introduce this disparity between with&without pushdown?

qianheng-aws (approved):

integ-test/src/test/java/org/opensearch/sql/ppl/RareCommandIT.java
LantaoJin: I added "keyword" subfield for the `gender`, agg pushdown `by gender` works in caclite.

penghuo: Why pushdown enable impact results?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchSortIndexScanRule.java
```diff
@@ -21,11 +21,11 @@ protected OpenSearchSortIndexScanRule(Config config) {
 
   @Override
   public void onMatch(RelOptRuleCall call) {
-    final LogicalSort sort = call.rel(0);
-    final CalciteLogicalIndexScan scan = call.rel(1);
+    final Sort sort = call.rel(0);
```
penghuo, line 24: @LantaoJin @qianheng-aws couple questions?
1. Is OpenSearchSortIndexScanRule intended to be a physical tweak rule for EnumerableRel?
2. Do we still need Logical plan optimization rule and CalciteLogicalIndexScan? Can all existing rules be expressed as physical tweak rules for EnumerableRel?

integ-test/src/test/java/org/opensearch/sql/ppl/RareCommandIT.java
LantaoJin: > Why pushdown enable impact results?

This result difference is not introduced by code changes in this PR. You can reproduce it in main branch when change the schema from
```
      "gender": {
        "type": "text",
        "fielddata": true
      },
```
to
```
      "gender": {
        "type": "text",
        "fields": {
          "keyword": {
            "type": "keyword",
            "ignore_above": 256
          }
        }
      },
```

The three results (v2 non-pushdown, v3 non-pushdown, v3 pushdown) are different.
| v2 non-pushdown | v3 non-pushdown | v3 pushdown |
| --- | --- | --- |
| "F", "VA", 8 | "F", "NV", 8 | "F", "AR", 8 |

But they are all correct. The rare command is not a deterministic command, it result depends on the order of return from OpenSearch and implementation (for v2). When the bucket pushdown works (by gender), the fetched data order is different with data order in non-pushdown.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchSortIndexScanRule.java
LantaoJin: EnumerableSort can generate from logical sort operator and logical join operator with physical SortMerge implementation. So I changed the `OpenSearchSortIndexScanRule` rule from logical layer to physical layer. It equals keeping the `OpenSearchSortIndexScanRule` in logical layer and adding a new rule for EnumerableSort + SortMergeJoin in physical layer.

Almost pushdown rules can move to physical layer but I didn't see any benefit to do that. We need logical plan rules. some optimization should happen in logical layer instead of physical. For example a LogicalJoin operator may generate multiple physical Join operators. rule applying in logical should be more efficient.

PR #3867, opened 2025-07-10
Support `reverse` command with Calcite

selsong:
### Description
Implement the `reverse` command in PPL to flip the result order of records. This command does not take any parameters and reverses the order of rows as they appear at that point in the pipeline.

Example:
```ppl
source=accounts | reverse
```
### Related Issues
N/A 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -99,6 +99,7 @@ import org.opensearch.sql.ast.tree.Project;
 import org.opensearch.sql.ast.tree.RareTopN;
 import org.opensearch.sql.ast.tree.Relation;
 import org.opensearch.sql.ast.tree.Rename;
+import org.opensearch.sql.ast.tree.Reverse;
 import org.opensearch.sql.ast.tree.Sort;
 import org.opensearch.sql.ast.tree.Sort.SortOption;
 import org.opensearch.sql.ast.tree.SubqueryAlias;
@@ -352,6 +353,20 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
   public RelNode visitHead(Head node, CalcitePlanContext context) {
     visitChildren(node, context);
     context.relBuilder.limit(node.getFrom(), node.getSize());
+    return context.relBuilder.peek();
+  }
+
+  @Override
+  public RelNode visitReverse(org.opensearch.sql.ast.tree.Reverse node, CalcitePlanContext context) {
+    visitChildren(node, context);
+    // Add ROW_NUMBER() column
+    RexNode rowNumber = context.relBuilder.aggregateCall(SqlStdOperatorTable.ROW_NUMBER)
+        .over().rowsTo(RexWindowBounds.CURRENT_ROW).as("__reverse_row_num__");
+    context.relBuilder.projectPlus(rowNumber);
+    // Sort by row number descending
+    context.relBuilder.sort(context.relBuilder.desc(context.relBuilder.field("__reverse_row_num__")));
+    // Remove row number column
+    context.relBuilder.projectExcept(context.relBuilder.field("__reverse_row_num__"));
     return context.relBuilder.peek();
   }
 
--- a/docs/user/ppl/cmd/reverse.rst
+++ b/docs/user/ppl/cmd/reverse.rst
@@ -11,7 +11,7 @@ reverse
 
 Description
 ============
-| Using ``reverse`` command to reverse the order of search results. The reverse command returns results in the opposite order from how they would normally be displayed, but does not affect which results are returned by the search.
+| Using ``reverse`` command to reverse the display order of search results. The same results are returned, but in reverse order.
 
 
 Syntax
--- a/docs/user/ppl/cmd/reverse.rst
+++ b/docs/user/ppl/cmd/reverse.rst
@@ -1,0 +1,114 @@
+=============
+reverse
+=============
+
+.. rubric:: Table of contents
+
+.. contents::
+   :local:
+   :depth: 2
+
+
+Description
+============
+| Using ``reverse`` command to reverse the order of search results. The reverse command returns results in the opposite order from how they would normally be displayed, but does not affect which results are returned by the search.
+
+
+Syntax
+============
+reverse
+
+
+* No parameters: The reverse command takes no arguments or options.
+
+
+Example 1: Basic reverse operation
+==================================
+
+The example shows reversing the order of all documents.
+
+PPL query::
+
+    os> source=accounts | fields account_number, age | reverse;
+    fetched rows / total rows = 4/4
+    +----------------+-----+
+    | account_number | age |
+    |----------------+-----|
+    | 6              | 36  |
+    | 18             | 33  |
+    | 1              | 32  |
+    | 13             | 28  |
+    +----------------+-----+
+
+
+Example 2: Reverse with sort
+============================
+
+The example shows reversing results after sorting by age in ascending order, effectively giving descending order.
+
+PPL query::
+
+    os> source=accounts | sort age | fields account_number, age | reverse;
+    fetched rows / total rows = 4/4
+    +----------------+-----+
+    | account_number | age |
+    |----------------+-----|
+    | 6              | 36  |
+    | 18             | 33  |
+    | 1              | 32  |
+    | 13             | 28  |
+    +----------------+-----+
+
+
+Example 3: Reverse with head
+============================
+
+The example shows using reverse with head to get the last 2 records from the original order.
+
+PPL query::
+
+    os> source=accounts | reverse | head 2 | fields account_number, age;
+    fetched rows / total rows = 2/2
+    +----------------+-----+
+    | account_number | age |
+    |----------------+-----|
+    | 6              | 36  |
+    | 18             | 33  |
+    +----------------+-----+
+
+
+Example 4: Double reverse
+=========================
+
+The example shows that applying reverse twice returns to the original order.
+
+PPL query::
+
+    os> source=accounts | reverse | reverse | fields account_number, age;
+    fetched rows / total rows = 4/4
+    +----------------+-----+
+    | account_number | age |
+    |----------------+-----|
+    | 13             | 28  |
+    | 1              | 32  |
+    | 18             | 33  |
+    | 6              | 36  |
+    +----------------+-----+
+
+
+Example 5: Reverse with complex pipeline
+=======================================
+
+The example shows reverse working with filtering and field selection.
+
+PPL query::
+
+    os> source=accounts | where age > 30 | fields account_number, age | reverse;
+    fetched rows / total rows = 3/3
+    +----------------+-----+
+    | account_number | age |
+    |----------------+-----|
+    | 6              | 36  |
+    | 18             | 33  |
+    | 1              | 32  |
+    +----------------+-----+
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteReverseCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteReverseCommandIT.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.remote;
+
+import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_BANK;
+import static org.opensearch.sql.util.MatcherUtils.rows;
+import static org.opensearch.sql.util.MatcherUtils.schema;
+import static org.opensearch.sql.util.MatcherUtils.verifyDataRowsInOrder;
+import static org.opensearch.sql.util.MatcherUtils.verifySchema;
+
+import java.io.IOException;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+import org.opensearch.sql.ppl.PPLIntegTestCase;
+
+public class CalciteReverseCommandIT extends PPLIntegTestCase {
+
+    @Override
+    public void init() throws Exception {
+        super.init();
+        enableCalcite();
+        disallowCalciteFallback();
+        loadIndex(Index.BANK);
+    }
+
+    @Test
+    public void testReverse() throws IOException {
+        JSONObject result = executeQuery(String.format("source=%s | fields account_number | reverse", TEST_INDEX_BANK));
+        verifySchema(result, schema("account_number", "bigint"));
+        verifyDataRowsInOrder(result, rows(32), rows(25), rows(20), rows(18), rows(13), rows(6), rows(1));
+    }
+
+    @Test
+    public void testReverseWithFields() throws IOException {
+        JSONObject result = executeQuery(String.format("source=%s | fields account_number, firstname | reverse", TEST_INDEX_BANK));
+        verifySchema(result, schema("account_number", "bigint"), schema("firstname", "string"));
+        verifyDataRowsInOrder(
+                result,
+                rows(32, "Dillard"),
+                rows(25, "Virginia"),
+                rows(20, "Elinor"),
+                rows(18, "Dale"),
+                rows(13, "Nanette"),
+                rows(6, "Hattie"),
+                rows(1, "Amber JOHnny"));
+    }
+
+    @Test
+    public void testReverseWithSort() throws IOException {
+        JSONObject result = executeQuery(String.format("source=%s | sort account_number | fields account_number | reverse", TEST_INDEX_BANK));
+        verifySchema(result, schema("account_number", "bigint"));
+        verifyDataRowsInOrder(result, rows(32), rows(25), rows(20), rows(18), rows(13), rows(6), rows(1));
+    }
+
+    @Test
+    public void testDoubleReverse() throws IOException {
+        JSONObject result = executeQuery(String.format("source=%s | fields account_number | reverse | reverse", TEST_INDEX_BANK));
+        verifySchema(result, schema("account_number", "bigint"));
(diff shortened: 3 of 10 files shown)


LantaoJin:
@selsong Thanks for starting contribution. ~~But I doubt that we need the `reverse` command for right now since~~
**Update:**
1. For new PPL commands, a Feature or RFC issue is required for syntax review before pull request.
2. This restriction can be related. ~~PPL as an unified query language, new commands and functions are required to be implemented both in PPL-on-OpenSearch and [PPL-on-Spark](https://github.com/opensearch-project/opensearch-spark/tree/main/ppl-spark-integration). But this won't be a problem if https://github.com/opensearch-project/sql/issues/3734 is done. Before that, new commands should be planned in release to avoid fragmented user experience. The current workaround is implementing it in two repositories.~~
3. ~~`reverse` could cause performance crisis when handle large dataset. A pushdown or restriction solution should be considered.~~ To unblock new command, performance improvement could be separated to a follow-up PR.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -355,6 +356,20 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
     return context.relBuilder.peek();
   }
 
+  @Override
+  public RelNode visitReverse(org.opensearch.sql.ast.tree.Reverse node, CalcitePlanContext context) {
+    visitChildren(node, context);
+    // Add ROW_NUMBER() column
+    RexNode rowNumber = context.relBuilder.aggregateCall(SqlStdOperatorTable.ROW_NUMBER)
+        .over().rowsTo(RexWindowBounds.CURRENT_ROW).as("__reverse_row_num__");
```
yuancu: `__reverse_row_num__` can be a shared static constant

docs/user/ppl/cmd/reverse.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Description
+============
+| Using ``reverse`` command to reverse the display order of search results. The same results are returned, but in reverse order.
+
+
+Syntax
+============
+reverse
+
```
yuancu, line 23: May need another section "Version" to indicate it's only available from version 3.2.0

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
```diff
@@ -32,6 +32,10 @@ public void onMatch(RelOptRuleCall call) {
 
     Integer limitValue = extractLimitValue(sort.fetch);
     Integer offsetValue = extractOffsetValue(sort.offset);
+    // Skip this rule if this is a reverse operation (indicated by row_number)
+    if (hasRowNumberFunction(sort)) {
```
yuancu: Such check can be moved ahead to the matching rule of the config at `OpenSearchLimitIndexScanRule.Config.DEFAULT`

Update: I found that it actually never step in to `hasRowNumberFunction` since a LogicalSort that reverses field `__row_number__` will not match this rule -- it fails the test `isLogicalSortLimit`

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
```diff
@@ -40,6 +44,18 @@ public void onMatch(RelOptRuleCall call) {
...
+  /**
+   * Check if the LogicalSort contains a row_number function, which indicates a reverse operation.
+   *
+   * @param sort The LogicalSort to check
+   * @return True if a row_number function is found, false otherwise
+   */
+  private boolean hasRowNumberFunction(LogicalSort sort) {
+    // Check if the sort has a row_number function in its digest
+    String digest = sort.getDigest();
+    return digest != null && digest.contains("row_number");
```
yuancu: Will the the digest of a logical sort ever contain `row_number`? You projected it as another column, it will refer it with its position instead of name.

I tested with the index bank with the ppl: `source=bank | reverse | fields address`. The digest I get of this logical sort is `LogicalSort.NONE.[20 DESC](input=RelSubset#2967,sort0=$20,dir0=DESC)`. The `20` corresponds to the added column in the following logical plan:

```
LogicalProject(address=[$2])
  LogicalSort(sort0=[$20], dir0=[DESC])
    LogicalProject(account_number=[$0], firstname=[$1], address=[$2], birthdate=[$3], date_value=[$4], gender=[$5], city=[$6], lastname=[$7], balance=[$8], employer=[$9], state=[$10], age=[$11], email=[$12], male=[$13], _id=[$14], _index=[$15], _score=[$16], _maxscore=[$17], _sort=[$18], _routing=[$19], __reverse_row_num__=[ROW_NUMBER() OVER ()])
      CalciteLogicalIndexScan(table=[[OpenSearch, bank]])
```

The digest doesn't contain the name of the column. You need another way to examine whether it's a reverse.

yuancu (commented):
Thank you for the contribution! The implementation is well structured and thoroughly tested. I left a few comments to discuss my concerns.

yuancu:
In the RFC #3873 , you mentioned the following case can be pushed down:

> The optimization is triggered when the logical plan has this exact shape:
> ```
> LogicalSort (reverse sort)
>    CalciteLogicalIndexScan
> ```
> ...
> If reverse immediately follows the source, pushdown to a descending index scan results in efficient queries.

However, in practice, they are not pushed down. For example, the physical plan of `source=bank | reverse` is as follows:

```
EnumerableCalc(expr#0..14=[{inputs}], proj#0..13=[{exprs}])
  EnumerableSort(sort0=[$14], dir0=[DESC])
    EnumerableWindow(window#0=[window(rows between UNBOUNDED PRECEDING and CURRENT ROW aggs [ROW_NUMBER()])])
      CalciteEnumerableIndexScan(table=[[OpenSearch, bank]], PushDownContext=[[PROJECT->[account_number, firstname, address, birthdate, date_value, gender, city, lastname, balance, employer, state, age, email, male]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"timeout":"1m","_source":{"includes":["account_number","firstname","address","birthdate","date_value","gender","city","lastname","balance","employer","state","age","email","male"],"excludes":[]}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```

Moreover, I doubt whether can it be pushed down. You added a new column to implement `reverse`. However, the new column isn't present in the datasource. Unless we support pushing down window function of `row_number()`, in which case we can push down logical sort after row number altogether, there is no way for the OpenSearch engine to sort by this non-existing column.

Nevertheless, I think it can be pushed down when there are preceding sort -- in this way, you can not rely on the added `__row_number__` column, but on the field to reverse the order of the results.

After fixing this, please kindly add tests in `CalciteExplainIT` to examine whether the generated physical plan is as expected.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReverseTest.java
```diff
@@ -0,0 +1,265 @@
...
+import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.schema.Statistic;
+import org.apache.calcite.schema.Statistics;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.test.CalciteAssert;
+import org.apache.calcite.tools.Frameworks;
+import org.junit.Test;
+
+/** Unit tests for {@code reverse} command in PPL. */
+public class CalcitePPLReverseTest extends CalcitePPLAbstractTest {
```
penghuo, line 12: In PPL module, we usually test the parser, and verifyPPLToSparkSQL. 
correctness teast should move to IT.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteReverseCommandIT.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.opensearch.sql.util.MatcherUtils.schema;
+import static org.opensearch.sql.util.MatcherUtils.verifyDataRowsInOrder;
+import static org.opensearch.sql.util.MatcherUtils.verifySchema;
+
+import java.io.IOException;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+import org.opensearch.sql.ppl.PPLIntegTestCase;
+
+public class CalciteReverseCommandIT extends PPLIntegTestCase {
```
penghuo, line 19: Could u add a test in ExplainIT for reverse command?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchLimitIndexScanRule.java
penghuo: Could u explain why we need to skip limit pushdown? What is logical plan before apply rule?

selsong: Thanks for pointing this out. After revisiting the issue, I realized the bug I encountered was actually due to an unrelated change on a different local branch. You're right that the limit pushdown logic itself is valid here, so I’ve reverted the unnecessary change to limit pushdown.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReverseTest.java
```diff
@@ -0,0 +1,160 @@
...
+                .defaultSchema(schema)
+                .traitDefs((List<RelTraitDef>) null)
+                .programs(Programs.heuristicJoinOrder(Programs.RULE_SET, true, 2));
+    }
+
+    @Test
+    public void testReverseParserSuccess() {
+        String ppl = "source=EMPLOYEES | reverse";
+        // Test that parser can successfully parse the reverse command
+        getRelNode(ppl);
```
yuancu: In such unit tests, the correctness of the logical plan and of the Spark SQL counterpart are also usually tested. You could refer to e.g. [CalcitePPLDedupTest](https://github.com/opensearch-project/sql/blob/main/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLDedupTest.java).

Btw, there are some existing tables that can be used for testing. E.g. you called `super(CalciteAssert.SchemaSpec.SCOTT_WITH_TEMPORAL)` in the constructor, the `EMP` table is set up for testing. You can reuse them unless you need a special schema.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLExplainIT.java
```diff
@@ -72,6 +72,18 @@ public void testExplainCommandSimple() throws IOException {
...
+  @Test
+  public void testExplainWithReverse() throws IOException {
+    String result = executeWithReplace("explain source=test | sort age | reverse | head 5");
+
+    // Verify that the plan contains a LogicalSort with fetch (from head 5)
+    assertTrue(result.contains("LogicalSort") && result.contains("fetch=[5]"));
+
+    // Verify that reverse added a ROW_NUMBER and another sort (descending)
+    assertTrue(result.contains("ROW_NUMBER()"));
+    assertTrue(result.contains("dir0=[DESC]"));
```
yuancu: I think this test is good in that it introduce minimum hard-coding strings. But adding the whole plan with `loadFromFile` also has its merits: code reader can directly refer to the generated plans to understand or check the correctness of the parsed query, without the need to execute the query by themselves.

You can also add the full plans later when you implement the push-down of the command.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReverseTest.java
selsong: Thanks! I modified the code to use the existing EMR table.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLExplainIT.java
selsong: Sounds good, I'll add the full loadFromFile later when I implement the pushdown.

selsong:
> 1. For new PPL commands, a Feature or RFC issue is required for syntax review before pull request.

An RFC is linked to this PR.

> 3. To unblock new command, performance improvement could be separated to a follow-up PR.

Yes, pushdown will be separated to a follow-up PR. Ran performance tests for various reverse command scenarios using the big5 dataset on EC2. Performance test statistics have been updated under **Performance** to the PR description

docs/user/ppl/cmd/reverse.rst
```diff
@@ -0,0 +1,117 @@
...
+Version
+=======
+3.2.0
+
+Syntax
+============
+reverse
+
+
+* No parameters: The reverse command takes no arguments or options.
```
penghuo, line 25: Could we add a note to call out limitation.
```
The reverse command processes the entire dataset. If applied directly to millions of records, it will consume significant memory resources on the coordinating node. Users should only apply the reverse command to smaller datasets, typically after aggregation operations.
```

penghuo (commented):
PR looks good.
@selsong  
1. Could you include the performance results from the current implementation in the PR description?
2. Could you create an issue to track the follow-up on window function pushdown?

docs/user/ppl/cmd/reverse.rst
selsong: Thanks! 
1. Yes, performance results from current implementation have been added to the PR description.
2. Issue #3924 has been created to track pushdown.

penghuo (dismissed):
@yuancu @LantaoJin  please take another look.

yuancu:
Looks good to me :)

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLExplainIT.java
```diff
@@ -72,6 +72,18 @@ public void testExplainCommandSimple() throws IOException {
     assertJsonEquals(expected, result);
   }
 
+  @Test
+  public void testExplainWithReverse() throws IOException {
```
LantaoJin: `CalcitePPLExplainIT` is just for `explain` command, please use `CalciteExplainIT` instead.

LantaoJin (changes requested):
Missing necessary updates in `PPLQueryDataAnonymizer` and `PPLQueryDataAnonymizerTest`

docs/user/ppl/cmd/reverse.rst
penghuo: add to index.rst. https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/index.rst

penghuo (approved):

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -355,6 +355,28 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
+    RexNode rowNumber =
+        context
+            .relBuilder
+            .aggregateCall(SqlStdOperatorTable.ROW_NUMBER)
+            .over()
+            .rowsTo(RexWindowBounds.CURRENT_ROW)
+            .as(REVERSE_ROW_NUM);
+    context.relBuilder.projectPlus(rowNumber);
+    // Sort by row number descending
+    context.relBuilder.sort(context.relBuilder.desc(context.relBuilder.field(REVERSE_ROW_NUM)));
```
LantaoJin, line 374: IMO, reverse by row_number should only apply to the plan (RelNode) which doesn't contain collation. For example:
`source=t | ... | sort - age | reverse` should apply sort by age asc, rather than row_number.
You can try:
```
RelCollation collation = context.relBuilder.peek().getTraitSet().getCollation();
if (collation == null || collation == RelCollations.EMPTY) {
  // Add ROW_NUMBER() column
    RexNode rowNumber =
   ...
} else {
  RelCollation reversedCollation = reverseCollation(collation);
  context.relBuilder.sort(reversedCollation)
}
```
Add `reverseCollation()` to PlanUtils.class
```
public static RelCollation reverseCollation(RelCollation original) {
  if (original == null || original.getFieldCollations().isEmpty()) {
        return original;
    }

    List<RelFieldCollation> reversedFields = new ArrayList<>();
    for (RelFieldCollation field : original.getFieldCollations()) {
        RelFieldCollation.Direction reversedDirection = 
            field.direction.reverse();
        
        RelFieldCollation reversedField = new RelFieldCollation(
            field.getFieldIndex(),
            reversedDirection,
            field.nullDirection
        );
        reversedFields.add(reversedField);
    }

    return RelCollations.of(reversedFields);
}
```

LantaoJin: Then add some tests:
```
source=t | ... | sort - age | reverse
source=t | ... | sort - age, + gender | reverse
source=t | ... | sort - age | reverse | reverse
source=t | ... | sort - age, + gender | reverse  | reverse
```

penghuo: @LantaoJin @selsong is it correct issue? or performance imporvement?
If it is performance improvement, we can track by issue https://github.com/opensearch-project/sql/issues/3924. And raise 2nd PR.

selsong: Thanks for the suggestion! This is a performance improvement, and it's already tracked under issue [#3924](https://github.com/opensearch-project/sql/issues/3924). The current implementation still produces correct results, but applying reverseCollation where applicable would avoid unnecessary computation. I'll address this enhancement in a follow-up PR.

LantaoJin: I don't think my suggestion is a pushdown enhancement, but is okey to refactor in separate PR.

Issue #3879, opened 2025-07-15
[FEATURE] Avoid using PIT search for simple query


qianheng-aws:
**Is your feature request related to a problem?**
For some simple queries whose plan or optimized plan is only a scan operator, we should avoid PIT search if the expected retrieved rows count(i.e. QUERY_SIZE_LIMIT) is less than the `max_result_window`. 

Examples for simple query:
```
source = table

// the filter and project can be push down
source = table | where ... | fields ...
```

**What solution would you like?**
Push down QUERY_SIZE_LIMIT into the final single scan operator.

**What alternatives have you considered?**
Add a limit operator on the top of the original plan and rely on Calcite optimizer to do push down. It has advantage that it will benefit some case where the plan is `Project-Scan` because of `SortProjectTransposeRule`, but the side effect is that it will change the plan by adding a limit operator.

**Do you have any additional context?**

PR #3880, opened 2025-07-15
Push down QUERY_SIZE_LIMIT

qianheng-aws:
### Description
Try to push down QUERY_SIZE_LIMIT if it's a simple plan with only scan.

This PR implements this by only pushing down QUERY_SIZE_LIMIT into the `OpenSearchRequestBuilder` of the final plan if it only contains a single scan operator.

This PR also fixes a bug in LIMIT push down by correct the cost computing logic. This bug will prevent multiple limit push down in some cases, see `CalciteExplainIT::testMultipleLimitExplain`.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3879

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
@@ -36,9 +36,9 @@ import org.opensearch.search.sort.ScoreSortBuilder;
 import org.opensearch.search.sort.SortBuilder;
 import org.opensearch.search.sort.SortBuilders;
 import org.opensearch.search.sort.SortOrder;
+import org.opensearch.sql.common.setting.Settings.Key;
 import org.opensearch.sql.data.type.ExprType;
 import org.opensearch.sql.opensearch.data.type.OpenSearchTextType;
-import org.opensearch.sql.common.setting.Settings.Key;
 import org.opensearch.sql.opensearch.request.OpenSearchRequestBuilder;
 import org.opensearch.sql.opensearch.storage.OpenSearchIndex;
 


core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -113,15 +115,16 @@ public static RelBuilder create(
...
+      FrameworkConfig config, JavaTypeFactory typeFactory, Integer querySizeLimit) {
     final Properties info = new Properties();
     if (config.getTypeSystem() != RelDataTypeSystem.DEFAULT) {
       info.setProperty(
           CalciteConnectionProperty.TYPE_SYSTEM.camelName(),
           config.getTypeSystem().getClass().getName());
     }
     try {
-      return new OpenSearchDriver().connect("jdbc:calcite:", info, null, typeFactory);
+      return new OpenSearchDriver(querySizeLimit).connect("jdbc:calcite:", info, null, typeFactory);
```
penghuo: Could we append Limit operator on each query instead of OpenSearchDriver constructor.

qianheng-aws: That's a workable options as well, I put it in the alternative solution.  https://github.com/opensearch-project/sql/issues/3879#issue-3231030764.

 It has side effect of changing the original plan if the append limit operator can not be push down. 
 
OPTION1: Push down QUERY_SIZE_LIMIT to the final single scan
- PROS: Won't change the plan, and the optimization process is efficient and straight forward
- CONS: Only improve on the restricted case of the single scan

OPTION2: Append LIMIT operator on the original plan
- PROS: Has improvement on more cases than single scan. e.g `Project-Scan` since it has `SortProjectTransposeRule` to swap `Limit` before the `Project`.
- CONS: Will change the final plan if Limit operator cannot be push down
 
 Which option do we prefer? @penghuo @LantaoJin

qianheng-aws: If we implement eval push down in the future, OPTION2 won't have its PROS since we can support pushing down all kinds of Project and left only a single Scan

LantaoJin: Option 1 is more like a specific optimization for the setting query_size. I prefer to option 2 with a new LogicalSort such `LogicalQueryLimit`. Anyway, I am ok to both Option 1 and 2.

penghuo: +1 on option2, 

> CONS: Will change the final plan if Limit operator cannot be push down

Could you elaborate on this? Are you suggesting that adding a LIMIT clause would change the result of the EXPLAIN plan? If that’s the concern, I’m okay with it. We explicitly enforce a querySizeLimit for every query anyway.


Latest impl is better, scanWithLimit is clean to me.

penghuo (dismissed):

LantaoJin (dismissed):
Note, the current implementation is option 1. I am also open to this option so I will approve it. @penghuo please merge it if the current implementation works for you too.

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
qianheng-aws: > Are you suggesting that adding a LIMIT clause would change the result of the EXPLAIN plan?

Yes.

Changing the plan will have suspicion of changing customers' intension although we do have restriction on the final results. 

And I was wondering why Hive or Spark doesn't do similar optimization of appending limit operator, it's said (by LLM) that the current implementation of counting row count on the final iteration has other advantages including:
1. Plan reuse for cases like pagination. Although we don't have such feature for PPL, but maybe needed in the future.
2. Better memory management. It currently doesn't applies for us since the whole process happens in one coordinator. But it will make sense once we change to distribution execution.
3. Keep the plan semantically equivalent to user's SQL.

So for long term consideration, keeping the plan unchanged makes more sense and is a standard practice.

LantaoJin (approved):

penghuo (approved):

LantaoJin:
@qianheng-aws please backport to 2.19-dev.

Issue #3877, opened 2025-07-14
[RFC] Add `table` Command to PPL (Calcite Engine)


aalva500-prog:
### Description
Implement a `table` command in OpenSearch PPL that provides field selection operations. This command will return search results with only the specified fields.

### Use Cases
- Selecting specific fields from search results
- Filtering result sets to include only fields of interest
- Controlling field order in results
- Using wildcard patterns for field selection

## 2. Final Syntax

```
table <wc-field-list>
```

### Parameters
- `<wc-field-list>`: A space or comma-delimited list of field names. Supports wildcard patterns using asterisk (*).

### Options
- Wildcard patterns: Supports various wildcard patterns using asterisk (*)
- Prefix wildcard: `EMP*` (matches fields starting with "EMP")
- Suffix wildcard: `*NO` (matches fields ending with "NO")
- Contains wildcard: `*E*` (matches fields containing "E")
- Mixed explicit and wildcard: `ENAME, EMP*, JOB`

### Basic Usage
```
source=EMP | table EMPNO, ENAME, JOB
```

Returns results containing only the EMPNO, ENAME, and JOB fields.

#### Using Renamed Fields
```
source=EMP | rename EMPNO as emp_id, ENAME as emp_name | table emp_id, emp_name, emp*
```

Returns results containing the emp_id, emp_name fields and any fields matching the emp* pattern.

#### Sorting and Filtering
```
source=EMP | where SAL > 1000 | sort - SAL | table ENAME, SAL, DEPTNO | head 3
```

Returns the top 3 results containing only ENAME, SAL, and DEPTNO fields.

#### Using Wildcards
```
source=EMP | table ENAME, JOB, DEPT*
```

Returns results containing ENAME, JOB, and any fields matching the DEPT* pattern.

#### Table with Evaluation
```
source=EMP | dedup DEPTNO | eval dept_type=case(DEPTNO=10, 'accounting' else 'other') | table EMPNO, dept_type
```

Returns results containing only EMPNO and dept_type fields.

#### Table with Multiple Wildcards
```
source=EMP | table *NAME, *NO, JOB
```

Returns results containing fields matching *NAME, *NO patterns, and the JOB field.

### Best Practices
- Place the `table` command at the end of search pipelines for optimal performance
- Use wildcards to select groups of related fields efficiently
- Perform field renaming before using the `table` command

### Optimization Notes
- The `table` command is a non-streaming, transforming command that processes the entire result set at once
[...]

PR #3885, opened 2025-07-15
Support `table` command in PPL (Calcite Engine)

aalva500-prog:
### Description
Implement a `table` command for OpenSearch PPL that provides field selection and projection functionality. The command allows users to explicitly specify which fields should be included in the query results, similar to the SELECT clause in SQL.

### Features Added
- **Core `table` command:** `table <field1>, <field2>, ... [, <fieldN>]`
- **Wildcard support:** Handles patterns like `log.*` to select multiple fields
- **Field expressions:** Supports field expressions and functions in projections
- **Calcite integration**: Full support in Calcite engine with proper field projection
- **Pipeline integration:** Works seamlessly with other PPL commands like `where` and `stats`

#### Grammar & Parsing
- Extended PPL parser with table command grammar
- Implemented AST node (Table.java) with visitor pattern support
- Added tableCommand rule to handle field projections

#### Calcite Engine Integration
- Added `visitTable()` method in `CalciteRelNodeVisitor`
- Implemented wildcard expansion for field patterns
- Proper field projection
- Maintains field order as specified in the command

### Usage Examples
```
-- Basic table with specific fields
source=accounts | table account_number, firstname, lastname

-- Wildcard field selection
source=logs | table time, host*, error_*

-- Mixed field specification
source=data | table timestamp, value*, status, response_time
```

### Related Issues
Resolves #[[3877](https://github.com/opensearch-project/sql/issues/3877)]

```
=== SIMPLE PERFORMANCE TEST ===
Testing working `table` command operations with direct curl

== BASIC TABLE OPERATIONS ===
🧪 Testing: Basic field selection
📝 Query: source=big5 | table `@timestamp`, `metrics.size` | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 16.11ms
   ✅ Iteration 2: 15.36ms
   ✅ Iteration 3: 15.02ms
   ✅ Iteration 4: 15.47ms
   ✅ Iteration 5: 15.23ms
📊 Average time: 15.43ms (5/5 successful)

🧪 Testing: Multiple fields with comma delimiter
📝 Query: source=big5 | table `@timestamp`, `metrics.size`, `metrics.tmin` | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 14.86ms
   ✅ Iteration 2: 14.74ms
   ✅ Iteration 3: 14.66ms
   ✅ Iteration 4: 14.67ms
   ✅ Iteration 5: 14.75ms
📊 Average time: 14.73ms (5/5 successful)

🧪 Testing: Multiple fields with space delimiter
📝 Query: source=big5 | table `@timestamp` `metrics.size` `metrics.tmin` | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 15.94ms
   ✅ Iteration 2: 16.19ms
   ✅ Iteration 3: 15.51ms
   ✅ Iteration 4: 15.13ms
   ✅ Iteration 5: 15.00ms
📊 Average time: 15.55ms (5/5 successful)

🧪 Testing: Single field selection
📝 Query: source=big5 | table `@timestamp` | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 14.00ms
   ✅ Iteration 2: 14.09ms
   ✅ Iteration 3: 14.66ms
   ✅ Iteration 4: 14.34ms
   ✅ Iteration 5: 14.37ms
📊 Average time: 14.29ms (5/5 successful)

🧪 Testing: Wildcard field selection - prefix
📝 Query: source=big5 | table metrics* | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 15.15ms
   ✅ Iteration 2: 14.82ms
   ✅ Iteration 3: 14.47ms
   ✅ Iteration 4: 14.02ms
   ✅ Iteration 5: 14.11ms
📊 Average time: 14.51ms (5/5 successful)

🧪 Testing: Wildcard field selection - suffix
📝 Query: source=big5 | table *size | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 14.08ms
   ✅ Iteration 2: 14.26ms
   ✅ Iteration 3: 13.91ms
   ✅ Iteration 4: 14.88ms
   ✅ Iteration 5: 14.65ms
📊 Average time: 14.35ms (5/5 successful)

🧪 Testing: Wildcard field selection - contains
📝 Query: source=big5 | table *cloud* | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 14.30ms
   ✅ Iteration 2: 14.29ms
   ✅ Iteration 3: 13.89ms
   ✅ Iteration 4: 14.47ms
   ✅ Iteration 5: 14.41ms
📊 Average time: 14.27ms (5/5 successful)

🧪 Testing: Mixed explicit and wildcard fields
📝 Query: source=big5 | table `@timestamp`, metrics* | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 14.24ms
   ✅ Iteration 2: 14.52ms
   ✅ Iteration 3: 14.59ms
   ✅ Iteration 4: 14.91ms
   ✅ Iteration 5: 14.96ms
📊 Average time: 14.64ms (5/5 successful)

🧪 Testing: Table with fields
📝 Query: source=big5 | fields `@timestamp`, `metrics.size` | table `@timestamp` | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 14.32ms
   ✅ Iteration 2: 14.11ms
   ✅ Iteration 3: 14.21ms
   ✅ Iteration 4: 13.87ms
   ✅ Iteration 5: 13.85ms
📊 Average time: 14.07ms (5/5 successful)

🧪 Testing: Table with all fields
📝 Query: source=big5 | table * | head 10
🔁 Running 5 iterations...
   ✅ Iteration 1: 15.27ms
   ✅ Iteration 2: 14.98ms
   ✅ Iteration 3: 14.72ms
   ✅ Iteration 4: 15.03ms
   ✅ Iteration 5: 15.36ms
📊 Average time: 15.07ms (5/5 successful)

=== LOAD TESTING ===
🧪 Testing: Large result set (100 rows)
📝 Query: source=big5 | table `@timestamp`, `metrics.size`, `metrics.tmin` | head 100
🔁 Running 3 iterations...
   ✅ Iteration 1: 16.65ms
   ✅ Iteration 2: 16.97ms
   ✅ Iteration 3: 16.88ms
📊 Average time: 16.83ms (3/3 successful)

🧪 Testing: Larger result set (500 rows)
📝 Query: source=big5 | table `@timestamp`, `metrics.size`, `metrics.tmin` | head 500
🔁 Running 3 iterations...
   ✅ Iteration 1: 24.33ms
   ✅ Iteration 2: 24.19ms
   ✅ Iteration 3: 23.31ms
📊 Average time: 23.94ms (3/3 successful)

🧪 Testing: Very large result set (1000 rows)
📝 Query: source=big5 | table `@timestamp`, `metrics.size`, `metrics.tmin` | head 1000
🔁 Running 2 iterations...
   ✅ Iteration 1: 43.14ms
   ✅ Iteration 2: 33.11ms
📊 Average time: 38.12ms (2/2 successful)

=== DIFFERENT FIELD COMBINATIONS ===
🧪 Testing: Few fields (2)
📝 Query: source=big5 | table `@timestamp`, `metrics.size` | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 15.93ms
   ✅ Iteration 2: 15.96ms
   ✅ Iteration 3: 15.76ms
📊 Average time: 15.88ms (3/3 successful)

🧪 Testing: Medium fields (5)
📝 Query: source=big5 | table `@timestamp`, `metrics.size`, `metrics.tmin`, `host`, `message` | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 17.22ms
   ✅ Iteration 2: 17.31ms
   ✅ Iteration 3: 17.95ms
📊 Average time: 17.49ms (3/3 successful)

🧪 Testing: Many fields (wildcard)
📝 Query: source=big5 | table * | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 19.80ms
   ✅ Iteration 2: 19.50ms
   ✅ Iteration 3: 18.42ms
📊 Average time: 19.24ms (3/3 successful)

🧪 Testing: Complex wildcard pattern
📝 Query: source=big5 | table *met*size* | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 15.72ms
   ✅ Iteration 2: 14.52ms
   ✅ Iteration 3: 14.28ms
📊 Average time: 14.84ms (3/3 successful)

🧪 Testing: Multiple wildcards
📝 Query: source=big5 | table *time*, met*, *size, *cloud* | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 16.02ms
   ✅ Iteration 2: 16.69ms
   ✅ Iteration 3: 16.48ms
📊 Average time: 16.39ms (3/3 successful)

🧪 Testing: Field ordering
📝 Query: source=big5 | table `metrics.tmin`, `@timestamp`, `metrics.size` | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 15.36ms
   ✅ Iteration 2: 15.31ms
   ✅ Iteration 3: 15.55ms
📊 Average time: 15.40ms (3/3 successful)

=== COMMAND COMBINATIONS ===
🧪 Testing: Table with filter
📝 Query: source=big5 | where `metrics.size` > 1000 | table `@timestamp`, `metrics.size`, `metrics.tmin` | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 280.30ms
   ✅ Iteration 2: 31.61ms
   ✅ Iteration 3: 31.16ms
📊 Average time: 114.35ms (3/3 successful)

🧪 Testing: Table with sort
📝 Query: source=big5 | sort `metrics.size` | table `@timestamp`, `metrics.size` | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 320.67ms
   ✅ Iteration 2: 27.24ms
   ✅ Iteration 3: 27.33ms
📊 Average time: 125.08ms (3/3 successful)

🧪 Testing: Table with limit
📝 Query: source=big5 | table `@timestamp`, `metrics.size` | head 5
🔁 Running 3 iterations...
   ✅ Iteration 1: 14.17ms
   ✅ Iteration 2: 13.48ms
   ✅ Iteration 3: 13.34ms
📊 Average time: 13.66ms (3/3 successful)

🧪 Testing: Complex query
📝 Query: source=big5 | where `metrics.size` > 1000 | sort - `metrics.size` | table `@timestamp`, `metrics.size`, `metrics.tmin` | head 3
🔁 Running 3 iterations...
   ✅ Iteration 1: 49.29ms
   ✅ Iteration 2: 33.13ms
   ✅ Iteration 3: 33.41ms
📊 Average time: 38.61ms (3/3 successful)

🧪 Testing: Table in middle of pipeline
📝 Query: source=big5 | where `metrics.size` > 1000 | table `@timestamp`, `metrics.size` | sort `metrics.size` | head 10
🔁 Running 3 iterations...
   ✅ Iteration 1: 71.95ms
   ✅ Iteration 2: 56.01ms
   ✅ Iteration 3: 56.22ms
📊 Average time: 61.39ms (3/3 successful)

=== BASELINE COMPARISON ===
🧪 Testing: Baseline (no table)
📝 Query: source=big5 | head 100
🔁 Running 5 iterations...
   ✅ Iteration 1: 21.45ms
   ✅ Iteration 2: 21.54ms
   ✅ Iteration 3: 21.36ms
   ✅ Iteration 4: 21.22ms
   ✅ Iteration 5: 21.35ms
📊 Average time: 21.38ms (5/5 successful)

🧪 Testing: Baseline with fields
📝 Query: source=big5 | fields `@timestamp`, `metrics.size` | head 100
🔁 Running 5 iterations...
   ✅ Iteration 1: 15.81ms
   ✅ Iteration 2: 15.46ms
   ✅ Iteration 3: 15.44ms
   ✅ Iteration 4: 16.05ms
   ✅ Iteration 5: 15.45ms
📊 Average time: 15.64ms (5/5 successful)

🧪 Testing: Production-like query (complex pipeline)
📝 Query: source=big5 | where `metrics.size` > 1000 | sort - `metrics.size` | table `@timestamp`, `metrics.size`, `metrics.tmin` | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 248.70ms
   ✅ Iteration 2: 39.25ms
   ✅ Iteration 3: 39.09ms
📊 Average time: 109.01ms (3/3 successful)

🧪 Testing: Production-like query (aggregation + table)
📝 Query: source=big5 | stats avg(`metrics.size`) as avg_size by `host` | sort - avg_size | table host, avg_size | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 40209.47ms
   ✅ Iteration 2: 39231.64ms
   ✅ Iteration 3: 42993.25ms
📊 Average time: 40811.45ms (3/3 successful)

=== CONCURRENT LOAD TEST ===
🚀 Running concurrent load test (3 parallel requests)...
Temp directory: /tmp/tmp.NhhxCGcUfU
Starting concurrent request 1...
Starting concurrent request 2...
Starting concurrent request 3...
Concurrent test results:
Request 1:
📊 Average time: 81.26ms (1/1 successful)

Request 2:
📊 Average time: 77.57ms (1/1 successful)

Request 3:
📊 Average time: 77.97ms (1/1 successful)


=== WHITESPACE DELIMITED FIELDS TESTING ===
🧪 Testing: Whitespace delimited fields - basic
📝 Query: source=big5 | table `@timestamp` `metrics.size` `metrics.tmin` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 16.34ms
   ✅ Iteration 2: 15.59ms
   ✅ Iteration 3: 15.25ms
📊 Average time: 15.72ms (3/3 successful)

🧪 Testing: Whitespace delimited fields - with wildcards
📝 Query: source=big5 | table `@timestamp` metrics* `host.name` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 15.42ms
   ✅ Iteration 2: 17.86ms
   ✅ Iteration 3: 16.33ms
📊 Average time: 16.53ms (3/3 successful)

🧪 Testing: Whitespace delimited fields - in complex pipeline
📝 Query: source=big5 | where `metrics.size` > 1000 | table `@timestamp` `metrics.size` `host.name` | sort `metrics.size` | head 10
🔁 Running 3 iterations...
   ✅ Iteration 1: 490.16ms
   ✅ Iteration 2: 59.20ms
   ✅ Iteration 3: 60.21ms
📊 Average time: 203.19ms (3/3 successful)

🧪 Testing: Whitespace delimited fields - single field
📝 Query: source=big5 | table `@timestamp` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 14.37ms
   ✅ Iteration 2: 13.73ms
   ✅ Iteration 3: 14.06ms
📊 Average time: 14.05ms (3/3 successful)

=== DELIMITER COMPARISON ===
🧪 Testing: Comma-delimited fields
📝 Query: source=big5 | table `@timestamp`, `metrics.size`, `metrics.tmin`, `host.name`, `message` | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 16.37ms
   ✅ Iteration 2: 16.39ms
   ✅ Iteration 3: 16.55ms
📊 Average time: 16.43ms (3/3 successful)

🧪 Testing: Whitespace-delimited fields
📝 Query: source=big5 | table `@timestamp` `metrics.size` `metrics.tmin` `host.name` `message` | head 50
🔁 Running 3 iterations...
   ✅ Iteration 1: 17.80ms
   ✅ Iteration 2: 17.85ms
   ✅ Iteration 3: 17.24ms
📊 Average time: 17.63ms (3/3 successful)

=== ADVANCED FIELD TESTING ===
🧪 Testing: Nested fields - agent
📝 Query: source=big5 | table `agent.name`, `agent.type`, `agent.version` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 15.05ms
   ✅ Iteration 2: 14.64ms
   ✅ Iteration 3: 14.92ms
📊 Average time: 14.87ms (3/3 successful)

🧪 Testing: Nested fields - aws.cloudwatch
📝 Query: source=big5 | table `aws.cloudwatch.log_group`, `aws.cloudwatch.log_stream`, `aws.cloudwatch.ingestion_time` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 14.71ms
   ✅ Iteration 2: 14.44ms
   ✅ Iteration 3: 14.14ms
📊 Average time: 14.43ms (3/3 successful)

🧪 Testing: Nested fields - data_stream
📝 Query: source=big5 | table `data_stream.dataset`, `data_stream.namespace`, `data_stream.type` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 14.52ms
   ✅ Iteration 2: 14.20ms
   ✅ Iteration 3: 14.41ms
📊 Average time: 14.37ms (3/3 successful)

🧪 Testing: Mixed field types
📝 Query: source=big5 | table `@timestamp`, `host.name`, `metrics.size`, `aws.cloudwatch.log_group` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 14.91ms
   ✅ Iteration 2: 15.18ms
   ✅ Iteration 3: 15.07ms
📊 Average time: 15.05ms (3/3 successful)

🧪 Testing: Special fields
📝 Query: source=big5 | table `_id`, `_index`, `_score` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 14.60ms
   ✅ Iteration 2: 14.97ms
   ✅ Iteration 3: 14.28ms
📊 Average time: 14.61ms (3/3 successful)

🧪 Testing: Event fields
📝 Query: source=big5 | table `event.dataset`, `event.id`, `event.ingested` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 14.58ms
   ✅ Iteration 2: 14.85ms
   ✅ Iteration 3: 14.74ms
📊 Average time: 14.72ms (3/3 successful)

🧪 Testing: Multi-field sorting
📝 Query: source=big5 | sort `metrics.size`, `@timestamp` | table `@timestamp`, `metrics.size`, `metrics.tmin` | head 20
🔁 Running 3 iterations...
   ✅ Iteration 1: 44382.67ms
   ✅ Iteration 2: 45540.15ms
   ✅ Iteration 3: 45333.47ms
📊 Average time: 45085.43ms (3/3 successful)

=== WILDCARD ALL FIELDS TESTING ===
🧪 Testing: Wildcard all fields (source=big5 | table *)
📝 Query: source=big5 | table * | head 10
🔁 Running 3 iterations...
   ✅ Iteration 1: 73.99ms
   ✅ Iteration 2: 15.43ms
   ✅ Iteration 3: 14.68ms
📊 Average time: 34.70ms (3/3 successful)

🧪 Testing: Direct query without table (baseline)
📝 Query: source=big5 | head 10
🔁 Running 3 iterations...
   ✅ Iteration 1: 15.08ms
   ✅ Iteration 2: 14.79ms
   ✅ Iteration 3: 14.61ms
📊 Average time: 14.82ms (3/3 successful)

=== SUMMARY ===
✅ Performance testing complete
💡 Key insights:
   - Check average times for each operation type
   - Compare table operations vs baseline
   - Look for performance degradation with larger result sets
   - Monitor concurrent request performance
   - Evaluate nested field access performance
   - Compare comma-delimited vs whitespace-delimited field performance
   - Evaluate table command performance in different pipeline positions

🎯 Next steps:
   - Optimize any slow operations
   - Test with production-like query volumes

==================================
🏁 PERFORMANCE TEST COMPLETE
==================================
```

### Check List


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1434,6 +1434,80 @@ private void flattenParsedPattern(
...
+        } else {
+          // Handle regular field reference
+          projectList.add(rexVisitor.analyze(expr, context));
+          explicitFieldNames.add(fieldName);
+        }
+      } else {
+        // Handle non-field expressions (functions, literals, etc.)
+        projectList.add(rexVisitor.analyze(expr, context));
+        // Generate a default name for expressions without explicit names
+        explicitFieldNames.add("expr" + projectList.size());
```
yuancu: I'm curious what are the cases where a selected item (`UnresolvedExpression`) is not a `Field`

From your implementation of `visitTableCommand`, it seems all items are `Field`. If so, this would be a dead code.

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,126 @@
...
+.. rubric:: Table of contents
+
+.. contents::
+   :local:
+   :depth: 2
+
+
+Description
+============
+The ``table`` command returns a table that is formed by only the fields that you specify in the arguments. Columns are displayed in the same order that fields are specified. Column headers are the field names. Rows are the field values. Each row represents an event.
```
yuancu: *event* is the terminology of splunk. I think we rarely use (or avoid using) event to represent a row in our documentation?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1434,6 +1434,80 @@ private void flattenParsedPattern(
...
+          List<String> matchedFields = expandWildcard(fieldName, availableFields);
+          for (String matchedField : matchedFields) {
+            // Add each matched field to the projection list
+            projectList.add(context.relBuilder.field(matchedField));
+            explicitFieldNames.add(matchedField);
+          }
+        } else {
+          // Handle regular field reference
+          projectList.add(rexVisitor.analyze(expr, context));
+          explicitFieldNames.add(fieldName);
```
yuancu, line 1479: It comes to me that: should we de-duplicate the field names? E.g. both `na*` and `*me` will match name, should it appear twice in the results?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,1044 @@
...
+   */
+  @Test
+  public void testTableWithPatternWildcard() throws IOException {
+    JSONObject actual =
+        executeQuery(String.format("source=%s | table account* | head 1", TEST_INDEX_ACCOUNT));
+
+    JSONArray datarows = actual.getJSONArray("datarows");
+    assertTrue("Should return exactly one row", datarows.length() == 1);
+    assertTrue(
+        "Should have at least account_number field", actual.getJSONArray("schema").length() >= 1);
```
yuancu: The first assert can be replaced with `verifyNumOfRows`.

The second assert can be replaced with `verifySchema`. The error message does not match what you check.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,1044 @@
...
+    for (int i = 0; i < schema.length(); i++) {
+      String fieldName = schema.getJSONObject(i).getString("name");
+      if ("account_number".equals(fieldName)) hasAccountNumber = true;
+      if ("firstname".equals(fieldName)) hasFirstname = true;
+      if ("age".equals(fieldName)) hasAge = true;
+    }
+
+    assertTrue("Should include account_number field", hasAccountNumber);
+    assertTrue("Should include firstname field", hasFirstname);
+    assertTrue("Should include age field", hasAge);
```
yuancu: Ditto.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,1044 @@
...
+            String.format(
+                "source=%s | sort age | table account_number, age | head 3", TEST_INDEX_ACCOUNT));
+
+    verifySchema(actual, schema("account_number", "bigint"), schema("age", "bigint"));
+
+    JSONArray datarows = actual.getJSONArray("datarows");
+    long prevAge = 0;
+    for (int i = 0; i < datarows.length(); i++) {
+      long currentAge = datarows.getJSONArray(i).getLong(1);
+      assertTrue("Ages should be in ascending order", currentAge >= prevAge);
```
yuancu: Can be replaced with `verifyDataRowsInOrder`

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,1044 @@
...
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | table account_number, state | stats count() by state | sort state",
+                TEST_INDEX_ACCOUNT));
+
+    verifySchema(actual, schema("count()", "bigint"), schema("state", "string"));
+
+    JSONArray datarows = actual.getJSONArray("datarows");
+    assertTrue("Should have multiple states in the result", datarows.length() > 10);
```
yuancu: Using `verifyNumOfRows` would be better that checking length manually. It would be even better to use `verifyDataRows`, since it can cover every functionality you are testing here (per the javadoc of this test).

You can go over all the rest tests and scrutinize them with the mentioned criteria.

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,126 @@
...
+
+
+Usage
+============
+The ``table`` command is a transforming command.
+
+
+Command Type
+============
+The ``table`` command is a non-streaming command. If you are looking for a streaming command similar to the ``table`` command, use the ``fields`` command.
```
yuancu: Since no other commands have such a streaming / non-streaming categorization, it feels abrupt to mention that this is non-streaming one. I think it's only meaningful if we have categorization for every command or if its `fields` counterpart is a streaming command.

One concern: isn't this documentation too similar to [that of Splunk](https://docs.splunk.com/Documentation/Splunk/9.4.2/SearchReference/Table)? I am worried that it may incur legal issues on copyright.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
aalva500-prog: @yuancu You're absolutely right, the Table AST node is constructed specifically for the PPL table command, and it only creates Field expressions in the projection list. I'll remove the dead code. Thanks!

docs/user/ppl/cmd/table.rst
aalva500-prog: @yuancu You're absolutely right! The "event" terminology is indeed Splunk-specific. This should be updated to use more generic terminology.

docs/user/ppl/cmd/table.rst
aalva500-prog: @yuancu You raise excellent points on both concerns:

- Streaming/non-streaming categorization : You're right that mentioning this without context for other commands feels abrupt and inconsistent.
- Copyright concerns : This is a very valid concern. The structure, phrasing, and even some examples appear very similar to Splunk documentation, which could indeed pose legal risks.

Let me revise the documentation to be more original and remove the streaming categorization. Thanks!

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
aalva500-prog: @yuancu, you're absolutely right to be concerned about duplication. The current implementation doesn't de-duplicate field names, which means if multiple wildcard patterns match the same field, it will appear multiple times in the result. I'll handle this issue. Thanks for raising!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: It should be handled. Please review, thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: It should be handled. Please review, thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: It should be handled. Please review, thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: It should be handled. Please review, thanks!

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
aalva500-prog: It should be handled. Please review, thanks!

#### Example query output:

Screenshot 2025-07-25 at 3 26 54 PM

yuancu: To be honest, I'm also not sure about. I was asking whether we should de-duplicate it 😂. As `fields a, a` seems to keep both of them.

For me, it should be reasonable to perform de-duplication in wildcard match, because users did not explicitly specify that they want it more than once as that it `fields`

core/src/main/java/org/opensearch/sql/ast/AbstractNodeVisitor.java
```diff
@@ -112,6 +113,21 @@ public T visitRelation(Relation node, C context) {
     return visitChildren(node, context);
   }
 
+  /**
+   * Visit a Table node in the AST.
+   *
+   * <p>This method processes a Table node which represents a data source in SQL queries. Tables are
+   * fundamental relation objects that serve as leaf nodes in the query plan.
```
yuancu: nitpick: are the nodes in question guaranteed to be leaf nodes? I noticed that the table node has a child field, which seems to represent the source data for the command. 

 I haven’t fully explored the AST myself, so please let me know if I’m misunderstanding or missing something.

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,121 @@
...
+
+
+Example 2: Using wildcards
+--------------------------
+
+PPL query::
+
+    os> source=employees | table ENAME, JOB, DEPT*;
+
+Returns results containing ENAME, JOB, and any fields matching the DEPT* pattern.
```
yuancu: Could you please also include the full result table for a more straightforward exemplification (as most other command documentations do).

I remember this doc used to have it. Sorry if any of my previous comments implied it's better to not include it.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+  @Test
+  public void testTableWithMixedFieldSpecification() throws IOException {
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | table firstname, a*, lastname | head 3", TEST_INDEX_ACCOUNT));
+
+    verifyNumOfRows(actual, 3);
+
+    verifyNumOfRows(actual, 3);
```
yuancu: Redundant check. Maybe you intended to use `verifySchema`?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+            String.format("source=%s | table *name, *number, age | head 3", TEST_INDEX_ACCOUNT));
+
+    verifySchema(
+        actual,
+        schema("firstname", "string"),
+        schema("lastname", "string"),
+        schema("account_number", "bigint"),
+        schema("age", "bigint"));
+
+    // Schema already verified above
```
yuancu: nitpick: meaningless comment

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+   *
+   * @throws IOException if query execution fails
+   */
+  @Test
+  public void testTableWithMultipleWildcardPatterns() throws IOException {
+    JSONObject actual =
+        executeQuery(
+            String.format("source=%s | table firstname, f*, a* | head 3", TEST_INDEX_ACCOUNT));
+
+    verifyNumOfRows(actual, 3);
```
yuancu: Should at least verify schema

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+  }
+
+  /**
+   * Tests space-delimited only syntax without commas. Verifies that fields can be specified with
+   * spaces only, without requiring commas as separators.
+   *
+   * @throws IOException if query execution fails
+   */
+  @Test
+  public void testTableSpaceDelimitedOnly() throws IOException {
```
yuancu: I think this test and `testTableSpaceDelimitedFieldOrder` can be merged.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+   * @throws IOException if query execution fails
+   */
+  @Test
+  public void testTableSpaceDelimitedFieldOrder() throws IOException {
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | table firstname account_number age | head 1", TEST_INDEX_ACCOUNT));
+
+    verifyNumOfRows(actual, 1);
```
yuancu: Should at least verify schema

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+   * demonstrates how to efficiently select related fields using wildcards.
+   *
+   * @throws IOException if query execution fails
+   */
+  @Test
+  public void testTableWithLargeFieldSetOptimization() throws IOException {
+    JSONObject actual =
+        executeQuery(String.format("source=%s | table a*, *name, s* | head 3", TEST_INDEX_ACCOUNT));
+
+    verifyNumOfRows(actual, 3);
```
yuancu: Should at least verify schema

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+   * @throws IOException if query execution fails
+   */
+  @Test
+  public void testTableWithDuplicateFields() throws IOException {
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | table account_number, account_number | head 1", TEST_INDEX_ACCOUNT));
+
+    verifyNumOfRows(actual, 1);
```
yuancu: ditto

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+  }
+
+  /**
+   * Tests the table command with field renaming. Verifies that the table command works correctly
+   * with renamed fields after applying the rename command.
+   *
+   * @throws IOException if query execution fails
+   */
+  @Test
+  public void testTableWithRenameAndWildcard() throws IOException {
```
yuancu: I think there is no wildcard in this and the above test?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+   * named fields don't duplicate when also matched by wildcards.
+   *
+   * @throws IOException if query execution fails
+   */
+  @Test
+  public void testTableFieldDeduplicationExplicitAndWildcard() throws IOException {
+    JSONObject actual =
+        executeQuery(String.format("source=%s | table age, a* | head 3", TEST_INDEX_ACCOUNT));
+
+    verifyNumOfRows(actual, 3);
```
yuancu: Should at least verify schema

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
```diff
@@ -0,0 +1,757 @@
...
+   *
+   * @throws IOException if query execution fails
+   */
+  @Test
+  public void testTableFieldDeduplicationMultipleOverlaps() throws IOException {
+    JSONObject actual =
+        executeQuery(
+            String.format("source=%s | table *name, first*, f* | head 3", TEST_INDEX_ACCOUNT));
+
+    verifyNumOfRows(actual, 3);
```
yuancu: ditto

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,121 @@
+=============
+table
+=============
```
yuancu, line 3: I think it would be better to mention its difference and similarities with `fields`, otherwise users may get confused about these two commands.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -113,6 +115,16 @@ showDataSourcesCommand
    : SHOW DATASOURCES
    ;
 
+tableCommand
+   : TABLE wcFieldList
+   | TABLE wcFieldExpressionList
```
yuancu: nitpick: the naming of the two is a little misleading. It sounds like they differ in that one is a list of expressions while another isn't. However, they actually differ in that whether they are comma-separated or space-separated.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
aalva500-prog: IMO you are right, we should de-deplicate in wildcard match. Also, there is no duplication in SPL, so I think your point is valid.

core/src/main/java/org/opensearch/sql/ast/AbstractNodeVisitor.java
aalva500-prog: Thank you for catching that! You're correct that Table nodes are not guaranteed to be leaf nodes. The Table class has a child field that represents the source data for the table operation, and the getChild() method returns either an empty list (when there's no child) or a singleton list containing the child node.

I'll update the documentation to accurately reflect that Table nodes can have children and remove the incorrect assertion that they are leaf nodes.

ppl/src/main/antlr/OpenSearchPPLParser.g4
aalva500-prog: I have changed the naming convention, as recommended.

docs/user/ppl/cmd/table.rst
aalva500-prog: I have updated the documentation, as recommended.

docs/user/ppl/cmd/table.rst
aalva500-prog: I have updated the documentation, as recommended.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTableCommandIT.java
aalva500-prog: I have refactored the whole class and consolidated the test cases. Please take a look. Thanks!

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,256 @@
...
+Comparison with fields Command
+==============================
+
+The ``table`` and ``fields`` commands are both used for field selection but have key differences:
+
+**Similarities:**
+
+* Both commands filter search results to include specific fields
+* Both support comma-delimited and space-delimited field lists
+* Both support wildcard patterns for field selection
```
yuancu: These two statements are incorrect w.r.t. the current implementation, where only `table` support space-delimited lists and wildcard patterns.

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,256 @@
...
+
+Arguments
+============
+<wc-field-list>
+  A space or comma-delimited list of field names with wildcard support. Supports various wildcard patterns using asterisk (*):
+  
+  * Prefix wildcard: ``EMP*`` (matches fields starting with "EMP")
+  * Suffix wildcard: ``*NO`` (matches fields ending with "NO")
+  * Contains wildcard: ``*E*`` (matches fields containing "E")
+  * Mixed explicit and wildcard: ``ENAME, EMP*, JOB``
```
yuancu, line 39: Please mention somewhere in this doc that resulting fields will be de-duplicated.

docs/user/ppl/cmd/table.rst
aalva500-prog: Yes, that's is correct, however, we are currently working on supporting wildcard and space delimiter in fields command. I can add those two sentences when the fields command has those features, np 😊

docs/user/ppl/cmd/table.rst
aalva500-prog: I have updated the doc file. Please review, thanks!

docs/user/ppl/cmd/table.rst
aalva500-prog: I have updated the doc file. Please review, thanks! 

Kindly note that I have provided the similarities and differences between the `fields` and `table` commands based on what it is currently implemented for the `fields` command. However, once the new features are implemented for the `fields` command, we will update this doc accordingly.

yuancu (approved):
Looks good to me. Thanks for your contribution :)

core/src/main/java/org/opensearch/sql/ast/tree/Table.java
```diff
@@ -0,0 +1,91 @@
...
+ * AST node representing a Table operation in the query plan.
+ *
+ * <p>This class represents a table operation in the Abstract Syntax Tree (AST). It contains a list
+ * of expressions to project (select) from the table and can have a child node representing the
+ * source of the data.
+ */
+@ToString
+@Getter
+@EqualsAndHashCode(callSuper = false)
+public class Table extends UnresolvedPlan {
```
LantaoJin, line 26: why not reuse `org.opensearch.sql.ast.tree.Project`? I didn't any difference between `table` and `project` in AST level. I'd like keep as simple as possible in AST.

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,264 @@
...
+
+
+Usage
+============
+The ``table`` command processes the entire result set and filters it to include only the specified fields. Duplicate field names in the field list are automatically de-duplicated.
+
+
+Field Renaming
+============
+The ``table`` command returns fields with their original names and does not support field renaming within the command itself. To rename fields, use the ``rename`` command before ``table``.
```
LantaoJin, line 49: > returns fields with their original names and does not support field renaming within the command itself.

what this means? I didn't get the point.

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,264 @@
...
++------------------+------------------------+------------------------+
+| **Deduplication**| Automatic              | Not applicable         |
++------------------+------------------------+------------------------+
+| **Default Behavior** | Always keeps specified fields | Keeps fields (+ is default) |
++------------------+------------------------+------------------------+
+
+**When to use table:**
+
+* When you want to control field order in results
+* When you only need to keep specific fields (no removal needed)
```
LantaoJin, line 85: Call out the `table` support wildcard.
I think the secondary core diff is `table` command is a truncate-able command, it means it wouldn't trigger PIT search or a limited PIT search.

LantaoJin:
Before diving into the code, I think [this](https://github.com/opensearch-project/sql/pull/3885#discussion_r2238278740) is needed to be clarified.

In SPL,  number of results of `table` returned based on settings `truncate_report`(false by default) and `max_count`(10000 by default). If the `truncate_report` set to true, it means the number of results is no larger than `max_count`. In another words, the PIT search won't be trigger or a limited number of PIT search will be trigger.
cc @penghuo and @brijos

docs/user/ppl/cmd/table.rst
aalva500-prog: @LantaoJin We have plans to support wildcard in `fields` command also.

docs/user/ppl/cmd/table.rst
aalva500-prog: @LantaoJin This statement is explaining a limitation of the proposed `table` command. Here's what it means:

- When you use `table`, it displays data using the original field names from your data source
- You cannot rename fields directly within the table command itself. For example, you can't do:
```
| table original_field_name AS new_name, another_field AS renamed_field
```
- You need to use the `rename` command before the `table` command:
```
| rename original_field_name AS new_name, another_field AS renamed_field
| table new_name, renamed_field
```

The key point is that field renaming and table are separate operations/commands that must be done in sequence, not combined into a single command. I hope that makes sense!

LantaoJin: It's confused me that the `fields` command does not support field renaming within the command itself either.

docs/user/ppl/cmd/table.rst
LantaoJin: > @LantaoJin We have plans to support wildcard in `fields` command also.

If that, seems we can implement the `table` command as "alias" of `fields`. No need to separate the implementation code.

aalva500-prog: @LantaoJin please refer to this comment: https://github.com/opensearch-project/sql/issues/3877#issuecomment-3134330809

penghuo: Although the table and fields commands appear similar, they have different syntax and behavior. The table command should be treated as a separate command.

aalva500-prog:
@LantaoJin @yuancu Our team decided not to implement the `table` command and just allow users to use either `table` or `fields` keyword, as if they were running the `fields` command as usual. The plan is to enrich the existing `fields` command with the SPL features. With this new approach, the following queries will be equivalent and return the same output:

```
source = "accounts" | fields account_number, first_name

source = "accounts" | table account_number, first_name

source = "accounts" | fields + account_number, first_name

source = "accounts" | table + account_number, first_name
```

penghuo:
> @LantaoJin @yuancu Our team decided not to implement the `table` command and just allow users to use either `table` or `fields` keyword, as if they were running the `fields` command as usual. The plan is to enrich the existing `fields` command with the SPL features. With this new approach, the following queries will be equivalent and return the same output:
> 
> ```
> source = "accounts" | fields account_number, first_name
> 
> source = "accounts" | table account_number, first_name
> 
> source = "accounts" | fields + account_number, first_name
> 
> source = "accounts" | table + account_number, first_name
> ```

Table will be a seperate command, it is alias of Field command, right?

LantaoJin:
`fields` and `table` are different in SPL, `fields` is a streaming command, `table` is non-streaming command. Current the PPL doesn't provide streaming mode, so the functionality of `table` has no difference with `fields`. +1 `table` will be an alias of `field` command.

PR #3886, opened 2025-07-15
Update the maven snapshot publish endpoint and credential

RyanL1997:
### Description
Update the maven snapshot publish endpoint and credential

Previous description from #3806 :
> Update the Maven Snapshots publish URL in accordance with the recent Sonatype migration.
[central.sonatype.org/publish/publish-portal-snapshots](https://central.sonatype.org/publish/publish-portal-snapshots/)
We have stored the onepassword token in this repo secrets and new credentials for Sonatypes username & password have been stored in onepassword. These credentials will be exported as env variables which used by maven publish.

### Related Issues
* Relate https://github.com/opensearch-project/opensearch-build/issues/5551
* Resolves #3856
* Resolves the conflict of #3806

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3892, opened 2025-07-17
[Backport 2.19] Update the maven snapshot publish endpoint and credential

RyanL1997:
### Description
Update the maven snapshot publish endpoint and credential

### Related Issues
* Backport #3886 

### Check List

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -18,7 +18,7 @@ jobs:
         id: lychee
         uses: lycheeverse/lychee-action@master
         with:
-          args: --accept=200,403,429,999  "./**/*.html" "./**/*.md" "./**/*.txt" --exclude "https://aws.oss.sonatype.*" "http://localhost*" "https://localhost" "https://odfe-node1:9200/" "https://community.tableau.com/docs/DOC-17978" ".*family.zzz" "https://pypi.python.org/pypi/opensearchsql/" "opensearch*" ".*@amazon.com" ".*email.com" "git@github.com" "http://timestamp.verisign.com/scripts/timstamp.dll" ".*/PowerBIConnector/bin/Release"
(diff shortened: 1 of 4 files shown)


RyanL1997:
seems like the security plugin didnt have snapshot for `2.19.3` yet: https://central.sonatype.com/artifact/org.opensearch.plugin/opensearch-security/2.19.2.0/dependents

LantaoJin:
why do we backport it to 2.19 rather than 2.x?

RyanL1997:
> why do we backport it to 2.19 rather than 2.x?

Since we have already moved to 3.x era for major version release, I think the 2.19 should be the last minor version we support for future 2.x patch releases. Similar thing was also applied to 1.3 branch when we first released 2.0 - when we do the backport of 1.x changes we directly backport it to 1.3. cc @LantaoJin

LantaoJin:
> > why do we backport it to 2.19 rather than 2.x?
> 
> Since we have already moved to 3.x era for major version release, I think the 2.19 should be the last minor version we support for future 2.x patch releases. Similar thing was also applied to 1.3 branch when we first released 2.0 - when we do the backport of 1.x changes we directly backport it to 1.3. cc @LantaoJin

Oh, sounds we had many 2.x backportings which should be backported to 2.19 instead. For example, https://github.com/opensearch-project/sql/compare/2.19...2.x

RyanL1997:
> Oh, sounds we had many 2.x backportings which should be backported to 2.19 instead. For example, https://github.com/opensearch-project/sql/compare/2.19...2.x

@LantaoJin If there is a fix for security vulnerability/maintenance (like this PR) we can backport it to 2.19 directly. New features should not be backport to 2.19 anymore.

penghuo (approved):

LantaoJin:
@RyanL1997 please manually backport to 2.19-dev if you think it was still required. The 2.19-dev branch is only for AOS internal backporting

Issue #3761, opened 2025-06-10
[FEATURE] Support function parameter coercion with Calcite


yuancu:
**Is your feature request related to a problem?**

Before introducing Calcite, when input parameter types of a function don't match with its expected types, it will try to coerce the parameter type with the expected types.

However, such mechanism isn't in place in current implementations with Calcite. This results in multiple problems. For example:
- when input is a literal and the expected type is a date, it will not automatically coerce the literal to a date. Instead, it requires that the function supports string type as an input. ([CalciteDateTimeFunctionIT.testUnixTimestampWithTimestampString](https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDateTimeFunctionIT.java))
- when the input is long and the expected type is only integer (`SqlTypeName.INTEGER`), it can't resolve the function call to a known implementation. 

**What solution would you like?**
Implement function type parameter coercion with Calcite. 

**What alternatives have you considered?**
N/A

**Do you have any additional context?**

- Type coercion rule before introducing Calcite: https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/data/type/WideningTypeRule.java
- This issue is also related: https://github.com/opensearch-project/sql/issues/3728

PR #3914, opened 2025-07-23
Support function argument coercion with Calcite

yuancu:
### Description

Support coercing argument types to expected ones with Calcite. This eliminates the necessities to implement multiple overrides for a function where the arguments can actually be implicitly cast.

This PR implements the alternative solution in RFC #3761 

### Implementation notes

I use `ExprType` as a common ground for type comparison and coercion. In comparison, the validation phase of Calcite uses `SqlTypeFamily` and `SqlTypeName`. However, as we have UDTs that are not covered by types in Calcite, using `ExprType` is more practical in our case.

### Related Issues
#3761

### Check List


yuancu:
Need #3919 to function properly.

yuancu:
Comparing IP is not pushed down as a range query (but as script). I am looking into this and will raise another issue and PR to fix it.

core/src/main/java/org/opensearch/sql/calcite/utils/PPLOperandTypes.java
```diff
@@ -27,40 +27,121 @@ private PPLOperandTypes() {}
       UDFOperandMetadata.wrap(
           (CompositeOperandTypeChecker) OperandTypes.INTEGER.or(OperandTypes.family()));
   public static final UDFOperandMetadata STRING =
-      UDFOperandMetadata.wrap((FamilyOperandTypeChecker) OperandTypes.STRING);
+      UDFOperandMetadata.wrap((FamilyOperandTypeChecker) OperandTypes.CHARACTER);
```
LantaoJin, line 30: How about change the `UDFOperandMetadata STRING` to `UDFOperandMetadata CHARACTER` since CHARACTER is the primary family. ditto for the rests.

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/FromUnixTimeFunction.java
```diff
@@ -56,10 +54,7 @@ public SqlReturnTypeInference getReturnTypeInference() {
 
   @Override
   public UDFOperandMetadata getOperandMetadata() {
-    return UDFOperandMetadata.wrap(
-        (CompositeOperandTypeChecker)
-            OperandTypes.NUMERIC.or(
-                OperandTypes.family(SqlTypeFamily.NUMERIC, SqlTypeFamily.STRING)));
+    return PPLOperandTypes.NUMERIC_OPTIONAL_STRING;
```
LantaoJin, line 57: IMO, it's more readable to keep the UDFOperandMetadata in the function definition:
```
  @Override
    return UDFOperandMetadata.wrap(
        (CompositeOperandTypeChecker)
            OperandTypes.NUMERIC.or(
                OperandTypes.family(SqlTypeFamily.NUMERIC, SqlTypeFamily.STRING)));
  }
```
```
  @Override
    return PPLOperandTypes.NUMERIC_OPTIONAL_STRING;
  }
```

Which one would be better?

core/src/main/java/org/opensearch/sql/calcite/utils/PPLOperandTypes.java
yuancu: I did not rename it since I wanted to make the distinction between PPL types and its underlying implementation with Calcite -- string is a type in PPL, while character (char & varchar) is its implementation with Calcite.

It also makes sense to rename `PPLOperandTypes.STRING` to `PPLOperandTypes.CHARACTER` though. This aligns the naming with its implementation. Which do you think is more reasonable?

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/FromUnixTimeFunction.java
yuancu: The second one complies with Calcite's convention when naming operand types in [OperandTypes.java](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java)

Besides, another motivation for making this update was to gather all defined operand types in a central repository so that when I have to make updates to them, it's less likely that I carelessly ignored one of them -- this happened this time when I updated the operand types.

LantaoJin: I don't think setting the operand type in each function class would make your updating this time any harder. And IMO , gathering all operand types in a central class make the function definition unreadable, except they are always reused and duplicated. I prefer to define acceptable operand types in each function even there is a little bit duplication.

yuancu: I agree that having the definition expanded will make what's under the hood more clear. Yet it may not necessarily be a detail that has to be reiterated at each function definition.

IMO, the latter is more readable. And this is exactly how Calcite defines operand type checkers. E.g.
```
public static final SqlSingleOperandTypeChecker NUMERIC_OPTIONAL_INTEGER =
      NUMERIC.or(NUMERIC_INTEGER);
```

The pattern is quite easy to grasp. The added benefit is that, when you want to create a new operand type checker, you can just type the operand types with this pattern -- the corresponding type checker will automatically jump out if they are already defined and used by other functions. On the other hand, without this, you will risking redefining what's already out there. What's worse, you may update one and neglect another even if two functions are supposed to have the same operand types. It occurred this time when I update function definitions by removing parameter types that are ought to be coerced.

With this centralized class, I also don't have to make the following statement even longer:
```
  public static final SqlOperator TIME_TO_SEC =
      adaptExprMethodToUDF(
              DateTimeFunctions.class,
              "exprTimeToSec",
              ReturnTypes.BIGINT_FORCE_NULLABLE,
              NullPolicy.ARG0,
              PPLOperandTypes.TIME_OR_TIMESTAMP_OR_STRING)
```

Moreover, previously, we already have some UDFs with their operand type checkers defined in `PPLOperandTypes` if the checker is shared, while those are not shared are defined directly in `getOperandMetadata`. This left a divided representation. It does no harm to unify and collect them into `PPLOperandTypes`.

LantaoJin: Okey, I see. The `PPLOperandTypes.xx` is not only used in `getOperandMetadata()` but also used in definition of custom `SqlOperator`.

yuancu:
Yes, all behaviors remain the same. If a function is declared to accept string, it will not coerce the string to a certain type. This is because it is ought to give certain response (like `null`) for malformatted input.

E.g. `timestamp(string/timestamp/date/time)` was not modified do `timestamp(timestamp/date/time)`.

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
```diff
@@ -53,6 +58,16 @@ public interface PPLTypeChecker {
...
 
+  /**
+   * Get a list of all possible parameter type combinations for the function.
+   *
+   * <p>This method is used to generate the allowed signatures for the function based on the
+   * parameter types.
+   *
+   * @return a list of lists, where each inner list represents an allowed parameter type combination
+   */
+  List<List<ExprType>> getParameterTypes();
```
qianheng-aws, line 69: [non-blocking]I would prefer using Type system defined in Calcite like `RelDataType` or SqlType.

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
```diff
@@ -0,0 +1,139 @@
...
+
+    // Iterate pairwise through the arguments and find the widest type
+    for (int i = 1; i < arguments.size(); i++) {
+      var type = OpenSearchTypeFactory.convertRelDataTypeToExprType(arguments.get(i).getType());
+      try {
+        if (areDateAndTime(widestType, type)) {
+          // If one is date and the other is time, we consider timestamp as the widest type
+          widestType = ExprCoreType.TIMESTAMP;
+        } else {
+          widestType = WideningTypeRule.max(widestType, type);
```
qianheng-aws, line 125: [non-blocking]Calcite has similar operation by using  `RelDataTypeFactory::leastRestrictive`. If it's functional equivalent as this, I would prefer use that one.

We may deprecate ExprValue and ExprType in the future. But for now, it's OK to keep using this one.

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
yuancu: I used `ExprType` since it has a full coverage of PPL types. I will also have to convert `RelDataType` to either `SqlTypeName` or `ExprType` for comparison purpose.

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
yuancu: Good idea, we could override `leastRestrictive` in favor of `RelDataType` in the future.

PR #3915, opened 2025-07-23
Convert like function call to wildcard query for Calcite filter pushdown

### LINKED ISSUE #3428: [BUG] calcite like function didn't work as expected

songkant-aws:
### Description
Enhancement: 

In V2 pushdown logic, like function is translated to OpenSearch DSL wildcard query but it's converted to UDF script in Calcite filter pushdown. This change makes sure Calcite has parity functionality.

Text field wildcard query has inconsistent behavior with SQL like function. So this PR also proposes to not pushdown like query for text field. Instead, we should only support pushdown like query for keyword field.

### Related Issues
Resolves #3428 

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteLikeQueryIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteLikeQueryIT.java
@@ -6,6 +6,7 @@
 package org.opensearch.sql.calcite.remote;
 
 import java.io.IOException;
+import org.junit.Assume;
 import org.junit.Test;
 import org.opensearch.sql.ppl.LikeQueryIT;
 
@@ -20,9 +21,7 @@ public class CalciteLikeQueryIT extends LikeQueryIT {
   @Override
   @Test
   public void test_convert_field_text_to_keyword() throws IOException {
-    // This test case is not applied to no pushdown case
-    if (isPushdownEnabled()) {
-      super.test_convert_field_text_to_keyword();
-    }
+    Assume.assumeTrue("Pushdown is not enabled, skipping this test.", isPushdownEnabled());
+    super.test_convert_field_text_to_keyword();
   }
 }


integ-test/src/test/java/org/opensearch/sql/ppl/LikeQueryIT.java
```diff
@@ -62,17 +62,17 @@ public void test_like_in_where_with_escaped_underscore() throws IOException {
   @Test
   public void test_like_on_text_field_with_one_word() throws IOException {
     String query =
-        "source=" + TEST_INDEX_WILDCARD + " | WHERE Like(TextBody, 'test*') | fields TextBody";
+        "source=" + TEST_INDEX_WILDCARD + " | WHERE Like(TextBody, 'test%') | fields TextBody";
     JSONObject result = executeQuery(query);
-    assertEquals(9, result.getInt("total"));
+    assertEquals(8, result.getInt("total"));
```
LantaoJin, line 67: Question:
With the patch, both in v2 and v3, predicate `WHERE Like(TextKeywordBody, 'test%')` will trigger wildcard query pushdown but predicate `WHERE Like(TextBody, 'test%')` won't trigger any pushdown. right?

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
```diff
@@ -435,6 +435,15 @@ public void testMultiFieldsRelevanceQueryFunctionExplain() throws IOException {
...
   }
 
+  @Test
+  public void testLikeFunctionExplain() throws IOException {
+    String expected = loadExpectedPlan("explain_like_function.json");
+    assertJsonEqualsIgnoreId(
+        expected,
+        explainQueryToString(
+            "source=opensearch-sql_test_index_account | where like(email, '%gmail%')"));
+  }
```
LantaoJin: Is the `email` a keyword or not? please add two tests for 
- keyword (e.g. firstname)
- text (e.g. address)

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteLikeQueryIT.java
```diff
@@ -6,66 +6,23 @@
...
-    super.test_like_on_text_field_with_greater_than_one_word();
-  }
-
-  @Override
-  @Test
-  @Ignore("https://github.com/opensearch-project/sql/issues/3428")
   public void test_convert_field_text_to_keyword() throws IOException {
-    super.test_convert_field_text_to_keyword();
+    // This test case is not applied to no pushdown case
+    if (isPushdownEnabled()) {
```
qianheng-aws: [Minor] Change to use `Assume` to skip the test if `isPushdownEnabled` is false.

```
Assume.assumeTrue("This test is only for push down enabled", isPushdownEnabled());
```

songkant-aws:
@penghuo Need to confirm some behavior.
SQL like function uses '_' to match single character and '%' to match zero or more characters. 
DSL wildcard query uses '?' to match single character and '*' to match zero or more characters.

When like function is pushed down as DSL wildcard query, V2 converts '_' and '%' to '?' and '\*' respectively but it doesn't handle '?' and '\*'. So when user enters "Amber*", he may just expect to match exactly "Amber*" term but DSL will do wildcard match to return "Amber*", "Amber A", "Amber B". 

Question: Since this seems to be incorrect but also a breaking change, should we correct this behavior? Should we totally align with SQL like function?

LantaoJin:
> @penghuo Need to confirm some behavior. SQL like function uses '_' to match single character and '%' to match zero or more characters. DSL wildcard query uses '?' to match single character and '*' to match zero or more characters.
> 
> When like function is pushed down as DSL wildcard query, V2 converts '_' and '%' to '?' and '*' respectively but it doesn't handle '?' and '*'. So when user enters "Amber*", he may just expect to match exactly "Amber*" term but DSL will do wildcard match to return "Amber*", "Amber A", "Amber B".
> 
> Question: Since this seems to be incorrect but also a breaking change, should we correct this behavior? Should we totally align with SQL like function?

I think we should fix this bug. Ref [here](https://docs.splunk.com/Documentation/Splunk/9.4.2/Search/Wildcards) and other databases SQL functions, the `_` and `%` are wildcard characters only for LIKE function. But `*` can be wildcard in other commands/functions, such **match** or **query_string**.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1137,10 +1148,24 @@ public QueryExpression notExists() {
...
+      String keywordField = OpenSearchTextType.toKeywordSubField(fieldName, this.rel.getExprType());
+      boolean isKeywordField = keywordField != null;
+      if (isKeywordField) {
+        builder =
+            wildcardQuery(
+                    keywordField, StringUtils.convertSqlWildcardToLucene(literal.stringValue()))
+                .caseInsensitive(true);
+        return this;
+      }
+      throw new UnsupportedOperationException("Like query is not supported for text field");
```
penghuo, line 1168: Text field can also be push down as script, Can we track it as enhancement issue?

And add a Notes to explain current limitation on Text field support in [LIKE](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/string.rst#like).

```
###
PUT {{baseUrl}}/demo-index

{
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 0
  },
  "mappings": {
    "properties": {
      "region": { "type": "text" },
      "sales": { "type": "double" }
    }
  }
}

###
POST {{baseUrl}}/demo-index/_bulk

{ "index": {} }
{ "region": "us-east", "sales": 100.12 }
{ "index": {} }
{ "region": "us-east", "sales": 200.34 }
{ "index": {} }
{ "region": "us-west", "sales": 300.65 }
{ "index": {} }
{ "region": "us-west", "sales": 400.78 }

###
POST {{baseUrl}}/demo-index/_search

{
  "derived": {
    "region_runtime": {
      "type": "keyword",
      "script": {
        "source": "emit(params._source['region'])"
      }
    }
  },
  "query": {
    "term": {
      "region_runtime": {
        "value": "us-east"
      }
    }
  }
}
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LikeQuery.java
```diff
@@ -29,4 +33,32 @@ protected WildcardQueryBuilder createBuilder(String field, String query) {
...
+   * Verify if the function supports like/wildcard query. Prefer to run wildcard query for keyword
+   * type field. For text type field, it doesn't support cross term match because OpenSearch
+   * internally break text to multiple terms and apply wildcard matching one by one, which is not
+   * same behavior with regular like function without pushdown.
+   *
+   * @param func function Input function expression
+   * @return boolean
+   */
+  @Override
+  public boolean canSupport(FunctionExpression func) {
```
penghuo, line 47: Is it for solving existing bug?

integ-test/src/test/java/org/opensearch/sql/ppl/LikeQueryIT.java
songkant-aws: Yes

integ-test/src/test/java/org/opensearch/sql/ppl/ExplainIT.java
songkant-aws: Done

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteLikeQueryIT.java
songkant-aws: Done

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
songkant-aws: Added limitation to LIKE function doc. 

I tried pushdown like function script for text field. However, getting ScriptDocValues for text field throws exception to not recommend to do it.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LikeQuery.java
songkant-aws: Yes, DSL wildcard matching for text field has different behavior with SQL like function because of tokenization.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
penghuo: Yes, text data does not have doc values, we should use source, e.g. `params._source['region']`. Let's track it in issue, not high priority.

penghuo (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
songkant-aws: Makes sense. Created a tracking issue: https://github.com/opensearch-project/sql/issues/3950. Hopefully, we can find a way to read script values from source.

qianheng-aws (approved):

PR #3921, opened 2025-07-25
Fix create PIT permissions issue

vamsimanohar:
### Description
Currently there is a bug where PPL is asking for search permission across all indices even though the user has permissions to make a PPL query on a particular index.

Issue is more detailed here: https://github.com/opensearch-project/security/issues/5508

Also, fixed few other issues with wrong error status codes in case of authorization errors.

Added more PPL permissions IT which covers the above bug.

### Root cause
* Transport PIT search calls require index information for correct security authorization.
* Rest PIT search calls automatically handles this in Rest layer by decoding PIT information into indices before transport layer.

### Related Issues

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/client/OpenSearchRestClient.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/client/OpenSearchRestClient.java
@@ -115,6 +115,11 @@ public class OpenSearchRestClient implements OpenSearchClient {
     return request.search(
         req -> {
           try {
+            // For RestClient with PIT: remove indices to avoid validation error
+            // "indices cannot be used with point in time"
+            if (req.source() != null && req.source().pointInTimeBuilder() != null) {
+              req = new SearchRequest().source(req.source());
+            }
             return client.search(req, RequestOptions.DEFAULT);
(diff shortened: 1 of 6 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/client/OpenSearchRestClient.java
```diff
@@ -115,6 +115,11 @@ public OpenSearchResponse search(OpenSearchRequest request) {
     return request.search(
         req -> {
           try {
+            // For RestClient with PIT: remove indices to avoid validation error
+            // "indices cannot be used with point in time"
```
ps48, line 119: Nit. May be we can link the original sec design here: https://github.com/opensearch-project/security/issues/2087

ps48 (approved):

Swiddis (approved):
It seems like the reason we weren't catching these already is because we're relying heavily on catch-rethrow as `RuntimeException`. We should revisit the error strategy at the top level to see if we can start being more specific with our error handling overall, I'd like to see a future where we have a plugin-specific error code index (like how many shell tools define all their exit codes).

penghuo (approved):

PR #3930, opened 2025-07-28
Skip script encoding when run explain with  'extended'

### LINKED ISSUE #3929: [FEATURE] More readable query explain for script pushdown
When the script pushdown is enabled, the script content will be serialized with Base64 encoding which is unreadable.

For explain command, the script doesn't need to serialize due to script execution is not required.

LantaoJin:
### Description
skip script encoding when run explain with 'extended':
- /_plugins/_ppl/_explain?format=extended
- `explain extended source=t | ...`

The RexNode expression `SCRIPT->=(-($1, 2), 30)` in explain command:
```
{
    "op": {
        "name": "=",
        "kind": "EQUALS",
        "syntax": "BINARY"
    },
    "operands": [{
        "op": {
            "name": "-",
            "kind": "MINUS",
            "syntax": "BINARY"
        },
        "operands": [{
            "input": 2,
            "name": "$2"
        }, {
            "literal": 2,
            "type": {
                "type": "INTEGER",
                "nullable": false
            }
        }],
        "type": {
            "type": "BIGINT",
            "nullable": true
        }
    }, {
        "literal": 30,
        "type": {
            "type": "INTEGER",
            "nullable": false
        }
    }]
}
```
### Related Issues
Resolves #3929 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
```diff
@@ -156,6 +156,7 @@ public void explain(
       ResponseListener<ExplainResponse> listener) {
     client.schedule(
         () -> {
+          CalcitePlanContext.isExplain.set(true);
```
qianheng-aws: Why not put this line inner try-catch? I'm worrying if we might add more code between this line and try-catch in the future, it will cause exception and lead to skipping `isExplain.remove()`, which will then pollute the thread pool

qianheng-aws:
The main concern from me is that, after this change, the DSL in explain output is no longer able to be executed. I think maybe we should skip decoding the script for execution as well.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
LantaoJin: done

LantaoJin:
> The main concern from me is that, after this change, the DSL in explain output is no longer able to be executed. I think maybe we should skip decoding the script for execution as well.

The entire json contains unreadable objects (Java Map of schema), the current the explain output in my PR only display the EXPR part. So maybe we should keep the decode string in explain and add a parameter to skip decode process in explain.

LantaoJin:
Refactor the code: skip script encoding when run the explain command in 'extended' mode. cc @qianheng-aws

LantaoJin:
Ping @qianheng-aws @penghuo

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
```diff
@@ -41,6 +41,8 @@
 public class RestPPLQueryAction extends BaseRestHandler {
   public static final String QUERY_API_ENDPOINT = "/_plugins/_ppl";
   public static final String EXPLAIN_API_ENDPOINT = "/_plugins/_ppl/_explain";
+  public static final String EXTENDED_EXPLAIN_API_ENDPOINT =
```
penghuo, line 44: if it is only for testing? move to test class

qianheng-aws:
Shall we add description in `explain.rst` to clarify the limitation of `explain extended` that its output DSL is not executable?

LantaoJin:
> Shall we add description in `explain.rst` to clarify the limitation of `explain extended` that its output DSL is not executable?

Seems we never mentioned the script in explain output can be executed.

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
```diff
@@ -36,6 +36,8 @@ public class CalcitePlanContext {
   public final QueryType queryType;
   public final Integer querySizeLimit;
 
+  public static final ThreadLocal<Boolean> skipEncoding = ThreadLocal.withInitial(() -> false);
```
dai-chen, line 39: Not sure if possible to avoid this global variable. If this is necessary, could you add clear javadoc for its purpose?

dai-chen (approved):
So this is only for extended mode? If we want to troubleshoot a query, we can still get executable DSL by explain right?

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
LantaoJin: Threadlocal variable should be static to avoid binding to one context instance (all context instances should share the same thread local variable). I will add javadoc in other PRs.

Issue #3387, opened 2025-03-06
[FEATURE] Calcite Engine Framework: pushdown whole project instead of reference


qianheng-aws:
**Is your feature request related to a problem?**
Calcite Engine Framework: pushdown whole project instead of reference

**What solution would you like?**
Using script push down

LantaoJin:
I am going to close this issue since the `ScriptFieldsPhase` is a `FetchSubPhase`. It means push down a script field would have less performance improvement. Instead, it requires additional script compiling.

penghuo:
I  believe ScriptFieldsPhase executes on each shard, not the coordination node.

1. Document-level data access requirement, Script fields support accessing document-specific data that only exists at the shard level: 
    - DocValues access: doc['field_name'].value requires direct access to Lucene's LeafReader
    - Source field access: Scripts can read from _source using params._source.field_name
    - Index reader dependency: Document data structures like DocValues are only available through the shard's IndexReader
2. Shard-level execution in fetch phase, The execution flow confirms shard-level processing:
    - FetchSearchPhase.executeFetch(): Sends fetch requests to individual shards (FetchSearchPhase.java)
    - Per-shard processing: Each shard executes its complete fetch pipeline, including ScriptFieldsPhase
    - Result collection: Each shard returns a FetchSearchResult containing computed script field values
    - Coordination-level merging: SearchPhaseController.merge() combines FetchSearchResult objects from all shards

@gaobinlong Could u help double confirm on this?

@LantaoJin What use case can leverage script_filed pushdown? e.g. `source=index | parse message https=(?<status>) | fields status`

gaobinlong:
> @gaobinlong Could u help double confirm on this?

Yes, by checking the code, ScriptFieldsPhase executes on each shard, FetchSearchPhase executes on coordinator node.

songkant-aws:
I will pick this up because we found OpenSearch DerivedFieldScript could address the concern of not evaluating project expressions on whole data rows. With whole project script pushdown, it will also unblock sort expression pushdown.

songkant-aws:
A concern to implement project pushdown with either ScriptFields or DerivedFieldScript. They don't support sorting according to the official document. 

ScriptField is kind of postprocessing based on input fields after each shard search phase, which means it doesn't support filtering.
DerivedFieldScript is a runtime field that is generated at search phase. Although it can support filtering, scoring and sorting is not supported yet.

songkant-aws:
[...]

PR #3940, opened 2025-07-30
Append limit operator for QUEERY_SIZE_LIMIT

qianheng-aws:
### Description
Since we don't have plan to implement [eval push down](https://github.com/opensearch-project/sql/issues/3387), we still need to append LIMIT operator explicitly for QUERY_SIZE_LIMIT, in order to cover the case of `Project - Scan`. See the source comment: https://github.com/opensearch-project/sql/pull/3880#discussion_r2209634187

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3879

### Check List


core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -250,8 +250,14 @@ public PhysicalPlan plan(LogicalPlan plan) {
...
 
-  public RelNode optimize(RelNode plan) {
-    return planner.customOptimize(plan);
+  /**
+   * Try to optimize the plan by appending a limit operator for QUERY_SIZE_LIMIT Don't add for
+   * `EXPLAIN` to avoid changing its output plan.
+   */
+  public RelNode optimize(RelNode plan, CalcitePlanContext context, boolean isExplain) {
+    if (isExplain) return plan;
+    context.relBuilder.limit(0, context.querySizeLimit);
```
LantaoJin: can you add a case in `ExplainIT`?

qianheng-aws: It won't change the output plan of explain.

penghuo: Add an IT to verify querySizeLimit pushdown.
Q, should we revert code of https://github.com/opensearch-project/sql/pull/3880?

LantaoJin: We can reuse [LogicalSystemLimit](https://github.com/opensearch-project/sql/pull/3749) in explain. The physical plan is no diff. Then we can verify the querySizeLimit pushdown in explain IT.

qianheng-aws: - Discussed with @LantaoJin  offline, I will add a new `LogicalSystemLimit` operator to represent the limit operator added by our system. It will be used in many other places  as well where we need to add a limitation. So it will contain a field of  `Type` to distinguish them.

- I will manually revert few lines related, although they are compatible. That https://github.com/opensearch-project/sql/pull/3880 also contains a bug fix related to limit push down.

integ-test/src/test/resources/expectedOutput/calcite/explain_filter_agg_push.json
```diff
@@ -1,6 +1,6 @@
...
+    "logical": "LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n  LogicalProject(avg_age=[$2], state=[$0], city=[$1])\n    LogicalAggregate(group=[{0, 1}], avg_age=[AVG($2)])\n      LogicalProject(state=[$7], city=[$5], age=[$8])\n        LogicalFilter(condition=[>($8, 30)])\n          CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n",
```
penghuo, line 3: we should supportLimit pushdown through aggregation. track with https://github.com/opensearch-project/sql/issues/3961

penghuo (approved):

PR #3941, opened 2025-07-30
Add auto-increment id to dynamic function signature

### LINKED ISSUE #3938: [BUG] Multiple calls to `SYSDATE` returns the same value
**What is the bug?**

[`SYSDATE`](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/datetime.rst#sysdate) returns the timestamp at which it executes. It is useful to track operation execution time. For example, `eval t1 = sysdate(5) | sort date | eval t2 = sysdate(6) | fields t1, t2` can give a hint on how long does it take to execute sort.

However, I found that multiple calls to `SYSDATE` with the same parameter within a query will return the same value.

E.g. `sd1` and `sd2` are different in this query: `source=dates | eval sd1 = sysdate(5) | sort date | eval sd2 = sysdate(6) | fields sd1, sd2`

| sd1 | sd2  |
| ----| ---- |
| 2025-07-22 06:16:33.13648 | 2025-07-22 06:16:33.136559 |

However, this query gives the same result: `source=dates | eval sd1 = sysdate(6) | sort date | eval sd2 = sysdate(6) | fields sd1, sd2`

| sd1 | sd2  |
| ----| ---- |
| 2025-07-22 06:19:07.006268 | 2025-07-22 06:19:07.006268 |

The physical plan gives a hint to the problem:
```
EnumerableCalc(expr#0=[{inputs}], expr#1=[6], expr#2=[SYSDATE($t1)], sd1=[$t2], sd2=[$t2])
  CalciteEnumerableIndexScan(table=[[OpenSearch, dates]], PushDownContext=[[PROJECT->[date], SORT->[{
  "date" : {
    "order" : "asc",
    "missing" : "_first"
  }
}]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"timeout":"1m","_source":{"includes":["date"],"excludes":[]},"sort":[{"date":{"order":"asc","missing":"_first"}}]}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```

Both `sd1` and `sd2` point to `$t2`.

However, it is worth mention that `sysdate` gives different results at different rows.

`source=dates | eval sd1 = sysdate(6) | sort date | eval sd2 = sysdate(6) | fields sd1, sd2`

| sd1 | sd2  |
| ----| ---- |
| 2025-07-30 02:30:54.107595 | 2025-07-30 02:30:54.107595 |
| 2025-07-30 02:30:54.107692 | 2025-07-30 02:30:54.107692 |

curl --location --globoff '{{opensearch}}/_plugins/_ppl' \
--data '{
  "query" : "source=dates | eval sd1 = sysdate(5) | sort date | eval sd2 = sysdate(6) | fields sd1, sd2"
}'

It returns 

 - OS: MacOS
 - Version 3.1.0
 - Plugins

_Originally posted by @yuancu in https://github.com/opensearch-project/sql/pull/3854#discussion_r2221351270_

LantaoJin:
### Description
`SYSDATE` is a dynamic function which cannot be resolved correctly in rule `ProjectToCalcRule`:
`ProjectToCalcRule` created a RexProgram in LogicalCalc but the `exprMap` to build RexProgram uses digest of RexCall as map key. Dynamic function such as `sysdate(6)` cannot be identified to diff expressions since their digests are same in PPL:
```
source=dates | eval sd1 = sysdate(6) | sort date | eval sd2 = sysdate(6) | fields sd1, sd2
```
Check the issue details in #3938.

The optimal solution should be fixed in Calcite side that `ProjectToCalcRule` identifies dynamic function with a special key instead of using digest of RexCall directly. But this PR workarounds the issue by adding an auto-increment id as one of operand of RexCall (function).

### Related Issues
Resolves #3938 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
@@ -27,6 +27,7 @@ import org.apache.calcite.sql.type.ReturnTypes;
 import org.apache.calcite.sql.type.SqlTypeFamily;
 import org.apache.calcite.sql.type.SqlTypeTransforms;
 import org.apache.calcite.sql.util.ReflectiveSqlOperatorTable;
+import org.apache.calcite.sql.validate.SqlMonotonicity;
 import org.apache.calcite.util.BuiltInMethod;
 import org.opensearch.sql.calcite.utils.PPLOperandTypes;
 import org.opensearch.sql.calcite.utils.PPLReturnTypes;
@@ -180,11 +181,15 @@ public class PPLBuiltinOperators extends ReflectiveSqlOperatorTable {
   public static final SqlOperator SECOND = new DatePartFunction(TimeUnit.SECOND).toUDF("SECOND");
   public static final SqlOperator MICROSECOND =
       new DatePartFunction(TimeUnit.MICROSECOND).toUDF("MICROSECOND");
-  public static final SqlOperator NOW = new CurrentFunction(ExprCoreType.TIMESTAMP).toUDF("NOW");
+  public static final SqlOperator NOW =
+      new CurrentFunction(ExprCoreType.TIMESTAMP)
+          .toUDF("NOW", false, false, SqlMonotonicity.INCREASING);
   public static final SqlOperator CURRENT_TIME =
-      new CurrentFunction(ExprCoreType.TIME).toUDF("CURRENT_TIME");
+      new CurrentFunction(ExprCoreType.TIME)
+          .toUDF("CURRENT_TIME", false, false, SqlMonotonicity.INCREASING);
   public static final SqlOperator CURRENT_DATE =
-      new CurrentFunction(ExprCoreType.DATE).toUDF("CURRENT_DATE");
+      new CurrentFunction(ExprCoreType.DATE)
+          .toUDF("CURRENT_DATE", false, false, SqlMonotonicity.INCREASING);
   public static final SqlOperator DATE_FORMAT =
       new FormatFunction(ExprCoreType.DATE).toUDF("DATE_FORMAT");
   public static final SqlOperator TIME_FORMAT =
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
```diff
@@ -357,11 +363,11 @@ public class PPLBuiltinOperators extends ReflectiveSqlOperatorTable {
   public static final SqlOperator MATCH_PHRASE_PREFIX =
       RELEVANCE_QUERY_FUNCTION_INSTANCE.toUDF("match_phrase_prefix");
   public static final SqlOperator SIMPLE_QUERY_STRING =
-      RELEVANCE_QUERY_FUNCTION_INSTANCE.toUDF("simple_query_string", false);
+      RELEVANCE_QUERY_FUNCTION_INSTANCE.toUDF("simple_query_string", false, false);
   public static final SqlOperator QUERY_STRING =
-      RELEVANCE_QUERY_FUNCTION_INSTANCE.toUDF("query_string", false);
+      RELEVANCE_QUERY_FUNCTION_INSTANCE.toUDF("query_string", false, false);
```
LantaoJin, line 368: @songkant-aws why these search relevant functions are `isDeterministic = false`?

songkant-aws: It's for resolving the over optimization issue in ReduceExpressionRule. 

Multi field relevance functions accept wildcard or multiple fields in the first argument string. i.e. "field*", "field1, field2". Calcite will optimize this as string literal instead of RexInputRef, which will require this UDF to have implementor. But this UDF definition is just for placeholder(shouldn't have implementor, aka non pushdown case), we can only support it when it's pushed down. To avoid rule optimization, we mark multi-field relevance function as nonDeterministic.

LantaoJin:
I will close this since the root cause is the sort was pushed down. Need to confirm that preventing the pushdown is a right behaviour.

PR #3959, opened 2025-07-31
Push down IP comparison as range query with Calcite

### LINKED ISSUE #3937: [BUG] IP comparison is not pushed down with Calcite
**What is the bug?**
Comparing IP is not pushed down as a range query but as script.

For this ppl: `source=weblogs | where host > '1.1.1.1' | fields host`
- With Calcite, comparing IP is pushed down as script:

```json
{
    "calcite": {
        "logical": "LogicalProject(host=[$0])\n  LogicalFilter(condition=[GREATER_IP($0, IP('1.1.1.1':VARCHAR))])\n    CalciteLogicalIndexScan(table=[[OpenSearch, weblogs]])\n",
        "physical": "CalciteEnumerableIndexScan(table=[[OpenSearch, weblogs]], PushDownContext=[[PROJECT->[host], SCRIPT->GREATER_IP($0, IP('1.1.1.1':VARCHAR))], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"query\":{\"script\":{\"script\":{\"source\":\"{\\\"langType\\\":\\\"calcite\\\",\\\"script\\\":\\\"rO0ABXNyABFqYXZhLnV0aWwuQ29sbFNlcleOq7Y6G6gRAwABSQADdGFneHAAAAADdwQAAAAGdAAHcm93VHlwZXQAensKICAiZmllbGRzIjogWwogICAgewogICAgICAidHlwZSI6ICJPVEhFUiIsCiAgICAgICJudWxsYWJsZSI6IHRydWUsCiAgICAgICJuYW1lIjogImhvc3QiCiAgICB9CiAgXSwKICAibnVsbGFibGUiOiBmYWxzZQp9dAAEZXhwcnQDeXsKICAib3AiOiB7CiAgICAibmFtZSI6ICJHUkVBVEVSX0lQIiwKICAgICJraW5kIjogIkdSRUFURVJfVEhBTiIsCiAgICAic3ludGF4IjogIkJJTkFSWSIKICB9LAogICJvcGVyYW5kcyI6IFsKICAgIHsKICAgICAgImlucHV0IjogMCwKICAgICAgIm5hbWUiOiAiJDAiCiAgICB9LAogICAgewogICAgICAib3AiOiB7CiAgICAgICAgIm5hbWUiOiAiSVAiLAogICAgICAgICJraW5kIjogIk9USEVSX0ZVTkNUSU9OIiwKICAgICAgICAic3ludGF4IjogIkZVTkNUSU9OIgogICAgICB9LAogICAgICAib3BlcmFuZHMiOiBbCiAgICAgICAgewogICAgICAgICAgImxpdGVyYWwiOiAiMS4xLjEuMSIsCiAgICAgICAgICAidHlwZSI6IHsKICAgICAgICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICAgICAgICJudWxsYWJsZSI6IGZhbHNlLAogICAgICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgICAgIH0KICAgICAgICB9CiAgICAgIF0sCiAgICAgICJjbGFzcyI6ICJvcmcub3BlbnNlYXJjaC5zcWwuZXhwcmVzc2lvbi5mdW5jdGlvbi5Vc2VyRGVmaW5lZEZ1bmN0aW9uQnVpbGRlciQxIiwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiT1RIRVIiLAogICAgICAgICJudWxsYWJsZSI6IHRydWUKICAgICAgfSwKICAgICAgImRldGVybWluaXN0aWMiOiB0cnVlLAogICAgICAiZHluYW1pYyI6IGZhbHNlCiAgICB9CiAgXSwKICAiY2xhc3MiOiAib3JnLm9wZW5zZWFyY2guc3FsLmV4cHJlc3Npb24uZnVuY3Rpb24uVXNlckRlZmluZWRGdW5jdGlvbkJ1aWxkZXIkMSIsCiAgInR5cGUiOiB7CiAgICAidHlwZSI6ICJCT09MRUFOIiwKICAgICJudWxsYWJsZSI6IHRydWUKICB9LAogICJkZXRlcm1pbmlzdGljIjogdHJ1ZSwKICAiZHluYW1pYyI6IGZhbHNlCn10AApmaWVsZFR5cGVzc3IAEWphdmEudXRpbC5IYXNoTWFwBQfawcMWYNEDAAJGAApsb2FkRmFjdG9ySQAJdGhyZXNob2xkeHA/QAAAAAAADHcIAAAAEAAAAAF0AARob3N0fnIAKW9yZy5vcGVuc2VhcmNoLnNxbC5kYXRhLnR5cGUuRXhwckNvcmVUeXBlAAAAAAAAAAASAAB4cgAOamF2YS5sYW5nLkVudW0AAAAAAAAAABIAAHhwdAACSVB4eA==\\\"}\",\"lang\":\"opensearch_compounded_script\",\"params\":{\"utcTimestamp\":1753841507558483000}},\"boost\":1.0}},\"_source\":{\"includes\":[\"host\"],\"excludes\":[]},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}]}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
    }
}
```

But in v2, it's actually pushed down as a range query:

```
{
    "root": {
        "name": "ProjectOperator",
        "description": {
            "fields": "[date]"
        },
        "children": [
            {
                "description": {
                    "request": "OpenSearchQueryRequest(indexName=dates, sourceBuilder={\"from\":0,\"size\":10000,\"timeout\":\"1m\",\"query\":{\"range\":{\"date\":{\"from\":\"2000-10-10\",\"to\":null,\"include_lower\":false,\"include_upper\":true,\"boost\":1.0}}},\"_source\":{\"includes\":[\"date\"],\"excludes\":[]},\"sort\":[{\"_doc\":{\"order\":\"asc\"}}]}, needClean=true, searchDone=false, pitId=s9y3QQEFZGF0ZXMWVzNCQ2U0UVNUX3Fzbmd0dXJuT3ZOdwAWRThCb1RVREhRYjZremM2SGZCTG9zZwAAAAAAAAAAGxZmUGtnT0ZWSFFNNmdEenZtcU95YnJnARZXM0JDZTRRU1RfcXNuZ3R1cm5Pdk53AAA=, cursorKeepAlive=1m, searchAfter=null, searchResponse=null)"
                },
                "children": []
            }
        ]
    }
}
```

```bash
curl --location 'http://localhost:9200/_plugins/_ppl/_explain' \
--header 'Content-Type: application/json' \
--data '{
  "query" : "source=dates | where date > '\''2000-10-10'\'' | fields date"
}
'
```

It should push down IP comparison as a range query like that in v2

 - OS: MacOS
 - Version 3.1.0 (abc5309a3f0d80e89b1a85a5d4e6c0ef8bdef0e2)
 - Plugins opensearch-sql

_Originally posted by @yuancu in https://github.com/opensearch-project/sql/issues/3914#issuecomment-3130552690_

yuancu:
### Description

IP comparison was pushed down as script. However, DSL supports range query with IP like the following:
```json
{"range":{"host":{"from":"1.1.1.1","to":null}}}
```

This PR push down IP comparison to range queries like above.

### Related Issues
Resolves #3937 

### Check List


core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
```diff
@@ -109,12 +110,19 @@ public class PPLBuiltinOperators extends ReflectiveSqlOperatorTable {
...
-  public static final SqlOperator EQUALS_IP = CompareIpFunction.equals().toUDF("EQUALS_IP");
-  public static final SqlOperator GREATER_IP = CompareIpFunction.greater().toUDF("GREATER_IP");
-  public static final SqlOperator GTE_IP = CompareIpFunction.greaterOrEquals().toUDF("GTE_IP");
-  public static final SqlOperator LESS_IP = CompareIpFunction.less().toUDF("LESS_IP");
-  public static final SqlOperator LTE_IP = CompareIpFunction.lessOrEquals().toUDF("LTE_IP");
+      CompareIpFunction.notEquals()
+          .withReverse(lookupOperator("NOT_EQUALS_IP"))
+          .toUDF("NOT_EQUALS_IP");
+  public static final SqlOperator EQUALS_IP =
+      CompareIpFunction.equals().withReverse(lookupOperator("EQUALS_IP")).toUDF("EQUALS_IP");
```
LantaoJin: can you add some doc in `withReverse()` to explain when should we use this API?

LantaoJin: And these changes are making the code harder to read and maintain. Can this `withReverse(lookupOperator("EQUALS_IP"))` be hidden? cc @ishaoxy as the `CompareIpFunction` author, do you understand what the new pattern code do?

ishaoxy: Truly a little hard for me. orz

yuancu: I created javadoc for this method.

Why I added this method:

1. `PredicateAnalyzer` relies on the `SqlKind` attribute of a comparator to convert comparison to DSL. E.g. if an operator has sql kind `SqlKind.GREATER_THAN_OR_EQUAL`, a DSL is created with `QueryExpression.create(pair.getKey()).gte(pair.getValue())`
2. To reuse this mechanism and process IP comparison in the same way, I override the `getKind` method of IP comparators with corresponding SQL kind.
3. When a `RelBuilder` normalizes the condition of a filter, it creates a hash code for the comparison operator. During this process, it will call `reverse()` of a `SqlOperator` to retrieve its reverse if it finds the comparison has to be normalized to a certain order.
4. This requires the `LESS_IP` UDF to have a reverse `GREATER_IP`. But they are defined sequentially, I have to find a way to assign their reverse later after all operators in `PPLBuiltinOperators` are instantiated.
5. That's where `withReverse` and `lookupOperator` come in. This delays the assignment of the reverse operator to the first time the `reverse()` of an IP comparison UDF is called -- by that time, all operators in `PPLBuiltinOperators` are already instantiated.

yuancu: Hiding it is a bit tricky for me. The reverse operator of a operator is defined at the same time with the operator itself in `PPLBuiltinOperators`. Therefore, looking it up will refer to `PPLBuiltinOperators`. If I hide `withReverse` by moving the lookup into `CompareIpFunction`, this will introduce circular import.

However, I can make the reverse operator an argument to the constructor of an IP comparator:

```java
public static final SqlOperator GREATER_IP =
      CompareIpFunction.greater(lookupOperator("LESS_IP")).toUDF("GREATER_IP");
```

And the `CompareIPFunction` class will be like:

```java
public class CompareIpFunction extends ImplementorUDF {
  private final SqlKind kind;
  /**
   * The reverse operator supplier that provides the reverse SQL operator
   *
   * <p>This method is used to establish the reversed relationship between comparison operators
   * (e.g., "less than" and "greater than"). When the query optimizer normalizes expressions, it may
   * need to transform "b > a" to "a < b".
   *
   * <p>E.g. in the {@code hashCode} method of {@link org.apache.calcite.rex.RexNormalize}#L115, it
   * always converts <i>B [comparator] A</i> to <i>A [reverse_comparator] B</i> if the ordinal of
   * the reverse of the comparator is smaller.
   *
   * <p>IP comparison functions use this to inform the optimizer that ip_less_than is the reverse of
   * ip_greater_than, allowing for proper query normalization.
   */
  private Supplier<SqlOperator> reverse;

  private CompareIpFunction(SqlKind kind, Supplier<SqlOperator> reverse) {
    super(new CompareImplementor(kind), NullPolicy.ANY);
    this.kind = kind;
    this.reverse = reverse;
  }

  public static CompareIpFunction greater(Supplier<SqlOperator> reverse) {
    return new CompareIpFunction(SqlKind.GREATER_THAN, reverse);
  }
...
```

Will this be more clear? If so, I'll update the implementation.

LantaoJin: No. maybe just hide the lookupOperator(),  change the interface of `withReverse` to
```
public static final SqlOperator EQUALS_IP =
      CompareIpFunction.equals().withReverse("EQUALS_IP").toUDF("EQUALS_IP");
```

yuancu: This will be a little tricky. `lookupOperator` relies on `PPLBuiltinOperators`. Hiding `lookupOperator` means moving operator lookup into `CompareIpFunction`. This will introduce cross dependency (`PPLBuiltinOperator -> CompareIpFunction` and `CompareIpFunction -> PPLBuiltinOperator`).

Besides, it will reduce coupling to have the `lookupOperator` staying within where the operators are defined. `CompareIPFunction` does not need to know anything about other operators or how will it be used.

qianheng-aws: > When a RelBuilder normalizes the condition of a filter, it creates a hash code for the comparison operator. During this process, it will call reverse() of a SqlOperator to retrieve its reverse if it finds the comparison has to be normalized to a certain order.

1. What's benefit of overriding this API? Or what's the main disadvantage of viewing `a < b` and `b > a` as 2 different expressions?

2. As for implementation, will it be better to add the reverse relationship unified in UserDefinedFunctionBuilder? Furthermore, if one operator has its reverse operator, we could invoke `toBinaryUDF` which has implementation of `reverse` based on functions' name. I think it's also fine not to add such API. Please refer to `SqlBinaryOperator` in Calcite, which implements their reverse based on SqlKind. By that way, we don't have to specify each functions' reverse here.

yuancu: 1. It is a required override since Calcite will invoke `reverse()` and perform subsequent operations since it assumes any `SqlOperator` with `SqlKind` like `LESS_THAN` must have a reverse operator.

2. I get it now. Yes, it'd doable. That will be neater and hide confusing details. I'll update the implementation.

yuancu: @qianheng-aws Refactored as suggested. Please review again. Thanks!

PR #3972, opened 2025-08-04
Support pushdown dedup with Calcite

### LINKED ISSUE #3913: [FEATURE] Support dedup pushdown with Calcite
[Dedup](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/dedup.rst) is a common PPL command which cannot be pushdown in v2.

Convert the dedup to [collapes search](https://docs.opensearch.org/docs/latest/search-plugins/searching-data/collapse-search/) in v3.

LantaoJin:
### Description
Support pushdown dedup with Calcite, if all conditions matched:
- number of retaining events no larger than 1
- dedup field list only contain one field
- the dedup field is one of field of index
- the dedup field type is keyword (or contain keyword subfield) or numeric
- the dedup option `keepempty=false` or missing

PPL query `source=account | fields account_number, gender, age | dedup gender`
Before:
```
  LogicalProject(account_number=[$0], gender=[$1], age=[$2])
    LogicalFilter(condition=[<=($3, 1)])
      LogicalProject(account_number=[$0], gender=[$1], age=[$2], _row_number_dedup_=[ROW_NUMBER() OVER (PARTITION BY $1 ORDER BY $1)])
        LogicalFilter(condition=[IS NOT NULL($1)])
          LogicalProject(account_number=[$0], gender=[$4], age=[$8])
            CalciteLogicalIndexScan(table=[[OpenSearch, account]])
```
After:
```
CalciteEnumerableIndexScan(table=[[OpenSearch, account]], PushDownContext=[
[PROJECT->[account_number, gender, age], FILTER->IS NOT NULL($1), COLLAPSE->gender, LIMIT->10000]

{
    "from": 0,
    "size": 10000,
    "timeout": "1m",
    "query": {
        "exists": {
            "field": "gender",
            "boost": 1.0
        }
    },
    "_source": {
        "includes": ["account_number", "gender", "age"],
        "excludes": []
    },
    "collapse": {
        "field": "gender.keyword"
    }
}
```

Besides, I open https://github.com/opensearch-project/sql/issues/3973 as followup to support complex `dedup` cases.

### Related Issues
Resolves #3913 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -155,6 +157,44 @@ private static RexNode constructCondition(List<RexNode> conditions, RexBuilder r
...
+      return null;
+    }
+    // For text, use its subfield if exists.
+    String field = OpenSearchTextType.toKeywordSubField(originalFieldName, fieldType);
+    if (field == null) {
+      LOG.debug("Cannot pushdown the dedup due to no keyword subfield for {}.", fieldName);
+      return null;
+    }
+    CalciteLogicalIndexScan newScan = this.copyWithNewSchema(finalOutput.getRowType());
+    newScan.pushDownContext.add(
```
penghuo, line 188: what is default sort order of opensearch collapse if no sort-key?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -155,6 +157,44 @@ private static RexNode constructCondition(List<RexNode> conditions, RexBuilder r
...
+    if (fieldType == null) {
+      // the fieldName must be one of index fields
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Cannot pushdown the dedup '{}' due to it is not a index field", fieldName);
+      }
+      return null;
+    }
+    ExprType originalExprType = fieldType.getOriginalExprType();
+    String originalFieldName = originalExprType.getOriginalPath().orElse(fieldName);
+    if (!ExprCoreType.numberTypes().contains(originalExprType)
```
penghuo, line 169: why number type is not supported?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
penghuo: tie-break logic in OpenSearch. shard_id, and _doc. 
this is OpenSearch default `search order`, Add it to doc?
```
    /**
     * if we need to tie-break since score / sort value are the same we first compare shard index (lower shard wins)
     * and then iff shard index is the same we use the hit index.
     */
    static boolean tieBreakLessThan(ShardRef first, ScoreDoc firstDoc, ShardRef second, ScoreDoc secondDoc) {
        final int firstShardIndex = first.getShardIndex(firstDoc);
        final int secondShardIndex = second.getShardIndex(secondDoc);
        // Tie break: earlier shard wins
        if (firstShardIndex < secondShardIndex) {
            return true;
        } else if (firstShardIndex > secondShardIndex) {
            return false;
        } else {
            // Tie break in same shard: resolve however the
            // shard had resolved it:
            assert first.hitIndex != second.hitIndex;
            return first.hitIndex < second.hitIndex;
        }
    }
```

penghuo: tie-break logic in OpenSearch. shard_id, and _doc.
documenting this behavior in the deduplication logic to clarify how OpenSearch handles tie-breaking by default.

    /**
     * if we need to tie-break since score / sort value are the same we first compare shard index (lower shard wins)
     * and then iff shard index is the same we use the hit index.
     */
    static boolean tieBreakLessThan(ShardRef first, ScoreDoc firstDoc, ShardRef second, ScoreDoc secondDoc) {
        final int firstShardIndex = first.getShardIndex(firstDoc);
        final int secondShardIndex = second.getShardIndex(secondDoc);
        // Tie break: earlier shard wins
        if (firstShardIndex < secondShardIndex) {
            return true;
        } else if (firstShardIndex > secondShardIndex) {
            return false;
        } else {
            // Tie break in same shard: resolve however the
            // shard had resolved it:
            assert first.hitIndex != second.hitIndex;
            return first.hitIndex < second.hitIndex;
        }
    }

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
LantaoJin: only supports keywords and numeric types. There is a `!` before the condition.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
LantaoJin: > what is default sort order of opensearch collapse if no sort-key?

The default sort order is added by other logic, for example `_doc`. I will check the sort behavior of dedup in v2 and spl.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -121,6 +121,7 @@ public double estimateRowCount(RelMetadataQuery mq) {
                 switch (action.type) {
                       case AGGREGATION -> mq.getRowCount((RelNode) action.digest);
                       case PROJECT, SORT -> rowCount;
+                      case COLLAPSE -> NumberUtil.multiply(rowCount, estimateRowCountFactor);
```
qianheng-aws: Seems we don't need to multiply rowCount by estimateRowCountFactor 2 times. There is such operation below already.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
```diff
@@ -0,0 +1,107 @@
...
+            .withOperandSupplier(
+                b0 ->
+                    b0.operand(LogicalProject.class)
+                        .oneInput(
+                            b1 ->
+                                b1.operand(LogicalFilter.class)
+                                    .predicate(
+                                        f ->
+                                            f.getCondition().getKind()
+                                                == SqlKind.LESS_THAN_OR_EQUAL)
```
qianheng-aws: Why not adding a more precise predicator here?  It only checks the operation is `<=` but not checking literal is 1 until `onMatch`. However, we have chance to skip the `onMatch` earlier here, then it has no needs to do predication for its children.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    final LogicalProject finalOutput = call.rel(0);
+    final LogicalFilter numOfDedupFilter = call.rel(1);
+    final LogicalProject projectWithWindow = call.rel(2);
+    final CalciteLogicalIndexScan scan = call.rel(3);
+    RexLiteral numLiteral =
+        PlanUtils.findLiterals(numOfDedupFilter.getCondition(), true).getFirst();
```
qianheng-aws: Seems not making sense to find the literal deeply. e.g. We don't want to do optimization for the filter `$? + 1 <= 1`, the only case we want to match is `$? <= 1`.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
```diff
@@ -0,0 +1,107 @@
...
+    List<RexWindow> windows = PlanUtils.getRexWindowFromProject(projectWithWindow);
+    if (windows.isEmpty() || windows.stream().anyMatch(w -> w.partitionKeys.size() > 1)) {
+      // TODO leverage inner_hits for multiple partition keys
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Cannot pushdown the dedup with multiple fields");
+      }
+      return;
+    }
+    final List<String> fieldNameList = projectWithWindow.getInput().getRowType().getFieldNames();
+    List<Integer> selectColumns = PlanUtils.getSelectColumns(windows.getFirst().partitionKeys);
```
qianheng-aws, line 51: Shall we check the size of `windows` is exactly 1?  Or maybe do such check earlier in the predicator.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
qianheng-aws: We also need to check the field in the condition is the `_row_number_dedup`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
LantaoJin: It's on purpose, I cannot figure out the selectivity such as FILTER does. So multiply the estimateRowCountFactor twice.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
LantaoJin: > We also need to check the field in the condition is the _row_number_dedup

Checked in rule config.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
LantaoJin: size of windows is exactly 1 for _row_number_dedup_.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
LantaoJin: For further optimization, the number could be `> 1`

core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -298,7 +297,7 @@ private FrameworkConfig buildFrameworkConfig() {
             .parserConfig(SqlParser.Config.DEFAULT) // TODO check
             .defaultSchema(opensearchSchema)
             .traitDefs((List<RelTraitDef>) null)
-            .programs(Programs.calc(DefaultRelMetadataProvider.INSTANCE))
+            .programs(Programs.standard())
```
LantaoJin, line 300: use Programs.standard() in building context because the actual execution uses it.

integ-test/src/test/java/org/opensearch/sql/calcite/tpch/CalcitePPLTpchIT.java
```diff
@@ -54,7 +54,7 @@ public void testQ1() throws IOException {
         schema("count_order", "bigint"));
     verifyDataRows(
         actual,
-        rows(
+        closeTo(
```
LantaoJin, line 57: fix flaky

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
qianheng-aws: As discussed offline, we'd better prevent going into `match` as we could as possible. So we need to put `validFilter ` here and it should only be changed like current code or update `validFilter` until in that PR for supporting `>1`.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchIndexScanRule.java
```diff
@@ -26,6 +26,14 @@ static boolean noAggregatePushed(AbstractCalciteIndexScan scan) {
     return table.unwrap(OpenSearchIndex.class) != null;
   }
 
+  static boolean isAggregatePushed(AbstractCalciteIndexScan scan) {
```
qianheng-aws: Are these 2 functions used anywhere?

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -347,4 +355,52 @@ static RexNode derefMapCall(RexNode rexNode) {
     }
     return rexNode;
   }
+
+  static List<RexLiteral> findLiterals(RexNode node, boolean deep) {
```
qianheng-aws: Seems never used

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchDedupPushdownRule.java
LantaoJin: > As discussed offline, we'd better prevent going into `match` as we could as possible. So we need to put `validFilter ` here and it should only be changed like current code or update `validFilter` until in that PR for supporting `>1`.

done

PR #3982, opened 2025-08-05
Fix DOUBLE to STRING cast rendering zero values in scientific notation

### LINKED ISSUE #3947: [BUG] Casting rounded float to string returns "0E0" instead of "0.0"
**What is the bug?**
Casting rounded float to string returns "0E0" instead of "0.0"

```
source=test002 
| eval vv=round((cast(1 as double)/100000)*100, 2) 
| eval sv=cast(vv as string) 
| fields sv

sv = "0E0"
```

```
sv = "0.0"
```

 - Plugins

n/a

n/a

yuancu:
### Description

When calcite is enabled, casting a double (not decimal) of value `0.0` to string will result in `0E0`. The root cause is analyzed [here](https://github.com/opensearch-project/sql/issues/3947#issuecomment-3153922774).

This PR fixed the issue by replacing the cast call to a `NUMBER_TO_STRING` call.

The logical plan of query `source=opensearch-sql_test_index_account | head 1 | eval d= cast(0 as double) | eval s=cast(d as string) | fields s` is now:

```
  LogicalProject(s=[NUMBER_TO_STRING(0.0E0:DOUBLE)])
    LogicalSort(fetch=[1])
      CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])
```

Additionally, decimals are also cast with `NUMBER_TO_STRING`. Decimals like `0.xxx` were cast to string as `.xxx` before this PR. Now they are  represented as `0.xxx`

### Related Issues
Resolves #3947 

### Check List


yuancu:
@dai-chen  Yes

```
SELECT CAST(CAST(0.0 AS DOUBLE) AS VARCHAR(50));
+--------+
| EXPR$0 |
+--------+
| 0E0    |
+--------+
```

But I think it's a convention rather than bug.

qianheng-aws:
> This PR fixed the issue by replacing the cast call to a FORMAT_NUMBER call with the decimal pattern: 0.0###############.

Why not just using ` Expressions.call(
              fromPrimitive.boxClass,
              "toString",
              operand)` directly?
 
 
 I think the main conflict is Calcite has decimal format for 0 of double/float type, while we don't want it.

LantaoJin:
> LogicalProject(s=[FORMAT_NUMBER(0.0E0:DOUBLE, '0.0###############')])

-1

@yuancu  you try the suggestion of https://github.com/opensearch-project/sql/pull/3982#issuecomment-3157521012?

yuancu:
@qianheng-aws @LantaoJin Thanks for pointing out. That will be much more straightforward.

Initially I wanted to avoid creating custom operators by reusing `SqlLibraryOperators.FormatNumber`. Yet now that I did create a new custom operator, leveraging `number.toString()` will be more straightforward.

qianheng-aws (approved):

core/src/main/java/org/opensearch/sql/expression/function/udf/math/NumberToStringFunction.java
```diff
@@ -0,0 +1,51 @@
...
+    return PPLOperandTypes.NUMERIC;
+  }
+
+  public static class NumberToStringImplementor implements NotNullImplementor {
+
+    @Override
+    public Expression implement(
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+      Expression operand = translatedOperands.get(0);
+      return Expressions.call(Expressions.box(operand), "toString");
```
qianheng-aws, line 48:
```    
Expressions.call(Primitive.ofBox(operand.getType()).getBoxClass(), "toString", operand)
```
would be a little better since it doesn't need to create a boxed object for non-nullable fields. But the performance difference should be very trivial

yuancu: I agree. Updated the implementation.

qianheng-aws (dismissed):

LantaoJin:
@yuancu could you fix the decimal-to-string issue too? https://github.com/opensearch-project/sql/issues/3947#issuecomment-3157751366

yuancu:
Fixed decimal casting. Reverted to `Expressions.call(Expressions.box(operand), "toString")` for a simpler and unified implementation. The performance degradation should be negligible.

dai-chen:
So our current approach is to add UDF for any conversion not supported by Calcite built-in CAST function, right? I'm thinking shall we just have one enhanced CAST function impl instead of adding UDF per conversion. Any thoughts?

yuancu:
Yes, our current approach is to reuse Calcite's built-in cast as much as possible until its behavior does not match expectation. It's feasible to have an enhanced CAST UDF to handle all cast.

Downside:
- throw away what's already implemented in `RexBuilder.makeCast`
- cannot directly reuse existing UDFs (E.g. DATE, TIME, TIMESTAMP) since you can not call another operator inside CAST UDF

Upside:
- Cleaner and more unified cast

LantaoJin (approved):

LantaoJin:
> So our current approach is to add UDF for any conversion not supported by Calcite built-in CAST function, right? I'm thinking shall we just have one enhanced CAST function impl instead of adding UDF per conversion. Any thoughts?

we could have a rule to remove this UDF in plan for Spark (if the table is spark table, or a flag, enable the rule of `RemovingNumberToStringUDF`.

PR #3971, opened 2025-08-04
SUM aggregation enhancement on operations with literal

qianheng-aws:
### Description
SUM aggregation enhancement on operations with literal

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/3967

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3853, opened 2025-07-07
Skipping codegen and compile for Scan only plan

LantaoJin:
### Description
Currently, when the Calcite engine is enabled, **​​all query plans**​​ -- regardless of their structure -- are converted to `EnumerableRel`. This means the plan is transformed into a `Linq4j` Expression, generates Java code, and undergoes just-in-time (JIT) compilation before execution.

For ​​simple or fully pushdown-compatible queries​​ (e.g., `source=t`, `source=t | where a=1`, `source=t | sort a`, or `source=t | where a=1 | sort b | head 10`), the optimized plans only contain node `EnumerableTableScan`. Despite their simplicity, these plans ​​unnecessarily undergo code generation and dynamic compilation​​, introducing overhead.

To ​​reduce codegen and compilation time​​, we propose converting such plans to `BindableRel` instead of `EnumerableRel`. This bypasses code generation and compilation entirely, ​​improving plan execution time by **~30%** (By benchmarking the simple queries by https://github.com/opensearch-project/sql/pull/3822)

### Related Issues
Resolves #3852 

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3983, opened 2025-08-05
ResourceMonitor only checks the memory health by calculating the memory usage after GC of old gen in v3

LantaoJin:
### Description
In v2, `ResourceMonitor` gets the memory usage by using total memory - free memory of runtime.
But in v3, Calcite will produce lots of **recyclable** memory garbage during its planning and implementation process. For q30, it produces around 70 mb in optimizing + 30 mb in implementation on local test, which can all be GC, ref https://github.com/opensearch-project/sql/pull/3971#discussion_r2251784085

This PR changes the behaviour of `ResourceMonitor` for v3:
ResourceMonitor only checks the memory health by calculating the memory usage after GC of old gen.

### Related Issues
Also can improve/resolve the issue of #3750 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/executor/QueryService.java
+++ b/core/src/main/java/org/opensearch/sql/executor/QueryService.java
@@ -112,11 +112,14 @@ public class QueryService {
         log.warn("Fallback to V2 query engine since got exception", t);
         executeWithLegacy(plan, queryType, listener, Optional.of(t));
       } else {
-        if (t instanceof Error) {
+        if (t instanceof Exception) {
+          listener.onFailure((Exception) t);
+        } else if (t instanceof VirtualMachineError) {
+          // throw and fast fail the VM errors such as OOM (same with v2).
+          throw t;
+        } else {
           // Calcite may throw AssertError during query execution.
           listener.onFailure(new CalciteUnsupportedException(t.getMessage(), t));
-        } else {
-          listener.onFailure((Exception) t);
         }
       }
     }
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.monitor;
+
+import com.sun.management.GarbageCollectionNotificationInfo;
+import java.lang.management.GarbageCollectorMXBean;
+import java.lang.management.ManagementFactory;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
+import javax.management.Notification;
+import javax.management.NotificationEmitter;
+import javax.management.NotificationListener;
+import javax.management.openmbean.CompositeData;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/** Get memory usage from GC notification listener, which is used in Calcite engine. */
+public class GCedMemoryUsage implements MemoryUsage {
+  private static final Logger LOG = LogManager.getLogger();
+
+  private GCedMemoryUsage() {
+    registerGCListener();
+  }
+
+  // Lazy initialize the instance to avoid register GCListener in v2.
+  private static class Holder {
+    static final MemoryUsage INSTANCE = new GCedMemoryUsage();
+  }
+
+  /**
+   * Get the singleton instance of GCedMemoryUsage.
+   *
+   * @return GCedMemoryUsage instance
+   */
+  public static MemoryUsage getInstance() {
+    return Holder.INSTANCE;
+  }
+
+  private final AtomicLong usage = new AtomicLong(-1);
+
+  @Override
+  public long usage() {
+    return usage.get();
+  }
+
+  @Override
+  public void setUsage(long value) {
+    usage.set(value);
+  }
+
+  private void registerGCListener() {
+    List<GarbageCollectorMXBean> gcBeans = ManagementFactory.getGarbageCollectorMXBeans();
+    for (GarbageCollectorMXBean gcBean : gcBeans) {
+      if (gcBean instanceof NotificationEmitter) {
+        NotificationEmitter emitter = (NotificationEmitter) gcBean;
+        emitter.addNotificationListener(new FullGCListener(), null, null);
+      }
+    }
+  }
+
+  private static class FullGCListener implements NotificationListener {
+    @Override
+    public void handleNotification(Notification notification, Object handback) {
+      if (notification
+          .getType()
+          .equals(GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION)) {
+        CompositeData cd = (CompositeData) notification.getUserData();
+        GarbageCollectionNotificationInfo info = GarbageCollectionNotificationInfo.from(cd);
+
+        String gcAction = info.getGcAction();
+        LOG.info("gcAction: {}", gcAction);
+        if (gcAction.contains("major")
+            || gcAction.contains("concurrent")
+            || gcAction.contains("old")
+            || gcAction.contains("full")) {
+          Map<String, java.lang.management.MemoryUsage> memoryUsageAfterGc =
+              info.getGcInfo().getMemoryUsageAfterGc();
+          String possibleOldGenKey = findOldGenKey(memoryUsageAfterGc.keySet());
+          if (possibleOldGenKey != null) {
+            java.lang.management.MemoryUsage oldGenUsage =
+                memoryUsageAfterGc.get(possibleOldGenKey);
+            getInstance().setUsage(oldGenUsage.getUsed());
+            LOG.info("Full GC detected, memory usage after GC is {} bytes.", getInstance().usage());
+          }
+        }
+      }
+    }
+
+    private String findOldGenKey(Set<String> memoryPoolNames) {
+      LOG.info("Memory pool names: {}", memoryPoolNames);
+      List<String> possibleOldGenKeys =
+          Arrays.asList("G1 Old Gen", "PS Old Gen", "CMS Old Gen", "Tenured Gen", "Old Gen");
+      for (String key : possibleOldGenKeys) {
+        if (memoryPoolNames.contains(key)) {
+          return key;
+        }
+      }
+      return null;
+    }
+  }
+}
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/MemoryUsage.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/MemoryUsage.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.monitor;
+
+public interface MemoryUsage {
+  long usage();
+
+  void setUsage(long usage);
+}
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchMemoryHealthy.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchMemoryHealthy.java
@@ -9,6 +9,7 @@ import com.google.common.annotations.VisibleForTesting;
 import java.util.concurrent.ThreadLocalRandom;
 import lombok.NoArgsConstructor;
 import lombok.extern.log4j.Log4j2;
+import org.opensearch.sql.common.setting.Settings;
 
 /** OpenSearch Memory Monitor. */
 @Log4j2
@@ -16,15 +17,26 @@ public class OpenSearchMemoryHealthy {
   private final RandomFail randomFail;
   private final MemoryUsage memoryUsage;
 
-  public OpenSearchMemoryHealthy() {
+  public OpenSearchMemoryHealthy(Settings settings) {
     randomFail = new RandomFail();
-    memoryUsage = new MemoryUsage();
+    memoryUsage =
+        isCalciteEnabled(settings)
+            ? GCedMemoryUsage.getInstance()
+            : RuntimeMemoryUsage.getInstance();
   }
 
   @VisibleForTesting
   public OpenSearchMemoryHealthy(RandomFail randomFail, MemoryUsage memoryUsage) {
     this.randomFail = randomFail;
     this.memoryUsage = memoryUsage;
+  }
+
+  private boolean isCalciteEnabled(Settings settings) {
+    if (settings != null) {
+      return settings.getSettingValue(Settings.Key.CALCITE_ENGINE_ENABLED);
+    } else {
+      return false;
+    }
   }
 
   /** Is Memory Healthy. Calculate based on the current heap memory usage. */
@@ -47,14 +59,6 @@ public class OpenSearchMemoryHealthy {
   static class RandomFail {
     public boolean shouldFail() {
       return ThreadLocalRandom.current().nextBoolean();
-    }
-  }
-
-  static class MemoryUsage {
-    public long usage() {
-      final long freeMemory = Runtime.getRuntime().freeMemory();
-      final long totalMemory = Runtime.getRuntime().totalMemory();
-      return totalMemory - freeMemory;
     }
   }
 
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/system/OpenSearchSystemIndexEnumerator.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/system/OpenSearchSystemIndexEnumerator.java
@@ -12,10 +12,14 @@ import lombok.ToString;
 import org.apache.calcite.linq4j.Enumerator;
 import org.opensearch.sql.data.model.ExprNullValue;
 import org.opensearch.sql.data.model.ExprValue;
+import org.opensearch.sql.exception.NonFallbackCalciteException;
+import org.opensearch.sql.monitor.ResourceMonitor;
 import org.opensearch.sql.opensearch.request.system.OpenSearchSystemRequest;
 
 /** Supports a simple iteration over a collection for OpenSearch system index */
 public class OpenSearchSystemIndexEnumerator implements Enumerator<Object> {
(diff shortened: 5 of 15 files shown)


core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -112,11 +112,14 @@ public void executeWithCalcite(
         log.warn("Fallback to V2 query engine since got exception", t);
         executeWithLegacy(plan, queryType, listener, Optional.of(t));
       } else {
-        if (t instanceof Error) {
+        if (t instanceof Exception) {
+          listener.onFailure((Exception) t);
+        } else if (t instanceof VirtualMachineError) {
+          // throw and fast fail the VM errors such as OOM (same with v2).
+          throw t;
+        } else {
```
LantaoJin, line 119: To align with behaviour of v2, OOM error will fail the whole JVM. We caught OOM error in previous implementation.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/system/OpenSearchSystemIndexEnumerator.java
```diff
@@ -25,10 +29,22 @@ public class OpenSearchSystemIndexEnumerator implements Enumerator<Object> {
...
+  public OpenSearchSystemIndexEnumerator(
+      List<String> fields, OpenSearchSystemRequest request, ResourceMonitor monitor) {
     this.fields = fields;
     this.request = request;
+    this.monitor = monitor;
+    this.queryCount = 0;
     this.current = null;
+    if (!this.monitor.isHealthy()) {
+      throw new NonFallbackCalciteException("insufficient resources to run the query, quit.");
+    }
```
LantaoJin, line 47: Healthy checking is missed in access OpenSearchSystemIndex.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public void handleNotification(Notification notification, Object handback) {
+      if (notification
+          .getType()
+          .equals(GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION)) {
+        CompositeData cd = (CompositeData) notification.getUserData();
+        GarbageCollectionNotificationInfo info = GarbageCollectionNotificationInfo.from(cd);
+
+        String gcAction = info.getGcAction();
+        LOG.info("gcAction: {}", gcAction);
```
penghuo: it should be debug log.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/MemoryUsage.java
```diff
@@ -0,0 +1,12 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.monitor;
+
+public interface MemoryUsage {
+  long usage();
+
+  void setUsage(long usage);
+}
```
dai-chen, line 13: Could you add missing javadoc? Meanwhile currently the API looks more like a data model class?

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
```diff
@@ -0,0 +1,107 @@
...
+    List<GarbageCollectorMXBean> gcBeans = ManagementFactory.getGarbageCollectorMXBeans();
+    for (GarbageCollectorMXBean gcBean : gcBeans) {
+      if (gcBean instanceof NotificationEmitter) {
+        NotificationEmitter emitter = (NotificationEmitter) gcBean;
+        emitter.addNotificationListener(new FullGCListener(), null, null);
+      }
+    }
+  }
+
+  private static class FullGCListener implements NotificationListener {
```
dai-chen: What will happen if this is no full GC happened? We fall back to RuntimeMemoryUsage?
Also any significant delay on usage update if full GC is not frequent?

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchMemoryHealthy.java
```diff
@@ -9,16 +9,20 @@
...
 public class OpenSearchMemoryHealthy {
   private final RandomFail randomFail;
   private final MemoryUsage memoryUsage;
 
-  public OpenSearchMemoryHealthy() {
+  public OpenSearchMemoryHealthy(Settings settings) {
     randomFail = new RandomFail();
-    memoryUsage = new MemoryUsage();
+    memoryUsage =
+        isCalciteEnabled(settings)
```
penghuo: Any concern of using GCedMemoryUsage as default, and deprecated RuntimeMemoryUsage?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/system/OpenSearchSystemIndexEnumerator.java
dai-chen: I think we discussed this previously. Can we do this in non-intrusive way? Maybe AOP/wrap on operator level?

dai-chen (commented):
I just want to understand the idea more deeply: the assumption is that most objects—such as Calcite’s internal structures—are short-lived, meaning they die in the Young Generation and do not survive into the Old Generation. As a result, the memory usage calculated from this approach reflects a lower bound of the actual memory pressure?

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: Not only full GCs are listened, but also concurrent GCs are notified, the name could be `NonYoungGcListener` or `OldGenGCListener`?

PS: CGC is quite frequent in G1 since it's not stop-the-world.

LantaoJin:
> I just want to understand the idea more deeply: the assumption is that most objects in Young Gen—such as Calcite’s internal structures—are short-lived, meaning they die after GC and do not survive into the Old Gen, right? As a result, the memory usage calculated from this approach reflects a lower bound of the actual memory pressure?

Not actually, the objects can survive into the Old Gen but can be easily GCed by concurrent GC (not YGC).

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchMemoryHealthy.java
LantaoJin: No tech concern, just not want to introduce any breaking experiences for v2.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/MemoryUsage.java
LantaoJin: Sure. But `Record` class is implicitly final and cannot be extended.

core/src/main/java/org/opensearch/sql/executor/QueryService.java
qianheng-aws: It think it's OK to keep align with v2. But will it be better to trigger GC and then check memory usage again before making the service crash? Anyway, the primary purpose of this PR should not be blocked by this.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/system/OpenSearchSystemIndexEnumerator.java
LantaoJin: > I think we discussed this previously. Can we do this in non-intrusive way? Maybe AOP/wrap on operator level?

Yes, technically we could wrap it on operator level. But it will introduce
1. A new logical node `LogicalMonitor`
2. A `ConverterRule` to convert a logical node to physical node
3. A new physical node such as `EnumerableMonitor` similar to `ResourceMonitorPlan` in v2. 
4. Optimization #3853 has to find another way to workaround because the `root.rel` (physical plan) won't be a `Scannable` any more.

Not quite sure it's still worth to do the AOP way now.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: deleted.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/MemoryUsage.java
LantaoJin: javadoc added.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: renamed to `OldGenGCListener`

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Override
+  public void setUsage(long value) {
+    usage.set(value);
+  }
+
+  private void registerGCListener() {
+    List<GarbageCollectorMXBean> gcBeans = ManagementFactory.getGarbageCollectorMXBeans();
+    for (GarbageCollectorMXBean gcBean : gcBeans) {
```
qianheng-aws, line 61: Since we don't really handle notification for young GC, why do we add listener to all NotificationEmitter?

I think the current implementation will go into `handleNotification`  very frequency but then fast return, since young GC happens more frequent than others.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
```diff
@@ -0,0 +1,109 @@
...
+  private static class OldGenGCListener implements NotificationListener {
+    @Override
+    public void handleNotification(Notification notification, Object handback) {
+      if (notification
+          .getType()
+          .equals(GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION)) {
+        CompositeData cd = (CompositeData) notification.getUserData();
+        GarbageCollectionNotificationInfo info = GarbageCollectionNotificationInfo.from(cd);
+
+        String gcAction = info.getGcAction();
```
qianheng-aws: [question]Comparing to GC action, will `getGcName` be better here?

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
```diff
@@ -0,0 +1,109 @@
...
+        if (gcAction.contains("major")
+            || gcAction.contains("concurrent")
+            || gcAction.contains("old")
+            || gcAction.contains("full")) {
+          Map<String, java.lang.management.MemoryUsage> memoryUsageAfterGc =
+              info.getGcInfo().getMemoryUsageAfterGc();
+          String possibleOldGenKey = findOldGenKey(memoryUsageAfterGc.keySet());
+          if (possibleOldGenKey != null) {
+            java.lang.management.MemoryUsage oldGenUsage =
```
qianheng-aws: Shall we add the memory usage of other spaces as well instead of only old gen?

dai-chen:
> > I just want to understand the idea more deeply: the assumption is that most objects in Young Gen—such as Calcite’s internal structures—are short-lived, meaning they die after GC and do not survive into the Old Gen, right? As a result, the memory usage calculated from this approach reflects a lower bound of the actual memory pressure?
> 
> Not actually, the objects can survive into the Old Gen but can be easily GCed by concurrent GC (not YGC).

Could you clarify the following points:

1. When we say "can be easily GCed," are we assuming that all objects in the Young Generation are short-lived (recyclable)? That's why I was asking whether this new calculation represents a lower bound for total memory usage.
2. In the description, we mentioned 70MB + 30MB—is that memory usage per query? If it's total, is it a major concern?
3. Have we rerun the benchmark with these new changes?

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
dai-chen: I see. So this means we also make assumption on GC type? Or this is applied to other popular collectors too?

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: no. `gcName` deeply depends on the GC type (G1、ZGC、Parallel GC) the name could be totally different for old gen garbage collection. `gcAction` is more stable.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: I will rename `oldGenUsage` to `mem` to avoid confusion. This size is not the memory size of old gen. Its the entire memory usage after old gen GC.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: Ok, will do a refactor to remove the notification for young.

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: done

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: done

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
LantaoJin: > Or this is applied to other popular collectors too?

Yes. The current code considered popular collectors such as G1, CMS, ZGC. But not all collectors in industry

LantaoJin:
> > > I just want to understand the idea more deeply: the assumption is that most objects in Young Gen—such as Calcite’s internal structures—are short-lived, meaning they die after GC and do not survive into the Old Gen, right? As a result, the memory usage calculated from this approach reflects a lower bound of the actual memory pressure?
> > 
> > 
> > Not actually, the objects can survive into the Old Gen but can be easily GCed by concurrent GC (not YGC).
> 
> Could you clarify the following points:
> 
> 1. When we say "can be easily GCed," are we assuming that all objects in the Young Generation are short-lived (recyclable)? That's why I was asking whether this new calculation represents a lower bound for total memory usage.
> 2. In the description, we mentioned 70MB + 30MB—is that memory usage per query? If it's total, is it a major concern?
> 3. Have we rerun the benchmark with these new changes?

1. No, **"can be easily GCed"** means the object can be GCed by Young GC or Concurrent GC. The PR doesn't care about how long an object can stay in Young Gen, instead, as long as the objects can be GCed in Young GC or Concurrent GC(Old Gen), in another word, "easy" to trigger and "easy" to execute (no STW). Again, YGC and CGC are all work in CMS and G1, in ZGC, there is no Young/Old Gen, but all GC are concurrent. So the latest code should work as well.
2.  "70MB + 30MB—is that memory usage per query " specified the Q30 of ClickBench query. It's just a user case. The major concern is after introducing Calcite, the memory usage could be much higher than v2 in optimizer stage. For a complex query (actually the Q30 is not complex enough), it may consume much higher memory than v2 since the Cascades style optimizer of Calcite really consumes memory to build equivalence set. But they should be easily GCed in YGC or CGC too. The old resource monitor may easy to trigger fast fail in v3.
3. Not yet. We can do it. But it may be no problem in benchmarking since the memory setting of EC2 is GB level. The failure is trigger easily in local test (laptop or Github CI).

core/src/main/java/org/opensearch/sql/executor/QueryService.java
LantaoJin: > But will it be better to trigger GC and then check memory usage again before making the service crash?

Maybe more discussion is required and track in a follow-up issue.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/system/OpenSearchSystemIndexEnumerator.java
dai-chen: Okay, probably in future we need this for other operators that possibly generate more records, such as cross/lateral join, right?

LantaoJin:
The IT could be still flaky in GitHub CI sometimes. But I cannot reproduce the resource limit issue in local with current implementation. But without this patch, resource limit issue fails easily in CalcitePPLClickBenchIT.test.

qianheng-aws (approved):

LantaoJin:
> The IT could be still flaky in GitHub CI sometimes. But I cannot reproduce the resource limit issue in local with current implementation. But without this patch, resource limit issue fails easily in CalcitePPLClickBenchIT.test.

Not see it again in latest two runs of CI. @dai-chen @penghuo can you check this PR again?

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchMemoryHealthy.java
```diff
@@ -27,6 +28,24 @@ public OpenSearchMemoryHealthy(RandomFail randomFail, MemoryUsage memoryUsage) {
     this.memoryUsage = memoryUsage;
   }
 
+  private MemoryUsage buildMemoryUsage(Settings settings) {
+    try {
+      return isCalciteEnabled(settings)
+          ? GCedMemoryUsage.getInstance()
+          : RuntimeMemoryUsage.getInstance();
+    } catch (MemoryUsageException e) {
```
penghuo: fallback to RuntimeMemoryUsage will not trigger; A runtime exception thrown during static field initialization is wrapped in ExceptionInInitializerError.

LantaoJin: fixed

qianheng-aws (approved):

PR #3986, opened 2025-08-06
Add sum and avg functions in eval

vamsimanohar:
### Description

### Related Issues

### Check List


dai-chen (commented):
I’m wondering if we evaluated other options. If I understand correctly, SUM(a, b, c) could be rewritten as an arithmetic expression, for example: `eval X = a + b + c`. Same works for `AVG`?

vamsimanohar:
> I’m wondering if we evaluated other implementation options. If I understand correctly, SUM(a, b, c) could be rewritten as an arithmetic expression, for example: `eval X = a + b + c`. The same approach also works for `AVG`?

Make sense..let me relook at the PR.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -293,6 +298,33 @@ private Function buildFunction(
...
+      throw new SyntaxCheckException(functionName + " function requires at least one argument");
+    }
+
+    List<UnresolvedExpression> arguments =
+        args.stream().map(this::visitFunctionArg).collect(Collectors.toList());
+
+    // Build the sum expression: a + b + c + ...
+    UnresolvedExpression functionExpr = arguments.get(0);
+    for (int i = 1; i < arguments.size(); i++) {
+      functionExpr = new Function("+", Arrays.asList(functionExpr, arguments.get(i)));
```
Swiddis: We're only doing recursion on one side here for the list, is that not going to cause issues for larger lists? Historically we've had to rewrite these to produce balanced trees. See e.g. https://github.com/opensearch-project/sql/pull/3660

penghuo: +1

vamsimanohar: sure will change.

Note:
if some one writes a large expression without using sum. Eg: 1 + 2 + 3.....
ANTLR creates a left leaning tree anyways. which requires a more involved change in grammar to optimize.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -271,9 +269,16 @@ public UnresolvedExpression visitCaseFunctionCall(
...
-        FUNCTION_NAME_MAPPING.getOrDefault(functionName.toLowerCase(Locale.ROOT), functionName),
-        ctx.functionArgs().functionArg());
+    final String mappedName =
+        FUNCTION_NAME_MAPPING.getOrDefault(functionName.toLowerCase(Locale.ROOT), functionName);
+
+    // Rewrite sum and avg functions to arithmetic expressions
+    if (SUM.getName().getFunctionName().equalsIgnoreCase(mappedName)
+        || AVG.getName().getFunctionName().equalsIgnoreCase(mappedName)) {
+      return rewriteSumAvgFunction(mappedName, ctx.functionArgs().functionArg());
+    }
```
dai-chen, line 279: np: I assume we will add more function like `min`, `max`, please see how to make this more maintainable. Maybe a generic reduce method like https://spark.apache.org/docs/latest/api/sql/index.html#reduce?

PR #3993, opened 2025-08-06
Support timechart command with Calcite

### LINKED ISSUE #3965: [RFC] Support `timechart` command in PPL
# Implement `timechart` command functionality in OpenSearch PPL
## Summary
Implement a `timechart` command in OpenSearch PPL that will allow users to generate temporal trend charts directly within PPL queries, streamlining the process of identifying patterns and anomalies in time-based datasets. It aggregates a specific field and returns a chart with time as the X-axis.

## Background
Currently, PPL users rely on the stats command for time-series data visualization and statistical analysis. While the stats command provides various aggregation computations and data processing capabilities, it was not specifically designed for time-series visualization use cases.
### Current PPL `stats` function capabilities:
* Verbose syntax for complex time-series queries
* No built-in visualization-ready formatting
* Manual sorting required (results not automatically time-ordered)
* No automatic time gap filling (missing time periods not handled)
* Span always appears as first column regardless of logical grouping preference

### Additional binning capabilities we want to add:
* Flexible span or bin parameter to control the length of time
* Other field: By default, when too many fields show up in the legend, timechart creates an OTHER field to capture the long tail
* by syntax, designed to accept only one grouping field
* Fill missing time periods with nulls

## Proposed Implementation
Create a dedicated `timechart` command with comprehensive syntax:

### Complete Timechart Command Syntax
```
... 
| timechart 
    [cont=<boolean>] 
    [fixedrange=<boolean>]
    [limit=<int>]
    [agg=<aggregation>] 
    [<bin-options>... ]
    ( (<single-aggregate> [by <split-by-clause>] ) | (<eval-expression>) by <split-by-clause> ) )
```

#### **`span=<value>`** - Fixed interval binning
source=logs | bin timestamp span=3600000 | stats count() by timestamp
#### **`bins=<number>`** - Fixed number of equal-width bins
source=sales | bin price bins=50 | stats count() by price
#### **`minspan=<value>`** - Automatic span calculation with minimum interval
source=metrics | bin timestamp minspan=300000 | stats count() by timestamp
#### **`aligntime=<value>`** - Time alignment for consistent bucketing
-- Align to earliest time
source=events | bin timestamp span=86400000 aligntime=earliest | stats count() by timestamp

-- Align to specific timestamp
source=logs | bin timestamp span=3600000 aligntime=1640995200000 | stats count() by timestamp
#### **`start=<num>` / `end=<num>`** - Range-bounded binning
source=data | bin value span=100 start=0 end=10000 | stats count() by value
### 2. Time-Based Binning
Support time scale units:

### 3. Field Aliasing

source=accounts | bin age span=10 AS age_group | stats count() by age_group
### Core Components
* **Grammar**: Extend PPL lexer/parser with timechart command tokens and syntax rules
* **AST**: Create/update Timechart node to support all parameters
* **Logic**: Implement timechart algorithms including other field
* **Integration**: Full Calcite integration with proper relational algebra generation

### Key Features
* Time alignment supporting `earliest`, `latest`, and specific timestamp values
* Type safety: time-specific parameters only apply to temporal fields
* Flexible parameter combinations for complex binning scenarios
* Locale-independent number formatting
* Backward compatibility with existing PPL commands

### Basic Time Series 
-- Grouped Time-Bucket Aggregation (Unlimited Series)
... | timechart span=1m limit=0 count() as errorCount by url
 
-- Time-based Throughput (No Grouping)
... | timechart count(_time) as tpm span=1m

-- Grouped Time-Bucket Aggregation (Default Series Filtering)
... | timechart span=1m count() as errorCount by url

### Test Coverage Required
 * Unit tests for bin command parsing and AST generation
* Integration tests for end-to-end functionality
* Calcite tests for logical plan generation
* Performance tests with various bin configurations

### Documentation
* `docs/user/ppl/cmd/timechart.rst` - Complete parameter documentation with examples

**Implementation Discussion**

selsong:
### Description
Implement timechart in PPL to accept basic span syntax and aggregation functions. 
TODO: 
- Format table to pivot on the specified field. 
- Incorporate bin command.

### Related Issues
Resolves #3965 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
+++ b/core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.ast.tree;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import org.opensearch.sql.ast.AbstractNodeVisitor;
+import org.opensearch.sql.ast.expression.Literal;
+import org.opensearch.sql.ast.expression.UnresolvedExpression;
+
+/**
+ * AST node represent Timechart operation.
+ */
+@Getter
+@ToString
+@EqualsAndHashCode(callSuper = false)
+@AllArgsConstructor
+public class Timechart extends UnresolvedPlan {
+  private UnresolvedPlan child;
+  private UnresolvedExpression spanExpression;
+  private UnresolvedExpression aggregateFunction;
+  private UnresolvedExpression byField;
+
+  @Override
+  public Timechart attach(UnresolvedPlan child) {
+    return new Timechart(child, spanExpression, aggregateFunction, byField);
+  }
+
+  @Override
+  public List<UnresolvedPlan> getChild() {
+    return ImmutableList.of(child);
+  }
+
+  @Override
+  public <T, C> T accept(AbstractNodeVisitor<T, C> nodeVisitor, C context) {
+    return nodeVisitor.visitTimechart(this, context);
+  }
+}
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -70,7 +70,6 @@ import org.opensearch.sql.ast.expression.Literal;
 import org.opensearch.sql.ast.expression.ParseMethod;
 import org.opensearch.sql.ast.expression.PatternMethod;
 import org.opensearch.sql.ast.expression.PatternMode;
-import org.opensearch.sql.ast.expression.SpanUnit;
 import org.opensearch.sql.ast.expression.UnresolvedExpression;
 import org.opensearch.sql.ast.expression.WindowFrame;
 import org.opensearch.sql.ast.expression.WindowFrame.FrameType;
@@ -1231,28 +1230,26 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     // Build group by list - span first, then by field if present
     List<UnresolvedExpression> groupExprList = new ArrayList<>();
     
-    // Add span expression with @timestamp alias
+    // Add span - use @timestamp field
+    UnresolvedExpression spanExpr;
     if (node.getSpanExpression() != null) {
-      groupExprList.add(new Alias("@timestamp", node.getSpanExpression()));
+      spanExpr = node.getSpanExpression();
     } else {
-      // Default span if none specified
-      UnresolvedExpression defaultSpan = AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1m"), SpanUnit.of("m"));
-      groupExprList.add(new Alias("@timestamp", defaultSpan));
+      // Default to 1 minute span if not specified
+      spanExpr = AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1"), null);
     }
+    groupExprList.add(spanExpr);
     
-    // Add by field if present with proper alias
+    // Add by field if present
     if (node.getByField() != null) {
-      UnresolvedExpression byField = node.getByField();
-      String byAlias = byField instanceof Field ? 
-          ((Field) byField).getField().toString() : byField.toString();
-      groupExprList.add(new Alias(byAlias, byField));
+      groupExprList.add(node.getByField());
     }
     
     // Perform aggregation with trimming
     Pair<List<RexNode>, List<AggCall>> aggregationAttributes =
         aggregateWithTrimming(groupExprList, aggExprList, context);
     
-    // Schema reordering for timechart: aggregation results first, then group by columns
+    // Schema reordering - aggregation results first, then group by columns
     List<RexNode> outputFields = context.relBuilder.fields();
     int numOfOutputFields = outputFields.size();
     int numOfAggList = aggExprList.size();
@@ -1263,41 +1260,36 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
         outputFields.subList(numOfOutputFields - numOfAggList, numOfOutputFields);
     reordered.addAll(aggRexList);
     
-    // Add group by columns (span field first, then by field)
-    List<RexNode> groupByFields = new ArrayList<>();
-    for (int i = 0; i < aggregationAttributes.getLeft().size(); i++) {
-      RexNode groupField = outputFields.get(i);
-      groupByFields.add(groupField);
+    // Add group by columns (timestamp and host)
+    for (int i = 0; i < numOfOutputFields - numOfAggList; i++) {
+      reordered.add(outputFields.get(i));
     }
-    reordered.addAll(groupByFields);
     
-    // Apply proper field names
+    // Create proper field names
     List<String> fieldNames = new ArrayList<>();
     
     // Get aggregation function name
     UnresolvedExpression aggFunc = node.getAggregateFunction();
-    String aggName = "aggr";
-    if (aggFunc instanceof AggregateFunction) {
-      aggName = ((AggregateFunction) aggFunc).getFuncName();
-    } else if (aggFunc instanceof Function) {
-
+    String aggName = "aggr"; // default
+    if (aggFunc instanceof org.opensearch.sql.ast.expression.AggregateFunction) {
+      aggName = ((org.opensearch.sql.ast.expression.AggregateFunction) aggFunc).getFuncName();
     }
     fieldNames.add(aggName);
     
-    fieldNames.add("@timestamp"); // timestamp field
-    
+    // Add by field name first (matches data order)
     if (node.getByField() != null) {
-      String byFieldName = node.getByField() instanceof Field ? 
-          ((Field) node.getByField()).getField().toString() : "host";
+      String byFieldName = node.getByField() instanceof org.opensearch.sql.ast.expression.Field ? 
+          ((org.opensearch.sql.ast.expression.Field) node.getByField()).getField().toString() : "host";
       fieldNames.add(byFieldName);
     }
+    
+    // Add timestamp field name last
+    fieldNames.add("@timestamp");
+    
     context.relBuilder.project(reordered, fieldNames);
     
     // Sort by time (span field) - always first group by field
-    if (!groupByFields.isEmpty()) {
-      RexNode timeField = groupByFields.get(0);
-      context.relBuilder.sort(timeField);
-    }
+    context.relBuilder.sort(context.relBuilder.field(0));
     
     return context.relBuilder.peek();
   }
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -103,6 +103,7 @@ import org.opensearch.sql.ast.tree.Sort;
 import org.opensearch.sql.ast.tree.Sort.SortOption;
 import org.opensearch.sql.ast.tree.SubqueryAlias;
 import org.opensearch.sql.ast.tree.TableFunction;
+import org.opensearch.sql.ast.tree.Timechart;
 import org.opensearch.sql.ast.tree.Trendline;
 import org.opensearch.sql.ast.tree.Trendline.TrendlineType;
 import org.opensearch.sql.ast.tree.UnresolvedPlan;
@@ -1216,6 +1217,68 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
             .collect(Collectors.toList());
     relBuilder.projectPlus(expandedFields);
     return relBuilder.peek();
+  }
+
+  @Override
+  public RelNode visitTimechart(
+      org.opensearch.sql.ast.tree.Timechart node, CalcitePlanContext context) {
+    visitChildren(node, context);
+    
+    // Build aggregation list with the provided aggregate function
+    List<UnresolvedExpression> aggExprList = List.of(node.getAggregateFunction());
+    
+    // Build group by list - span first, then by field if present
+    List<UnresolvedExpression> groupExprList = new ArrayList<>();
+    
+    // Add span - use @timestamp field
+    UnresolvedExpression spanExpr;
+    if (node.getSpanExpression() != null) {
+      spanExpr = node.getSpanExpression();
+    } else {
+      // Default to 1 minute span if not specified
+      spanExpr = AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1m"), null);
+    }
+    groupExprList.add(spanExpr);
+    
+    // Add by field if present
+    if (node.getByField() != null) {
+      groupExprList.add(node.getByField());
+    }
+    
+    // Perform aggregation with trimming
+    Pair<List<RexNode>, List<AggCall>> aggregationAttributes =
+        aggregateWithTrimming(groupExprList, aggExprList, context);
+    
+    // Schema reordering - aggregation results first, then group by columns
+    List<RexNode> outputFields = context.relBuilder.fields();
+    int numOfOutputFields = outputFields.size();
+    int numOfAggList = aggExprList.size();
+    List<RexNode> reordered = new ArrayList<>(numOfOutputFields);
+    
+    // Add aggregation results first
+    List<RexNode> aggRexList =
+        outputFields.subList(numOfOutputFields - numOfAggList, numOfOutputFields);
+    reordered.addAll(aggRexList);
+    
+    // Add group by columns
+    List<RexNode> aliasedGroupByList =
+        aggregationAttributes.getLeft().stream()
+            .map(this::extractAliasLiteral)
+            .flatMap(Optional::stream)
+            .map(ref -> ((RexLiteral) ref).getValueAs(String.class))
+            .map(context.relBuilder::field)
+            .map(f -> (RexNode) f)
+            .toList();
+    reordered.addAll(aliasedGroupByList);
+    context.relBuilder.project(reordered);
+    
+    // Sort by time (span field) - always first group by field
+    if (!aliasedGroupByList.isEmpty()) {
+      RexNode timeField = aliasedGroupByList.get(0);
+      context.relBuilder.sort(timeField);
+    }
+    
+    return context.relBuilder.peek();
   }
 
   @Override
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -70,6 +70,7 @@ import org.opensearch.sql.ast.expression.Literal;
 import org.opensearch.sql.ast.expression.ParseMethod;
 import org.opensearch.sql.ast.expression.PatternMethod;
 import org.opensearch.sql.ast.expression.PatternMode;
+import org.opensearch.sql.ast.expression.SpanUnit;
 import org.opensearch.sql.ast.expression.UnresolvedExpression;
 import org.opensearch.sql.ast.expression.WindowFrame;
 import org.opensearch.sql.ast.expression.WindowFrame.FrameType;
@@ -1221,76 +1222,36 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
 
   @Override
   public RelNode visitTimechart(
-      org.opensearch.sql.ast.tree.Timechart node, CalcitePlanContext context) {
+          org.opensearch.sql.ast.tree.Timechart node, CalcitePlanContext context) {
     visitChildren(node, context);
-    
-    // Build aggregation list with the provided aggregate function
-    List<UnresolvedExpression> aggExprList = List.of(node.getAggregateFunction());
-    
-    // Build group by list - span first, then by field if present
+
+    // Build group by list with proper span expression
     List<UnresolvedExpression> groupExprList = new ArrayList<>();
-    
-    // Add span - use @timestamp field
+
+    // Add span expression for time bucketing (this is essential for timechart)
     UnresolvedExpression spanExpr;
     if (node.getSpanExpression() != null) {
       spanExpr = node.getSpanExpression();
     } else {
       // Default to 1 minute span if not specified
-      spanExpr = AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1"), null);
+      spanExpr = AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1m"), null);
     }
     groupExprList.add(spanExpr);
-    
-    // Add by field if present
+
+    // Add by field if present (for multi-series timechart)
     if (node.getByField() != null) {
       groupExprList.add(node.getByField());
     }
-    
-    // Perform aggregation with trimming
-    Pair<List<RexNode>, List<AggCall>> aggregationAttributes =
-        aggregateWithTrimming(groupExprList, aggExprList, context);
-    
-    // Schema reordering - aggregation results first, then group by columns
-    List<RexNode> outputFields = context.relBuilder.fields();
-    int numOfOutputFields = outputFields.size();
-    int numOfAggList = aggExprList.size();
-    List<RexNode> reordered = new ArrayList<>(numOfOutputFields);
-    
-    // Add aggregation results first
-    List<RexNode> aggRexList =
-        outputFields.subList(numOfOutputFields - numOfAggList, numOfOutputFields);
-    reordered.addAll(aggRexList);
-    
-    // Add group by columns (timestamp and host)
-    for (int i = 0; i < numOfOutputFields - numOfAggList; i++) {
-      reordered.add(outputFields.get(i));
-    }
-    
-    // Create proper field names
-    List<String> fieldNames = new ArrayList<>();
-    
-    // Get aggregation function name
-    UnresolvedExpression aggFunc = node.getAggregateFunction();
-    String aggName = "aggr"; // default
-    if (aggFunc instanceof org.opensearch.sql.ast.expression.AggregateFunction) {
-      aggName = ((org.opensearch.sql.ast.expression.AggregateFunction) aggFunc).getFuncName();
-    }
-    fieldNames.add(aggName);
-    
-    // Add by field name first (matches data order)
-    if (node.getByField() != null) {
-      String byFieldName = node.getByField() instanceof org.opensearch.sql.ast.expression.Field ? 
-          ((org.opensearch.sql.ast.expression.Field) node.getByField()).getField().toString() : "host";
-      fieldNames.add(byFieldName);
-    }
-    
-    // Add timestamp field name last
-    fieldNames.add("@timestamp");
-    
-    context.relBuilder.project(reordered, fieldNames);
-    
-    // Sort by time (span field) - always first group by field
+
+    // Build aggregation list
+    List<UnresolvedExpression> aggExprList = List.of(node.getAggregateFunction());
+
+    // Perform aggregation with time-based grouping
+    aggregateWithTrimming(groupExprList, aggExprList, context);
+
+    // Sort by the first field (which should be the time span)
     context.relBuilder.sort(context.relBuilder.field(0));
-    
+
     return context.relBuilder.peek();
   }
 
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -70,6 +70,7 @@ import org.opensearch.sql.ast.expression.Literal;
 import org.opensearch.sql.ast.expression.ParseMethod;
 import org.opensearch.sql.ast.expression.PatternMethod;
 import org.opensearch.sql.ast.expression.PatternMode;
+import org.opensearch.sql.ast.expression.SpanUnit;
 import org.opensearch.sql.ast.expression.UnresolvedExpression;
 import org.opensearch.sql.ast.expression.WindowFrame;
 import org.opensearch.sql.ast.expression.WindowFrame.FrameType;
@@ -1230,26 +1231,28 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     // Build group by list - span first, then by field if present
     List<UnresolvedExpression> groupExprList = new ArrayList<>();
     
-    // Add span - use @timestamp field
-    UnresolvedExpression spanExpr;
+    // Add span expression with @timestamp alias
     if (node.getSpanExpression() != null) {
-      spanExpr = node.getSpanExpression();
+      groupExprList.add(new Alias("@timestamp", node.getSpanExpression()));
     } else {
-      // Default to 1 minute span if not specified
-      spanExpr = AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1m"), null);
+      // Default span if none specified
+      UnresolvedExpression defaultSpan = AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1m"), SpanUnit.of("m"));
+      groupExprList.add(new Alias("@timestamp", defaultSpan));
     }
-    groupExprList.add(spanExpr);
     
-    // Add by field if present
+    // Add by field if present with proper alias
     if (node.getByField() != null) {
-      groupExprList.add(node.getByField());
+      UnresolvedExpression byField = node.getByField();
+      String byAlias = byField instanceof Field ? 
+          ((Field) byField).getField().toString() : byField.toString();
+      groupExprList.add(new Alias(byAlias, byField));
     }
     
     // Perform aggregation with trimming
     Pair<List<RexNode>, List<AggCall>> aggregationAttributes =
         aggregateWithTrimming(groupExprList, aggExprList, context);
     
-    // Schema reordering - aggregation results first, then group by columns
+    // Schema reordering for timechart: aggregation results first, then group by columns
     List<RexNode> outputFields = context.relBuilder.fields();
     int numOfOutputFields = outputFields.size();
     int numOfAggList = aggExprList.size();
@@ -1260,21 +1263,39 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
         outputFields.subList(numOfOutputFields - numOfAggList, numOfOutputFields);
     reordered.addAll(aggRexList);
     
-    // Add group by columns
-    List<RexNode> aliasedGroupByList =
-        aggregationAttributes.getLeft().stream()
-            .map(this::extractAliasLiteral)
-            .flatMap(Optional::stream)
-            .map(ref -> ((RexLiteral) ref).getValueAs(String.class))
-            .map(context.relBuilder::field)
-            .map(f -> (RexNode) f)
-            .toList();
-    reordered.addAll(aliasedGroupByList);
-    context.relBuilder.project(reordered);
+    // Add group by columns (span field first, then by field)
+    List<RexNode> groupByFields = new ArrayList<>();
+    for (int i = 0; i < aggregationAttributes.getLeft().size(); i++) {
+      RexNode groupField = outputFields.get(i);
+      groupByFields.add(groupField);
+    }
+    reordered.addAll(groupByFields);
+    
+    // Apply proper field names
+    List<String> fieldNames = new ArrayList<>();
+    
+    // Get aggregation function name
+    UnresolvedExpression aggFunc = node.getAggregateFunction();
+    String aggName = "aggr";
+    if (aggFunc instanceof AggregateFunction) {
+      aggName = ((AggregateFunction) aggFunc).getFuncName();
+    } else if (aggFunc instanceof Function) {
+
+    }
+    fieldNames.add(aggName);
+    
+    fieldNames.add("@timestamp"); // timestamp field
+    
+    if (node.getByField() != null) {
+      String byFieldName = node.getByField() instanceof Field ? 
+          ((Field) node.getByField()).getField().toString() : "host";
+      fieldNames.add(byFieldName);
+    }
+    context.relBuilder.project(reordered, fieldNames);
     
     // Sort by time (span field) - always first group by field
-    if (!aliasedGroupByList.isEmpty()) {
-      RexNode timeField = aliasedGroupByList.get(0);
+    if (!groupByFields.isEmpty()) {
+      RexNode timeField = groupByFields.get(0);
       context.relBuilder.sort(timeField);
     }
     
--- a/ppl/src/main/antlr/OpenSearchPPLParser.g4
+++ b/ppl/src/main/antlr/OpenSearchPPLParser.g4
@@ -70,6 +70,7 @@ commands
    | expandCommand
    | flattenCommand
    | reverseCommand
+   | timechartCommand
    ;
 
 commandName
@@ -99,6 +100,7 @@ commandName
    | EXPAND
    | FLATTEN
    | TRENDLINE
+   | TIMECHART
    | EXPLAIN
    | REVERSE
    ;
@@ -145,6 +147,15 @@ sortCommand
 
 reverseCommand
    : REVERSE
+   ;
+
+timechartCommand
+   : TIMECHART (spanClause)? statsFunction (BY fieldExpression)?
+   ;
+
+spanLiteral
+   : integerLiteral timespanUnit
+   | stringLiteral
    ;
 
 evalCommand
--- a/ppl/src/main/antlr/OpenSearchPPLParser.g4
+++ b/ppl/src/main/antlr/OpenSearchPPLParser.g4
@@ -150,7 +150,7 @@ reverseCommand
    ;
 
 timechartCommand
-   : TIMECHART (spanClause)? statsFunction (BY fieldExpression)?
+   : TIMECHART (spanClause | SPAN EQUAL spanLiteral)? statsFunction (BY fieldExpression)?
    ;
 
 spanLiteral
--- a/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
+++ b/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
@@ -396,7 +396,7 @@ public class AstBuilder extends OpenSearchPPLParserBaseVisitor<UnresolvedPlan> {
       // Default span if none specified
       spanExpression = AstDSL.span(
           AstDSL.field("@timestamp"), 
-          AstDSL.stringLiteral("1m"), 
+          AstDSL.intLiteral(1),
           org.opensearch.sql.ast.expression.SpanUnit.of("m"));
     }
     UnresolvedExpression aggregateFunction = internalVisitExpression(ctx.statsFunction());
--- a/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
+++ b/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
@@ -79,6 +79,7 @@ import org.opensearch.sql.ast.tree.Reverse;
 import org.opensearch.sql.ast.tree.Sort;
 import org.opensearch.sql.ast.tree.SubqueryAlias;
 import org.opensearch.sql.ast.tree.TableFunction;
+import org.opensearch.sql.ast.tree.Timechart;
 import org.opensearch.sql.ast.tree.Trendline;
 import org.opensearch.sql.ast.tree.UnresolvedPlan;
 import org.opensearch.sql.ast.tree.Window;
@@ -380,6 +381,19 @@ public class AstBuilder extends OpenSearchPPLParserBaseVisitor<UnresolvedPlan> {
   @Override
   public UnresolvedPlan visitReverseCommand(OpenSearchPPLParser.ReverseCommandContext ctx) {
     return new Reverse();
+  }
+
+  /** Timechart command. */
+  @Override
+  public UnresolvedPlan visitTimechartCommand(OpenSearchPPLParser.TimechartCommandContext ctx) {
+    UnresolvedExpression spanExpression = ctx.spanClause() != null 
+        ? internalVisitExpression(ctx.spanClause()) 
+        : null;
+    UnresolvedExpression aggregateFunction = internalVisitExpression(ctx.statsFunction());
+    UnresolvedExpression byField = ctx.fieldExpression() != null 
+        ? internalVisitExpression(ctx.fieldExpression()) 
+        : null;
+    return new Timechart(null, spanExpression, aggregateFunction, byField);
   }
 
   /** Eval command. */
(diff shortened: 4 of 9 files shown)


core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
```diff
@@ -0,0 +1,61 @@
...
+import org.opensearch.sql.ast.expression.UnresolvedExpression;
+
+/** AST node represent Timechart operation. */
+@Getter
+@ToString
+@EqualsAndHashCode(callSuper = false)
+@AllArgsConstructor
+public class Timechart extends UnresolvedPlan {
+  private UnresolvedPlan child;
+  private UnresolvedExpression spanExpression;
```
dai-chen: This can be generalized to bin option expression? I assume in future this could be span, bins etc.

core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
```diff
@@ -0,0 +1,61 @@
...
+  private Integer limit;
+  private Boolean useOther;
+
+  public Timechart(
+      UnresolvedPlan child,
+      UnresolvedExpression spanExpression,
+      UnresolvedExpression aggregateFunction,
+      UnresolvedExpression byField) {
+    this(child, spanExpression, aggregateFunction, byField, null, true);
+  }
```
dai-chen: I think constructor can only take required arguments, e.g., `new TimeChart(aggFunc).span(...).by(...)`

core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
```diff
@@ -0,0 +1,61 @@
...
+
+/** AST node represent Timechart operation. */
+@Getter
+@ToString
+@EqualsAndHashCode(callSuper = false)
+@AllArgsConstructor
+public class Timechart extends UnresolvedPlan {
+  private UnresolvedPlan child;
+  private UnresolvedExpression spanExpression;
+  private UnresolvedExpression aggregateFunction;
```
dai-chen, line 26: This should be a list?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -147,6 +149,20 @@ reverseCommand
...
+   ;
+
+useOtherValue
+   : booleanLiteral
+   | ident
+   ;
+
+spanLiteral
+   : integerLiteral timespanUnit
+   | stringLiteral
```
dai-chen: I think we can simplify this like `relevanceArg`? Also we can consider extracting common parameters like bin options so timechart and bin command can share.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTimechartCommandIT.java
```diff
@@ -0,0 +1,342 @@
...
+        rows("2024-07-01 00:11:00", null, null, null, null, null, 1, null, null),
+        rows("2024-07-01 00:12:00", null, null, null, null, null, null, null, 1),
+        rows("2024-07-01 00:13:00", null, null, null, null, null, null, 1, null),
+        rows("2024-07-01 00:14:00", null, null, null, null, null, 1, null, null),
+        rows("2024-07-01 00:15:00", null, null, null, null, null, null, null, 1),
+        rows("2024-07-01 00:16:00", null, null, 1, null, null, null, null, null),
+        rows("2024-07-01 00:17:00", null, null, null, 1, null, null, null, null),
+        rows("2024-07-01 00:18:00", 1, null, null, null, null, null, null, null),
+        rows("2024-07-01 00:19:00", null, 1, null, null, null, null, null, null),
+        rows("2024-07-01 00:20:00", null, null, null, null, 1, null, null, null));
```
dai-chen: I feel this is too much to assert. Probably we can make test data smaller?
Same for examples in `timechart.rst `

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -134,6 +138,29 @@ public void testExplainWithReverse() throws IOException {
     assertTrue(result.contains("dir0=[DESC]"));
   }
 
+  @Test
+  public void testExplainWithTimechart() throws IOException {
```
dai-chen: Just curious what the explain output looks like? I'm thinking showing the explain of second query plan is more useful?

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryRequest.java
```diff
@@ -131,6 +137,48 @@ public PPLQueryRequest toPPLQueryRequest() {
     PPLQueryRequest pplQueryRequest = new PPLQueryRequest(pplQuery, jsonContent, path, format);
     pplQueryRequest.sanitize(sanitize);
     pplQueryRequest.style(style);
+
+    // Extract timechart parameters from AST
+    extractTimechartParametersFromAST(pplQueryRequest);
+
```
dai-chen: I think we should avoid touching this class as well as `PPLQueryRequest`. Need to deep dive into your dynamic pivot executor to see where is the best place to implement this two phase execution.

core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
selsong: Currently, this PR only supports a single aggregation function instead of a list that handles multiple for P0 functionality. I can add multiple aggregation function support in later PR.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTimechartCommandIT.java
selsong: Thanks! Agreed, made test data smaller. events_many_hosts.json still has 11 hosts since it is the minimum to test default limit=10. Updated IT and rst doc too.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
selsong: I added explain_timechart.json to show the explain output for logical and physical plan. I updated PR description to show that current approach is OpenSearch native aggregation then post-processing pivot in TimechartResponseFormatter, so there is no 2nd query plan. The explain output only includes aggregation pushdown. 

The OpenSearch single query is:
```
{
  "aggs": {
    "composite": {
      "sources": [
        {"time_bucket": {"date_histogram": {"field": "timestamp", "interval": "1m"}}},
        {"host": {"terms": {"field": "host.keyword"}}}
      ]
    },
    "aggs": {"avg_cpu": {"avg": {"field": "cpu_usage"}}}
  }
}
```
This single query returns:
```
{
  "buckets": [
    {"key": {"time_bucket": "2024-01-01T00:00:00", "host": "web-01"}, "avg_cpu": {"value": 45.2}},
    {"key": {"time_bucket": "2024-01-01T00:00:00", "host": "db-01"}, "avg_cpu": {"value": 78.1}},
    {"key": {"time_bucket": "2024-01-01T00:01:00", "host": "web-01"}, "avg_cpu": {"value": 52.3}}
  ]
}
```

protocol/src/main/java/org/opensearch/sql/protocol/response/format/TimechartResponseFormatter.java
```diff
@@ -0,0 +1,291 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.Singular;
+import org.opensearch.sql.protocol.response.QueryResult;
+
+/**
+ * JSON response formatter specifically for timechart command. It transforms the data from
+ * [timestamp, field, value] format to a pivot table format: [timestamp, field1_value, field2_value,
+ * ...].
+ */
+public class TimechartResponseFormatter extends JsonResponseFormatter<QueryResult> {
```
ykmr1224: Why do we need specific formatter for Timechart? Isn't it just a subset of Json?

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java
```diff
@@ -156,7 +157,16 @@ private ResponseListener<ExecutionEngine.QueryResponse> createListener(
     } else if (format.equals(Format.VIZ)) {
       formatter = new VisualizationResponseFormatter(pplRequest.style());
     } else {
-      formatter = new SimpleJsonResponseFormatter(JsonResponseFormatter.Style.PRETTY);
+      // Check if this is a timechart query
+      if (pplRequest.getRequest().toLowerCase().contains("timechart")) {
```
ykmr1224: nit: I think we don't need to nest the if block. (let's just do like `} else if (pplRequest... ) {`)

ykmr1224: Would it mis-judge when query somehow contains string `timechart`?

core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
```diff
@@ -0,0 +1,64 @@
...
+  private UnresolvedExpression byField;
+  private Integer limit;
+  private Boolean useOther;
+
+  public Timechart(UnresolvedPlan child, UnresolvedExpression aggregateFunction) {
+    this(child, null, aggregateFunction, null, null, true);
+  }
+
+  public Timechart span(UnresolvedExpression binExpression) {
+    return new Timechart(child, binExpression, aggregateFunction, byField, limit, useOther);
```
ykmr1224: This looks weird to me. Should we reuse `this` and just set span? Same for the following methods.
It might be better using `@Builder` lombok annotation.

protocol/src/main/java/org/opensearch/sql/protocol/response/format/TimechartResponseFormatter.java
```diff
@@ -0,0 +1,291 @@
...
+
+    json.total(response.size()).size(response.size());
+
+    response.columnNameTypes().forEach((name, type) -> json.column(new Column(name, type)));
+
+    json.datarows(fetchDataRows(response));
+    return json.build();
+  }
+
+  private Object buildPivotJsonObject(QueryResult response, List<String> columnNames) {
```
Swiddis: This is a long function (136 lines), consider breaking it into simpler components next time. Helps with testing + reasoning through how it works.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1218,6 +1221,79 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
...
+    context.relBuilder.distinct();
+    RelNode distinctQuery = context.relBuilder.peek();
+
+    // Restore the original state
+    context.relBuilder.push(currentState);
+
+    // Store the necessary information for the dynamic pivot operation
+    // This will be used by the execution engine to perform the two-phase approach
+    Map<String, String> dynamicPivotInfo = new HashMap<>();
+    dynamicPivotInfo.put("timeField", timeField);
```
ykmr1224: Should we add constants (might be enum) for "timeField", "byField", "valueField", "limit", and "useOther"?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1218,6 +1221,79 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
     return relBuilder.peek();
   }
 
+  @Override
+  public RelNode visitTimechart(
+      org.opensearch.sql.ast.tree.Timechart node, CalcitePlanContext context) {
```
Swiddis, line 1410: This method's also pretty long & it's hard to parse what the high-level approach is. Effectively need to read all lines to understand what it's doing.

I'm a big fan of [Ousterhout's commenting notes](https://web.stanford.edu/~ouster/cgi-bin/cs190-spring16/lecture.php?topic=comments).

selsong: Thanks, added high level comment.

protocol/src/main/java/org/opensearch/sql/protocol/response/format/TimechartResponseFormatter.java
selsong: Thanks, refactored into simpler components.

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryRequest.java
```diff
@@ -131,6 +137,45 @@ public PPLQueryRequest toPPLQueryRequest() {
     PPLQueryRequest pplQueryRequest = new PPLQueryRequest(pplQuery, jsonContent, path, format);
     pplQueryRequest.sanitize(sanitize);
     pplQueryRequest.style(style);
+    extractTimechartParameters(pplQueryRequest);
     return pplQueryRequest;
   }
+
+  private void extractTimechartParameters(PPLQueryRequest pplQueryRequest) {
```
dai-chen: Since there are follow up work, I wonder can we avoid such changes. Can we simply assume the second column is always the one to pivot? I think we can merge this for now if we can avoid touching these PPL classes.

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+Version
+=======
+3.3.0
+
+Syntax
+============
+timechart [span=<time_interval>] [limit=<number>] [useother=<boolean>] <aggregation_function> [by <field>]
+
+* span: optional. Specifies the time interval for grouping data.
+  * Default: 1m (1 minute)
```
penghuo, line 31: Doc does not format correcttly. The Default should in newline? https://github.com/selsong/sql/blob/f3d987f46ec42189b1484abf3a56a61cb60ab500/docs/user/ppl/cmd/timechart.rst

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+    * week (w)
+    * month (M, case sensitive)
+    * quarter (q)
+    * year (y)
+
+* limit: optional. Specifies the maximum number of distinct values to display when using the "by" clause.
+  * Default: 10
+  * When there are more distinct values than the limit, the additional values are grouped into an "OTHER" category.
+  * Set to 0 to show all distinct values without any limit.
+  * Position: Should be specified after the span parameter and before the aggregation function if both are used.
```
penghuo: Why restrict on position? Is there impl limitation?

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+
+* useother: optional. Controls whether to create an "OTHER" category for values beyond the limit.
+  * Default: true
+  * When set to false, only the top N values (based on limit) are shown without an "OTHER" column.
+  * When set to true, values beyond the limit are grouped into an "OTHER" category.
+  * Only applies when using the "by" clause and when there are more distinct values than the limit.
+
+* aggregation_function: mandatory. The aggregation function to apply to each time bucket.
+  * Currently, only a single aggregation function is supported.
+  * Available functions: count(), avg(), sum(), min(), max()
```
penghuo: We only support 5?, instead of all [stats function](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/stats.rst#aggregation-functions).

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java
selsong: Thanks. You're right I think the check can be more robust. I added more robust detection.

protocol/src/main/java/org/opensearch/sql/protocol/response/format/TimechartResponseFormatter.java
selsong: The TimechartResponseFormatter transforms data from [timestamp, field, value] format to **pivot table format** [timestamp, field1_value, field2_value, ...] for visualization compatibility. It handles timechart-specific operations like data pivoting, limit enforcement with "OTHER" grouping, and score-based column sorting that the standard JSON formatter doesn't support. This is approach 3 chosen for now, with more discussion on the Dynamic Pivot Implementation Approach on the RFC #3965

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+This example counts events for each hour and groups them by host.
+
+PPL query::
+
+    source=events | timechart span=1h count() by host
+
+Result::
+
+    +---------------------+-------+--------+--------+
+    | $f2                 | db-01 | web-01 | web-02 |
```
penghuo: why column name is $f2, it should be `@timestamp`?

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+    * hour (h)
+    * day (d)
+    * week (w)
+    * month (M, case sensitive)
+    * quarter (q)
+    * year (y)
+
+* limit: optional. Specifies the maximum number of distinct values to display when using the "by" clause.
+  * Default: 10
+  * When there are more distinct values than the limit, the additional values are grouped into an "OTHER" category.
```
penghuo: add explaintaion of `most distinct`

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+PPL query::
+
+    source=events | timechart span=1m count() by host
+
+Result::
+
+    +---------------------+-------+--------+--------+
+    | $f2                 | db-01 | web-01 | web-02 |
+    +---------------------+-------+--------+--------+
+    | 2024-07-01 00:00:00 | null  | 1      | null   |
```
penghuo: count result should be 0, instead of null?

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+    +---------------------+-------+--------+--------+
+
+Example 3: Calculate average CPU usage by minute
+==============================================
+
+This example calculates the average CPU usage for each minute without grouping by any field.
+
+PPL query::
+
+    source=events | timechart span=1m avg(cpu_usage)
```
penghuo: what is result?

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+    | 2024-07-01 00:02:00 | null    | 1       | null    |
+    | 2024-07-01 00:03:00 | 1       | null    | null    |
+    | 2024-07-01 00:04:00 | null    | null    | 1       |
+    +---------------------+---------+---------+---------+
+
+Example 5: Using the limit parameter
+==================================
+
+When there are many distinct values in the "by" field, the timechart command will display the top values based on the limit parameter and group the rest into an "OTHER" category.
+This query will display the top 2 hosts with the highest CPU usage values, and group the remaining hosts into an "OTHER" category.
```
penghuo: This query will display the top 2 hosts with the highest CPU usage values -> 
This query will display the top 2 hosts with the highest sum of avg CPU usage values?

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,224 @@
...
+Note: The limit parameter must be specified after the span parameter. The following syntax is correct::
+
+    source=events | timechart span=1m limit=2 avg(cpu_usage) by host
+
+Result::
+
+    +---------------------+--------+--------+-------+
```
penghuo: does example 5 and 6 using same dataset? if yes, in example 5, web-01, avg(cpu)=55.03, but in example 6, web-2, avg(cpu)=55.03, seems not correct?

docs/user/ppl/cmd/timechart.rst
selsong: We support all stats functions. Fixed documentation to match.

docs/user/ppl/cmd/timechart.rst
selsong: 2nd one, highest sum of avg, updated.

docs/user/ppl/cmd/timechart.rst
selsong: Updated doc. Examples 5 and 6 use different datasets: Example 5 uses the ``events`` dataset with fewer hosts for simplicity, while Example 6 uses the ``events_many_hosts`` dataset with 11 distinct hosts.

docs/user/ppl/cmd/timechart.rst
selsong: Updated

docs/user/ppl/cmd/timechart.rst
selsong: Updated, Supports all positions now.

docs/user/ppl/cmd/timechart.rst
selsong: Updated.

docs/user/ppl/cmd/timechart.rst
selsong: Updated

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryRequest.java
selsong: Moved parameter extraction to PPLService file

docs/user/ppl/cmd/timechart.rst
selsong: Fixed.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTimechartCommandIT.java
```diff
@@ -0,0 +1,279 @@
...
+    enableCalcite();
+    disallowCalciteFallback();
+
+    // Create events index with timestamp data
+    String eventsMapping =
+        "{\"mappings\":{\"properties\":{\"@timestamp\":{\"type\":\"date\"},\"host\":{\"type\":\"keyword\"},\"service\":{\"type\":\"keyword\"},\"response_time\":{\"type\":\"integer\"},\"status_code\":{\"type\":\"integer\"},\"bytes_sent\":{\"type\":\"long\"},\"cpu_usage\":{\"type\":\"double\"},\"memory_usage\":{\"type\":\"double\"},\"region\":{\"type\":\"keyword\"},\"environment\":{\"type\":\"keyword\"}}}}";
+    if (!isIndexExist(client(), "events")) {
+      createIndexByRestClient(client(), "events", eventsMapping);
+      loadDataByRestClient(client(), "events", "src/test/resources/events_test.json");
+    }
```
penghuo: You can use loadIndex() method. e.g. in CalciteArrayFunctionIT.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTimechartCommandIT.java
```diff
@@ -0,0 +1,279 @@
...
+    // Create index without @timestamp field
+    String noTimestampMapping =
+        "{\"mappings\":{\"properties\":{\"name\":{\"type\":\"keyword\"},\"occupation\":{\"type\":\"keyword\"},\"country\":{\"type\":\"keyword\"},\"salary\":{\"type\":\"integer\"},\"year\":{\"type\":\"integer\"},\"month\":{\"type\":\"integer\"}}}}";
+    if (!isIndexExist(client(), "no_timestamp")) {
+      createIndexByRestClient(client(), "no_timestamp", noTimestampMapping);
+      loadDataByRestClient(client(), "no_timestamp", "src/test/resources/occupation.json");
+    }
+
+    // Test should throw exception for missing @timestamp field
+    assertThrowsWithReplace(
```
penghuo: Add Assert error message

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTimechartCommandIT.java
```diff
@@ -0,0 +1,279 @@
...
+        () -> {
+          executeQuery("source=no_timestamp | timechart count()");
+        });
+  }
+
+  @Test
+  public void testTimechartWithMinuteSpanNoGroupBy() throws IOException {
+    JSONObject result = executeQuery("source=events | timechart span=1m avg(cpu_usage)");
+    verifySchema(result, schema("@timestamp", "timestamp"), schema("$f1", "double"));
+    assertEquals(5, result.getInt("total"));
```
penghuo, line 114: Add verifyDataRows

ppl/src/main/java/org/opensearch/sql/ppl/PPLService.java
```diff
@@ -107,4 +107,154 @@ private AbstractPlan plan(
...
+   * @param queryListener {@link ResponseListener} for query response
+   * @param explainListener {@link ResponseListener} for explain response
+   * @return {@link AbstractPlan} execution plan
+   */
+  public AbstractPlan createPlan(
+      PPLQueryRequest request,
+      ResponseListener<QueryResponse> queryListener,
+      ResponseListener<ExplainResponse> explainListener) {
+    return plan(request, queryListener, explainListener);
+  }
```
penghuo: nit: createPlan is redanudent , make plan function public for testting purpose.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -134,6 +138,22 @@ public void testExplainWithReverse() throws IOException {
...
+    // Create events index with timestamp data for timechart test
+    String eventsMapping = TestUtils.getMappingFile("events_index_mapping.json");
+    if (!isIndexExist(client(), "events")) {
+      createIndexByRestClient(client(), "events", eventsMapping);
+      loadDataByRestClient(client(), "events", "src/test/resources/events_test.json");
+    }
```
penghuo: Use loadIndex()

integ-test/src/test/resources/expectedOutput/calcite/explain_timechart.json
```diff
@@ -0,0 +1,6 @@
+{
+  "calcite": {
+    "logical": "LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n  LogicalProject($f3=[$1], host=[$0], $f2=[$2])\n    LogicalAggregate(group=[{0, 2}], agg#0=[AVG($1)])\n      LogicalProject(host=[$0], cpu_usage=[$2], $f3=[SPAN($1, 1, 'm')])\n        CalciteLogicalIndexScan(table=[[OpenSearch, events]])\n",
+    "physical": "EnumerableLimit(fetch=[10000])\n  EnumerableCalc(expr#0..2=[{inputs}], $f3=[$t1], host=[$t0], $f2=[$t2])\n    CalciteEnumerableIndexScan(table=[[OpenSearch, events]], PushDownContext=[[AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0, 2},agg#0=AVG($1))], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"size\":0,\"timeout\":\"1m\",\"aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"sources\":[{\"host\":{\"terms\":{\"field\":\"host\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\"}}},{\"$f3\":{\"date_histogram\":{\"field\":\"@timestamp\",\"missing_bucket\":true,\"missing_order\":\"first\",\"order\":\"asc\",\"fixed_interval\":\"1m\"}}}]},\"aggregations\":{\"$f2\":{\"avg\":{\"field\":\"cpu_usage\"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
```
penghuo: Formatter should process results after `| timechart limit=x`, currently, it only process 10K results at max.

ppl/src/main/java/org/opensearch/sql/ppl/PPLService.java
selsong: removed, unused code

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1331,6 +1334,237 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
...
+      List<UnresolvedExpression> args = new ArrayList<>();
+      if (aggFunc.getField() != null) {
+        args.add(aggFunc.getField());
+      }
+      if (aggFunc.getArgList() != null) {
+        args.addAll(aggFunc.getArgList());
+      }
+
+      if (args.isEmpty() || funcName.equals("count")) {
+        // Special case for count() to show as just "count" instead of "count(AllFields())"
```
Swiddis: Interesting that this only affects count and not any other aggregations? Guess it's that the aggregations are checked earlier, but then why is `count` not among them?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1331,6 +1334,237 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
...
+            sb.append(((Field) args.get(i)).getField().toString());
+          } else {
+            sb.append(args.get(i).toString());
+          }
+        }
+        sb.append(")");
+        return sb.toString();
+      }
+    } else {
+      return "value";
```
Swiddis: As a style point, this should be a guard clause. "if (!aggregateFunction instanceof AggregateFunction) return "value";`, then you don't need `else`. Makes the method easier to read. (Same for the `isEmpty` check above)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
selsong: count() internally uses AllFields() as its argument, while other aggregations like avg(field) have explicit fields that should be shown. We only need to hide the internal AllFields() implementation detail for count and show just count instead of count(AllFields())

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
selsong: Thanks! Fixed.

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,253 @@
...
+
+  * Default: true
+  * When set to false, only the top N values (based on limit) are shown without an "OTHER" column.
+  * When set to true, values beyond the limit are grouped into an "OTHER" category.
+  * Only applies when using the "by" clause and when there are more distinct values than the limit.
+
+* **aggregation_function**: mandatory. The aggregation function to apply to each time bucket.
+
+  * Currently, only a single aggregation function is supported.
+  * Available functions: All standard aggregation functions supported by stats are supported.
```
penghuo: All aggregation functions supported by stats command are supported.

And add hyperlink of stats command doc.

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,253 @@
...
+    * week (w)
+    * month (M, case sensitive)
+    * quarter (q)
+    * year (y)
+
+* **limit**: optional. Specifies the maximum number of distinct values to display when using the "by" clause.
+
+  * Default: 10
+  * When there are more distinct values than the limit, the additional values are grouped into an "OTHER" category.
+  * The "most distinct" values are determined by calculating the sum of the aggregation values across all time intervals for each distinct field value. The top N values with the highest sums are displayed individually, while the rest are grouped into the "OTHER" category.
```
penghuo, line 48: the rest are grouped into OTHER category if userOther not equal to false.

docs/user/ppl/cmd/timechart.rst
```diff
@@ -0,0 +1,253 @@
...
+  * Set to 0 to show all distinct values without any limit.
```
penghuo: if limit=0, userother is false, right?

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
```diff
@@ -419,6 +420,11 @@ public String visitReverse(Reverse node, String context) {
     return StringUtils.format("%s | reverse", child);
   }
 
+  @Override
+  public String visitTimechart(Timechart node, String context) {
+    return "source=t | timechart count() by host";
```
penghuo: why always by host? use visitHead as example.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -165,6 +167,25 @@ reverseCommand
...
+   ;
+
+timechartParameter
+   : (spanClause | SPAN EQUAL spanLiteral)
+   | timechartArg
+   ;
+
+timechartArg
+   : LIMIT EQUAL integerLiteral
+   | USEOTHER EQUAL (booleanLiteral | ident)
```
penghuo, line 184: nit: userother="abc" also valid? support booleanLiteral should be enough?

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
```diff
@@ -205,7 +205,8 @@ public void execute(
                     () -> {
                       try (PreparedStatement statement = OpenSearchRelRunners.run(context, rel)) {
                         ResultSet result = statement.executeQuery();
-                        buildResultSet(result, rel.getRowType(), context.querySizeLimit, listener);
+                        buildResultSet(
+                            result, rel.getRowType(), context.querySizeLimit, context, listener);
```
penghuo: All change in this file can be reverted?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -571,6 +573,57 @@ public UnresolvedPlan visitReverseCommand(OpenSearchPPLParser.ReverseCommandCont
     return new Reverse();
   }
 
+  /** Timechart command. */
+  @Override
+  public UnresolvedPlan visitTimechartCommand(OpenSearchPPLParser.TimechartCommandContext ctx) {
+    UnresolvedExpression binExpression = null;
+    // Extract limit parameter if present
+    Integer limit = null;
```
penghuo: default limit=10

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -571,6 +573,57 @@ public UnresolvedPlan visitReverseCommand(OpenSearchPPLParser.ReverseCommandCont
...
+          }
+        }
+      }
+    }
+
+    // If no span was specified, use the default
+    if (binExpression == null) {
+      binExpression =
+          AstDSL.span(AstDSL.field("@timestamp"), AstDSL.intLiteral(1), SpanUnit.of("m"));
+    }
```
penghuo: nit, binExpression = AstDSL.span(AstDSL.field("@timestamp"), AstDSL.intLiteral(1), SpanUnit.of("m")); could be default value of binExpression on L579.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -571,6 +573,57 @@ public UnresolvedPlan visitReverseCommand(OpenSearchPPLParser.ReverseCommandCont
...
+
+    UnresolvedExpression aggregateFunction = internalVisitExpression(ctx.statsFunction());
+    UnresolvedExpression byField =
+        ctx.fieldExpression() != null ? internalVisitExpression(ctx.fieldExpression()) : null;
+
+    // The limit and useOther parameters are stored in the Timechart AST node
+    // and will be used by the TimechartResponseFormatter
```
penghuo: Remove this comments, it is not reflect latest impl.

ppl/src/main/antlr/OpenSearchPPLParser.g4
selsong: useother=t and useother=f are also valid.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLTimechartTest.java
```diff
@@ -0,0 +1,399 @@
...
+
+import static org.junit.Assert.assertNotNull;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+import org.apache.calcite.DataContext;
+import org.apache.calcite.config.CalciteConnectionConfig;
+import org.apache.calcite.linq4j.Enumerable;
+import org.apache.calcite.linq4j.Linq4j;
```
penghuo, line 16: Add test to translate PPL to SQL. follow CalcitePPLAggregationTest

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -571,6 +573,57 @@ public UnresolvedPlan visitReverseCommand(OpenSearchPPLParser.ReverseCommandCont
...
+      } else if (paramCtx.timechartArg() != null) {
+        OpenSearchPPLParser.TimechartArgContext argCtx = paramCtx.timechartArg();
+        if (argCtx.LIMIT() != null && argCtx.integerLiteral() != null) {
+          limit = Integer.parseInt(argCtx.integerLiteral().getText());
+        } else if (argCtx.USEOTHER() != null) {
+          if (argCtx.booleanLiteral() != null) {
+            useOther = Boolean.parseBoolean(argCtx.booleanLiteral().getText());
+          } else if (argCtx.ident() != null) {
+            String useOtherValue = argCtx.ident().getText().toLowerCase();
+            useOther = "true".equals(useOtherValue) || "t".equals(useOtherValue);
```
penghuo: we should fail fast if there is typo, e.g. useOther="treu"

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -571,6 +573,57 @@ public UnresolvedPlan visitReverseCommand(OpenSearchPPLParser.ReverseCommandCont
...
+    for (OpenSearchPPLParser.TimechartParameterContext paramCtx : ctx.timechartParameter()) {
+      if (paramCtx.spanClause() != null) {
+        binExpression = internalVisitExpression(paramCtx.spanClause());
+      } else if (paramCtx.spanLiteral() != null) {
+        // Convert span=1h to span(@timestamp, 1h)
+        binExpression = internalVisitExpression(paramCtx.spanLiteral());
+      } else if (paramCtx.timechartArg() != null) {
+        OpenSearchPPLParser.TimechartArgContext argCtx = paramCtx.timechartArg();
+        if (argCtx.LIMIT() != null && argCtx.integerLiteral() != null) {
+          limit = Integer.parseInt(argCtx.integerLiteral().getText());
```
penghuo, line 591: validate limit is non negative number.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1331,6 +1334,408 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
...
+    UnresolvedExpression spanExpr =
+        node.getBinExpression() != null
+            ? node.getBinExpression()
+            : AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1m"), null);
+
+    List<UnresolvedExpression> groupExprList = Arrays.asList(spanExpr);
+
+    // Handle no by field case
+    if (node.getByField() == null) {
+      String valueFunctionName = getValueFunctionName(node.getAggregateFunction());
```
penghuo, line 1420: How stats command handle name? we should reuse it.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1331,6 +1334,408 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
...
+    // Extract parameters
+    UnresolvedExpression spanExpr =
+        node.getBinExpression() != null
+            ? node.getBinExpression()
+            : AstDSL.span(AstDSL.field("@timestamp"), AstDSL.stringLiteral("1m"), null);
+
+    List<UnresolvedExpression> groupExprList = Arrays.asList(spanExpr);
+
+    // Handle no by field case
+    if (node.getByField() == null) {
```
penghuo, line 1419: if is no by field, it is as same as `stats agg span`, right? so we can reuse all code in visitStats()? or In AST, we can simply no by field case as Stats?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1331,6 +1334,408 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
...
+  }
+
+  /** Transforms timechart command into SQL-based operations. */
+  @Override
+  public RelNode visitTimechart(
+      org.opensearch.sql.ast.tree.Timechart node, CalcitePlanContext context) {
+    visitChildren(node, context);
+
+    // Extract parameters
+    UnresolvedExpression spanExpr =
```
penghuo: AST already did it? Bin Expression is non-null value

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1331,6 +1334,408 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
...
+        // Detect field indices first
+        FieldIndices indices =
+            detectFieldIndices(context.relBuilder.peek().getRowType().getFieldNames(), byFieldName);
+
+        // Project to ensure correct field order: [@timestamp, host, value]
+        context.relBuilder.project(
+            context.relBuilder.alias(context.relBuilder.field(indices.timeIndex), "@timestamp"),
+            context.relBuilder.alias(context.relBuilder.field(indices.byIndex), byFieldName),
+            context.relBuilder.alias(
+                context.relBuilder.field(indices.valueIndex), valueFunctionName));
```
penghuo: @LantaoJin  Is there better way do do this? add alias in aggregation builder?

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
selsong: Yes, reverted changes to this file. Now using visitAggregation for the case of 1 aggregation function no group by field.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLTimechartTest.java
selsong: Done

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
```diff
@@ -56,6 +56,9 @@ public class CalcitePlanContext {
 
   @Getter public Map<String, RexLambdaRef> rexLambdaRefMap;
 
+  /** Store original field names for timechart queries to preserve aggregation function names */
+  @Getter @Setter private Map<Integer, String> originalFieldNames = new HashMap<>();
```
qianheng-aws: Seems never used. Please check and remove it.

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -54,6 +54,14 @@ public interface PlanUtils {
...
   String ROW_NUMBER_COLUMN_NAME_SUBSEARCH = "_row_number_subsearch_";
 
+  // Timechart constants
+  String TIMECHART_TIME_FIELD = "timeField";
+  String TIMECHART_BY_FIELD = "byField";
+  String TIMECHART_VALUE_FIELD = "valueField";
+  String TIMECHART_LIMIT = "limit";
+  String TIMECHART_USE_OTHER = "useOther";
+  String TIMECHART_OTHER_CATEGORY = "OTHER";
+
```
qianheng-aws: Used nowhere.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
qianheng-aws: https://github.com/opensearch-project/sql/blob/98c1b6aa5a4f4c16ddd63e3eef395751e01f55e0/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java#L612-L617

For stats command, the name is generated directly from the query text when building AST. Maybe we should do similar thing in `visitTimechartCommand`.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -216,6 +218,16 @@ public void testExplainWithReverse() throws IOException {
     assertTrue(result.contains("dir0=[DESC]"));
   }
 
+  @Test
+  public void testExplainWithTimechart() throws IOException {
+    var result = explainQueryToString("source=events | timechart span=1m avg(cpu_usage) by host");
```
qianheng-aws, line 223: > // Use zero-filling for count aggregations, standard result for others

Please add explain test case for count aggregation as it gets a totally different plan actually.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1365,6 +1369,330 @@ public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
...
+
+    try {
+      // Step 1: Initial aggregation - IMPORTANT: order is [spanExpr, byField]
+      groupExprList = Arrays.asList(spanExpr, byField);
+      aggregateWithTrimming(groupExprList, List.of(node.getAggregateFunction()), context);
+
+      // First rename the timestamp field (2nd to last) to @timestamp
+      List<String> fieldNames = context.relBuilder.peek().getRowType().getFieldNames();
+      List<String> renamedFields = new ArrayList<>(fieldNames);
+      renamedFields.set(fieldNames.size() - 2, "@timestamp");
```
qianheng-aws, line 1468: The reason why we find that there is `flip` for group by keys is because of the registry mechanism for aggregate key in Calcite.

https://github.com/apache/calcite/blob/41e1ab8b8c18b186730b30cd2cbebffe807d86f3/core/src/main/java/org/apache/calcite/tools/RelBuilder.java#L2509-L2519

It's actually not `flip` but appending new expressions(i.e. span(...) here) to the original expressions. So although we constructed group by expr to be `[spanExpr, byField]`, but got `[byField, spanExpr]` in the end.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLTimechartTest.java
```diff
@@ -0,0 +1,417 @@
...
+            + "ORDER BY 1 NULLS LAST";
+    verifyPPLToSparkSQL(root, expectedSparkSql);
+  }
+
+  @Test
+  public void testTimechartWithBy() {
+    String ppl = "source=events | timechart count() by host";
+
+    RelNode root = getRelNode(ppl);
+    String expectedSparkSql =
```
penghuo: ~~SQL query does not reflect actual logical plan, Let's ignore it, and create a issue to root cause.~~ It is calcite relnode to sql issue, column should be nullable.

qianheng-aws (approved):
The functionality of this PR LGTM. Thanks for your contribution! @selsong

PR #4001, opened 2025-08-07
Allow equal expression as a function argument

### LINKED ISSUE #3992: [BUG] IF function contain equal expression return ExpressionEvaluationException
**What is the bug?**
* `=` expression works as expected in eval
```
source=logs-otel | parse body 'HTTP/1.1 (?<httpstatus>\\d+)' | eval status200=httpstatus='200' | fields httpstatus, status200
```
* `=` throw expression when used inside if expression
```
source=logs-otel | parse body 'HTTP/1.1 (?<httpstatus>\\d+)' | eval status200=if(httpstatus='200', 1, 0) | fields httpstatus, status200

{
  "error": {
    "reason": "Invalid Query",
    "details": "IF function expects {[BOOLEAN,ANY,ANY]}, but got [STRING,INTEGER,INTEGER]",
  },
  "status": 400
}
```

No exception.

n/a

```
###
PUT {{baseUrl}}/_plugins/_query/settings

{
  "transient" : {
    "plugins.calcite.enabled" : true
  }
}


###
POST {{baseUrl}}/logs-otel/_bulk

{"index": {}}
{"regionId": 1, "body": "HTTP/1.1 200 UT response_timeout - \"-\" 380 24 15000"}
{"index": {}}
{"regionId": 1, "body": "HTTP/1.1 301 UT response_timeout - \"-\" 380 24 15000"}
{"index": {}}
{"regionId": 2, "body": "HTTP/1.1 404 UT response_timeout - \"-\" 380 24 15000"}
{"index": {}}
{"regionId": 2, "body": "HTTP/1.1 504 UT response_timeout - \"-\" 380 24 15000"}
{"index": {}}
{"regionId": 2, "body": "HTTP/1.1 900 UT response_timeout - \"-\" 380 24 15000"}
```

yuancu:
### Description

This PR allows comparison expressions with equal as a function argument.

Previously, arguments in expressions like `eval res = foo(param1=1, param2='b')` would be treated as named arguments. They would match the following rule:

```antlr
functionArg
   : (ident EQUAL)? functionArgExpression
   ;
```

However, this prevented users from passing comparison expressions like `filed=target` as a parameter, which are widely used in condition functions like `IF`, `CASE`, etc.

I found that named parameters are used exclusively in table functions. This PR fixes the problem by eliminating name parameters in other functions.

### Related Issues
Resolves #3992

### Check List

### DIFF
--- a/ppl/src/main/antlr/OpenSearchPPLParser.g4
+++ b/ppl/src/main/antlr/OpenSearchPPLParser.g4
@@ -518,7 +518,7 @@ tableSource
    ;
 
 tableFunction
-   : qualifiedName LT_PRTHS functionArgs RT_PRTHS
+   : qualifiedName LT_PRTHS namedFunctionArgs RT_PRTHS
    ;
 
 // fields
@@ -593,10 +593,17 @@ functionArgs
    : (functionArg (COMMA functionArg)*)?
    ;
 
-functionArg
-   : (ident EQUAL)? functionArgExpression
+namedFunctionArgs
+   : (namedFunctionArg (COMMA namedFunctionArg)*)?
    ;
 
+functionArg
+   : functionArgExpression
+   ;
+
+namedFunctionArg
+   : (ident EQUAL)? functionArgExpression
+   ;
 
 functionArgExpression
    : lambda
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLConditionBuiltinFunctionIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLConditionBuiltinFunctionIT.java
@@ -195,6 +195,19 @@ public class CalcitePPLConditionBuiltinFunctionIT extends PPLIntegTestCase {
   }
 
   @Test
+  public void testIfWithEquals() throws IOException {
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | eval jake = if(name='Jake', 1, 0) | fields name, jake",
+                TEST_INDEX_STATE_COUNTRY));
+
+    verifySchema(actual, schema("name", "string"), schema("jake", "int"));
+
+    verifyDataRows(actual, rows("Jake", 1), rows("Hello", 0), rows("John", 0), rows("Jane", 0));
+  }
+
+  @Test
   public void testIsPresent() throws IOException {
     JSONObject actual =
         executeQuery(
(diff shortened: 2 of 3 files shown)


ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -593,10 +593,17 @@ functionArgs
...
+   ;
+
 functionArg
-   : (ident EQUAL)? functionArgExpression
+   : functionArgExpression
    ;
 
+namedFunctionArg
+   : (ident EQUAL)? functionArgExpression
+   ;
```
dai-chen, line 606: If users name the argument in the meanwhile, it will be `if(isNameJake=name='Jake' ...)` ?

yuancu: After the change, only table functions support named arguments. `IF` is not one of them. `if(isNameJake=name='Jake' ...)` will not be a valid expression. For table functions, based on the following rules:

```antlr
tableFunction
   : qualifiedName LT_PRTHS namedFunctionArgs RT_PRTHS
   ;

namedFunctionArg
   : (ident EQUAL)? functionArgExpression
   ;

functionArgExpression
   : lambda
   | logicalExpression
   ;

logicalExpression
   : NOT logicalExpression                                      # logicalNot
   | left = logicalExpression AND right = logicalExpression     # logicalAnd
   | left = logicalExpression XOR right = logicalExpression     # logicalXor
   | left = logicalExpression OR right = logicalExpression      # logicalOr
   | expression                                                 # logicalExpr
   ;
```

I think you are right.

BTW, I haven't found any official definitions for table functions in OpenSearch, although they present in the grammar file. Do you know some examples of them?

dai-chen: I see. Could you create follow up issue if so? To support both, we may need breaking change right?
I recall table function was related to PromQL support. Ref: https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/admin/connectors/prometheus_connector.rst#query_range-table-function

dai-chen (approved):

ppl/src/main/antlr/OpenSearchPPLParser.g4
yuancu: Thank you for the source! Do you mean to support named arguments for all functions by *to support both*? I am a little confused in why should we do so. Afaik, named arguments isn't mentioned anywhere in PPL specifications. And we had no plan to support out-of-order arguments even for functions with named arguments. Moreover, it's hard to tell whether `if(a=bool_column, ...)` is a named argument or a logical expression if we allow named arguments for all functions.

dai-chen: Sorry for confusion. I asked because I saw `if(predicate:code=200` in SPL doc: https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-manual/functions/naming-function-arguments. Never mind if this is not planned.

yuancu:
@qianheng-aws Can you help merge this PR? Thanks!

ppl/src/main/antlr/OpenSearchPPLParser.g4
yuancu: Thanks for the explanation. That's a specific grammar. They use `:` to avoid the dilemma of telling apart logical comparison and assignment,

yuancu:
@LantaoJin Should this PR be backported?

PR #4006, opened 2025-08-07
fix snapshot uploading

ahkcs:
Update snapshot uploading

I changed the path to fetch the publish-snapshot.sh(https://github.com/opensearch-project/opensearch-build/blob/main/publish/publish-snapshot.sh) from opensearch-build-libraries to opensearch-build since this file no longer exists in opensearch-build-libraries

### DIFF
--- a/.github/workflows/publish-async-query-core.yml
+++ b/.github/workflows/publish-async-query-core.yml
@@ -71,7 +71,7 @@ jobs:
 
       - uses: actions/checkout@v4
         with:
-          repository: 'opensearch-project/opensearch-build-libraries'
+          repository: 'opensearch-project/opensearch-build'
           path: 'build'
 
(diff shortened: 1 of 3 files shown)


dai-chen (approved):

.github/workflows/publish-async-query-core.yml
```diff
@@ -71,7 +71,7 @@ jobs:
 
       - uses: actions/checkout@v4
         with:
-          repository: 'opensearch-project/opensearch-build-libraries'
+          repository: 'opensearch-project/opensearch-build'
```
noCharger, line 74: can we make it a var so it can be reused in multiple files?

noCharger:
Let's add backport labels to corresponding branches

noCharger (approved):

Issue #3827, opened 2025-06-27
[BUG] SPAN has different behavior between pushdown and non-pushdown for timestamp before 1971


qianheng-aws:
**What is the bug?**
The PPL:
```
source = test | eval ts = timestamp("1961-04-12 09:10:00") | stats count() by span(ts, 1h) | head 1
```
get results:
```
{
  "schema": [
    {
      "name": "count()",
      "type": "bigint"
    },
    {
      "name": "span(ts,1h)",
      "type": "timestamp"
    }
  ],
  "datarows": [
    [
      6005,
      "1961-04-12 10:00:00"
    ]
  ],
  "total": 1,
  "size": 1
}
```

`"1961-04-12 09:10:00"` is rounded to `"1961-04-12 10:00:00"`. However, in DSL, it should be rounded to "1961-04-12 09:00:00"

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - OS: [e.g. Mac]
 - Version [e.g. 3.1.0-SNAPSHOT]
 - Plugins SQL

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

PR #4017, opened 2025-08-11
Fix span on negative timestamp

qianheng-aws:
### Description
Fix span on negative timestamp

### Related Issues
Resolves  https://github.com/opensearch-project/sql/issues/3827

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/utils/DateTimeUtils.java
+++ b/core/src/main/java/org/opensearch/sql/utils/DateTimeUtils.java
@@ -36,7 +36,8 @@ public class DateTimeUtils {
    * @return Rounded date/time value in utc millis
    */
   public static long roundFloor(long utcMillis, long unitMillis) {
-    return utcMillis - utcMillis % unitMillis;
+    long res = utcMillis - utcMillis % unitMillis;
+    return (utcMillis < 0 && res != utcMillis) ? res - unitMillis : res;
   }
 
   /**
@@ -65,7 +66,9 @@ public class DateTimeUtils {
         (zonedDateTime.getYear() - initDateTime.getYear()) * 12L
             + zonedDateTime.getMonthValue()
             - initDateTime.getMonthValue();
-    long monthToAdd = (monthDiff / interval - 1) * interval;
+    long multiplier = monthDiff / interval - 1;
+    if (monthDiff < 0 && monthDiff % interval != 0) --multiplier;
+    long monthToAdd = multiplier * interval;
     return initDateTime.plusMonths(monthToAdd).toInstant().toEpochMilli();
   }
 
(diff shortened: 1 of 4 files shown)


LantaoJin (dismissed):

LantaoJin (approved):

core/src/test/java/org/opensearch/sql/utils/DateTimeUtilsTest.java
```diff
@@ -16,6 +16,7 @@
 import java.time.format.DateTimeFormatter;
 import java.util.concurrent.TimeUnit;
 import org.junit.jupiter.api.Test;
+import org.opensearch.sql.planner.physical.collector.Rounding.DateTimeUnit;
 
 public class DateTimeUtilsTest {
```
dai-chen, line 21: np: this is actually UT for `DateTimeUnit`?

core/src/main/java/org/opensearch/sql/utils/DateTimeUtils.java
```diff
@@ -65,7 +66,9 @@ public static long roundMonth(long utcMillis, int interval) {
         (zonedDateTime.getYear() - initDateTime.getYear()) * 12L
             + zonedDateTime.getMonthValue()
             - initDateTime.getMonthValue();
-    long monthToAdd = (monthDiff / interval - 1) * interval;
+    long multiplier = monthDiff / interval - 1;
+    if (monthDiff < 0 && monthDiff % interval != 0) --multiplier;
+    long monthToAdd = multiplier * interval;
```
dai-chen, line 71: Not sure if I understand correct: this can be abstract as

```
static long roundMonth(long utcMillis, int interval)   { return roundByMonths(utcMillis, interval); }
static long roundQuarter(long utcMillis, int interval) { return roundByMonths(utcMillis, interval * 3); }
static long roundYear(long utcMillis, int interval)    { return roundByMonths(utcMillis, interval * 12); }
```

dai-chen (approved):

core/src/test/java/org/opensearch/sql/utils/DateTimeUtilsTest.java
qianheng-aws: No, It's actually testing the `round` API for each DateTimeUnit, and that APIs are only called by `DateTimeUtilsT`, so I put the test case here.

core/src/main/java/org/opensearch/sql/utils/DateTimeUtils.java
qianheng-aws: Agree, seems that could be abstracted. This focus on bugfix and just keep the exist implementation without too much refinement. Maybe that refinement could be done next time if we need further related change.

Issue #4007, opened 2025-08-07
[FEATURE] Enhance search command to support free text and wildcards.


vamsimanohar:
Ask:

Support following types of queries.

*  Searching for terms and phrases with out field value. 
   * `search User Not Found`
   *  `search "User Not Found"`
* Searching with wildcard characters
   * `search ccnumber="4232*1232" H*l*o`

penghuo:
Search command required `search field is indexed as a text type, with all punctuation and spaces preserved in the index`. e.g.

```
###
PUT {{baseUrl}}/log00001
Content-Type: application/x-ndjson

{
  "settings": {
    "analysis": {
      "analyzer": {
        "bicla": {
          "type": "custom",
          "tokenizer": "my_tokenizer"
        }
      },
      "tokenizer": {
        "my_tokenizer": {
          "type": "pattern",
          "pattern": "(\\w+)|(\\p{Punct})|(\\s)",
          "group": 0
        }
      }
    }
  },
  "mappings": {
    "properties": {
      "message": {
        "type": "text",
        "analyzer": "bicla"
      }
    }
  }
}

###
POST {{baseUrl}}/log00001/_bulk
Content-Type: application/x-ndjson

{ "index": { "_id": 1 } }
{"message": "[2025-08-26] logtype=ws:access error"}
{ "index": { "_id": 2 } }
{"message": "[2025-08-26] logtype!=ws:access this is unexpected"}
{ "index": { "_id": 3 } }
{"message": "[2025-08-26] ws:access logtype this is unexpected"}
{ "index": { "_id": 4 } }
{"message": "[2025-08-26] logtype=ws:access   error"}


### Return Doc1 and Doc4
POST {{baseUrl}}/log00001/_search
Content-Type: application/json

{
  "query": {
    "match_phrase": {
      "message": {
        "query": "logtype=ws:access"
      }
    }
  }
}

### Return Doc1
POST {{baseUrl}}/log00001/_search
Content-Type: application/json

{
  "query": {
    "match_phrase": {
      "message": {
        "query": "logtype=ws:access error"
      }
    }
  }
}

### Return Doc2
POST {{baseUrl}}/log00001/_search
Content-Type: application/json

{
  "query": {
    "match_phrase": {
      "message": {
        "query": "logtype!=ws:access"
      }
    }
  }
}

### Return Doc4
POST {{baseUrl}}/log00001/_search
Content-Type: application/json

{
  "query": {
    "match_phrase": {
      "message": {
[...]

PR #4018, opened 2025-08-11
Make fields optional parameter in multi field relevance function.

vamsimanohar:
### Description

This PR is the first step in making it easier to search in PPL. We want to support a simple "free text search" like the one described in [this issue](https://github.com/opensearch-project/sql/issues/4007)

The goal is to let you write a simple search like this:
`search source = my_index "fatal error"`

And our system will automatically translate it to this behind the scenes:
`search source = my_index | where simple_query_string("fatal error")`

### Changes in this PR

To prepare for the new free text search, I'm updating how a few of our search functions work.

Now, you won't always have to provide a list of fields to search in. The `fields` parameter is now optional for functions like `simple_query_string` and `multi_match`.

If you don't specify any fields, the query will automatically search the default fields set up for that index (specifically, in the `index.query.default_field` setting).

Updated Documentation can be found here: [PPL Relevance Functions](https://github.com/vamsimanohar/sql/blob/search_text/docs/user/ppl/functions/relevance.rst)

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/udf/RelevanceQueryFunction.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/udf/RelevanceQueryFunction.java
@@ -32,8 +32,11 @@ public class RelevanceQueryFunction extends ImplementorUDF {
   }
 
   /*
-   * Starting from the 3rd parameter, they are optional parameters for relevance queries.
-   * Different query has different parameter set, which will be validated in dedicated query builder
+   * The first parameter is always required (either fields or query).
+   * The second parameter is query when fields are present, otherwise it's the first parameter.
+   * Starting from the 3rd parameter (or 2nd when no fields), they are optional parameters for relevance queries.
+   * Different query has different parameter set, which will be validated in dedicated query builder.
+   * Query parameter is always required and cannot be null.
    */
   @Override
   public UDFOperandMetadata getOperandMetadata() {
@@ -55,7 +58,7 @@ public class RelevanceQueryFunction extends ImplementorUDF {
                         SqlTypeFamily.MAP,
                         SqlTypeFamily.MAP,
                         SqlTypeFamily.MAP),
-                    i -> i > 1 && i < 14) // Parameters 3-14 are optional
+                    i -> i > 0 && i < 14) // Parameters 3-14 are optional
                 .or(
                     OperandTypes.family(
                         ImmutableList.of(
@@ -84,7 +87,7 @@ public class RelevanceQueryFunction extends ImplementorUDF {
                             SqlTypeFamily.MAP,
                             SqlTypeFamily.MAP,
                             SqlTypeFamily.MAP),
-                        i -> i > 1 && i < 25))); // Parameters 3-25 are optional
+                        i -> i > 0 && i < 25))); // Parameters 3-25 are optional
   }
 
   public static class RelevanceQueryImplementor implements NotNullImplementor {
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/relevance/MultiFieldQuery.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/relevance/MultiFieldQuery.java
@@ -30,13 +30,15 @@ abstract class MultiFieldQuery<T extends QueryBuilder> extends RelevanceQuery<T>
   }
 
   @Override
+  protected int getMinimumParameterCount() {
+    return 1; // Only requires query parameter, fields is optional
+  }
+
+  @Override
   public T createQueryBuilder(List<NamedArgumentExpression> arguments) {
     // Extract 'fields' and 'query'
-    var fields =
-        arguments.stream()
-            .filter(a -> a.getArgName().equalsIgnoreCase("fields"))
-            .findFirst()
-            .orElseThrow(() -> new SemanticCheckException("'fields' parameter is missing."));
+    var fieldsOpt =
+        arguments.stream().filter(a -> a.getArgName().equalsIgnoreCase("fields")).findFirst();
 
     var query =
         arguments.stream()
@@ -44,9 +46,16 @@ abstract class MultiFieldQuery<T extends QueryBuilder> extends RelevanceQuery<T>
             .findFirst()
             .orElseThrow(() -> new SemanticCheckException("'query' parameter is missing"));
 
-    var fieldsAndWeights =
-        fields.getValue().valueOf().tupleValue().entrySet().stream()
-            .collect(ImmutableMap.toImmutableMap(e -> e.getKey(), e -> e.getValue().floatValue()));
+    ImmutableMap<String, Float> fieldsAndWeights;
+    if (fieldsOpt.isPresent()) {
+      fieldsAndWeights =
+          fieldsOpt.get().getValue().valueOf().tupleValue().entrySet().stream()
+              .collect(
+                  ImmutableMap.toImmutableMap(e -> e.getKey(), e -> e.getValue().floatValue()));
+    } else {
+      // Default to searching all fields if no fields specified
+      fieldsAndWeights = ImmutableMap.of();
+    }
 
     return createBuilder(fieldsAndWeights, query.getValue().valueOf().stringValue());
   }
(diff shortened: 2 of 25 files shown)


core/src/main/java/org/opensearch/sql/expression/function/udf/RelevanceQueryFunction.java
```diff
@@ -32,8 +32,11 @@ public SqlReturnTypeInference getReturnTypeInference() {
   }
 
   /*
-   * Starting from the 3rd parameter, they are optional parameters for relevance queries.
-   * Different query has different parameter set, which will be validated in dedicated query builder
+   * The first parameter is always required (either fields or query).
+   * The second parameter is query when fields are present, otherwise it's the first parameter.
+   * Starting from the 3rd parameter (or 2nd when no fields), they are optional parameters for relevance queries.
+   * Different query has different parameter set, which will be validated in dedicated query builder.
+   * Query parameter is always required and cannot be null.
```
Swiddis, line 39: Can you add a doc explaining what the fields actually do here? I can follow the field parsing but it's not clear what their use is, or how this is related to "operand metadata"

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/relevance/MultiFieldQuery.java
```diff
@@ -63,21 +72,29 @@ public T createQueryBuilder(List<NamedArgumentExpression> arguments) {
...
-                  RexLiteral fieldLiteral = (RexLiteral) fieldAndWeightNodes.get(i);
-                  RexLiteral weightLiteral = (RexLiteral) fieldAndWeightNodes.get(i + 1);
-                  String field =
-                      ((NlsString) Objects.requireNonNull(fieldLiteral.getValue())).getValue();
-                  Float weight =
-                      ((Double) Objects.requireNonNull(weightLiteral.getValue())).floatValue();
-                  return Map.entry(field, weight);
-                })
-            .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+    ImmutableMap<String, Float> fields;
```
Swiddis, line 75: Is it cleaner to create the ImmutableMap once here instead of building a Map from the fields list and freezing it at the end?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/relevance/SimpleQueryStringQuery.java
```diff
@@ -21,7 +21,12 @@ public SimpleQueryStringQuery() {
   @Override
   protected SimpleQueryStringBuilder createBuilder(
       ImmutableMap<String, Float> fields, String query) {
-    return QueryBuilders.simpleQueryStringQuery(query).fields(fields);
+    SimpleQueryStringBuilder builder = QueryBuilders.simpleQueryStringQuery(query);
```
Swiddis, line 24: If I'm understanding right, this whole PR is just plumbing for this part where we build a query string and add some fields, right?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -538,17 +538,27 @@ private List<UnresolvedExpression> multiFieldRelevanceArguments(
...
+        ctx.getRuleContexts(OpenSearchPPLParser.RelevanceFieldAndWeightContext.class);
+    if (fieldContexts != null && !fieldContexts.isEmpty()) {
+      var fields =
+          new RelevanceFieldList(
+              fieldContexts.stream()
+                  .collect(
+                      Collectors.toMap(
+                          f -> StringUtils.unquoteText(f.field.getText()),
+                          f -> (f.weight == null) ? 1F : Float.parseFloat(f.weight.getText()))));
+      builder.add(new UnresolvedArgument("fields", fields));
```
Swiddis, line 624: I notice we're reimplementing this "build if present" logic multiple times for fields (I count at least 4).

Is there some sort of safer helper we can provide to make this easier for callers, or are they all doing different things?

core/src/main/java/org/opensearch/sql/expression/function/udf/RelevanceQueryFunction.java
vamsimanohar: I have already added the documentation in description for fields.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/relevance/SimpleQueryStringQuery.java
vamsimanohar: yes...fields is not a required parameter, if not provided it will default to index.default.query_field setting.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/relevance/MultiFieldQuery.java
vamsimanohar: I didn't get  it..would you elaborate more?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
vamsimanohar: Are you talking about 

```
   if (!fields.isEmpty()) {
      builder = builder.fields(fields);
    }
 ```
 
 Are you taking about this?

Comment (vamsimanohar) [reply ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java:624]: These are only 2-3 lines with different builder types. I don't see pull into helper functions would make it any better.

File: core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java [hunk 5]
```diff
@@ -647,6 +647,17 @@ public RexNode visitRelevanceFieldList(RelevanceFieldList node, CalcitePlanConte
                       context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.DOUBLE),
                       true));
             });
+
+    // Handle empty field list case - create an empty map with explicit types
+    if (varArgRexNodeList.isEmpty()) {
```
penghuo: If read default fields from index.query.default_field setting, then other change is not necessary?

vamsimanohar: The current changes are not big in the other files except for more test files and also would leave the setting dependency to underlying DSL implementation.

Also, the above setting is a index specific setting, I need to somehow pass index into the context and add additional dependency on get Index settings call where the changes would be become more or less similar.


@penghuo  let me know if you still want to do that for other reasons.

penghuo: make sense, I thought it is cluster setting. 
for multiiple indicies case, DSL will handle it?

vamsimanohar: Good one, I need to see the behavior. IMO, it should handle.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
songkant-aws, line 624: With this check in AstBuilder, it will not produce RelevanceFieldList, right? In that case, I think you don't need the empty list check logic in `CalciteRexNodeVisitor.visitRelevanceFieldList` logic because it will not analyze that RexNode.

Another option is to add empty RelevanceFieldList here anyway. Then no UDF metadata check logic is changed. The function will look like: `simple_query_string(NULL AS MAP<VARCHAR, DOUBLE>, RexCall(MAP...), ...)`. And it may probably simplify some logic in PredicateAnalyzer by checking NULL literal instead of verifying parameters on different startIndices.

vamsimanohar: My bad, this file changes are not removed. let me see the other approach if it reduces the code.

vamsimanohar: I went with removing the the changes in RelevanceFieldList as the second approach is forcing me to change some V2 engine files and tests.

vamsimanohar:
The integ test failure is unrelated to this PR

Swiddis (approved):

songkant-aws (approved):

PR #4019, opened 2025-08-11
Add wildcard support for rename command

ritvibhatt:
### Description

### Related Issues

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
+++ b/core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
@@ -25,6 +25,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import java.util.Set;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -90,6 +91,7 @@ import org.opensearch.sql.ast.tree.Values;
 import org.opensearch.sql.common.antlr.SyntaxCheckException;
 import org.opensearch.sql.data.model.ExprMissingValue;
 import org.opensearch.sql.data.type.ExprCoreType;
+import org.opensearch.sql.data.type.ExprType;
 import org.opensearch.sql.datasource.DataSourceService;
 import org.opensearch.sql.exception.SemanticCheckException;
 import org.opensearch.sql.expression.DSL;
@@ -129,6 +131,7 @@ import org.opensearch.sql.planner.logical.LogicalWindow;
 import org.opensearch.sql.planner.physical.datasource.DataSourceTable;
 import org.opensearch.sql.storage.Table;
 import org.opensearch.sql.utils.ParseUtils;
+import org.opensearch.sql.utils.WildcardRenameUtils;
 
 /**
  * Analyze the {@link UnresolvedPlan} in the {@link AnalysisContext} to construct the {@link
@@ -309,20 +312,27 @@ public class Analyzer extends AbstractNodeVisitor<LogicalPlan, AnalysisContext> 
   @Override
   public LogicalPlan visitRename(Rename node, AnalysisContext context) {
     LogicalPlan child = node.getChild().get(0).accept(this, context);
+    
+    // Get available fields from current schema
+    Set<String> availableFields = getAvailableFieldNames(context);
+    
     ImmutableMap.Builder<ReferenceExpression, ReferenceExpression> renameMapBuilder =
         new ImmutableMap.Builder<>();
     for (Map renameMap : node.getRenameList()) {
-      Expression origin = expressionAnalyzer.analyze(renameMap.getOrigin(), context);
-      // We should define the new target field in the context instead of analyze it.
+      String originPattern = ((Field) renameMap.getOrigin()).getField().toString();
+      
       if (renameMap.getTarget() instanceof Field) {
-        ReferenceExpression target =
-            new ReferenceExpression(
-                ((Field) renameMap.getTarget()).getField().toString(), origin.type());
-        ReferenceExpression originExpr = DSL.ref(origin.toString(), origin.type());
-        TypeEnvironment curEnv = context.peek();
-        curEnv.remove(originExpr);
-        curEnv.define(target);
-        renameMapBuilder.put(originExpr, target);
+        String targetPattern = ((Field) renameMap.getTarget()).getField().toString();
+        
+        // Validate pattern compatibility for wildcards
+        if (WildcardRenameUtils.isWildcardPattern(originPattern) && 
+            !WildcardRenameUtils.validatePatternCompatibility(originPattern, targetPattern)) {
+          throw new SemanticCheckException(String.format(
+              "Wildcard count mismatch between source '%s' and target '%s' patterns", 
+              originPattern, targetPattern));
+        }
+
+        expandWildcardRename(originPattern, targetPattern, availableFields, renameMapBuilder, context);
       } else {
         throw new SemanticCheckException(
             String.format("the target expected to be field, but is %s", renameMap.getTarget()));
@@ -830,5 +840,58 @@ public class Analyzer extends AbstractNodeVisitor<LogicalPlan, AnalysisContext> 
     groupByList.add(patternsField);
     groupByList.addAll(node.getPartitionByList());
     return new Aggregation(aggExprs, ImmutableList.of(), groupByList);
+  }
+
+  /**
+   * Get available field names from current type environment.
+   *
+   * @param context the analysis context
+   * @return set of available field names
+   */
+  private Set<String> getAvailableFieldNames(AnalysisContext context) {
+    TypeEnvironment currentEnv = context.peek();
+    return currentEnv.lookupAllFields(Namespace.FIELD_NAME).keySet();
+  }
+
+  /**
+   * Expand wildcard rename patterns to concrete field mappings.
+   *
+   * @param sourcePattern the source wildcard pattern
+   * @param targetPattern the target wildcard pattern
+   * @param availableFields set of available field names
+   * @param renameMapBuilder builder for rename mappings
+   * @param context the analysis context
+   */
+  private void expandWildcardRename(
+      String sourcePattern,
+      String targetPattern,
+      Set<String> availableFields,
+      ImmutableMap.Builder<ReferenceExpression, ReferenceExpression> renameMapBuilder,
+      AnalysisContext context) {
+
+    List<String> matchingFields = WildcardRenameUtils.matchFieldNames(sourcePattern, availableFields);
+
+    if (matchingFields.isEmpty()) {
+      throw new SemanticCheckException(
+          String.format("No fields match the wildcard pattern '%s'", sourcePattern));
+    }
+
+    TypeEnvironment curEnv = context.peek();
+
+    for (String fieldName : matchingFields) {
+      String newName = WildcardRenameUtils.applyWildcardTransformation(
+          sourcePattern, targetPattern, fieldName);
+
+      Symbol fieldSymbol = new Symbol(Namespace.FIELD_NAME, fieldName);
+      ExprType fieldType = curEnv.resolve(fieldSymbol);
+
+      ReferenceExpression origin = DSL.ref(fieldName, fieldType);
+      ReferenceExpression target = new ReferenceExpression(newName, fieldType);
+
+      curEnv.remove(origin);
+      curEnv.define(target);
+
+      renameMapBuilder.put(origin, target);
+    }
   }
 }
--- a/core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
+++ b/core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.utils;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * Utility class for handling wildcard patterns in rename operations.
+ * Supports shell-style (*) wildcards.
+ */
+public class WildcardRenameUtils {
+  
+  /**
+   * Check if pattern contains any supported wildcards.
+   *
+   * @param pattern the pattern to check
+   * @return true if pattern contains * wildcards
+   */
+  public static boolean isWildcardPattern(String pattern) {
+    return pattern.contains("*");
+  }
+  
+  /**
+   * Check if pattern is a single wildcard that matches all fields.
+   *
+   * @param pattern the pattern to check
+   * @return true if pattern is exactly "*"
+   */
+  public static boolean isFullWildcardPattern(String pattern) {
+    return "*".equals(pattern);
+  }
+  
+  
+  /**
+   * Convert wildcard pattern to regex with capture groups.
+   *
+   * @param pattern the wildcard pattern
+   * @return regex pattern with capture groups
+   */
+  public static String wildcardToRegex(String pattern) {
+    String[] parts = pattern.split("\\*", -1);
+    return Arrays.stream(parts)
+        .map(Pattern::quote)
+        .collect(Collectors.joining("(.*)"));
+  }
+  
+  /**
+   * Match field names against wildcard pattern.
+   *
+   * @param wildcardPattern the pattern to match against
+   * @param availableFields set of available field names
+   * @return list of matching field names, sorted
+   */
+  public static List<String> matchFieldNames(String wildcardPattern, Set<String> availableFields) {
+    if (!isWildcardPattern(wildcardPattern)) {
+      // No wildcards
+      return availableFields.contains(wildcardPattern) 
+          ? List.of(wildcardPattern) 
+          : List.of();
+    }
+    
+    if (isFullWildcardPattern(wildcardPattern)) {
+      // Single wildcard matches all available fields
+      return availableFields.stream()
+          .sorted()
+          .collect(Collectors.toList());
+    }
+    
+    String regexPattern = "^" + wildcardToRegex(wildcardPattern) + "$";
+    Pattern pattern = Pattern.compile(regexPattern, Pattern.CASE_INSENSITIVE);
+    
+    return availableFields.stream()
+        .filter(field -> pattern.matcher(field).matches())
+        .sorted()
+        .collect(Collectors.toList());
+  }
+  
+  /**
+   * Apply wildcard transformation to generate new field name.
+   *
+   * @param sourcePattern the source wildcard pattern
+   * @param targetPattern the target wildcard pattern
+   * @param actualFieldName the actual field name to transform
+   * @return transformed field name
+   * @throws IllegalArgumentException if patterns don't match or are invalid
+   */
+  public static String applyWildcardTransformation(
+      String sourcePattern, 
+      String targetPattern, 
+      String actualFieldName) {
+    
+    // No wildcards in either pattern
+    if (!isWildcardPattern(sourcePattern) && !isWildcardPattern(targetPattern)) {
+      return targetPattern;
+    }
+    
+    // Both are full wildcards
+    if (isFullWildcardPattern(sourcePattern) && isFullWildcardPattern(targetPattern)) {
+      return actualFieldName;
+    }
+
+    if (isFullWildcardPattern(sourcePattern)) {
+      // Replace * in target with the actual field name
+      return targetPattern.replace("*", actualFieldName);
+    }
+
+    String sourceRegex = "^" + wildcardToRegex(sourcePattern) + "$";
+    Pattern sourceP = Pattern.compile(sourceRegex, Pattern.CASE_INSENSITIVE);
+    Matcher matcher = sourceP.matcher(actualFieldName);
+    
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException(
+          String.format("Field '%s' does not match pattern '%s'", 
+              actualFieldName, sourcePattern));
+    }
+
+    String result = targetPattern;
+
+    for (int i = 1; i <= matcher.groupCount(); i++) {
+      String capturedValue = matcher.group(i);
+      
+      int index = result.indexOf("*");
+      if (index >= 0) {
+        result = result.substring(0, index) + capturedValue + result.substring(index + 1);
+      }
+    }
+    
+    return result;
+  }
+  
+  /**
+   * Validate that source and target patterns have matching wildcard counts.
+   *
+   * @param sourcePattern the source pattern
+   * @param targetPattern the target pattern
+   * @return true if patterns are compatible
+   */
+  public static boolean validatePatternCompatibility(String sourcePattern, String targetPattern) {
+    int sourceWildcards = countWildcards(sourcePattern);
+    int targetWildcards = countWildcards(targetPattern);
+    return sourceWildcards == targetWildcards;
+  }
+  
+  /**
+   * Count the number of wildcards in a pattern.
+   *
+   * @param pattern the pattern to analyze
+   * @return number of wildcard characters
+   */
+  private static int countWildcards(String pattern) {
+    return (int) pattern.chars().filter(ch -> ch == '*').count();
+  }
+  
+}
--- a/ppl/src/main/antlr/OpenSearchPPLParser.g4
+++ b/ppl/src/main/antlr/OpenSearchPPLParser.g4
@@ -1077,6 +1077,7 @@ tableIdent
 
 wildcard
    : ident (MODULE ident)* (MODULE)?
+   | STAR
    | SINGLE_QUOTE wildcard SINGLE_QUOTE
    | DOUBLE_QUOTE wildcard DOUBLE_QUOTE
    | BACKTICK wildcard BACKTICK


docs/user/ppl/cmd/rename.rst
```diff
@@ -18,8 +18,8 @@ Syntax
 ============
 rename <source-field> AS <target-field>["," <source-field> AS <target-field>]...
 
-* source-field: mandatory. The name of the field you want to rename.
-* field list: mandatory. The name you want to rename to.
+* source-field: mandatory. The name of the field you want to rename. Supports wildcard patterns using ``*``.
+* target-field: mandatory. The name you want to rename to. Must have same number of wildcards as the source.
```
penghuo, line 22: Add a test case of multiple rename, e.g. `rename *name as *_name, *_name as *@name`

integ-test/src/test/java/org/opensearch/sql/ppl/RenameCommandIT.java
```diff
@@ -43,12 +41,51 @@ public void testRenameMultiField() throws IOException {
...
+    verifyColumn(result, columnName("old_firstname"), columnName("old_lastname"));
+  }
+
+  @Test
+  public void testRenameWildcardWithMultipleCaptures() throws IOException {
+    JSONObject result =
+        executeQuery(
+            "source="
+                + TEST_INDEX_ACCOUNT
+                + " | fields firstname, lastname | rename *first* as *FIRST*");
```
penghuo: Does `rename f*r*t* as F*R*T*` supported?

core/src/test/java/org/opensearch/sql/utils/WildcardRenameUtilsTest.java
```diff
@@ -0,0 +1,124 @@
...
+    // Test exact match (no wildcards)
+    List<String> exactMatch = WildcardRenameUtils.matchFieldNames("age", fields);
+    assertEquals(List.of("age"), exactMatch);
+    List<String> exactNoMatch = WildcardRenameUtils.matchFieldNames("xyz", fields);
+    assertTrue(exactNoMatch.isEmpty());
+  }
+
+  @Test
+  void testApplyWildcardTransformation() {
+    assertEquals(
```
penghuo, line 80: add a test case for partial matched. e.g. `f*r*tname, F*R*Tname`, it should be FiRTname?

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -311,10 +314,56 @@ public LogicalPlan visitRename(Rename node, AnalysisContext context) {
...
-      // We should define the new target field in the context instead of analyze it.
-      if (renameMap.getTarget() instanceof Field) {
+      if (!(renameMap.getTarget() instanceof Field)) {
+        throw new SemanticCheckException(
+            String.format("the target expected to be field, but is %s", renameMap.getTarget()));
+      }
+
+      if (renameMap.getOrigin() instanceof Field
+          && WildcardRenameUtils.isWildcardPattern(
+              ((Field) renameMap.getOrigin()).getField().toString())) {
```
penghuo: What happen if origin field name include speicial * character, e.g. 
```
rename `field*` as `myfield*`
```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -311,10 +314,56 @@ public LogicalPlan visitRename(Rename node, AnalysisContext context) {
...
     ImmutableMap.Builder<ReferenceExpression, ReferenceExpression> renameMapBuilder =
         new ImmutableMap.Builder<>();
+
+    Set<String> availableFields = getAvailableFieldNames(context);
+
     for (Map renameMap : node.getRenameList()) {
-      Expression origin = expressionAnalyzer.analyze(renameMap.getOrigin(), context);
-      // We should define the new target field in the context instead of analyze it.
-      if (renameMap.getTarget() instanceof Field) {
+      if (!(renameMap.getTarget() instanceof Field)) {
```
dai-chen: Are all these changes for V2 engine support?

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -311,10 +314,56 @@ public LogicalPlan visitRename(Rename node, AnalysisContext context) {
...
+        }
+
+        // Handle wildcard rename
+        List<String> matchingFields =
+            WildcardRenameUtils.matchFieldNames(fieldName, availableFields);
+
+        if (matchingFields.isEmpty()) {
+          throw new SemanticCheckException(
+              String.format("No fields match the pattern '%s'", fieldName));
+        }
```
penghuo: If match failed, query should works, but no field renamed, right?

docs/user/ppl/cmd/rename.rst
```diff
@@ -59,6 +59,25 @@ PPL query::
...
 
+
+Example 3: Rename with wildcards
+=================================
+
+The example shows renaming multiple fields using wildcard patterns.
+
+PPL query::
+
+    os> source=accounts | rename *name as *_name | fields first_name, last_name;
```
dai-chen, line 85: Is multiple pattern supported? e.g., `rename *name as *_name, *age as *_age`.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -311,10 +314,56 @@ public LogicalPlan visitRename(Rename node, AnalysisContext context) {
     LogicalPlan child = node.getChild().get(0).accept(this, context);
     ImmutableMap.Builder<ReferenceExpression, ReferenceExpression> renameMapBuilder =
         new ImmutableMap.Builder<>();
+
+    Set<String> availableFields = getAvailableFieldNames(context);
+
     for (Map renameMap : node.getRenameList()) {
```
penghuo: Why chagne Analyzer? It is for calcite.enable=false? If yes, it is not required.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -292,19 +293,53 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
     visitChildren(node, context);
     List<String> originalNames = context.relBuilder.peek().getRowType().getFieldNames();
     List<String> newNames = new ArrayList<>(originalNames);
+
     for (org.opensearch.sql.ast.expression.Map renameMap : node.getRenameList()) {
```
penghuo, line 470: does cascading rename use case supported?,  e.g. ` e.g. rename *name as *_name, *_name as *@name`

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
ritvibhatt: Yes, removed all of the changes for v2 support

integ-test/src/test/java/org/opensearch/sql/ppl/RenameCommandIT.java
ritvibhatt: Yes, added test for partial match to ```CalcitePPLRenameIT```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
ritvibhatt: Can't match/replace wildcard characters in field names, this is SPL behavior as well. Added a note in ```rename.rst```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
ritvibhatt: Yes, updated to just not rename anything if wildcard doesn't match any fields

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
ritvibhatt: Removed changes for v2 engine support

docs/user/ppl/cmd/rename.rst
ritvibhatt: Yes it is, added as example 4 in ```rename.rst```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ritvibhatt: Updated to support cascading rename for both wildcard and no wildcard and added test in ```CalcitePPLRenameIT```

docs/user/ppl/cmd/rename.rst
```diff
@@ -18,8 +18,13 @@ Syntax
 ============
 rename <source-field> AS <target-field>["," <source-field> AS <target-field>]...
 
-* source-field: mandatory. The name of the field you want to rename.
-* field list: mandatory. The name you want to rename to.
+* source-field: mandatory. The name of the field you want to rename. Supports wildcard patterns using ``*``.
```
penghuo: Add since 3.3 to indicate this is a new feature

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -382,22 +383,54 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
...
+      if (renameMap.getOrigin() instanceof Field
+          && WildcardRenameUtils.isWildcardPattern(
+              ((Field) renameMap.getOrigin()).getField().toString())) {
+        String sourcePattern = ((Field) renameMap.getOrigin()).getField().toString();
+        String targetPattern = ((Field) renameMap.getTarget()).getField().toString();
+
+        if (!WildcardRenameUtils.validatePatternCompatibility(sourcePattern, targetPattern)) {
+          throw new SemanticCheckException(
+              "Source and target patterns have different wildcard counts");
+        }
```
penghuo: Add IT for failure case, and assert error message.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -382,22 +383,54 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
...
+                  sourcePattern, targetPattern, fieldName);
+
+          int fieldIndex = newNames.indexOf(fieldName);
+          if (fieldIndex >= 0) {
+            newNames.set(fieldIndex, newName);
+            context.relBuilder.rename(newNames);
+          } else {
+            throw new SemanticCheckException(
+                String.format("the wildcard matched field %s cannot be resolved", fieldName));
+          }
```
penghuo: Add IT for failure case, and assert error message.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
penghuo: Is it expected?
```
###
POST {{baseUrl}}/tttt/_bulk

{ "index": { "_id": 1 } }
{ "value*": 10, "id123": 1 }


###

{
  "query": "source=tttt "
}

###
POST {{baseUrl}}/_plugins/_ppl

{
  "query": "source=tttt | rename `value*` as cde*"
}

### Results
{
  "schema": [
    {
      "name": "cde*",
      "type": "bigint"
    },
    {
      "name": "id123",
      "type": "bigint"
    }
  ],
  "datarows": [
    [
      10,
      1
    ]
  ],
  "total": 1,
  "size": 1
}
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ritvibhatt: Changed to update field names after each rename clause so next rename can see changes instead of all at once after all renames. Added ```testCascadingRename``` and ```testCascadingRenameWithWildcard``` in ```CalcitePPLRenameIT```

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
ritvibhatt: Yes, the ```*``` will match the ```*``` in the other field name but it can't be replaced in the rename, so trying to rename ```value*``` as ```cde``` won't work

penghuo (dismissed):

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -382,22 +383,52 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
...
+
+      Set<String> availableFields = new HashSet<>(newNames);
+      List<String> matchingFields =
+          WildcardRenameUtils.matchFieldNames(sourcePattern, availableFields);
+
+      for (String fieldName : matchingFields) {
+        String newName;
+        newName =
+            WildcardRenameUtils.applyWildcardTransformation(
+                sourcePattern, targetPattern, fieldName);
```
ykmr1224: nit: any reason to separate line? (can we simply do `String newName = ...;`)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -382,22 +383,52 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
...
+
+        // If target field already exists, remove field before renaming source
+        int existingFieldIndex = newNames.indexOf(newName);
+        if (existingFieldIndex != -1 && !fieldName.equals(newName)) {
```
ykmr1224: Can we simply do like `if (newNames.contains(newName) && ...)`?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -382,22 +383,52 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
...
+      List<String> matchingFields =
+          WildcardRenameUtils.matchFieldNames(sourcePattern, availableFields);
+
+      for (String fieldName : matchingFields) {
+        String newName;
+        newName =
+            WildcardRenameUtils.applyWildcardTransformation(
+                sourcePattern, targetPattern, fieldName);
+
+        // If target field already exists, remove field before renaming source
```
ykmr1224: Better extracting as `removeFieldIfExist(...)`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -382,22 +383,52 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
...
+        if (existingFieldIndex != -1 && !fieldName.equals(newName)) {
+          newNames.remove(newName);
+          context.relBuilder.projectExcept(context.relBuilder.field(newName));
+        }
+
+        int fieldIndex = newNames.indexOf(fieldName);
+        if (fieldIndex != -1) {
+          newNames.set(fieldIndex, newName);
+        }
+        context.relBuilder.rename(newNames);
```
ykmr1224: Does this mean we call `rename` repeatedly?
Do we need to call it multiple time? Is it enough to call it once at the end of this method?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -382,22 +383,52 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
...
+      }
+      // if source field doesn't exist but target does, remove target field from results
+      if (matchingFields.isEmpty() && newNames.contains(targetPattern)) {
+        newNames.remove(targetPattern);
```
ykmr1224: Is it find to directly match with targetPattern? (Does this mean we don't care wildcard here?)

core/src/test/java/org/opensearch/sql/utils/WildcardRenameUtilsTest.java
```diff
@@ -0,0 +1,129 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.LinkedHashSet;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class WildcardRenameUtilsTest {
```
ykmr1224, line 17: nit: Missing `public`

core/src/test/java/org/opensearch/sql/utils/WildcardRenameUtilsTest.java
```diff
@@ -0,0 +1,129 @@
...
+  void testWildcardToRegex() {
+    assertEquals("\\Q\\E(.*)\\Qname\\E", WildcardRenameUtils.wildcardToRegex("*name"));
+    assertEquals("\\Qname\\E(.*)\\Q\\E", WildcardRenameUtils.wildcardToRegex("name*"));
+    assertEquals(
+        "\\Q\\E(.*)\\Q_\\E(.*)\\Q_field\\E", WildcardRenameUtils.wildcardToRegex("*_*_field"));
+  }
+
+  @Test
+  void testMatchFieldNames() {
+    LinkedHashSet<String> fields = new LinkedHashSet<>();
```
ykmr1224: Why LinkedHashSet? Plus, it is better using immutable collection such as `ImmutableSet.of(...)` whenever possible. (mutable object tend to cause problem, and use of immutable collection is recommended)

Let's rename it to `availableFields` for better clarity.

core/src/test/java/org/opensearch/sql/utils/WildcardRenameUtilsTest.java
```diff
@@ -0,0 +1,129 @@
...
+    List<String> firstFields = WildcardRenameUtils.matchFieldNames("first*", fields);
+    assertEquals(List.of("firstname"), firstFields);
+
+    // Test full wildcard - matches all fields
+    List<String> allFields = WildcardRenameUtils.matchFieldNames("*", fields);
+    assertEquals(List.of("firstname", "lastname", "age", "address", "fullname"), allFields);
+
+    // Test no matches
+    List<String> noMatch = WildcardRenameUtils.matchFieldNames("*xyz", fields);
+    assertTrue(noMatch.isEmpty());
```
ykmr1224: nit: It is better splitting method than making code block by comment. Same for other places.

core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Match field names against wildcard pattern.
+   *
+   * @param wildcardPattern the pattern to match against
+   * @param availableFields set of available field names
+   * @return list of matching field names
+   */
+  public static List<String> matchFieldNames(String wildcardPattern, Set<String> availableFields) {
```
ykmr1224: I was not sure why we need to use `Set` for `availableFields`

core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
```diff
@@ -0,0 +1,138 @@
...
+      String sourcePattern, String targetPattern, String actualFieldName) {
+
+    // Both are full wildcards
+    if (isFullWildcardPattern(sourcePattern) && isFullWildcardPattern(targetPattern)) {
+      return actualFieldName;
+    }
+
+    if (isFullWildcardPattern(sourcePattern)) {
+      // Replace * in target with the actual field name
+      return targetPattern.replace("*", actualFieldName);
```
ykmr1224: Does this mean all the `*` will be replaced with actualFieldName?

core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    String sourceRegex = "^" + wildcardToRegex(sourcePattern) + "$";
+    Pattern sourceP = Pattern.compile(sourceRegex);
+    Matcher matcher = sourceP.matcher(actualFieldName);
```
ykmr1224: nit: I think we don't need `sourceP` variable. (I would extract convert from `sourcePattern` to matcher.

core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
```diff
@@ -0,0 +1,138 @@
...
+          String.format("Field '%s' does not match pattern '%s'", actualFieldName, sourcePattern));
+    }
+
+    String result = targetPattern;
+
+    for (int i = 1; i <= matcher.groupCount(); i++) {
+      String capturedValue = matcher.group(i);
+
+      int index = result.indexOf("*");
+      if (index >= 0) {
```
ykmr1224, line 108: Does this mean we will ignore in case result has fewer `*`? It might be safer to throw exception if we don't expect this case. If we intentionally ignore this case, let's clearly mention it by a comment.

docs/user/ppl/cmd/rename.rst
```diff
@@ -59,6 +74,64 @@ PPL query::
...
 
+
+Example 3: Rename with wildcards
+=================================
+
+The example shows renaming multiple fields using wildcard patterns. (Requires Calcite query engine)
+
+PPL query::
+
+    PPL> source=accounts | rename *name as *_name | fields first_name, last_name;
```
ykmr1224: The pattern with `PPL>` won't be executed by doctest. (It needs to be `os>`)
I'm working on fix to enable Calcite in [this PR](https://github.com/opensearch-project/sql/pull/4100)
(We can fix it later if my PR takes time to merge)

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLRenameIT.java
```diff
@@ -94,6 +85,12 @@ public void testRenameToMetaField() throws IOException {
         schema("country", "string"),
         schema("year", "int"),
         schema("month", "int"));
+    verifyDataRows(
+        result,
+        rows("Jake", "USA", "California", 4, 2023, 70),
+        rows("Hello", "USA", "New York", 4, 2023, 30),
+        rows("John", "Canada", "Ontario", 4, 2023, 25),
+        rows("Jane", "Canada", "Quebec", 4, 2023, 20));
```
ykmr1224: Better extracting it as many test case verify same data rows.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -140,7 +140,7 @@ wcFieldList
    ;
 
 renameCommand
-   : RENAME renameClasue (COMMA renameClasue)*
+   : RENAME renameClasue (COMMA? renameClasue)*
```
ykmr1224, line 154: Can we omit COMMA? Should we have test case if it is intended change?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -1119,6 +1119,7 @@ tableIdent
 
 wildcard
    : ident (MODULE ident)* (MODULE)?
+   | STAR
```
ykmr1224: Is it safe to add here? (This could affect other commands)

dai-chen (dismissed):
Thanks for the changes!

ykmr1224 (changes requested):
Field name could contain `.` such as nested field. Wondering if that works fine with current implementation.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ritvibhatt: Updated to make one line

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ritvibhatt: Updated!

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ritvibhatt: updated

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ritvibhatt: You're right calling it once at the end is enough, updated!

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ritvibhatt: Yes, if the target pattern does have wildcards but the source pattern didn't match anything, nothing should change since there is nothing to replace the wildcards in the source pattern with so it won't match any existing fields

core/src/test/java/org/opensearch/sql/utils/WildcardRenameUtilsTest.java
ritvibhatt: Updated to use immutable collection and renamed

core/src/test/java/org/opensearch/sql/utils/WildcardRenameUtilsTest.java
ritvibhatt: Updated

core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
ritvibhatt: changed to use collection

core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
ritvibhatt: Updated so that consecutive wildcards will throw an error unless renaming to same thing (ex ```rename **name as **name``` won't change any names and won't throw error) or both target and source are full wildcard (ex ```rename ** as **``` won't change any names and won't throw error). So every * wont be replaced by the full field name now

core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
ritvibhatt: Updated

core/src/main/java/org/opensearch/sql/utils/WildcardRenameUtils.java
ritvibhatt: Updated to throw exception!

docs/user/ppl/cmd/rename.rst
ritvibhatt: Updated

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLRenameIT.java
ritvibhatt: Updated

ppl/src/main/antlr/OpenSearchPPLParser.g4
ritvibhatt: Added tests for delimiting with space

ppl/src/main/antlr/OpenSearchPPLParser.g4
ritvibhatt: Updated to use ```renameFieldExpression``` in ```renameClasue``` so STAR is only added to rename

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -415,25 +416,52 @@ public RelNode visitRename(Rename node, CalcitePlanContext context) {
...
       }
+
+      String sourcePattern = ((Field) renameMap.getOrigin()).getField().toString();
+      String targetPattern = ((Field) renameMap.getTarget()).getField().toString();
+
+      if (WildcardRenameUtils.isWildcardPattern(sourcePattern)
+          && !WildcardRenameUtils.validatePatternCompatibility(sourcePattern, targetPattern)) {
+        throw new SemanticCheckException(
+            "Source and target patterns have different wildcard counts");
+      }
```
ykmr1224, line 483: nit&optional: It might be simpler to put this logic into validatePatternCompatibility and throw exception from there. (simply call `WildcardRenameUtils.validatePatternCompatibility` from this method)

ykmr1224 (approved):

RyanL1997 (approved):

LantaoJin:
Backporing is required manually in this PR. @ykmr1224

ritvibhatt:
@LantaoJin I raised the manual backport (#4250), trying to figure out test failures

Issue #4014, opened 2025-08-08
[RFC] Support `eval` command in PPL


ahkcs:
### Description
The PPL `eval` command has critical feature gaps compared to SPL that block essential data processing workflows. Most notably, string concatenation operations fail with runtime errors, and the command lacks OpenSearch DSL optimization.

### Environment
- OpenSearch version: [Please specify]
- PPL Plugin version: [Please specify]

#### 1. String Concatenation Failure (P0 - CRITICAL)
**Current Behavior:** String concatenation using the `+` operator fails with a runtime error.

**Expected Behavior:** String concatenation should work seamlessly like in SPL.

**Error Message:**
```
java.lang.RuntimeException: java.sql.SQLException: Error while preparing plan
LogicalProject... full_name=[+(+($1, ' '), $10)]
```

#### 2. Integer Division Precision Loss (P2 - MEDIUM)
**Current Behavior:** PPL uses strict integer division, losing decimal precision.

**Expected Behavior:** Auto-promotion to decimal types for division operations (SPL behavior).

**Example:**
- SPL: `balance / 1000` → `16.623, 39.225, 28.838`
- PPL: `balance / 1000` → `16, 39, 28`

#### 3. Performance Architecture Limitation (P1 - HIGH)
**Current Behavior:** The eval command is not rewritten to OpenSearch DSL and executes only on the coordination node.

**Expected Behavior:** Eval operations should be pushed down to OpenSearch DSL for distributed execution.

#### String Concatenation Issue:
```bash
# This fails in PPL but works in SPL
curl -X POST "localhost:9200/_plugins/_ppl" \
-H "Content-Type: application/json" \
-d '{
  "query": "source=accounts | eval full_name = firstname+\" \"+lastname | sort account_number | fields account_number, firstname, lastname, full_name | head 3"
}'
```

#### Integer Division Issue:
```bash
# Returns integer results instead of decimals
curl -X POST "localhost:9200/_plugins/_ppl" \
-H "Content-Type: application/json" \
-d '{
  "query": "source=accounts | eval balance_k = balance / 1000 | sort account_number | fields account_number, balance, balance_k | head 3"
}'
```

### Impact
- **String Operations:** Blocks all text processing workflows in eval expressions
- **Numeric Precision:** Mathematical calculations lose precision unless developers explicitly use decimal literals
- **Performance:** Potential performance bottleneck for large-scale data processing

#### Priority 1: String Concatenation Support (P0)
[...]

PR #4020, opened 2025-08-11
Fix PPL eval command string concatenation with + operator

ahkcs:
# Fix PPL eval command string concatenation with + operator

## Description

This PR fixes a critical issue where string concatenation using the + operator in PPL eval commands would fail with a runtime error during Calcite query planning. Related #4014

## Problem

PPL queries with string concatenation using the + operator were failing with:
```
java.sql.SQLException: Error while preparing plan
LogicalProject... full_name=[+(+($1, ' '), $10)]
```

Example failing query:
```sql
source=accounts | eval full_name = firstname + " " + lastname | fields account_number, firstname, lastname, full_name
```

## Example Usage

After this fix, the following PPL queries now work correctly:

```sql
# Basic string concatenation
source=accounts | eval full_name = firstname + " " + lastname
```


core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -488,6 +488,22 @@ public RexNode visitFunction(Function node, CalcitePlanContext context) {
...
+    // Handle string concatenation with + operator
+    if ("+".equals(node.getFuncName()) && arguments.size() == 2) {
+      RexNode left = arguments.get(0);
+      RexNode right = arguments.get(1);
+
+      // Check if both operands are strings
+      if (left.getType().getSqlTypeName() == SqlTypeName.VARCHAR
+          || left.getType().getSqlTypeName() == SqlTypeName.CHAR) {
+        if (right.getType().getSqlTypeName() == SqlTypeName.VARCHAR
+            || right.getType().getSqlTypeName() == SqlTypeName.CHAR) {
```
ykmr1224: Do we convert types implicitly? In that case, should we handle other types?

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -488,6 +488,22 @@ public RexNode visitFunction(Function node, CalcitePlanContext context) {
       }
     }
 
+    // Handle string concatenation with + operator
+    if ("+".equals(node.getFuncName()) && arguments.size() == 2) {
+      RexNode left = arguments.get(0);
```
ykmr1224: Let's extract and make `visitFunction` smaller for better readability.

dai-chen: +1. Is it possible to leverage current visitor pattern to dispatch instead of adding more and more conditional checks in a single visit method?

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ahkcs: Currently, we don't perform implicit type conversion - we only handle the case where both operands are already string types (VARCHAR/CHAR) or both numbers(already supported)

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ahkcs: I have implemented a dispatch method to resolve this

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -497,6 +503,64 @@ public RexNode visitFunction(Function node, CalcitePlanContext context) {
...
+   */
+  private RexNode tryHandleSpecialOperators(
+      Function node, List<RexNode> arguments, CalcitePlanContext context) {
+    switch (node.getFuncName()) {
+      case "+":
+        return tryHandleStringConcatenation(arguments, context);
+      default:
+        return null;
+    }
+  }
```
LantaoJin: This fixing can address the request but the exception logic is tricky to me. It skips the TypeChecker. Can we refactor it similar to https://github.com/opensearch-project/sql/pull/3821? cc @yuancu for further suggestions

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
```diff
@@ -354,4 +354,24 @@ public void testDependedLateralEval() {
             + "GROUP BY `DEPTNO`";
     verifyPPLToSparkSQL(root, expectedSparkSql);
   }
+
+  @Test
+  public void testEvalStringConcatenationWithPlus() {
```
Swiddis: Do these tests cover what happens if there are null entries in the dataset?

ahkcs: I have added more tests to cover null entries scenario

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ahkcs: I have refactored

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -760,8 +760,36 @@ void populate() {
       registerOperator(AND, SqlStdOperatorTable.AND);
       registerOperator(OR, SqlStdOperatorTable.OR);
       registerOperator(NOT, SqlStdOperatorTable.NOT);
-      registerOperator(ADD, SqlStdOperatorTable.PLUS);
-      registerOperator(ADDFUNCTION, SqlStdOperatorTable.PLUS);
+
+      // Register ADD with custom implementation that handles both numeric and string types
+      register(
```
qianheng-aws, line 694: I would recommend registering  `+` for string in a separate function signature. The current builder should support registering multiple times for the same function name with different signatures. And then its type checker should be simpler, should using PPLFamilyTypeChecker directly.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -760,8 +760,36 @@ void populate() {
...
+
+                if (arg1IsString && arg2IsString) {
+                  return builder.makeCall(SqlStdOperatorTable.CONCAT, arg1, arg2);
+                } else {
+                  return builder.makeCall(SqlStdOperatorTable.PLUS, arg1, arg2);
+                }
+              },
+              new PPLTypeChecker.PPLAddTypeChecker()));
+      register(
+          ADDFUNCTION,
```
qianheng-aws: Do we really want register `ADD` function for string?

qianheng-aws (changes requested):

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -760,8 +760,36 @@ void populate() {
       registerOperator(AND, SqlStdOperatorTable.AND);
       registerOperator(OR, SqlStdOperatorTable.OR);
       registerOperator(NOT, SqlStdOperatorTable.NOT);
-      registerOperator(ADD, SqlStdOperatorTable.PLUS);
-      registerOperator(ADDFUNCTION, SqlStdOperatorTable.PLUS);
+
+      // Register ADD with custom implementation that handles both numeric and string types
+      register(
+          ADD,
+          createFunctionImpWithTypeChecker(
```
yuancu: This function is removed in #4024 as it is unnecessarily involved. Please refactor accordingly following #4024.

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
```diff
@@ -601,4 +601,71 @@ private static String formatExprSignatures(List<List<ExprType>> signatures) {
                     + "]")
         .collect(Collectors.joining(","));
   }
+
+  /**
+   * A custom type checker for the ADD operator that supports both numeric addition and string
+   * concatenation.
+   */
+  class PPLAddTypeChecker implements PPLTypeChecker {
```
yuancu: Please remove this from `PPLTypeChecker` interface. Please use `PPLTypeChecker.wrapComposite` instead.

yuancu (changes requested):
I suggest you re-implement this PR in the following way (after #4024 is approved):

```java
register(
        ADD,
        (RexBuilder builder, RexNode... args) ->
                builder.makeCall(SqlStdOperatorTable.PLUS, args),
        PPLTypeChecker.wrapFamily(OperandTypes.NUMERIC_NUMERIC));
register(
        ADD,
        (RexBuilder builder, RexNode... args) ->
                builder.makeCall(SqlStdOperatorTable.CONCAT, args),
        PPLTypeChecker.wrapFamily(OperandTypes.STRING_STRING));
```


This registers the same function with two different implementations, controlled via type checkers.


P.s. please make the PR title more specific

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
qianheng-aws: I think just `registerOperator(PLUS, SqlStdOperatorTable.CONCAT);` could implement this without specifying type checker, does it? @yuancu 

`registerOperator` already has capability to generate type checker itself automatically.

yuancu: You are right, I updated my suggestions on implementation to the following:

```java
register(
        ADD,
        (RexBuilder builder, RexNode... args) ->
                builder.makeCall(SqlStdOperatorTable.PLUS, args),
        PPLTypeChecker.wrapFamily(OperandTypes.NUMERIC_NUMERIC));

registerOperator(ADD, SqlStdOperatorTable.CONCAT);
```

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: Removed

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: Updated implementation

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
ahkcs: Removed

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: Updated implementation

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -689,7 +689,21 @@ void populate() {
...
+          ADD,
+          (RexBuilder builder, RexNode... args) -> builder.makeCall(SqlStdOperatorTable.PLUS, args),
+          new PPLTypeChecker.PPLFamilyTypeChecker(SqlTypeFamily.NUMERIC, SqlTypeFamily.NUMERIC));
+
+      // Register ADD (+ symbol) for string concatenation
+      register(
+          ADD,
+          (RexBuilder builder, RexNode... args) ->
+              builder.makeCall(SqlStdOperatorTable.CONCAT, args),
+          new PPLTypeChecker.PPLFamilyTypeChecker(SqlTypeFamily.STRING, SqlTypeFamily.STRING));
```
yuancu: Is there any specific reason that you did not leverage `registerOperator(ADD, SqlStdOperatorTable.CONCAT);`?

> P.s. please make the PR title more specific. I.e. what enhancement it is.

A reminder ⬆️

P.s. please take some rest on Sunday 😂

ahkcs: I assumed both registrations needed explicit type checking, I have made changes according to your suggestion

P.S. Thanks for the reminder about rest! 😄 Will update the PR title to be more specific.

Swiddis (dismissed):

ykmr1224 (dismissed):

yuancu (commented):
I think there should be an integration test to verify result correctness of string concatenation with +.

LantaoJin (dismissed):

LantaoJin:
> I think there should be an integration test to verify result correctness of string concatenation with +.

@ahkcs normally, we don't use `CalcitePPL*Test` to verify the result since the `CalcitePPL*Test` leverages Calcite catalog instead of ours which may have some logic gaps. `CalcitePPL*Test` is useful to verify the pure Calcite plan and SparkSQL generated. Instead, you can add an integration test in `CalcitePPL*IT`

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
yuancu: I doubt that these tests really cover cases where null entries exist -- these tests only verify the correctness of plans. It's only during the execution of these plans that one will know what happens when there are null operands.

ahkcs:
@qianheng-aws @yuancu Added IT tests

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
```diff
@@ -354,4 +354,102 @@ public void testDependedLateralEval() {
...
+
+    String expectedSparkSql =
+        "SELECT `EMPNO`, `ENAME`, `JOB`, `ENAME` || ' ' || `JOB` `full_name`\n"
+            + "FROM `scott`.`EMP`\n"
+            + "LIMIT 3";
+    verifyPPLToSparkSQL(root, expectedSparkSql);
+  }
+
+  @Test
+  public void testEvalStringConcatenationWithNullableField() {
```
yuancu: Can be removed since concatenating string with null operands is actually verified in IT

LantaoJin (dismissed):

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
ahkcs: Removed

qianheng-aws (approved):

LantaoJin (approved):

Issue #4005, opened 2025-08-07
`coalesce` PPL function (Enhancement - Calcite)


aalva500-prog:
# `coalesce` (Enhancement - Calcite)

### Description
This RFC proposes enhancements to the existing PPL `coalesce` function to improve usability and compatibility. Currently, PPL `coalesce` enforces strict type matching and requires existing fields. The proposed enhancements include: automatic type coercion for mixed data types, support for non-existent fields, and empty field handling. These improvements will make `coalesce` more flexible while maintaining its core functionality of returning the first non-null value from a list of arguments.

### Use Cases
- Field normalization across different data sources (e.g., "clientip" vs "ipaddress")
- Handling missing data with fallback values
- Consolidating similar fields from different systems into a unified field

## 2. Final Syntax

```
source=<index> | eval <field>=coalesce(<field1>, <field2>, ..., <fieldN>)
```

### Parameters
- `<field>`: The output field name to store the coalesced result
- `<field1>, <field2>, ..., <fieldN>`: Field names or literal values to evaluate in order
- Supports mixed data types with automatic type coercion
- Can include literal values as fallback options
- Allows non-existent fields and empty field handling

### Options
- Mixed field names and literal values (no type constraints)
- Support for all data types (string, numeric, etc.) with automatic coercion
- Unlimited number of arguments
- Support for empty fields (" ")
- Support for non-existent/dummy fields

### Basic Usage
```
PPL> source=accounts | eval result = coalesce(employer, firstname, lastname) | fields result, firstname, lastname, employer
```

### Advanced Usage
```
PPL> source=firewall | eval src_ip = coalesce(src_ip,sourceip,source_ip,sip,ip) | fields src_ip
```

```
PPL> source=logs | eval ip = coalesce(clientip, ipaddress, "unknown") | fields ip
```

```
PPL> source=data | eval result = coalesce(field1, 123, "fallback") | fields result
```

### Best Practices
- Use `coalesce` for field normalization before aggregation operations
- Mix field names and literal values as needed with automatic type handling
- Place literal fallback values at the end of the argument list
- Leverage empty field support for comprehensive data handling

### Optimization Notes
- Place coalesce operations early in the query pipeline when possible to reduce processing overhead
- Consider using coalesce for field normalization before aggregation operations

### Technical Approach
[...]

PR #4041, opened 2025-08-14
PPL Coalesce Function Enhancement for Calcite - Advanced Type Handling and Field Support

aalva500-prog:
# PPL Coalesce Function Enhancement for Calcite - Advanced Type Handling and Field Support

## Description

This PR implements significant enhancements to the existing PPL `coalesce` function in Calcite to address current limitations with strict type checking and field requirements. The enhancement transforms coalesce from a restrictive function that requires same-type arguments and existing fields to a flexible function that handles mixed data types, non-existent fields, and empty values gracefully.

### Features Implemented

**✅ Feature 1: Automatic Type Coercion**
- Removed strict same-type requirement for all arguments
- Implemented intelligent type coercion for mixed data types
- Example: `source=data | eval result = coalesce(field1, 123, "fallback")`

**✅ Feature 2: Non-existent Field Support**
- Added support for dummy/non-existent fields in coalesce arguments
- Graceful handling of missing fields without runtime errors
- Example: `source=firewall | eval src_ip = coalesce(src_ip, sourceip, dummy_field, "unknown")`

**✅ Feature 3: Empty Field Handling**
- Support for empty strings ("") and whitespace strings (" ") as valid values
- Empty strings are treated as valid non-null values, not skipped
- Example: `source=accounts | eval result = coalesce("", firstname)` returns ""

### Calcite Changes
- Updated coalesce function to support mixed data types with automatic type coercion
- Enhanced field resolution to handle non-existent fields without errors
- Modified value processing to treat empty strings as valid non-null values

## Usage Examples

```ppl
# Feature 1: Automatic Type Coercion
source=data | eval result = coalesce(field1, 123, "fallback")
source=logs | eval value = coalesce(numeric_field, "0", 999)

# Feature 2: Non-existent Field Support  
source=firewall | eval src_ip = coalesce(src_ip, sourceip, dummy_field, "unknown")
source=logs | eval ip = coalesce(clientip, ipaddress, non_existent_field, "127.0.0.1")

# Feature 3: Empty Field Handling
source=accounts | eval result = coalesce("", firstname)        # Returns ""
source=accounts | eval result = coalesce(" ", firstname)       # Returns " "
source=accounts | eval result = coalesce(firstname, "")        # Returns firstname or ""

# Combined Features Example
source=firewall | eval src_ip = coalesce(src_ip, sourceip, dummy_field, "unknown")  # Uses all 3 features
```
### Testing Plan
- **Type Coercion**: Test mixed data type handling and automatic conversions
- **Non-existent Fields**: Test graceful handling of missing/dummy fields
- **Empty Values**: Test empty string and whitespace processing
- **Backward Compatibility**: Verify existing queries continue to work
- **Performance**: Validate no significant performance degradation

## Performance Impact

Type coercion and field resolution changes may have minimal performance overhead. Performance testing will be conducted with big5 dataset to ensure acceptable impact.

## Breaking Changes

**None** - This PR maintains full backward compatibility while adding enhanced functionality.

## Functionality Improvements

| Feature | Before Enhancement | After Enhancement |
|---------|-------------------|-------------------|
| Mixed Types | ❌ Runtime Error | ✅ Automatic Coercion |
| Non-existent Fields | ❌ Runtime Error | ✅ Graceful Handling |
| Empty Strings | ❌ Not Supported | ✅ Treated as Valid |
| Type Safety | ✅ Strict Checking | ✅ Intelligent Coercion |

## Documentation

PPL command documentation is updated in this PR to reflect the new coalesce behavior including type coercion, non-existent field handling, and empty value support.

## Best Practices

- **Field Ordering**: Place most specific/reliable fields first in coalesce arguments
- **Type Consistency**: While mixed types are supported, prefer consistent types when possible
- **Fallback Values**: Always include a literal fallback value as the last argument
- **Performance**: Use coalesce early in the query pipeline for optimal performance

## BIG5 COALESCE PERFORMANCE TEST
Dataset: big5
Test Date: Tue Aug 19 2025

### Baseline:
`source=big5 | head <size>`
head 10: Avg = 19ms, P90 = 25ms
head 100: Avg = 32ms, P90 = 38ms
head 1K: Avg = 123ms, P90 = 124ms
head 10K: Avg = 1170ms, P90 = 1222ms
head 100K: Avg = 1109ms, P90 = 1134ms

`source=big5 | eval result = coalesce(`host.name`, 'unknown') | head <size>`
head 10: Avg = 37ms, P90 = 40ms
head 100: Avg = 40ms, P90 = 44ms
head 1K: Avg = 147ms, P90 = 164ms
head 10K: Avg = 1167ms, P90 = 1177ms
head 100K: Avg = 1160ms, P90 = 1196ms

`source=big5 | eval result = coalesce(`host.name`, `agent.name`, 'unknown') | head <size>`
head 10: Avg = 29ms, P90 = 33ms
head 100: Avg = 46ms, P90 = 49ms
head 1K: Avg = 157ms, P90 = 199ms
head 10K: Avg = 1194ms, P90 = 1221ms
head 100K: Avg = 1166ms, P90 = 1203ms

`source=big5 | eval result = coalesce(`host.name`, 'unknown') | fields result, `host.name` | head <size>`
head 10: Avg = 25ms, P90 = 27ms
head 100: Avg = 28ms, P90 = 28ms
head 1K: Avg = 56ms, P90 = 57ms
head 10K: Avg = 367ms, P90 = 391ms
head 100K: Avg = 347ms, P90 = 354ms

### Feature testing:
`source=big5 | eval result = coalesce(`host.name`, 123, 'fallback') | head <size>`
head 10: Avg = 33ms, P90 = 37ms
head 100: Avg = 48ms, P90 = 52ms
head 1K: Avg = 166ms, P90 = 196ms
head 10K: Avg = 1138ms, P90 = 1163ms
head 100K: Avg = 1181ms, P90 = 1206ms

`source=big5 | eval result = coalesce(`host.name`, dummy_field, 'unknown') | head <size>`
head 10: Avg = 32ms, P90 = 38ms
head 100: Avg = 40ms, P90 = 43ms
head 1K: Avg = 135ms, P90 = 137ms
head 10K: Avg = 1193ms, P90 = 1239ms
head 100K: Avg = 1185ms, P90 = 1247ms

`source=big5 | eval result = coalesce('', `host.name`) | head <size>`
head 10: Avg = 45ms, P90 = 62ms
head 100: Avg = 37ms, P90 = 39ms
head 1K: Avg = 148ms, P90 = 172ms
head 10K: Avg = 1177ms, P90 = 1177ms
head 100K: Avg = 1143ms, P90 = 1152ms

## Related Issues
Adresses #[4005](https://github.com/opensearch-project/sql/issues/4005)

### Check List
- [✅] New functionality includes testing.
- [✅] New functionality has been documented.
- [✅] New functionality has javadoc added.
- [✅] New functionality has a user manual doc added.
- [✅] Commits are signed per the DCO using `--signoff`.
- [✅] Public documentation issue/PR [created](https://github.com/opensearch-project/sql/issues/4005).


dai-chen (commented):
I have some high level questions:

1. Is Automatic Type Coercion applied to all other functions as well? I'm just thinking 
2. Besides UDF, did we consider other approaches? e.g., handle nonexistent or empty field in Analyzer layer?

aalva500-prog:
Hi @dai-chen, thanks for raising these points! Please find below my answers, as well as implementation and design decision:

### 1) Is Automatic Type Coercion applied to all other functions as well?
As far as I'm concerned, no, automatic type coercion is NOT applied to all other functions. I'm not aware if the enhancements for the coalesce function are going to be applied to other functions. We may need to verify and confirm with the team.

Currently, the below is what we have:

**For condition functions:**
- COALESCE & IFNULL: Get Calcite's automatic type coercion
- IS_NULL, IS_NOT_NULL, ISPRESENT: Don't need type coercion (single operand)
- EARLIEST, LATEST: Custom UDFs without automatic coercion
- IF, CASE: Use manual type checking

**For `eval` related functions (Legacy engine):**
- No automatic type coercion - require strict type matching
- Use manual type checking and casting
- Implemented in the legacy expression evaluation system

**Other functions WITHOUT Automatic Type Coercion:**
- UDF Functions: Use strict type checking via PPLTypeChecker
- Math Functions: Implement manual type handling (e.g., DivideFunction)
- Custom Functions: Require explicit type matching

### 2) Besides UDF, did we consider other approaches? e.g., handle nonexistent or empty field in Analyzer layer?
Yes, other approaches were considered:
1. Analyzer Layer Handling (like eval functions):
    - Yes, this could handle non-existent fields in ExpressionAnalyzer.visitIdentifier()
    - However, it would require changes to core analysis pipeline
    - Also, it would work for legacy engine but not leverage Calcite's capabilities

2. Native Calcite Integration: Direct use of SqlStdOperatorTable.COALESCE
    - This is the standard Calcite operator
    - Type inference occurs in Calcite's leastRestrictive() method - Called by Calcite internally
    - Error thrown by Calcite's type system - When it cannot find a common type
       ```
        Cannot infer return type for COALESCE; operand types: [VARCHAR, INTEGER, VARCHAR]
       ```

### 3) Why UDF Approach Was Chosen:
- Separation of Concerns: Field handling is done at the visitor level, not analyzer level. Type coercion is done at Calcite level
- Calcite Integration: Leverages Calcite's engine. UDF bridges PPL syntax to Calcite's native operators seamlessly
- Type System Compatibility: Leverages Calcite's sophisticated type coercion automatically
- Minimal Impact: Doesn't require changes to core analysis pipeline
- Performance: Calcite's optimized type handling vs manual eval function logic
- Future-proof: Works with Calcite engine evolution

**The implementation shows a hybrid approach:**
- Field resolution handled in CalciteRexNodeVisitor (visitor layer)
- Type coercion delegated to Calcite's native COALESCE operator
- UDF wrapper provides the bridge between PPL and Calcite

### 4) Why Creating Our Own Enhanced Coalesce Function:

The data type evaluation happens entirely within Calcite's core libraries, not in this project's codebase. This project only calls SqlStdOperatorTable.COALESCE, and Calcite handles the rest.

Currently COALESCE Data Type Evaluation in this Project works in the following way:

1. Type Checking:
The following error  comes from Calcite's core libraries:
  ```
  Cannot infer return type for COALESCE; operand types: [VARCHAR, INTEGER, VARCHAR]
  ```

The Calcite libraries are external dependencies that contain the rigid type system causing the mixed-type issues.

2. How COALESCE Type Evaluation Works and Where It Happens:
    * Calcite's SqlStdOperatorTable.COALESCE - This is the standard Calcite operator. 
    * Type inference occurs in Calcite's leastRestrictive() method - Called by Calcite internally
    * Error thrown by Calcite's type system - When it cannot find a common type

3. Currently, in this Project's Code:
    - PPLFuncImpTable.java
    ```
     return builder.makeCall(SqlStdOperatorTable.COALESCE, processedArgs);
     ```

    - ExtendedRexBuilder.java:
    ```
    public RelDataType commonType(RexNode... nodes) {
         return this.getTypeFactory()
             .leastRestrictive(Arrays.stream(nodes).map(RexNode::getType).toList());
    }
    ```

4. Calcite Libraries Location:
    * Gradle Dependencies (build.gradle):
        - api('org.apache.calcite:calcite-core:1.38.0')
        - api 'org.apache.calcite:calcite-linq4j:1.38.0'

   * Physical JAR Files:
      ```
       ~/.gradle/caches/modules-2/files-2.1/org.apache.calcite/calcite-core/1.38.0/
      ├── calcite-core-1.38.0.jar                    # Compiled classes
      ├── calcite-core-1.38.0-sources.jar           # Source code
     └── calcite-core-1.38.0-javadoc.jar           # Documentation
      ```

5. COALESCE Implementation Files:
    * SqlCoalesceFunction.java - Main COALESCE function definition
    * RexImpTable$CoalesceImplementor.class - Runtime implementation

6. Key Code in SqlCoalesceFunction.java:
    - SqlCoalesceFunction
      ```
      public SqlCoalesceFunction() {
        super("COALESCE",
            SqlKind.COALESCE,
            ReturnTypes.LEAST_RESTRICTIVE.andThen(SqlTypeTransforms.LEAST_NULLABLE), // ← TYPE INFERENCE HERE
      null,
            OperandTypes.SAME_VARIADIC,  // ← REQUIRES SAME TYPES
            SqlFunctionCategory.SYSTEM);
      }
      ```
    
     - COALESCE Rewrite Logic:
        ```
          @Override public SqlNode rewriteCall(SqlValidator validator, SqlCall call) {
              // COALESCE gets rewritten to CASE statement:
          // COALESCE(a, b, c) becomes:
          // CASE WHEN a IS NOT NULL THEN a
          //      WHEN b IS NOT NULL THEN b  
          //      ELSE c END
          }
        ```

    - The Problem:
        * ReturnTypes.LEAST_RESTRICTIVE - Calls leastRestrictive() method
        * OperandTypes.SAME_VARIADIC - - Expects all operands to be the same type
        * When mixed types like [VARCHAR, INTEGER, VARCHAR]are passed:
            - leastRestrictive() returns null
            - Calcite throws "Cannot infer return type for COALESCE"

 6. The Solution Needed:
      To fix this, we would need to either:
      * Modify Calcite's type inference system (not feasible)
      * Register our custom function instead of using SqlStdOperatorTable.COALESCE (This is the approach I'm implementing)

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
```diff
@@ -0,0 +1,74 @@
...
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return opBinding -> {
+      for (int i = 0; i < opBinding.getOperandCount(); i++) {
+        if (!opBinding.isOperandNull(i, false)) {
+          return opBinding.getOperandType(i);
+        }
+      }
+      return opBinding.getOperandCount() > 0
+          ? opBinding.getOperandType(0)
+          : opBinding.getTypeFactory().createSqlType(SqlTypeName.VARCHAR);
```
yuancu: (I haven't read through the implementation, just came across this PR.)

A high level question: why are we allowing incompatible types for coalesce function? How will it know what type should the result be in advance?

For example, in `eval res = coalesce(string, integer, ip, geopoint, date, timestamp, time, nested, struct)`, there are null values for different columns in different rows, `res` may be either of the 9 types. Then what should be the type of the final result column `res`?

ahkcs:
For this test: `CrossClusterCoalesceIT.java`
I think we should add our test cases in `CrossClusterSearchIT` instead of creating our own ones?

aalva500-prog:
@ahkcs If the command/function enhancements you are implementing only works when Calcite is enabled, adding test cases in this `CrossClusterSearchIT` file will give errors. Maybe we can create a cross-cluster test file for Calcite only and add all our tets there.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -544,6 +561,7 @@ public RexNode visitInSubquery(InSubquery node, CalcitePlanContext context) {
           "The number of columns in the left hand side of an IN subquery does not match the number"
               + " of columns in the output of subquery");
     }
+
```
RyanL1997: Lets remove the unnecessary change

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -622,9 +640,6 @@ public RexNode visitCase(Case node, CalcitePlanContext context) {
     return context.rexBuilder.makeCall(SqlStdOperatorTable.CASE, caseOperands);
   }
 
-  /*
```
RyanL1997: any reson we need to remove this?

core/src/main/java/org/opensearch/sql/calcite/utils/UserDefinedFunctionUtils.java
```diff
@@ -344,4 +345,18 @@ public UDFOperandMetadata getOperandMetadata() {
       }
     };
   }
+
+  public static org.opensearch.sql.data.model.ExprValue enhancedCoalesce(
+      org.opensearch.sql.data.model.ExprValue... args) {
+    for (org.opensearch.sql.data.model.ExprValue arg : args) {
```
RyanL1997: nit: maybe not inline import

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    return ExprValueUtils.nullValue();
+  }
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return opBinding -> {
+      for (int i = 0; i < opBinding.getOperandCount(); i++) {
+        if (!opBinding.isOperandNull(i, false)) {
+          return opBinding.getOperandType(i);
```
RyanL1997: To me, this returns the type of the first non-null operand, not the first non-null value. At compile time, you don't know which value will be non-null at runtime.

e.g.
```bash
coalesce(null_field, "string", 123)
```
If `null_field` is nullable string type, this returns `STRING`, but what if at runtime `null_field` is `null` and "string" is also `null`? Then `123` (`INTEGER`) would be returned, causing a type mismatch.

Let me know if I missed some context.

core/src/main/java/org/opensearch/sql/calcite/utils/UserDefinedFunctionUtils.java
```diff
@@ -344,4 +345,18 @@ public UDFOperandMetadata getOperandMetadata() {
...
+      if (arg != null && !arg.isNull() && !arg.isMissing()) {
+        // Check for empty strings
+        if (arg.type().typeName().equals("STRING") && arg.stringValue().trim().isEmpty()) {
```
RyanL1997: - Is that possible to enum comparison instead of `equals()`?
- Should we do any safety check to make sure the conversion is safe before calling `stringValue()`?
- `trim().isEmpty()` means `" "` becomes empty - is this intended?

RyanL1997 (commented):
Hi @aalva500-prog , thanks for taking this on, and I just left some comments.

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
aalva500-prog: You are right to question this @yuancu. Please see my answer below and thanks for your insights!

The current implementation replicates SPL's behavior, where `coalesce` performs no type checking and returns the value and data type of the first non-null argument.

Your point about type safety is well-taken. The current implementation has a fundamental limitation: it permits incompatible types and uses a "first non-null type" approach for return type inference, which can cause runtime type errors.

The challenge is that Calcite requires compile-time type determination, but with mixed incompatible types, the actual return type depends on runtime data availability.

To address this concern, I propose the following approach:

1. Try to find a least restrictive common type for compatible types
2. If that fails (incompatible types), fall back to VARCHAR/STRING as the universal type

That being explained, while your example `coalesce(string, integer, ip, geopoint, date, timestamp, time, nested, struct)` is syntactically valid in SPL, it represents an uncommon use case. In practice, coalesce operations with such diverse data types may encounter type compatibility issues and should be used judiciously. I'll keep you posted when I implement a new solution, thanks again!

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
aalva500-prog: Not really, I'll put it back, thanks for catching this!

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
aalva500-prog: You are right, however, your description matches with how SPL works with different data types in `coalesce`. Nevertheless, I think I can improve this logic and handle it differently by doing an automatic type coercion, thanks for bringing this up!

core/src/main/java/org/opensearch/sql/calcite/utils/UserDefinedFunctionUtils.java
aalva500-prog: sure, I can do that, np!

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
aalva500-prog: sounds good!

core/src/main/java/org/opensearch/sql/calcite/utils/UserDefinedFunctionUtils.java
aalva500-prog: Hi @RyanL1997, thanks for these observations, please find below the changes I'll perform and the answer to your question about `trim.isEmpty()`:

1. Enum comparison: I'll change it from `arg.type().typeName().equals("STRING")` to `ExprCoreType.STRING.equals(arg.type())` which is more type-safe and efficient.

2. Safety check: The enum comparison `ExprCoreType.STRING.equals(arg.type())` inherently provides the safety check - Then I'll only call stringValue() when I've confirmed the type is STRING.

3. `trim().isEmpty()` behavior: Yes, this is intentional. The enhanced coalesce skips whitespace-only strings treating them as "empty" values to find the next non-empty argument. What I want in this case is to skip not just null/missing values but also empty strings.

aalva500-prog: @RyanL1997 made some changes to my code, please review, thanks!

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
aalva500-prog: @RyanL1997 made some changes to my code, please review, thanks!

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
aalva500-prog: @yuancu made some changes to my code, please review, thanks!

integ-test/src/test/java/org/opensearch/sql/security/CrossClusterCoalesceIT.java
```diff
@@ -0,0 +1,98 @@
...
+import static org.opensearch.sql.util.MatcherUtils.verifyColumn;
+
+import java.io.IOException;
+import lombok.SneakyThrows;
+import org.json.JSONObject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.opensearch.sql.ppl.PPLIntegTestCase;
+
+/** Cross Cluster Search tests for enhanced coalesce function with Calcite. */
```
RyanL1997: nit: remove these verbose comments

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -466,25 +472,37 @@ private List<RelDataType> modifyLambdaTypeByFunction(
...
-    for (UnresolvedExpression arg : args) {
-      if (arg instanceof LambdaFunction) {
-        CalcitePlanContext lambdaContext =
-            prepareLambdaContext(
-                context, (LambdaFunction) arg, arguments, node.getFuncName(), null);
-        RexNode lambdaNode = analyze(arg, lambdaContext);
-        if (node.getFuncName().equalsIgnoreCase("reduce")) { // analyze again with calculate type
-          lambdaContext =
+
+    boolean isCoalesce = "coalesce".equalsIgnoreCase(node.getFuncName());
```
RyanL1997, line 476: The `inCoalesceFunction` flag in context might not be thread-safe if the visitor is used concurrently. Also, consider if nested `COALESCE` calls are properly handled - the flag might get incorrectly reset if `COALESCE` is nested within another `COALESCE`. So can you confirm the above?

RyanL1997:
Also, may need a rebase and resolve the conflict

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLFunctionTypeTest.java
```diff
@@ -52,14 +52,18 @@ public void testComparisonWithDifferentType() {
             + " but got [STRING,INTEGER]");
   }
 
+  @Test
+  public void testCoalesceWithSameType() {
+    String ppl = "source=EMP | eval coalesce_name = coalesce(ENAME, 'Jack') | fields coalesce_name";
+    getRelNode(ppl);
```
dai-chen: No assertion?

docs/user/ppl/functions/condition.rst
```diff
@@ -267,6 +274,56 @@ Example::
...
+    +---------+----------+---------+
+    | result  | employer | balance |
+    |---------+----------+---------|
+    | Pyrami  | Pyrami   | 39225   |
+    | Netagy  | Netagy   | 32838   |
+    | Quility | Quility  | 4180    |
+    | 5686    | null     | 5686    |
+    +---------+----------+---------+
+
+Non-existent Field Handling::
```
dai-chen, line 314: Also highlight this in Behaviors section above?

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
```diff
@@ -0,0 +1,108 @@
...
+              .map(
+                  operand ->
+                      (Expression)
+                          Expressions.call(
+                              ExprValueUtils.class,
+                              "fromObjectValue",
+                              Expressions.convert_(operand, Object.class)))
+              .toList();
+
+      Expression returnTypeName = Expressions.constant(call.getType().getSqlTypeName().toString());
```
dai-chen, line 40: "the least restrictive common type among all arguments", this is determined by Calcite?

dai-chen (approved):
Thanks for the changes!

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
aalva500-prog: @RyanL1997 Thanks for raising this point. The current implementation handles your scenario with nested coalesce. Please see the screenshots below for reference.

1) non-existing field in outer coalesce correctly returns field in inner coalesce:
Screenshot 2025-08-28 at 11 06 42 AM

2) existing field in outer coalesce is correctly returned:
Screenshot 2025-08-28 at 11 09 12 AM 

3) non-existing field in both inner/outer coalesce:
Screenshot 2025-08-28 at 11 29 50 AM

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
aalva500-prog: Hi @dai-chen, yes, the "least restrictive common type" is determined by Calcite's built-in type system in the `getReturnTypeInference()` method. The type inference does not happen in `createImplementor()`. The type inference happens earlier during Calcite's query planning phase when it calls `getReturnTypeInference()`. If Calcite can't find a common type, it falls back to `VARCHAR` as the most permissive type that can represent any value as a string. So `createImplementor()` doesn't determine the type - it just accesses the type that was already computed by the `getReturnTypeInference()` method during an earlier planning phase. The `createImplementor()` method generates the runtime execution code that uses the pre-determined type.

#### Summary of how EnhancedCoalesceFunction works:
1) Function Registration: In `PPLBuiltinOperators.java` as `ENHANCED_COALESCE = new EnhancedCoalesceFunction().toUDF("COALESCE")`
2) UDF Creation: Called by `UserDefinedFunctionBuilder.toUDF()` to create the `SqlUserDefinedFunction`
3) Query Planning: Calcite uses it during type inference to determine the `COALESCE` result type
4) Type Inference Flow:
     - Planning Phase: `getReturnTypeInference()` → Calcite's `leastRestrictive()` determines common type (e.g., `INTEGER`, `VARCHAR`)
    - Code Generation: `createImplementor()` → `call.getType()` retrieves the already-determined type
    - Runtime: `enhancedCoalesceWithType()` receives the type name and coerces the result accordingly

RyanL1997 (approved):

PR #4044, opened 2025-08-15
Redefine type checkers of isnull, isnotnull, and ispresent to IGNORE

### LINKED ISSUE #4004: [BUG][Calcite] IsNotNull doesn't support struct type
**What is the bug?**
 IsNotNull doesn't support struct type

1. Create index with struct field:
```
PUT {{baseUrl}}/test/
{
  "properties": {
    "object_field": {
      "type": "object"
    }
}
```

2. Run ppl:
```
source=test | where isNotNull(object_field)
```
and it will throw exception:
```
{
  "error": {
    "reason": "Invalid Query",
    "details": "IS_NOT_NULL function expects {[ANY]}, but got [STRUCT]",
  },
  "status": 400
}
```
while it's supported in v2 engine.

It should run successfully

 - Plugins

It's blocked by TypeChecker, because ANY is incompatible with MAP in Calcite.

yuancu:
### Description

Update the type checkers of isnull, isnotnull, and ispresent to support null checking of struct types (and possibly map types).

### Related Issues
Resolves #4004

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchIndexScanRule.java
```diff
@@ -62,4 +69,39 @@ static boolean isLogicalSortLimit(LogicalSort sort) {
...
+   * OpenSearch DSL does not handle these conditions correctly.
+   *
+   * @param filter The filter operation to check
+   * @return true if the condition should be excluded from pushdown, false otherwise
+   */
+  static boolean isConditionExcluded(Filter filter) {
+    RexNode condition = filter.getCondition();
+    if (condition instanceof RexCall conditionCall) {
+      return FILTER_PUSHDOWN_EXCLUDED_FUNCTION.contains(conditionCall.getKind())
+          && conditionContainsNestedField(condition, filter.getRowType().getFieldList());
```
LantaoJin: seems this checker only cover a simple Filter such as `where isnull(a.b)`, rather than a complex filter expression such as `where a.b = 'xx' and isnull(a.c)`

yuancu: Thanks for reminding! I'll double check.

yuancu: Fixed

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1050,6 +1048,20 @@ void populate() {
                               builder.makeLiteral(" "),
                               arg))),
           PPLTypeChecker.family(SqlTypeFamily.ANY));
+      // Re-define the type checker for is not null, is present, and is null since their original
+      // type checker ANY isn't compatible with struct types.
+      register(
```
qianheng-aws: How about adding a new method like `registerOperator(FUNC_NAME, OPERATOR, TYPE_CHECKER)` to simplify this?

yuancu: Did not do so since I worried that it might mix levels of abstraction. Yet having such a method does simplify many duplicated snippets. Updated the implementation.

LantaoJin (approved):

PR #4056, opened 2025-08-18
Implement `reverse` performance optimization

### LINKED ISSUE #3924: [FEATURE] Support `reverse` pushdown with Calcite
**Problem**
The current implementation of `reverse` in PPL uses a ROW_NUMBER window function to invert row order, which is not pushed down to the data source. This results in local execution that can be costly on large datasets.

Without pushdown, queries using `reverse` on large datasets may suffer performance degradation due to local sorting and window function computation.

For example, the query source=big5 | reverse | head 10
fails because reverse is applied on the entire big5 dataset before head 10 is applied, so the dataset is too big to be reversed.

**Goals**
- Detect logical plan patterns where pushdown is possible.
- Flip sort direction in pushdown when applicable to implement reverse.
- Double reverses or 2 Consecutive reverses should be a no-op.
- Disallow or fallback gracefully if conditions are not met.
- Add configuration flags or warnings for use on large unbounded datasets.

`reverse` can be pushed down when there are preceding sort, with no filter and limit in between.
In other cases, reverse cannot be pushed down with the current implementation, since there isn't a __reverse_row_number__ in the index, unless we push down the window function as well.

**Alternatives**
- Tried sort by desc on _id but there may be performance issues on this field since the default is sort by asc on _id
- Can try sort by desc on _doc

**Additional context**
RFC to implement reverse: #3873 
Reverse implementation PR: #3867

selsong:
### Description
This PR updates the Calcite planner to handle `reverse` correctly by flipping sort directions in pushdown.  
It also documents and tests edge cases such as double or consecutive `reverse` usage, even though these are not expected in normal user queries.

### 1. Flip sort direction
**Query:**
```
source = accounts | sort + balance, - firstname | reverse
```

**Logical Plan:**
```
LogicalSystemLimit(sort0=[$3], sort1=[$1], dir0=[DESC-nulls-last], dir1=[ASC-nulls-first], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10])
  LogicalSort(sort0=[$3], sort1=[$1], dir0=[DESC-nulls-last], dir1=[ASC-nulls-first])
    LogicalSort(sort0=[$3], sort1=[$1], dir0=[ASC-nulls-first], dir1=[DESC-nulls-last])
      CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])
```

**Physical Plan:**
```
CalciteEnumerableIndexScan(table=[[OpenSearch, accounts]], PushDownContext=[[PROJECT->[account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname], SORT->[
  {
    \"balance\" : {
      \"order\" : \"desc\",
      \"missing\" : \"_last\"
    }
  },
  {
    \"firstname.keyword\" : {
      \"order\" : \"asc\",
      \"missing\" : \"_first\"
    }
  }
], LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={
  \"from\":0,
  \"size\":10000,
  \"timeout\":\"1m\",
  \"_source\":{
    \"includes\":[\"account_number\",\"firstname\",\"address\",\"balance\",\"gender\",\"city\",\"employer\",\"state\",\"age\",\"email\",\"lastname\"],
    \"excludes\":[]
  },
  \"sort\":[
    {\"balance\":{\"order\":\"desc\",\"missing\":\"_last\"}},
    {\"firstname.keyword\":{\"order\":\"asc\",\"missing\":\"_first\"}}
  ]
}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
```
Explanation:
The single reverse flips the original sort order (balance ASC, firstname DESC) to the reversed order (balance DESC, firstname ASC).

### 2. Double reverse
**Query:**
```
source = accounts | sort + balance, - firstname | reverse | reverse
```

**Logical Plan:**
```
LogicalSystemLimit(sort0=[$3], sort1=[$1], dir0=[ASC-nulls-first], dir1=[DESC-nulls-last], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10])
  LogicalSort(sort0=[$3], sort1=[$1], dir0=[ASC-nulls-first], dir1=[DESC-nulls-last])   <-- final sort
    LogicalSort(sort0=[$3], sort1=[$1], dir0=[DESC-nulls-last], dir1=[ASC-nulls-first]) <-- after 1st reverse
      LogicalSort(sort0=[$3], sort1=[$1], dir0=[ASC-nulls-first], dir1=[DESC-nulls-last]) <-- original sort
        CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])
```

**Physical Plan:**
```
CalciteEnumerableIndexScan(table=[[OpenSearch, accounts]], PushDownContext=[[PROJECT->[account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname], SORT->[
  {
    \"balance\" : {
      \"order\" : \"asc\",
      \"missing\" : \"_first\"
    }
  },
  {
    \"firstname.keyword\" : {
      \"order\" : \"desc\",
      \"missing\" : \"_last\"
    }
  }
], LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={
  \"from\":0,
  \"size\":10000,
  \"timeout\":\"1m\",
  \"_source\":{
    \"includes\":[\"account_number\",\"firstname\",\"address\",\"balance\",\"gender\",\"city\",\"employer\",\"state\",\"age\",\"email\",\"lastname\"],
    \"excludes\":[]
  },
  \"sort\":[
    {\"balance\":{\"order\":\"asc\",\"missing\":\"_first\"}},
    {\"firstname.keyword\":{\"order\":\"desc\",\"missing\":\"_last\"}}
  ]
}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
```
Explanation:
The two reverses cancel each other out. The final sort order is the same as the original query (balance ASC, firstname DESC).

### 3. Reverse after multiple sorts
**Query:**
```
source = accounts | sort + balance | sort - firstname | reverse
```

**Logical Plan:**
```
LogicalSystemLimit(sort0=[$1], dir0=[ASC-nulls-first], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10])
  LogicalSort(sort0=[$1], dir0=[ASC-nulls-first])
    LogicalSort(sort0=[$1], dir0=[DESC-nulls-last])
      LogicalSort(sort0=[$3], dir0=[ASC-nulls-first])
        CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])
```

**Physical Plan:**
```
CalciteEnumerableIndexScan(table=[[OpenSearch, accounts]], PushDownContext=[[PROJECT->[account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname], SORT->[
  {
    \"firstname.keyword\" : {
      \"order\" : \"asc\",
      \"missing\" : \"_first\"
    }
  }
], LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={
  \"from\":0,
  \"size\":10000,
  \"timeout\":\"1m\",
  \"_source\":{
    \"includes\":[\"account_number\",\"firstname\",\"address\",\"balance\",\"gender\",\"city\",\"employer\",\"state\",\"age\",\"email\",\"lastname\"],
    \"excludes\":[]
  },
  \"sort\":[
    {\"firstname.keyword\":{\"order\":\"asc\",\"missing\":\"_first\"}}
  ]
}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
```
Explanation:
The last sort + reverse overrides the earlier sort + balance. The final pushed-down order is firstname ASC.

### Related Issues
Resolves #3924 

### Check List


Swiddis (approved):
I'm curious about the cases where the double reverse optimization matters in practice, seems unlikely that someone would reverse something twice like that except by accident

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -360,20 +362,43 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
   @Override
   public RelNode visitReverse(
       org.opensearch.sql.ast.tree.Reverse node, CalcitePlanContext context) {
+    // Check if the child is also a Reverse node
+    if (node.getChild().size() == 1
+        && node.getChild().get(0) instanceof org.opensearch.sql.ast.tree.Reverse) {
```
dai-chen: Is logical optimization rule better way to do this? Or this is just edge case and usually users won't do `reverse | reverse`?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -360,20 +362,43 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
-            .rowsTo(RexWindowBounds.CURRENT_ROW)
-            .as(REVERSE_ROW_NUM);
-    context.relBuilder.projectPlus(rowNumber);
-    // Sort by row number descending
-    context.relBuilder.sort(context.relBuilder.desc(context.relBuilder.field(REVERSE_ROW_NUM)));
-    // Remove row number column
-    context.relBuilder.projectExcept(context.relBuilder.field(REVERSE_ROW_NUM));
+
+    RelCollation collation = context.relBuilder.peek().getTraitSet().getCollation();
+    if (collation == null || collation == RelCollations.EMPTY) {
```
dai-chen: Is this missing UT/IT coverage?

Also I recall we have UT to verify (Spark) SQL translated from PPL, could you add test there?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
penghuo: +1, calcite opt-rule does not cover this?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -360,20 +362,43 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
-            .over()
-            .rowsTo(RexWindowBounds.CURRENT_ROW)
-            .as(REVERSE_ROW_NUM);
-    context.relBuilder.projectPlus(rowNumber);
-    // Sort by row number descending
-    context.relBuilder.sort(context.relBuilder.desc(context.relBuilder.field(REVERSE_ROW_NUM)));
-    // Remove row number column
-    context.relBuilder.projectExcept(context.relBuilder.field(REVERSE_ROW_NUM));
+
+    RelCollation collation = context.relBuilder.peek().getTraitSet().getCollation();
```
songkant-aws: Be careful of using `getTraitSet()` to get fieldCollations because there is an edge case when it returns a different instance than `RelCollation`.

Sometimes, it returns a RelCompositeTrait instance containing all collation equivalent candidates. For example, this child PPL query `source=EMP | sort SAL | eval a = cast(SAL as double) | where a > 2` will return RelCompositeTrait instance like `[[5 ASC-nulls-first], [8]]`. The eighth projected column is a cast expression but has the equivalent collation with `[5 ASC-nulls-first]`. The traitSet put equivalences together into a RelCompositeTrait.

A better way is probably `List<RelCollation> collations =  context.relBuilder.getCluster().getMetadataQuery().collations(context.relBuilder.peek())`. You can pick the first valid RelCollation.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
songkant-aws: I think it's not necessary to check duplicate `Reverse` nodes. Calcite will ensure the top sort collation by overriding the child's collation with current one.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
songkant-aws: Calcite has some issue of carrying NullDirection when deducing collation for expressions.  See: https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/rel/metadata/RelMdCollation.java#L358-L360.

If we really care about reversing null direction, another option is we can find the closest sort node and reverse based on that collation.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: > I think it's not necessary to check duplicate `Reverse` nodes. Calcite will ensure the top sort collation by overriding the child's collation with current one.

For the previous implementation (adding row_number window function), Calcite can do nothing.

+1 to add a opt-rule, it could cover more cases:
```
... | reverse | reverse
... | reverse | fields ... | reverse
... | reverse | where ... | reverse
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: ref: https://github.com/opensearch-project/sql/blob/main/DEVELOPER_GUIDE.rst#new-ppl-command-checklist

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -567,19 +569,30 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
-            .as(REVERSE_ROW_NUM);
-    context.relBuilder.projectPlus(rowNumber);
-    // Sort by row number descending
-    context.relBuilder.sort(context.relBuilder.desc(context.relBuilder.field(REVERSE_ROW_NUM)));
-    // Remove row number column
-    context.relBuilder.projectExcept(context.relBuilder.field(REVERSE_ROW_NUM));
+    
+    // Check if there's an existing sort to reverse
+    List<RelCollation> collations = 
+        context.relBuilder.getCluster().getMetadataQuery().collations(context.relBuilder.peek());
```
RyanL1997, line 593: The `null` check for collations is sufficient, but accessing `collations.get(0)` could throw an `IndexOutOfBoundsException` if the list is empty. Maybe we can add an explicit check:

```java
RelCollation collation = collations != null && !collations.isEmpty() ? collations.get(0) : null;
if (collation != null && !collation.getFieldCollations().isEmpty()) {
    RelCollation reversedCollation = PlanUtils.reverseCollation(collation);
    context.relBuilder.sort(reversedCollation);
} else {
    // Fallback to ROW_NUMBER
}
```
This ensures no exception occurs if `collations` is empty.

core/src/main/java/org/opensearch/sql/calcite/plan/OpenSearchRules.java
```diff
@@ -8,12 +8,19 @@
...
+import org.opensearch.sql.calcite.rule.SortReverseOptimizationRule;
 
 public class OpenSearchRules {
   private static final PPLAggregateConvertRule AGGREGATE_CONVERT_RULE =
       PPLAggregateConvertRule.Config.SUM_CONVERTER.toRule();
 
-  public static final List<RelOptRule> OPEN_SEARCH_OPT_RULES =
+  public static final List<ConverterRule> OPEN_SEARCH_OPT_RULES = ImmutableList.of();
+  
+  public static final List<RelOptRule> OPTIMIZATION_RULES =
```
RyanL1997: nit: Consider using a more descriptive name like `OPEN_SEARCH_OPTIMIZATION_RULES` to align with `OPEN_SEARCH_OPT_RULES_OLD`.

core/src/main/java/org/opensearch/sql/calcite/plan/OpenSearchRules.java
```diff
@@ -8,12 +8,19 @@
...
   private static final PPLAggregateConvertRule AGGREGATE_CONVERT_RULE =
       PPLAggregateConvertRule.Config.SUM_CONVERTER.toRule();
 
-  public static final List<RelOptRule> OPEN_SEARCH_OPT_RULES =
+  public static final List<ConverterRule> OPEN_SEARCH_OPT_RULES = ImmutableList.of();
+  
+  public static final List<RelOptRule> OPTIMIZATION_RULES = 
+      ImmutableList.of(SortReverseOptimizationRule.INSTANCE);
+
+  public static final List<RelOptRule> OPEN_SEARCH_OPT_RULES_OLD =
```
RyanL1997: nit: This suggests a deprecation or transition. Add a comment explaining why this was renamed and whether it’s temporary or permanent.

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
```diff
@@ -0,0 +1,65 @@
...
+    LogicalSort optimizedSort = LogicalSort.create(
+        innerSort.getInput(),
+        outerSort.getCollation(),
+        outerSort.offset,
+        outerSort.fetch);
+    
+    call.transformTo(optimizedSort);
+  }
+  
+  private boolean hasSameField(LogicalSort outerSort, LogicalSort innerSort) {
```
RyanL1997: I saw that this method only checks the first field collation - may this miss cases where multiple fields are sorted? Or should we extend it to compare all fields?

I was thinking about something like this:
```java
private boolean hasSameField(LogicalSort outerSort, LogicalSort innerSort) {
    List<RelFieldCollation> outerFields = outerSort.getCollation().getFieldCollations();
    List<RelFieldCollation> innerFields = innerSort.getCollation().getFieldCollations();
    if (outerFields.isEmpty() || innerFields.isEmpty()) {
        return false;
    }
    return outerFields.size() == innerFields.size() &&
           IntStream.range(0, outerFields.size())
                   .allMatch(i -> outerFields.get(i).getFieldIndex() == innerFields.get(i).getFieldIndex());
}
```

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -217,6 +216,23 @@ public void testExplainWithReverse() throws IOException {
...
+
+  @Test
+  public void testExplainWithReversePushdownMultipleFields() throws IOException {
+    String query = "source=opensearch-sql_test_index_account | sort - age, + firstname | reverse";
+    var result = explainQueryToString(query);
+    String expected = loadFromFile("expectedOutput/calcite/explain_reverse_pushdown_multiple.json");
+    assertJsonEqualsIgnoreId(expected, result);
+  }
+
+<<<<<<< HEAD
```
RyanL1997: need to remove this

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -391,4 +394,36 @@ public Void visitInputRef(RexInputRef inputRef) {
...
+    if (original == null || original.getFieldCollations().isEmpty()) {
+      return original;
+    }
+
+    List<RelFieldCollation> reversedFields = new ArrayList<>();
+    for (RelFieldCollation field : original.getFieldCollations()) {
+      RelFieldCollation.Direction reversedDirection = field.direction.reverse();
+      
+      // Handle null direction properly - reverse it as well
+      RelFieldCollation.NullDirection reversedNullDirection = field.nullDirection;
```
RyanL1997: nit: can be simplified as 

```java
RelFieldCollation.NullDirection reversedNullDirection =
    field.nullDirection == RelFieldCollation.NullDirection.FIRST ? RelFieldCollation.NullDirection.LAST :
    field.nullDirection == RelFieldCollation.NullDirection.LAST ? RelFieldCollation.NullDirection.FIRST :
    field.nullDirection;
```

core/src/main/java/org/opensearch/sql/calcite/plan/OpenSearchRules.java
selsong: Thanks! renamed for clarity

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
selsong: Fixed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
selsong: Thank you all for the excellent feedback about using Calcite optimization rules! After implementing and testing the SortDirectionOptRule, I discovered that you were right - Calcite's built-in optimization already handles sort merging effectively. The physical plans were identical with or without the custom rule.

Based on this, I've removed the custom optimization rule and now rely entirely on Calcite's mature optimization framework. The core improvement remains in PlanUtils: reverse detection that flips sort directions instead of using window functions.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
selsong: Thanks! I switched from getTraitSet().getCollation() to getMetadataQuery().collations() to properly handle RelCompositeTrait instances and collation equivalences.
Null direction preservation: The optimization rule now properly handles and reverses null directions when flipping sort collations.

selsong:
> I'm curious about the cases where the double reverse optimization matters in practice, seems unlikely that someone would reverse something twice like that except by accident

Thanks for the note! I agree. It’s extremely unlikely that users would intentionally issue a double reverse and there is no usage as of now. This PR is now only focused on flipping the sort direction, and Calcite's default optimization takes care of the double reverse.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
selsong: Fixed. Thanks!

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
selsong: Fixed thanks!

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
selsong: Added UT. Thanks!

core/src/main/java/org/opensearch/sql/calcite/plan/OpenSearchRules.java
```diff
@@ -16,6 +17,11 @@ public class OpenSearchRules {
   public static final List<RelOptRule> OPEN_SEARCH_OPT_RULES =
       ImmutableList.of(AGGREGATE_CONVERT_RULE);
 
+  public static final List<RelOptRule> OPTIMIZATION_RULES = ImmutableList.of();
```
dai-chen: This is not necessary now?

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
```diff
@@ -0,0 +1,117 @@
...
+
+import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.rel.logical.LogicalSort;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.sql.calcite.plan.LogicalSystemLimit;
+
+/** Combines sort then reverse into 1 sort. */
+public class SortReverseOptimizationRule extends RelOptRule {
```
dai-chen: rename this more generic because `reverse` is PPL concept?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -217,6 +216,30 @@ public void testExplainWithReverse() throws IOException {
...
 
+  @Test
+  public void testExplainWithReversePushdown() throws IOException {
+    String query = "source=opensearch-sql_test_index_account | sort - age | reverse";
+    var result = explainQueryToString(query);
+    String expected =
+        isPushdownEnabled()
+            ? loadFromFile("expectedOutput/calcite/explain_reverse_pushdown_single.json")
+            : loadFromFile(
+                "expectedOutput/calcite_no_pushdown/explain_reverse_pushdown_single.json");
```
dai-chen: `loadExpectedPlan` handles this already?

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public boolean matches(RelOptRuleCall call) {
+    LogicalSort outerSort = call.rel(0);
+    LogicalSort innerSort = call.rel(1);
+
+    LOG.debug("SortReverseOptimizationRule.matches() called");
+    LOG.debug("Outer sort: {}", outerSort);
+    LOG.debug("Inner sort: {}", innerSort);
+    LOG.debug("Inner sort input: {}", innerSort.getInput());
```
dai-chen: merge into single logging like line 95?

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
```diff
@@ -0,0 +1,117 @@
...
+
+    // Don't optimize if outer sort is a LogicalSystemLimit
+    if (call.rel(0) instanceof LogicalSystemLimit) {
+      LOG.debug("Skipping: outer sort is LogicalSystemLimit");
+      return false;
+    }
+
+    // Don't optimize if inner sort has a fetch limit (head/limit before sort)
+    // This preserves limit-then-sort semantics
```
dai-chen: Could you try to add UT which may be more readable than comment? If not possible, probably you can add high level comment with example query plan tree above.

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
```diff
@@ -0,0 +1,117 @@
...
+package org.opensearch.sql.calcite.rule;
+
+import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.rel.logical.LogicalSort;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.sql.calcite.plan.LogicalSystemLimit;
+
+/** Combines sort then reverse into 1 sort. */
```
yuancu: I'm a little about curious what's the behavior without this rule.  Won't multiple sorts on the same fields be merged?

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
```diff
@@ -0,0 +1,117 @@
...
+  public static final SortReverseOptimizationRule INSTANCE = new SortReverseOptimizationRule();
+
+  private SortReverseOptimizationRule() {
+    super(
+        operand(LogicalSort.class, operand(LogicalSort.class, any())),
+        "SortReverseOptimizationRule");
+  }
+
+  @Override
+  public boolean matches(RelOptRuleCall call) {
```
yuancu: Are there specific reasons on overriding the method instead of using a config?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReverseTest.java
```diff
@@ -19,13 +19,7 @@ public void testReverseParserSuccess() {
...
     RelNode root = getRelNode(ppl);
     String expectedLogical =
-        ""
-            + "LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5],"
-            + " COMM=[$6], DEPTNO=[$7])\n"
-            + "  LogicalSort(sort0=[$8], dir0=[DESC])\n"
-            + "    LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4],"
-            + " SAL=[$5], COMM=[$6], DEPTNO=[$7], __reverse_row_num__=[ROW_NUMBER() OVER ()])\n"
-            + "      LogicalTableScan(table=[[scott, EMP]])\n";
+        "LogicalSort(sort0=[$0], dir0=[DESC])\n" + "  LogicalTableScan(table=[[scott, EMP]])\n";
```
yuancu, line 22: In my understanding, it falls back to the original implementation with window functions when there is no preceding sort. In this test, if there is not reverse, are the fields sort by `$0` (EMPNO) by default?

If not, this test case should fallback.

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
selsong: Yes, added UT and made comment more readable.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
selsong: Fixed

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
selsong: Removed opt rule

core/src/main/java/org/opensearch/sql/calcite/plan/OpenSearchRules.java
selsong: Agreed, removed

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReverseTest.java
selsong: Yes, EMP table has implicit ordering by EMPNO ($0) in test data. This is detected as an existing sort and converts `reverse` to a simple DESC sort instead of using window functions.

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
selsong: I think in this case override provides more control for complex matching logic like checking fetch limits and intermediate nodes.

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
selsong: You're right. Multiple sorts on the same fields are merged so without this rule physical plans are identical due to Calcite's built-in physical optimization. I think this rule ensures consistent optimization across different query patterns, especially in post-aggregation scenarios where physical rules might not apply.

songkant-aws: Could you provide an example where default physical rule might not apply? 

There is a `SortRemoveRule` in Calcite that changes the traits based on the Sort collation requirement. For this two sorts case, I think it will pick one of them anyway. Seems this optimization might be overlapped with existing rule.

In physical plan rules, Calcite will ensure query do not lose sort semantics by adding `AbstractConverter` node over applicable nodes that require sorting. The `AbstractConverter` node checks if the child collation satisfies the expected collation to decide whether to add a sort between parent and child. The sort added by `AbstractConverter` should be winner of the two sorts.

core/src/main/java/org/opensearch/sql/calcite/rule/SortDirectionOptRule.java
```diff
@@ -0,0 +1,128 @@
...
+    boolean matches = hasSameFieldWithOppositeDirection(outerSort, innerSort);
+    LOG.debug("Same field with opposite direction: {}", matches);
+    return matches;
+  }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    LogicalSort outerSort = call.rel(0);
+    org.apache.calcite.rel.RelNode intermediate = call.rel(1);
+    LogicalSort innerSort = call.rel(2);
```
songkant-aws: I think this pattern needs careful consideration. There are lots of cases to be considered. 

If intermediate is a Project node without WINDOW function, the pattern is supposed to be optimized by `SortProjectTransposeRule`. And it will be changed to Project - outerSort - innerSort. Other bunch of rules may optimize the changed pattern to keep outerSort.

If the Project contains WINDOW function, we should not change outerSort under the intermediate node. Because it may change its semantics.

If the intermediate node is aggregation, suppose the sort is on the group by column, I think the outerSort over aggregated result will be more efficient because the row count will be reduced a lot after aggregation. Check if Calcite can make it like outerSort(groupByColumn) - Aggregate(count) by groupByColumn

selsong: Thanks, you're right about these cases. Given that Calcite's built-in optimization already handles these cases correctly and efficiently, I've removed the custom optimization rule entirely. This eliminates the risk of interfering with Calcite's sophisticated rule interactions while preserving the core performance improvement.

core/src/main/java/org/opensearch/sql/calcite/rule/SortReverseOptimizationRule.java
selsong: Thank you both for the insightful discussion! You were absolutely correct, and I was initially wrong in my assessment. After further testing and analysis, I've confirmed that:
Calcite's built-in optimization is sufficient - SortRemoveRule and AbstractConverter nodes handle sort merging effectively
Physical plans are identical with or without the custom rule, so the custom rule was redundant 
Based on your feedback, I've removed the SortDirectionOptRule entirely. The core improvement remains: reverse detection that flips existing sort directions. Calcite's built-in optimization then handles the sort merging at the physical level.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReverseTest.java
yuancu: Why is EMP table **implicitly** ordered by EMPNO?  How is the order inferred / detected?

Update: I tried debugging, and found that the table does contain a collation trait with EMPNO sorted even with `source=EMP`. I don't know where is this collation from though.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -238,8 +238,7 @@ public void testFilterFunctionScriptPushDownExplain() throws Exception {
   @Test
   public void testExplainWithReverse() throws IOException {
     String result =
-        executeWithReplace(
-            "explain source=opensearch-sql_test_index_account | sort age | reverse | head 5");
+        executeWithReplace("explain source=opensearch-sql_test_index_account | reverse | head 5");
```
yuancu: Can you please verify the full logical plan of this query. As this is the only test without a preceding collation, it helps to ensure the logical is correct over time with manually executing the query.

selsong: Added full logical plan. Thanks!

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReverseTest.java
dai-chen: interesting, that's why there is no test case for using `ROW_NUMBER` in this test right?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -586,19 +587,30 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
-            .rowsTo(RexWindowBounds.CURRENT_ROW)
-            .as(REVERSE_ROW_NUM);
-    context.relBuilder.projectPlus(rowNumber);
-    // Sort by row number descending
-    context.relBuilder.sort(context.relBuilder.desc(context.relBuilder.field(REVERSE_ROW_NUM)));
-    // Remove row number column
-    context.relBuilder.projectExcept(context.relBuilder.field(REVERSE_ROW_NUM));
+
+    // Check if there's an existing sort to reverse
+    List<RelCollation> collations =
```
dai-chen, line 592: So we assume the head of this collation list must come from the `sort` command right before `reverse` command right? Is it always true or the reverse logic below works even though it's not true?

selsong: 1. Yes, this test uses the EMP table which has a natural ordering by EMPNO. When reverse is called, Calcite's metadata query detects this existing collation and reverses it directly, so we see LogicalSort(sort0=[$0], dir0=[DESC]) instead of ROW_NUMBER. The ROW_NUMBER fallback only occurs when there's no existing collation.
2. No, the assumption isn't always true, but I think the logic works correctly when its not true. getMetadataQuery().collations() returns all collations Calcite detects (natural orderings, index orderings, sort operations), not just the immediate sort command. We take the first available RelCollation (which may contain multiple sort fields) and reverse all its field collations at once.

dai-chen: For Q2: Take query `... | sort | X | Y | Z | reverse` for example, we should be good as long as commands in-between (X, Y, Z) doesn't rely on the output order of `sort` command, right? Just thinking any counterexample, such as `... | sort | head | reverse`?

I'm thinking is it too early to do this in `visitReverse`? Essentially, our `CalciteRelNodeVisitor` converts AST to its **logical representation** by RelNode. Just wondering what does it look like if we simply translate reverse to logical sort operator here?

selsong: Great question about edge cases! I tested both `| sort | head 10 | reverse` and `| sort | eval | head | reverse` patterns and they work correctly. In the first part of visitReverse the SORT ASC and LIMIT for head are combined into `LogicalSort(sort0=[$0], dir0=[ASC-nulls-first], fetch=[10])` and then the reverseCollation is added with `LogicalSort(sort0=[$0], dir0=[DESC-nulls-last])`. The physical plan pushes SORT ASC + LIMIT N to OpenSearch, then applies DESC sort on the limited results. Let me know if there are any other edge cases, Thanks!

integ-test/src/test/resources/expectedOutput/calcite/explain_double_reverse_no_op.json
```diff
@@ -0,0 +1,6 @@
...
+    "physical": "EnumerableCalc(expr#0..3=[{inputs}], proj#0..1=[{exprs}])\n  EnumerableLimit(fetch=[10000])\n    EnumerableSort(sort0=[$3], dir0=[DESC])\n      EnumerableWindow(window#0=[window(rows between UNBOUNDED PRECEDING and CURRENT ROW aggs [ROW_NUMBER()])])\n        EnumerableSort(sort0=[$2], dir0=[DESC])\n          EnumerableWindow(window#0=[window(rows between UNBOUNDED PRECEDING and CURRENT ROW aggs [ROW_NUMBER()])])\n            CalciteEnumerableIndexScan(table=[[OpenSearch, events]], PushDownContext=[[PROJECT->[cpu_usage, @timestamp]], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"_source\":{\"includes\":[\"cpu_usage\",\"@timestamp\"],\"excludes\":[]}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
+  }
+}
```
yuancu, line 6: Is this explain result and ``explain_double_reverse_sort_no_op`` used somewhere? I guess they were used for the elimination rule of double reverse. Then you delete the rule and relevant tests

ahkcs:
@selsong Can you follow up on this? If not, we can take it over

ahkcs:
Moving the PR to #4775

Swiddis:
Supplanted by #4775

PR #4071, opened 2025-08-19
Support pushdown sort by simple expressions

### LINKED ISSUE #3990: [FEATURE] Push down sort on RexCall of monotonic
This query
```
source = test |   eval b = a + 1 | sort b
```
is equivalent to:
```
source = test | sort a | eval b = a + 1
```

since `a + 1` is monotonic and so it's collation equals to `a`.

We can transform the query into the latter one and then has chance to push down its sort into scan.

songkant-aws:
### Description
This PR aims to resolve simple sort expression pushdown problem without prerequisite of project pushdown optimization.

A PPL query may contain a sort over projected trivial expression like: `source = test |  eval b = a + 1 | sort b`

This optimization PR will optimize this PPL into `source = test | sort a | eval b = a + 1` so that the problem is translated to pushdown field sort on column a. Also, sorting by field is supposed to be faster than sorting by script considering OpenSearch internal optimization.

### Related Issues
Resolves #3990 

### Check List

### DIFF
--- a/opensearch/src/test/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtilTest.java
+++ b/opensearch/src/test/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtilTest.java
@@ -199,8 +199,7 @@ public class OpenSearchRelOptUtilTest {
     srcInput = rexBuilder.makeInputRef(srcType, 1);
     dstType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP);
     cast = rexBuilder.makeCast(dstType, srcInput);
-    result =
-        OpenSearchRelOptUtil.getOrderEquivalentInputInfo(cast);
+    result = OpenSearchRelOptUtil.getOrderEquivalentInputInfo(cast);
     assertExpectedInputInfo(result, 1, false);
 
     // Cast from integer to double
@@ -208,8 +207,7 @@ public class OpenSearchRelOptUtilTest {
     srcInput = rexBuilder.makeInputRef(srcType, 1);
     dstType = typeFactory.createSqlType(SqlTypeName.DOUBLE);
     cast = rexBuilder.makeCast(dstType, srcInput);
-    result =
-        OpenSearchRelOptUtil.getOrderEquivalentInputInfo(cast);
+    result = OpenSearchRelOptUtil.getOrderEquivalentInputInfo(cast);
     assertExpectedInputInfo(result, 1, false);
 
     // Cast from integer to float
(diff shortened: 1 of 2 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/SortProjectExprTransposeRule.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    final Sort sort = call.rel(0);
+    final Project project = call.rel(1);
+
+    if (sort.getConvention() != project.getConvention()) {
+      return;
+    }
```
yuancu: I'm curious when will their conventions be different. From the entrance rule, it seems they will always be `LogicalSort` and `LogicalProject` with `NONE` convention.

songkant-aws: For this case, it should be the same. The original rule targets to optimize both logical and physical plans. So I used this check but forgot to remove it. I can remove it once I verify it's not needed.

songkant-aws:
> It seems that when the sorted field is not projected in the final output, there will be an added sort at the top.
> 
> E.g.
> 
> * `source=opensearch-sql_test_index_account | eval b = balance + 1 | sort b | fields b`:
>   ```
>   EnumerableCalc(expr#0=[{inputs}], expr#1=[1], expr#2=[+($t0, $t1)], balance=[$t0], $f1=[$t2])
>     CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[SORT->[{
>     "balance" : {
>       "order" : "asc",
>       "missing" : "_first"
>     }
>   }], LIMIT->10000, PROJECT->[balance]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":10000,"timeout":"1m","_source":{"includes":["balance"],"excludes":[]},"sort":[{"balance":{"order":"asc","missing":"_first"}}]}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
>   ```
> * but for `source=opensearch-sql_test_index_account | eval b = balance + 1 | sort b | fields b`
>   ```
>   EnumerableSort(sort0=[$0], dir0=[ASC-nulls-first])
>     EnumerableCalc(expr#0=[{inputs}], expr#1=[1], expr#2=[+($t0, $t1)], $f0=[$t2])
>       CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[balance], SORT->[{
>     "balance" : {
>       "order" : "asc",
>       "missing" : "_first"
>     }
>   }], LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":10000,"timeout":"1m","_source":{"includes":["balance"],"excludes":[]},"sort":[{"balance":{"order":"asc","missing":"_first"}}]}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
>   ```

Nice catch! It happens when project doesn't carry the equivalent input field in its project node expressions. In `ExpandCollationOnProjectExprRule`, I use Project itself collation to check if it satisfies the target collation(aka top sort collation). But since Project has only trivial simple expression output, it will contain empty RelCollation []. So in this case, the rule will skip due to the empty fromCollation.

I need to check if it's safe to use Project.getInput() collation instead of collations on Project itself. The problem should be solved with that fix.

qianheng-aws (approved):

opensearch/src/test/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtilTest.java
```diff
@@ -0,0 +1,246 @@
...
+
+    // Cast from low precision to high precision
+    srcType = typeFactory.createSqlType(SqlTypeName.DECIMAL);
+    srcInput = rexBuilder.makeInputRef(srcType, 1);
+    dstType =
+        typeFactory.createSqlType(
+            SqlTypeName.DECIMAL, srcType.getPrecision() + 4, srcType.getScale() + 4);
+    cast = rexBuilder.makeCast(dstType, srcInput);
+    result = OpenSearchRelOptUtil.getOrderEquivalentInputInfo(cast);
+    assertExpectedInputInfo(result, 1, false);
```
yuancu, line 229: A converse test like the following may help:

```java
// Cast from high precision to low precision
srcType = typeFactory.createSqlType(SqlTypeName.DECIMAL);
srcInput = rexBuilder.makeInputRef(srcType, 1);
dstType =
        typeFactory.createSqlType(
                SqlTypeName.DECIMAL, srcType.getPrecision() - 4, srcType.getScale());
cast = rexBuilder.makeCast(dstType, srcInput);
result = OpenSearchRelOptUtil.getOrderEquivalentInputInfo(cast);
assertFalse(result.isPresent());
```

yuancu (approved):
LGTM

Issue #4082, opened 2025-08-20
[RFC] Add `regex` Command to PPL (Calcite Engine)


RyanL1997:
## Problem Statement
In data analysis and log processing workflows, users frequently need to filter records based on pattern matching within field values, for example:

- Filtering log entries that contain specific error patterns or codes
- Extracting records where email addresses match certain domain patterns  
- Finding entries with usernames following specific naming conventions
- Identifying records with IP addresses, phone numbers, or other structured data patterns

Currently, PPL does not support regex-based filtering, forcing users to rely on exact string matching with `where` clauses or external processing tools. This limits query flexibility for complex pattern-based data extraction scenarios.

## Current State
- PPL supports `where` for exact field value matching, but lacks pattern-based filtering capabilities
- PPL provides `match` command for full-text search, which uses OpenSearch's analyzers and scoring but doesn't support regex patterns
- No native support for regular expression matching in the Calcite engine
- Users must resort to complex string functions or external tools for pattern matching

## Long-Term Goals
- Provide intuitive regex pattern matching capabilities within PPL pipelines
- Enable efficient server-side regex filtering to reduce data transfer overhead
- Establish consistent patterns for text processing commands in PPL
- Support both positive and negative regex matching scenarios

### Syntax
```bash
<source> | [commands] | regex <field>=<pattern> | [commands]
<source> | [commands] | regex <field>!=<pattern> | [commands]  
```

- `field`: The field name to apply regex matching against
- `pattern`: Java regex pattern string (supports standard regex metacharacters)
- `!=`: Negated matching - returns records that do NOT match the pattern
- `=`: Positive matching - returns records that match the pattern

### Semantics
**Expected Behavior:**

- **Action**: Filters records based on regex pattern matching against specified field values
- **Matching Type**: Partial matching - pattern can match anywhere within the field value
- **Pattern Support**: Java regex syntax including:
  - Character classes: `[a-z]`, `[^0-9]`, `\d`, `\w`, `\s`
  - Quantifiers: `*`, `+`, `?`, `{n}`, `{n,m}`
  - Anchors: `^` (start), `$` (end)
  - Groups: `(abc)`, `(?:abc)`
  - Alternation: `a|b`
  - Escape sequences: `\.`, `\\`, `\n`, `\t`
- **Case Sensitivity**: Matching is case-sensitive by default
- **Null Handling**: Returns null for null field values or patterns
- **Data Preservation**: Does not modify field values, only filters records

### Examples
```bash
[...]

PR #4083, opened 2025-08-20
[Feature] Implementation of `regex` Command In PPL

RyanL1997:
### Description
 Implementation of `regex` Command In PPL

### Details:
- Core logic of `regex`
```bash
<source> | [commands] | regex <field>=<pattern> | [commands]
<source> | [commands] | regex <field>!=<pattern> | [commands]  
```

- field: The field name to apply regex matching against
- pattern: Java regex pattern string (supports standard regex metacharacters)
- `!=`: Negated matching - returns records that do NOT match the pattern
- `=`: Positive matching - returns records that match the pattern

For the handling of default field (`_source` / `_raw`): 
- Tracking at issue https://github.com/opensearch-project/sql/issues/4111

### Related Issues
* Relate https://github.com/opensearch-project/sql/issues/4082

### Check List


core/src/main/java/org/opensearch/sql/expression/parse/RegexCommonUtils.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /**
+   * Convert ExprValue to string, handling non-string types. Based on decision to auto-convert
+   * non-string fields to strings.
+   *
+   * @param value The ExprValue to convert
+   * @return String representation of the value
+   */
+  public static String toStringValue(ExprValue value) {
```
ahkcs: This method seems unused

core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
```diff
@@ -380,6 +381,7 @@ public class PPLBuiltinOperators extends ReflectiveSqlOperatorTable {
       RELEVANCE_QUERY_FUNCTION_INSTANCE.toUDF("multi_match", false);
   public static final SqlOperator NUMBER_TO_STRING =
       new NumberToStringFunction().toUDF("NUMBER_TO_STRING");
+  public static final SqlOperator REGEX_MATCH = SqlLibraryOperators.REGEXP_CONTAINS;
```
ahkcs: If we are going to use this operator, I think we don't need another alias

ahkcs:
Could you add `CrossClusterSearchIT` and `PPLQueryDataAnonymizerTest` as well?

core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
RyanL1997: good catch, this was some left over from my original poc. Fixed.

core/src/main/java/org/opensearch/sql/expression/parse/RegexCommonUtils.java
RyanL1997: fixed

RyanL1997:
> Could you add CrossClusterSearchIT and PPLQueryDataAnonymizerTest as well?

~~Will add unit test and these test soon since i would like to have the general core logic to be reviewed first.~~

Added.

core/src/main/java/org/opensearch/sql/ast/tree/Regex.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /** Pattern. */
+  private final Literal pattern;
+
+  /** Child Plan. */
+  @Setter private UnresolvedPlan child;
+
+  public Regex(UnresolvedExpression field, String operator, Literal pattern) {
+    this.field = field;
+    this.negated = "!=".equals(operator);
```
ykmr1224: Should we define constants for `!=` and `=`? I see multiple use of them.

core/src/main/java/org/opensearch/sql/expression/operator/predicate/RegexMatch.java
```diff
@@ -0,0 +1,79 @@
...
+import org.opensearch.sql.data.type.ExprType;
+import org.opensearch.sql.expression.Expression;
+import org.opensearch.sql.expression.ExpressionNodeVisitor;
+import org.opensearch.sql.expression.env.Environment;
+import org.opensearch.sql.expression.parse.RegexCommonUtils;
+
+/**
+ * Expression for regex matching using Java's built-in regex engine. Supports standard Java regex
+ * features including named groups, lookahead/lookbehind, backreferences, and inline flags. Uses
+ * find() for partial matching to align with SPL semantics.
```
ykmr1224: Should we eliminate other product name?

core/src/main/java/org/opensearch/sql/expression/operator/predicate/RegexMatch.java
```diff
@@ -0,0 +1,79 @@
...
+  @Override
+  public ExprValue valueOf(Environment<Expression, ExprValue> valueEnv) {
+    ExprValue fieldValue = field.valueOf(valueEnv);
+    ExprValue patternValue = pattern.valueOf(valueEnv);
+
+    if (fieldValue.isNull()
+        || fieldValue.isMissing()
+        || patternValue.isNull()
+        || patternValue.isMissing()) {
+      return ExprValueUtils.booleanValue(false);
```
ykmr1224: Could this actually happen? In that case, is this behavior align with requirements?

core/src/main/java/org/opensearch/sql/expression/operator/predicate/RegexMatch.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    String text = fieldValue.stringValue();
+    String regex = patternValue.stringValue();
+
+    if (text == null) {
+      return ExprValueUtils.booleanValue(false);
```
ykmr1224: Should `negated` affect this result?

core/src/main/java/org/opensearch/sql/expression/parse/RegexCommonUtils.java
```diff
@@ -0,0 +1,104 @@
...
+   * Get compiled pattern from cache or compile and cache it.
+   *
+   * @param regex The regex pattern string
+   * @return Compiled Pattern object
+   * @throws PatternSyntaxException if the regex is invalid
+   */
+  public static Pattern getCompiledPattern(String regex) {
+    // Check cache size and clear if needed (simple LRU-like behavior)
+    if (patternCache.size() > MAX_CACHE_SIZE) {
+      patternCache.clear();
```
ykmr1224: I think we should use LinkedHashMap to remove oldest one. (This implementation remove everything once reached max size.)

We might be able to use LRUMap from Apache commons.

core/src/main/java/org/opensearch/sql/expression/parse/RegexExpression.java
```diff
@@ -36,32 +33,30 @@ public class RegexExpression extends ParseExpression {
...
   }
 
   @Override
   ExprValue parseValue(ExprValue value) throws ExpressionEvaluationException {
     String rawString = value.stringValue();
-    Matcher matcher = regexPattern.matcher(rawString);
-    if (matcher.matches()) {
-      return new ExprStringValue(matcher.group(identifierStr));
+
+    // Extract the specific named group
```
ykmr1224: nit: Let's don't put redundant comment.

core/src/main/java/org/opensearch/sql/expression/parse/RegexExpression.java
```diff
@@ -36,32 +33,30 @@ public class RegexExpression extends ParseExpression {
...
+    if (extracted != null) {
+      return new ExprStringValue(extracted);
     }
     log.debug("failed to extract pattern {} from input ***", regexPattern.pattern());
     return new ExprStringValue("");
   }
 
   /**
-   * Get list of derived fields based on parse pattern.
+   * Get list of derived fields based on parse pattern. Delegates to shared utility.
```
ykmr1224: Why do we need to delegate? Should we directly call the delegated method from the class which calls this method?

docs/user/ppl/cmd/regex.rst
```diff
@@ -0,0 +1,184 @@
+=============
+regex
```
ykmr1224, line 2: Let's note supported version.

docs/user/ppl/cmd/regex.rst
```diff
@@ -0,0 +1,184 @@
...
+============
+| The ``regex`` command filters search results by matching field values against a regular expression pattern. Only documents where the specified field matches the pattern are included in the results.
+
+Syntax
+============
+regex <field> = <pattern>
+regex <field> != <pattern>
+regex <pattern>
+
+* field: optional. The field name to match against. If not specified, the pattern will be matched against the default field.
```
ykmr1224: Can we clarify about `default field`?

docs/user/ppl/cmd/regex.rst
```diff
@@ -0,0 +1,184 @@
...
+
+Regular Expression Engine
+==========================
+
+The regex command uses Java's built-in regular expression engine, which supports:
+
+* **Standard regex features**: Character classes, quantifiers, anchors
+* **Named capture groups**: ``(?<name>pattern)`` syntax
+* **Lookahead/lookbehind**: ``(?=...)`` and ``(?<=...)`` assertions
+* **Case sensitivity**: Patterns are case-sensitive by default
```
ykmr1224: Does `by default` indicate it can be changed?

core/src/main/java/org/opensearch/sql/ast/tree/Regex.java
RyanL1997: fixed

core/src/main/java/org/opensearch/sql/expression/operator/predicate/RegexMatch.java
RyanL1997: good catch, fixed

docs/user/ppl/cmd/regex.rst
RyanL1997: This is a good question, since we are having this tracking issue (https://github.com/opensearch-project/sql/issues/4111) to track the default field's on going design, do you think we should mention the current limitation in the doc just for now? The current limitation is that "for rex and regex, user have to specified a field". cc @penghuo

core/src/main/java/org/opensearch/sql/expression/operator/predicate/RegexMatch.java
RyanL1997: Nice catch, I rechecked the logic:
- The field null/missing checks are necessary and correct - they filter out documents that don't have the field being searched, which is standard regex filtering behavior and throw an exception until #4111 is implemented.
```bash
curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d'{"query": "source=accounts | regex \"*.on$\" "}' | jq

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   219  100   172  100    47    257     70 --:--:-- --:--:-- --:--:--   327
{
  "error": {
    "reason": "Invalid Query",
    "details": "Regex command requires a field to be specified",
  },
  "status": 400
}
```

-  Removed patternValue.isNull() and patternValue.isMissing() checks since patterns are always literal strings in the current implementation
- Kept only the field null/missing checks which are necessary and correct

core/src/main/java/org/opensearch/sql/expression/operator/predicate/RegexMatch.java
RyanL1997: The field validation happens at query analysis time before our code executes, which catches non-existent fields early. Our null/missing checks handle the runtime cases where fields exist in the schema but have null/missing values in specific documents.

So, negation correctly doesn't affect this behavior - null/missing values are filtered out in both regular and negated regex, which aligns with standard filtering semantics.

```bash
# Case 1: =
❯   curl -X POST "localhost:9200/_plugins/_ppl" \
    -d'{"query": "source=accounts | regex middlename=\".*\" "}' | jq

{
  "error": {
    "reason": "Invalid Query",
    "details": "field [middlename] not found; input fields are: [account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname, _id, _index, _score, _maxscore, _sort, _routing]",
  },
  "status": 400
}

# Case 2: !=
❯   curl -X POST "localhost:9200/_plugins/_ppl" \
    -d'{"query": "source=accounts | regex middlename!=\".*\" "}' | jq

{
  "error": {
    "reason": "Invalid Query",
    "details": "field [middlename] not found; input fields are: [account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname, _id, _index, _score, _maxscore, _sort, _routing]",
  },
  "status": 400
}
```

core/src/main/java/org/opensearch/sql/expression/parse/RegexCommonUtils.java
RyanL1997: Thanks for pointing it out. Implemented proper LRU cache using `LinkedHashMap` with access-order tracking and automatic eviction of oldest entries. This replaces the previous 'clear everything' approach with efficient removal of only the least recently used patterns when cache limit is reached. The implementation is thread-safe using `Collections.synchronizedMap`. Could you take a look again?

core/src/main/java/org/opensearch/sql/expression/parse/RegexExpression.java
RyanL1997: fixed.

core/src/main/java/org/opensearch/sql/expression/parse/RegexExpression.java
RyanL1997: Good catch. Removed the unnecessary delegation layer. `ParseUtils` now calls `RegexCommonUtils.getNamedGroupCandidates()` directly instead of going through `RegexExpression.getNamedGroupCandidates()`. This eliminates the wrapper method that added no value and simplifies the call chain.

docs/user/ppl/cmd/regex.rst
RyanL1997: I've updated the documentation to be more accurate. The case sensitivity can be changed using Java's built-in inline flags like `(?i)` for case-insensitive matching, `(?m)` for multiline mode, `(?s)` for dotall mode, etc. This isn't something we implemented ourselves - it's standard Java regex functionality that we inherit through `Pattern.compile()`

 I've updated the documentation to:
  - Remove the confusing 'by default' language
  - List common inline flags with examples
  - Add a reference to the official Java Pattern documentation (https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) for complete coverage of all available modes and features

core/src/main/java/org/opensearch/sql/ast/tree/Regex.java
```diff
@@ -0,0 +1,61 @@
...
+import org.opensearch.sql.ast.AbstractNodeVisitor;
+import org.opensearch.sql.ast.expression.Literal;
+import org.opensearch.sql.ast.expression.UnresolvedExpression;
+
+/** AST node represent Regex filtering operation. */
+@Getter
+@ToString
+@EqualsAndHashCode(callSuper = false)
+public class Regex extends UnresolvedPlan {
+  /** Operator constants. */
```
ykmr1224: nit: Let's remove redundant comments (I see many). Those would be visual noise for reader, and could waste tokens for LLMs. I'd recommend to add common instruction for LLM to avoid unneeded comments.

core/src/main/java/org/opensearch/sql/expression/parse/RegexCommonUtils.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+/**
+ * Common utilities for regex operations. Provides pattern caching and consistent matching behavior.
+ */
+public class RegexCommonUtils {
```
ykmr1224, line 20: Let's add unit tests.

docs/user/ppl/cmd/regex.rst
ykmr1224: If it is mandatory now, let's describe it as mandatory. We should update it once we made it optional.

docs/user/ppl/cmd/regex.rst
```diff
@@ -0,0 +1,143 @@
...
+    | 1              | VA    |
+    +----------------+-------+
+
+
+Limitations
+===========
+
+* **String fields only**: The regex command currently only supports string fields. Using it on numeric or boolean fields will result in an error
+* **Performance**: Complex regex patterns may impact query performance, especially on large datasets
+* **Memory usage**: Pattern compilation results are cached, but very large numbers of unique patterns may consume memory
```
ykmr1224: Do we need to note this here? I don't see much risk to just cache 1000 regex patterns.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteRegexCommandIT.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.remote;
+
+import org.opensearch.sql.ppl.RegexCommandIT;
+
+public class CalciteRegexCommandIT extends RegexCommandIT {
```
ykmr1224: Is it meant to execute test defined in RegexCommandIT? It looks confusing to me.

integ-test/src/test/java/org/opensearch/sql/ppl/RegexCommandIT.java
```diff
@@ -0,0 +1,91 @@
...
+import java.io.IOException;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+
+public class RegexCommandIT extends PPLIntegTestCase {
+
+  @Override
+  public void init() throws Exception {
+    super.init();
+    enableCalcite();
```
ykmr1224: Is it intentionally enabled here as well?

core/src/main/java/org/opensearch/sql/ast/tree/Regex.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /** Pattern. */
+  private final Literal pattern;
+
+  /** Child Plan. */
+  @Setter private UnresolvedPlan child;
+
+  public Regex(UnresolvedExpression field, String operator, Literal pattern) {
+    this.field = field;
+    this.negated = NOT_EQUALS_OPERATOR.equals(operator);
```
ykmr1224: Should we receive `nagated` as a parameter? (Why do we need to convert from String here?)

docs/user/ppl/cmd/regex.rst
RyanL1997: good point. removed

core/src/main/java/org/opensearch/sql/ast/tree/Regex.java
RyanL1997: Good catch, fixed.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteRegexCommandIT.java
RyanL1997: Good catch and this is an left over design since I was trying to support this command in legacy engine. Bur now I think this is should be supported in calcite only. So i have relocated this under calcite IT and removed this file entirely.

integ-test/src/test/java/org/opensearch/sql/ppl/RegexCommandIT.java
RyanL1997: addressed by relocate this entirely under calcite IT - this was an left over design that i was trying to support this in legacy engine. But now, it should be only supported when calcite enabled.

core/src/main/java/org/opensearch/sql/expression/ExpressionNodeVisitor.java
```diff
@@ -101,4 +102,13 @@ public T visitWhen(WhenClause node, C context) {
...
     return visitNode(node, context);
   }
+
+  public T visitRegex(RegexMatch node, C context) {
+    T result = defaultResult();
+    T fieldResult = node.getField().accept(this, context);
+    result = aggregateResult(result, fieldResult);
+    T patternResult = node.getPattern().accept(this, context);
+    result = aggregateResult(result, patternResult);
+    return result;
```
dai-chen: Any reason not to visitChildren like others?

docs/user/ppl/cmd/regex.rst
RyanL1997: updated

core/src/main/java/org/opensearch/sql/expression/parse/RegexCommonUtils.java
RyanL1997: added

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -174,6 +175,31 @@ public RelNode visitFilter(Filter node, CalcitePlanContext context) {
...
 
+  @Override
+  public RelNode visitRegex(Regex node, CalcitePlanContext context) {
+    visitChildren(node, context);
+
+    // TODO: Handle the default field case, tracking by
+    // https://github.com/opensearch-project/sql/issues/4111
+    if (node.getField() == null) {
+      throw new IllegalArgumentException("Regex command requires a field to be specified");
+    }
```
dai-chen: Is this possible? Field is not optional in grammar right?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -174,6 +175,31 @@ public RelNode visitFilter(Filter node, CalcitePlanContext context) {
...
+
+    RexNode fieldRex = rexVisitor.analyze(node.getField(), context);
+    RexNode patternRex = rexVisitor.analyze(node.getPattern(), context);
+
+    RexNode regexCondition =
+        context.rexBuilder.makeCall(
+            org.apache.calcite.sql.fun.SqlLibraryOperators.REGEXP_CONTAINS, fieldRex, patternRex);
```
dai-chen: Is this default behavior of visitFunction?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -241,6 +243,14 @@ pathElement
 pathArrayAccess
    : LT_CURLY (INTEGER_LITERAL)? RT_CURLY
    ;
+regexCommand
+    : REGEX regexExpr
+    ;
+
+regexExpr
+    : field=qualifiedName EQUAL pattern=stringLiteral
+    | field=qualifiedName NOT_EQUAL pattern=stringLiteral
```
dai-chen: np: can be merged into one (EQUAL | NOT_EQUAL)

core/src/main/java/org/opensearch/sql/expression/operator/predicate/RegexMatch.java
```diff
@@ -0,0 +1,76 @@
...
+import org.opensearch.sql.expression.parse.RegexCommonUtils;
+
+/**
+ * Expression for regex matching using Java's built-in regex engine. Supports standard Java regex
+ * features including named groups, lookahead/lookbehind, backreferences, and inline flags. Uses
+ * find() for partial matching semantics.
+ */
+@ToString
+@EqualsAndHashCode
+public class RegexMatch implements Expression {
```
dai-chen: Is this for V2 engine?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: This is just for now. I do have an issue to track the default field implementation for both `regex` and `rex` : https://github.com/opensearch-project/sql/issues/4111

core/src/main/java/org/opensearch/sql/expression/operator/predicate/RegexMatch.java
RyanL1997: Good catch, and this was an left over from my poc branch which I was trying to support it for legacy engine. Since now, we are targeting in calcite only, so I have removed the above.

core/src/main/java/org/opensearch/sql/expression/ExpressionNodeVisitor.java
RyanL1997: > Good catch, and this was an left over from my poc branch which I was trying to support it for legacy engine. Since now, we are targeting in calcite only, so I have removed the above.

This can also be addressed by the above. - I have removed it.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: Oh, i saw the issue, this is not being triggered at all.

RyanL1997: removed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: If you were asking the default behavior between `=` and `!=` then, yes, the default behavior is the positive match (`field = 'pattern'`) using `REGEXP_CONTAINS`. For the negated case (`field != 'pattern'`), we wrap the result with `NOT`. This matches the grammar where `regexExpr` can be either `field=pattern` or `field!=pattern`, with the equals case being the base case and not-equals being the negated version.

ppl/src/main/antlr/OpenSearchPPLParser.g4
RyanL1997: Fixed.

core/src/main/java/org/opensearch/sql/ast/tree/Regex.java
RyanL1997: fixed.

core/src/test/java/org/opensearch/sql/analysis/AnalyzerTest.java
```diff
@@ -1929,4 +1929,18 @@ public void brain_patterns_command() {
 
     assertAnalyzeEqual(expectedPlan, patterns);
   }
+
+  @Test
+  public void regex_command_throws_unsupported_operation_exception_in_legacy_engine() {
```
ykmr1224: nit: seems this test file has different naming pattern.

docs/user/ppl/cmd/regex.rst
```diff
@@ -0,0 +1,141 @@
+=============
```
ykmr1224, line 1: You need to add this file name to `docs/category.json` to test it with doctest.

RyanL1997: fixed

core/src/test/java/org/opensearch/sql/analysis/AnalyzerTest.java
RyanL1997: kinda hard to fully align with them, but I have renamed it to `regex_command_throws_unsupported_exception_with_legacy_engine`.

dai-chen (dismissed):
Thanks for the changes!

docs/user/ppl/cmd/regex.rst
```diff
@@ -0,0 +1,150 @@
...
+
+
+Example 5: Case-sensitive matching
+==================================
+
+The example demonstrates that regex matching is case-sensitive by default.
+
+PPL query (case-sensitive - no match)::
+
+    source=accounts | regex state="virginia" | fields account_number, state;
```
ykmr1224: missing `os>` prefix. And you need to put expected output from next line.
Same for the below one.

RyanL1997: oh got it.

RyanL1997: fixed

Swiddis (approved):

RyanL1997:
Will manual backport it.

ykmr1224:
Can you do manual backport or see what is conflicting?
It is blocking https://github.com/opensearch-project/sql/pull/4214

PR #3654, opened 2025-05-22
Add DISTINCT_COUNT_APPROX function

xinyual:
### Description
This PR add DISTINCT_COUNT_APPROX function. It uses HyperLogLog++ algorithm.

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]
#3353 

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4084, opened 2025-08-20
Support distinct_count/dc in eventstats

### LINKED ISSUE #4052: [FEATURE] PPL `distinct_count`/`dc` function support for `eventstats` command

Yes. Currently `distinct_count`/`dc` work in `stats command, but it’s unavailable in `eventstats` command. This prevents annotating each event with the distinct count per group (or globally), which is a common log-analytics need by `eventstats` command.

1. Enable both in `eventstats` grammar and translate it to `COUNT(DISTINCT field) OVER (...)`.
2. Support general SQL window function pushdown optimization (tracked separately).

N/A

Test result:

```
opensearchsql> source=accounts | stats distinct_count(gender), dc(age);
+--------------------------+-----------+
| distinct_count(gender)   | dc(age)   |
|--------------------------+-----------|
| 2                        | 5         |
+--------------------------+-----------+

opensearchsql> source=accounts | eventstats distinct_count(age);
{'reason': 'Invalid Query', 'details': "[distinct_count] is not a valid term at this part of the query: '...counts | eventstats distinct_count' <-- HERE. Expecting one of 22 possible tokens. Some examples: 'AVG', 'COUNT', 'MAX', 'MIN', 'SUM', ...", 'type': 'SyntaxCheckException'}
```

ahkcs:
### Description
This PR implements support for `distinct_count()` and `dc()` functions in the PPL `eventstats` command, enabling users to calculate distinct counts as window functions with optional partitioning.

  ## Usage Examples

  ```sql
  -- Basic distinct count
  source=accounts | eventstats dc(state) as distinct_states

  -- Using distinct_count function name
  source=accounts | eventstats distinct_count(country) as unique_countries

  -- With partitioning
  source=accounts | eventstats dc(state) as state_count by gender
```

  ##Query Translation

```
  Input: source=accounts | eventstats dc(state) by gender
  Translates to: APPROX_DISTINCT_COUNT(state) OVER (PARTITION BY gender)
```

### Related Issues
Resolves #4052 

### Check List

### CODE REVIEW

File: docs/user/ppl/cmd/eventstats.rst [hunk 1]
```diff
@@ -277,6 +277,41 @@ Example::
...
+------------------
+
+Description
+>>>>>>>>>>>
+
+Usage: DISTINCT_COUNT(expr), DC(expr). Returns the approximate number of distinct values of expr using HyperLogLog++ algorithm. Both ``DISTINCT_COUNT`` and ``DC`` are equivalent and provide the same functionality.
+
+Example::
+
+    PPL> source=accounts | eventstats dc(state) as distinct_states;
```
dai-chen: np: probably we can use one example here for both distinct_count and dc.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -401,6 +402,11 @@ scalarWindowFunctionName
    | NTILE
    ;
 
+distinctCountWindowFunction
+   : DISTINCT_COUNT
+   | DC
```
dai-chen: Is `scalarWindowFunctionName` the rule for all window functions?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -558,8 +558,13 @@ public UnresolvedExpression visitBetween(OpenSearchPPLParser.BetweenContext ctx)
...
   public UnresolvedExpression visitWindowFunction(OpenSearchPPLParser.WindowFunctionContext ctx) {
-    Function f =
-        buildFunction(ctx.windowFunctionName().getText(), ctx.functionArgs().functionArg());
+    Function f;
+    if (ctx.windowFunctionName().distinctCountWindowFunction() != null) {
+      // All distinct count variants (dc, distinct_count, distinct_count_approx) use the UDAF
+      f = buildFunction("distinct_count_approx", ctx.functionArgs().functionArg());
+    } else {
+      f = buildFunction(ctx.windowFunctionName().getText(), ctx.functionArgs().functionArg());
+    }
```
dai-chen: Is this change still required if we add both to `scalarWindowFunctionName` in above comment?

dai-chen (commented):
Could you add UT in `CalcitePPLEventstatsTest` as well?

docs/user/ppl/cmd/eventstats.rst
ahkcs: Updated

ahkcs:
> Could you add UT in `CalcitePPLEventstatsTest` as well?

Added UT

ppl/src/main/antlr/OpenSearchPPLParser.g4
ahkcs: Updated to use under `scalarWindowFunctionName`

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ahkcs: Removed change

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1122,6 +1123,15 @@ void populate() {
           wrapSqlOperandTypeChecker(
               SqlStdOperatorTable.COUNT.getOperandTypeChecker(), COUNT.name(), false));
 
+      register(
```
dai-chen: This was missing previously? How this DISTINCT_COUNT_APPROX function registered previously?

dai-chen (approved):
Thanks for the changes!

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: It was previously registered here: OpenSearchExecutionEngine.java:281-287
I added registration in PPLFuncImpTable for Unit Tests

dai-chen: Is it registered twice? Should we remove it in OpenSearchExecutionEngine? cc: @yuancu

docs/user/ppl/cmd/eventstats.rst
```diff
@@ -277,6 +277,28 @@ Example::
     +----------------+-----------+----------------------+---------+--------+--------+----------+-------+-----+-----------------------+----------+--------------------+
 
 
+DISTINCT_COUNT, DC
```
penghuo: Add since 3.3 in doc

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: Seems like removing it on `OpenSearchExecutionEngine` would cause integration test to fail

dai-chen: Also I'm a little confused after another look at our `distinct_count` and `distinct_count_approx` function. Here we translate all to `distinct_count_approx` because Calcite window function doesn't support `COUNT(DISTINCT)`. Need to confirm if:

1. There is inconsistency in stats with or without pushdown
2. We either clarify all distinct function estimated or avoid pushing down for accurate result as SPL's

ahkcs: Hi @xinyual and @yuancu,

Could you help confirm the current behavior of the `dc` (distinct count) function in the `stats` command under two scenarios:

1. **Pushdown enabled**
2. **Pushdown disabled**

Specifically:

* Is the result of the `dc` function in `stats` exact or approximate in each case?
* For context, in the `eventstats` implementation, `dc` currently returns approximate results. I'd like to know if the same holds true for `stats` both with and without pushdown.

Thanks!

yuancu: @dai-chen Discussed with @xinyual , the primary reason that `DISTINCT_COUNT_APPROX` is registered in `OpenSearchExecutionEngine` is because it relies on a library from OpenSearch: `org.opensearch.search.aggregations.metrics.HyperLogLogPlusPlus`. PPL tries to not depend on any data engine specific implementation. Therefore, the function is registered externally in `OpenSearchExecutionEngine`.

This discussion may be helpful: https://github.com/opensearch-project/sql/pull/3654#discussion_r2115053978


For your questions:

> There is inconsistency in stats with or without pushdown

No, currently the `distinct_count_approx` in ppl uses the same implementation as in OpenSearch DSL(with `org.opensearch.search.aggregations.metrics.HyperLogLogPlusPlus`). 

> We either clarify all distinct function estimated or avoid pushing down for accurate result as SPL's

The approximation behavior is described here: [Cardinality aggregations - OpenSearch Doucmentation](https://docs.opensearch.org/latest/aggregations/metric/cardinality/#controlling-precision). Although it should be fine if we leave out this description since it aligns with OpenSearch's behavior, it is indeed better to mention the approximated behavior in the document.

yuancu: @ahkcs I think you can use the existing implementation of `APPROX_COUNT_DISTINCT`, without the need to register a new one with Calcite's implementation, which may cause inconsistency with the pushed-down results. You can do so by simply removing this register statement.

integ-test/src/test/resources/expectedOutput/calcite/explain_eventstats_distinct_count.json
```diff
@@ -0,0 +1,6 @@
+{
+  "calcite": {
+    "logical": "LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n  LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10], distinct_states=[APPROX_DISTINCT_COUNT($7) OVER (PARTITION BY $4)])\n    CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n",
+    "physical": "EnumerableLimit(fetch=[10000])\n  EnumerableWindow(window#0=[window(partition {4} aggs [APPROX_DISTINCT_COUNT($7)])])\n    CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname]], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"_source\":{\"includes\":[\"account_number\",\"firstname\",\"address\",\"balance\",\"gender\",\"city\",\"employer\",\"state\",\"age\",\"email\",\"lastname\"],\"excludes\":[]}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"
```
yuancu, line 4: It's not pushed down. Reusing existing implementation may solve this.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: @yuancu Thanks for the feedback! I tried removing the register statement as suggested, but it causes "Cannot resolve function: DISTINCT_COUNT_APPROX" errors during Calcite query planning, even though the OpenSearch implementation exists in OpenSearchExecutionEngine.java. 
In my initial implementation, I didn't have this register statement but the added unit tests requires this registration

integ-test/src/test/resources/expectedOutput/calcite/explain_eventstats_distinct_count.json
ahkcs: @yuancu You're right that it's not being pushed down. However, removing the registration in PPLFuncImpTable won't solve the push-down issue. The CalciteExplainIT goes through the external OpenSearch execution path and would use the existing implementation in `OpenSearchExecutionEngine.java`

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
yuancu: Hmm... that's weird. After I removed the function you registered, I tried to run this query: `source=opensearch-sql_test_index_account | eventstats dc(age) as unique_age by state`. It gets executed without a problem. It resolves to the function that we registered externally.

**Update**: I found that unit tests will indeed fail. This is because the function is registered externally when starting the cluster with OpenSearch engine, yet there isn't really any cluster spin up in the unit tests.

I think we can let go of the unit tests? There is no need to introduce such duplication / inconsistency for the sake of logical plan validation -- you can validate it in explain IT instead.

integ-test/src/test/resources/expectedOutput/calcite/explain_eventstats_distinct_count.json
yuancu: `stats dc(field)` can be pushed down -- `source=opensearch-sql_test_index_account | stats dc(age) as unique_age` is explained as the following plan:

```
{
  "calcite": {
    "logical": """LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalAggregate(group=[{}], unique_age=[COUNT(DISTINCT $0)])
    LogicalProject(age=[$8])
      CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])
""",
    "physical": """EnumerableLimit(fetch=[10000])
  CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[AGGREGATION->rel#22:LogicalAggregate.NONE.[](input=RelSubset#21,group={},unique_age=COUNT(DISTINCT $0))], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"unique_age":{"cardinality":{"field":"age"}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
"""
  }
}
```

~~You can imitate how `stats dc(field)` is pushed down to enable the same for  `eventstats dc(field)`.~~

**Update**: I found that all of the existing `eventstats agg_function(field)` are not pushed down. Maybe it's hard to push down window functions..

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: Updated implementation to remove UT and registration

dai-chen:
Hi @yuancu @ahkcs Based on my testing below, when `distinct_count` is not pushed down, Calcite executes it using its built-in implementation, which produces accurate results. When it is pushed down, it is translated into an OpenSearch cardinality aggregation, which returns approximate results. Is that correct?

| Command | **Without Pushdown** | **With Pushdown** |
|---|---|---|
| `stats distinct_count(field)` | **Exact** `COUNT(DISTINCT field)` computed in Calcite. | Uses `cardinality` aggregation: **close to accurate** if distinct ≤ `precision_threshold`, otherwise **approximate**. |
| `eventstats distinct_count(field)` | Currently uses the same HLL++ as cardinality aggregation via `distinct_count_approx` | N/A (Not clear how to push down window function yet) |


<pre>
# With pushdown
opensearchsql> explain source=accounts | stats distinct_count(age);
OpenSearchException({'calcite': {
'logical': 'LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n
  LogicalAggregate(group=[{}], distinct_count(age)=[COUNT(DISTINCT $0)])\n
    LogicalProject(age=[$2])\n
      CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])\n',
'physical': 'EnumerableLimit(fetch=[10000])\n
  CalciteEnumerableIndexScan(table=[[OpenSearch, accounts]],
    PushDownContext=[[AGGREGATION->rel#362:LogicalAggregate.NONE.[](input=RelSubset#361,
      group={},distinct_count(age)=COUNT(DISTINCT $0))],
      OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m",
        <b>"aggregations":{"distinct_count(age)":{"cardinality":{"field":"age"}}}},</b>
        requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n'}})

# Without pushdown
opensearchsql> explain source=accounts | stats distinct_count(age);
OpenSearchException({'calcite': {
'logical': 'LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n
  LogicalAggregate(group=[{}], distinct_count(age)=[COUNT(DISTINCT $0)])\n
    LogicalProject(age=[$2])\n
      CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])\n',
'physical': 'EnumerableLimit(fetch=[10000])\n
  EnumerableAggregate(group=[{}], <b>distinct_count(age)=[COUNT($0)])</b>\n
    EnumerableAggregate(group=[{2}])\n
      CalciteEnumerableIndexScan(table=[[OpenSearch, accounts]])\n'}})
</pre>

If there is inconsistency, I'm thinking the option may include:

1. **Keep current implementation**: Document that distinct_count / dc return exact (or close-to-exact) results for small cardinalities (≤ 3k by default) and approximate results above that. Ref: https://docs.opensearch.org/latest/aggregations/metric/cardinality/#controlling-precision
2. **Use accurate count function consistently**: Always compute exact distinct counts, either by reusing Calcite’s built-in implementation or by registering a new function. Document that this may lead to heavy resource usage and need to benchmark what's the performance.

cc: @penghuo @LantaoJin

penghuo:
> Keep current implementation: Document that distinct_count / dc return exact (or close-to-exact) results for small cardinalities (≤ 3k by default) and approximate results above that. Ref: https://docs.opensearch.org/latest/aggregations/metric/cardinality/#controlling-precision

+1, agree, it is short-term solution, dc/est_dc is same when pushdown to OpenSearch.

dai-chen (approved):
Thanks for the changes!

Please open a follow up PR to clarify the accuracy in documentation. Also we can consider extract aggregation functions doc shared by both stats and eventstats command.

PR #4094, opened 2025-08-21
Add documents on how to develop a UDF / UDAF

### LINKED ISSUE #4043: [FEATURE] Add documents on how to develop a new UDF/UDAF

Current frameworks for creating new user-defined functions (UDF) / user-defined aggregation functions (UDAF) is a little bit involved. Developers has to take some time to dive into different parts of the code base to learn how to create a new UDF / UDAF.

We should create a document somewhere with examples to instruct developers on how to create a new UDF / UDAF.

*originally discussed at https://github.com/opensearch-project/sql/pull/4024#pullrequestreview-3113758922*

yuancu:
### Description

This PR creates a dev doc *Developing PPL Functions* under folder docs/dev.

Additionally, it refactors the registration of aggregation functions to simplify its logic, making it similar to the creation and registration of non-aggregation functions to reduce learning barriers.

### Related Issues
Resolves #4043 

### Check List

### DIFF
--- a/DEVELOPER_GUIDE.rst
+++ b/DEVELOPER_GUIDE.rst
@@ -211,133 +211,13 @@ Java files are formatted using `Spotless <https://github.com/diffplug/spotless>`
    * - Javadoc format can be maintained by wrapping javadoc with `<pre></pre>` HTML tags
    * - Strings can be formatted on multiple lines with a `+` with the correct indentation for the string.
 
+Development Guidelines
+----------------------
 
-New PPL Command Checklist
-=========================
+For detailed development documentation, please refer to the `development documentation <docs/dev/index.md>`_. For specific guidance on implementing PPL components, see the following resources:
 
-If you are working on contributing a new PPL command, please read this guide and review all items in the checklist are done before code review. You also can leverage this checklist to guide how to add new PPL command.
-
-Prerequisite
-------------
-
-| ✅ Open an RFC issue before starting to code:
-- Describe the purpose of the new command
-- Include at least syntax definition, usage and examples
-- Implementation options are welcome if you have multiple ways to implement it
-
-| ✅ Obtain PM review approval for the RFC:
-- If PM unavailable, consult repository maintainers as alternative
-- An offline meeting might be required to discuss the syntax and usage
-
-Coding & Tests
---------------
-
-| ✅ Lexer/Parser Updates:
-- Add new keywords to OpenSearchPPLLexer.g4
-- Add grammar rules to OpenSearchPPLParser.g4
-- Update ``commandName`` and ``keywordsCanBeId``
-| ✅ AST Implementation:
-- Add new tree nodes under package ``org.opensearch.sql.ast.tree``
-- Prefer reusing ``Argument`` for command arguments **over** creating new expression nodes under ``org.opensearch.sql.ast.expression``
-| ✅ Visitor Pattern:
-- Add ``visit*`` in ``AbstractNodeVisitor``
-- Overriding ``visit*`` in ``Analyzer``, ``CalciteRelNodeVisitor`` and ``PPLQueryDataAnonymizer``
-| ✅ Unit Tests:
-- Extend ``CalcitePPLAbstractTest``
-- Keep test queries minimal
-- Include ``verifyLogical()`` and ``verifyPPLToSparkSQL()``
-| ✅ Integration tests (pushdown):
-- Extend ``PPLIntegTestCase``
-- Use complex real-world queries
-- Include ``verifySchema()`` and ``verifyDataRows()``
-| ✅ Integration tests (Non-pushdown):
-- Add test class to ``CalciteNoPushdownIT``
-| ✅ Explain tests:
-- Add tests to ``ExplainIT`` or ``CalciteExplainIT``
-| ✅ Unsupported in v2 test:
-- Add a test in ``NewAddedCommandsIT``
-| ✅ Anonymizer tests:
-- Add a test in ``PPLQueryDataAnonymizerTest``
-| ✅ Cross-cluster Tests (optional, nice to have):
-- Add a test in ``CrossClusterSearchIT``
-| ✅ User doc:
-- Add a xxx.rst under ``docs/user/ppl/cmd`` and link the new doc to ``docs/user/ppl/index.rst``
-
-Developing PPL Functions
-========================
-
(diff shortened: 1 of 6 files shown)


DEVELOPER_GUIDE.rst
```diff
@@ -217,24 +209,33 @@ Java files are formatted using `Spotless <https://github.com/diffplug/spotless>`
    * - Javadoc format can be maintained by wrapping javadoc with `<pre></pre>` HTML tags
    * - Strings can be formatted on multiple lines with a `+` with the correct indentation for the string.
 
-New PPL Command Checklist
-=========================
+Developing PPL Commands and Functions
```
LantaoJin: This will break the the [link](https://github.com/opensearch-project/sql/blob/fcdb7882132442e8ad88c15535f3e82f7d979264/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L13).

I think we you can add a new section out of `New PPL Command Checklist`

yuancu: I thought it was too heavy to have two sections for developing functions and commands, considering most of other sections are high level instructions.

But somehow I cannot add lower level sections to this document. I updated them to separate sections.

DEVELOPER_GUIDE.rst
```diff
@@ -217,21 +211,24 @@ Java files are formatted using `Spotless <https://github.com/diffplug/spotless>`
...
+| ✅ Open an RFC issue describing the command:
+- Specify the purpose and need for the new command
+- Include syntax definition, usage and examples
+- Outline implementation options if multiple approaches exist
+
 | ✅ Obtain PM review approval for the RFC:
-- If PM unavailable, consult repository maintainers as alternative
-- An offline meeting might be required to discuss the syntax and usage
+- Consult repository maintainers if PM is unavailable
+- Be prepared for meetings to discuss syntax and usage details
```
yuancu: Modified by LLM unintentionally, should be ok though

DEVELOPER_GUIDE.rst
```diff
@@ -267,6 +264,81 @@ Coding & Tests
...
+Developing User-Defined Functions (UDFs)
+----------------------------------------
+
+| ✅ Creating UDFs: A user-defined function is an instance of ``SqlOperator`` that transforms input row expressions (``RexNode``) into a new one. There are three approaches to implementing UDFs:
+
+- Use existing Calcite operators: Leverage operators already declared in Calcite's ``SqlStdOperatorTable`` or ``SqlLibraryOperators``, and defined in ``RexImpTable.java``
+- Adapt existing static methods: Convert Java static methods to UDFs using utility functions like ``UserDefinedFunctionUtils.adaptExprMethodToUDF``
+- Implement from scratch
+
+  * Implement the ``ImplementorUDF`` interface
```
qianheng-aws: Please call out that developers should ensure their UDF calculating logic is efficient by implementing data-irrelevant logic during the compilation phase rather than at runtime. In other words, that part of logic should preferably use ling4j's expression instead of internal static method calls.

qianheng-aws: It's usually ignored in our PRs implementation for UDF.

qianheng-aws (approved):

DEVELOPER_GUIDE.rst
```diff
@@ -217,21 +211,24 @@ Java files are formatted using `Spotless <https://github.com/diffplug/spotless>`
...
 
 Prerequisite
 ------------
 
-| ✅ Open an RFC issue before starting to code:
-- Describe the purpose of the new command
-- Include at least syntax definition, usage and examples
-- Implementation options are welcome if you have multiple ways to implement it
+| ✅ Open an RFC issue describing the command:
+- Specify the purpose and need for the new command
```
LantaoJin: Is the new description really better than the original?

The LLM generated content changed the level of requirements. Some info I tried to callout is missing. For example, offline meeting is not necessary, syntax, usage and examples are minimal requirements of RFC, implementation details are not necessary in RFC, etc.

Can you revert the LLM generated suggestions except grammar fixing?

LantaoJin (commented):
~~why this PR includes code refactoring?~~

Swiddis:
This should go in the dev docs at `docs/dev` instead of being put directly in the developer guide.

Dev guide should be more high level on how to build the project and maybe what the core entrypoints are. Making this its own doc will also give more room to add detail in e.g. lines like "Creating UDFs: A user-defined function is an instance of `SqlOperator`" (What is `SqlOperator`?)

DEVELOPER_GUIDE.rst
yuancu: Sorry for the unnecessary change. I have reverted them.

DEVELOPER_GUIDE.rst
yuancu: Thanks. I added such comments. I'm thinking about adding an example so that it will be easier to comprehend. Does the following example fit the case?

```java
@Override
public Expression implement(RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
     // Don't
    SqlTypeName type = call.getOperands().getFirst().getType().getTypeName();
    List<Expression> operands = Lists.concat(translatedOperands, List.of(Expressions.constant(type));
    return Expressions.call(Implementor.class, "allInOneOp", operands)

    // Do
    String opName = type == FLOAT ? "floatOp" : "integerOp";
    return Expressions.call(Implementor.class, , translatedOperands);
}

public static int allInOneOp(Number n, SqlTypeFamily t){
    if (t == FLOAT) { ... }
    else if (t == INTEGER) { .... }
}

public static int floatOp(float n){
    ...
}

public static int integerOp(integer n) {
   ...
}
```

yuancu:
> This should go in the dev docs at `docs/dev` instead of being put directly in the developer guide.
> 
> Dev guide should be more high level on how to build the project and maybe what the core entrypoints are. Making this its own doc will also give more room to add detail in e.g. lines like "Creating UDFs: A user-defined function is an instance of `SqlOperator`" (What is `SqlOperator`?)

That makes sense. I have moved both the guidance for commands and functions from DEVELOPER_GUIDE to docs/dev

qianheng-aws (approved):

ykmr1224:
> @yuancu
Should we backport this to 2.19-dev?
It is causing backport failure for https://github.com/opensearch-project/sql/pull/4100

PR #4106, opened 2025-08-22
strftime function implementation

vamsimanohar:
### Summary

  Implements the STRFTIME function for OpenSearch SQL PPL with Calcite engine, providing
  timestamp formatting functionality.

 #### Description

  This PR adds the STRFTIME function that formats Unix timestamps into human-readable
  date/time strings using format specifiers.

 ####  Key Features

  - Unix Timestamp Support: Accepts Unix timestamps in seconds
  - Extended Support:
    - Handles millisecond timestamps (auto-detected when value > 100000000000)
    - Supports nanosecond precision through floating-point seconds
    - Works with PPL timestamp types (from now(), from_unixtime(), etc.)
  - Type Safety: Explicitly accepts only NUMERIC and TIMESTAMP types
    - String inputs are NOT supported (use unix_timestamp() for conversion)

   #### Implementation Details

  - Input Types Supported:
    - INTEGER/LONG: Unix timestamps in seconds or milliseconds (auto-detected)
    - DOUBLE/FLOAT: Unix timestamps with fractional seconds
    - TIMESTAMP: From functions like now(), from_unixtime(), timestamp()
  - Format Specifiers: Full support for standard strftime format codes
    - Date: %Y (year), %m (month), %d (day), etc.
    - Time: %H (hour), %M (minute), %S (second), etc.
    - Shortcuts: %F (YYYY-MM-DD), %T (HH:MM:SS), etc.

   #### Usage Examples

```
  -- Format Unix timestamp
  source=logs | eval formatted = strftime(1521467703, "%Y-%m-%d %H:%M:%S")
  -- Result: "2018-03-19 13:55:03"
```

```
  -- Format with milliseconds
  source=logs | eval formatted = strftime(1521467703123, "%F %T")
  -- Result: "2018-03-19 13:55:03"
```

```
  -- Format current time
  source=logs | eval current = strftime(now(), "%B %d, %Y")
  -- Result: "September 02, 2025"
```

```
  -- Convert string date first (strings not directly supported)
  source=logs | eval ts = unix_timestamp("2018-03-19 13:55:03")
             | eval formatted = strftime(ts, "%m/%d/%Y")
  -- Result: "03/19/2018"
```

   #### Notes

  - All timestamps are interpreted as UTC timezone
  - Text formatting uses language-neutral Locale.ROOT for consistency
  - Calcite may attempt implicit type conversion for string literals, but the function
  itself only accepts NUMERIC and TIMESTAMP types

### Related Issues
Resolves #[st]

### Check List


core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java
```diff
@@ -191,6 +193,7 @@ public Expression visitRelevanceFieldList(RelevanceFieldList node, AnalysisConte
   @Override
   public Expression visitFunction(Function node, AnalysisContext context) {
     FunctionName functionName = FunctionName.of(node.getFuncName());
+    validateCalciteOnlyFunction(functionName);
```
dai-chen, line 196: Is this necessary since we're enabling Calcite by default in 3.3? I assume there are many functions only work with Calcite?

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/StrftimeFunction.java
```diff
@@ -0,0 +1,205 @@
...
+ * Implementation of the STRFTIME function. This function takes a UNIX timestamp (in seconds) and
+ * formats it according to POSIX-style format specifiers.
+ *
+ * <p>All timestamps are interpreted as UTC and all text formatting uses Locale.ROOT for consistent,
+ * language-neutral output regardless of server locale settings.
+ *
+ * <p>Signatures: - (DOUBLE, STRING) -> STRING - (LONG, STRING) -> STRING - (TIMESTAMP, STRING) ->
+ * STRING - (STRING, STRING) -> STRING
+ */
+public class StrftimeFunction extends ImplementorUDF {
```
dai-chen, line 39: Does our existing V2 function can help simplify here? e.g, https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/datetime.rst#date-format, https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/datetime.rst#from-unixtime

integ-test/src/test/resources/expectedOutput/calcite/explain_strftime_function.json
```diff
@@ -0,0 +1,6 @@
+{
+  "calcite": {
+    "logical": "LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n  LogicalSort(fetch=[1])\n    LogicalProject(formatted_date=[STRFTIME(1521467703, '%Y-%m-%d':VARCHAR)])\n      CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n",
```
dai-chen, line 3: np: do we need such test for UDF?

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/StrftimeFunction.java
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Format the datetime using the format string.
+   *
+   * @param dateTime The datetime to format
+   * @param format The format string
+   * @return Formatted string or null if formatting fails
+   */
+  private static String formatDateTime(ZonedDateTime dateTime, String format) {
+    ExprValue result = StrftimeFormatterUtil.formatZonedDateTime(dateTime, format);
+    return result.stringValue();
```
noCharger, line 183: Could this cause NPE if the result is null?

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/StrftimeFunction.java
vamsimanohar: Date_format uses mysql style identifiers and
strftime time uses POSIX style.
Do you see a reuse?

core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java
vamsimanohar: It is part of checklist written to send out the same message for all these calcite only functions.

integ-test/src/test/resources/expectedOutput/calcite/explain_strftime_function.json
vamsimanohar: Not really, these are also part of checklist.

core/src/main/java/org/opensearch/sql/expression/datetime/StrftimeFormatterUtil.java
```diff
@@ -0,0 +1,249 @@
...
+  /**
+   * Extract the first 10 digits from a timestamp to get UNIX seconds. This handles timestamps that
+   * may be in milliseconds or nanoseconds.
+   *
+   * @param timestamp The timestamp value
+   * @return UNIX timestamp in seconds
+   */
+  public static long extractUnixSeconds(double timestamp) {
+    // Return as-is if negative or already in valid seconds range
+    if (timestamp <= MAX_UNIX_TIMESTAMP) {
```
noCharger: this logic will not handle large negative millisecond timestamp

core/src/main/java/org/opensearch/sql/expression/datetime/StrftimeFormatterUtil.java
```diff
@@ -0,0 +1,249 @@
...
+    matcher.appendTail(result);
+
+    return result.toString();
+  }
+
+  /** Parse precision value for subsecond formats. */
+  private static int parsePrecision(String precisionStr, String type) {
+    if (precisionStr != null) {
+      return Integer.parseInt(precisionStr);
+    }
```
noCharger, line 176:
```
    public static int parseInt(String s) throws NumberFormatException {
        return parseInt(s, 10);
    }
```

Do we need to handle exception here?

core/src/main/java/org/opensearch/sql/expression/datetime/StrftimeFormatterUtil.java
```diff
@@ -0,0 +1,249 @@
...
+        String replacement = entry.getValue().format(dateTime);
+        result = result.replace(specifier, replacement);
+      }
+    }
+    return result;
+  }
+
+  /** Handle %N and %Q subsecond formats with optional precision. */
+  private static String handleSubSecondFormats(String format, ZonedDateTime dateTime) {
+    StringBuffer result = new StringBuffer();
```
noCharger: Is it necessary to use StringBuffer instead of StringBuilder here?

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/StrftimeFunction.java
vamsimanohar: Only if dataTime is null and it doesn't happen as there enough null checks in the calling methods.

core/src/main/java/org/opensearch/sql/expression/datetime/StrftimeFormatterUtil.java
vamsimanohar: Not a checked exception right, let it bubble to the customer.

core/src/main/java/org/opensearch/sql/expression/datetime/StrftimeFormatterUtil.java
vamsimanohar: sure

core/src/main/java/org/opensearch/sql/expression/datetime/StrftimeFormatterUtil.java
vamsimanohar: Good one..although the above function is used only in unit tests...we are n't handling negative timestamps properly. Will make that change.

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/StrftimeFunction.java
dai-chen: what's the difference?

core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java
dai-chen: Okay, I wasn't aware of that. So we need to add all Calcite only functions to the list you created?

vamsimanohar: I checked again, its only for commands. https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md
```
Unsupported in v2 test:
Add a test in NewAddedCommandsIT
```

I think we can skip for functions. I thought its good to have.

core/src/main/java/org/opensearch/sql/expression/function/udf/datetime/StrftimeFunction.java
vamsimanohar:
```
✅ Specifiers unique to MySQL 

%D → Day of month with English suffix (1st, 2nd, 3rd…).

%h → Hour (01–12) (alias for %I in MySQL).

%i → Minutes (00–59).

%r → 12-hour time with AM/PM (hh:mm:ss AM/PM).

%U / %u / %V / %v / %X / %x → MySQL’s special week number / year specifiers tied to WEEK() function modes.

%M → Full month name (January..December).

⚠️ In POSIX, %M = minutes, not month.

%c → Month numeric (0–12).

⚠️ In POSIX, %c = full date/time string.

%k → Hour (0–23, no leading zero).

%l → Hour (1–12, no leading zero).

%s → Seconds (00–59).

⚠️ In POSIX, %s = Unix epoch timestamp.

%a → Abbreviated weekday (Sun..Sat).

⚠️ In POSIX, %a = abbreviated weekday too, but starts with Monday (Mon..Sun).
```

```
✅ Specifiers unique to POSIX/Java 

(not present in MySQL)

%A → Full weekday name (Monday..Sunday).

%B → Full month name (January..December).

%C → Century (two digits).

%E z → Timezone offset in minutes from UTC.

%F → ISO 8601 date (YYYY-MM-DD).

%g / %G → ISO week-based year (2-digit / 4-digit).

%N → Subsecond digits with precision (nanoseconds).

%Q → Fractional seconds with precision (default ms).

%s → Unix epoch seconds.

%x → Date (MM/dd/yyyy format).

%X → Time (HH:mm:ss format).

%z / %:z / %::z / %:::z → Timezone offset formats.

%Z → Timezone abbreviation (EST, PDT).
```

penghuo (approved):

PR #4107, opened 2025-08-22
FollowUp: Disable GCedMemoryUsage if no concurrent GC MXBean

LantaoJin:
### Description
https://github.com/opensearch-project/sql/pull/3983 enables `GCedMemoryUsage` for Calcite. But when the Java version is less than 21, there is no `ConcurrentMXBean` returned from `ManagementFactory.getGarbageCollectorMXBeans()`. Fallback it to `RuntimeMemoryUsage`.

### Related Issues

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/ResourceMonitorIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/ResourceMonitorIT.java
@@ -13,6 +13,7 @@ import org.json.JSONObject;
 import org.junit.Test;
 import org.opensearch.client.ResponseException;
 import org.opensearch.sql.common.setting.Settings;
+import org.opensearch.sql.opensearch.monitor.GCedMemoryUsage;
 
 public class ResourceMonitorIT extends PPLIntegTestCase {
 
@@ -27,11 +28,7 @@ public class ResourceMonitorIT extends PPLIntegTestCase {
     // update plugins.ppl.query.memory_limit to 1%
     updateClusterSettings(
         new ClusterSetting("persistent", Settings.Key.QUERY_MEMORY_LIMIT.getKeyValue(), "1%"));
-    // ClickBench Q30 is high memory consumption query.
-    String query = sanitize(loadFromFile("clickbench/queries/q30.ppl"));
-
-    ResponseException exception =
-        expectThrows(ResponseException.class, () -> executeQueryMultipleTimes(query));
+    ResponseException exception = expectThrows(ResponseException.class, this::executeQuery);
     assertEquals(500, exception.getResponse().getStatusLine().getStatusCode());
     assertThat(
         exception.getMessage(),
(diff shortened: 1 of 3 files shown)


integ-test/src/test/java/org/opensearch/sql/ppl/ResourceMonitorIT.java
```diff
@@ -21,17 +19,19 @@ public class ResourceMonitorIT extends PPLIntegTestCase {
...
   }
 
   @Test
   public void queryExceedResourceLimitShouldFail() throws IOException {
     // update plugins.ppl.query.memory_limit to 1%
     updateClusterSettings(
         new ClusterSetting("persistent", Settings.Key.QUERY_MEMORY_LIMIT.getKeyValue(), "1%"));
-    String query = String.format("search source=%s age=20", TEST_INDEX_DOG);
+    // ClickBench Q30 is high memory consumption query.
+    String query = sanitize(loadFromFile("clickbench/queries/q30.ppl"));
```
LantaoJin: The original test might not trigger GC in single run. Use ClickBench Q30 and execute it 10 times to make sure some CGC could be triggered.

LantaoJin:
No need to backport to 2.19-dev. The fixing is included in https://github.com/opensearch-project/sql/pull/4105

integ-test/src/test/java/org/opensearch/sql/ppl/ResourceMonitorIT.java
penghuo: Does this mean the new solution is less aggressive toward failed queries compared to a real memory check?
And in both case, CircuitBreaker still not triggered? Does it means our solution is still aggregesive?

opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/GCedMemoryUsage.java
```diff
@@ -55,13 +56,24 @@ public void setUsage(long value) {
...
   }
 
+  public static boolean initialized() {
+    return initialized;
+  }
+
   private void registerGCListener() {
-    boolean registered = false;
     List<GarbageCollectorMXBean> gcBeans = ManagementFactory.getGarbageCollectorMXBeans();
+    if (gcBeans.stream()
```
penghuo, line 65: Did we get runtime exception on 2.19-dev branch? Which JDK version this feature depend on?

LantaoJin: Haven't backported to 2.19-dev. I am still working on backporting https://github.com/opensearch-project/sql/pull/4105. Then the JDK version < 20, there is no specific `G1 Concurrent GC` Monitoring registered, ref https://github.com/openjdk/jdk/blob/jdk-19-ga/src/hotspot/share/gc/g1/g1MonitoringSupport.cpp#L91. From JDK 20, `G1 Concurrent GC` was registered in Monitoring, ref https://github.com/openjdk/jdk/blob/jdk-20-ga/src/hotspot/share/gc/g1/g1MonitoringSupport.cpp#L93.

This PR is to verify if there a Concurrent MXBean registered in JDK before initialing a GCedMemoryUsage.

qianheng-aws (approved):

Issue #4108, opened 2025-08-22
[RFC] Add `rex` Command to PPL (Calcite Engine)


RyanL1997:
## Problem Statement

In data analysis and log processing workflows, users frequently need to extract structured information from unstructured text fields and perform text transformations, for example:

- Extracting email usernames and domains from user registration data
- Parsing log entries to extract timestamp, severity level, and error codes
- Transforming sensitive data by masking or anonymizing field values
- Extracting multiple values from comma-separated or space-delimited fields
- Converting field formats for standardization (e.g., phone numbers, addresses)

Currently, PPL lacks comprehensive text processing capabilities, forcing users to rely on basic string functions or external processing tools. This limits the ability to perform complex field extraction and transformation tasks within PPL pipelines.

## Current State

- PPL provides basic string functions like `substr`, `concat`, and `trim`
- Limited pattern-based field extraction capabilities
- No native support for named capture groups or complex regex operations
- Users must resort to multiple commands or external tools for text processing workflows
- No standardized approach for field masking or data anonymization

## Long-Term Goals

- Provide comprehensive text processing capabilities within PPL pipelines
- Enable efficient server-side pattern matching and field extraction
- Support both field extraction and in-place text transformation operations
- Establish consistent patterns for text processing commands in PPL
- Enable complex data parsing workflows for log analysis and data preparation

### High-Level Functionality

The `rex` command provides two primary modes for text processing:

1. **Field Extraction Mode (Default)**: Extract structured data from text using named capture groups
2. **Text Transformation Mode**: Modify field values using find-and-replace operations

#### Field Extraction Mode
```bash
source | rex [field=<field>] "<pattern>" [max_match=<int>] [offset_field=<string>]
```

#### Text Transformation Mode
```bash
source | rex [field=<field>] mode=sed "<sed-expression>"
```

### Parameters

- **field**: Source field to process (defaults to _raw field handling)
- **pattern**: Regular expression with named capture groups for extraction
- **max_match**: Controls number of matches extracted (1=single value, >1=array, 0=unlimited)
- **offset_field**: Creates field tracking character positions of matches
- **mode=sed**: Enables text transformation mode using sed-style expressions

#### Basic Named Group Extraction
```bash
# Extract email components
source=accounts | rex field=email "(?<username>[^@]+)@(?<domain>[^.]+)\.(?<tld>.+)" | fields username, domain, tld

# Parse log entries  
[...]

PR #4109, opened 2025-08-22
[Feature] Core Implementation of `rex` Command In PPL

RyanL1997:
### Description
Core logic implementation of `rex` Command In PPL 

### Details:
  - Core logic of `rex` (Extract Mode Only)
  ```bash
  <source> | [commands] | rex field=<field> "<pattern>" | [commands]
  <source> | [commands] | rex field=<field> mode=extract "<pattern>" | [commands]  
  <source> | [commands] | rex field=<field> "<pattern>" max_match=<n> | [commands]
  ```
  - `field`: The field name to extract data from
  - `pattern`: Java regex pattern with named capture groups (?<name>...) for field extraction
  - `mode=extract`: Extract mode for field extraction (default and only supported mode)
  - `max_match`: Maximum number of matches to capture (returns array of matches)

**Supported Functionality:**
- **Single extraction**: Extract one match per named capture group
- **Multi pattern extraction**: Extract from multiple named capture groups in one pattern  
- **Max match**: Use `max_match=N` for multiple occurrences with both single and multiple groups

**Note**: This implementation focuses on extract mode only. SED mode and offset field functionality have been intentionally excluded to keep this PR focused and manageable.

For the handling of default field (`_source` / `_raw`): 
- Tracking at issue https://github.com/opensearch-project/sql/issues/4111

### Related Issues
* Relate #4108 

### Related Analysis
* UDF vs Native Calcite Function: https://github.com/opensearch-project/sql/issues/4108#issuecomment-3231105679

### Check List
- [x] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/DEVELOPER_GUIDE.rst#new-ppl-command-checklist) all confirmed.


docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
+=============
+rex
```
penghuo, line 2: Add since 3.3

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+Description
+============
+| The ``rex`` command extracts fields from a raw text field using regular expression named capture groups and optionally filters events that match the extraction pattern.
+
+
+Syntax
+============
+rex [mode=<mode>] field=<field> <pattern> [max_match=<int>] [offset_field=<string>]
+
+* field: mandatory. The field must be a text field to extract data from.
```
penghuo: text field -> string field.  As text has specific meaning in OpenSearch, it will confuse OpenSearch user.

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+* pattern: mandatory string. The regular expression pattern with named capture groups used to extract new fields. Pattern must contain at least one named capture group using ``(?<name>pattern)`` syntax.
+* mode: optional. Either ``extract`` (default) or ``sed``. In extract mode, creates new fields from named capture groups. In sed mode, performs text substitution on the field.
+* max_match: optional integer (default=1). Maximum number of matches to extract. If greater than 1, extracted fields become arrays.
+* offset_field: optional string. Field name to store the character offset positions of matches.
+
+Named Capture Group Limitations
```
penghuo: Move this sections into limitations, it is clearly state in Java regex doc.

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+Example 1: Basic Field Extraction
+==================================
+
+Extract username and domain from email addresses using named capture groups.
+
+PPL query::
+
+    os> source=accounts | rex field=email "(?<username>[^@]+)@(?<domain>[^.]+)" | fields email, username, domain | head 3 ;
+    fetched rows / total rows = 3/3
+    +-----------------------+-------------+---------+
```
penghuo: Format bellow table, https://github.com/RyanL1997/sql/blob/d31ab7567c0449191c1e68cb9355d2edc08baa48/docs/user/ppl/cmd/rex.rst

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+    | amberduke@pyrami.com  | amberduke   | pyrami  |
+    | hattiebond@netagy.com | hattiebond  | netagy  |
+    | nanettebates@quility.com | nanettebates | quility |
+    +-----------------------+-------------+---------+
+
+
+Example 2: Extract with Filtering
+==================================
+
+The rex command automatically filters out events that don't match the pattern.
```
penghuo: > automatically filters out events

IMO, if address value don't match pattern streetnum and streetname should be null, but the entire envent should be keeped. right?

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+.. rubric:: Table of contents
+
+.. contents::
+   :local:
+   :depth: 2
+
+
+Description
+============
+| The ``rex`` command extracts fields from a raw text field using regular expression named capture groups and optionally filters events that match the extraction pattern.
```
penghuo: > optionally filters

Is there any paramaters control it?

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+
+
+Syntax
+============
+rex [mode=<mode>] field=<field> <pattern> [max_match=<int>] [offset_field=<string>]
+
+* field: mandatory. The field must be a text field to extract data from.
+* pattern: mandatory string. The regular expression pattern with named capture groups used to extract new fields. Pattern must contain at least one named capture group using ``(?<name>pattern)`` syntax.
+* mode: optional. Either ``extract`` (default) or ``sed``. In extract mode, creates new fields from named capture groups. In sed mode, performs text substitution on the field.
```
penghuo: Add explaination of sed syntax.

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+* max_match: optional integer (default=1). Maximum number of matches to extract. If greater than 1, extracted fields become arrays.
+* offset_field: optional string. Field name to store the character offset positions of matches.
```
penghuo: what is offset_field format, comma seperated key=value pair string?

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+
+
+Limitations
+===========
+
+There are several important limitations with the rex command:
+
+**Named Capture Group Naming:**
+
+- Named capture groups cannot contain underscores due to Java regex limitations
```
penghuo, line 181: We can call-out only java regex is supported, and avoid mention Perl and other flavors in multiple place

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+- Pattern must contain at least one named capture group
+- Regular capture groups ``(...)`` without names are not allowed
+- Pattern must use Java regex syntax, not Perl or other flavors
+
+**Field Behavior:**
+
+- rex automatically filters out events that don't match the pattern
+- In extract mode, creates new fields from named capture groups
+- In sed mode, modifies the original field content
+- When max_match > 1, extracted fields become arrays
```
penghuo: It is not limitation right?

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+
+**Performance Considerations:**
+
+- Complex patterns with large max_match values may impact performance
+- Consider the specificity of patterns to avoid unnecessary processing
+- Use anchors (^ $) appropriately to improve matching efficiency
```
penghuo: What we want to call-out? do we have releated to test-data?

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,220 @@
...
+
+**Integration:**
+
+- Extracted fields can be used in subsequent PPL commands (where, stats, sort, etc.)
+- sed mode modifications affect the field for subsequent operations
+- offset_field provides debugging information for pattern matching
```
penghuo: What we want to call-out in here?, it is not limitation.

integ-test/src/test/java/org/opensearch/sql/calcite/CalciteNoPushdownIT.java
```diff
@@ -82,6 +82,8 @@
   CalcitePrometheusDataSourceCommandsIT.class,
   CalciteQueryAnalysisIT.class,
   CalciteRareCommandIT.class,
+  CalciteRexCommandIT.class,
+  CalciteRenameCommandIT.class,
```
penghuo, line 90: remove CalciteRenameCommandIT

integ-test/src/test/java/org/opensearch/sql/ppl/RexCommandIT.java
```diff
@@ -0,0 +1,181 @@
...
+
+  @Override
+  public void init() throws Exception {
+    super.init();
+    enableCalcite();
+    loadIndex(Index.ACCOUNT);
+  }
+
+  @Test
+  public void testRexBasicFieldExtraction() throws IOException {
```
penghuo: what if rex extract group field name conflict with existing field name?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -268,6 +268,26 @@ public void testExplainOnAggregationWithSumEnhancement() throws IOException {
                 TEST_INDEX_BANK)));
   }
 
+  @Test
+  public void testRexExplain() throws IOException {
+    String query =
+        "source=opensearch-sql_test_index_account | rex field=lastname \\\"(?<initial>^[A-Z])\\\" |"
```
penghuo, line 436: `\\\"` can be simpliy to `\"`

docs/user/ppl/cmd/rex.rst
RyanL1997: Yes, currently is like this:
```bash
# comma separator, pattern order
username=0-12,domain=14-22
```

Do we need to switch to some of the other format like the following ?
```bash
# & separator
domain=14-22&username=0-12 
```

docs/user/ppl/cmd/rex.rst
RyanL1997: Fixed, added a version section here: https://github.com/opensearch-project/sql/pull/4109/files#diff-1207ed670a85a160ac986d8775072cb8ae114a599f4c076dae3e740c4d5d687aR16

docs/user/ppl/cmd/rex.rst
RyanL1997: Added.

docs/user/ppl/cmd/rex.rst
RyanL1997: nice catch, and i have fixed all the tables in the rst doc

docs/user/ppl/cmd/rex.rst
RyanL1997: This is not a limitation.

docs/user/ppl/cmd/rex.rst
RyanL1997: For this section, I was trying to callout some best practice for how to construct the rex command. For example, this specific line you highlighted, I mentioned the to use specific patterns when possible (e.g., `"(?<email>[^@]+@[^.]+\.com)"` vs `"(?<email>.*@.*)"`), because anchored patterns reduce unnecessary backtracking in regex engine. 

However, just like you suggested in one of the other comments, we can just let user to reference to the official java regular expression doc for these best practice following tips, since it is more focused on the regular expression usage itself, not the actual rex command.

docs/user/ppl/cmd/rex.rst
RyanL1997: This is not limitation. I was trying to add some description of potential use case of `rex` command. I have removed this section, for following reason:
- checked the other commands' rst file, none of them has this specific section, and they just directly jump into the examples

According to the above, I have removed it. Thanks for pointing it out.

docs/user/ppl/cmd/rex.rst
RyanL1997: fixed.

integ-test/src/test/java/org/opensearch/sql/calcite/CalciteNoPushdownIT.java
RyanL1997: good catch, fixed.

docs/user/ppl/cmd/rex.rst
RyanL1997: Fixed, I have removed the wording of: 
> "- Pattern must use Java regex syntax, not Perl or other flavors"

docs/user/ppl/cmd/rex.rst
RyanL1997: I have actually removed this "Field Behavior" section from "Limitation", for following reasons:
- First, these are not limitations
- We have covered this content in our "Syntax" section

According to the above, I have removed them entirely. Thanks for calling it out.

docs/user/ppl/cmd/rex.rst
RyanL1997: fixed.

docs/user/ppl/cmd/rex.rst
RyanL1997: In the current revision, I have removed this "Performance Consideration" section which were related to these regex usage specific tips. Instead, I have added a line to reference user to the official java doc for more info:
> - For detailed Java regex pattern syntax and usage, refer to the `official Java Pattern documentation <https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html>`_

at https://github.com/opensearch-project/sql/pull/4109/files#diff-1207ed670a85a160ac986d8775072cb8ae114a599f4c076dae3e740c4d5d687aR194

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
RyanL1997: Since the test constructs JSON directly in Java, I need another layer of escaping like this, here is why:
  - First `\` escapes the second `\` in Java
  - Second `\` escapes the `"` in JSON
  - Result: `\\\"` in Java code → `\"` in JSON → `"` when parsed
  
  I have also done the experiment of changing it to `\"` and it does give me the following 400:
```bash
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Failed to parse request payload"}],"type":"illegal_argument_exception","reason":"Failed to parse request payload","caused_by":{"type":"j_s_o_n_exception","reason":"Expected a ',' or '}' at 78 [character 76 line 2]"}},"status":400}
```

ppl/src/test/java/org/opensearch/sql/ppl/parser/AstBuilderTest.java
```diff
@@ -971,6 +971,6 @@ private String mappingTable(String indexName) {
   @Test(expected = IllegalArgumentException.class)
   public void testBinCommandDuplicateParameter() {
     // Test that duplicate parameters throw an exception
-    plan("search source=test | bin field span=10 span=20");
+    plan("search source=test | bin index_field span=10 span=20");
```
RyanL1997, line 1080: Hi @ahkcs , could you review this change to see if it is making sense? After I rebased with you bin cmd change, I need this change since the syntax `field` as been added into the grammar for rex command, so this test instead of returning a illegal argument error, it starts returning antlr error for complained about the grammar usage of this:

```bash
ArgumentFactoryTest > testBinCommandDuplicateParameter FAILED
    java.lang.Exception: Unexpected exception, expected<java.lang.IllegalArgumentException> but was<org.opensearch.sql.common.antlr.SyntaxCheckException>
        at org.junit.internal.runners.statements.ExpectException.evaluate(ExpectException.java:30)
...
```
I got the above from a CI run (https://github.com/opensearch-project/sql/actions/runs/17224987607/job/48867804413?pr=4109#step:4:5421) , and I have also verified this by my local testing.

ahkcs: This change makes sense, we should apply this fix on all similar test cases

RyanL1997: Thanks for the confirmation. Yes, and seems like this is the only place that we need to fix it - the CIs are passing now.

integ-test/src/test/java/org/opensearch/sql/ppl/RexCommandIT.java
RyanL1997: Currently, it will overwrite the existing fields when rex extracts fields with conflicting names. Here is an example:

```bash
 curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d'{
    "query": "source=accounts | rex field=lastname \"(?<firstname>[A-Z])\" | head 3"
  }' | jq

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1375  100  1285  100    90  16475   1153 --:--:-- --:--:-- --:--:-- 17857
{
  "schema": [
    {
      "name": "account_number",
      "type": "bigint"
    },
    {
      "name": "address",
      "type": "string"
    },
    {
      "name": "balance",
      "type": "bigint"
    },
    {
      "name": "gender",
      "type": "string"
    },
    {
      "name": "city",
      "type": "string"
    },
    {
      "name": "employer",
      "type": "string"
    },
    {
      "name": "state",
      "type": "string"
    },
    {
      "name": "age",
      "type": "bigint"
    },
    {
      "name": "email",
      "type": "string"
    },
    {
      "name": "lastname",
      "type": "string"
    },
    {
      "name": "firstname",
      "type": "string"
    }
  ],
  "datarows": [
    [
      1,
      "880 Holmes Lane",
      39225,
      "M",
      "Brogan",
      "Pyrami",
      "IL",
      32,
      "amberduke@pyrami.com",
      "Duke",
      "D"
    ],
    [
      6,
      "671 Bristol Street",
      5686,
      "M",
      "Dante",
      "Netagy",
      "TN",
      36,
      "hattiebond@netagy.com",
      "Bond",
      "B"
    ],
    [
      13,
      "789 Madison Street",
      32838,
      "F",
      "Nogal",
      "Quility",
      "VA",
      28,
      "nanettebates@quility.com",
      "Bates",
      "B"
    ]
  ],
  "total": 3,
  "size": 3
}
```

docs/user/ppl/cmd/rex.rst
RyanL1997: Yes, you are correct on this, and I have fixed the behavior, for example:
```bash
# Trying to match numeric value in first name field - no match
curl -X POST "http://localhost:9200/_plugins/_ppl" \
    -d '{
      "query": "source=accounts | rex field=firstname \"(?<digits>[0-9]+)\" | head 3"
    }'

{
  "schema": [
    {
      "name": "account_number",
      "type": "bigint"
    },
    {
      "name": "firstname",
      "type": "string"
    },
    {
      "name": "address",
      "type": "string"
    },
    {
      "name": "balance",
      "type": "bigint"
    },
    {
      "name": "gender",
      "type": "string"
    },
    {
      "name": "city",
      "type": "string"
    },
    {
      "name": "employer",
      "type": "string"
    },
    {
      "name": "state",
      "type": "string"
    },
    {
      "name": "age",
      "type": "bigint"
    },
    {
      "name": "email",
      "type": "string"
    },
    {
      "name": "lastname",
      "type": "string"
    },
    {
      "name": "digits",
      "type": "string"
    }
  ],
  "datarows": [
    [
      1,
      "Amber",
      "880 Holmes Lane",
      39225,
      "M",
      "Brogan",
      "Pyrami",
      "IL",
      32,
      "amberduke@pyrami.com",
      "Duke",
      null
    ],
    [
      6,
      "Hattie",
      "671 Bristol Street",
      5686,
      "M",
      "Dante",
      "Netagy",
      "TN",
      36,
      "hattiebond@netagy.com",
      "Bond",
      null
    ],
    [
      13,
      "Nanette",
      "789 Madison Street",
      32838,
      "F",
      "Nogal",
      "Quility",
      "VA",
      28,
      "nanettebates@quility.com",
      "Bates",
      null
    ]
  ],
  "total": 3,
  "size": 3
}%  
```

docs/user/ppl/cmd/rex.rst
RyanL1997: Good catch and this description is misleading. Reference to the comment: https://github.com/opensearch-project/sql/pull/4109#discussion_r2301887745, I have changed this behavior and re-wrote the description as: 
> The ``rex`` command extracts fields from a raw text field using regular expression named capture groups.

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,199 @@
...
+
+Syntax
+============
+rex [mode=<mode>] field=<field> <pattern> [max_match=<int>] [offset_field=<string>]
+
+* field: mandatory. The field must be a string field to extract data from.
+* pattern: mandatory string. The regular expression pattern with named capture groups used to extract new fields. Pattern must contain at least one named capture group using ``(?<name>pattern)`` syntax.
+* mode: optional. Either ``extract`` (default) or ``sed``. In extract mode, creates new fields from named capture groups. In sed mode, performs text substitution on the field.
+
+  **Sed mode syntax:**
```
penghuo: Please reformat the doc. https://github.com/RyanL1997/sql/blob/202a69e3338f97d99542b623744f1c4e9058df15/docs/user/ppl/cmd/rex.rst

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,199 @@
...
+    | amberduke@pyrami.com     | amberduke    | pyrami  |
+    | hattiebond@netagy.com    | hattiebond   | netagy  |
+    | nanettebates@quility.com | nanettebates | quility |
+    +--------------------------+--------------+---------+
+
+
+Example 2: Handling Non-matching Patterns
+==========================================
+
+The rex command returns all events, setting extracted fields to null for non-matching patterns.
```
penghuo: the following test results should demo the null values.

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,199 @@
...
+Limitations
+===========
+
+There are several important limitations with the rex command:
+
+**Named Capture Group Naming:**
+
+- Named capture groups cannot contain underscores due to Java regex limitations
+- Group names must start with a letter and contain only letters and digits
+- Use ``(?<username>...)`` not ``(?<user_name>...)``
```
penghuo: remove it, add as example to explain groups can't contain understore.

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,199 @@
...
+
+
+Example 3: Multiple Matches with max_match
+===========================================
+
+Extract multiple words from address field using max_match parameter.
+
+PPL query::
+
+    os> source=accounts | rex field=address "(?<words>[A-Za-z]+)" max_match=3 | fields address, words | head 3 ;
```
penghuo: address only include 2 match words, it does not demostrate how max_match=3 impact results.

docs/user/ppl/cmd/rex.rst
RyanL1997: Update: I have changed the output format to use `&` seprator. Here is an example:

```bash
    -d'{"query": "source=accounts | rex field=email \"(?<user>[^@]+)@(?<domain>.+)\" offset_field=positions | fields email, user, domain, positions | head 5"}' | jq

{
  "schema": [
    {
      "name": "email",
      "type": "string"
    },
    {
      "name": "user",
      "type": "string"
    },
    {
      "name": "domain",
      "type": "string"
    },
    {
      "name": "positions",
      "type": "string"
    }
  ],
  "datarows": [
    [
      "amberduke@pyrami.com",
      "amberduke",
      "pyrami.com",
      "domain=10-19&user=0-8"
    ],
    [
      "hattiebond@netagy.com",
      "hattiebond",
      "netagy.com",
      "domain=11-20&user=0-9"
    ],
    [
      "nanettebates@quility.com",
      "nanettebates",
      "quility.com",
      "domain=13-23&user=0-11"
    ],
    [
      "daleadams@boink.com",
      "daleadams",
      "boink.com",
      "domain=10-18&user=0-8"
    ],
    [
      "elinorratliff@scentric.com",
      "elinorratliff",
      "scentric.com",
      "domain=14-25&user=0-12"
    ]
  ],
  "total": 5,
  "size": 5
}
```

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -219,6 +221,20 @@ parseCommand
    : PARSE (source_field = expression) (pattern = stringLiteral)
    ;
 
+rexCommand
+    : REX rexExpr
```
penghuo, line 280: Add new keyword to keywordsCanBeId list

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -219,6 +221,20 @@ parseCommand
...
+
+rexExpr
+    : pattern=stringLiteral (rexOption)*
+    | FIELD EQUAL field=qualifiedName (rexOption)* pattern=stringLiteral (rexOption)*
+    ;
+
+rexOption
+    : MAX_MATCH EQUAL maxMatch=integerLiteral
+    | OFFSET_FIELD EQUAL offsetField=qualifiedName
+    | MODE EQUAL SED
```
penghuo: mode=sed and mode=extract should both works

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
```diff
@@ -511,6 +511,37 @@ public void testPatterns() {
                 + " variable_count_threshold=5"));
   }
 
+  @Test
+  public void testRexCommand() {
+    assertEquals(
+        "source=t | rex field=message \"(?<user>[A-Z]+)\"",
+        anonymize("source=t | rex field=message \"(?<user>[A-Z]+)\""));
```
penghuo, line 562: expect field=message is normalize to field=*

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -219,6 +221,20 @@ parseCommand
    : PARSE (source_field = expression) (pattern = stringLiteral)
    ;
 
+rexCommand
```
penghuo, line 279: add new keyword to keywordsCanBeId list

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLRexTest.java
```diff
@@ -0,0 +1,224 @@
...
+    super(CalciteAssert.SchemaSpec.SCOTT_WITH_TEMPORAL);
+  }
+
+  @Test
+  public void testRexBasicFieldExtraction() {
+    String ppl = "source=EMP | rex field=ENAME '(?<first>[A-Z]).*' | fields ENAME, first";
+    RelNode root = getRelNode(ppl);
+    String expectedLogical =
+        "LogicalProject(ENAME=[$1], first=[REX_EXTRACT($1, '(?<first>[A-Z]).*', 1)])\n"
+            + "  LogicalFilter(condition=[REGEXP_CONTAINS($1, '([A-Z]).*')])\n"
```
penghuo: why there is LogicalFilter?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -173,6 +176,110 @@ public RelNode visitFilter(Filter node, CalcitePlanContext context) {
...
+    }
+
+    List<RexNode> newFields = new ArrayList<>();
+    List<String> newFieldNames = new ArrayList<>();
+
+    for (int i = 0; i < namedGroups.size(); i++) {
+      RexNode extractCall;
+      if (node.getMaxMatch().isPresent() && node.getMaxMatch().get() != 1) {
+        extractCall =
+            PPLFuncImpTable.INSTANCE.resolve(
```
penghuo, line 233: Why register UDF? Calcite support REGEXP_EXTRACT already

docs/user/ppl/cmd/rex.rst
RyanL1997: yes i fixed the query

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: Thanks for calling it out, and I have tried to switch everything by using the calcite function, and the following is the current implementation:

**Fully Optimized:**

- REX_SED - Complete native Calcite optimization
   - `REGEXP_REPLACE_PG_3` for basic substitution
   - `REGEXP_REPLACE_PG_4` for flagged substitution (`g`, `i`)
   - `REGEXP_REPLACE_5` for nth occurrence
   - `TRANSLATE3` for transliteration (`y/from/to/`)

**Optimally Hybrid:**

- REX_EXTRACT - Best possible optimization given Calcite limitations
   - Native `REGEXP_EXTRACT` for single capture groups
   - Custom `REX_EXTRACT` for multiple capture groups (Calcite limitation)

**Cannot Optimize:**

- `REX_EXTRACT_MULTI` - Calcite doesn't support multiple groups
- `REX_OFFSET` - Calcite doesn't support multiple groups + different output format

Referencing to the local Calcite env + source code of Calcite library, I have also done some analysis and illustrated why we cannot replace other UDFs with some Calcite functions, please reference to: https://github.com/opensearch-project/sql/issues/4108#issuecomment-3231105679

docs/user/ppl/cmd/rex.rst
RyanL1997: fixed

docs/user/ppl/cmd/rex.rst
RyanL1997: make sense. Fixed

docs/user/ppl/cmd/rex.rst
RyanL1997: Yes, and could you review it again? 

Couple changes I made: 
- reformatted the doc
- add the description of return type into each example

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLRexTest.java
RyanL1997: fixed, this was coming from stalled test before we remove the filter operator. good catch

ppl/src/main/antlr/OpenSearchPPLParser.g4
RyanL1997: i rechecked the requirement, and yes. I have implemented it and edited the doc

ppl/src/main/antlr/OpenSearchPPLParser.g4
RyanL1997: It looks like EXTRACT is already properly included in the keywordsCanBeId lists in all the relevant PPL parser grammar files. 

`/ppl/src/main/antlr/OpenSearchPPLParser.g4)`: `EXTRACT` is included at line 1236: https://github.com/opensearch-project/sql/blob/994bdd5588ab5fa5b2bfd897ace786fc0c046d14/ppl/src/main/antlr/OpenSearchPPLParser.g4#L1236

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
RyanL1997: For this one, I was actually referencing to the other regular expression related commands:

https://github.com/opensearch-project/sql/blob/994bdd5588ab5fa5b2bfd897ace786fc0c046d14/ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java#L470-L488

So it seems like for both parse and grok, they did not anonymize the field. So should we do that? Or keep aligning with them?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -174,6 +176,79 @@ public RelNode visitFilter(Filter node, CalcitePlanContext context) {
     return context.relBuilder.peek();
   }
 
+  @Override
+  public RelNode visitRex(Rex node, CalcitePlanContext context) {
+    visitChildren(node, context);
+
+    RexNode fieldRex = rexVisitor.analyze(node.getField(), context);
```
penghuo: node.getField() could return null valu? If Rex does not support default field, disable it in ANTRL.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -174,6 +176,79 @@ public RelNode visitFilter(Filter node, CalcitePlanContext context) {
...
+        extractCall =
+            PPLFuncImpTable.INSTANCE.resolve(
+                context.rexBuilder,
+                BuiltinFunctionName.REX_EXTRACT_MULTI,
+                fieldRex,
+                context.rexBuilder.makeLiteral(patternStr),
+                context.relBuilder.literal(i + 1),
+                context.relBuilder.literal(node.getMaxMatch().get()));
+      } else {
+        if (namedGroups.size() == 1) {
```
penghuo: If REX_EXTRACT can support INTERNAL_REGEXP_EXTRACT, if..else can be simplified.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -174,6 +176,79 @@ public RelNode visitFilter(Filter node, CalcitePlanContext context) {
...
+      throw new IllegalArgumentException(
+          "Rex pattern must contain at least one named capture group");
+    }
+
+    List<RexNode> newFields = new ArrayList<>();
+    List<String> newFieldNames = new ArrayList<>();
+
+    for (int i = 0; i < namedGroups.size(); i++) {
+      RexNode extractCall;
+      if (node.getMaxMatch().isPresent() && node.getMaxMatch().get() != 1) {
```
penghuo: in multiple group case, instead of call regex match multiple times, we can simpleiy to 1 time. e.g. regex_matches can return array[] include multiple matched results.
create a issue to track perf improvement

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractMultiFunction.java
```diff
@@ -0,0 +1,90 @@
...
+          RexExtractMultiFunction.class,
+          "extractMultipleGroups",
+          field,
+          pattern,
+          groupIndex,
+          maxMatch);
+    }
+  }
+
+  public static List<String> extractMultipleGroups(
```
penghuo, line 66: Add UT

core/src/main/java/org/opensearch/sql/expression/function/udf/RexOffsetFunction.java
```diff
@@ -0,0 +1,92 @@
...
+    public Expression implement(
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+      Expression field = translatedOperands.get(0);
+      Expression pattern = translatedOperands.get(1);
+
+      return Expressions.call(RexOffsetFunction.class, "calculateOffsets", field, pattern);
+    }
+  }
+
+  public static String calculateOffsets(String text, String patternStr) {
```
penghuo: add UT

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
songkant-aws: +1

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractFunction.java
```diff
@@ -0,0 +1,71 @@
...
+      Pattern compiledPattern = Pattern.compile(pattern);
+      Matcher matcher = compiledPattern.matcher(text);
+
+      if (matcher.find() && groupIndex > 0 && groupIndex <= matcher.groupCount()) {
+        return matcher.group(groupIndex);
+      }
+
+      return null;
+    } catch (Exception e) {
+      return null;
```
songkant-aws, line 62: Is it a legit behavior to fail open? Not rethrow exception to early stop the execution?

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractFunction.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  public static String extractGroup(String text, String pattern, int groupIndex) {
+    if (text == null || pattern == null) {
+      return null;
+    }
+
+    try {
+      Pattern compiledPattern = Pattern.compile(pattern);
```
songkant-aws, line 56: [Low priority] I'm thinking a nice to have feature is to cache compiled pattern. Alternatively, define it as a global Linq4j expression. See Calcite example: https://github.com/apache/calcite/blob/44b57985eaeb0ef0c1eda2447aa75b5855259356/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java#L461-L475

Maybe we can create an issue to track perf improvement if it's not feasible

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractFunction.java
```diff
@@ -0,0 +1,71 @@
...
+      Expression field = translatedOperands.get(0);
+      Expression pattern = translatedOperands.get(1);
+      Expression groupIndex = translatedOperands.get(2);
+
+      return Expressions.call(RexExtractFunction.class, "extractGroup", field, pattern, groupIndex);
+    }
+  }
+
+  public static String extractGroup(String text, String pattern, int groupIndex) {
+    if (text == null || pattern == null) {
```
songkant-aws: How about the blank text or pattern?

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractMultiFunction.java
```diff
@@ -0,0 +1,90 @@
...
+
+  public RexExtractMultiFunction() {
+    super(new RexExtractMultiImplementor(), NullPolicy.ARG0);
+  }
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return call ->
+        call.getTypeFactory()
+            .createArrayType(call.getTypeFactory().createSqlType(SqlTypeName.VARCHAR, 2000), -1);
```
songkant-aws: Since you allow returning null value, add UT to test if this returnType works for null as well. I remember it needs to call out nullable for array type as well

ppl/src/main/antlr/OpenSearchPPLParser.g4
songkant-aws: It seems we allow the repeated instances of rexOption here. For example, `search ... | rex field=msg mode=sed mode=extract '(?<first>[A-Z]).*' maxMatch=1 maxMatch=5`

We may add validation here and throw exception before building RelNode.

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
```diff
@@ -419,6 +420,32 @@ public String visitReverse(Reverse node, String context) {
...
+    String field = visitExpression(node.getField());
+    String pattern = "\"" + node.getPattern().toString() + "\"";
+    StringBuilder command = new StringBuilder();
+
+    // Build the base command
+    if (node.getMode() == Rex.RexMode.SED) {
+      command.append(String.format("%s | rex field=%s mode=sed %s", child, field, pattern));
+    } else {
+      command.append(String.format("%s | rex field=%s %s", child, field, pattern));
+    }
```
songkant-aws: Can simplify this logic to print explicit mod?

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractMultiFunction.java
RyanL1997: added

core/src/main/java/org/opensearch/sql/expression/function/udf/RexOffsetFunction.java
RyanL1997: added

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: good catch, fixed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: Yes and I have removed it. Rechecked the doc, we have also mentioned it as mandatory for now.

RyanL1997:
Btw, for better readability, Im currently working on splitting this PR into separate parts:
- Part 1: Extract (single pattern / multi pattern) + max_match
- Part 2: Sed Mode + Offset Field

Turning it into draft just for now.

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
RyanL1997: good point, fixed.

docs/user/ppl/cmd/rex.rst
```diff
@@ -0,0 +1,171 @@
+=============
+rex
```
penghuo, line 2: Add to calcite-enable test. https://github.com/opensearch-project/sql/blob/main/docs/category.json#L59

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractFunction.java
RyanL1997: I actually realized that we don't need this logic - Query planner already validates null inputs via `NullPolicy.ARG0`
```bash
curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d'{
    "query": "source=accounts | rex field=nonexistent \"(?<test>.+)\" | fields nonexistent, test | head 1"
  }' | jq

{
  "error": {
    "reason": "Invalid Query",
    "details": "field [nonexistent] not found; input fields are: [account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname, _id, _index, _score, _maxscore, _sort, _routing]",
  },
  "status": 400
}
```

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractFunction.java
RyanL1997: yes, this is the same behavior of the requirment

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractFunction.java
```diff
@@ -0,0 +1,71 @@
...
+    try {
+      Pattern compiledPattern = Pattern.compile(pattern);
+      Matcher matcher = compiledPattern.matcher(text);
+
+      if (matcher.find() && groupIndex > 0 && groupIndex <= matcher.groupCount()) {
+        return matcher.group(groupIndex);
+      }
+      return null;
+    } catch (PatternSyntaxException e) {
+      throw new IllegalArgumentException(
```
RyanL1997, line 64: Hi @penghuo, I have rechecked the tequrement - and yes for pattern failure it should be catched. The cuurent behavior is like this: 
```bash
 curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d'{
    "query": "source=accounts | rex field=email \"(?<invalid>[\" | fields email, invalid | head 1"
  }' | jq

{
  "error": {
    "reason": "Invalid Query",
    "details": "Error in 'rex' command: Encountered the following error while compiling the regex '(?<invalid>[': Unclosed character class near index 11\n(?<invalid>[\n           ^",
  },
  "status": 400
}
```

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractFunction.java
penghuo: +1, Add issue to track it. per row pattern compile is expensive.

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractMultiFunction.java
RyanL1997: Yes, I have noticed the same - the return type needs to properly handle nullability for both the array itself and its elements.

 Changes Made:

  1. Fixed Return Type Inference: Updated getReturnTypeInference() to explicitly create a nullable array type:
  ```java
  return call -> {
    var elementType = call.getTypeFactory().createSqlType(SqlTypeName.VARCHAR, 2000);
    return call.getTypeFactory().createArrayType(
        call.getTypeFactory().createTypeWithNullability(elementType, true), -1);
  };
```
  2. This ensures the array type is properly marked as nullable in Calcite's type system.
  3. Added Unit Tests: Created testNullReturnTypeCompatibility() to verify the function correctly returns null for:
    - No regex matches found
    - Invalid group indices
    - Empty input strings

**Comment by ahkcs** [comment]
This branch currently has merge conflicts

File: ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java [hunk 39]
```diff
@@ -966,6 +967,49 @@ public UnresolvedPlan visitRegexCommand(OpenSearchPPLParser.RegexCommandContext
...
+                        ? " or adjust the plugins.ppl.rex.max_match.limit setting."
+                        : "."),
+                maxMatch.get(),
+                maxMatchLimit));
+      } else {
+        effectiveMaxMatch = maxMatch.get();
+      }
+    } else {
+      effectiveMaxMatch = 1;
+    }
```
penghuo, line 1023: nit: consider simpley if..else..
int effectiveMaxMatch = maxMatch.else(1)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -204,6 +206,52 @@ public RelNode visitRegex(Regex node, CalcitePlanContext context) {
...
+        // Use REX_EXTRACT_MULTI for multiple matches
+        extractCall =
+            PPLFuncImpTable.INSTANCE.resolve(
+                context.rexBuilder,
+                BuiltinFunctionName.REX_EXTRACT_MULTI,
```
penghuo, line 235: REX_EXTRACT_MULTI already cover max_match=1? we can remove else... REX_EXTRACT? Any concern?

docs/user/ppl/cmd/rex.rst
RyanL1997: Yes I think  I did: https://github.com/opensearch-project/sql/pull/4109/files#diff-8b39b5d54adc60c65d50023a1f1989881fe5ae44778a73a4c8f750f35a770dd2R62

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: good point. However, I have re-checked the requirement, and I think the correct behavior should be: 
- when `max_match = 1` (specified as 1 or default), the return type should be a single string ->  handle by `REX_EXTRACT` only
- when `max_match > 1`, the return type should be an array -> handle by `REX_EXTRACT_MULTI`

So, I think we can keep the same, and I have modified the condition to make sure multi extraction should not be able to handle single extraction case.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
RyanL1997: fixed.

penghuo (approved):

core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractFunction.java
RyanL1997: follow up issue: https://github.com/opensearch-project/sql/issues/4235

PR #4122, opened 2025-08-24
[Backport 2.19-dev] Support distinct_count/dc in eventstats

ahkcs:
### Description
Support distinct_count/dc in eventstats #4084

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.sql.calcite.remote;
 
+import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_BANK;
 import static org.opensearch.sql.util.MatcherUtils.assertJsonEqualsIgnoreId;
 
 import java.io.IOException;
@@ -214,6 +215,40 @@ public class CalciteExplainIT extends ExplainIT {
                 + " address_length = length(address) | stats count() by address_length");
     String expected = loadFromFile("expectedOutput/calcite/explain_script_push_on_text.json");
     assertJsonEqualsIgnoreId(expected, result);
+  }
+
+  @Test
+  public void testEventstatsDistinctCountExplain() throws IOException {
+    Assume.assumeTrue("This test is only for push down enabled", isPushdownEnabled());
+    String query =
+        "source=opensearch-sql_test_index_account | eventstats dc(state) as distinct_states";
+    var result = explainQueryToString(query);
+    String expected = loadFromFile("expectedOutput/calcite/explain_eventstats_dc.json");
+    assertJsonEqualsIgnoreId(expected, result);
+  }
+
+  @Test
+  public void testEventstatsDistinctCountFunctionExplain() throws IOException {
+    Assume.assumeTrue("This test is only for push down enabled", isPushdownEnabled());
+    String query =
+        "source=opensearch-sql_test_index_account | eventstats distinct_count(state) as"
+            + " distinct_states by gender";
+    var result = explainQueryToString(query);
+    String expected = loadFromFile("expectedOutput/calcite/explain_eventstats_distinct_count.json");
+    assertJsonEqualsIgnoreId(expected, result);
+  }
+
+  // Only for Calcite, as v2 gets unstable serialized string for function
+  @Test
+  public void testExplainOnAggregationWithSumEnhancement() throws IOException {
+    String expected = loadExpectedPlan("explain_agg_with_sum_enhancement.json");
+    assertJsonEqualsIgnoreId(
+        expected,
+        explainQueryToString(
+            String.format(
+                "source=%s | stats sum(balance), sum(balance + 100), sum(balance - 100),"
+                    + " sum(balance * 100), sum(balance / 100) by gender",
+                TEST_INDEX_BANK)));
   }
 
   /**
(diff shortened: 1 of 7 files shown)


integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -238,19 +237,6 @@ public void testEventstatsDistinctCountFunctionExplain() throws IOException {
...
-  public void testExplainOnAggregationWithSumEnhancement() throws IOException {
-    String expected = loadExpectedPlan("explain_agg_with_sum_enhancement.json");
-    assertJsonEqualsIgnoreId(
-        expected,
-        explainQueryToString(
-            String.format(
-                "source=%s | stats sum(balance), sum(balance + 100), sum(balance - 100),"
-                    + " sum(balance * 100), sum(balance / 100) by gender",
-                TEST_INDEX_BANK)));
-  }
```
LantaoJin: why this IT was deleted? do we miss any backporting?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLEventstatsIT.java
```diff
@@ -598,7 +598,7 @@ public void testEventstatDistinctCount() throws IOException {
     JSONObject actual =
         executeQuery(
             String.format(
-                "source=%s | eventstats dc(state) as dc_state", TEST_INDEX_STATE_COUNTRY));
+                "source=%s | eventstats dc(state) as dc_state | fields name, country, state, month, year, age, dc_state", TEST_INDEX_STATE_COUNTRY));
```
LantaoJin, line 601: why `fields` required in these ITs.

ahkcs: It will cause the CI to fail, local testing was successful without `fields`

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
ahkcs: This IT was deleted because it was not originally in 2.19-dev branch. And this IT was not created by #4084 so I dropped it in this backport

LantaoJin: Oh, I see.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLEventstatsIT.java
LantaoJin: > It will cause the CI to fail, local testing was successful without `fields`

If the original tests were successful without this PR, sounds a potential bug is introducing in this PR.

ahkcs: The CI failure is this PR was caused by the order of the schema, and this failure also is not able to be reproduced locally. I suspect that it's triggered by different testing environment in CI and local. Do you think we should also add this `fields` to main for consistency in this case?

LantaoJin: > and this failure also is not able to be reproduced locally. I suspect that it's triggered by different testing environment in CI and local. 

The question is how the PR introduced a reproduced CI failure. we need to figure out the reason.

ahkcs:
> In addition to Java-version-specific fixes in backporting, do not simply modify original tests during backporting. At minimum, provide an explanation for why you're bypassing the issue rather than resolving it fundamentally.

The CI failure was reproduced locally by switching to Java 11. Since different Java version will cause the order of the schema to be different, should we add the `fields` on both main and 2.19-dev backport to ensure output consistency for different Java versions?

Swiddis:
It's not just this PR, I pulled https://github.com/opensearch-project/sql/commit/f10ffd50a08f0b167b4d684798456bf7d0b419d9 locally (base for this branch) and it looks like others of the NoPushdown tests are failing with shuffled results on temurin-11. e.g. `Fillnull`.

LantaoJin:
> > In addition to Java-version-specific fixes in backporting, do not simply modify original tests during backporting. At minimum, provide an explanation for why you're bypassing the issue rather than resolving it fundamentally.
> 
> The CI failure was reproduced locally by switching to Java 11. Since different Java version will cause the order of the schema to be different, should we add the `fields` on both main and 2.19-dev backport to ensure output consistency for different Java versions?

Do you mean when move to Java 11, even without your PR, the CI will fail either. If yes, we can add `fields`. But if this failures only happened with your PR, we still need to investigate why the PR changed something.

ahkcs:
> Do you mean when move to Java 11, even without your PR, the CI will fail either. If yes, we can add `fields`. But if this failures only happened with your PR, we still need to investigate why the PR changed something.

The CI failure is specifically for my `distinct_count` test cases, no other tests failed

For testing in main, the project requires Java 21 minimum. In 2.19-dev branch, I have confirmed that CI failure can be reproduced locally with Java 11. When using Java 21, the integration test will pass locally.

By `local testing` I mean the local branch for this PR

ykmr1224:
@ahkcs @LantaoJin 
I think we can put `@Ignore` to the failing test cases for this specific PR, but we should fix it to return result with consistent order regardless of the environment.

ykmr1224 (approved):

LantaoJin:
FYI. This PR is intended to resolve the issue of result ordering inconsistencies across different Java versions. https://github.com/opensearch-project/sql/pull/3740

PR #4123, opened 2025-08-25
Implement `Append` command with Calcite

### LINKED ISSUE #4078: [RFC] Support `Append` command in PPL
**Problem Statement**
Implement `append` command to simulate SQL standard union all operation. The `append` command syntax is quite similar to the `appendcols` command, which is already implemented during Calcite V3 engine development. Although their syntax are similar, their final combined result formats are quite different. The key difference is that the `append` command functions as a special union operator in SQL, while the `appendcols` operates as a special join operator in SQL.

**Syntax**

`append <subsearch>`

**Core Functionality**
OpenSearch doesn’t have similar functionality as append command. So we need to prioritize the new command development.

The append command appends the results of subsearch to the current results vertically. Say, the current search outputs 1000 rows with fields schema [a, b] and the subsearch outputs another 1000 rows with different fields schema [a, c, d], the final output will contain 2000 rows with the combined fields of two searches, aka [a, b, c, d].

The following table illustrates the concept. The green colored results are the main query. The yellow colored results are the subquery output. When combining results, the command fill any missing fields with NULL values.

Image

**Implementation**
OpenSearch PPL has already supported subsearch syntax starting with search command, follows a serious of other commands. To support the core functionality, we just need to implement union operation over two result rows for the main query and the subquery. Standard SQL union operator (like that in Calcite) only supports union two datasets with the same field schema. But we can achieve the functionality by a simple workaround approach. The problem can be translated to a similar SQL query like:

```
select a, b, NULL as c, NULL as d from main_query
union all
select a, NULL as b, c, d from sub_query
```

**Proposed Solution**

* Analyze current query in Calcite RelNode visitor. Get its rowType (output fields schema).
* Analyze subquery in Calcite RelNode visitor. Get its rowType.
* Merge two schemas and add projects over two queries separately. The missing fields in each project will be filled with constant NULL values
* Perform union all operation over two projected RelNodes.
* Support different indices query in subsearch

**Long-Term Goals**
* Support subsearch timeout to avoid time consuming query
* Allow users to set maximum output rows
* Allow users to inherit timerange columns from upstream.

**Pushdown Support**

Don’t support it. It’s an operation that is suitable for execution on coordinator node. We should let the pushdown work in main query and subquery internally.

**Limitation**

Standard SQL union operator requires the exactly the same field schema(including the field types). If both the main query and the subquery has a same name field but with different types, the standard union could error out. A workaround could be using different field alias for conflicting field.

songkant-aws:
### Description
Implement `Append` command in V3 Calcite engine. This `Append` command is quite similar to standard SQL union all operator.

### Related Issues
Resolves #4078 

### Check List


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1159,6 +1161,76 @@ public RelNode visitAppendCol(AppendCol node, CalcitePlanContext context) {
...
   }
 
+  @Override
+  public RelNode visitAppend(Append node, CalcitePlanContext context) {
+    // 1. Resolve main plan
+    visitChildren(node, context);
+
+    // 2. Build subsearch tree (attach relation to subsearch)
+    UnresolvedPlan relation = getRelation(node);
+    transformPlanToAttachChild(node.getSubSearch(), relation);
```
yuancu: Are we going to support searching another index in the subsearch, instead of the same index as the main query? In the current implementation, this seems to be unsupported.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1159,6 +1161,76 @@ public RelNode visitAppendCol(AppendCol node, CalcitePlanContext context) {
     }
   }
 
+  @Override
+  public RelNode visitAppend(Append node, CalcitePlanContext context) {
```
ykmr1224, line 1222: Let's organize it better by extracting methods.

ykmr1224, line 1222: Can we share logic with `appendcol`?

songkant-aws: No, I think they're different logic. Can't share it with `appendcol`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
songkant-aws: Added that support in latest commit, though we haven't reached the agreement in RFC. 

I'm inclined to support this syntax anyway.

ppl/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -41,7 +41,7 @@ ML:                                 'ML';
 FILLNULL:                           'FILLNULL';
 FLATTEN:                            'FLATTEN';
 TRENDLINE:                          'TRENDLINE';
-APPENDCOL:                          'APPENDCOL';
+APPENDCOLS:                         'APPENDCOLS';
```
LantaoJin: It's a breaking change. cc @penghuo @brijos

core/src/main/java/org/opensearch/sql/ast/tree/Append.java
```diff
@@ -0,0 +1,47 @@
...
+@ToString
+@EqualsAndHashCode(callSuper = false)
+@RequiredArgsConstructor
+public class Append extends UnresolvedPlan {
+
+  private final UnresolvedPlan searchPlan;
+
+  private final UnresolvedPlan subSearch;
+
+  private UnresolvedPlan child;
```
LantaoJin, line 28: what's the different between `searchPlan` and `child`?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: > Let's organize it better by extracting methods.

Extracting methods and Util classes makes the logic more hard to understand, unless the logic could be reused.

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -391,4 +392,9 @@ public Void visitInputRef(RexInputRef inputRef) {
     visitor.visitEach(rexNodes);
     return selectedColumns;
   }
+
+  static boolean isEmptyValuesPlan(UnresolvedPlan plan) {
```
LantaoJin: this `PlanUtils` is for Calcite RelNode actually.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1176,6 +1178,77 @@ public RelNode visitAppendCol(AppendCol node, CalcitePlanContext context) {
...
+    // 1. Resolve main plan
+    visitChildren(node, context);
+
+    // 2. Build subsearch tree
+    if (PlanUtils.isEmptyValuesPlan(node.getSearchPlan())) {
+      node.getSearchPlan().accept(this, context);
```
LantaoJin: can you explain what is `empty values`? I didn't see any IT for this.
IMO, the searchPlan is child of the `node` and you have already visited its children in step 1.

docs/user/ppl/cmd/append.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. contents::
+   :local:
+   :depth: 2
+
+
+Description
+============
+| (Experimental)
+| (From 3.2.0)
```
LantaoJin: remove

docs/user/ppl/cmd/append.rst
```diff
@@ -0,0 +1,131 @@
...
+| Using ``append`` command to append the result of a sub-search and attach it as additional rows to the bottom of the input search results (The main search).
+The command aligns columns with the same field names and types. For different column fields between the main search and sub-search, NULL values are filled in the respective rows.
+
+Version
+=======
+3.2.0
```
LantaoJin: 3.3.0

docs/user/ppl/cmd/append.rst
```diff
@@ -0,0 +1,131 @@
...
+      "acknowledged": true,
+      "persistent": {
+        "plugins": {
+          "calcite": {
+            "enabled": "true"
+          }
+        }
+      },
+      "transient": {}
+    }
```
LantaoJin: I think we can remove this since we plan to enable calcite in 3.3.0

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -68,7 +68,8 @@ commands
    | mlCommand
    | fillnullCommand
    | trendlineCommand
-   | appendcolCommand
+   | appendcolsCommand
+   | appendCommand
```
LantaoJin: seems they are all missing in `commandName`

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAppendTest.java
```diff
@@ -0,0 +1,143 @@
...
+            + "SELECT *\n"
+            + "FROM `scott`.`EMP`\n"
+            + "WHERE `DEPTNO` = 20";
+    verifyPPLToSparkSQL(root, expectedSparkSql);
+  }
+
+  @Test
+  public void testAppendEmptySearchCommand() {
+    List<String> testPPLs =
+        Arrays.asList("source=EMP | append [ | where DEPTNO = 20 ]", "source=EMP | append [ ]");
```
LantaoJin: Supporting `append [ ]` is on purpose?

LantaoJin (commented):
Can you add a test in ExplainIT and CrossClusterSearchIT?

ppl/src/main/antlr/OpenSearchPPLLexer.g4
LantaoJin: Maybe you could separate to two PRs, since one of it should be marked to `breaking`.

songkant-aws: That's fair. I can make `appendcols` another PR.

core/src/main/java/org/opensearch/sql/ast/tree/Append.java
songkant-aws: The `child` is the main query. However, the `searchPlan` is the first optional searchCommand node starting in subsearch.

This is for matching optional `searchCommand` in the head of subsearch for `append` or `appendcols` command. 

For example, in future, this query is a valid query: `search | append [ | inputlookup myexcel | fields myfield ]`
In above query case, the searchCommand between square brackets is optional because `inputlookup` command is a command similar to TableFunction. Users have freedom to choose whether data comes from search or local input. The searchPlan is used for parsing empty searchCommand as a 0 row * 0 col `LogicalValues[[]]` RelNode.

Another reason is to use searchPlan to handle an edge case of appending empty subresults. Some other pipeline language has similar functionality to allow subsearch outputs empty result. For example, `search | append [ ]` and `search | append [ | fields a, b, c ]` are the same because subsearch start with 0 row * 0 col input. The syntax is legit to append empty result. It is equivalent to main query.

Calcite is a strong schema engine, parsing a RelNode like 
```
Project(a = [$0], b = [$1], ..)
    LogicalValues[[]]
```
will throw exception when either building RelNode or at runtime. Because Project cannot find any 'a' column from the input.

I'm open to this discussion, we have two options here:
1. Since we're using strong schema engine, we should just throw exception. This option is simple but may have inconsistent behavior with other pipeline language.
2. We add logics to use empty values like `LogicalValues[[]]` to handle edge cases. A better way is probably to  determine whether subsearch RelNode's leaf is empty values. But still it has the problem of building RelNode successfully.

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
songkant-aws: It's used in handling edge case of determining RelNode in subsearch as described in https://github.com/opensearch-project/sql/pull/4123#discussion_r2308997609. But I'm thinking it can be moved to AstBuilder as well.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
songkant-aws: See my explanation in: https://github.com/opensearch-project/sql/pull/4123#discussion_r2308997609

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAppendTest.java
songkant-aws: Yes. It's on purpose to test the edge case of appending 0 row * 0 col subresult. See my explanation in: https://github.com/opensearch-project/sql/pull/4123#discussion_r2308997609

ppl/src/main/antlr/OpenSearchPPLParser.g4
songkant-aws: done

docs/user/ppl/cmd/append.rst
songkant-aws: Removed

docs/user/ppl/cmd/append.rst
songkant-aws: Done

docs/user/ppl/cmd/append.rst
songkant-aws: Removed

core/src/main/java/org/opensearch/sql/ast/tree/Append.java
songkant-aws: I refactor the code a bit to avoid confusing name of `searchPlan`

songkant-aws: I choose to throw exception for empty subsearch input for now because it needs some dirty work to preprocess ast tree to achieve a not useful use case. Not supporting it by throwing exception is elegant.

LantaoJin: > I choose to throw exception for empty subsearch input for now because it needs some dirty work to preprocess ast tree to achieve a not useful use case. Not supporting it by throwing exception is elegant.

@songkant-aws I think the the previous implementation looks good to me except the confusing name of `searchPlan`, maybe we need it back with another name such as `emptySource`

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAppendTest.java
LantaoJin: Get it, please support the corner case `append [ ]` and `append [ | fields a,b,c ]`

core/src/main/java/org/opensearch/sql/ast/tree/Append.java
songkant-aws: I changed back the empty source support, and for nested case.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAppendTest.java
songkant-aws: Done

core/src/main/java/org/opensearch/sql/ast/EmptySourcePropagateVisitor.java
```diff
@@ -0,0 +1,121 @@
...
+  @Override
+  public UnresolvedPlan visitAppendCol(AppendCol node, Void context) {
+    UnresolvedPlan subSearch = node.getSubSearch().accept(this, context);
+    UnresolvedPlan child = node.getChild().get(0).accept(this, context);
+    return new AppendCol(node.isOverride(), subSearch).attach(child);
+  }
+
+  // TODO: Revisit lookup logic here but for now we don't see use case yet
+  @Override
+  public UnresolvedPlan visitLookup(Lookup node, Void context) {
```
LantaoJin, line 70: None of tests/ITs cover the `visitLookup` and `visitJoin` you added here. Please add tests/ITs for empty source for `lookup`/`join`

docs/category.json
```diff
@@ -57,6 +57,8 @@
   ],
   "ppl_cli_calcite": [
     "user/ppl/cmd/append.rst",
+    "user/ppl/cmd/stats.rst",
```
LantaoJin: duplicated to L63

songkant-aws: Removed duplicate line

core/src/main/java/org/opensearch/sql/ast/EmptySourcePropagateVisitor.java
songkant-aws: Added tests. For now, our join or lookup command syntax enforces the searchCommand to be existent in the right child. So I can't cover the case of empty source of right child. Empty source for left child test cases are added.

LantaoJin:
@songkant-aws please fix the conflicts. @qianheng-aws @yuancu do one of you have a chance to take another review after conflicts resolved?

PR #4127, opened 2025-08-25
Adding big5 `coalesce` queries

aalva500-prog:
### Description
Adding big5 queries for the new `coalesce` function enhancements.

### Related Issues
https://github.com/opensearch-project/sql/issues/4005

### Check List

### DIFF
--- a/integ-test/src/test/resources/big5/queries/coalesce_empty_string_priority.ppl
+++ b/integ-test/src/test/resources/big5/queries/coalesce_empty_string_priority.ppl
@@ -1,0 +1,3 @@
+source = big5
+| eval result = coalesce('', `host.name`)
+| head 10
--- a/integ-test/src/test/resources/big5/queries/coalesce_numeric_fallback.ppl
+++ b/integ-test/src/test/resources/big5/queries/coalesce_numeric_fallback.ppl
@@ -1,0 +1,3 @@
+source = big5
+| eval result = coalesce(`host.name`, 123, 'fallback')
+| head 10
--- a/integ-test/src/test/resources/big5/queries/coalesce_nonexistent_field_fallback.ppl
+++ b/integ-test/src/test/resources/big5/queries/coalesce_nonexistent_field_fallback.ppl
@@ -1,0 +1,3 @@
+source = big5
+| eval result = coalesce(`host.name`, dummy_field, 'unknown')
+| head 10


integ-test/src/test/resources/big5/queries/coalesce_empty_string_priority.ppl
```diff
@@ -0,0 +1,3 @@
+source = big5
```
LantaoJin: What's purpose to add the ppl files? Please update the `PPLBig5IT`.

aalva500-prog: The purpose of adding the ppl files is to have the source query for new features/enhancements in commands/functions that include benchmark. We are testing performance in big5 data set whenever we improve an existing command/function or create a new one. Hope that makes sense.

aalva500-prog: Hi @LantaoJin, I tried to add the queries to the `PPLBig5IT` test suite as recommended. During the integration test execution, I have identified two issues that require clarification:

**Issue 1: Calcite Configuration in PPLBig5IT**
The `PPLBig5IT` class disables Calcite by default in its initialization method. If the new command or enhancements are designed to function exclusively with Calcite enabled, the tests will fail in this configuration.

```java
@Override
    super.init();
    loadIndex(Index.BIG5);
    disableCalcite();
}
```

**Issue 2: Missing Field in CalcitePPLBig5IT**
When executing tests through CalcitePPLBig5IT, I encounter a field resolution error for `host.name`. The error indicates this field is not present in the current dataset, though it appears to be part of the Big5 dataset mappings used in our EC2 performance testing environment.

```java
org.opensearch.client.ResponseException: method [POST], host [http://[::1]:61814], URI [/_plugins/_ppl], status line [HTTP/1.1 400 Bad Request]
{
  "error": {
    "reason": "Invalid Query",
    "details": "field [host.name] not found; input fields are: [agent, agent.ephemeral_id, agent.id, agent.name, agent.type, agent.version, process, process.name, log, log.file, log.file.path, message, tags, cloud, cloud.region, input, input.type, @timestamp, ecs, ecs.version, data_stream, data_stream.dataset, data_stream.namespace, data_stream.type, meta, meta.file, host, metrics, metrics.size, metrics.tmin, aws, aws.cloudwatch, aws.cloudwatch.ingestion_time, aws.cloudwatch.log_group, aws.cloudwatch.log_stream, event, event.dataset, event.id, event.ingested, _id, _index, _score, _maxscore, _sort, _routing]",
  },
  "status": 400
}
```

Could you please confirm whether the `host.name` field should be included in the `Big5` dataset used for integration testing in this project? This will help determine if we need to update the dataset or modify the test queries accordingly.

Thank you for your guidance on resolving these issues.

integ-test/src/test/java/org/opensearch/sql/calcite/big5/PPLBig5IT.java
```diff
@@ -17,7 +17,7 @@
 
 @FixMethodOrder(MethodSorters.JVM)
 public class PPLBig5IT extends PPLIntegTestCase {
```
ykmr1224, line 19: It can be a separate task, but we might want to scan the files under `big5/queries` and automatically execute all.

ykmr1224 (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/big5/PPLBig5IT.java
aalva500-prog: Yeah, that's a great idea.

integ-test/src/test/resources/big5/queries/coalesce_empty_string_priority.ppl
```diff
@@ -0,0 +1,3 @@
...
+| eval result = coalesce('', `metrics.size`)
```
noCharger: This appears to be an edge case query. Any implementation nuances you'd like to benchmark?

integ-test/src/test/resources/big5/queries/coalesce_numeric_fallback.ppl
```diff
@@ -0,0 +1,3 @@
+source = big5
+| eval result = coalesce(`metrics.size`, 123, 'fallback')
```
noCharger: Does field type matter for `coalesce`?

noCharger (commented):
Can we add field with null value to just demonstrate functionality?

integ-test/src/test/resources/big5/queries/coalesce_numeric_fallback.ppl
aalva500-prog: @noCharger No, we have removed strict same-type requirement for all arguments. The new implementation of the `coalesce` function performs automatic type coercion for mixed data types.

LantaoJin (approved):

integ-test/src/test/resources/big5/queries/coalesce_empty_string_priority.ppl
aalva500-prog: @noCharger Yes, you are right, this is an edge case. I just wanted to test the new functionality with `big5` data set, as the new `coalesce` function supports empty strings ("") and whitespace strings (" ") as valid values. For example, empty strings are treated as valid non-null values, not skipped. So, this query will return an empty string always. Personally, I believe this functionality will be barely used, so maybe I can remove this query.

aalva500-prog:
@noCharger To correctly demonstrate `null` values handling, we need to verify the data in the whole `big5` data set to see if there are fields with a `null` value assigned. However, with the new implementation, we can demonstrate `null` values handling with `non-existing` fields also, as `non-existing` fields are treated as `null`. 

For example, the following query will always skip `dummy_field` field. Then, it will continue to the `host.name` field and check if it contains a `null` value. If it does contain a `null` value, it will then check if `metrics.size` is `non-null`. Finally, if both `host.name` and `metrics.size` are null, it will fallback to the `'unknown'` string literal. 

```
source = big5
| eval result = coalesce(dummy_field, `host.name`, `metrics.size`, 'unknown')
| head 10
```

Here we are demonstrating mixed data type handling in the same `coalesce` function, along with `non-existing` fields treated as `null` and string literal fallback. I think for the purpose of this exercise, we can leave this query only. Please let me know what you think, thanks!

noCharger:
> @noCharger To correctly demonstrate `null` values handling, we need to verify the data in the whole `big5` data set to see if there are fields with a `null` value assigned. However, with the new implementation, we can demonstrate `null` values handling with `non-existing` fields also, as `non-existing` fields are treated as `null`.
> 
> For example, the following query will always skip `dummy_field` field. Then, it will continue to the `host.name` field and check if it contains a `null` value. If it does contain a `null` value, it will then check if `metrics.size` is `non-null`. Finally, if both `host.name` and `metrics.size` are null, it will fallback to the `'unknown'` string literal.
> 
> ```
> source = big5
> | eval result = coalesce(dummy_field, `host.name`, `metrics.size`, 'unknown')
> | head 10
> ```
> 
> Here we are demonstrating mixed data type handling in the same `coalesce` function, along with `non-existing` fields treated as `null` and string literal fallback. I think for the purpose of this exercise, we can leave this query only. Please let me know what you think, thanks!

Sounds good, do we still need all cases or just one query demonstrate all?

aalva500-prog:
@noCharger IMO, one query should be enough. I'll update the PR and let you know.

noCharger (approved):

PR #4138, opened 2025-08-27
Support refering to implicit `@timestamp` field in span

### LINKED ISSUE #4136: [FEATURE] Support implicit `@timestamp` field access in span

Currently, we have to specify a field to be used by the span function.  E.g.  `source=my-index | stats count() by span(@timestamp, 1h)`

For the ease of use, we'd like to make `@timestamp` a default field when users do not explicitly specify a field. I.e. we are going to support syntax like `source=my-index | stats count() by span(1h)`, which will implicitly leverage `@timestamp` field. 

--

--

yuancu:
### Description

This PR allows makes the query `source=my-index | stats count() by span(1h)` equivalent to `source=my-index | stats count() by span(@timestamp, 1h)`

### Related Issues
Resolves #4136 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -46,6 +46,7 @@ import org.opensearch.sql.ast.expression.Case;
 import org.opensearch.sql.ast.expression.Cast;
 import org.opensearch.sql.ast.expression.Compare;
 import org.opensearch.sql.ast.expression.EqualTo;
+import org.opensearch.sql.ast.expression.Field;
 import org.opensearch.sql.ast.expression.Function;
 import org.opensearch.sql.ast.expression.In;
 import org.opensearch.sql.ast.expression.Interval;
@@ -67,6 +68,7 @@ import org.opensearch.sql.ast.expression.subquery.ExistsSubquery;
 import org.opensearch.sql.ast.expression.subquery.InSubquery;
 import org.opensearch.sql.ast.expression.subquery.ScalarSubquery;
 import org.opensearch.sql.ast.tree.UnresolvedPlan;
+import org.opensearch.sql.calcite.plan.OpenSearchConstants;
 import org.opensearch.sql.calcite.utils.OpenSearchTypeFactory;
 import org.opensearch.sql.calcite.utils.PlanUtils;
 import org.opensearch.sql.common.utils.StringUtils;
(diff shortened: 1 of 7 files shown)


LantaoJin (approved):
@yuancu can you fix the conflicts

yuancu:
> @yuancu can you fix the conflicts

Fixed. Thanks for reminding!

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -355,7 +357,19 @@ public RexNode visitAlias(Alias node, CalcitePlanContext context) {
 
   @Override
   public RexNode visitSpan(Span node, CalcitePlanContext context) {
-    RexNode field = analyze(node.getField(), context);
+    RexNode field;
+    if (node.getField() != null) {
```
qianheng-aws: Will it be better to set `field` in span to be `new Field(new QualifiedName(OpenSearchConstants.IMPLICIT_FIELD_TIMESTAMP))` if field is null? Then we don't need any change here.

And I'm wondering if v2 works well since we only make change for calcite visitor.

yuancu: I referred it here because I thought the AST level should be consistent with users' input. E.g. `BY span(1h)` should be parsed to `Span(value=1, unit=hour, field=null)`. Subsequent implicit reference can be added later because this semantic of how absent field should be interpreted is in a higher level than syntax tree construction.

But it's also not bizarre to refer it while building AST tree. I'll modify the implementation accordingly.

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
```diff
@@ -729,9 +729,13 @@ public String visitAggregateFunction(AggregateFunction node, String context) {
...
     @Override
     public String visitSpan(Span node, String context) {
-      String field = analyze(node.getField(), context);
+      String field = node.getField() != null ? analyze(node.getField(), context) : null;
       String value = analyze(node.getValue(), context);
-      return StringUtils.format("span(%s, %s %s)", field, value, node.getUnit().getName());
+      if (field != null) {
+        return StringUtils.format("span(%s, %s %s)", field, value, node.getUnit().getName());
+      } else {
+        return StringUtils.format("span(%s %s)", value, node.getUnit().getName());
```
LantaoJin: no test for verification?

qianheng-aws (approved):

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
yuancu: Reverted this change to `Span` as span will always come with a field with latest implementation -- I inject `@timestamp` to spans without a field specified in the AST layer now.

RyanL1997:
> This PR is stalled because it has been open for 2 weeks with no activity.

Hi @yuancu , is this implementation still in track?

LantaoJin (approved):

LantaoJin (approved):

qianheng-aws (approved):

PR #4142, opened 2025-08-27
Add shortcut for count()

noCharger:
### Description
Add shortcut for count()

### Related Issues
N/A

### Check List
~~- [ ] New functionality has javadoc added.~~ N/A
~~- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/DEVELOPER_GUIDE.rst#new-ppl-command-checklist) all confirmed.~~ N/A
~~- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).~~ N/A
~~- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).~~ N/A

### DIFF
--- a/docs/user/dql/aggregations.rst
+++ b/docs/user/dql/aggregations.rst
@@ -129,12 +129,11 @@ The aggregation could has expression as arguments::
 COUNT Aggregations
 ------------------
 
-Besides regular identifiers, ``COUNT`` aggregate function also accepts arguments such as ``*`` or literals like ``1``. The ``C()`` function can be used as an abbreviation for ``COUNT()`` with no arguments. The meaning of these different forms are as follows:
+Besides regular identifiers, ``COUNT`` aggregate function also accepts arguments such as ``*`` or literals like ``1``. The meaning of these different forms are as follows:
 
 1. ``COUNT(field)`` will count only if given field (or expression) is not null or missing in the input rows.
 2. ``COUNT(*)`` will count the number of all its input rows.
 3. ``COUNT(1)`` is same as ``COUNT(*)`` because any non-null literal will count.
-4. ``C()`` is equivalent to ``COUNT()`` and ``COUNT(*)``.
 
 Aggregation Functions
 =====================
--- a/docs/user/ppl/cmd/stats.rst
+++ b/docs/user/ppl/cmd/stats.rst
@@ -80,7 +80,7 @@ COUNT
 Description
 >>>>>>>>>>>
 
-Usage: Returns a count of the number of expr in the rows retrieved by a SELECT statement.
+Usage: Returns a count of the number of expr in the rows retrieved by a SELECT statement. The ``C()`` function can be used as an abbreviation for ``COUNT()``.
 
 Example::
 
@@ -91,6 +91,14 @@ Example::
     |---------|
     | 4       |
     +---------+
+
+    os> source=accounts | stats c();
+    fetched rows / total rows = 1/1
+    +-----+
+    | c() |
+    |-----|
+    | 4   |
+    +-----+
 
 SUM
 ---


Swiddis (dismissed):
Do we need to worry about other cases where people expect `count` as a string literal representing count? e.g. I reviewed [this condition match](https://github.com/opensearch-project/sql/pull/3993/files/8da537f2c713c06e52c6afa0e97264cd831fac99#diff-518ddbde8cc139f2be35ffaeccd6dca4468147cc06f24f3458792e05584a61daR1350) yesterday, but also stuff like [SelectResultSet](https://github.com/opensearch-project/sql/blob/b452f07c2706939c74620eedbdfc4893c6a9afff/legacy/src/main/java/org/opensearch/sql/legacy/executor/format/SelectResultSet.java#L345-L346) or [Flint's aggregate function map](https://github.com/opensearch-project/sql/blob/b452f07c2706939c74620eedbdfc4893c6a9afff/async-query-core/src/main/java/org/opensearch/sql/spark/validator/FunctionType.java#L61)

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAggregationTest.java
```diff
@@ -30,6 +30,19 @@ public void testSimpleCount() {
 
     String expectedSparkSql = "" + "SELECT COUNT(*) `c`\n" + "FROM `scott`.`EMP`";
     verifyPPLToSparkSQL(root, expectedSparkSql);
+
+    ppl = "source=EMP | stats c() as count_emp";
```
ykmr1224, line 34: nit: Let's separate method for a different test case. Same for other tests.

docs/user/dql/aggregations.rst
```diff
@@ -129,11 +129,12 @@ The aggregation could has expression as arguments::
 COUNT Aggregations
 ------------------
 
-Besides regular identifiers, ``COUNT`` aggregate function also accepts arguments such as ``*`` or literals like ``1``. The meaning of these different forms are as follows:
+Besides regular identifiers, ``COUNT`` aggregate function also accepts arguments such as ``*`` or literals like ``1``. The ``C()`` function can be used as an abbreviation for ``COUNT()`` with no arguments. The meaning of these different forms are as follows:
```
ykmr1224: Looks like this doc is for SQL. Should we fix PPL doc?

docs/user/dql/aggregations.rst
```diff
@@ -129,11 +129,12 @@ The aggregation could has expression as arguments::
...
 
 1. ``COUNT(field)`` will count only if given field (or expression) is not null or missing in the input rows.
 2. ``COUNT(*)`` will count the number of all its input rows.
 3. ``COUNT(1)`` is same as ``COUNT(*)`` because any non-null literal will count.
+4. ``C()`` is equivalent to ``COUNT()`` and ``COUNT(*)``.
```
ykmr1224: Wondering if we need to handle `c(1)` or `c(*)` as well.

docs/user/dql/aggregations.rst
RyanL1997: nit: Consider adding a usage example for `C()` to make it clearer for users

RyanL1997 (approved):
In generally lgtm.

noCharger:
> Do we need to worry about other cases where people expect `count` as a string literal representing count? e.g. I reviewed [this condition match](https://github.com/opensearch-project/sql/pull/3993/files/8da537f2c713c06e52c6afa0e97264cd831fac99#diff-518ddbde8cc139f2be35ffaeccd6dca4468147cc06f24f3458792e05584a61daR1350) yesterday, but also stuff like [SelectResultSet](https://github.com/opensearch-project/sql/blob/b452f07c2706939c74620eedbdfc4893c6a9afff/legacy/src/main/java/org/opensearch/sql/legacy/executor/format/SelectResultSet.java#L345-L346) or [Flint's aggregate function map](https://github.com/opensearch-project/sql/blob/b452f07c2706939c74620eedbdfc4893c6a9afff/async-query-core/src/main/java/org/opensearch/sql/spark/validator/FunctionType.java#L61)

@selsong will add support for count without parentheses

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAggregationTest.java
noCharger: > Let's separate method for a different test case.

Having separate test methods for count() and c() doesn't make much sense since both functions produce identical logical plans, results, and the only difference is the parser recognizing c() as an alias for count(). This creates unnecessary code duplication and maintenance overhead

ykmr1224: If only one thing is tested in a test method, it would be very easy to identify what is going wrong. You might want to refer this post: https://stackoverflow.com/questions/61400/what-makes-a-good-unit-test

ykmr1224 (approved):

PR #4152, opened 2025-08-27
Search command revamp.

### LINKED ISSUE #4007: [FEATURE] Enhance search command to support free text and wildcards.

vamsimanohar:
### Description

**Note** : Although the PR says 70 file changes, most of them are test file changes and PPL query changes in old files. Below are the important files to review.

``` 
 Parser & Grammar

  - OpenSearchPPLParser.g4 - Grammar extensions for search expression syntax
  - AstBuilder.java - Constructs Search AST from parse tree
  - AstExpressionBuilder.java - Builds search expression tree

  Search Expression AST Classes

  - Search.java - Enhanced Search AST node with expression support
  - SearchExpression.java - Base interface for search expressions
  - SearchComparison.java - Field comparison operators (=, !=, <, >, <=, >=)
  - SearchIn.java - IN operator implementation
  - Search{And,Or,Not}.java - Boolean operators
  - SearchLiteral.java - Free text search support

  Core Integration Points

  - Analyzer.java - Integrates search expression analysis into query pipeline
  - CalciteRelNodeVisitor.java - Converts Search nodes to relational operations for Calcite execution path
  ```

  **Summary**
  The search command has been redesigned with well-defined full text search functionality instead of being an extended where clause with source support. It now supports new search
   expressions that translate directly to OpenSearch's query_string DSL for optimal performance.

  The new boundary for the search command: it is designed exclusively for full-text search, always pushed down as the first operation, and limited to Lucene query syntax support. Any additional filtering should be achieved using the where command.

  The revamped search command leverages https://lucene.apache.org/core/2_9_4/queryparsersyntax.html and only supports:
  - Field comparisons: search source=logs status=200 AND method="GET"
  - Boolean operators: AND, OR, NOT with proper precedence handling
  - IN operator: search source=logs status IN (200, 201, 204)
  - Range queries: search source=logs responseTime>100 AND responseTime<=500
  - Free text search: search source=logs "error message"

  All search expressions are pushed down to OpenSearch as query_string queries, ensuring efficient data filtering at the storage layer before entering the processing pipeline.

  Solution

  Key Design Decisions

  1. Separate AST Hierarchy: Introduced dedicated SearchExpression classes distinct from regular Expression classes, maintaining clean separation between search-time filtering and
  query-time evaluation
  2. Direct Query DSL Translation: Search expressions bypass logical/physical planning stages and translate directly to OpenSearch Query DSL in the Analyzer phase for optimal
  performance
  3. Grammar Extension: Extended PPL parser to support rich search syntax while preserving full backward compatibility with existing queries

**How It Works**

  - Parser constructs a Search node containing a SearchExpression tree from the query
  - Analyzer visits the Search node and transforms expressions directly into OpenSearch query_string queries
  - Queries execute at the OpenSearch level, filtering data efficiently before it enters the processing pipeline

### Related Issues
Resolves #4007

### Check List

### CODE REVIEW

File: docs/user/ppl/cmd/search.rst [hunk 1]
```diff
@@ -62,3 +62,274 @@ PPL query::
     | 13             | Nanette   | 789 Madison Street | 32838   | F      | Nogal  | Quility  | VA    | 28  | null                 | Bates    |
     +----------------+-----------+--------------------+---------+--------+--------+----------+-------+-----+----------------------+----------+
 
+Search Expression
```
penghuo, line 25: Re-order the doc, Explain search expression, then example.

docs/user/ppl/cmd/search.rst
```diff
@@ -62,3 +62,274 @@ PPL query::
...
+
+Syntax
+------
+The search expression syntax supports:
+
+* Field-value comparisons: ``field=value``, ``field!=value``, ``field>value``, ``field>=value``, ``field<value``, ``field<=value``
+* Boolean operators: ``AND``, ``OR``, ``NOT``
+* Grouping with parentheses: ``(expression)``
+* IN operator for multiple values: ``field IN (value1, value2, value3)``
+* Text search without field (searches default field): ``"search term"``
```
penghuo: what does search term means? match phrase?
does search command support `search error`? use case is match doc  which default field include error.

docs/user/ppl/cmd/search.rst
```diff
@@ -62,3 +62,274 @@ PPL query::
...
+     - When field exists
+     - When field doesn't exist
+   * - ``field != "value"``
+     - Returns if field ≠ value
+     - **Does NOT return**
+   * - ``NOT field = "value"``
+     - Excludes if field = value
+     - **Returns**
+
+Example 7: Text Search (Default Field) and Phrase Search
```
penghuo: let's highlight example 7, this is most powerfull case.

docs/user/ppl/cmd/search.rst
```diff
@@ -62,3 +62,274 @@ PPL query::
...
+=========================================================
+
+Search text without specifying a field searches the default field::
+
+    os> search "Amber" source=accounts;
```
penghuo: what is difference of `search "Amber"` vs `search Amber`?

docs/user/ppl/cmd/search.rst
```diff
@@ -62,3 +62,274 @@ PPL query::
...
+   :header-rows: 1
+
+   * - Search For
+     - PPL Syntax
+     - Example
+   * - ``prefix*`` (wildcard)
+     - ``field=prefix*``
+     - Matches "prefix", "prefixABC", etc.
+   * - ``prefix*`` (literal)
+     - ``field="prefix\\*"``
```
penghuo: Previous section callout the following Lucene special characters are automatically escaped in values. Is it conflict.

docs/user/ppl/cmd/search.rst
```diff
@@ -62,3 +62,274 @@ PPL query::
...
+Field Types and Search Behavior
+--------------------------------
+* **Text fields**: Analyzed for full-text search
+* **Keyword fields**: Exact value matching
+* **Numeric fields**: Support range queries (>, <, >=, <=)
+* **Date fields**: Support range queries with date values
+
+Operator Precedence
+-------------------
+Parentheses are used to ensure proper operator precedence in complex expressions.
```
penghuo: search expression section already explain this, not need to re-emphasise.

docs/user/ppl/cmd/search.rst
```diff
@@ -62,3 +62,274 @@ PPL query::
...
+* Wildcards work on both keyword and text fields
+* For text fields, wildcards apply to individual tokens after analysis
+* Leading wildcards (``*term``) may impact performance on large datasets
+
+Field Types and Search Behavior
+--------------------------------
+* **Text fields**: Analyzed for full-text search
+* **Keyword fields**: Exact value matching
+* **Numeric fields**: Support range queries (>, <, >=, <=)
+* **Date fields**: Support range queries with date values
```
penghuo: we want to call-out limitation of each data type in this section? e.g., 
* keyword field support wildcard and exactally match
* text field limitation on support wildcard search
```
{"message": "[2025-08-26] logtype=ws:access error"}

message="*=ws:access*" does not match.
```

penghuo (changes requested):

docs/user/ppl/cmd/search.rst
vamsimanohar: Changed to `error` and added more details.

docs/user/ppl/cmd/search.rst
vamsimanohar: My bad, we automatically escape  all the special chards except for /, *, ?. 

These characters are not escaped as they are used for wildcard queries. If someone needs to search for these characters, they need to escape with backslash / .

docs/user/ppl/cmd/search.rst
vamsimanohar: sure, also there are some edge cases with current implementation, which I am going to come with an issue.

docs/user/ppl/cmd/search.rst
vamsimanohar: same...no difference

docs/user/ppl/cmd/search.rst
vamsimanohar: @penghuo  updated the text.
Will update further for date types and any missing types after testing.

docs/user/ppl/cmd/search.rst
```diff
@@ -16,20 +16,93 @@ Description
...
+    GET /accounts/_settings/index.query.default_field
+    
+    PUT /accounts/_settings
+    {
+      "index.query.default_field": "firstname,lastname,email"
+    }
+
+**Free-text search** - searching default field::
+
+    os> search "Amber" source=accounts;
```
penghuo: 1. add test case without double quote, `search Amber source=accounts`
2. explain difference of token w/o double quote in search expression section.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -113,9 +113,65 @@ commandName
...
+ : searchLiteral                                     # literalTerm
+ | fieldComparison                                   # comparisonTerm
+ | fieldInList                                       # inListTerm
+ ;
+
+// Standalone search literal - quotes have special meaning
+searchLiteral
+   : DQUOTA_STRING      // Double quotes = phrase search (preserve quotes)
+   | SQUOTA_STRING      // Single quotes searching for quotes
+   | BQUOTA_STRING      // Backticks will also be searched.
```
penghuo: When should use single quote and backtic? e.g.

###

{ "index": { "_id": 1 } }
{"message": "(1+1)=2"}
{ "index": { "_id": 2 } }
{"message": "The Right Way"}


`"search source=log00001 '(0+1)=2'"` should not match?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -156,15 +159,36 @@ public UnresolvedPlan visitSubSearch(OpenSearchPPLParser.SubSearchContext ctx) {
...
+        combined = searchExprs.getFirst();
+      } else {
+        // Multiple separate search expressions should each be wrapped in parentheses
+        // before being combined with AND (e.g., "a=1 b=-1" becomes "(a:1) AND (b:-1)")
+        List<SearchExpression> wrappedExprs =
+            searchExprs.stream().map(SearchGroup::new).collect(Collectors.toList());
+        combined = wrappedExprs.getFirst();
+        for (int i = 1; i < wrappedExprs.size(); i++) {
+          combined = new SearchAnd(combined, wrappedExprs.get(i));
+        }
```
penghuo: nit, simply for-loop as 
```
combined = searchExprs.stream()
    .map(SearchGroup::new)
    .reduce((acc, expr) -> new SearchAnd(acc, expr))
    .orElse(null);
```

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -156,15 +159,36 @@ public UnresolvedPlan visitSubSearch(OpenSearchPPLParser.SubSearchContext ctx) {
...
+      // Combine multiple expressions with AND
+      SearchExpression combined;
+      if (searchExprs.size() == 1) {
+        combined = searchExprs.getFirst();
+      } else {
+        // Multiple separate search expressions should each be wrapped in parentheses
+        // before being combined with AND (e.g., "a=1 b=-1" becomes "(a:1) AND (b:-1)")
+        List<SearchExpression> wrappedExprs =
+            searchExprs.stream().map(SearchGroup::new).collect(Collectors.toList());
+        combined = wrappedExprs.getFirst();
```
penghuo: searchExprs.size() could be 0, then combined is null?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -812,4 +784,136 @@ public UnresolvedExpression visitNumericSpanValue(
   public UnresolvedExpression visitLogWithBaseSpan(OpenSearchPPLParser.LogWithBaseSpanContext ctx) {
     return org.opensearch.sql.ast.dsl.AstDSL.stringLiteral(ctx.getText());
   }
+
+  // Visitor methods for search expressions
+  @Override
+  public SearchExpression visitGroupedExpression(OpenSearchPPLParser.GroupedExpressionContext ctx) {
```
penghuo, line 768: Is better to create SearchExpressionAnalyzer instead?

SearchExpression is only used by search command, and SearchExpression is not composite with other expression.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchRelevanceFunctionPushdownRule.java
```diff
@@ -0,0 +1,118 @@
...
+import org.immutables.value.Value;
+import org.opensearch.sql.opensearch.storage.scan.CalciteLogicalIndexScan;
+
+/**
+ * Planner rule that always pushes down filters containing relevance functions (like query_string,
+ * match, multi_match, etc.) to OpenSearch, regardless of the pushdown settings. This ensures
+ * relevance functions are always executed by OpenSearch for optimal performance and functionality.
+ */
+@Value.Enclosing
+public class OpenSearchRelevanceFunctionPushdownRule
```
penghuo, line 28: https://github.com/opensearch-project/sql/pull/3834/files, does this PR already cover query_string push down?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -812,4 +784,136 @@ public UnresolvedExpression visitNumericSpanValue(
...
+      String rawText = ctx.numericLiteral().getText();
+      return new SearchLiteral(rawText);
+    } else if (ctx.SQUOTA_STRING() != null || ctx.BQUOTA_STRING() != null) {
+      // Single quotes or backticks - literal search
+      String text = ctx.getText();
+      return new SearchLiteral(new Literal(text, DataType.STRING));
+    } else if (ctx.tableQualifiedName() != null) {
+      // Table name
+      UnresolvedExpression tableExpr = visit(ctx.tableQualifiedName());
+      return new SearchLiteral(tableExpr);
```
penghuo: what is use case of query_string search table name?

ppl/src/main/antlr/OpenSearchPPLParser.g4
penghuo: "search source=log00001 \"a'b\""

```
{ "index": { "_id": 3 } }
{"message": "a'b"}
{ "index": { "_id": 4 } }
{"message": "'a'b"}
```

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchRelevanceFunctionPushdownRule.java
vamsimanohar: This is used in case pushdown not enabled scenario.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
vamsimanohar: This is to support as much as possible in unquoted strings.


In unquoted Strings, I am only supporting now DQUOTA_STRING and ID only which covers only alphanumeric. Anything else should be inside  double quites.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchRelevanceFunctionPushdownRule.java
vamsimanohar: In case if customer disables pushdown setting? search still would be pushed down. Without this there is a lot of refactoring to existing ITs where simple queries like source=index age=10 are failing in NoCalcitePushDown

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
vamsimanohar:
```
if (ctx.searchExpression().isEmpty()) {
      return visitFromClause(ctx.fromClause());
} 
```

This removes the case where searchExprs.size() = 0 case...but I can be futher defensive and add this condition at relevant place.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
vamsimanohar: You mean SearchExpressionBuilder?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchRelevanceFunctionPushdownRule.java
vamsimanohar: Created an issue to remove the above rule when the setting is removed in future. 

Meta Issue: https://github.com/opensearch-project/sql/issues/4260

Actual Issue: https://github.com/opensearch-project/sql/issues/4262

docs/user/ppl/cmd/search.rst
```diff
@@ -16,20 +16,197 @@ Description
...
-* search: search keywords, which could be ignore.
+* search: search keyword, which could be ignored.
 * index: mandatory. search command must specify which index to query from. The index name can be prefixed by "<cluster name>:" for cross-cluster search.
-* bool-expression: optional. any expression which could be evaluated to boolean value.
+* search-expression: optional. Search expression that gets converted to OpenSearch query_string function which takes in Lucene Query Syntax.
 
+Search Expression
+=================
+
+The search command supports search expressions that are converted to OpenSearch ``query_string`` function calls.
```
yuancu: *OpenSearch query_string queries* or *PPL query string function* may more accurate? Please make a link to the query string doc page to make it easier for users who did not know query string: https://docs.opensearch.org/latest/query-dsl/full-text/query-string/. (Alternatively, a link to PPL query string function also works)

integ-test/src/test/resources/big5/queries/asc_sort_timestamp_can_match_shortcut.ppl
```diff
@@ -1,3 +1,3 @@
-source = big5 match(`process.name`, 'kernel')
```
yuancu, line 1: Can you confirm that this is not a breaking change? i.e. there is no customer use case for match in search queries. Other discontinued logical expression syntax includes `search ... exists [subsearch]`, `search ... in [subsearch]`, etc.

cc @LantaoJin

core/src/main/java/org/opensearch/sql/ast/expression/SearchComparison.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  private final Field field;
+  private final Operator operator;
+  private final UnresolvedExpression value;
+
+  @Override
+  public String toQueryString() {
+    String fieldName = QueryStringUtils.escapeFieldName(field.getField().toString());
```
yuancu, line 48: Should the field be validated somewhere? It seem it feeds whatever user feeds into the query string. A wrong field name will never get alter.

docs/user/ppl/cmd/search.rst
```diff
@@ -16,20 +16,197 @@ Description
...
+-------------
+
+**Search Literals (standalone terms):**
+
+* **Unquoted**: Only supports alphanumeric characters, hyphens (``-``), underscores (``_``), and wildcards (``*``, ``?``)
+  
+  - ``search error source=logs`` - searches for "error"
+  - ``search error-code source=logs`` - searches for "error-code"
+  - ``search user_123 source=logs`` - searches for "user_123"
+  - ``search error* source=logs`` - wildcard search for terms starting with "error"
```
yuancu: Can you make it explicit that multiple unquoted literals result in an AND relationship? e.g. `search a b` is equivalent to `search a AND b`.

This came to me because the default behavior of string query seem to be OR. E.g.
```json
GET time_test/_search
{
  "query": {
    "query_string": {
      "query": "value:8945 7623"
    }
  },
  "_source": {
    "includes": [
      "value"
    ]
  }
}
```

Results:
```json
{   
    "hits": [
      {
        "_index": "time_test",
        "_id": "1",
        "_score": 1,
        "_source": {
          "value": 8945
        }
      },
      {
        "_index": "time_test",
        "_id": "2",
        "_score": 1,
        "_source": {
          "value": 7623
        }
      }
    ]
}

File: core/src/main/java/org/opensearch/sql/ast/expression/QueryStringUtils.java [hunk 19]
```diff
@@ -0,0 +1,71 @@
...
+   * Escape Lucene/query_string special characters. Special characters: + - && || ! ( ) { } [ ] ^ "
+   * ~ : \ / Note: * and ? are NOT escaped to allow wildcard pattern matching
+   *
+   * @param text the text to escape
+   * @return escaped text with wildcards preserved
+   */
+  public static String escapeLuceneSpecialCharacters(String text) {
+    // List of special characters that need escaping (excluding * and ? for wildcard support)
+    String specialChars = "+-&|!(){}[]^\"~:\\/";
+
```
yuancu: Can be static final

docs/user/ppl/cmd/search.rst
```diff
@@ -16,20 +16,197 @@ Description
...
+
+* **Quoted (double quotes)**: Required for spaces and special characters. Creates a phrase search.
```
yuancu: Single quote will be parsed as a string literal, which should also work. Isn't it?

integ-test/src/test/java/org/opensearch/sql/ppl/SearchCommandIT.java
```diff
@@ -67,4 +68,645 @@ public void searchCommandWithoutSourceShouldFailToParse() throws IOException {
...
+                TEST_INDEX_OTEL_LOGS));
+    verifyDataRows(
+        result,
+        rows(
+            "Failed to parse JSON with special characters: {\"key\": \"value with \\\"quotes\\\""
+                + " and: $@#%^&*()[]{}<>|\\/?\"}"));
+  }
+
+  @Test
+  public void testSearchWithIPAddress() throws IOException {
```
yuancu, line 546: Can you also add a test for timestamp match?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -117,9 +117,61 @@ commandName
...
+
+// Field IN list - quotes are just string delimiters
+fieldInList
+ : fieldExpression IN LT_PRTHS comparisonValueList RT_PRTHS          # fieldInValues
+ ;
+
+// Comparison value - reuse literalValue which already handles signed numbers
+comparisonValue
+   : literalValue          // Includes all literal types with proper sign support
+   | ID
```
yuancu: Just curious, why is ID accepted here? I tried input a random ID without quote, it becomes `null` after visited

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -117,9 +117,61 @@ commandName
...
+
+searchTerm
+ : searchLiteral                                     # literalTerm
+ | fieldComparison                                   # comparisonTerm
+ | fieldInList                                       # inListTerm
+ ;
+
+// Standalone search literal - quotes have special meaning
+searchLiteral
+   : DQUOTA_STRING      // Double quotes = phrase search (preserve quotes) "vamsi[" OR manohar
```
yuancu: Can it be `stringLiteral` instead?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -117,9 +117,61 @@ commandName
...
+   | ID
+   | numericLiteral
+   | booleanLiteral
+   | datetimeLiteral
```
yuancu: I tried with datetimeLiteral:

```
source=bank  timestamp '2018-06-23 00:00:00'
```

It produces no result. While `source=bank | where birthdate=date '2018-06-23 00:00:00'` returns one hit.

yuancu (commented):
Can you add tests in ExplainIT for main cases?

integ-test/src/test/java/org/opensearch/sql/ppl/SearchCommandIT.java
vamsimanohar: added.

ppl/src/main/antlr/OpenSearchPPLParser.g4
vamsimanohar: Can you tell me which case is not working?

Unquoted ID is supported in SPL.
In lucene query as well, there is a fundamental different between quoted and unquoted. We want to expand unquoted support to more than ID in future.

core/src/main/java/org/opensearch/sql/ast/expression/SearchComparison.java
vamsimanohar: I was actually fixing this .

This is for `first name` -> first\ name in Lucene Query.

integ-test/src/test/resources/big5/queries/asc_sort_timestamp_can_match_shortcut.ppl
vamsimanohar: That support is actually a side effect of brining logicalExpression to where clause and the idea of the above revamp is to define a separate path for search which is always pushdown and as close as possible to query_string support with lucene query syntax.

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
```diff
@@ -33,7 +33,7 @@ public class PPLQueryDataAnonymizerTest {
 
   @Test
   public void testSearchCommand() {
-    assertEquals("source=t | where a = ***", anonymize("search source=t a=1"));
+    assertEquals("source=t a:***", anonymize("search source=t a=1"));
```
penghuo, line 36: the expectation is `search source=t a=***`, track with issue. https://github.com/opensearch-project/sql/issues/4290

docs/user/ppl/cmd/search.rst
```diff
@@ -16,49 +16,447 @@ Description
...
-* search: search keywords, which could be ignore.
+* search: search keyword, which could be ignored.
 * index: mandatory. search command must specify which index to query from. The index name can be prefixed by "<cluster name>:" for cross-cluster search.
-* bool-expression: optional. any expression which could be evaluated to boolean value.
+* search-expression: optional. Search expression that gets converted to OpenSearch `query_string <https://docs.opensearch.org/latest/query-dsl/full-text/query-string/>`_ function which uses `Lucene Query Syntax <https://lucene.apache.org/core/2_9_4/queryparsersyntax.html>`_.
 
+Search Expression
+=================
+
+The search expression syntax supports:
```
penghuo, line 28: Reformat the doc. https://github.com/vamsimanohar/sql/blob/2bddfad672f603cf8b085a130e2b6063eac17dc9/docs/user/ppl/cmd/search.rst

docs/user/ppl/cmd/search.rst
```diff
@@ -16,49 +16,447 @@ Description
...
+* **Free text search**: ``error`` or ``"error message"`` - Searches the default field configured by the ``index.query.default_field`` setting (defaults to ``*`` which searches all fields)
```
penghuo: not sure, `Free text search` should be `Full text search`

docs/user/ppl/cmd/search.rst
```diff
@@ -16,49 +16,447 @@ Description
...
+    {
+      "index.query.default_field": "firstname,lastname,email"
+    }
+
+Field Types and Search Behavior
+================================
+
+**Text Fields**: Full-text search, phrase search
+* ``search message="error occurred" source=logs``
+* Limitations: Wildcards apply to terms after analysis, not entire field value. So it might give unexpected results and not recommended.
```
penghuo: Could we remove So it might give unexpected results and not recommended. IMO, it is confuse, full text search support message="error occ*", and no performance issue.

doctest/test_data/otellogs.json
```diff
@@ -0,0 +1,60 @@
+{"index": {"_id": "1"}}
```
penghuo, line 1: Thanks for add otellogs dataset.

docs/user/ppl/cmd/search.rst
vamsimanohar: I am not sure..what shouild we call when we dont specify field.

integ-test/src/test/resources/big5/queries/desc_sort_timestamp_no_can_match_shortcut.ppl
```diff
@@ -1,3 +1,3 @@
-source = big5 match(`process.name`, 'kernel')
+source = big5 | where match(`process.name`, 'kernel')
```
penghuo: why not use `process.name=kernel`

integ-test/src/test/resources/big5/queries/sort_numeric_asc_with_match.ppl
```diff
@@ -1,3 +1,3 @@
-source = big5 match(`log.file.path`, '/var/log/messages/solarshark')
+source = big5 log.file.path=\"/var/log/messages/solarshark\"
```
penghuo, line 1: `\"` is required by json escape, right?

penghuo (approved):

core/src/main/java/org/opensearch/sql/utils/QueryStringUtils.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.utils;
+
+/** Utility class for query_string syntax operations. */
+public class QueryStringUtils {
```
ykmr1224, line 9: I'd recommend having comprehensive unit test for this class.

core/src/main/java/org/opensearch/sql/ast/expression/SearchComparison.java
```diff
@@ -0,0 +1,70 @@
...
+      return symbol;
+    }
+  }
+
+  private final Field field;
+  private final Operator operator;
+  private final SearchLiteral value;
+
+  @Override
+  public String toQueryString() {
```
ykmr1224, line 47: Are we good with null safety?

core/src/main/java/org/opensearch/sql/ast/expression/SearchComparison.java
```diff
@@ -0,0 +1,70 @@
...
+          + fieldName
+          + ":"
+          + valueStr
+          + " )";
+      case GREATER_THAN -> fieldName + ":>" + valueStr;
+      case GREATER_OR_EQUAL -> fieldName + ":>=" + valueStr;
+      case LESS_THAN -> fieldName + ":<" + valueStr;
+      case LESS_OR_EQUAL -> fieldName + ":<=" + valueStr;
+      default -> fieldName + ":" + valueStr;
+    };
```
ykmr1224: It is better stick to Java 11 syntax yet, since we need to backport to 2.19... (This will cause manual change when backporting.)

docs/user/ppl/cmd/search.rst
```diff
@@ -16,49 +16,457 @@ Description
 
 Syntax
```
ykmr1224, line 17: Can we distinguish what is supported after version 3.3?

vamsimanohar: No, we are good with all the features from previous versions. We unnecessarily complicated the functionality by having logicalExpression in search clause but now we are in parity with what is documented in previous versions.

core/src/main/java/org/opensearch/sql/ast/expression/SearchComparison.java
vamsimanohar: yes, Also, I have unit tests with `search source=index`.

core/src/main/java/org/opensearch/sql/utils/QueryStringUtils.java
vamsimanohar: This is covered indirectly via AstSearchExpressionTest. I just verified line and branch coverage. One line missing 100% branch coverage, will add few more tests into it.

ykmr1224: I think it is more important to cover major cases for Lucene special chars in unit test than just achieving 100% line/branch coverage, since we rely on this class to escape those characters.

vamsimanohar: Have covered all cases with new tests thanks.

penghuo (approved):

PR #3878, opened 2025-07-14
Support ```bin``` command with Calcite

ahkcs:
## Summary

This PR implements a complete separation of the `bin` command's `span` functionality from aggregation `span` logic, ensuring backward compatibility while introducing robust SPL-compatible binning behavior. This includes support for parameters such as `span`, `bins`, `minspan`, `start`, `end`, and `aligntime`.

---
## Related Issues

Resolves #3876 

## Test Results

Tests were run on the following real datasets:

* **`accounts.json`**: 1000 records, `age` values 20–40, `balance` values \~1000–50000
* **`time_test_data.json`**: 100 records, timestamps from 2025-07-28 to 2025-08-01

---

### 1. `span` Parameter (Numeric)

**Query**:

```
source=accounts | bin age span=10 | stats count() by age | head 5
```

**Result**:

```json
[
  [504, "30-40"],
  [451, "20-30"],
  [45,  "40-50"]
]
```

---

### 2. `bins` Parameter

**Query**:

```
source=accounts | bin age bins=3 | stats count() by age | head 5
```

**Result**:

```json
[
  [504, "30-40"],
  [451, "20-30"],
  [45,  "40-50"]
]
```

---

### 3. `minspan` Parameter

**Query**:

```
source=accounts | bin balance minspan=5000 | stats count() by balance | head 5
```

**Result**:

```json
[
  [187, "30000-40000"],
  [215, "40000-50000"],
  [213, "10000-20000"],
  [168, "0-10000"],
  [217, "20000-30000"]
]
```

---

### 4. `start` and `end` Parameters

**Query**:

```
source=accounts | bin balance span=10000 start=5000 end=45000 | stats count() by balance | head 5
```

**Result**:

```json
[
  [187, "30000-40000"],
  [215, "40000-50000"],
  [213, "10000-20000"],
  [168, "0-10000"],
  [217, "20000-30000"]
]
```

**Logical Plan**:

---

### 5. `aligntime="@d"` Parameter

**Query**:

```
source=time_test | bin @timestamp span=1hour aligntime=\"@d+3h\" | fields @timestamp | head 5
```

**Result**:

```json
 ["2025-07-28 00:00:00"],
 ["2025-07-28 01:00:00"],
 ["2025-07-28 02:00:00"],
 ["2025-07-28 03:00:00"],
 ["2025-07-28 04:00:00"]
```

---

Bin Performance Test

```
=== BIN COMMAND PERFORMANCE TEST (SIMPLIFIED) ===
[...]

PR #4163, opened 2025-08-28
`Bin` command big5 queries

ahkcs:
### Description
Adding big5 queries used for performance testing in the new features implemented for the `bin` command.

### Related Issues
#3878

### DIFF
--- a/integ-test/src/test/resources/big5/queries/bin_aligntime.ppl
+++ b/integ-test/src/test/resources/big5/queries/bin_aligntime.ppl
@@ -1,0 +1,3 @@
+source = big5
+| bin `@timestamp` span=12h aligntime="@d+3h"
+| head 10
--- a/integ-test/src/test/resources/big5/queries/bin_bins.ppl
+++ b/integ-test/src/test/resources/big5/queries/bin_bins.ppl
@@ -1,0 +1,3 @@
+source = big5
+| bin `metrics.size` bins=5
+| head 10
--- a/integ-test/src/test/resources/big5/queries/bin_default.ppl
+++ b/integ-test/src/test/resources/big5/queries/bin_default.ppl
@@ -1,0 +1,3 @@
+source = big5
+| bin `metrics.size`
+| head 10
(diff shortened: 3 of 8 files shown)


LantaoJin:
please update `CalcitePPLBig5IT`

integ-test/src/test/java/org/opensearch/sql/calcite/big5/CalcitePPLBig5IT.java
```diff
@@ -16,4 +18,52 @@ public void init() throws Exception {
     super.init();
     enableCalcite();
   }
+
+  @Test
+  public void bin_aligntime() throws IOException {
+    String ppl = sanitize(loadFromFile("big5/queries/bin_aligntime.ppl"));
```
ykmr1224: QQ: Why do we need to load from separate file? Query seems simple and it should be simpler to write here.

ahkcs: I think we are adding this IT to verify if the queries we put under the big5 queries folder works, also to be consistent with merged PR: https://github.com/opensearch-project/sql/pull/4127

Swiddis (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/big5/CalcitePPLBig5IT.java
Swiddis: ^ keeping it separate helps us copy the queries as benchmark cases, we want to ideally just copy the contents of `big5/queries` and generate workloads from it.

noCharger (commented):
@ahkcs  What queries generate the most overhead? If the implementation does not diverge, they can unite into one.

ahkcs:
> @ahkcs What queries generate the most overhead? If the implementation does not diverge, they can unite into one.

I have removed the other test cases and kept

1. bin_span_time.ppl - Time-based binning
2. bin_span_log.ppl - Logarithmic binning
3. bin_bins.ppl - Magnitude-based binning

noCharger (approved):

PR #4177, opened 2025-08-29
Updating documentation for fields and table commands

aalva500-prog:
### Description
Updating documentation foe `fields` and `table` commands.

### Check List

### DIFF
--- a/docs/user/ppl/cmd/fields.rst
+++ b/docs/user/ppl/cmd/fields.rst
@@ -13,8 +13,7 @@ Description
 ============
 Using ``field`` command to keep or remove fields from the search result.
 
-.. note::
-   Enhanced field features are available when the Calcite engine is enabled. For detailed documentation on enhanced features, see `fields-enhanced <fields-enhanced.rst>`_
+Enhanced field features are available when the Calcite engine is enabled with 3.3+ version. When Calcite is disabled, only basic comma-delimited field selection is supported.
 
 Syntax
 ============
@@ -24,14 +23,17 @@ field [+|-] <field-list>
 * field list: mandatory. comma-delimited keep or remove fields.
 
 
+Basic Examples
+==============
+
 Example 1: Select specified fields from result
-==============================================
+----------------------------------------------
 
 The example show fetch account_number, firstname and lastname fields from search results.
 
 PPL query::
 
-    os> source=accounts | fields account_number, firstname, lastname;
+    ppl> source=accounts | fields account_number, firstname, lastname;
     fetched rows / total rows = 4/4
     +----------------+-----------+----------+
     | account_number | firstname | lastname |
@@ -43,13 +45,13 @@ PPL query::
     +----------------+-----------+----------+
 
 Example 2: Remove specified fields from result
-==============================================
+----------------------------------------------
 
 The example show fetch remove account_number field from search results.
 
 PPL query::
 
-    os> source=accounts | fields account_number, firstname, lastname | fields - account_number ;
+    ppl> source=accounts | fields account_number, firstname, lastname | fields - account_number ;
     fetched rows / total rows = 4/4
     +-----------+----------+
     | firstname | lastname |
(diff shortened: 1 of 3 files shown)


Swiddis (approved):

docs/user/ppl/cmd/fields.rst
```diff
@@ -24,14 +23,17 @@ field [+|-] <field-list>
...
 Example 1: Select specified fields from result
-==============================================
+----------------------------------------------
 
 The example show fetch account_number, firstname and lastname fields from search results.
 
 PPL query::
 
-    os> source=accounts | fields account_number, firstname, lastname;
+    ppl> source=accounts | fields account_number, firstname, lastname;
```
ykmr1224: This needs to be `os>` to be executed in doctest.

docs/user/ppl/cmd/fields.rst
```diff
@@ -24,14 +23,17 @@ field [+|-] <field-list>
 * field list: mandatory. comma-delimited keep or remove fields.
 
 
+Basic Examples
+==============
+
 Example 1: Select specified fields from result
-==============================================
+----------------------------------------------
```
ykmr1224, line 30: I think we use `=`

docs/user/ppl/cmd/table.rst
```diff
@@ -0,0 +1,69 @@
...
+.. rubric:: Table of contents
+
+.. contents::
+   :local:
+   :depth: 2
+
+
+Description
+============
+The ``table`` command is an alias for the ``fields`` command and provides the same field selection capabilities. It allows you to keep or remove fields from the search result using enhanced syntax options.
```
ykmr1224: Let's link to `fields` doc from here.
And I think we want to rather simply ask reader to check `fields` doc to avoid future inconsistency of the docs (I can easily imagine we update only `fields` doc and forget updating `table` doc)

ykmr1224 (changes requested):

docs/user/ppl/cmd/fields.rst
aalva500-prog: @ykmr1224 I think I will have to add the `fields.rst` to calcite category then for the new improvements your made in doctest?

docs/user/ppl/cmd/table.rst
aalva500-prog: I don't think the doctest actually test the `table` command. Should we add it?

docs/user/ppl/cmd/fields.rst
aalva500-prog: @ykmr1224 I'm using `-------------` here because this is as sub-header of the "Basic Examples", which is using `======`. The same thing is happening in the "Enhanced Features (Version 3.3.0)". If I use `======`, they will be at the same level. I want to highlight/separate concerns between the basic functionalities and the new enhancements for better understanding. For example:

Basic Examples
==============

Example 1: Select specified fields from result
----------------------------------------------

Example 2: Remove specified fields from result
----------------------------------------------

Enhanced Features (Version 3.3.0)
===========================================

Example 3: Space-delimited field selection
-------------------------------------------

Example 4: Prefix wildcard pattern
-----------------------------------

ykmr1224: I see. We might want to fix other files, then.

docs/user/ppl/cmd/table.rst
ykmr1224: I mean removing all the details from `table` doc, and ask reader to go to `fields` doc to avoid redundancy and inconsistency in the future.

aalva500-prog: Makes sense, let's confirm with Anas next week, thanks!

docs/user/ppl/cmd/fields.rst
aalva500-prog: Maybe we can keep all of them at the same level, as you suggest. However, let's confirm with Anas next week. I think he mentioned that we should mark new features that need Calcite with 3.3+ version explicitly in the documentation.

anasalkouz (approved):

noCharger (approved):

ykmr1224 (approved):

Swiddis (approved):

docs/user/ppl/index.rst
```diff
@@ -110,6 +110,8 @@ The query start with search command and then flowing a set of command delimited
 
   - `reverse command <cmd/reverse.rst>`_
 
+  - `table command <cmd/table.rst>`_
```
Swiddis, line 113: Instead of a separate doc we could also just have this link into `fields`, not too important though.

PR #4185, opened 2025-08-29
Implement spath partial document traversal

Swiddis:
### Description
Permits `spath` to partially traverse document fields, which helps with parsing some types of nested structures.

Core idea is that `spath field=data inner.key` should be able to parse `{"data": {"inner": "{\"key\": 0}"}}`

### Related Issues
N/A

### Check List


core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
```diff
@@ -44,17 +47,80 @@ public <T, C> T accept(AbstractNodeVisitor<T, C> nodeVisitor, C context) {
...
+      }
+      if (name.length() > result.length() && matchField.charAt(name.length()) == '.') {
+        result = name;
+      }
+    }
+
+    return result;
+  }
+
+  private String unescape(String s) {
```
ykmr1224: nit: should it be `unquote` or `stripBackticks`?

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
```diff
@@ -44,17 +47,80 @@ public <T, C> T accept(AbstractNodeVisitor<T, C> nodeVisitor, C context) {
...
+   * We want input=outer, path=inner.data to match records like `{ "outer": { "inner": "{\"data\":
+   * 0}" }}`. To rewrite this as eval, that means we need to detect the longest prefix match in the
+   * fields (`outer.inner`) and parse `data` out of it. We need to match on segments, so
+   * `outer.inner` shouldn't match `outer.inner_other`.
+   *
+   * @return The field from the RelBuilder with the most overlap, or inField if none exists.
+   */
+  private String computePathField(RelBuilder builder) {
+    RelDataType rowType = builder.peek().getRowType();
+    List<String> fieldNames = rowType.getFieldNames();
```
ykmr1224: nit: it would become clearer if we rename it to `rowFieldNames` (to clearly distinguish with the input field names)

docs/user/ppl/cmd/spath.rst
```diff
@@ -78,3 +80,19 @@ PPL query::
...
     +--------+
+
+Example 4: Field traversal
+============================
+
+SPath can also traverse plain documents, in which case it acts similarly to renaming.
+
+PPL query::
+
+    PPL> source=sample | spath input=regular_object path=field;
```
ykmr1224: Let's fix it to `os>` and verify doctest pass.

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
```diff
@@ -44,17 +47,80 @@ public <T, C> T accept(AbstractNodeVisitor<T, C> nodeVisitor, C context) {
...
+    return this.inField + "." + this.path;
+  }
+
+  /**
+   * We want input=outer, path=inner.data to match records like `{ "outer": { "inner": "{\"data\":
+   * 0}" }}`. To rewrite this as eval, that means we need to detect the longest prefix match in the
+   * fields (`outer.inner`) and parse `data` out of it. We need to match on segments, so
+   * `outer.inner` shouldn't match `outer.inner_other`.
+   *
+   * @return The field from the RelBuilder with the most overlap, or inField if none exists.
```
ykmr1224, line 87: I am confused. Is `input` parameter for specifying where we want to read JSON from? This description looks like we are reading JSON which includes `outer` attribute inside.

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
```diff
@@ -44,17 +47,80 @@ public <T, C> T accept(AbstractNodeVisitor<T, C> nodeVisitor, C context) {
...
+    for (String name : fieldNames) {
+      if (!matchField.startsWith(name)) {
+        continue;
+      }
+      if (name.length() > result.length() && matchField.charAt(name.length()) == '.') {
+        result = name;
+      }
+    }
+
+    return result;
```
ykmr1224, line 98: Let's extract this part and give a good name for readability. It was difficult for me to understand the intent of the logic.

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
Swiddis: Input is the outermost field we want to start extracting the inner values from, so on something like `{ "outer": { "inner": "{\"data\": 0}" }}` then `input=outer` means that Spath will be processing the document `{ "inner": "{\"data\": 0}" }`. Then `path=inner.data` would access the value `0`, or `path=inner` would access the value `"{\"data\": 0}"`

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
Swiddis: Trying to think of how to extract this, the core loop depends on every field in scope except `rowType` so it's tricky to make it smaller in a useful way...

docs/user/ppl/cmd/spath.rst
Swiddis: Calcite doctests weren't ready when I originally wrote that, will set up a data set & proper test cases soon

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
ykmr1224: Does that mean `outer` is a column? In that case I think we want to separate that from JSON string.

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
Swiddis: Simplified it a bit -- focused on the inner matching logic instead of the loop

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
Swiddis: Yeah, I think in the future it makes sense to remove the `input` field and just navigate directly to the specified path. For this change, the intended behavior is to allow this type of mixing so you don't need to worry about where exactly the boundary is. I might cut a future PR to make `input` be an empty string by default?

Swiddis:
Doctest failures unrelated? Can see the spath ones passed, metadata wasn't touched
image

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
ykmr1224: I feel it is making thing more complicated.
In my opinion, it is simpler and easier to understand if `input` simply point the column containing JSON, and `path` specifies the path in JSON.

But I am open if others feels current approach is better.

PR #4190, opened 2025-09-02
Run unit test suites in parallel

Swiddis:
### Description
Our unit tests don't have any cross-dependency, so we can freely turn on parallelization for them. Makes builds a bit faster locally. (At least the parts where most of the feedback is -- async is still slow but now it runs at the end instead of the start.)

### Related Issues
Slow local builds. I want `./gradlew -x doctest -x integTest` to be fast.

### Check List

### DIFF
--- a/async-query-core/build.gradle
+++ b/async-query-core/build.gradle
@@ -85,6 +85,7 @@ spotless {
 }
 
 test {
+    maxParallelForks = Runtime.runtime.availableProcessors()
     useJUnitPlatform()
     testLogging {
         events "skipped", "failed"
--- a/async-query/build.gradle
+++ b/async-query/build.gradle
@@ -53,6 +53,7 @@ dependencies {
 }
 
 test {
+    maxParallelForks = Runtime.runtime.availableProcessors()
     useJUnitPlatform {
         includeEngines("junit-jupiter")
     }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -94,6 +94,7 @@ spotless {
 }
 
 test {
+    maxParallelForks = Runtime.runtime.availableProcessors()
     useJUnitPlatform()
     testLogging {
         events "skipped", "failed"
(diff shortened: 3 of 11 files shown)


vamsimanohar:
I haven't tested out but thanks for doing this. 
These small things save a lot of time. The other day maven repository order killed my time.

LantaoJin:
Does this fixing save time of integ-test for building locally with `--parallel` only, or does it reduce the time of workflow in GitHub CI either?

Swiddis:
It reduces the github CI but not very much -- the individual tasks are parallelized now (so e.g. `core:test` will be faster), but the tasks are globally still serial. It has the most effect locally where you can `-x integTest --parallel`.

The problem with enabling parallelism between tasks right now is it seems there's an invisible dependency for `integTest`, so just `gradle build --parallel` is failing (even though `integTest` itself is serial). I've been meaning to figure out what it is. The high CPU usage also causes issues for doctest. In the meantime, I also want to make integTest itself parallel, but [haven't been able to get it to even try](https://github.com/opensearch-project/OpenSearch/issues/19186).

Something we could do is break up our testing action into unit tests, integTest, and doctest as three different actions. Then we run `./gradlew integTest`, `./gradlew doctest`, and `./gradlew build -x integTest -x doctest --parallel`. That'd give us much faster CI overall.

Swiddis:
> Something we could do is break up our testing action into unit tests, integTest, and doctest as three different actions. Then we run ./gradlew integTest, ./gradlew doctest, and ./gradlew build -x integTest -x doctest --parallel. That'd give us much faster CI overall.

#4193, looks like it works. If preferred, close this in favor of that, but I'd rather get this merged quicker since it has immediate benefit for local development, and don't want to block this on any issues with those (riskier) CI updates.

penghuo (commented):
Thanks. 
Could u update doc also. https://github.com/opensearch-project/sql/blob/main/DEVELOPER_GUIDE.rst#building-and-running-tests

LantaoJin (dismissed):

Swiddis:
Test failures pending #4215, was passing before doc update

LantaoJin (approved):

Issue #4047, opened 2025-08-15
[FEATURE] PPL `eventstats` command enhancement


dai-chen:
**Is your feature request related to a problem?**

The `eventstats` command was added in https://github.com/opensearch-project/sql/pull/3585, but there remain important gaps compared with Splunk SPL’s `eventstats`, affecting feature parity.

**What solution would you like?**

<img width="1149" height="879" alt="Image" src="https://github.com/user-attachments/assets/066c4f57-de44-4f79-9874-8795ee9abbe4" />

**What alternatives have you considered?**

N/A

**Do you have any additional context?**

Execution plan & priorities summary:

- P0
    - Close gaps in common aggregation functions (tracked separately)
        - https://github.com/opensearch-project/sql/pull/4084
        - https://github.com/opensearch-project/sql/pull/4100
    - Support aggregation wildcard * (tracked under the stats enhancement issue) [TBD]
- P0.5
    - Investigate the possibility of pushdown optimizations for SQL window function
    - Add a memory limit guard for large inputs.
- P1:
    - Deliver full parity for the `allnum` parameter
    - Add support for more aggregation functions

PR #4212, opened 2025-09-03
Add earliest/latest aggregate function for eventstats PPL command

ykmr1224:
### Description
- Add earliest/latest aggregate function for eventstats PPL command

### Related Issues
- https://github.com/opensearch-project/sql/issues/4047

### Check List


dai-chen:
CI is failing on doctest?

ykmr1224:
> CI is failing on doctest?

Oh, seems I did wrong while merging and `category.json` was invalid... fixed now.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -483,6 +483,7 @@ eventstatsAggTerm
 
 windowFunction
    : windowFunctionName LT_PRTHS functionArgs RT_PRTHS
+   | earliestLatestFunction
```
dai-chen: is earliest and latest function already in `statsFunctionName`?

docs/user/ppl/cmd/eventstats.rst
```diff
@@ -93,16 +93,16 @@ Usage: Returns a count of the number of expr in the rows retrieved by a SELECT s
 
 Example::
 
-    PPL> source=accounts | eventstats count();
+    os> source=accounts | fields account_number, gender, age | eventstats count() | sort account_number;
```
dai-chen, line 96: Thanks for the fix! This is much clear with fewer fields in example.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -717,12 +717,36 @@ public UnresolvedExpression visitBetween(OpenSearchPPLParser.BetweenContext ctx)
...
   @Override
   public UnresolvedExpression visitWindowFunction(OpenSearchPPLParser.WindowFunctionContext ctx) {
-    Function f =
-        buildFunction(ctx.windowFunctionName().getText(), ctx.functionArgs().functionArg());
+    Function f;
+    if (ctx.windowFunctionName() != null) {
+      // Standard window function
+      f = buildFunction(ctx.windowFunctionName().getText(), ctx.functionArgs().functionArg());
+    } else if (ctx.earliestLatestFunction() != null) {
+      f = handleEarliestLatestWindowFunction(ctx.earliestLatestFunction());
```
dai-chen: Could you clarify why we need special logic for earliest/latest? I thought it's the same as other regular window function.

ppl/src/main/antlr/OpenSearchPPLParser.g4
ykmr1224: No, it is defined separately in `earliestLatestFunction`

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ykmr1224: It is to handle default time field, but let me see if I can improve it...

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -717,8 +703,12 @@ public UnresolvedExpression visitBetween(OpenSearchPPLParser.BetweenContext ctx)
...
   @Override
   public UnresolvedExpression visitWindowFunction(OpenSearchPPLParser.WindowFunctionContext ctx) {
-    Function f =
-        buildFunction(ctx.windowFunctionName().getText(), ctx.functionArgs().functionArg());
+    Function f;
+    if (ctx.windowFunctionName() != null) {
+      f = buildFunction(ctx.windowFunctionName().getText(), ctx.functionArgs().functionArg());
+    } else {
+      throw new SyntaxCheckException("Invalid window function");
+    }
```
dai-chen: Is this possible with ANTLR?

docs/user/ppl/cmd/eventstats.rst
```diff
@@ -221,17 +221,16 @@ Usage: VAR_POP(expr). Returns the population standard variance of expr.
 
 Example::
 
-    PPL> source=accounts | eventstats var_pop(age);
+    os> source=accounts | fields account_number, gender, age | eventstats var_pop(age) | sort account_number;
```
dai-chen, line 224: Just wonder is there any way to avoid adding a sort in each example query? Why some queries below doesn't require sorting?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -517,14 +517,13 @@ statsAggTerm
...
-   : statsFunctionName LT_PRTHS valueExpression RT_PRTHS        # statsFunctionCall
-   | (COUNT | C) LT_PRTHS evalExpression RT_PRTHS               # countEvalFunctionCall
+   : (COUNT | C) LT_PRTHS evalExpression RT_PRTHS               # countEvalFunctionCall
    | (COUNT | C) (LT_PRTHS RT_PRTHS)?                           # countAllFunctionCall
    | PERCENTILE_SHORTCUT LT_PRTHS valueExpression RT_PRTHS      # percentileShortcutFunctionCall
    | (DISTINCT_COUNT | DC | DISTINCT_COUNT_APPROX) LT_PRTHS valueExpression RT_PRTHS    # distinctCountFunctionCall
    | takeAggFunction                                            # takeAggFunctionCall
    | percentileApproxFunction                                   # percentileApproxFunctionCall
-   | earliestLatestFunction                                     # earliestLatestFunctionCall
+   | statsFunctionName LT_PRTHS functionArgs RT_PRTHS           # statsFunctionCall
```
dai-chen, line 582: The change from `valueExpression` to `functionArgs` will allow logical expression as stats function input right? If this is intentional, could you double check if any impact or test to cover this?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -289,7 +290,16 @@ public UnresolvedExpression visitSortField(SortFieldContext ctx) {
   /** Aggregation function. */
   @Override
   public UnresolvedExpression visitStatsFunctionCall(StatsFunctionCallContext ctx) {
-    return new AggregateFunction(ctx.statsFunctionName().getText(), visit(ctx.valueExpression()));
+    return buildAggregateFunction(
+        ctx.statsFunctionName().getText(), ctx.functionArgs().functionArg());
```
RyanL1997, line 271: Just being curious, Where is the `@timestamp` default handled now? The documentation still claims it defaults to `@timestamp`, but I don't see this logic in the new implementation.

docs/user/ppl/cmd/eventstats.rst
```diff
@@ -290,16 +289,111 @@ For details on algorithm accuracy and precision control, see the `OpenSearch Car
...
+EARLIEST (Since 3.3)
+---------------------
+
+Description
+>>>>>>>>>>>
+
+Usage: EARLIEST(field [, time_field]). Return the earliest value of a field based on timestamp ordering. This function enriches each event with the earliest value found within the specified grouping.
+
+* field: mandatory. The field to return the earliest value for.
+* time_field: optional. The field to use for time-based ordering. Defaults to @timestamp if not specified.
```
RyanL1997, line 312: But looking at the test expectations:

`earliest_message=[ARG_MIN($2, $3) OVER ()]`

Where `$3` appears to be `@timestamp` field. How is this default determined? There's no explicit logic I found that says "if only one argument, use `@timestamp` as second argument."

dai-chen:
Hi @xinyual Could you help take another look since you worked on this before?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ykmr1224: Very good question. It is handled in [PPLFuncImpTable](https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java#L1060C28-L1060C44)

docs/user/ppl/cmd/eventstats.rst
ykmr1224: see above comment

docs/user/ppl/cmd/eventstats.rst
ykmr1224: I think the inconsistent order is coming from OpenSearch query result. We need dive deep to come up with consistent order.

Fixed examples to have sort command.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ykmr1224: Nice catch! Fixed.

xinyual:
> Hi @xinyual Could you help take another look since you worked on this before?

LGTM

ppl/src/main/antlr/OpenSearchPPLParser.g4
ykmr1224: I am checking the validation, and found parameter was not validated for window functions. Working on the fix.

docs/user/ppl/cmd/eventstats.rst
dai-chen: Okay, just thinking can we handle this in test framework itself, otherwise all tests requires sorting?

ykmr1224:
I found argument count/type validation for window functions were not working, and added verification + tests for it.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -533,6 +533,7 @@ public RexNode visitWindowFunction(WindowFunction node, CalcitePlanContext conte
                   (arguments.isEmpty() || arguments.size() == 1)
                       ? Collections.emptyList()
                       : arguments.subList(1, arguments.size());
+              PPLFuncImpTable.INSTANCE.validateAggFunctionSignature(functionName, field, args);
```
dai-chen, line 536: Is this validation missing for all window function because we bypass Calcite SQL validator?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1162,36 +1169,38 @@ void populate() {
...
-   * @return a string in the format [type1,type2,...] representing the argument types
-   */
-  private static String getActualSignature(List<RelDataType> argTypes) {
-    return "["
-        + argTypes.stream()
-            .map(OpenSearchTypeFactory::convertRelDataTypeToExprType)
-            .map(Objects::toString)
-            .collect(Collectors.joining(","))
-        + "]";
+  static List<RexNode> resolveTimeField(List<RexNode> argList, CalcitePlanContext ctx) {
```
dai-chen, line 1213: We may reuse this in any function with implicit @timestamp field? https://github.com/opensearch-project/sql/pull/4138

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ykmr1224: It is because we are directly converting window functions into other built-in function and passing only required parameters. Unneeded parameters were just ignored during that process and didn't raise error.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ykmr1224: This method relies on the position of time field param, and cannot be directly reused, but we should unify the logic to refer implicit(default) timestamp field.

Added tracking issue: https://github.com/opensearch-project/sql/issues/4275

ppl/src/main/antlr/OpenSearchPPLParser.g4
ykmr1224: fixed.

LantaoJin:
@yuancu , could you review this PR?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -407,25 +405,40 @@ public void registerExternalAggOperator(
     aggExternalFunctionRegistry.put(functionName, Pair.of(signature, handler));
   }
 
+  public void validateAggFunctionSignature(
```
Swiddis, line 408: suggestion: Consider using an `AggFunctionSignature` class instead of shotgun validation

From a safety standpoint, it would be better design to create a record type and construct it with validation. Then any methods that require the function signature can take an `AggFunctionSignature`, and it's guaranteed that the input is valid since you can't construct the object otherwise.

Otherwise, we rely on the rest of the code to "just know" whether the input is already validated or not.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -129,7 +128,7 @@ public void supportPartialPushDown() throws IOException {
   @Ignore("We've supported script push down on text field")
   @Test
   public void supportPartialPushDown_NoPushIfAllFailed() throws IOException {
-    Assume.assumeTrue("This test is only for push down enabled", isPushdownEnabled());
+    enabledOnlyWhenPushdownIsEnabled();
```
Swiddis, line 131: question: Could this be an annotation instead of a method?

Might be nicer to read, but not sure how involved creating custom annotations is.

ykmr1224: That's a good point. I think we can do that. Let's me take a look and address in a separate PR if it is feasible (and reasonable).

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ykmr1224: It is actually using the signature coming from function registry (this is just delegating validation). It needs to be done here since we are converting some aggregate/window function to different ones, and add/remove/change attributes.

ykmr1224:
@yuancu merging now, but please let me know if you see any risk or concern.

PR #4215, opened 2025-09-03
Fix AnalyzerTest

ykmr1224:
### Description
- Fix AnalyzerTest
- Follow up for https://github.com/opensearch-project/sql/pull/4214

Issue #3606, opened 2025-05-06
[RFC] Enable Merge Queues


Swiddis:
**Problem Statement**
Inspired over the past year by [Tyler Cipriani's post](https://tylercipriani.com/blog/2022/02/17/its-still-not-rocket-science/) and using them on other projects: we currently don't have any post-merge protection for our branch. This causes issues with merge skew: two colliding PRs can cause a failing main branch (while passing individually).

![Merge queue example: without queues, branches A and B may pass but A+B fails. With queues, B will be tested as A+B instead.](https://github.com/user-attachments/assets/19a629b4-6aa3-46f7-86a6-be650ffe44b0)

**Current State**
Because we don't have this protection, we're pretty regularly susceptible to `main` breaking due to external factors or collisions. If we wanted to be safe, we would need to effectively manually do the same thing as a merge queue: rerun tests right before merging. When we do this, it causes a lot of time to be spent waiting synchronously on CI.

It also means that there's a natural incentive to "just merge," and ignore CI entirely. It's up to maintainer discipline whether the tests always pass or not.

**Proposal**
We can enable [merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) for our repo. Doing this will:
- Enable us to more confidently merge PRs while protecting the main branch.
- Speed up merging: the queue can be run in parallel.
- Guarantee that `main` is always passing tests.
- Incentivize fixing flaky tests, as these will have a greater visible impact in the queue.
- Prevent instances of accidentally merging PRs with failing tests.

The primary drawback is: the queue can get blocked if an upstream change breaks. It's up to the team whether we want to prioritize always passing tests, or avoiding the inconvenience of waiting on patches for upstream breakage. There's a permissions role that lets you override the merge queue in an emergency.

**Implementation Discussion**
We submit a request to the Org's `.github`: https://github.com/opensearch-project/.github/issues/new?template=GITHUB_REQUEST_TEMPLATE.yaml. From there, all we need to do is add a `merge_group:` trigger to relevant CI actions (most obviously the tests, possibly other actions too). Then the queue will be active for all merged PRs.

gaiksaya:
I was going through the [documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) for merge queues and came across this paragraph:


[...]

PR #4216, opened 2025-09-03
Add merge_group trigger to test workflows

Swiddis:
### Description
Re: #4215, it seems prudent to actually act on #3606 now -- since automerge relies on branch protections to block on CI, but enabling branch protections globally means non-admins can never override them.

### Related Issues
#3606 

### Check List

### DIFF
--- a/.github/workflows/integ-tests-with-security.yml
+++ b/.github/workflows/integ-tests-with-security.yml
@@ -9,6 +9,7 @@ on:
     paths:
       - 'integ-test/**'
       - '.github/workflows/integ-tests-with-security.yml'
+  merge_group:
 
 jobs:
   Get-CI-Image-Tag:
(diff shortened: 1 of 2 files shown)


Swiddis:
> So PR ready to merge will stay in the queue until manual approved?

The merge queue needs to actually be enabled, see https://github.com/opensearch-project/.github/issues/381

But the flow is that once the PR goes in the queue, it will try to merge it with `main` and run tests, and it will reject the PR if tests fail. If the tests pass, it's actually merged. That solves a few problems for us:
- We can no longer accidentally merge PRs with failing tests, the queue would need to be manually overrided (which it usually shouldn't)
- We can hit "merge" on PRs with failing tests if we believe `main` has a fix, since it'll test in the queue before actually merging (and cancel the merge if tests are still failing after the fix)
- We can safely bulk-merge without worrying about conflicts between PRs (any issues will be kicked out before the actual merge)

PR #4222, opened 2025-09-03
Don't recreate indices on every doctest

Swiddis:
### Description
Since I've been working on build timing lately: the vast majority of our time in doctest is spent repeatedly creating and destroying test indices with `refresh=wait_for`.

Originally I just wanted to rewrite this to run all the indexing in parallel, but it turns out: our doctests don't do any index modifications. There's no real reason to be repeatedly destroying and reindexing all this data. Disabling this behavior (in addition to parallel index creation) makes doctest run in ~13 seconds locally:

```
/workplace/sawiddis/sql/doctest/../docs/user/dql/expressions.rst
Doctest: expressions.rst ... ok
/workplace/sawiddis/sql/doctest/../docs/user/general/comments.rst
Doctest: comments.rst ... ok
...
/workplace/sawiddis/sql/doctest/../docs/user/ppl/cmd/stats.rst
Doctest: stats.rst ... ok

----------------------------------------------------------------------
Ran 54 tests in 12.838s
```

(Also, I found a missing task dependency that was causing `./gradlew -x integTest --parallel` to fail to run doctest)

### Related Issues
N/A

### Check List

### DIFF
--- a/doctest/test_docs.py
+++ b/doctest/test_docs.py
@@ -11,6 +11,7 @@ import re
 import subprocess
 import sys
 import unittest
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from functools import partial
 
 import click
@@ -205,6 +206,7 @@ class TestDataManager:
     
     def __init__(self):
         self.client = OpenSearch([ENDPOINT], verify_certs=True)
+        self.is_loaded = False
     
     def load_file(self, filename, index_name):
         mapping_file_path = './test_mapping/' + filename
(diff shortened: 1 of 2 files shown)


doctest/test_docs.py
```diff
@@ -282,7 +299,7 @@ def set_up_test_indices_without_calcite(test):
 
 
 def tear_down(test):
-    get_test_data_manager().cleanup_indices()
+    pass
```
ps48, line 302: Thanks for the PR I was thinking about improving this as well. Have a question, will this lead to cascading test failures if the first test failed to load indexes?

Swiddis: We run doctest with `--fastfail` so if the first test failed, the rest wouldn't even run. Same behavior as now, if any individual run has an index failure. (Here, the indexing exception would be raised via `future.result()`)

ps48 (approved):

Issue #4203, opened 2025-09-02
[FEATURE] PPL `first`/`last` aggregate function


ykmr1224:
Add first/last aggregate function for PPL (Calcite)

### FIRST Function
**Syntax**: `FIRST(field)`
**Description**: Returns the first value of the specified field in natural document order
**Parameters**:
- `field` (required): The field name to get the first value from
**Return Type**: Same as input field type (nullable)
**Behavior**: 
- Returns the first occurrence of the field value in the result set
- Uses natural document order (no sorting applied)
- Returns NULL if no records exist or field is NULL in first record

### LAST Function
**Syntax**: `LAST(field)`
**Description**: Returns the last value of the specified field in natural document order
**Parameters**:
- `field` (required): The field name to get the last value from
**Return Type**: Same as input field type (nullable)
**Behavior**:
- Returns the last occurrence of the field value in the result set
- Uses reverse natural document order (equivalent to sorting by `_doc DESC`)
- Returns NULL if no records exist or field is NULL in last record

### Usage Examples
```ppl
# Basic usage
source=logs | stats first(message), last(status) by host

# Combined with other aggregations
source=metrics | stats first(cpu_usage), last(memory_usage), count(), avg(response_time) by server

# Sequential processing after sorting
source=events | sort timestamp | stats first(event_type), last(event_data) by session_id

# With eventstats (if extended)
source=transactions | eventstats first(amount) by account_id
```

### Comparison with Related Functions
| Function | Ordering | Use Case | Parameters |
|----------|----------|----------|------------|
| `FIRST(field)` | Natural document order | Sequential processing | Single field |
| `LAST(field)` | Reverse document order | Sequential processing | Single field |
| `EARLIEST(field)` | Time-based (ASC) | Time-based analysis | Field + optional time field |
| `LATEST(field)` | Time-based (DESC) | Time-based analysis | Field + optional time field |
| `TAKE(field, n)` | Natural document order | Multiple record retrieval | Field + count |

ykmr1224:
I checked possibility to reuse `earliest`/`latest` implementation for `first`/`last`.

[...]

PR #4223, opened 2025-09-04
Support first/last aggregation functions for PPL

ahkcs:
<html><head></head><body><h1>Add <code inline="">FIRST()</code> and <code inline="">LAST()</code> aggregate functions</h1>
<p><strong>Resolves:</strong> #4203</p>
<h2>Summary</h2>
<p>This PR implements <code inline="">FIRST(field)</code> and <code inline="">LAST(field)</code> for PPL , returning the first and last values of a field in <strong>natural index (document) order</strong>.</p>
<ul>
<li>
<p><strong>New Aggregates:</strong> <code inline="">FIRST(field)</code>, <code inline="">LAST(field)</code></p>
</li>
<li>
<p><strong>Document Order Processing:</strong> Natural document order</p>
</li>
<li>
<p><strong>OpenSearch Pushdown:</strong> Translates to <code inline="">top_hits</code> with <code inline="">size: 1</code> for optimal performance</p>
</li>
<li>
<p><strong>Calcite Integration:</strong> Implemented as Calcite UDAFs</p>
</li>
</ul>

<h2>Function Specs</h2>
<h3>Add first/last aggregate function for PPL (Calcite)</h3>
<h4>FIRST Function</h4>
<ul>
<li>
<p><strong>Syntax:</strong> <code inline="">FIRST(field)</code></p>
</li>
<li>
<p><strong>Description:</strong> Returns the first value of the specified field in natural document order.</p>
</li>
<li>
<p><strong>Parameters:</strong></p>
<ul>
<li>
<p><code inline="">field</code> <em>(required)</em>: The field name to get the first value from.</p>
</li>
</ul>
</li>
<li>
<p><strong>Return Type:</strong> Same as input field type <em>(nullable)</em>.</p>
</li>
<li>
<p><strong>Behavior:</strong></p>
<ul>
<li>
<p>Returns the first occurrence of the field value in the result set (or the first per group when used with <code inline="">BY</code>).</p>
</li>
<li>
<p>Uses natural document order (no sorting applied).</p>
</li>
<li>
<p>Returns <strong>NULL</strong> if no records exist, or if the field is <strong>NULL</strong> in the first record of the bucket.</p>
</li>
</ul>
</li>
</ul>
<h4>LAST Function</h4>
<ul>
<li>
<p><strong>Syntax:</strong> <code inline="">LAST(field)</code></p>
</li>
<li>
<p><strong>Description:</strong> Returns the last value of the specified field in natural document order.</p>
</li>
<li>
<p><strong>Parameters:</strong></p>
<ul>
<li>
<p><code inline="">field</code> <em>(required)</em>: The field name to get the last value from.</p>
</li>
</ul>
</li>
<li>
<p><strong>Return Type:</strong> Same as input field type <em>(nullable)</em>.</p>
</li>
<li>
<p><strong>Behavior:</strong></p>
<ul>
<li>
<p>Returns the last occurrence of the field value in the result set (or the last per group when used with <code inline="">BY</code>).</p>
</li>
<li>
<p>Uses <strong>reverse</strong> natural document order (equivalent to sorting by <code inline="">_doc DESC</code>).</p>
</li>
<li>
<p>Returns <strong>NULL</strong> if no records exist, or if the field is <strong>NULL</strong> in the last record of the bucket.</p>
</li>
</ul>
</li>
</ul>
<hr>
<h2>Usage Examples</h2>
<pre><code class="language-ppl">
-- Basic usage
source=logs | stats first(message), last(status) by host

-- Combined with other aggregations
source=metrics | stats first(cpu_usage), last(memory_usage), count(), avg(response_time) by server

-- Sequential processing after sorting
source=events | sort timestamp | stats first(event_type), last(event_data) by session_id
</code></pre>

</ul></body></html>


dai-chen (commented):
A high level question: I see Calcite supports `FIRST/LAST_VALUE` as window function, can we simply register it as aggregate function?

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/FirstLastParser.java
```diff
@@ -0,0 +1,43 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.opensearch.search.SearchHit;
+import org.opensearch.search.aggregations.Aggregation;
+import org.opensearch.search.aggregations.metrics.TopHits;
+
+/** Parser for FIRST and LAST aggregate functions using TopHits aggregation. */
+@EqualsAndHashCode
+@RequiredArgsConstructor
+public class FirstLastParser implements MetricParser {
```
dai-chen: I think Parser should be correspondent to DSL query. Can we make this generic for top hit parser?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -519,6 +520,10 @@ earliestLatestFunction
    : (EARLIEST | LATEST) LT_PRTHS valueExpression (COMMA timeField = valueExpression)? RT_PRTHS
    ;
 
+firstLastFunction
+   : (FIRST | LAST) LT_PRTHS valueExpression RT_PRTHS
```
dai-chen: just add both to `statsFunctionName`?

docs/user/ppl/cmd/stats.rst
```diff
@@ -30,6 +30,10 @@ The following table dataSources the aggregation functions and also indicates how
 +----------+-------------+-------------+
 | LIST     | Ignore      | Ignore      |
 +----------+-------------+-------------+
+| FIRST    | Return NULL | Return NULL |
++----------+-------------+-------------+
+| LAST     | Return NULL | Return NULL |
```
dai-chen: For this table, my understanding is how the aggregate function handle NULL. If we say "return NULL", that means we return NULL if we see any NULL value?

ahkcs:
> A high level question: I see Calcite supports `FIRST/LAST_VALUE` as window function, can we simply register it as aggregate function?

That’s a great question! Calcite does support `FIRST_VALUE` and `LAST_VALUE` as window functions. After experimenting, though, we found that while registration and logical plan generation work fine, Calcite doesn’t provide the necessary conversion rules to run them in aggregate contexts. This leads to a `CannotPlanException: Missing conversion is LogicalAggregate[convention: NONE -> ENUMERABLE]`.

In practice, that means we’d need to build additional Calcite infrastructure to bridge this gap—which would end up being more effort than using our current UDAF approach. So I think while it’s technically possible, sticking with UDAF is the more straightforward path for now in my opinion.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/FirstLastParser.java
ahkcs: Moved the logic to TopHitsParser

ppl/src/main/antlr/OpenSearchPPLParser.g4
ahkcs: Updated

docs/user/ppl/cmd/stats.rst
ahkcs: Removed

ahkcs: For `first` and `last`, it will return NULL if no records exist or field is NULL in last record, but it seems like this info shouldn't be in this table

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/FirstAggFunction.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Override
+  public FirstAccumulator add(FirstAccumulator acc, Object... values) {
+    Object candidateValue = values[0];
+    // Accept the first value encountered, including NULL values
+    // This ensures we return NULL if the first record has a NULL field value
+    if (!acc.hasValue()) {
+      acc.setValue(candidateValue);
+    }
```
Swiddis: Is this always guaranteed to run in order? Not sure if we have to worry about parallelism here

Swiddis (dismissed):

docs/user/ppl/cmd/stats.rst
dai-chen: Could you confirm if this align with SPL or SPL will ignore null and return first non-null if any. Also please clarify this in user manual. Thanks!

ahkcs: After confirmation, it turns out that SPL will return the first non-null value for `first` and `last`, I have updated the implementation and documentation to comply with that

ahkcs: I have added UT for testing null handling

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/FirstAggFunction.java
ahkcs: Good catch! I've made both functions thread-safe by adding synchronized setValue() and volatile fields

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/FirstAggFunction.java
```diff
@@ -0,0 +1,66 @@
...
+  public Object result(FirstAccumulator accumulator) {
+    return accumulator.value();
+  }
+
+  @Override
+  public FirstAccumulator add(FirstAccumulator acc, Object... values) {
+    Object candidateValue = values[0];
+    // Only accept the first non-null value encountered
+    // Skip null values to find the first actual value
+    if (candidateValue != null && !acc.hasValue()) {
```
dai-chen: `!acc.hasValue` is already done inside Accumulator, probably we can remove the check here as well as the getter?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -365,6 +365,32 @@ public void testExplainOnEarliestLatestWithCustomTimeField() throws IOException
...
+        explainQueryToString(
+            String.format(
+                "source=%s | stats first(firstname) as first_name, last(firstname) as"
+                    + " last_name by gender",
+                TEST_INDEX_BANK)));
+  }
+
+  // Only for Calcite
+  @Test
+  public void testExplainOnFirstLastDifferentFields() throws IOException {
```
dai-chen: These looks like very similar test. Maybe only keep this one?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -320,6 +321,25 @@ yield switch (functionName) {
                   .size(helper.inferValue(args.getLast(), Integer.class))
                   .from(0),
               new TopHitsParser(aggFieldName));
+          case FIRST -> {
```
dai-chen, line 327: try to add UT for this?

ahkcs: Added UT

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
ahkcs: Removed `testExplainOnFirstLastDifferentFields`

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/FirstAggFunction.java
ahkcs: Removed

docs/user/ppl/cmd/stats.rst
```diff
@@ -463,13 +467,83 @@ Example with custom time field::
...
+    +------------------+--------+
+    | first(firstname) | gender |
+    |------------------+--------|
+    | Nanette          | F      |
+    | Amber            | M      |
+    +------------------+--------+
+
+Example with multiple aggregations::
+
+    os> source=accounts | stats first(firstname), last(firstname), count() by gender;
```
dai-chen: np: I see some examples here and below do both `first` and `last`?

opensearch/src/test/java/org/opensearch/sql/opensearch/request/AggregateAnalyzerTest.java
```diff
@@ -345,6 +347,137 @@ void analyze_groupBy_TextWithoutKeyword() {
     assertEquals("[field] must not be null", exception.getCause().getMessage());
   }
 
+  @Test
+  void analyze_firstAggregation() throws ExpressionNotAnalyzableException {
+    AggregateCall firstCall =
```
dai-chen: np: If you want, probably you can reuse helper class `AggregationTestBuilder` below. Does line 794 throw exception in the case of null Project?

dai-chen (approved):
Thanks for the changes!

docs/user/ppl/cmd/stats.rst
ahkcs: Removed `last` in this example

opensearch/src/test/java/org/opensearch/sql/opensearch/request/AggregateAnalyzerTest.java
ahkcs: Thanks for the suggestion! I've refactored both analyze_firstAggregation() and analyze_lastAggregation() tests to use the AggregationTestBuilder helper class.
Regarding line 794 - It would throw a ClassCastException when the aggregate input is a LogicalTableScan instead of a Project. I've fixed this by adding proper type checking

Swiddis (approved):

PR #4224, opened 2025-09-04
Support absolute time range in search command 

### LINKED ISSUE #4135: [FEATURE] Support earliest and latest in search command

Currently, earliest and latest options are supported in conditions ([doc](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/condition.rst#earliest)) and aggregation functions (#4100), but not yet supported in search command. E.g. `search source=emps earliest=-1d`

Support earliest and latest parameter in search command

--

--

yuancu:
### Description
Support absolute time range modifier in search command. 

Examples:

```
source=any earliest='2020-12-10'
source=any earliest='2020-12-10' latest='2025-09-10 13:00:12'
source=any latest=now
```

Queries with earliest and latest time modifiers will be converted to scan followed by filters with the implicit timestamp field `@timestamp`. For example, `source=any earliest='2020-12-10'` is converted to `source=any | where @timestamp >= '2020-12-10'`

### Related Issues
Resolves #4135 

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/SearchCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/SearchCommandIT.java
@@ -8,8 +8,11 @@ package org.opensearch.sql.ppl;
 import static org.opensearch.sql.legacy.TestsConstants.*;
 import static org.opensearch.sql.util.MatcherUtils.columnName;
 import static org.opensearch.sql.util.MatcherUtils.rows;
+import static org.opensearch.sql.util.MatcherUtils.schema;
 import static org.opensearch.sql.util.MatcherUtils.verifyColumn;
 import static org.opensearch.sql.util.MatcherUtils.verifyDataRows;
+import static org.opensearch.sql.util.MatcherUtils.verifyNumOfRows;
+import static org.opensearch.sql.util.MatcherUtils.verifySchema;
 
 import java.io.IOException;
 import org.json.JSONObject;
@@ -24,6 +27,7 @@ public class SearchCommandIT extends PPLIntegTestCase {
     super.init();
     loadIndex(Index.BANK);
     loadIndex(Index.DOG);
+    loadIndex(Index.TIME_TEST_DATA);
   }
 
   @Test
@@ -66,5 +70,30 @@ public class SearchCommandIT extends PPLIntegTestCase {
       assertTrue(e.getMessage().contains("RuntimeException"));
       assertTrue(e.getMessage().contains(SYNTAX_EX_MSG_FRAGMENT));
     }
+  }
+
+  @Test
+  public void testSearchWithEarliest() throws IOException {
+    JSONObject result1 =
+        executeQuery(
+            String.format(
+                "search source=%s earliest='2025-08-01 03:47:41' | fields @timestamp",
+                TEST_INDEX_TIME_DATA));
+    verifySchema(result1, schema("@timestamp", "timestamp"));
+    verifyDataRows(result1, rows("2025-08-01 03:47:41"));
+
+    JSONObject result0 =
+        executeQuery(
+            String.format(
+                "search source=%s earliest='2025-08-01 03:47:42' | fields @timestamp",
+                TEST_INDEX_TIME_DATA));
+    verifyNumOfRows(result0, 0);
+
+    JSONObject result2 =
+        executeQuery(
+            String.format(
+                "search source=%s earliest='2025-08-01 02:00:55' | fields @timestamp",
+                TEST_INDEX_TIME_DATA));
+    verifyDataRows(result2, rows("2025-08-01 02:00:56"), rows("2025-08-01 03:47:41"));
   }
 }
--- a/ppl/src/main/antlr/OpenSearchPPLParser.g4
+++ b/ppl/src/main/antlr/OpenSearchPPLParser.g4
@@ -113,7 +113,7 @@ commandName
    ;
 
 searchCommand
-   : (SEARCH)? (logicalExpression)* fromClause (logicalExpression)*     # searchFrom
+   : (SEARCH)? (timeRangeExpression | logicalExpression)* fromClause (timeRangeExpression | logicalExpression)*     # searchFrom
    ;
 
 describeCommand
@@ -624,6 +624,15 @@ singleFieldRelevanceFunction
 // Field is a list of columns
 multiFieldRelevanceFunction
    : multiFieldRelevanceFunctionName LT_PRTHS (LT_SQR_PRTHS field = relevanceFieldAndWeight (COMMA field = relevanceFieldAndWeight)* RT_SQR_PRTHS COMMA)? query = relevanceQuery (COMMA relevanceArg)* RT_PRTHS
+   ;
+
+timeRangeExpression
+   : (EARLIEST | LATEST) EQUAL timeRangeValue
+   ;
+
+timeRangeValue
+   : NOW
+   | stringLiteral
    ;
 
 // tables
--- a/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
+++ b/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.sql.ppl.parser;
 
+import static org.opensearch.sql.ast.dsl.AstDSL.field;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.*;
 import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.BinaryArithmeticContext;
 import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.BooleanLiteralContext;
@@ -38,6 +39,8 @@ import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.SpanClause
 import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.StatsFunctionCallContext;
 import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.StringLiteralContext;
 import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.TableSourceContext;
+import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.TimeRangeExpressionContext;
+import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.TimeRangeValueContext;
 import static org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.WcFieldExpressionContext;
 
 import com.google.common.collect.ImmutableList;
@@ -88,8 +91,10 @@ import org.opensearch.sql.ast.expression.subquery.ExistsSubquery;
 import org.opensearch.sql.ast.expression.subquery.InSubquery;
 import org.opensearch.sql.ast.expression.subquery.ScalarSubquery;
 import org.opensearch.sql.ast.tree.Trendline;
+import org.opensearch.sql.calcite.plan.OpenSearchConstants;
 import org.opensearch.sql.common.antlr.SyntaxCheckException;
 import org.opensearch.sql.common.utils.StringUtils;
+import org.opensearch.sql.expression.function.BuiltinFunctionName;
 import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser;
 import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.BinaryArithmeticContext;
 import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.BooleanLiteralContext;
@@ -811,5 +816,34 @@ public class AstExpressionBuilder extends OpenSearchPPLParserBaseVisitor<Unresol
   @Override
   public UnresolvedExpression visitLogWithBaseSpan(OpenSearchPPLParser.LogWithBaseSpanContext ctx) {
     return org.opensearch.sql.ast.dsl.AstDSL.stringLiteral(ctx.getText());
+  }
+
+  /** Process time range value expressions (NOW or string literal) */
+  @Override
+  public UnresolvedExpression visitTimeRangeValue(TimeRangeValueContext ctx) {
+    if (ctx.NOW() != null) {
+      // Convert NOW keyword to function call NOW()
+      return buildFunction(BuiltinFunctionName.NOW.name(), List.of());
+    } else if (ctx.stringLiteral() != null) {
+      return visit(ctx.stringLiteral());
+    } else {
+      throw new IllegalArgumentException("Unknown time range value type");
+    }
+  }
+
+  /**
+   * Process time range expressions (EARLIEST='value' or LATEST='value') It creates a Comparison
+   * filter like @timestamp >= timeRangeValue
+   */
+  @Override
+  public UnresolvedExpression visitTimeRangeExpression(TimeRangeExpressionContext ctx) {
+    // EARLIEST != null --> @timestamp >= timeRangeValue
+    String operator =
+        ctx.EARLIEST() != null
+            ? BuiltinFunctionName.GTE.getName().toString()
+            : BuiltinFunctionName.LTE.getName().toString();
+    UnresolvedExpression timestampField = field(OpenSearchConstants.IMPLICIT_FIELD_TIMESTAMP);
+    UnresolvedExpression timeValue = visit(ctx.timeRangeValue());
+    return new Compare(operator, timestampField, timeValue);
   }
 }
--- a/core/src/main/java/org/opensearch/sql/calcite/plan/OpenSearchConstants.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/plan/OpenSearchConstants.java
@@ -20,6 +20,8 @@ public interface OpenSearchConstants {
 
   String METADATA_FIELD_ROUTING = "_routing";
 
+  String IMPLICIT_FIELD_TIMESTAMP = "@timestamp";
+
   java.util.Map<String, ExprType> METADATAFIELD_TYPE_MAP =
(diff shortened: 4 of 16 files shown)


vamsimanohar:
can you review this: https://github.com/opensearch-project/sql/pull/4152
I am changing the grammar for search expression and its functionality.
We might need to reevaluate the implementation.

yuancu:
Hi @vamsimanohar, thank you for reminding! I think the functionality does not overlap, I'll re-implement this based on your grammar.

vamsimanohar:
@yuancu  Few things to keep in mind.

`
We fundamentally want search command to be always push down and translate to `query_string` function in DSL which follows Lucene query syntax.  Anything we add to search command should get translated to lucene query.
`

I think it should be possible with your changes.

core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EarliestFunction.java
```diff
@@ -66,6 +66,6 @@ public static Boolean earliest(Object... inputs) {
     ZonedDateTime earliest =
         getRelativeZonedDateTime(
             expression, ZonedDateTime.ofInstant(clock.instant(), clock.getZone()));
-    return earliest.isBefore(candidateDatetime);
+    return !earliest.isAfter(candidateDatetime);
```
yuancu, line 69: Modified because `earliest` should be *greater or equal*.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -761,6 +762,60 @@ multiFieldRelevanceFunction
...
+   | NOW LT_PRTHS RT_PRTHS
+   | DECIMAL_LITERAL
+   | INTEGER_LITERAL
+   | stringLiteral
+   | timeSnap
+   | (PLUS | MINUS) (integerLiteral)? timeModifierUnit (timeSnap)?
+   ;
+
+timeSnap
+  : AT timeModifierUnit;
```
yuancu: Asking for help: currently, `AT timeModifierUnit` can not be matched. Since antlr is parsed bottom-up, plus that a shorter rule matching longer text will be prioritized, `AT timeModifierUnit` will always be matched as an identifier by the following rule:

```antlr
// OpenSearchPPLLexer.g4
ID:                                 ID_LITERAL;
fragment ID_LITERAL:                ([@*A-Z_])+?[*A-Z_\-0-9]*;
```

```antlr
// OpenSearchPPLParser.g4
ident
   : (DOT)? ID
```

As a result, I can not write something like `source=t earliest=-10h@day`, but have to write it as `source=t earliest='-10h@day'` because `-10h@day` will be matched to the shorter rule `ident`.

I tried to make the grammar context aware, but had no luck with multiple solutions. Is there any suggestion for a workaround?

core/build.gradle
```diff
@@ -74,6 +74,7 @@ dependencies {
     testImplementation group: 'org.hamcrest', name: 'hamcrest-library', version: "${hamcrest_version}"
     testImplementation group: 'org.mockito', name: 'mockito-core', version: "${mockito_version}"
     testImplementation group: 'org.mockito', name: 'mockito-junit-jupiter', version: "${mockito_version}"
+    testImplementation group: 'org.opensearch', name: 'opensearch', version: "${opensearch_version}"
```
yuancu: This dependency is added for the access of `org.opensearch.common.time.DateMathParser` -- I want to make sure that the parsed OpenSearch date math like `now-1d/M-2M` returns the intended instant for `-1d@q`

vamsimanohar:
@yuancu  hey can you update the description with latest implementation where you are translating to Lucene query.

ppl/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -514,13 +515,19 @@ HR:                                 'HR';
 HRS:                                'HRS';
 HOURS:                              'HOURS';
 DAYS:                               'DAYS';
+WEEKS:                              'WEEKS';
```
vamsimanohar: can you add these new terms to searchKeyWords or `keywordCanbeId`

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -122,7 +122,8 @@ searchCommand
    ;
 
 searchExpression
- : LT_PRTHS searchExpression RT_PRTHS                 # groupedExpression
+ : timeModifier                                       # timeModifierExpression
```
vamsimanohar, line 125: For my understanding: This implies we can write only timeModifier in searchExpression.
Can you update search.rst with these changes.

ppl/src/main/antlr/OpenSearchPPLParser.g4
vamsimanohar: @yuancu  have you figured out anything on this?

integ-test/src/test/java/org/opensearch/sql/ppl/SearchCommandIT.java
```diff
@@ -917,4 +921,52 @@ public void testSearchWithDateINOperator() throws IOException {
         rows("2024-01-15 10:30:00.123456789", "INFO"),
         rows("2024-01-15 10:30:01.23456789", "ERROR"));
   }
+
+  @Test
+  public void testSearchWithAbsoluteEarliestAndNow() throws IOException {
+    JSONObject result1 =
+        executeQuery(
+            String.format(
+                "search source=%s earliest='2025-08-01 03:47:41' latest=now | fields @timestamp",
```
penghuo, line 1010: What if earliest is a field name? Should add backtick?

ppl/src/main/antlr/OpenSearchPPLParser.g4
penghuo: +1, please update doc.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -122,7 +122,8 @@ searchCommand
...
+ | LT_PRTHS searchExpression RT_PRTHS                 # groupedExpression
  | NOT searchExpression                               # notExpression
```
penghuo, line 127: do timemodifier support NOT? e.g. `search source=index NOT latest=now`

core/build.gradle
penghuo: parseRelativeTime() is called by PPL module only, right? if move parseRelativeTime to PPL module, does opensearch depedency still needed?

penghuo (changes requested):

ppl/src/main/antlr/OpenSearchPPLParser.g4
yuancu: No, I haven't :/

core/build.gradle
yuancu: Yes, it is only called by PPL module. If I still want to assure the correctness of the parsed instant, I'll have to move this dependency to the PPL module.

Is there any concern over this dependency? If it's for the performance, as the dependency is only test time, I assume there would be no harm to the runtime performance.

ppl/src/main/antlr/OpenSearchPPLLexer.g4
yuancu: Added `timeModifierUnit` to `searchableKeyWord`. I wonder how does this change the parsing behavior

ppl/src/main/antlr/OpenSearchPPLParser.g4
yuancu: Yes. Time modifiers are converted conditions with `SearchComparison`. Any further operation that is applicable to search comparison, such as NOT, OR, AND, are applicable to time modifiers.

integ-test/src/test/java/org/opensearch/sql/ppl/SearchCommandIT.java
yuancu: Yes, they can be accessed with backticks. I have added a test case for such scenarios and updated docs.

ppl/src/main/antlr/OpenSearchPPLParser.g4
yuancu: Thanks for pointing out! The documentation is updated.

ppl/src/main/antlr/OpenSearchPPLParser.g4
yuancu: Update: the problem is solved by creating a new lexer `TIME_SNAP` that matches `@ time_snap_unit` and placing it before `ID`.

docs/user/ppl/cmd/search.rst
```diff
@@ -44,6 +45,35 @@ The search expression syntax supports:
...
+**Examples of Time Modifier Values**:
+
+* ``earliest=now`` - From current time
+* ``latest='2024-12-31 23:59:59'`` - Until a specific date
+* ``earliest=-7d`` - From 7 days ago
+* ``latest='+1d@d'`` - Until tomorrow at start of day
+* ``earliest='-1month@month'`` - From start of previous month
+* ``latest=1754020061`` - Until a unix timestamp (August 1, 2025 03:47:41 at UTC)
+
+Read more details on time modifiers `here <https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/functions/ppl-datetime.md#relative_timestamp>`_.
```
vamsimanohar, line 70: Is `relative_timestamp` correct place, can we move time modifiers to a separate md file?

docs/user/ppl/cmd/search.rst
```diff
@@ -44,6 +45,35 @@ The search expression syntax supports:
 * Unquoted: ``status=active``, ``code=ERR-401``
 * Quoted: ``email="user@example.com"``, ``message="server error"``
 
+**Time Modifiers**: Filter search results by time range using the implicit ``@timestamp`` field. Time modifiers support the same formats as the `EARLIEST and LATEST condition functions <https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/condition.rst#earliest>`_:
+
+1. **Current time**: ``now`` or ``now()`` - the current time
+2. **Absolute format**: ``MM/dd/yyyy:HH:mm:ss`` or ``yyyy-MM-dd HH:mm:ss``
+3. **Unix timestamp**: Numeric values (seconds since epoch) like ``1754020060.123``
+4. **Relative format**: ``(+|-)<time_integer><time_unit>[+<...>]@<snap_unit>`` - Time offset from current time
```
vamsimanohar, line 53: Just for clarification, `snap_unit` will always be start of what is specified and no relation to `now` right?

core/src/main/java/org/opensearch/sql/utils/DateTimeUtils.java
```diff
@@ -170,7 +168,19 @@ public static LocalDate extractDate(ExprValue value, FunctionProperties function
 
   public static ZonedDateTime getRelativeZonedDateTime(String input, ZonedDateTime baseTime) {
     try {
-      Instant parsed = LocalDateTime.parse(input, DIRECT_FORMATTER).toInstant(ZoneOffset.UTC);
+      Instant parsed;
+      try {
+        parsed = LocalDateTime.parse(input, DIRECT_FORMATTER).toInstant(ZoneOffset.UTC);
+      } catch (DateTimeParseException ignored) {
+        try {
+          parsed =
```
vamsimanohar: What is this handling?

ppl/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -514,19 +515,34 @@ HR:                                 'HR';
...
 MON:                                'MON';
 MONTHS:                             'MONTHS';
+QTR:                                'QTR';
+QTRS:                               'QTRS';
+QUARTERS:                           'QUARTERS';
+YR:                                 'YR';
+YRS:                                'YRS';
+YEARS:                              'YEARS';
```
vamsimanohar: You might need to rebase after https://github.com/opensearch-project/sql/pull/4334

docs/user/ppl/cmd/search.rst
vamsimanohar: can I got to start 2months before.

`15m@2mon`?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -892,4 +894,56 @@ public SearchLiteral visitSearchLiteral(OpenSearchPPLParser.SearchLiteralContext
...
+      // OpenSearch time range accepts "now" as a reference to the current time
+      osDateMathExpression = ctx.NOW().getText().toLowerCase(Locale.ROOT);
+    } else {
+      // Process absolute and relative time modifier values
+      String pplTimeModifier =
+          ctx.stringLiteral() != null
+              ? (String) ((Literal) visit(ctx.stringLiteral())).getValue()
+              : ctx.getText().strip();
+      // Parse a PPL time modifier to OpenSearch date math expression
+      osDateMathExpression = DateTimeUtils.resolveTimeModifier(pplTimeModifier);
```
vamsimanohar, line 905: Is this an existing function?

vamsimanohar:
@yuancu  Did we cover every usecase mentioned in the description? Lets try to close this PR by end of the day. I can stay late and get on a call with you for review.

You probably might need to refactor from this PR:https://github.com/opensearch-project/sql/pull/4334

docs/user/ppl/cmd/search.rst
yuancu: >  Just for clarification, snap_unit will always be start of what is specified and no relation to now right?

Span unit will be the start of what is specified. E.g. `@mon` align the time expression before it to the start of that month.

In our implementation for week and quarter, it has some connection with the current time though. Because in opensearch datemath, there is no alignment to quarter, we have to manually calculate how many months to shift based on when is it now. For week, because the start of week is different in PPL and OpenSearch is difference, a now reference is also necessary to convert the week day shift. Please refer to `core/src/test/java/org/opensearch/sql/utils/DateTimeUtilsTest.java` for more details.

> can I got to start 2months before. `15m@2mon`?

Nope, you'll have to use something like  `-15m-2month`, or `-15m-2month@month` if you want the alignment.

core/src/main/java/org/opensearch/sql/utils/DateTimeUtils.java
yuancu: This is to make the existing [earliest and latest condition functions](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/condition.rst#earliest) also support formats like `YYYY-MM-DD HH:mm:ss` and `YYYY-MM-DDTHH:mm:ssZ`, it used to support only one format `MM/dd/yyyy:HH:mm:ss`, which is rather alien to PPL, compared with other date time functions.

ppl/src/main/antlr/OpenSearchPPLLexer.g4
yuancu: Thanks, removed

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
yuancu: Nope, it is implemented from scratch in this PR.

docs/user/ppl/cmd/search.rst
yuancu: Actually, this points to a documentation of opensearch Spark project. I did so because the [earliest condition function](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/condition.rst#earliest) also points to this doc. I thought it's not bad to have a single source of truth. Or do you think it would be better if we have a replicate of it in our repo in a separate file.

yuancu:
> @yuancu Did we cover every usecase mentioned in the description? Lets try to close this PR by end of the day. I can stay late and get on a call with you for review.
> 
> You probably might need to refactor from this PR:#4334

Unit test and explaining IT covers all of them. I did not cover some relative time modifier in integration test because they either have something to do with the current time, which may be flaky to mock and test. I'll try to cover more of them in the integration test.  E.g. there exists unknown latency after I write a timestamp and before the query is executed.

core/src/main/java/org/opensearch/sql/utils/DateTimeUtils.java
vamsimanohar: can we loop through formatters instead of deep nested. It will be easy to add more in future.

yuancu: Yes that would be better. Thanks for the suggestion

ppl/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -517,6 +518,16 @@ NUMERIC_ID : DEC_DIGIT+ ID_LITERAL;
 
 // LITERALS AND VALUES
 //STRING_LITERAL:                     DQUOTA_STRING | SQUOTA_STRING | BQUOTA_STRING;
+fragment WEEK_SNAP_UNIT:            'W' [0-7];
+fragment TIME_SNAP_UNIT:              'S' | 'SEC' | 'SECOND'
```
vamsimanohar, line 522: Dont we want to support plural like SECONDS, MINUTES  like SPANLENGTH

yuancu: Nope, time snap does not support plural forms.

core/src/main/java/org/opensearch/sql/utils/DateTimeUtils.java
yuancu: Fixed

core/build.gradle
vamsimanohar: we want core modules to not depend on any opensearch dependencies. This helps in our future plans to publish ppl as a library or running sql engine as a separate service.

If we can avoid, its better to avoid.

yuancu: Thanks for the explanation! I've removed the dependency and related tests.

vamsimanohar (approved):

Issue #3961, opened 2025-07-31
[FEATURE] Support Limit pushdown through aggregation.


penghuo:
**Is your feature request related to a problem?**
```
  @Test
  public void testFilterAndAggPushDownExplain() throws IOException {
    String expected = loadExpectedPlan("explain_filter_agg_push.json");
    assertJsonEqualsIgnoreId(
        expected,
        explainQueryToString(
            "source=opensearch-sql_test_index_account"
                + "| where age > 30 "
                + "| stats avg(age) AS avg_age by state, city"));
  }
```
* Output
```
{
  "calcite": {
    "logical": "LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n  LogicalProject(avg_age=[$2], state=[$0], city=[$1])\n    LogicalAggregate(group=[{0, 1}], avg_age=[AVG($2)])\n      LogicalProject(state=[$7], city=[$5], age=[$8])\n        LogicalFilter(condition=[>($8, 30)])\n          CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n",
[...]

PR #4228, opened 2025-09-04
Push down limit operator into aggregation bucket size

qianheng-aws:
### Description
Push down limit operator into aggregation bucket size. Also adjusting the digest of limit push down to include `offset` information as well.

### Related Issues
Resolves  https://github.com/opensearch-project/sql/issues/3961

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -291,15 +292,27 @@ public CalciteLogicalIndexScan pushDownAggregate(Aggregate aggregate, Project pr
...
     try {
-      CalciteLogicalIndexScan newScan = this.copyWithNewSchema(getRowType());
-      newScan.pushDownContext.add(
-          PushDownAction.of(
-              PushDownType.LIMIT,
-              limit,
-              requestBuilder -> requestBuilder.pushDownLimit(limit, offset)));
-      return newScan;
+      if (pushDownContext.isAggregatePushed()) {
+        // Push down the sort into the aggregation bucket
```
penghuo: Is it accurate? should be push down the limit into the aggregation bucket?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -291,15 +292,27 @@ public CalciteLogicalIndexScan pushDownAggregate(Aggregate aggregate, Project pr
...
+        boolean pushed =
+            pushDownContext.getAggPushDownAction().pushDownLimitIntoBucketSize(limit + offset);
+        if (!pushed && offset > 0) return null;
```
penghuo, line 301: if hasAggregatorInSortBy, limit should not been pushed
`source=log00001 | stats sum(age) as age by vi | sort age | head 1`, limit 1 should not pushed to composite-aggregation.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
qianheng-aws: It's a little bit misleading. It's actually pushed down into the size of composite aggregation. Will change to that saying.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
qianheng-aws: In the DSL of composite aggregation, the sort operation is executed previous to size cutting.  That's exactly the execution sequence we need for this PPL.  

See the test case in `StatsCommandIT::testStatsWithLimit`, if we add test case like:
```
source=%s | stats avg(balance) as a by age | sort - age | head 5
```
The final results is:
```
...
    "datarows": [
        [
            null,
            36
        ],
        [
            48086,
            34
        ],
        [
            4180,
            33
        ],
        [
            39225,
            32
        ],
        [
            32838,
            28
        ]
    ],
    "size": 5
}
```
It meets our expectation.


What's debatable is the case of PPL:
```
source=%s | stats avg(balance) as a by age | head 5 | sort - age
```
I think it's both OK whether push down sort after limit since we don't ensure a order for limit operator. So the output of limit operator is not ensured `idempotent `. But since we already has restriction in `OpenSearchSortIndexScanRule` to prevent push down sort after limit, the behavior can be align.

qianheng-aws: > source=log00001 | stats sum(age) as age by vi | sort age | head 1

This query will combine sort + limit into a single `LogicalSort` operator, and it will cause an issue by push down it as a limit directly.

This is actually an robust bug already exist in `OpenSearchLimitIndexScanRule`. We need to check the collation before pushing down the `LogicalSort` as a limit.

Add tests for verifying the data and plan separately in the latest [commit](https://github.com/opensearch-project/sql/pull/4228/commits/1835f81e99ca4a4188279fd4bd6ebce4b76b893a).

LantaoJin:
I saw some failures after this PR merged. ref https://github.com/opensearch-project/sql/actions/runs/17625084933/job/50079822319 @qianheng-aws

Issue #4200, opened 2025-09-02
[FEATURE] Add support for `median(<value>)`


aalva500-prog:
**Is your feature request related to a problem?**
Yes. PPL lacks SPL’s `median` function `median(<field>)`.

`median(<field>)` and `percentile50(<field>)` are functionally equivalent - they both return the 50th percentile value of the specified field.

However, there are some subtle differences:

- `median()` is specifically designed for finding the middle value and may have slight performance optimizations for this exact use case.
- `percentile50()` is part of the more general percentile function family, which can calculate any percentile (1-99)

In practice, both will give you the same result for the 50th percentile calculation. The choice between them is often a matter of code readability and intent - `median()` makes it immediately clear you want the median value, while `percentile50()` shows you're working with percentile calculations.

**What solution would you like?**
Support this shortcut and rewrite it to `perc50(<value>)`, which is already supported with pushdown optimization.

**What alternatives have you considered?**
N/A

**Do you have any additional context?**
N/A

RyanL1997:
#4234 merged

PR #4234, opened 2025-09-05
Add support for `median(<value>)`

aalva500-prog:
### Description
PPL lacks `median` function `median(<field>)` in stats aggregation functions.

This PR implements the support of the `median` shortcut and rewrite it to `perc50(<value>)`, which is already supported with pushdown optimization.

`median(<field>)` and `percentile50(<field>)` are functionally equivalent - they both return the 50th percentile value of the specified field.

However, there are some subtle differences:

- `median()` is specifically designed for finding the middle value and may have slight performance optimizations for this exact use case.
- percentile50() is part of the more general percentile function family, which can calculate any percentile (1-99)

In practice, both will give you the same result for the 50th percentile calculation. The choice between them is often a matter of code readability and intent - median() makes it immediately clear you want the median value, while percentile50() shows you're working with percentile calculations.

### Related Issues
Resolves #[4200]<https://github.com/opensearch-project/sql/issues/4200>

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
@@ -123,6 +123,7 @@ import static org.opensearch.sql.expression.function.BuiltinFunctionName.MATCH_P
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.MATCH_PHRASE_PREFIX;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.MAX;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.MD5;
+import static org.opensearch.sql.expression.function.BuiltinFunctionName.MEDIAN;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.MICROSECOND;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.MIN;
 import static org.opensearch.sql.expression.function.BuiltinFunctionName.MINUTE;
@@ -253,6 +254,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.sql.calcite.CalcitePlanContext;
 import org.opensearch.sql.calcite.utils.OpenSearchTypeFactory;
+import org.opensearch.sql.calcite.utils.PPLOperandTypes;
 import org.opensearch.sql.calcite.utils.PlanUtils;
 import org.opensearch.sql.calcite.utils.UserDefinedFunctionUtils;
 import org.opensearch.sql.exception.ExpressionEvaluationException;
@@ -1115,6 +1117,24 @@ public class PPLFuncImpTable {
               extractTypeCheckerFromUDF(PPLBuiltinOperators.PERCENTILE_APPROX),
               PERCENTILE_APPROX.name(),
               false));
+
+      // Register MEDIAN as alias for PERCENTILE_APPROX with 50th percentile
+      register(
+          MEDIAN,
+          (distinct, field, argList, ctx) -> {
+            List<RexNode> medianArgList = new ArrayList<>();
+            medianArgList.add(
+                ctx.rexBuilder.makeExactLiteral(
+                    BigDecimal.valueOf(50.0))); // hardcoded 50th percentile
+            medianArgList.add(ctx.rexBuilder.makeFlag(field.getType().getSqlTypeName()));
+            return UserDefinedFunctionUtils.makeAggregateCall(
+                PPLBuiltinOperators.PERCENTILE_APPROX,
+                List.of(field),
+                medianArgList,
+                ctx.relBuilder);
+          },
+          wrapSqlOperandTypeChecker(
+              PPLOperandTypes.NUMERIC.getInnerTypeChecker(), MEDIAN.name(), false));
 
       register(
           EARLIEST,
--- a/core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
@@ -198,6 +198,7 @@ public enum BuiltinFunctionName {
   TAKE(FunctionName.of("take")),
   // t-digest percentile which is used in OpenSearch core by default.
   PERCENTILE_APPROX(FunctionName.of("percentile_approx")),
+  MEDIAN(FunctionName.of("median")),
   EARLIEST(FunctionName.of("earliest")),
   LATEST(FunctionName.of("latest")),
   DISTINCT_COUNT_APPROX(FunctionName.of("distinct_count_approx")),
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1118,6 +1120,21 @@ void populate() {
               PERCENTILE_APPROX.name(),
               false));
 
+      register(
```
dai-chen, line 1127: I'm thinking what's the pros and cons between registering a new function vs. rewriting in AST layer?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1118,6 +1120,21 @@ void populate() {
...
+          MEDIAN,
+          (distinct, field, argList, ctx) -> {
+            List<RexNode> medianArgList = new ArrayList<>();
+            medianArgList.add(ctx.rexBuilder.makeExactLiteral(BigDecimal.valueOf(50.0)));
+            medianArgList.add(ctx.rexBuilder.makeFlag(field.getType().getSqlTypeName()));
```
Swiddis: nit: `List.of`?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1118,6 +1120,21 @@ void populate() {
               PERCENTILE_APPROX.name(),
               false));
 
+      register(
+          MEDIAN,
+          (distinct, field, argList, ctx) -> {
+            List<RexNode> medianArgList = new ArrayList<>();
+            medianArgList.add(ctx.rexBuilder.makeExactLiteral(BigDecimal.valueOf(50.0)));
```
RyanL1997: nit: Consider extracting 50.0 as a constant.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
RyanL1997: Should there be `null` checking for `field.getType()`? What happens if the field type is `null`?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1118,6 +1120,21 @@ void populate() {
               PERCENTILE_APPROX.name(),
               false));
 
+      register(
+          MEDIAN,
+          (distinct, field, argList, ctx) -> {
```
RyanL1997, line 1129: Question: The `distinct` and `argList` parameters are not used. Should there be validation that:

- `distinct` is false (since `MEDIAN` doesn't support `DISTINCT`)?
- `argList` is empty (since `MEDIAN` takes no additional arguments)?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
RyanL1997: Also, just for my knowledge: why MEDIAN is implemented as `PERCENTILE_APPROX(field, 50.0)`?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
aalva500-prog: Hi @dai-chen, please find below why I chose registering a new function instead of rewriting it in AST layer:

- MEDIAN is a legitimate aggregate function, not syntax sugar. Function registration keeps it in the proper semantic layer where aggregate functions belong, maintaining clean separation between function logic and query planning infrastructure.

- Consistency: All other aggregate functions (AVG, COUNT, SUM) use function registration. IMO, MEDIAN should follow the same pattern for code consistency.

- Function registration provides automatic type checking, error handling, and validation through the existing PPLTypeChecker framework. AST rewriting would require manually implementing these safeguards or risk runtime failures.

- Maintainability: All functions logic stays in one location, so changes, bug fixes, or enhancements are contained and don't require touching multiple planning components. AST rewriting scatters the logic across visitors, rules, and planners.

I understand AST rewriting could be preferred when performance optimization is critical, however, I don't think in this specific case there is a great difference, but I could be wrong. That being said, please let me know if you think the current implementation can cause a significant performance bottleneck and I should be using AST rewriting approach instead. Thanks!

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
aalva500-prog: Good catch! If `field.getType()` returns `null`, calling `getSqlTypeName()` on it would throw a `NullPointerException`. This affects both `MEDIAN` and `PERCENTILE_APPROX` functions, btw.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
aalva500-prog: Good suggestion! Let me extract the 50.0 value as a constant.

To answer your question, `MEDIAN` is implemented as `PERCENTILE_APPROX(field, 50.0)` because:

1) Mathematical equivalence: The `median` is by definition the `50th percentile` - the value that splits the data in half.

2) Code reuse: Instead of implementing a separate median algorithm, I'm leveraging the existing `PERCENTILE_APPROX` function which already handles:

- Approximate percentile calculations (efficient for large datasets)
- Type handling and validation
- Integration with the underlying execution engine

3) Performance: PERCENTILE_APPROX uses an approximate algorithm (like t-digest) that's much faster than exact median calculation, especially for large datasets where you'd need to sort all values.

4) Consistency: Both functions will behave the same way in terms of approximation accuracy and performance characteristics. Adding to this point, if you look into the `BuiltinFunctionName` class, the mapping for the `percentile` function also uses the `PERCENTILE_APPROX` function:

```
          .put("percentile", BuiltinFunctionName.PERCENTILE_APPROX)
```

In general, the "APPROX" part means it's trading some precision for speed - perfectly acceptable for most analytical use cases where you need fast results on large datasets rather than exact values.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
aalva500-prog: You're absolutely right, the `distinct` and `argList` parameters are not used in the implementation, and there should be validation to ensure they meet the expected constraints for MEDIAN. Let me add the necessary validation, thanks!

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
aalva500-prog: Good catch! I'll simplify to use `List.of()`.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
dai-chen: I think the main benefit of current approach is exposing the same function set in SQL. We can discuss more later. Thanks!

PR #4241, opened 2025-09-08
[Feature] Implementation of mode sed and offset_field in rex PPL command

RyanL1997:
### Description
Implementation of mode sed and offset_field in rex PPL command

### Related Issues
* Related #4108 
* Related #4109 

### Check List


ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -287,6 +287,8 @@ rexExpr
 rexOption
     : MAX_MATCH EQUAL maxMatch=integerLiteral
     | MODE EQUAL EXTRACT
+    | MODE EQUAL SED
```
dai-chen: np: `mode=(EXTRACT | SED)` ?

docs/user/ppl/cmd/rex.rst
```diff
@@ -19,11 +19,23 @@ Version
...
+  - **sed mode**: Performs text substitution on the field using sed-style patterns:
+
+    - ``s/pattern/replacement/`` - Replace first occurrence
+    - ``s/pattern/replacement/g`` - Replace all occurrences (global)
+    - ``s/pattern/replacement/n`` - Replace only the nth occurrence (where n is a number)
+    - ``y/from_chars/to_chars/`` - Character-by-character transliteration
+    - Backreferences: ``\1``, ``\2``, etc. reference captured groups in replacement
+
 * max_match: optional integer (default=1). Maximum number of matches to extract. If greater than 1, extracted fields become arrays. The value 0 means unlimited matches, but is automatically capped to the configured limit (default: 10, configurable via ``plugins.ppl.rex.max_match.limit``).
+* offset_field: optional string. Field name to store the character offset positions of matches.
```
dai-chen: This option only works for extract mode or both?

docs/user/ppl/cmd/rex.rst
```diff
@@ -77,7 +89,41 @@ PPL query::
...
+
+PPL query::
+
+    os> source=accounts | rex field=email "(?<username>[^@]+)@(?<domain>[^.]+)" offset_field=matchpos | fields email, username, domain, matchpos | head 2 ;
+    fetched rows / total rows = 2/2
+    +-----------------------+------------+--------+---------------------------+
+    | email                 | username   | domain | matchpos                  |
+    |-----------------------+------------+--------+---------------------------|
+    | amberduke@pyrami.com  | amberduke  | pyrami | domain=10-15&username=0-8 |
+    | hattiebond@netagy.com | hattiebond | netagy | domain=11-16&username=0-9 |
```
dai-chen, line 122: `domain` is second capture group in regex but comes first here? Is it by design?

core/src/main/java/org/opensearch/sql/expression/function/udf/RexOffsetFunction.java
```diff
@@ -0,0 +1,92 @@
...
+            offsetPairs.add(groupName + "=" + start + "-" + (end - 1));
+          }
+        }
+        groupIndex++;
+      }
+
+      java.util.Collections.reverse(offsetPairs);
+      return offsetPairs.isEmpty() ? null : String.join("&", offsetPairs);
+    } catch (Exception e) {
+      return null;
```
dai-chen: Probably capture the specific exception type? So we return null if any error?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2108,4 +2126,118 @@ private void buildExpandRelNode(
       context.relBuilder.rename(names);
     }
   }
+
+  /**
+   * Creates an optimized sed call using native Calcite functions where possible. Falls back to
+   * custom REX_SED for complex cases.
+   */
```
dai-chen: Could you elaborate this design in PR description? Also where is the REX_SED and fallback logic? I may miss it somewhere

RyanL1997: Good catch, I should actually remove this comment, and here is the reason:

In the previous design version I developed, I created an UDF called `REX_SED`, however, I have completely removed these logic, and start using all the native calcite sql functions for the entire SED feature. 
- Here is the original comment I discussed with @penghuo : https://github.com/opensearch-project/sql/pull/4109#discussion_r2305922467
- Here is the related investigation I did in my original RFC for the usage of UDF: https://github.com/opensearch-project/sql/issues/4108#issuecomment-3231105679

But yes, let me transfer some of the above info the PR description and remove this misleading comment.

RyanL1997: fixed.

ppl/src/main/antlr/OpenSearchPPLParser.g4
RyanL1997: fixed.

docs/user/ppl/cmd/rex.rst
RyanL1997: Good observation and yes, this behavior aligns with the requirement.

docs/user/ppl/cmd/rex.rst
RyanL1997: This is a very good question. I have rechecked the requirement also, and confimed`offset_field` should be only working on the extraction mode. I added an clear exception msg:

```bash
❯   curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d'{
      "query": "source=accounts | rex field=email mode=sed offset_field=email \"s/@.*/@newcompany.com/\" | fields firstname, email | head 5"
    }' | jq

{
  "error": {
    "reason": "Invalid Query",
    "details": "Rex command: offset_field cannot be used with mode=sed. The offset_field option is only supported in extract mode.",
  },
  "status": 400
}
```

And also I have added a test, and updated the doc.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: Btw I have also updated the PR description with a section of "Design Details".

core/src/main/java/org/opensearch/sql/expression/function/udf/RexOffsetFunction.java
RyanL1997: Good point. I checked the logic, exceptions can happen should be `PatternSyntaxException` - When regex pattern compilation fails.

So that I have also updated the above logic, and improve the better error handling instead of silently returning `null`. Here is an example:

```bash
curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d'{
      "query": "source=accounts | rex field=email \"(?<user>[^@)+@(?<domain>.+)\" offset_field=pos | fields email, pos | head 1"
  }'

{
  "error": {
    "reason": "Invalid Query",
    "details": "Invalid regex pattern in rex command: Unclosed character class near index 26\n(?<user>[^@)+@(?<domain>.+)",
  },
  "status": 400
}% 
```
And this also aligns with the requirement.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2136,4 +2154,115 @@ private void buildExpandRelNode(
       context.relBuilder.rename(names);
     }
   }
+
+  /** Creates an optimized sed call using native Calcite functions */
+  private RexNode createOptimizedSedCall(
```
dai-chen, line 2159: np: If sed/regex is common across different commands, it may worth creating dedicated AST expression nodes and separate these into its own visit method in the future.

docs/user/ppl/cmd/rex.rst
RyanL1997: Actually, checked with @dai-chen offline, it should be ordered alphabetically :/
I have changed into something like the following.

```bash
curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d'{
      "query": "source=accounts | rex field=email \"(?<user>[a-zA-Z0-9._%+-]+)@(?<domain>[a-zA-Z0-9.-]+)\\.(?<tld>[a-zA-Z]{2,})\" offset_field=positions | fields email, user, domain, tld, positions | head 3"
  }' | jq

{
  "schema": [
    {
      "name": "email",
      "type": "string"
    },
    {
      "name": "user",
      "type": "string"
    },
    {
      "name": "domain",
      "type": "string"
    },
    {
      "name": "tld",
      "type": "string"
    },
    {
      "name": "positions",
      "type": "string"
    }
  ],
  "datarows": [
    [
      "amberduke@pyrami.com",
      "amberduke",
      "pyrami",
      "com",
      "domain=10-15&tld=17-19&user=0-8"
    ],
    [
      "hattiebond@netagy.com",
      "hattiebond",
      "netagy",
      "com",
      "domain=11-16&tld=18-20&user=0-9"
    ],
    [
      "nanettebates@quility.com",
      "nanettebates",
      "quility",
      "com",
      "domain=13-19&tld=21-23&user=0-11"
    ]
  ],
  "total": 3,
  "size": 3
}
```

dai-chen (approved):

Issue #3975, opened 2025-08-04
[RFC] Add replace Command to PPL (Calcite Engine)


manasvinibs:
### Problem Statement:

Currently, OpenSearch PPL lacks a direct string replacement capability, requiring users to use complex workarounds or multiple commands to achieve simple string replacements. This limitation impacts data processing efficiency and user experience when performing text transformations.

### Current State
Currently, OpenSearch PPL does not have a direct "replace" command in its standard command set. However, we can achieve string replacement functionality using the eval command with string functions. 

```
source=myindex
| eval new_field = replace(field_name, 'text_to_find', 'text_to_replace')
```

### Proposed Goal
Implement a PPL 'replace' command that efficiently transforms field values in search results through string replacement operations, while maintaining data integrity and optimal performance in OpenSearch.

#### In Scope:

1. Replace field values in search results with specified values
   - Support single and multiple field replacements
   - Handle case-sensitive and case-insensitive replacements
   - Process wildcard patterns
   - Support multiple replacement pairs

2. Field Processing
   - Replace values in original fields only
   - Skip generated fields (from stats/eval)
   - Support default behavior for unspecified fields

3. Functionality
   - String value replacement
   - Pattern-based matching
   - Field-specific operations
   - Streaming data processing

 #### Out of Scope:
  - Replacement in generated fields (stats/eval)
  - Regular expression support (initial phase)
  - Dynamic value generation
  - Complex transformations

#### Basic syntax:

`source=index | replace <string_to_replace> WITH <replacement_string>`

#### With field specification
`
source=index | replace <string_to_replace> WITH <replacement_string> IN field1, field2
`

#### Multiple replacements

```
source=index | replace 
    <string1> WITH <replacement1>
    <string2> WITH <replacement2> 
    IN field1, field2
```

#### Examples:
```

# 1. Basic String Replacement
source=logs 
| replace "error" WITH "ERROR"

# 2. Field-Specific Replacement
source=logs 
| replace "error" WITH "ERROR" IN message, log_level

# 3. Multiple Replacements
source=logs 
| replace 
    "error" WITH "ERROR"
    "warning" WITH "WARNING"
    "info" WITH "INFO"

# 4. Case Sensitivity Option
source=logs 
| replace CASE-SENSITIVE "Error" WITH "ERROR"

# 5. Wildcard Support
source=logs 
| replace *error* WITH "ERROR"

# 6. Multiple Field Support
source=logs 
[...]

PR #4248, opened 2025-09-08
Add replace command with Calcite

manasvinibs:
### Description
Implement the replace command in PPL to replace text patterns in specified fields. This PR includes the grammar implementation and basic replacement functionality. This implementation reuses the existing replace functionality. Missing/new features for regex support is added in a separate PR.

### Syntax
`<source> | replace '<pattern>' WITH '<replacement>' IN field_list`

pattern: The text pattern to search for (case-sensitive)
replacement: The text to replace matches with
field_list: Comma-separated list of fields to perform replacement in
Replace creates new fields with prefix 'new_' containing the replaced text, while preserving original fields.

#### Expected Behavior:

- Action: Creates new fields with replaced text values
- Scope: Operates only on specified fields
- Data Preservation: Original fields remain unchanged
- Case Sensitivity: Text literal matching is case-sensitive
- Pattern Type: Only supports literal string patterns (wildcards or regex will be added in a separate PR)

#### Implementation Approach:

- Creates new fields with 'new_' prefix for each specified field
- Performs literal string replacement in the specified fields
- Maintains original field values
- Validates field existence and pattern/replacement values

### Example Queries

```
-- Replace in single field
source=logs | replace 'error' WITH 'ERROR' IN message

-- Replace in multiple fields 
source=logs | replace 'USA' WITH 'United States' IN country, state

-- Replace with other commands
source=logs | where level='error' | replace 'error' WITH 'ERROR' IN message | sort @timestamp

-- Replace and select specific fields
source=logs | replace 'error' WITH 'ERROR' IN message | fields message, new_message
```

#### Output Schema
For each field specified in IN clause, a new field is created:

- Original field: remains unchanged
- New field: prefixed with 'new_' containing replaced text

#### Example:

Input fields: message="error occurred", level="error"
After replace 'error' WITH 'ERROR' IN message, level:
- message: "error occurred"
- new_message: "ERROR occurred"
- level: "error"
- new_level: "ERROR"

### Related Issues

https://github.com/opensearch-project/sql/issues/3975

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
+++ b/core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
@@ -84,6 +84,7 @@ import org.opensearch.sql.ast.tree.Regex;
 import org.opensearch.sql.ast.tree.Relation;
 import org.opensearch.sql.ast.tree.RelationSubquery;
 import org.opensearch.sql.ast.tree.Rename;
+import org.opensearch.sql.ast.tree.Replace;
 import org.opensearch.sql.ast.tree.Reverse;
 import org.opensearch.sql.ast.tree.Rex;
 import org.opensearch.sql.ast.tree.Search;
@@ -786,6 +787,11 @@ public class Analyzer extends AbstractNodeVisitor<LogicalPlan, AnalysisContext> 
   @Override
   public LogicalPlan visitCloseCursor(CloseCursor closeCursor, AnalysisContext context) {
     return new LogicalCloseCursor(closeCursor.getChild().get(0).accept(this, context));
+  }
+
+  @Override
+  public LogicalPlan visitReplace(Replace node, AnalysisContext context) {
+    throw getOnlyForCalciteException("Replace");
   }
 
   @Override
--- a/core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
+++ b/core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.ast.tree;
+
+import com.google.common.collect.ImmutableList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+import org.opensearch.sql.ast.AbstractNodeVisitor;
+import org.opensearch.sql.ast.expression.DataType;
+import org.opensearch.sql.ast.expression.Field;
+import org.opensearch.sql.ast.expression.Literal;
+import org.opensearch.sql.ast.expression.UnresolvedExpression;
+
+@Getter
+@Setter
+@ToString
+@EqualsAndHashCode(callSuper = false)
+public class Replace extends UnresolvedPlan {
+  private final UnresolvedExpression pattern;
+  private final UnresolvedExpression replacement;
+  private final List<Field> fieldList;
+  private UnresolvedPlan child;
+
+  public Replace(
+      UnresolvedExpression pattern, UnresolvedExpression replacement, List<Field> fieldList) {
+    this.pattern = pattern;
+    this.replacement = replacement;
+    this.fieldList = fieldList;
+    validate();
+  }
+
+  public void validate() {
+    if (pattern == null) {
+      throw new IllegalArgumentException("Pattern expression cannot be null in Replace command");
+    }
+    if (replacement == null) {
+      throw new IllegalArgumentException(
+          "Replacement expression cannot be null in Replace command");
+    }
+
+    // Validate pattern is a string literal
+    if (!(pattern instanceof Literal && ((Literal) pattern).getType() == DataType.STRING)) {
+      throw new IllegalArgumentException("Pattern must be a string literal in Replace command");
+    }
+
+    // Validate replacement is a string literal
+    if (!(replacement instanceof Literal && ((Literal) replacement).getType() == DataType.STRING)) {
+      throw new IllegalArgumentException("Replacement must be a string literal in Replace command");
+    }
+
+    if (fieldList == null || fieldList.isEmpty()) {
+      throw new IllegalArgumentException(
+          "Field list cannot be empty in Replace command. Use IN clause to specify the field.");
+    }
+
+    Set<String> uniqueFields = new HashSet<>();
+    List<String> duplicates =
+        fieldList.stream()
+            .map(field -> field.getField().toString())
+            .filter(fieldName -> !uniqueFields.add(fieldName))
+            .collect(Collectors.toList());
+
+    if (!duplicates.isEmpty()) {
+      throw new IllegalArgumentException(
+          String.format("Duplicate fields [%s] in Replace command", String.join(", ", duplicates)));
+    }
+  }
+
+  @Override
+  public Replace attach(UnresolvedPlan child) {
+    if (null == this.child) {
+      this.child = child;
+    } else {
+      this.child.attach(child);
+    }
+    return this;
+  }
+
+  @Override
+  public List<UnresolvedPlan> getChild() {
+    return this.child == null ? ImmutableList.of() : ImmutableList.of(this.child);
+  }
+
+  @Override
+  public <T, C> T accept(AbstractNodeVisitor<T, C> nodeVisitor, C context) {
+    return nodeVisitor.visitReplace(this, context);
+  }
+}
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -112,6 +112,7 @@ import org.opensearch.sql.ast.tree.RareTopN;
 import org.opensearch.sql.ast.tree.Regex;
 import org.opensearch.sql.ast.tree.Relation;
 import org.opensearch.sql.ast.tree.Rename;
+import org.opensearch.sql.ast.tree.Replace;
 import org.opensearch.sql.ast.tree.Rex;
 import org.opensearch.sql.ast.tree.SPath;
 import org.opensearch.sql.ast.tree.Search;
@@ -144,6 +145,7 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
 
   private final CalciteRexNodeVisitor rexVisitor;
   private final CalciteAggCallVisitor aggVisitor;
+  private static final String NEW_FIELD_PREFIX = "new_";
 
   public CalciteRelNodeVisitor() {
     this.rexVisitor = new CalciteRexNodeVisitor(this);
@@ -2178,6 +2180,40 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     } else {
       throw new CalciteUnsupportedException("Explicit values node is unsupported in Calcite");
     }
+  }
+
+  @Override
+  public RelNode visitReplace(Replace node, CalcitePlanContext context) {
+    visitChildren(node, context);
+
+    List<String> fieldNames = context.relBuilder.peek().getRowType().getFieldNames();
+    RexNode patternNode = rexVisitor.analyze(node.getPattern(), context);
+    RexNode replacementNode = rexVisitor.analyze(node.getReplacement(), context);
+
+    List<RexNode> projectList = new ArrayList<>();
+    List<String> newFieldNames = new ArrayList<>();
+
+    // First add all original fields
+    for (String fieldName : fieldNames) {
+      RexNode fieldRef = context.relBuilder.field(fieldName);
+      projectList.add(fieldRef);
+      newFieldNames.add(fieldName);
+    }
+
+    // Then add new fields with replaced content using new_ prefix
+    for (Field field : node.getFieldList()) {
+      String fieldName = field.getField().toString();
+      RexNode fieldRef = context.relBuilder.field(fieldName);
+
+      RexNode replaceCall =
+          context.relBuilder.call(
+              SqlStdOperatorTable.REPLACE, fieldRef, patternNode, replacementNode);
+      projectList.add(replaceCall);
+      newFieldNames.add(NEW_FIELD_PREFIX + fieldName);
+    }
+
+    context.relBuilder.project(projectList, newFieldNames);
+    return context.relBuilder.peek();
   }
 
   private void buildParseRelNode(Parse node, CalcitePlanContext context) {
--- a/docs/user/ppl/cmd/replace.rst
+++ b/docs/user/ppl/cmd/replace.rst
@@ -1,0 +1,107 @@
+=============
+replace
+=============
+
+.. rubric:: Table of contents
+
+.. contents::
+ :local:
+ :depth: 2
+
+
+Description
+============
+| Using ``replace`` command to replace text in one or more fields in the search result.
+* The command creates new fields with *new_* prefix for replaced content (e.g., replacing text in 'country' creates 'new_country')
+* If a field with *new_* prefix already exists (e.g., 'new_country'), a number will be appended to create a unique field name (e.g., 'new_country0')
+
+
+Version
+=======
+3.3.0
+
+
+Syntax
+============
+replace '<pattern>' WITH '<replacement>' IN <field-name>[, <field-name>]...
+
+Note: This command is only available when Calcite engine is enabled.
+
+* pattern: mandatory. The text pattern you want to replace. Currently supports only plain text literals (no wildcards or regular expressions).
+* replacement: mandatory. The text you want to replace with.
+* field list: mandatory. One or more field names where the replacement should occur.
+
+
+Example 1: Replace text in one field
+====================================
+
+The example shows replacing text in one field.
+
+PPL query::
+
+ os> source=accounts | replace "IL" WITH "Illinois" IN state | fields state, new_state;
+ fetched rows / total rows = 4/4
+ +-------+-----------+
+ | state | new_state |
+ |-------+-----------|
+ | IL    | Illinois  |
+ | TN    | TN        |
+ | VA    | VA        |
+ | MD    | MD        |
+ +-------+-----------+
+
+
+Example 2: Replace text in multiple fields
+==========================================
+
+The example shows replacing text in multiple fields.
+
+PPL query::
+
+ os> source=accounts | replace "IL" WITH "Illinois" IN state, address | fields state, address, new_state, new_address;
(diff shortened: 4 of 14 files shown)


core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -700,6 +701,12 @@ public LogicalPlan visitCloseCursor(CloseCursor closeCursor, AnalysisContext con
     return new LogicalCloseCursor(closeCursor.getChild().get(0).accept(this, context));
   }
 
+  @Override
+  public LogicalPlan visitReplace(Replace node, AnalysisContext context) {
+    throw new UnsupportedOperationException(
+        "Replace is supported only when " + CALCITE_ENGINE_ENABLED.getKeyValue() + "=true");
```
Swiddis: There's a util for this nowadays: https://github.com/opensearch-project/sql/blob/c1866861518427b2d7b0714baa51dcbb4b6e21e7/core/src/main/java/org/opensearch/sql/calcite/utils/CalciteUtils.java#L15

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,76 @@
...
+      throw new IllegalArgumentException(
+          "Replacement expression cannot be null in Replace command");
+    }
+    String patternStr = pattern.toString();
+    if (patternStr.contains("+")
+        || patternStr.contains("-")
+        || patternStr.contains("*")
+        || patternStr.contains("/")) {
+      throw new IllegalArgumentException(
+          "Expression is not allowed in replace pattern. Only string literals are supported.");
```
Swiddis: A little confused by this logic, we don't allow strings that contain these characters? Or if we're trying to forbid any expressions at all, I'm not sure just the 4 arithmetic operators will be sufficient (e.g. if someone tries to boolean `and` or something).

If we don't want an expression, the grammar should specify a `stringLiteral` or something along those lines, instead of `expression`.

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,76 @@
...
+  public Replace(
+      UnresolvedExpression pattern, UnresolvedExpression replacement, List<Field> fieldList) {
+    this.pattern = pattern;
+    this.replacement = replacement;
+    this.fieldList = fieldList;
+    validate();
+  }
+
+  public Replace(UnresolvedExpression pattern, UnresolvedExpression replacement) {
+    this(pattern, replacement, ImmutableList.of());
```
RyanL1997: This constructor allows an empty field list, which conflicts with the documentation stating that the field list is mandatory: 
> * field list: mandatory. One or more field names where the replacement should occur.

(see `docs/user/ppl/cmd/replace.rst:22`). 

So maybe either remove this constructor or update the documentation to clarify the behavior when no fields are specified?

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private void validate() {
+    if (pattern == null) {
```
RyanL1997, line 42: The validation logic for the pattern expression hardcodes checks for mathematical operators (`+`, `-`, `*`, `/`) on L44-47. This is may miss other invalid expressions in my opinion. Maybe consider validating that pattern is a string literal by checking its type (e.g., Literal with `DataType.STRING`)?

Something like this:
```java
if (!(pattern instanceof Literal && ((Literal) pattern).getType() == DataType.STRING)) {
    throw new IllegalArgumentException("Replace pattern must be a string literal.");
}
```

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
RyanL1997: +1 for @Swiddis 's question, and I also left a comment in this validation method in general.

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+    String patternStr = pattern.toString();
+    if (patternStr.contains("+")
+        || patternStr.contains("-")
+        || patternStr.contains("*")
+        || patternStr.contains("/")) {
+      throw new IllegalArgumentException(
+          "Expression is not allowed in replace pattern. Only string literals are supported.");
+    }
+    if (fieldList == null || fieldList.isEmpty()) {
```
RyanL1997, line 60: do we need to add a validation for duplicate fields in the `fieldList` to prevent redundant `REPLACE` operations?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1476,4 +1477,38 @@ private void buildExpandRelNode(
...
+    // Then add new fields with replaced content using new_ prefix
+    for (Field field : node.getFieldList()) {
+      String fieldName = field.getField().toString();
+      RexNode fieldRef = context.relBuilder.field(fieldName);
+
+      RexNode replaceCall =
+          context.relBuilder.call(
+              SqlStdOperatorTable.REPLACE, fieldRef, patternNode, replacementNode);
+      projectList.add(replaceCall);
+      newFieldNames.add("new_" + fieldName);
```
RyanL1997: nit: maybe a constant instead? (e.g., `NEW_FIELD_PREFIX = "new_"`)

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -131,6 +133,10 @@ renameCommand
    : RENAME renameClasue (COMMA renameClasue)*
    ;
 
+replaceCommand
+   : REPLACE pattern=expression WITH replacement=expression IN fieldList
```
RyanL1997: Question: do we need to update the grammar to enforce `stringLiteral` for `expression`? I saw that `Replace.java` restricts patterns to string literals.

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,95 @@
+=============
```
ykmr1224, line 1: Let's add this file to `docs/category.json` so doctest tests examples in it.

ppl/src/main/antlr/OpenSearchPPLParser.g4
ykmr1224: +1

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,95 @@
...
+.. rubric:: Table of contents
+
+.. contents::
+ :local:
+ :depth: 2
+
+
+Description
+============
+| Using ``replace`` command to replace text in one or more fields in the search result. The replaced text appears in new fields with prefix *new_*.
```
ykmr1224: Is it our intended behavior to introduce new attribute?

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,95 @@
...
+
+
+Syntax
+============
+replace '<pattern>' WITH '<replacement>' IN <field-name>[, <field-name>]...
+
+* pattern: mandatory. The text pattern you want to replace.
+* replacement: mandatory. The text you want to replace with.
+* field list: mandatory. One or more field names where the replacement should occur.
```
ykmr1224, line 32: Should we allow omitting field list and apply replace to all the fields?

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,95 @@
...
+Description
+============
+| Using ``replace`` command to replace text in one or more fields in the search result. The replaced text appears in new fields with prefix *new_*.
+
+
+Syntax
+============
+replace '<pattern>' WITH '<replacement>' IN <field-name>[, <field-name>]...
+
+* pattern: mandatory. The text pattern you want to replace.
```
ykmr1224: Let's clarify that we currently support only plain text pattern. (`pattern` could indicate wildcard/regex/etc.)

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Since replace command only supports plain string literals, you can use LIKE command with replace for pattern matching needs.
+
+PPL query::
+
+ os> source=test_logs | where LIKE(message, '%error error%') | eval new_message = replace(message, "error", "ERROR");
+ fetched rows / total rows = 1/1
+ +-------------------------------+-------------------------------+
+ | message                       | new_message                   |
+ |------------------------+--------------------------------------|
```
ykmr1224: nit: I see broken table format.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1476,4 +1477,38 @@ private void buildExpandRelNode(
...
+    List<String> newFieldNames = new ArrayList<>();
+
+    // First add all original fields
+    for (String fieldName : fieldNames) {
+      RexNode fieldRef = context.relBuilder.field(fieldName);
+      projectList.add(fieldRef);
+      newFieldNames.add(fieldName);
+    }
+
+    // Then add new fields with replaced content using new_ prefix
```
ykmr1224: What happens if `new_xxx` already exist?

ykmr1224 (commented):
Wild card support for rename command is implemented in https://github.com/opensearch-project/sql/pull/4019
Let's utilize the logic when we implement wild card support for `replace`.

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,95 @@
...
+=============
+
+.. rubric:: Table of contents
+
+.. contents::
+ :local:
+ :depth: 2
+
+
+Description
```
ykmr1224, line 12: Let's add notes for supported version, and also it is supported only when Calcite enabled.

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
manasvinibs: Thanks for pointing this! Yes the idea was to avoid any non-string literals for pattern text. I have updated the grammar token itself to string literal to handle this use case.

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
manasvinibs: Updated in the next revision

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
manasvinibs: Good call! Updated in the next revision.

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
manasvinibs: Added in the updated rev.

docs/user/ppl/cmd/replace.rst
manasvinibs: Currently with this implementation I'm reusing existing eval replace function which also exposes the replaced values into new fields. I'm mimicking the same behavior with direct command. But, if we think instead of adding new field, we should replace on the original one I'm open to explore that option. Let me know your thoughts.

docs/user/ppl/cmd/replace.rst
manasvinibs: As the current implementation is reusing the existing eval replace, providing one or more field name is mandatory. Replacing to all the fields can be weighed in as a follow-up as it would require changes to replace function itself as well as considering performance impact to the changes.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
manasvinibs: I tested this behavior and I see system is automatically handling the field name collision by appending a number ("new_country0") and resolving field name conflicts gracefully. I have documented this behavior.

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,107 @@
...
+============
+| Using ``replace`` command to replace text in one or more fields in the search result.
+* The command creates new fields with *new_* prefix for replaced content (e.g., replacing text in 'country' creates 'new_country')
+* If a field with *new_* prefix already exists (e.g., 'new_country'), a number will be appended to create a unique field name (e.g., 'new_country0')
+
+
+Version
+=======
+3.2.0
```
ykmr1224: Should it be 3.3.0?

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,92 @@
...
+@Setter
+@ToString
+@EqualsAndHashCode(callSuper = false)
+public class Replace extends UnresolvedPlan {
+  private final UnresolvedExpression pattern;
+  private final UnresolvedExpression replacement;
+  private final List<Field> fieldList;
+  private UnresolvedPlan child;
+
+  public Replace(
```
ykmr1224, line 33: Let's add unit test to verify validation logic.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2136,6 +2138,40 @@ public RelNode visitValues(Values values, CalcitePlanContext context) {
...
+    for (Field field : node.getFieldList()) {
+      String fieldName = field.getField().toString();
+      RexNode fieldRef = context.relBuilder.field(fieldName);
+
+      RexNode replaceCall =
+          context.relBuilder.call(
+              SqlStdOperatorTable.REPLACE, fieldRef, patternNode, replacementNode);
+      projectList.add(replaceCall);
+      newFieldNames.add(NEW_FIELD_PREFIX + fieldName);
```
ykmr1224, line 2212: Looks it does not check existing field and add suffix number as written in the doc.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -372,6 +373,26 @@ public UnresolvedPlan visitRenameCommand(RenameCommandContext ctx) {
             .collect(Collectors.toList()));
   }
 
+  /** Replace command. */
+  @Override
+  public UnresolvedPlan visitReplaceCommand(OpenSearchPPLParser.ReplaceCommandContext ctx) {
+    UnresolvedExpression pattern = AstDSL.stringLiteral(removeQuotes(ctx.pattern.getText()));
+    UnresolvedExpression replacement =
+        AstDSL.stringLiteral(removeQuotes(ctx.replacement.getText()));
```
ykmr1224: I think we should follow the pattern other commands (`grok`, `parse`) do to handle string literal:

```
  public UnresolvedPlan visitGrokCommand(OpenSearchPPLParser.GrokCommandContext ctx) {
    UnresolvedExpression sourceField = internalVisitExpression(ctx.source_field);
    Literal pattern = (Literal) internalVisitExpression(ctx.pattern);

    return new Parse(ParseMethod.GROK, sourceField, pattern, ImmutableMap.of());
  }
```

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
```diff
@@ -267,6 +268,34 @@ public String visitRename(Rename node, String context) {
...
+    }
+
+    // Build the replace command string
+    return StringUtils.format(
+        "%s | replace %s WITH %s%s", child, pattern, replacement, fieldListStr);
+  }
+
+  private String visitFieldExpression(Field field) {
+    return field.toString();
+  }
```
ykmr1224: nit: why is this method needed? Field::toString won't work?

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
```diff
@@ -546,6 +546,13 @@ public void testGrok() {
         anonymize("source=t | grok email '.+@%{HOSTNAME:host}' | fields email, host"));
   }
 
+  @Test
+  public void testReplaceCommand() {
+    assertEquals(
+        "source=EMP | replace *** WITH *** IN Field(field=fieldname, fieldArgs=[])",
+        anonymize("source=EMP | replace \"value\" WITH \"newvalue\" IN fieldname"));
```
ykmr1224, line 561: nit: better testing with multiple fields and without `IN` param.

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
```diff
@@ -267,6 +268,34 @@ public String visitRename(Rename node, String context) {
...
+
+    // Handle field list if present
+    String fieldListStr = "";
+    if (node.getFieldList() != null && !node.getFieldList().isEmpty()) {
+      fieldListStr =
+          " IN "
+              + node.getFieldList().stream()
+                  .map(this::visitFieldExpression)
+                  .collect(Collectors.joining(", "));
+    }
```
ykmr1224: Field list is mandatory, isn't it?

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
manasvinibs: Sure! Updated in the latest revision

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
manasvinibs: Yes every replace command adds a `new_` column and do not conflict with existing column names. Let me know if you think we should change the behavior.

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
manasvinibs: Added

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
manasvinibs: Yes that's right! I've removed the unnecessary check.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: I was wondering the following doc description is right since this logic doesn't check existence of the field. (Is it something automatically done?)

```* If a field with *new_* prefix already exists (e.g., 'new_country'), a number will be appended to create a unique field name (e.g., 'new_country0')```

PR #4264, opened 2025-09-09
handle double negation

ahkcs:
### Description
Try to resolve the CI failure in #4036 

Adding double negation handling logic to resolve the following issue:

Issue: The query not firstname not in ('Amber', 'Dale') should return 2 rows but returns 998 rows in 3.3.
 I ran explain queries to it, and here is the result:
  - Working (direct IN): {"bool": {"should": [{"term": "Amber"}, {"term": "Dale"}]}}
  - Broken (double negation): {"bool": {"must_not": [{"bool": {"must_not": [{"bool": {"should": [{"term": "Amber"}, {"term": "Dale"}]}}]}}]}}
It seems like the double negation was not handled properly in version 3.3

### DIFF
--- a/core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java
+++ b/core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java
@@ -153,6 +153,13 @@ public class ExpressionAnalyzer extends AbstractNodeVisitor<Expression, Analysis
 
   @Override
   public Expression visitNot(Not node, AnalysisContext context) {
+    // Handle double negation: NOT(NOT(X)) -> X
+    if (node.getExpression() instanceof Not) {
+      Not innerNot = (Not) node.getExpression();
+      // Double negation detected: return the inner expression directly
+      return innerNot.getExpression().accept(this, context);
+    }
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java
```diff
@@ -153,6 +153,13 @@ public Expression visitXor(Xor node, AnalysisContext context) {
 
   @Override
   public Expression visitNot(Not node, AnalysisContext context) {
+    // Handle double negation: NOT(NOT(X)) -> X
```
qianheng-aws, line 156: Shall we make such change in PredicateAnalyzer? That's the place we create query builder directly. So no transformation will happen between our expression and dsl.

ahkcs: If we apply the fix in `ExpressionAnalyzer`, we can prevent the nested `must_not[must_not[...]]` structure from appearing in explain queries. 
Also, in the failing test case, the query uses Legacy engine with calcite disabled, so `PredicateAnalyzer` was never used

PR #4297, opened 2025-09-15
Support `Streamstats` command with calcite

### LINKED ISSUE #4207: [RFC] Implement streamstats command in PPL
**Problem Statement**
The streamstats command is conceptually different from the traditional stats command because it performs cumulative or rolling statistical calculations in a streaming manner as events are processed, rather than computing final aggregates after the entire dataset is scanned.

While OpenSearch PPL already supports aggregation through the stats command, it does not currently provide streaming aggregations or window-based cumulative metrics similar to streamstats. The key difference is that stats in PPL aggregates data globally or by groups in a single result row per group, whereas streamstats updates and outputs running calculations for every incoming event.

**Current State**
Explains the existing setup or context and its shortcomings.

**Long-Term Goals**
* Describe the ideal outcome or future state you aim to achieve.
* What are the primary objectives of this solution should fulfill in the long run?
* How will this solution ensure sustainability or scalability over time?
* Consider whether the solution addresses the root problem and aligns with the overall goals.
* How confident are we that this solution solves the problem?

**Proposal**
Summarizes the suggested solution or improvement.

**Approach**
Details the steps or strategy to implement the proposal.

**Alternative**
Alternative that can serve as workarounds to partially or temporarily solve the problem.

**Implementation Discussion**

ishaoxy:
### Description
Support `Streamstats` command with arguments below:

```
streamstats
  [current=<bool>]
  [window=<int>]
  [global=<bool>]
  [reset_before="("<eval-expression>")"]
  [reset_after="("<eval-expression>")"]
  stats-agg-term
  [by-clause]
```
Also rule out aggregator in PPLAggregateMergeRule.
  
## Implementation Details

The implementation handles three distinct execution paths, depending on the combination of `window`, `global`, `group`, and `reset` arguments:

**Why This Design**

1. Default path can rely on native SQL OVER because there is no global/window-with-reset complexity.

2. **Specific SQL limitations:**

- Native SQL OVER clauses cannot implement **per-group sliding windows over the entire stream** . However, we want to combine a global sequence with group-level partitioning. In SQL, a window is either **global** without a BY clause or **partitioned by a group** with a BY clause; you cannot have a “global sequence plus per-group sliding frame” in one OVER.

- `ROWS BETWEEN ... PRECEDING` cannot take a variable (it only supports constants like `1 PRECEDING`, `1+1 PRECEDING`).

3. `Global + window + group` want "per-group sliding windows over entire stream," but SQL window functions do not allow fully flexible frame boundaries combined with lateral joins. Hence, we simulate it via `ROW_NUMBER() + correlated join + aggregate`.

4. Reset path introduces segment semantics (`seg_id`) that cannot be represented natively in SQL OVER clauses. Each reset creates a new frame partition. By default, reset behaves like a global window, but when grouping exists, it applies per-group aggregation within each reset segment. So I use helper columns (before_flag, after_flag, seg_id) and a correlated join ensures correctness.

**1. Default Path (No global in use / no reset)**

- Window functions are translated directly using visitWindowFunction().
- Calcite OVER clauses are generated for each aggregate.
- SQL-like plan:

```
SELECT *, <window_function_over(...)>
FROM source
```

**2. global=true + window > 0 + group exists**

To support sliding windows over the entire stream with optional grouping:

- A global sequence column (ROW_NUMBER() OVER (ORDER BY ...) AS seq) is added.
- Correlated LEFT JOINs simulate the sliding window using seq and by-clause equality filters.
- Each window function is converted into a standard aggregate (AggregateFunction) and executed within the correlated subquery.
- SQL-like plan:

```
WITH t AS (
  SELECT x.*,
         ROW_NUMBER() OVER (ORDER BY /* default ordering */) AS seq
  FROM source x
)
SELECT t.*, agg.*
FROM t
LEFT JOIN LATERAL (
  SELECT SUM(age) AS sum_age
  FROM t r
  WHERE r.seq BETWEEN t.seq - (:window - 1) AND t.seq
    AND r.gender IS NOT DISTINCT FROM t.gender
) AS agg ON TRUE;
```

**3. Reset Path (reset_before / reset_after defined)**

When `reset_before` or `reset_after` exist:

- Helper columns are added:

1. __stream_seq__: global row number.
2. __reset_before_flag__ / __reset_after_flag__: flags for reset conditions.
3. __seg_id__: segment ID, computed via SUM over flags to identify partitions.

- Correlated LEFT JOIN + aggregate simulates the frame while respecting segment boundaries (seg_id) and optional group filtering.
- SQL-like plan:

```
WITH base AS (
  SELECT s.*,
         ROW_NUMBER() OVER (ORDER BY /* default */) AS seq,
         CASE WHEN (/reset_before predicate/) THEN 1 ELSE 0 END AS before_flag,
         CASE WHEN (/reset_after  predicate/) THEN 1 ELSE 0 END AS after_flag
  FROM source s
),
seg AS (
  SELECT b.*,
         COALESCE(SUM(before_flag) OVER (ORDER BY seq ROWS UNBOUNDED PRECEDING), 0)
       + COALESCE(SUM(after_flag)  OVER (ORDER BY seq ROWS UNBOUNDED PRECEDING
                                        RANGE BETWEEN UNBOUNDED PRECEDING AND 1 PRECEDING), 0)
       AS seg_id
  FROM base b
)
SELECT t.*, agg.*
FROM seg t
LEFT JOIN LATERAL (
  SELECT /* window aggregates: SUM(age), AVG(salary), ... */
  FROM seg r
  WHERE (
    CASE
      WHEN :window = 0 AND :current  THEN r.seq <= t.seq
      WHEN :window = 0 AND NOT :current THEN r.seq <  t.seq
      WHEN :current  THEN r.seq BETWEEN t.seq - (:window - 1) AND t.seq
      ELSE               r.seq BETWEEN t.seq - :window AND t.seq - 1
    END
  )
  AND r.seg_id = t.seg_id
  AND (r.gender IS NOT DISTINCT FROM t.gender) -- optional by-clause
) AS agg ON TRUE;
```

### Related Issues
Resolves #4207

### Check List


ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -419,6 +423,90 @@ public UnresolvedPlan visitEventstatsCommand(OpenSearchPPLParser.EventstatsComma
...
+        return WindowFrame.of(
+            WindowFrame.FrameType.ROWS, (windowSize - 1) + " PRECEDING", "CURRENT ROW");
+      } else {
+        // N PRECEDING to 1 PRECEDING
+        return WindowFrame.of(WindowFrame.FrameType.ROWS, windowSize + " PRECEDING", "1 PRECEDING");
+      }
+    } else {
+      // Default: running total
+      if (current) {
+        return WindowFrame.of(WindowFrame.FrameType.ROWS, "UNBOUNDED PRECEDING", "CURRENT ROW");
```
yuancu: Can be replaced with `WindowFrame.toCurrentRow()` for robustness

core/src/main/java/org/opensearch/sql/ast/tree/StreamWindow.java
```diff
@@ -0,0 +1,57 @@
...
+import lombok.Getter;
+import lombok.ToString;
+import org.opensearch.sql.ast.AbstractNodeVisitor;
+import org.opensearch.sql.ast.expression.Argument;
+import org.opensearch.sql.ast.expression.UnresolvedExpression;
+
+@Getter
+@ToString
+@EqualsAndHashCode(callSuper = false)
+public class StreamWindow extends UnresolvedPlan {
```
yuancu, line 19: How's this class different from the `Window` class? It seems they are identical apart from `argExprList`, which stores the options / arguments of `streamstats` command. But the argExprList is never used (or will they be used in the future?).

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -419,6 +423,90 @@ public UnresolvedPlan visitEventstatsCommand(OpenSearchPPLParser.EventstatsComma
...
+    }
+
+    // Current argument
+    Literal currentArg = arguments.get("current");
+    if (currentArg != null) {
+      Object value = currentArg.getValue();
+      try {
+        current = (Boolean) value;
+      } catch (Exception e) {
+        throw new IllegalArgumentException("Invalid current flag: " + value, e);
```
yuancu: Can you also add tests for these erroneous cases?

yuancu (commented):
Looks good to me. Thanks for your contribution. I left some comments for minor suggestions.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
ishaoxy: Replaced, thanks for reminding.

core/src/main/java/org/opensearch/sql/ast/tree/StreamWindow.java
ishaoxy: Now refactored.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
ishaoxy: Added an IT case for negative window size.

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
+=============
+streamstats
+=============
```
LantaoJin: The number of `=` should equal the length of text

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+
+.. rubric:: Table of contents
+
+.. contents::
+   :local:
+   :depth: 2
+
+
+Description
+============
```
LantaoJin: ditto

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+| The ``streamstats`` command is used to calculate cumulative or rolling statistics as events are processed in order. Unlike ``stats`` or ``eventstats`` which operate on the entire dataset at once, it computes values incrementally on a per-event basis, often respecting the order of events in the search results. It allows you to generate running totals, moving averages, and other statistics that evolve with the stream of events.
+
+| Key aspects of `streamstats`:
```
LantaoJin: `|` at the beginning of line means starting a new line. So no need to add empty new line if you add `|` at the beginning of line

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+
+1. It computes statistics incrementally as each event is processed, making it suitable for time-series and sequence-based analysis.
+2. Supports arguments such as window (for sliding window calculations) and current (to control whether the current event included in calculation).
+3. Retains all original events and appends new fields containing the calculated statistics.
+4. Particularly useful for calculating running totals, identifying trends, or detecting changes over sequences of events.
+
+| Difference between ``stats``, ``eventstats`` and ``streamstats``
+All of these commands can be used to generate aggregations such as average, sum, and maximum, but they have some key differences in how they operate and what they produce:
+
```
LantaoJin, line 26: Add `|` in L24 or insert an empty new line before L24. Delete `|` in L23 or delete L22.

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+ * ``eventstats``: Based on all relevant events, then the result is added back to each event in the group.
+ * ``streamstats``: Calculations occur progressively as each event is processed; can be scoped by window.
+* Use Cases:
+ * ``stats``: When only aggregated results are needed (e.g., counts, averages, sums).
+ * ``eventstats``: When aggregated statistics are needed alongside original event data.
+ * ``streamstats``: When a running total or cumulative statistic is needed across event streams.
+
+Version
+=======
+3.3.0
```
LantaoJin: can be deleted

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+
+Syntax
+======
+streamstats [current=<bool>] [window=<int>] [global=<bool>] [reset_before="("<eval-expression>")"] [reset_after="("<eval-expression>")"] <function>... [by-clause]
+
+
+* function: mandatory. A aggregation function or window function.
+
+* current: optional.
+
```
LantaoJin: delete empty line between two bullet statements.

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+ * Syntax: current=<boolean>
+ * Description: If true, the search includes the given, or current, event in the summary calculations. If false, the search uses the field value from the previous event.
+ * Default: true
```
LantaoJin: Combine lines 57~59 into
```
 * current: optional. If true, the search includes the given, or current, event in the summary calculations. If false, the search uses the field value from the previous event. **Default:** true.
```

LantaoJin: ditto for rest

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+* window: optional.
+
+ * Syntax: window=<integer>
+ * Description: Specifies the number of events to use when computing the statistics.
+ * Default: 0, which means that all previous and current events are used.
+
+* global: optional.
+
+ * Syntax: global=<boolean>
+ * Description: Used only when the window argument is set. Defines whether to use a single window, global=true, or to use separate windows based on the by clause. If global=false and window is set to a non-zero value, a separate window is used for each group of values of the field specified in the by clause.
```
LantaoJin: > Used only when the window argument is set.

This line confused me. As the default value of `global` is `true`, and default value of `window` is `0`. What if the `window` and `global` all not set? Or what if we set global=true and window=0?

LantaoJin: And what if we set `global=false` and `window=4` but no `by` clause?

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+ * Syntax: by [span-expression,] [field,]...
+ * Description: The by clause could be the fields and expressions like scalar functions and aggregation functions. Besides, the span clause can be used to split specific field into buckets in the same interval, the stats then does the aggregation by these span buckets.
+ * Default: If no <by-clause> is specified, all events are processed as a single group and running statistics are computed across the entire event stream.
+
+* span-expression: optional, at most one.
+
+ * Syntax: span(field_expr, interval_expr)
+ * Description: The unit of the interval expression is the natural unit by default. If the field is a date and time type field, and the interval is in date/time units, you will need to specify the unit in the interval expression. For example, to split the field ``age`` into buckets by 10 years, it looks like ``span(age, 10)``. And here is another example of time span, the span to split a ``timestamp`` field into hourly intervals, it looks like ``span(timestamp, 1h)``.
+
+* Available time unit:
```
LantaoJin: refactor L96~L366 according to https://github.com/opensearch-project/sql/pull/4562/files#diff-75c70b8c047f6bd1daaea269784eeacdd4c6bba49cb4a34dd3e84a0d8aa6cffd

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+-----
+
+Description
+>>>>>>>>>>>
+
+Usage: Returns a count of the number of expr in the rows retrieved by a SELECT statement.
+
+Example::
+
+    PPL> source=accounts | streamstats count();
```
LantaoJin: Update all `PPL>` to `os>`

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+    | 13             | Nanette   | 789 Madison Street   | 32838   | F      | Nogal  | Quility  | VA    | 28  | null                  | Bates    | 1               | 1               |
+    | 1              | Amber     | 880 Holmes Lane      | 39225   | M      | Brogan | Pyrami   | IL    | 32  | amberduke@pyrami.com  | Duke     | 1               | 1               |
+    | 6              | Hattie    | 671 Bristol Street   | 5686    | M      | Dante  | Netagy   | TN    | 36  | hattiebond@netagy.com | Bond     | 2               | 2               |
+    | 18             | Dale      | 467 Hutchinson Court | 4180    | M      | Orick  | null     | MD    | 33  | daleadams@boink.com   | Adams    | 3               | 3               |
+    +----------------+-----------+----------------------+---------+--------+--------+----------+-------+-----+-----------------------+----------+-----------------+-----------------+
+
+
+Configuration
+=============
+This command requires Calcite enabled.
```
LantaoJin: delete this section

LantaoJin (commented):
haven't review the code details, just need to refactor the user doc and confirm the behaviour

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+The global argument is only applicable when a window argument is set. It defines how the window is applied in relation to the grouping fields:
+
+* global=true: a global window is applied across all rows, but the calculations inside the window still respect the by groups.
+* global=false: the window itself is created per group, meaning each group gets its own independent window.
+
+This example shows how to calculate the running average of age across accounts by country, using a global window.
+
+PPL query::
+
+    PPL> source=state_country | streamstats window=2 global=true avg(age) as running_avg by country ;
```
LantaoJin: seems the result ordered by country
```
source=state_country |  sort country | streamstats window=2 global=true avg(age) as running_avg by country
```

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,476 @@
...
+
+
+Example 4: Use the reset_before and reset_after arguments to reset statistics
+=============================================================================
+
+This example calculates the running average of age across accounts by country, with resets applied.
+
+PPL query::
+
+    PPL> source=state_country | streamstats current=false reset_before=age>34 reset_after=age<25 avg(age) as avg_age by country;
```
LantaoJin: ditto

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
```diff
@@ -0,0 +1,942 @@
...
+
+public class CalciteStreamstatsCommandIT extends PPLIntegTestCase {
+  @Override
+  public void init() throws Exception {
+    super.init();
+    enableCalcite();
+
+    loadIndex(Index.STATE_COUNTRY);
+    loadIndex(Index.STATE_COUNTRY_WITH_NULL);
+    loadIndex(Index.BANK_TWO);
```
LantaoJin, line 25: remove this line

docs/user/ppl/index.rst
```diff
@@ -112,6 +112,8 @@ The query start with search command and then flowing a set of command delimited
 
   - `stats command <cmd/stats.rst>`_
 
+  - `streamstats command <cmd/streamstats.rst>`_
```
LantaoJin, line 115: please add streamstats.rst to category.json as well

ppl/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -110,6 +111,11 @@ DEDUP_SPLITVALUES:                  'DEDUP_SPLITVALUES';
 PARTITIONS:                         'PARTITIONS';
 ALLNUM:                             'ALLNUM';
 DELIM:                              'DELIM';
+CURRENT:                            'CURRENT';
+WINDOW:                             'WINDOW';
+GLOBAL:                             'GLOBAL';
+RESET_BEFORE:                       'RESET_BEFORE';
+RESET_AFTER:                        'RESET_AFTER';
```
LantaoJin, line 118: these three seems missing in `keywordsCanBeId`

LantaoJin (commented):
Please add some test` in `CalciteExplainIT.java`

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
```diff
@@ -0,0 +1,942 @@
...
+import static org.opensearch.sql.util.MatcherUtils.*;
+
+import java.io.IOException;
+import java.util.List;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+import org.opensearch.client.Request;
+import org.opensearch.sql.ppl.PPLIntegTestCase;
+
+public class CalciteStreamstatsCommandIT extends PPLIntegTestCase {
```
LantaoJin, line 18: add some tests with other command
eventstats + streamstats:
```
source = t | eventstats ... | streamstats ...
```
sort + streamstats:
```
source = t | sort ... | streamstats ...
```
left join with streamstats:
```
source = t | left join left=l right=r on l.key=r.key [ source = tt | streamstats ... ] 
```
streamstats in in-subsearch:
```
source = t | where a in [ source = tt | streamstats ...]
```

ishaoxy: fixed all the doc issues and added all the tests mentioned above. @LantaoJin

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -604,6 +604,28 @@ public void testEventstatsDistinctCountFunctionExplain() throws IOException {
     assertJsonEqualsIgnoreId(expected, result);
   }
 
+  @Test
+  public void testStreamstatsDistinctCountExplain() throws IOException {
+    enabledOnlyWhenPushdownIsEnabled();
```
LantaoJin, line 17: why enable this test only for pushdown?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -604,6 +604,28 @@ public void testEventstatsDistinctCountFunctionExplain() throws IOException {
...
+    String query =
+        "source=opensearch-sql_test_index_account | streamstats dc(state) as distinct_states";
+    var result = explainQueryYaml(query);
+    String expected = loadFromFile("expectedOutput/calcite/explain_streamstats_dc.yaml");
+    assertYamlEqualsIgnoreId(expected, result);
+  }
+
+  @Test
+  public void testStreamstatsDistinctCountFunctionExplain() throws IOException {
+    enabledOnlyWhenPushdownIsEnabled();
```
LantaoJin, line 32: ditto

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
ishaoxy: I previously imitated the behavior of `testEventstatsDistinctCountExplain()`, and I'm not sure why it was set to only pushdown. Now fixed.

LantaoJin (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -729,6 +748,41 @@ public void testExplainOnEventstatsEarliestLatestNoGroupBy() throws IOException
...
+        expected,
+        explainQueryYaml(
+            String.format(
+                "source=%s | streamstats earliest(message, created_at) as earliest_message,"
+                    + " latest(message, created_at) as latest_message by level",
+                TEST_INDEX_LOGS)));
+  }
+
+  @Test
+  public void testExplainOnStreamstatsEarliestLatestNoGroupBy() throws IOException {
```
yuancu, line 806: Can you add a few more complex explain ITs (e.g. covering all 3 plan paths)?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
```diff
@@ -0,0 +1,1095 @@
...
+                "source=%s | streamstats window=2 global=true avg(age) as avg by country",
+                TEST_INDEX_STATE_COUNTRY));
+
+    verifyDataRows(
+        actual2,
+        rows("John", "Canada", "Ontario", 4, 2023, 25, 25),
+        rows("Jane", "Canada", "Quebec", 4, 2023, 20, 22.5),
+        rows("Jake", "USA", "California", 4, 2023, 70, 70),
+        rows("Hello", "USA", "New York", 4, 2023, 30, 50),
+        rows("Jay", "USA", "Quebec", 4, 2023, 40, 40));
```
yuancu, line 442: I'm still a little confused by this and the above example.

If I get it correctly, by setting `window=2` and `global=true`, a single window is used (for all the data?). If so, why the window is reset for Jake's row?

And why is the window reset again in Jay's row? This seems to be the only difference that is brought by `gloabl`. But for me, it seems Jay is within the window frame of size two in both exaple.

ishaoxy: You can review this example. The details of global is shown here. https://github.com/opensearch-project/sql/blob/3cf6c0618a420a0fa1070e58beca27e0fbb8ee5d/docs/user/ppl/cmd/streamstats.rst#L148

qianheng-aws:
[question]
2. global=true + window > 0 + group exists
Is 
```
WITH t AS (
  SELECT x.*,
         ROW_NUMBER() OVER (ORDER BY /* default ordering */) AS seq
  FROM source x
)
SELECT t.*, agg.*
FROM t
LEFT JOIN LATERAL (
  SELECT SUM(age) AS sum_age
  FROM t r
  WHERE r.seq BETWEEN t.seq - (:window - 1) AND t.seq
    AND r.gender IS NOT DISTINCT FROM t.gender
) AS agg ON TRUE;
```
equal to
```
SELECT 
    t.*,
    SUM(age) OVER (
        PARTITION BY gender 
        ORDER BY seq 
        ROWS BETWEEN (:window - 1) PRECEDING AND CURRENT ROW
    ) as sum_age
FROM (
    SELECT x.*,
           ROW_NUMBER() OVER (ORDER BY /* default ordering */) AS seq
    FROM source x
) t;
```
?

3. Reset Path (reset_before / reset_after defined)
I don't get the implementation presented by the SQL in our description. Seems it implements the reset_after and reset_before of accumulation by splitting all rows into several groups based on `seg_id`. But is it really correct and do we have results comparison with SPL?

qianheng-aws:
Do we have explain test cases for  feature 2(i.e. global) and 3(i.e. reset_before, reset_after)?

docs/user/ppl/cmd/streamstats.rst
```diff
@@ -0,0 +1,229 @@
...
+    | Hello | USA     | New York   | 4     | 2023 | 30  |
+    | John  | Canada  | Ontario    | 4     | 2023 | 25  |
+    | Jane  | Canada  | Quebec     | 4     | 2023 | 20  |
+    | Jim   | Canada  | B.C        | 4     | 2023 | 27  |
+    | Peter | Canada  | B.C        | 4     | 2023 | 57  |
+    | Rick  | Canada  | B.C        | 4     | 2023 | 70  |
+    | David | USA     | Washington | 4     | 2023 | 40  |
+    +-------+---------+------------+-------+------+-----+
+
+* global=true: The window slides across all rows globally (following their input order), but inside each window, aggregation is still computed by country. So we process the data stream row by row to build the sliding window with size 2. We can see that David and Rick are in a window.
```
qianheng-aws, line 173: > 'We can see that David and Rick are in a window.'

I don't get this. They are not even adjacent.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
LantaoJin: +1 for different arguments combinations.

docs/user/ppl/cmd/streamstats.rst
qianheng-aws: Please add a sort on `seq_id` so the output could keep its original sequence. It's hard to identify the correctness of results under the current sequence.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
ishaoxy: ok will do.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
yuancu: Ah ah, I get it. The displayed order is different from the data's original order. The original table should be the following:

| ID  | Name  | Age | State      | Country | Year | Month |
|-----|-------|-----|------------|---------|------|-------|
| 1   | Jake  | 70  | California | USA     | 2023 | 4     |
| 2   | Hello | 30  | New York   | USA     | 2023 | 4     |
| 3   | John  | 25  | Ontario    | Canada  | 2023 | 4     |
| 4   | Jane  | 20  | Quebec     | Canada  | 2023 | 4     |
| 5   | Jay  | 40  | Quebec     | USA  | 2023 | 4     |

Therefore, when `global=false`, `avg(age)` with a window size of 2 should be the following

| ID  | Name  | Age | State      | Country | Year | Month | avg(age) |
|-----|-------|-----|------------|---------|------|-------|----------|
| 1   | Jake  | 70  | California | USA     | 2023 | 4     |  (window usa) 70 | 
| 2   | Hello | 30  | New York   | USA     | 2023 | 4     | 50 |
| 3   | John  | 25  | Ontario    | Canada  | 2023 | 4     | (window canada) 25 |
| 4   | Jane  | 20  | Quebec     | Canada  | 2023 | 4     | 22.5 |
| 5   | Jay  | 40 | Quebec     | USA  | 2023  | 4 | (window usa continues from 2nd row) 35 |

When `global=true`, the result should be

| ID  | Name  | Age | State      | Country | Year | Month | avg(age) |
|-----|-------|-----|------------|---------|------|-------|----------|
| 1   | Jake  | 70  | California | USA     | 2023 | 4     |  (global window) 70 | 
| 2   | Hello | 30  | New York   | USA     | 2023 | 4     | 50 |
| 3   | John  | 25  | Ontario    | Canada  | 2023 | 4     | (avg(age) over line 2,3 by country) 25 |
| 4   | Jane  | 20  | Quebec     | Canada  | 2023 | 4     | 22.5 |
| 5   | Jay  | 40 | Quebec     | USA  | 2023  | 4 | (avg(age) over line 4,5 by country) 40 |

ishaoxy: yes! I think this explanation is absolutely right. Thank you sooo much!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
ishaoxy: Added.

docs/user/ppl/cmd/streamstats.rst
ishaoxy: You can refer to the explanaion in this comment. https://github.com/opensearch-project/sql/pull/4297#discussion_r2464708559

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1531,6 +1532,298 @@ private void validateFillNullTypeCompatibility(
...
+    } else { // global + window + by condition
+      RexNode frame = buildFrameFilter(context, node, outerSeq, rightSeq);
+      RexNode group = buildGroupFilter(context, groupList, v.get());
+      filter = context.relBuilder.and(frame, group);
+    }
+    context.relBuilder.filter(filter);
+
+    // aggregate all window functions on right side
+    List<AggCall> aggCalls = buildAggCallsForWindowFunctions(node.getWindowFunctionList(), context);
+    context.relBuilder.aggregate(context.relBuilder.groupKey(), aggCalls);
```
yuancu, line 1703: note (for reviewers): the group key is empty because the partition is done with `buildGroupFilter(context, groupList, v.get())`

yuancu (dismissed):
LGTM

ishaoxy:
> SELECT 
>     t.*,
>     SUM(age) OVER (
>         PARTITION BY gender 
>         ORDER BY seq 
>         ROWS BETWEEN (:window - 1) PRECEDING AND CURRENT ROW
>     ) as sum_age
> FROM (
>     SELECT x.*,
>            ROW_NUMBER() OVER (ORDER BY /* default ordering */) AS seq
>     FROM source x
> ) t;

I think the second case is the same as the default solution window=n, global=false+by, but it cannot achieve the case of window=n, global=true+by, which is why I used the former expression.

And for Reset path, I have done a lot of experiments on SPL, and I think our current implementation is consistent with the behavior of it. @qianheng-aws

docs/user/ppl/cmd/streamstats.rst
ishaoxy: I have discussed with Lantao before about the sort problem. Since the result is correct, I think we can keep this behavior (sort with partition) now. Maybe change the sort in the future enhancement pr? cc @LantaoJin

LantaoJin: > Please add a sort on seq_id so the output could keep its original sequence.

Good idea.

> Maybe change the sort in the future enhancement pr?

Checked with @qianheng-aws , I think we should add the `sort on seq_id` to align the behaviour with SPL, or the result would confuse anyone who run the command. @ishaoxy

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1642,9 +1642,27 @@ public RelNode visitStreamWindow(StreamWindow node, CalcitePlanContext context)
...
     // Default
+    RexNode streamSeq =
+        PlanUtils.makeOver(
+            context,
+            BuiltinFunctionName.ROW_NUMBER,
+            null,
+            List.of(),
+            List.of(),
+            List.of(),
+            WindowFrame.toCurrentRow());
```
LantaoJin: You'd better to create a new row_number window function RexNode using code similar to following which could has an opportunity to add an alias for the row_number(), then we can optimize in pushdown rule in future.
```
    RexNode rowNumber =
        context
            .relBuilder
            .aggregateCall(SqlStdOperatorTable.ROW_NUMBER)
            .over()
            .partitionBy(dedupeFields)
            .orderBy(dedupeFields)
            .rowsTo(RexWindowBounds.CURRENT_ROW)
            .as(ROW_NUMBER_COLUMN_FOR_DEDUP);
```

ishaoxy: fixed.

ishaoxy:
The "global" path behavior is influenced by pr #4703  and resulted in an incorrect performance. Thanks to @yuancu 's help, now aggregator is excluded in PPLAggregateMergeRule in this pr.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1604,53 +1604,64 @@ public RelNode visitStreamWindow(StreamWindow node, CalcitePlanContext context)
...
     List<RexNode> overExpressions =
         node.getWindowFunctionList().stream().map(w -> rexVisitor.analyze(w, context)).toList();
     context.relBuilder.projectPlus(overExpressions);
 
     // resort when there is by condition
     if (hasGroup) {
-      context.relBuilder.sort(context.relBuilder.field(SEQ_COL));
+      context.relBuilder.sort(context.relBuilder.field(ROW_NUMBER_COLUMN_NAME_STREAMSTATS));
+      context.relBuilder.projectExcept(
+          context.relBuilder.field(ROW_NUMBER_COLUMN_NAME_STREAMSTATS));
```
LantaoJin, line 1636: `context.relBuilder.projectExcept` was out of the `hasGroup` branch, now you move it into the branch. Is it on purpose?

integ-test/src/test/resources/expectedOutput/calcite/explain_streamstats_distinct_count.yaml
```diff
@@ -1,9 +1,15 @@
...
-        CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])
+      LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10], distinct_states=[$18])
+        LogicalSort(sort0=[$17], dir0=[ASC])
+          LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10], _id=[$11], _index=[$12], _score=[$13], _maxscore=[$14], _sort=[$15], _routing=[$16], __stream_seq__=[$17], distinct_states=[DISTINCT_COUNT_APPROX($7) OVER (PARTITION BY $4 ROWS UNBOUNDED PRECEDING)])
+            LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10], _id=[$11], _index=[$12], _score=[$13], _maxscore=[$14], _sort=[$15], _routing=[$16], __stream_seq__=[ROW_NUMBER() OVER ()])
```
LantaoJin, line 7: why there was only one LogicalProject before, now are two LogicalProject nodes.

LantaoJin: is it related to https://github.com/opensearch-project/sql/pull/4297#discussion_r2485205863?

ishaoxy: Yes, it is due to the sort by ROW_NUMBER added in the default path.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ishaoxy: Yes, this change is just a small optimization. It only creates the `row_num` column and sorts and deletes it when `hasGroup` is true in the default path. Previously, `row_num` would be created in all cases but only sorted when `hasGroup` is true. I think this meets the expectations and does not introduce any logical changes.

PR #4300, opened 2025-09-15
Add support for writing resources and include more tests

lezzago:
### Description
Add support for writing resources and include more tests

### Related Issues
Resolves N/A

### Check List

### DIFF
--- a/direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/WriteDirectQueryResourcesRequest.java
+++ b/direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/WriteDirectQueryResourcesRequest.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.directquery.rest.model;
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.Map;
+
+/*
+ * @opensearch.experimental
+ */
+@Data
+@NoArgsConstructor
+public class WriteDirectQueryResourcesRequest {
+  private String dataSource;
+  private DirectQueryResourceType resourceType;
+  private String resourceName;
+  private String request;
+  private Map<String, String> RequestOptions;
+
+  /**
+   * Sets the resource type from a string value.
+   *
+   * @param resourceTypeStr The resource type as a string
+   */
+  public void setResourceTypeFromString(String resourceTypeStr) {
+    if (resourceTypeStr != null) {
+      this.resourceType = DirectQueryResourceType.fromString(resourceTypeStr);
+    }
+  }
+}
--- a/direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClient.java
+++ b/direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClient.java
@@ -96,4 +96,12 @@ public interface PrometheusClient extends DataSourceClient {
    * @throws IOException If there is an issue with the request
    */
   JSONArray getAlertmanagerSilences() throws IOException;
+
+  /**
+   * Creates a silence in Alertmanager.
+   *
+   * @return JSONArray containing the silences
+   * @throws IOException If there is an issue with the request
+   */
+  String createAlertmanagerSilences(String silenceJson) throws IOException;
 }
--- a/direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
+++ b/direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
@@ -19,6 +19,7 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
+import okhttp3.RequestBody;
 import okhttp3.Response;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -278,11 +279,37 @@ public class PrometheusClientImpl implements PrometheusClient {
     String baseUrl = alertmanagerUri.toString().replaceAll("/$", "");
     String queryUrl = String.format("%s/api/v2/silences", baseUrl);
 
-    logger.debug("Making Alertmanager silences request: {}", queryUrl);
+    logger.debug("Making Get Alertmanager silences request: {}", queryUrl);
     Request request = new Request.Builder().url(queryUrl).build();
     Response response = this.alertmanagerHttpClient.newCall(request).execute();
 
     return readAlertmanagerResponse(response);
+  }
+
+  @Override
+  public String createAlertmanagerSilences(String silenceJson) throws IOException {
+    String baseUrl = alertmanagerUri.toString().replaceAll("/$", "");
+    String queryUrl = String.format("%s/api/v2/silences", baseUrl);
+
+    logger.debug("Making Create Alertmanager silences request: {}", queryUrl);
+    Request request = new Request.Builder()
+            .url(queryUrl)
+            .header("Content-Type", "application/json")
+            .post(RequestBody.create(silenceJson.getBytes(StandardCharsets.UTF_8)))
+            .build();
+    Response response = this.alertmanagerHttpClient.newCall(request).execute();
+
+    if (response.isSuccessful()) {
+      return Objects.requireNonNull(response.body()).string();
+    } else {
+      String errorBody = response.body() != null ? response.body().string() : "No response body";
+      logger.error(
+              "create Alertmanager Silence request failed with code: {}, error body: {}", response.code(), errorBody);
+      throw new org.opensearch.sql.prometheus.exception.PrometheusClientException(
+              String.format(
+                      "Alertmanager request failed with code: %s. Error details: %s",
+                      response.code(), errorBody));
+    }
   }
 
   /**
--- a/direct-query-core/src/main/java/org/opensearch/sql/prometheus/query/PrometheusQueryHandler.java
+++ b/direct-query-core/src/main/java/org/opensearch/sql/prometheus/query/PrometheusQueryHandler.java
@@ -20,6 +20,8 @@ import org.opensearch.sql.datasource.query.QueryHandler;
 import org.opensearch.sql.directquery.rest.model.ExecuteDirectQueryRequest;
 import org.opensearch.sql.directquery.rest.model.GetDirectQueryResourcesRequest;
 import org.opensearch.sql.directquery.rest.model.GetDirectQueryResourcesResponse;
+import org.opensearch.sql.directquery.rest.model.WriteDirectQueryResourcesRequest;
+import org.opensearch.sql.directquery.rest.model.WriteDirectQueryResourcesResponse;
 import org.opensearch.sql.prometheus.client.PrometheusClient;
 import org.opensearch.sql.prometheus.exception.PrometheusClientException;
 import org.opensearch.sql.prometheus.model.MetricMetadata;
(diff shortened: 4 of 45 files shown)


direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/WriteDirectQueryResourcesRequest.java
```diff
@@ -0,0 +1,35 @@
...
+ * @opensearch.experimental
+ */
+@Data
+@NoArgsConstructor
+public class WriteDirectQueryResourcesRequest {
+  private String dataSource;
+  private DirectQueryResourceType resourceType;
+  private String resourceName;
+  private String request;
+  private Map<String, String> RequestOptions;
```
joshuali925, line 23: would the values ever need to be non-string?

direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClient.java
```diff
@@ -99,4 +99,12 @@ Map<String, List<MetricMetadata>> getAllMetrics(Map<String, String> queryParams)
...
    */
   JSONArray getAlertmanagerSilences() throws IOException;
+
+  /**
+   * Creates a silence in Alertmanager.
+   *
+   * @return JSONArray containing the silences
+   * @throws IOException If there is an issue with the request
+   */
+  String createAlertmanagerSilences(String silenceJson) throws IOException;
```
joshuali925, line 109: return annotation doesn't match return type, is it JSONArray as string?

direct-query/src/main/java/org/opensearch/sql/directquery/rest/RestDirectQueryResourcesManagementAction.java
```diff
@@ -77,6 +83,13 @@ public List<Route> routes() {
...
+            String.format(
+                Locale.ROOT,
+                "%s/alertmanager/api/v2/{resourceType}",
+                BASE_DIRECT_QUERY_RESOURCES_URL)),
+        // Only support creating alert silences for prometheus for now
+        new Route(
+            POST,
             String.format(
                 Locale.ROOT,
                 "%s/alertmanager/api/v2/{resourceType}",
```
joshuali925, line 95: I thought we are going to have POST for both write and read resources?

direct-query/src/main/java/org/opensearch/sql/directquery/transport/format/DirectQueryRequestConverter.java
```diff
@@ -15,6 +15,9 @@
 import org.opensearch.sql.prometheus.model.PrometheusQueryType;
 import org.opensearch.sql.spark.rest.model.LangType;
 
+/*
+ * @opensearch.experimental
+ */
 @UtilityClass
 /*
  * @opensearch.experimental
```
joshuali925: nit. duplicated annotation?

direct-query/src/main/java/org/opensearch/sql/directquery/transport/format/DirectQueryResourcesRequestConverter.java
```diff
@@ -11,7 +11,11 @@
...
 import org.opensearch.sql.directquery.rest.model.DirectQueryResourceType;
 import org.opensearch.sql.directquery.rest.model.GetDirectQueryResourcesRequest;
+import org.opensearch.sql.directquery.rest.model.WriteDirectQueryResourcesRequest;
 
+/*
+ * @opensearch.experimental
+ */
 @UtilityClass
 /*
  * @opensearch.experimental
```
joshuali925: nit. duplicated annotation?

direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/GetDirectQueryResourcesActionRequest.java
```diff
@@ -21,11 +24,11 @@ public class GetDirectQueryResourcesActionRequest extends ActionRequest {
   private final GetDirectQueryResourcesRequest directQueryRequest;
 
   public GetDirectQueryResourcesActionRequest(GetDirectQueryResourcesRequest directQueryRequest) {
+    super();
     this.directQueryRequest = directQueryRequest;
   }
 
   public GetDirectQueryResourcesActionRequest(StreamInput in) throws IOException {
-    super(in);
```
joshuali925, line 28: curious why remove this?

direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/WriteDirectQueryResourcesRequest.java
lezzago: These are the request options so String should be safe here. The datasource that takes in this request to deserialize it to a non-string like an int if needed.

direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClient.java
lezzago: Good catch, will update the annotation to be of String.

direct-query/src/main/java/org/opensearch/sql/directquery/rest/RestDirectQueryResourcesManagementAction.java
lezzago: This one single route was only added and it was a create alert manager resources, so it had to be POST. I think that since this is experimental, lets keep GET for now in read resources and add POST as a followup if needed.

direct-query/src/main/java/org/opensearch/sql/directquery/transport/format/DirectQueryRequestConverter.java
lezzago: removing

direct-query/src/main/java/org/opensearch/sql/directquery/transport/format/DirectQueryResourcesRequestConverter.java
lezzago: removing

direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/GetDirectQueryResourcesActionRequest.java
lezzago: I was having test failures because this would require the input stream to have all the required params for the parent classes like needed a field like nodeId. Below is an example failure from the test `testStreamConstructorWithAllFields` in `GetDirectQueryResourcesActionRequestTest`.
```
> Task :direct-query:test FAILED
GetDirectQueryResourcesActionRequestTest > testStreamConstructorWithAllFields() FAILED
    java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because "nodeId" is null
        at org.opensearch.core.tasks.TaskId.readFromStream(TaskId.java:99)
        at org.opensearch.transport.TransportRequest.<init>(TransportRequest.java:74)
        at org.opensearch.action.ActionRequest.<init>(ActionRequest.java:58)
        at org.opensearch.sql.directquery.transport.model.GetDirectQueryResourcesActionRequest.<init>(GetDirectQueryResourcesActionRequest.java:29)
        at org.opensearch.sql.directquery.transport.model.GetDirectQueryResourcesActionRequestTest.testStreamConstructorWithAllFields(GetDirectQueryResourcesActionRequestTest.java:54)

```

joshuali925 (approved):

direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/WriteDirectQueryResourcesRequest.java
```diff
@@ -0,0 +1,35 @@
...
+/*
+ * @opensearch.experimental
+ */
+@Data
+@NoArgsConstructor
+public class WriteDirectQueryResourcesRequest {
+  private String dataSource;
+  private DirectQueryResourceType resourceType;
+  private String resourceName;
+  private String request;
```
vamsimanohar, line 22: is this requestBody?

direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/WriteDirectQueryResourcesRequest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Map;
+
+/*
+ * @opensearch.experimental
+ */
+@Data
+@NoArgsConstructor
+public class WriteDirectQueryResourcesRequest {
+  private String dataSource;
+  private DirectQueryResourceType resourceType;
```
vamsimanohar, line 20: Is this class form earlier PRs? Is this going to be a list of all resource types from all datasource types?

direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
```diff
@@ -281,13 +282,39 @@ public JSONArray getAlertmanagerSilences() throws IOException {
...
+            .build();
+    Response response = this.alertmanagerHttpClient.newCall(request).execute();
+
+    if (response.isSuccessful()) {
+      return Objects.requireNonNull(response.body()).string();
+    } else {
+      String errorBody = response.body() != null ? response.body().string() : "No response body";
+      logger.error(
+              "create Alertmanager Silence request failed with code: {}, error body: {}", response.code(), errorBody);
+      throw new org.opensearch.sql.prometheus.exception.PrometheusClientException(
```
vamsimanohar: is this fully qualified name necessary? 
weirdly llms are generating fully qualified names I have faced this issue earlier.

direct-query-core/src/main/java/org/opensearch/sql/prometheus/query/PrometheusQueryHandler.java
```diff
@@ -192,4 +194,34 @@ public GetDirectQueryResourcesResponse<?> getResources(
...
+                        String createdSilence = client.createAlertmanagerSilences(request.getRequest());
+                        return WriteDirectQueryResourcesResponse.withList(List.of(createdSilence));
+                      }
+                      default:
+                        throw new IllegalArgumentException(
+                                "Invalid prometheus resource type: " + request.getResourceType());
+                    }
+                  } catch (IOException e) {
+                    LOG.error("Error getting resources", e);
+                    throw new org.opensearch.sql.prometheus.exception.PrometheusClientException(
```
vamsimanohar: Nit same as above.

direct-query-core/src/main/java/org/opensearch/sql/prometheus/client/PrometheusClientImpl.java
lezzago: Ack, it was from the existing code that I copied, but good catch. I will see if I can clean any of this up anywhere else in the modules.

direct-query/src/main/java/org/opensearch/sql/directquery/transport/TransportWriteDirectQueryResourcesRequestAction.java
```diff
@@ -0,0 +1,72 @@
...
+/*
+ * @opensearch.experimental
+ */
+public class TransportWriteDirectQueryResourcesRequestAction
+    extends HandledTransportAction<
+        WriteDirectQueryResourcesActionRequest, WriteDirectQueryResourcesActionResponse> {
+
+  private final DirectQueryExecutorService directQueryExecutorService;
+
+  public static final String NAME = "indices:data/write/direct_query_resources";
```
vamsimanohar: Is `indices:data` correct namespace? What is the rationale behind this? 

` public static final String NAME = "cluster:admin/opensearch/ql/datasources/read";`

For example we are using the above for GetDatasourcesQueryRequest.  The rationale here is that datasources are cluster level resource which admin can configure.

vamsimanohar: [NIT] Also, we need to create an issue for making changes in security dashboards plugin to include all the new permissions.

direct-query-core/src/main/java/org/opensearch/sql/prometheus/query/PrometheusQueryHandler.java
lezzago: done

direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/WriteDirectQueryResourcesRequest.java
lezzago: This class is from earlier PRs. Currently this is the list of all resource types. When this goes to a new plugin, there needs to be some re-architecturing to fetch this automatically for each datasource supported.

direct-query-core/src/main/java/org/opensearch/sql/directquery/rest/model/WriteDirectQueryResourcesRequest.java
lezzago: Yes, we pass through the request body.

lezzago:
> LGTM. I think we still need to think through on what to use for permissions.
> 
> `cluster:admin/opensearch/direct_query/write/resources`

Is this for more granular level permissions on what data sources the users would have access to? But I agree before going GA with the feature, we need to have this get more feedback from the community as well.

PR #4305, opened 2025-09-16
Updating coalesce documentation

aalva500-prog:
### Description
Current COALESCE Documentation Status
✅ Well Covered - OpenSearch 3.1+ COALESCE function:

The documentation already includes comprehensive coverage of the native COALESCE function:

Syntax and usage - ✅ Documented with clear usage pattern
Common use cases and examples - ✅ Multiple detailed examples including:
Basic field fallback scenarios
Empty string handling
Mixed data types with auto coercion
Non-existent field handling
Version specification - ✅ Clearly marked as "Version: 3.1.0"
Behavior details - ✅ Comprehensive behavior section covering:
Null and missing value handling
Empty string treatment
Type coercion rules
Return type determination
❌ Missing - Pre-3.1 COALESCE alternative patterns:
The documentation is missing the requested section for pre-3.1 OpenSearch versions. Specifically missing:

No documentation for nested IFNULL patterns as COALESCE alternatives
No examples of the pattern you mentioned: ifnull(cast(attributes.http.response.status_code as int), ifnull(cast(attributes.http.status_code as int), -1))
No version guidance for users on older OpenSearch versions
No observability use case examples for varying field names
❌ Missing - Performance considerations:
The current documentation doesn't include performance considerations for the COALESCE function.

❌ Missing - Limitations or restrictions:
While the documentation covers behavior well, it doesn't explicitly call out limitations or restrictions.

### Related Issues
Resolves #[4199]

### Check List
- [] New functionality includes testing.
- [] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/docs/user/ppl/functions/condition.rst
+++ b/docs/user/ppl/functions/condition.rst
@@ -255,11 +255,24 @@ Argument type: all the supported data type. Supports mixed data types with autom
 Return type: determined by the least restrictive common type among all arguments, with fallback to string if no common type can be determined
 
 Behavior:
+
 - Returns the first value that is not null and not missing (missing includes non-existent fields)
 - Empty strings ("") and whitespace strings (" ") are considered valid values
 - If all arguments are null or missing, returns null
 - Automatic type coercion is applied to match the determined return type
 - If type conversion fails, the value is converted to string representation
+- For best results, use arguments of the same data type to avoid unexpected type conversions
+
+Performance Considerations:
+
+- Optimized for multiple field evaluation, more efficient than nested IFNULL patterns
+- Evaluates arguments sequentially, stopping at the first non-null value
+- Consider field order based on likelihood of containing values to minimize evaluation overhead
+
+Limitations:
+
+- Type coercion may result in unexpected string conversions for incompatible types
+- Performance may degrade with very large numbers of arguments
 
 Example::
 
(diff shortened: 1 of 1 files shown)


docs/user/ppl/functions/condition.rst
```diff
@@ -324,6 +337,47 @@ Non-existent Field Handling::
...
+
+Version: Pre-3.1
+
+For OpenSearch versions prior to 3.1, COALESCE-like functionality can be achieved using nested IFNULL statements. This pattern is particularly useful in observability use cases where field names may vary across different data sources.
+
+Usage: ifnull(field1, ifnull(field2, ifnull(field3, default_value)))
+
+Example - HTTP Status Code Fallback::
+
+    PPL> source=http_logs | eval status_code = ifnull(cast(attributes.http.response.status_code as int), ifnull(cast(attributes.http.status_code as int), -1)) | fields status_code, attributes.http.response.status_code, attributes.http.status_code
```
ykmr1224: We should use `os>` instead of `PPL>` to enable doctest.

docs/user/ppl/functions/condition.rst
```diff
@@ -324,6 +337,47 @@ Non-existent Field Handling::
     | Dale    | Dale      |
     +---------+-----------+
 
+Pre-3.1 Alternative: Nested IFNULL Pattern
```
ykmr1224: `Pre-3.1` sounds vague to me. `(Version <= 3.1)` or `(Only 3.0/3.1)` might be better notation? (I was not sure if it applies to 3.1 or not)

docs/user/ppl/functions/condition.rst
aalva500-prog: Seems like this file is using `PPL>` in a lot of commands. I'll fix all of them.

docs/user/ppl/functions/condition.rst
aalva500-prog: The nested `IFNULL` pattern should work on all versions, whereas `coalesce` only works on the newest versions. Maybe we can keep it as follows to avoid confusion:

```
Alternative: Nested IFNULL Pattern
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

For OpenSearch versions prior to 3.1, COALESCE-like functionality can be achieved using nested IFNULL statements. This pattern is particularly useful in observability use cases where field names may vary across different data sources.

Usage: ifnull(field1, ifnull(field2, ifnull(field3, default_value)))
```

docs/user/ppl/functions/condition.rst
```diff
@@ -324,6 +337,33 @@ Non-existent Field Handling::
...
+    | Dale    | null     | Dale      | Adams    |
+    +---------+----------+-----------+----------+
+
+Notes:
+
+- No automatic type coercion (requires explicit casting when needed)
+- More verbose syntax compared to native COALESCE
+- Nested structure can become complex with many fallback fields
+- Upgrade to OpenSearch 3.1+ for native COALESCE support with better performance and cleaner syntax
+
```
penghuo: Move to IFNULL section, as another example to explain how it works.

Swiddis (approved):

docs/user/ppl/functions/condition.rst
aalva500-prog: Sure, I'll do that, thanks!

penghuo (approved):

PR #4320, opened 2025-09-17
Spotless precommit: apply instead of check

Swiddis:
### Description
Following https://github.com/opensearch-project/sql/pull/4306#discussion_r2354100847: want to playtest this before merging since it might have weird staging behavior.

### Related Issues
N/A

### Check List


Swiddis:
upd: turns out that precommit stashes unstaged patches before running, so don't need to worry about staging at all when we get to the exec.

ps48 (dismissed):
Feels great to see this addition 🥇  😄 . I have invested good numbers of hours waiting for the CI just to see it fail on Spotless where rest of all integ, unit tests have passed.

ps48:
@Swiddis Should we have both `apply` first and then `check` later? What if spotlessApply isn't able to fix all issues, something like an `import missing`

Swiddis:
> I have invested good numbers of hours waiting for the CI just to see it fail on Spotless where rest of all integ, unit tests have passed.

I think we all have -- hits most when you remember to check before the original push, but then forget to re-run it after applying PR feedback that doesn't change functionality

RyanL1997 (dismissed):

Swiddis:
> @Swiddis Should we have both apply first and then check later? What if spotlessApply isn't able to fix all issues, something like an import missing

I looked around and I don't see any rules we're using that can't be autoapplied, but we can future-proof it

RyanL1997 (approved):

ps48 (approved):

Issue #4275, opened 2025-09-11
[FEATURE] Unify the logic to lookup default time field


ykmr1224:
Several PPL functions utilize default timestamp field `@timestamp`, but those are identified by separate logic.
We should refactor and unify the logic for better maintainability.

https://github.com/opensearch-project/sql/pull/4212
https://github.com/opensearch-project/sql/pull/4138

RyanL1997:
Transferring from @anasalkouz : Also add the capability of handling other provided time field.

PR #4332, opened 2025-09-18
Support `multisearch` command in calcite

ahkcs:
### Description
  This PR introduces the multisearch command, a new PPL command that allows combining results from multiple search subsearches into a single unified result set.

  Resolves: #4348

  **Key Features:**
  - Combines results from 2+ search subsearches with different filtering criteria
  - **Supports ALL PPL commands** within subsearches (where, eval, fields, stats, sort, dedup, etc.)
  - Enables timestamp-based result interleaving for time-series data analysis
  - Uses efficient UNION ALL + ORDER BY architecture for optimal performance

  **Example Usage:**
  ```sql
  -- Basic age group analysis
  source=accounts | multisearch
    [search source=accounts | where age < 30 | eval age_group = "young"]
    [search source=accounts | where age >= 30 | eval age_group = "adult"]
  | stats count by age_group

  -- Success rate monitoring pattern
  source=logs | multisearch
    [search source=logs | where status="success" | eval result="success"]
    [search source=logs | where status!="success" | eval result="total"]
  | stats count(eval(result="success")) as success_count, count() as total_count

```

### CODE REVIEW

File: docs/user/ppl/cmd/multisearch.rst [hunk 1]
```diff
@@ -0,0 +1,269 @@
...
+
+* **Comparative Analysis**: Compare metrics across different segments, regions, or time periods
+* **Success Rate Monitoring**: Calculate success rates by comparing successful vs. total operations
+* **Multi-source Data Combination**: Merge data from different indices or apply different filters to the same source
+* **A/B Testing Analysis**: Combine results from different test groups for comparison
+* **Time-series Data Merging**: Interleave events from multiple sources based on timestamps
+
+Version
+=======
+3.0.0
```
dai-chen: since 3.0.0?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -451,6 +453,14 @@ appendCommand
    : APPEND LT_SQR_PRTHS searchCommand? (PIPE commands)* RT_SQR_PRTHS
    ;
 
+multisearchCommand
+   : MULTISEARCH multisearchSubsearch (multisearchSubsearch)*
```
dai-chen: reuse `subSearch`?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -1035,6 +1036,34 @@ public UnresolvedPlan visitAppendCommand(OpenSearchPPLParser.AppendCommandContex
...
+          subsearchCtx.commands().stream()
+              .map(this::visit)
+              .reduce(searchCommandInSubSearch, (r, e) -> e.attach(r));
+
+      subsearches.add(fullSubsearch);
+    }
+
+    if (subsearches.size() < 2) {
+      throw new SemanticCheckException("At least two searches must be specified");
+    }
```
dai-chen, line 1041: enforce this in grammar if required?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1540,6 +1544,193 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
     return context.relBuilder.peek();
   }
 
+  @Override
+  public RelNode visitMultisearch(Multisearch node, CalcitePlanContext context) {
```
dai-chen, line 1671: not sure where the complexity comes. I thought each subsearch can be translated by existing visit logic?
I checked the similar visitAppend and it seems even more complex...

docs/user/ppl/cmd/multisearch.rst
ahkcs: Updated

ppl/src/main/antlr/OpenSearchPPLParser.g4
ahkcs: Updated to reuse `subSearch`

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
ahkcs: enforced

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: simplified, the visitMultisearch method currently processes each subsearch using existing visitor logic, aligns their schemas for union compatibility, creates a Calcite union of all results, and also checks for timestamp-based ordering.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1575,6 +1578,110 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+      List<RelNode> subsearchNodes, CalcitePlanContext context) {
+    // Collect all unique field names in order of first appearance
+    Set<String> allFieldNames = new java.util.LinkedHashSet<>();
+    for (RelNode node : subsearchNodes) {
+      for (RelDataTypeField field : node.getRowType().getFieldList()) {
+        allFieldNames.add(field.getName());
+      }
+    }
+
+    List<String> fieldOrder = new ArrayList<>(allFieldNames);
```
ykmr1224: Why do we need a copy?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1575,6 +1578,110 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+            context.rexBuilder.makeInputRef(
+                context.relBuilder.peek(), timestampFieldRef.getIndex());
+        context.relBuilder.sort(context.relBuilder.desc(timestampRef));
+      }
+    }
+
+    return context.relBuilder.peek();
+  }
+
+  private List<RelNode> alignSchemasForUnion(
```
ykmr1224: I think this logic can be single path (simply go through each subsearchNodes, collect field name, and decide fieldOrder, then add to projections. )

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1575,6 +1578,110 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+
+  /**
+   * Finds the timestamp field for multisearch ordering.
+   *
+   * @param rowType The row type to search for timestamp fields
+   * @return The name of the timestamp field, or null if not found
+   */
+  private String findTimestampField(RelDataType rowType) {
+    // Look for common timestamp field names
+    String[] timestampFieldNames = {"_time", "@timestamp", "timestamp", "time"};
```
ykmr1224: This should be common among commands. Currently, we should support `@timestamp`.
ref: https://github.com/opensearch-project/sql/issues/4275

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -151,6 +152,32 @@ public void testExplainIsEmpty() throws IOException {
...
+  @Test
+  public void testExplainMultisearchBasic() throws IOException {
+    String query =
+        "source=opensearch-sql_test_index_account | multisearch [search"
+            + " source=opensearch-sql_test_index_account | where age < 30 | eval age_group ="
+            + " 'young'] [search source=opensearch-sql_test_index_account | where age >= 30 | eval"
+            + " age_group = 'adult'] | stats count by age_group";
+    var result = explainQueryToString(query);
+    String expected = loadExpectedPlan("explain_multisearch_basic.json");
+    assertJsonEqualsIgnoreId(expected, result);
```
ykmr1224: Let's use `assertYamlEqualsJsonIgnoreId` for readable plan comparison. You can check the test case `supportSearchSargPushDown_singleRange` for an example.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
```diff
@@ -0,0 +1,378 @@
...
+  @Test
+  public void testBasicMultisearch() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | multisearch "
+                    + "[search source=%s | where age < 30 | eval age_group = \\\"young\\\"] "
+                    + "[search source=%s | where age >= 30 | eval age_group = \\\"adult\\\"] "
+                    + "| stats count by age_group | sort age_group",
+                TEST_INDEX_ACCOUNT, TEST_INDEX_ACCOUNT, TEST_INDEX_ACCOUNT));
```
ykmr1224: Let's have a test with different indices to test schema merge.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMultisearchTest.java
```diff
@@ -0,0 +1,368 @@
...
+    verifyPPLToSparkSQL(root, expectedSparkSql);
+  }
+
+  // ========================================================================
+  // Custom Table Implementation for Timestamp Testing
+  // ========================================================================
+
+  /** Custom table implementation with timestamp fields for multisearch testing. */
+  @RequiredArgsConstructor
+  static class TimeDataTable implements ScannableTable {
```
ykmr1224, line 322: Why do we need custom table implementation?

ykmr1224 (commented):
Question: what happens if sub-searchs have different type for the same field name?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1575,6 +1578,110 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
     return context.relBuilder.peek();
   }
 
+  @Override
+  public RelNode visitMultisearch(Multisearch node, CalcitePlanContext context) {
+    // Process each subsearch using existing visit logic
```
RyanL1997: lets remove these verbose comments

docs/user/ppl/cmd/multisearch.rst
```diff
@@ -0,0 +1,269 @@
...
+* **Prohibited Commands**: Non-streaming commands like ``stats``, ``sort``, ``dedup`` are not allowed within subsearches
+* **Schema Compatibility**: Fields with the same name across subsearches should have compatible types
+
+Configuration
+=============
+This command requires Calcite enabled.
+
+Enable Calcite::
+
+	>> curl -H 'Content-Type: application/json' -X PUT localhost:9200/_plugins/_query/settings -d '{
```
RyanL1997: Correct me if im wrong: do we actually need this section for configuration?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMultisearchTest.java
RyanL1997: +1 for Tomo's question

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1575,6 +1578,110 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+      }
+    }
+
+    return context.relBuilder.peek();
+  }
+
+  private List<RelNode> alignSchemasForUnion(
+      List<RelNode> subsearchNodes, CalcitePlanContext context) {
+    // Collect all unique field names in order of first appearance
+    Set<String> allFieldNames = new java.util.LinkedHashSet<>();
```
RyanL1997: Consider using the imported `java.util.HashSet` or add proper imports instead of fully qualified names.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1575,6 +1578,110 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+
+  private RelDataType findFieldTypeInNodes(String fieldName, List<RelNode> nodes) {
+    for (RelNode node : nodes) {
+      RelDataTypeField field = node.getRowType().getField(fieldName, false, false);
+      if (field != null) {
+        return field.getType();
+      }
+    }
+    // Fallback to VARCHAR if not found (shouldn't happen in practice)
+    return nodes.get(0).getCluster().getTypeFactory().createSqlType(SqlTypeName.VARCHAR);
```
RyanL1997: This fallback assumes `VARCHAR` but should consider using a more appropriate type or handle the case where no field type is found more gracefully. Or do we actually need this fallback?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -451,6 +453,10 @@ appendCommand
    : APPEND LT_SQR_PRTHS searchCommand? (PIPE commands)* RT_SQR_PRTHS
    ;
 
+multisearchCommand
+   : MULTISEARCH LT_SQR_PRTHS subSearch RT_SQR_PRTHS LT_SQR_PRTHS subSearch RT_SQR_PRTHS (LT_SQR_PRTHS subSearch RT_SQR_PRTHS)*
```
RyanL1997: The grammar enforces minimum two subsearches at parse time, but the error message in tests suggests runtime validation. Consider adding a more descriptive parse error.

docs/user/ppl/cmd/multisearch.rst
```diff
@@ -0,0 +1,269 @@
...
+
+* result-processing: optional. Commands applied to the merged results.
+
+ * Description: After the multisearch operation, you can apply any PPL command to process the combined results, such as ``stats``, ``sort``, ``head``, etc.
+
+Limitations
+===========
+
+* **Minimum Subsearches**: At least two subsearches must be specified
+* **Streaming Commands Only**: Subsearches can only contain streaming commands (``where``, ``eval``, ``fields``, ``head``, ``rename``)
```
RyanL1997: The limitations section mentions "Streaming Commands Only" but the implementation and tests show non-streaming commands are now supported. Remove or update this section.

docs/user/ppl/cmd/multisearch.rst
```diff
@@ -0,0 +1,269 @@
...
+* **A/B Testing Analysis**: Combine results from different test groups for comparison
+* **Time-series Data Merging**: Interleave events from multiple sources based on timestamps
+
+Version
+=======
+3.3.0
+
+Syntax
+======
+multisearch [search subsearch1] [search subsearch2] ... [search subsearchN]
```
RyanL1997: The syntax documentation is inconsistent with the actual grammar. L39 shows optional minimum, but L41 states "At least two search subsearches must be specified."

RyanL1997 (commented):
Hi @ahkcs , thanks for the change. I just left some comments.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Removed copy

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Updated to single path

ahkcs:
> Question: what happens if sub-searchs have different type for the same field name?

Currently we'll throw exception, for example: `class java.lang.Integer cannot be cast to class java.math.BigDecimal`
I have added IT tests for this behavior

docs/user/ppl/cmd/multisearch.rst
ahkcs: Updated

docs/user/ppl/cmd/multisearch.rst
ahkcs: Updated

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Removed fallback

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java

docs/user/ppl/cmd/multisearch.rst
ahkcs: Removed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: removed

ppl/src/main/antlr/OpenSearchPPLParser.g4
ahkcs: Updated, removed enforcement and updated error handling, added corresponding test cases

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMultisearchTest.java
ahkcs: We need this custom table implementation to test for timestamp interleaving behavior, the existing tables are not enough for testing timestamp interleaving

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
ahkcs: Updated

penghuo:
> Question: what happens if sub-searchs have different type for the same field name?

+1. is it common issue for append and multisearch command? if yes, consider fix it in seperate PR.

ahkcs:
> > Question: what happens if sub-searchs have different type for the same field name?
> 
> +1. is it common issue for append and multisearch command? if yes, consider fix it in seperate PR.

Created github issue: https://github.com/opensearch-project/sql/issues/4383 to track schema merging for `append` and `multisearch` command

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Hi Tomo, I am currently using this method as a temp solution to detect timestamp field, this can be updated with future implementation

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
ahkcs: Added test case for schema merge between different indices without type conflict, for cases with type conflict, we will resolve in a separate PR, currently tracked with this github issue: https://github.com/opensearch-project/sql/issues/4383

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: I still see two for loops on subsearchNodes. Let me if you didn't understand my intension.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: OK, but should we put `@timestamp` as first priority?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Hi Tomo, I understand your intention, but after implementation and testing, I found that using one for loop isn't feasible due to a dependency constraint: we need the complete field set from all nodes before creating projections for any node.

  In a single loop approach:
  - Node A gets projected with fields [name, age] (only fields seen so far)
  - Node B gets projected with fields [name, age, city] (after discovering city)
  - Result: Schema mismatch → Union fails with "Cannot compute compatible row type"

  The two loops are functionally required:
  1. Loop 1: Discover all fields across all nodes → [name, age, city]
  2. Loop 2: Project each node with complete schema → Node A: [name, age, NULL], Node B: [name, age, city]

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Updated implementation to put @timestamp as first priority

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
songkant-aws: Thanks for raising such problem. I'm wondering what's the data and case that triggers `RexCall cannot be cast to RexLiteral` error. Initial `append` has such test case to cover schema merge: https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAppendCommandIT.java#L216-L238

I think there is a missing corner case. Suspect if it's related to implicit casting(maybe we should use some SafeCast mechanism to return null instead of error) but I couldn't reproduce the error in my local.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
dai-chen: So we need this because UNION in SQL requires the same output schema but our multisearch doesn't right? Is this logic the same the one in `visitAppend`? If so, can we try to deduplicate as much as possible?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
ahkcs: Thanks for the clarification, after some further investigation, it does seem like a corner case caused by `birthdate` field in `bank` index:
Error can be reproduced with this IT:
```
@Test
  public void testAppendAsymmetricWithBirthdateOnly() throws IOException {
    JSONObject actual =
        executeQuery(
            String.format(
                Locale.ROOT,
                "source=%s | fields account_number, age | append [ source=%s | fields account_number, age, birthdate ]",
                TEST_INDEX_ACCOUNT,
                TEST_INDEX_BANK));

    // birthdate should be NULL-filled for ACCOUNT records
    verifySchemaInOrder(
        actual,
        schema("account_number", "bigint"),
        schema("age", "bigint"),
        schema("birthdate", "timestamp"),
        schema("age0", "int"));
  }
  ```
  I'll update the github issue accordingly as this is not a type conflict issue

File: core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java (hunk 6 above)
Comment (ahkcs) [reply core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java]: Updated implementation to refactor and reuse logic in `visitAppend`

File: integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java (hunk 9 above)
Comment (songkant-aws) [reply integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java]: Thanks for the reproduced test case. Found the problem. `birthdate` field is a timestamp. We uses User-Defined-Type to handle this type because PPL timestamp storage could be either STRING or LONG. Internally, we convert it to our UDT type to represent it.

Sometimes we need make sure type output is consistent and convert output(maybe LONG) to UDT type EXPR_TIMESTAMP(physically STRING type). See overridden logic here: https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/calcite/ExtendedRexBuilder.java#L128-L152
The UDT will be transferred to a RexCall (our own Cast UDF), which can't be cast to RexLiteral while calling Calcite super method of makeCast at 4th line here:
```
0  public RexLiteral makeNullLiteral(RelDataType type) {
1   if (!type.isNullable()) {
2      type = typeFactory.createTypeWithNullability(type, true);
3    }
4    return (RexLiteral) makeCast(type, constantNull);
5  }
```

Comment (songkant-aws) [reply integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java]: So the fix is: Support UDT type cast to NULL literal

**Comment by RyanL1997** [comment]
lets resolve the conflict

File: core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java [hunk 18]
```diff
@@ -1633,63 +1635,206 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+
+          // Add NULL projection for this field in all previous nodes
+          for (List<RexNode> prevProjection : allProjections) {
+            prevProjection.add(context.rexBuilder.makeNullLiteral(unmatchedField.getType()));
+          }
+
+          // Add actual field reference for current node
+          currentProjection.add(context.rexBuilder.makeInputRef(currentNode, j));
+        }
       }
```
ykmr1224: This logic seems very difficult to understand and error-prone since loops are correlated each other. Can't we simply go through currentFields and branch based on if existing field contain the name already or not?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1633,63 +1635,206 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+      List<RelDataTypeField> currentFields = currentNode.getRowType().getFieldList();
+
+      // Create field map for current node
+      Map<String, RelDataTypeField> currentFieldMap =
+          currentFields.stream()
+              .collect(
+                  Collectors.toMap(
+                      RelDataTypeField::getName,
+                      field -> field,
+                      (existing, replacement) -> existing)); // Keep first occurrence if duplicates
```
ykmr1224: Do we need to rebuild this map every time?

Wondering if duplicate is expected here. (should it already be deduped?)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1633,63 +1635,206 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+    }
+
+    // Fallback: check other common timestamp field names
+    String[] fallbackTimestampNames = {"_time", "timestamp", "time"};
+    for (String fieldName : fallbackTimestampNames) {
+      RelDataTypeField field = rowType.getField(fieldName, false, false);
+      if (field != null) {
+        return fieldName;
+      }
+    }
```
ykmr1224: Why do we need to separate the logic for "@timestamp" and others?

ahkcs: Hi Tomo, I think I may have misunderstood your comment: https://github.com/opensearch-project/sql/pull/4332#discussion_r2380480581
How do you think I should put @timestamp as first priority?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Hi Tomo,  I think we do need to rebuild it for each iteration because each iteration processes a different node, and each node has its own distinct field list

For duplicate, I think it's not expected here as Calcite validates that a RelNode cannot have duplicate field names in its schema, I have updated to remove duplication merging

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Hi Tomo, thanks for the feedback! You can also refer to my previous comment for this: 
https://github.com/opensearch-project/sql/pull/4332#discussion_r2380545639

  The algorithm has two fundamentally different responsibilities:

  1. Build projection in schema order
    - Iterates through names (unified schema) to maintain consistent field ordering
    - Adds field references or NULLs for each schema position
  2.  Discover new fields in field order
    - Iterates through currentFields to find fields not yet in schema
    - Extends schema and backfills previous nodes

Example: If Node 0 has [firstname, age, balance] and Node 1 has [city, firstname], Node 1's projection must be [firstname, NULL, NULL, city] - maintaining schema order. Iterating only through currentFields [city, firstname] can't directly produce this ordering.

ahkcs: I also updated to have clearer comments for better understanding

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: I think we can simply write like:

```
    String[] candidates = {"@timestamp", "_time", "timestamp", "time"};
    for (String fieldName : candidates) {
      RelDataTypeField field = rowType.getField(fieldName, false, false);
      if (field != null) {
        return fieldName;
      }
    }
```
(result should be same as the current implementation, unless I misunderstand something)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1633,63 +1635,204 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+      List<RexNode> currentProjection = new ArrayList<>();
+
+      // Phase 1: Build projection for existing schema fields (in schema order)
+      // This ensures the output maintains the established field order from previous nodes
+      for (int i = 0; i < names.size(); i++) {
+        String schemaFieldName = names.get(i);
+        RelDataTypeField baseField = baseFields.get(Math.min(i, baseFields.size() - 1));
+        RelDataTypeField currentField = currentFieldMap.get(schemaFieldName);
+
+        if (currentField != null && currentField.getType().equals(baseField.getType())) {
```
ykmr1224: why do we need Math.min here? And what happens when i exceeds the baseFields.size() - 1?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1633,63 +1635,204 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+    if (nodes.size() == 1) {
+      return new SchemaUnificationResult(nodes, nodes.get(0).getRowType().getFieldNames());
+    }
+
+    // Strategy: Use first node as the "main" schema, then merge others one by one
+    // This mimics append behavior where there's a main query + subsearches
+
+    RelNode firstNode = nodes.get(0);
+    List<RelDataTypeField> baseFields = firstNode.getRowType().getFieldList();
+    List<String> names = new ArrayList<>();
```
ykmr1224: can we give better name for this variable?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: Now I see what it actually does...

Sorry if my earlier comment confused you (regarding 1 path/2 path), but I think it would become simpler if we first go through all the nodes to come up with final schema, and use that to project each node. (I might still misunderstand the logic, but this is due to very long method containing many loops/if blocks. More extraction and smaller function would help code reader understand the logic. And that leads to better review feedbacks.)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Hi Tomo, thanks for the suggestion! I have updated the code
Now the First pass collects all fields from all nodes into a unified schema, then each node is projected against that schema
And I extracted it into 3 smaller methods for better readability:
    - buildUnifiedSchema() - Discovers all unique (name, type) combinations across nodes
    - buildProjectionForNode() - Projects a single node to match the unified schema
    - buildUnifiedSchemaWithConflictResolution()

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Renamed names to unifiedFieldNames for better clarity

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Hi Tomo, the Math.min was a bug. When processing multiple nodes, the names list grows as new fields are discovered, but baseFields stays the same size as the first node. When i >= baseFields.size(), we'd get an IndexOutOfBoundsException without Math.min, but using it gave us the wrong type (always the last field's type).

I've fixed it by maintaining a unifiedTypeMap that tracks the correct type for each field as the schema grows, eliminating the need for Math.min.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: Thanks for the fix. It became far better!

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1633,63 +1636,226 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+      }
+    }
+
+    return context.relBuilder.peek();
+  }
+
+  /** Result of schema unification with conflict resolution. */
+  private static class SchemaUnificationResult {
+    private final List<RelNode> projectedNodes;
+    private final List<String> unifiedFieldNames;
```
ykmr1224: Looks like we no longer use unifiedFieldNames. Is this class still needed?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1633,63 +1636,226 @@ public RelNode visitAppend(Append node, CalcitePlanContext context) {
...
+
+  /**
+   * Builds a unified schema for multiple nodes with type conflict resolution. Uses the same
+   * strategy as append command - renames conflicting fields to avoid type conflicts.
+   *
+   * @param nodes List of RelNodes to unify schemas for
+   * @param context Calcite plan context
+   * @return SchemaUnificationResult containing projected nodes and unified field names
+   */
+  private SchemaUnificationResult buildUnifiedSchemaWithConflictResolution(
```
ykmr1224: Can we move these functions to a separate class?
CalciteRelNodeVisitor is already huge, and it's better reducing the size for maintainability.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Removed as no longer needed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Thanks for the suggestion! I have extracted these function to a new class `SchemaUnifier` for better maintainability

doctest/test_data/time_test_data.json
```diff
@@ -0,0 +1,200 @@
+{"index":{"_id":"1"}}
```
dai-chen, line 1: np: do we really need so many test data to demonstrate the command usage in documentation?

ahkcs: I was thinking we can reuse this dataset for future doctest for timestamp

ahkcs: I updated to trim the dataset to 20 records

integ-test/src/test/resources/doctest/templates/dql/metadata.rst
```diff
@@ -10,3 +10,110 @@ Metadata Queries
    :depth: 1
 
 
+Querying Metadata
```
ykmr1224: What is this change for?

ykmr1224 (dismissed):

integ-test/src/test/resources/doctest/templates/dql/metadata.rst
ahkcs: Removed this change as this is a template

ykmr1224 (approved):

dai-chen:
Do we add and benchmark Big5 query for each new command like this?

ahkcs:
> Do we add and benchmark Big5 query for each new command like this?

I think we can add IT tests to benchmark Big5 queries like this, I'll open a separate PR for this:
https://github.com/opensearch-project/sql/pull/4163

ahkcs:
> Do we add and benchmark Big5 query for each new command like this?

For `Multisearch` command, I think the result will be exactly the same as `Append` command as they share the same implementation

RyanL1997 (approved):

PR #4333, opened 2025-09-19
Add max/min eval functions

ritvibhatt:
### Description

### Related Issues

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/udf/math/MaxFunction.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/udf/math/MaxFunction.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.expression.function.udf.math;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import org.apache.calcite.adapter.enumerable.NotNullImplementor;
+import org.apache.calcite.adapter.enumerable.NullPolicy;
+import org.apache.calcite.adapter.enumerable.RexToLixTranslator;
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.linq4j.tree.Expressions;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.type.SqlTypeFamily;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+import org.opensearch.sql.expression.function.UDFOperandMetadata;
+
+/**
+ * MAX(value1, value2, ...) returns the maximum value from the arguments.
+ * For mixed types, strings have higher precedence than numbers.
+ */
+public class MaxFunction extends ImplementorUDF {
+
+  public MaxFunction() {
+    super(new MaxImplementor(), NullPolicy.ANY);
+  }
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return opBinding -> opBinding.getTypeFactory().createSqlType(SqlTypeName.ANY);
+  }
+
+  @Override
+  public UDFOperandMetadata getOperandMetadata() {
+    return null;
+  }
+
+  public static class MaxImplementor implements NotNullImplementor {
+
+    @Override
+    public Expression implement(
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+      return Expressions.call(
+          MaxImplementor.class,
+          "max",
+          Expressions.newArrayInit(Object.class, translatedOperands));
+    }
+
+    public static Object max(Object[] args) {
+      return findMax(args);
+    }
+
+    private static Object findMax(Object[] args) {
+      if (args == null) {
+        return null;
+      }
+
+      return Arrays.stream(args)
+          .filter(Objects::nonNull)
+          .reduce(MaxImplementor::compareMax)
+          .orElse(null);
+    }
+
+    private static Object compareMax(Object a, Object b) {
+      boolean aIsNumeric = isNumeric(a);
+      boolean bIsNumeric = isNumeric(b);
+
+      if (aIsNumeric != bIsNumeric) {
+        return aIsNumeric ? b : a;
+      }
+      if (aIsNumeric) {
+        return ((Number) a).doubleValue() >= ((Number) b).doubleValue() ? a : b;
+      }
+      return a.toString().compareTo(b.toString()) >= 0 ? a : b;
+    }
+
+    private static boolean isNumeric(Object obj) {
+      return obj instanceof Number;
+    }
+  }
+}
--- a/core/src/main/java/org/opensearch/sql/expression/function/udf/math/MinFunction.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/udf/math/MinFunction.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.expression.function.udf.math;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import org.apache.calcite.adapter.enumerable.NotNullImplementor;
+import org.apache.calcite.adapter.enumerable.NullPolicy;
+import org.apache.calcite.adapter.enumerable.RexToLixTranslator;
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.linq4j.tree.Expressions;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.type.SqlTypeFamily;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+import org.opensearch.sql.expression.function.UDFOperandMetadata;
+
+/**
+ *MIN(value1, value2, ...) returns the minimum value from the arguments.
+ * For mixed types, numbers have higher precedence than strings.
+ */
+public class MinFunction extends ImplementorUDF {
+
+  public MinFunction() {
+    super(new MinImplementor(), NullPolicy.ANY);
+  }
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return opBinding -> opBinding.getTypeFactory().createSqlType(SqlTypeName.ANY);
+  }
+
+  @Override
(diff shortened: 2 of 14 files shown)


core/src/main/java/org/opensearch/sql/expression/function/udf/math/MaxFunction.java
```diff
@@ -0,0 +1,83 @@
...
+        return null;
+      }
+
+      return Arrays.stream(args)
+          .filter(Objects::nonNull)
+          .reduce(MaxImplementor::compareMax)
+          .orElse(null);
+    }
+
+    private static Object compareMax(Object a, Object b) {
```
RyanL1997: Do we need input validation in the `compareMax`/`compareMin` methods to handle edge cases:
- What happens if we get incomparable types (like arrays or objects)? <- is this possible?
- nit: Should we throw a more specific exception with a clear message in these cases?

core/src/main/java/org/opensearch/sql/expression/function/udf/math/MinFunction.java
```diff
@@ -0,0 +1,83 @@
...
+        return null;
+      }
+
+      return Arrays.stream(args)
+          .filter(Objects::nonNull)
+          .reduce(MinImplementor::compareMin)
+          .orElse(null);
+    }
+
+    private static Object compareMin(Object a, Object b) {
```
RyanL1997: see the above comment for `compareMax`.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
```diff
@@ -517,4 +517,106 @@ public void testValuesAggregationWithGroupBy() {
...
+            + " COMM=[$6], DEPTNO=[$7], a=[MIN(5, 30, $7, 'banana':VARCHAR, 'Door':VARCHAR, $1)])\n"
+            + "  LogicalTableScan(table=[[scott, EMP]])\n";
+    verifyLogical(root, expectedLogical);
+
+    String expectedSparkSql =
+        "SELECT `EMPNO`, `ENAME`, `JOB`, `MGR`, `HIREDATE`, `SAL`, `COMM`, `DEPTNO`, MIN(5, 30,"
+            + " `DEPTNO`, 'banana', 'Door', `ENAME`) `a`\n"
+            + "FROM `scott`.`EMP`";
+    verifyPPLToSparkSQL(root, expectedSparkSql);
+  }
```
RyanL1997, line 621: What should be the behavior with `null` values (will the `null` be ignored?):
```bash
source=accounts | eval result = MAX(age, null, 30) | fields age, result
```

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
```diff
@@ -517,4 +517,106 @@ public void testValuesAggregationWithGroupBy() {
...
+
+    String expectedSparkSql =
+        "SELECT `EMPNO`, `ENAME`, `JOB`, `MGR`, `HIREDATE`, `SAL`, `COMM`, `DEPTNO`, MIN('banana',"
+            + " 'Door', `ENAME`) `a`\n"
+            + "FROM `scott`.`EMP`";
+    verifyPPLToSparkSQL(root, expectedSparkSql);
+  }
+
+  @Test
+  public void testEvalMinOnNumericAndString() {
```
RyanL1997, line 607: What should be the behavior of mixed numeric types (int, long, float, double)?

e.g.
```bash
source=accounts | eval result = MAX(age, 30.5, 28L) | fields age, result
```

core/src/main/java/org/opensearch/sql/expression/function/udf/math/MaxFunction.java
ritvibhatt: For arrays/objects it gets converted to a string when comparing so it will be compared lexicographically

core/src/main/java/org/opensearch/sql/expression/function/udf/math/MinFunction.java
ritvibhatt: Same as max, arrays/objects it gets converted to a string when comparing so it will be compared lexicographically

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
ritvibhatt: Just updated so null values will be ignored unless all of the values are null, in which case it will return null. Added tests in ```CalcitePPLEvalMaxMinFunctionIT```

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
ritvibhatt: Should still be compared numerically and return the largest value from those

docs/user/ppl/functions/statistical.rst
```diff
@@ -0,0 +1,111 @@
...
+Statistical Functions
+======================
+
+.. rubric:: Table of contents
+
+.. contents::
+   :local:
+   :depth: 1
+
+.. versionadded:: 3.3.0
```
penghuo: the format looks not correct. https://github.com/ritvibhatt/sql/blob/7203266ac0d716e116e5ea0a9af1ed8902dda008/docs/user/ppl/functions/statistical.rst

core/src/main/java/org/opensearch/sql/expression/function/udf/math/MaxFunction.java
```diff
@@ -0,0 +1,83 @@
...
+      boolean aIsNumeric = isNumeric(a);
+      boolean bIsNumeric = isNumeric(b);
+
+      if (aIsNumeric != bIsNumeric) {
```
penghuo: what is expectataion is max(4, "2")? should be 4, right?

penghuo:
> A high level QQ: is supporting max(int, string) a hard requirement for now?
> 
> I'm just thinking otherwise:
> 
> 1. We probably can reuse the existing `array_min/max` function;
> 2. The implicit conversion (string -> numeric) can be supported in #4349. cc: @penghuo

array_min/max required all arguments have same type. 

max/min required speical handling, implicit cast may not work, e.g. the expactation is 
```
max(20, "4")    should return 20
max(20, "4a")   should return 4a
```
* convert fields to int does not work
```
array_max(20, 4)      should return 20
array_max(20, null)   should return 20
```

* convert fields to string does not work
```
array_max("20", "4")    should return "4"
array_max("20", "4a")   should return "4a"
```

dai-chen:
@ritvibhatt I synced with @penghuo offline. For Q2, it doesn’t seem to be a type conversion issue as I originally thought, but rather a data sorting one. For example, we could define a custom comparator in Java to handle the new sorting rule for numeric and string values, and then apply it in min/max/sort APIs. Perhaps we can do something similar within the Calcite accumulator?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
RyanL1997: Maybe relate to the solution of this https://github.com/opensearch-project/sql/issues/4349. Cuz I think 28L should be consider as a string. But yes, it is pointless to compare the string to numeric value from its concept.

RyanL1997: Not a blocker for now.

core/src/main/java/org/opensearch/sql/data/utils/MaxTypeComparator.java
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.data.utils;
+
+import java.util.Comparator;
+
+/** Comparator for MAX operations where strings have higher precedence than numbers. */
+public class MaxTypeComparator implements Comparator<Object> {
```
dai-chen: Is `MinTypeComparator` simply the inverse of `MaxTypeComparator`? I'm thinking of only one comparator and use it in Min/MaxFunction below by Java Stream's min/max(comparator) API.

Also this may worth adding dedicated UT to show its behavior for num-num, num-string, string-string etc.

core/src/main/java/org/opensearch/sql/data/utils/MixedTypeComparator.java
```diff
@@ -0,0 +1,38 @@
...
+  public int compare(Object a, Object b) {
+    boolean aIsNumeric = isNumeric(a);
+    boolean bIsNumeric = isNumeric(b);
+
+    // For same types compare directly
+    if (aIsNumeric == bIsNumeric) {
+      if (aIsNumeric) {
+        return Double.compare(((Number) a).doubleValue(), ((Number) b).doubleValue());
+      } else {
+        return Integer.compare(a.toString().compareTo(b.toString()), 0);
```
dai-chen, line 27: what does this Integer.compare with 0 meaning?

core/src/main/java/org/opensearch/sql/data/utils/MixedTypeComparator.java
```diff
@@ -0,0 +1,38 @@
...
+      }
+    }
+    // For mixed types, strings are considered larger than numbers (non-numeric values are treated
+    // as strings)
+    return aIsNumeric ? -1 : 1;
+  }
+
+  private static boolean isNumeric(Object obj) {
+    return obj instanceof Number;
```
dai-chen, line 36: Could you confirm whether comparison between "1" and "2" also considered as numerical comparison?

dai-chen (approved):
Thanks for the changes!

docs/user/ppl/functions/statistical.rst
ritvibhatt: Updated the doc

core/src/main/java/org/opensearch/sql/expression/function/udf/math/MaxFunction.java
ritvibhatt: Max(4, "2") should result in "2" since strings are always considered larger than numeric values

core/src/main/java/org/opensearch/sql/data/utils/MaxTypeComparator.java
ritvibhatt: Updated and added tests thank you!

core/src/main/java/org/opensearch/sql/data/utils/MixedTypeComparator.java
ritvibhatt: It will normalize it so if the string comparison returns a negative it will make it -1 and if it returns a positive it will be 1. Don't think that is necessary, can remove and just leave the string comparison

core/src/main/java/org/opensearch/sql/data/utils/MixedTypeComparator.java
ritvibhatt: No they will be compared as strings so max("9", "21") will return "9"

Swiddis (approved):

Issue #4282, opened 2025-09-11
[FEATURE] Push down bucket aggregation with sorting by metrics to `multi-terms` aggregation


LantaoJin:
**Is your feature request related to a problem?**
The query
```
| stats count() as cnt by a, b | sort cnt
```
will push down to a composite aggregation with 1000 buckets. So the `sort by cnt` only sorts on the 1000 records. The result is incorrect.

**What solution would you like?**
Push down such case to `multi-terms` aggregation https://docs.opensearch.org/latest/aggregations/bucket/multi-terms/

**What alternatives have you considered?**

**Do you have any additional context?**

LantaoJin:
For single bucket, it can be pushed to 
```
{
  "aggs": {
    "vi": {
      "terms": {
        "field": "vi",
        "size": 1000,
        "min_doc_count": 1,
        "order": {
          "a": "desc"
        }
      },
      "aggregations": {
        "a": {
          "sum": {
            "field": "age"
          }
        }
      }
    }
  }
}
```

penghuo:
@LantaoJin I found Composite aggregation perf could be better then Term aggregation in some case. for example, clickbench Q17, 
* Q17-Term run 2635ms, 
* Q17-Composite run 1325ms.

```
### Clear cache in between.
POST {{baseUrl}}/hits/_cache/clear

### Q17 - Term
GET {{baseUrl}}/hits/_search
Content-Type: application/x-ndjson

{
  "size": 0,
  "_source" : false,
  "track_total_hits" : -1,
  "aggs": {
    "user_groups": {
      "terms": {
        "field": "UserID",
        "size": 10
      },
      "aggs": {
        "search_phrase_groups": {
          "terms": {
            "field": "SearchPhrase",
            "size": 10
          }
        }
      }
    }
  }
}

### Q17 - Composite
GET {{baseUrl}}/hits/_search
Content-Type: application/x-ndjson

{
  "size" : 0,
  "_source" : false,
  "track_total_hits" : -1,
  "aggregations" : {
    "groupby" : {
      "composite" : {
        "size" : 10,
        "sources" : [
          {
            "UserIDTerm" : {
              "terms" : {
                "field" : "UserID",
                "missing_bucket" : true,
                "order" : "asc"
              }
            }
          },
          {
            "SearchPhraseTerm" : {
              "terms" : {
                "field" : "SearchPhrase",
                "missing_bucket" : true,
                "order" : "asc"
              }
            }
          }
        ]
      }
[...]

Issue #4278, opened 2025-09-11
[BUG] Sorting rules not applied to bucket aggregations


Swiddis:
**What is the bug?**
While messing with aggregation group-by on a dataset with 50000 records, I noticed that it's not processing all the buckets for queries:

```
> source = nginx_logs
   | stats count() by method, bytes
   | stats count();

fetched rows / total rows = 1/1
+---------+
| count() |
|---------|
| 1000    |  <--- Should be at least 10804, according to below
+---------+

> source = nginx_logs
   | dedup method, bytes
   | stats count();

fetched rows / total rows = 1/1
+---------+
| count() |
|---------|
| 10804   |
+---------+
```

The bucket count limitation is understandable (they're memory-intensive and shipping many thousands of buckets would exhaust bandwidth), but it does cause issues when trying to work with sorted aggregations. In this dataset, the top buckets should all be `GET`, not `DELETE`.

```
> source = nginx_logs
   | stats count() by method, bytes
   | sort -`count()`
   | head;

fetched rows / total rows = 10/10
+---------+--------+-------+
| count() | method | bytes |
|---------+--------+-------|
| 13      | DELETE | 81    |
| 12      | DELETE | 74    |
| 10      | DELETE | 71    |
| 10      | DELETE | 95    |
| 10      | DELETE | 51    |
| 10      | DELETE | 88    |
| 9       | DELETE | 78    |
| 9       | DELETE | 53    |
| 9       | DELETE | 49    |
| 9       | DELETE | 55    |
+---------+--------+-------+
```

This equivalently leads to incorrect `top` output:

```
> source = nginx_logs
   | top bytes by method
   | head;
   
fetched rows / total rows = 10/10
+--------+-------+-------+
| method | bytes | count |
|--------+-------+-------|
| DELETE | 81    | 13    |
| DELETE | 74    | 12    |
| DELETE | 51    | 10    |
| DELETE | 71    | 10    |
| DELETE | 88    | 10    |
| DELETE | 95    | 10    |
| DELETE | 44    | 9     |
| DELETE | 49    | 9     |
| DELETE | 52    | 9     |
| DELETE | 53    | 9     |
+--------+-------+-------+
```

**How can one reproduce the bug?**
1. Create a dataset with at least 1000 different buckets according to some sort of grouping
2. Observe incorrect results when trying to collect/sort them

**What is the expected behavior?**
[...]

PR #4335, opened 2025-09-19
Scaffolding: top/rare useOther and percentage arguments

Swiddis:
### Description
PR 1/3: this PR adds some parsing logic and does a bit of plumbing related to upcoming arguments. This doesn't include the actual implementation. useOther and showPerc will be the other two PRs.

Depending on how complex those are, I might also be able to hit #4278 and #4282 in the same go, since [composite buckets of terms support sorting by count](https://docs.opensearch.org/latest/aggregations/bucket/terms/). I'd also want to switch to the rare_terms aggregator for `rare`, per the warning on the linked doc.

Exactly how to get a total for other/percentage is still somewhat open. I thought earlier that I could subtract from a cheap `_count` call, but I didn't notice at the time that you could also supply filters before the inputs which invalidates that method. Going to poke at the DSL a bit more & see what we end up with (worst case we end up going by all returned buckets, which hits issues if there's too many buckets).

### Related Issues
N/A

### Check List


ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -301,12 +301,13 @@ logSpanValue
    : LOG_WITH_BASE                                                   # logWithBaseSpan
    ;
 
+// TODO support arbitrary argument ordering
```
Swiddis: todo: hitting this tomorrow, opening the PR for initial feedback.

core/src/main/java/org/opensearch/sql/ast/tree/args/RareTopNArguments.java
```diff
@@ -0,0 +1,112 @@
...
+  public static final String USE_OTHER = "useOther";
+
+  private int noOfResults = 10;
+  private String countField = "count";
+  private boolean showCount = true;
+  private String percentField = "percent";
+  private boolean showPerc = false;
+  private boolean useOther = false;
+
+  public RareTopNArguments(List<Argument> arguments) {
```
RyanL1997, line 43: btw do we need to handle case of invalid noOfResults? I know it is currently positive in the constructor. Is there a case that it can be set, for example, like -5?

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
```diff
@@ -279,15 +279,15 @@ public void testTopCommandWithNAndGroupBy() {
   public void testRareCommandWithGroupByWithCalcite() {
     when(settings.getSettingValue(Key.CALCITE_ENGINE_ENABLED)).thenReturn(true);
     assertEquals(
-        "source=t | rare 10 countield='count' showcount=true a by b",
```
RyanL1997: Just being curious: was this a typo in before?

Swiddis: Yeah the typo was there before

Swiddis:
Doc changes will come when the features are actually supported -- they'll currently just raise errors so no point documenting (they'd fail doctest anyway)

.pre-commit-config.yaml
```diff
@@ -3,11 +3,6 @@ repos:
     hooks:
       - id: spotless-format
         name: Spotless Format
-        entry: bash -c './gradlew spotlessApply && git add -u'
-        language: system
-        pass_filenames: false
-      - id: spotless-check
-        name: Spotless Post-format Check
-        entry: bash -c './gradlew spotlessCheck'
+        entry: bash -c './gradlew spotlessApply spotlessCheck && git add -u'
```
Swiddis, line 6: small build change, makes precommit faster (check always runs after apply)

Swiddis:
Closing as the feature has been deprioritized for a while & this keeps causing merge conflicts, will re-open if I work on it again

Issue #4058, opened 2025-08-18
[Enhancement] Better error handling for `JOIN` query limitation


RyanL1997:
## Description
According to the official documentation: [OpenSearch doc - JOIN does not support aggregations on the joined result](https://docs.opensearch.org/latest/search-plugins/sql/limitation/#join-does-not-support-aggregations-on-the-joined-result), JOIN queries in today's SQL plugin has certain limitations for aggregated join result. However, currently we do not have certain error handling to capture these limitations to warn users.

## Reproduce of the issue
- Env Setup - latest version of opensearch
- Construct a JOIN query with some aggregate functions, e.g.
```sql
SELECT entity.country_code, count(interaction_aggregate.interaction_id), count(if(interaction_aggregate.is_1to1visit, interaction_aggregate.interaction_id, null))
FROM interaction_aggregate
JOIN entity on entity.entity_key = interaction_aggregate.entity_key
GROUP BY entity.country_code LIMIT 10
```

- Execute the above query in the latest OS version, and from the response, it will output the column with aliasing issue - `count(b.channel) c` incorrectly generates duplicate columns `a.c` and `b.c` with `null` values instead of the expected aggregation result

- In the query explanation, the symptoms are:
  - No aggregation operator in the logical or physical plan
    - "`Project [ columns=[interaction_aggregate_0.COUNT, hcp_1.COUNT, hcp_1.country_code] ]`" in response indicates that the SQL engine is treating the `COUNT(...)` expressions as literal field names, not as functions to be evaluated, because it cannot resolve an aggregation function over joined data. 

  - `JOIN` followed by projection, not aggregation
    - The physical plan is using a `BlockHashJoin`, then directly projecting the `COUNT` columns from both sources
    ```bash
     "schema": [
       {"name": "interaction_aggregate_0.COUNT(interaction_id)", "type": "long"},
       ...
     ],
     "datarows": [[null, "CO", null], ...]
    ```

## Expectation/Exit Criteria
- SQL plugin can properly detect the usage of `JOIN` with aggregation
- Throw an error and point to the correct official documentation for user's awareness.
- Proper test cases for the enhanced behavior

RyanL1997:
Targeting  ~~`v3.4.0`~~ `v3.3.0`

PR #4344, opened 2025-09-22
[Enhancement] Add error handling for known limitation of sql `JOIN`

RyanL1997:
Signed-off-by: Jialiang Liang <jiallian@amazon.com>
### Description
Better error handling for known limitation of sql JOIN

example:
```bash
❯   curl -X POST "localhost:9200/_plugins/_sql" -H 'Content-Type: application/json' -d'
  {
    "query": "SELECT SUM(a.balance) FROM accounts a JOIN accounts b ON a.account_number = b.account_number"
  }'

{
  "error": {
    "details": "JOIN queries do not support aggregations on the joined result. For more information, see https://docs.opensearch.org/3.2/search-plugins/sql/limitation/#join-does-not-support-aggregations-on-the-joined-result",
    "type": "SqlParseException"
  },
  "status": 400
}% 
```

### Related Issues
* Resolve https://github.com/opensearch-project/sql/issues/4058
* Official doc: https://docs.opensearch.org/latest/search-plugins/sql/limitation/#join-does-not-support-aggregations-on-the-joined-result

### Check List

### DIFF
--- a/legacy/src/main/java/org/opensearch/sql/legacy/antlr/visitor/AntlrSqlParseTreeVisitor.java
+++ b/legacy/src/main/java/org/opensearch/sql/legacy/antlr/visitor/AntlrSqlParseTreeVisitor.java
@@ -291,21 +291,6 @@ public class AntlrSqlParseTreeVisitor<T extends Reducible>
           }
 
           @Override
-          public Boolean visitTableSources(TableSourcesContext ctx) {
-            if (ctx.tableSource().size() > 1) {
-              return true;
-            }
-
-            for (int i = 0; i < ctx.tableSource().size(); i++) {
-              Boolean hasJoinInTableSource = visit(ctx.tableSource(i));
-              if (hasJoinInTableSource != null && hasJoinInTableSource) {
-                return true;
-              }
-            }
-            return false;
-          }
-
-          @Override
           protected Boolean defaultResult() {
             return false;
           }
--- a/legacy/src/test/java/org/opensearch/sql/legacy/antlr/visitor/AntlrSqlParseTreeVisitorTest.java
+++ b/legacy/src/test/java/org/opensearch/sql/legacy/antlr/visitor/AntlrSqlParseTreeVisitorTest.java
@@ -108,13 +108,6 @@ public class AntlrSqlParseTreeVisitorTest {
   }
 
   @Test
-  public void visitImplicitJoinWithGroupByShouldThrowException() {
-    exceptionRule.expect(RuntimeException.class);
-    exceptionRule.expectMessage("JOIN queries do not support aggregations on the joined result.");
-    visit("SELECT COUNT(*) FROM testIndex t1, testIndex t2 WHERE t1.id = t2.id GROUP BY t1.field");
-  }
-
-  @Test
   public void visitExplicitJoinWithGroupByShouldThrowException() {
     exceptionRule.expect(RuntimeException.class);
     exceptionRule.expectMessage("JOIN queries do not support aggregations on the joined result.");


legacy/src/main/java/org/opensearch/sql/legacy/parser/SqlParser.java
```diff
@@ -399,10 +400,63 @@ public JoinSelect parseJoinSelect(SQLQueryExpr sqlExpr) throws SqlParseException
...
 
+  /**
+   * Validates that JOIN queries do not use aggregations (GROUP BY or aggregate functions). This
+   * limitation is documented in the official OpenSearch SQL documentation.
+   */
+  private void validateJoinWithoutAggregations(MySqlSelectQueryBlock query)
+      throws SqlParseException {
+    String errorMessage =
+        "JOIN queries do not support aggregations on the joined result. For more information, see"
+            + " https://docs.opensearch.org/latest/search-plugins/sql/limitation/#join-does-not-support-aggregations-on-the-joined-result";
```
Swiddis: suggestion: Can we fetch the cluster version and use it instead of `latest`?

If we ever fix this constraint, this error message will be invalid on all the older versions that output it

legacy/src/main/java/org/opensearch/sql/legacy/parser/SqlParser.java
```diff
@@ -399,10 +400,63 @@ public JoinSelect parseJoinSelect(SQLQueryExpr sqlExpr) throws SqlParseException
 
     updateJoinLimit(query.getLimit(), joinSelect);
 
-    // todo: throw error feature not supported:  no group bys on joins ?
+    validateJoinWithoutAggregations(query);
```
Swiddis: suggestion (if-minor): Should we move a lot of this logic to the construction of the `JoinSelect`?

Seems weird to construct the object and then use its setters to fill in everything, instead of just passing in the subquery block and letting it figure itself out. This also would guarantee that this validation is applied everywhere that JoinSelect is constructed.

legacy/src/test/java/org/opensearch/sql/legacy/unittest/parser/SqlParserTest.java
```diff
@@ -1784,4 +1784,134 @@ private boolean conditionExist(
     }
     return false;
   }
+
+  /** JOIN Aggregation Validation Tests */
+  private void expectJoinAggregationException() {
+    thrown.expect(SqlParseException.class);
+    thrown.expectMessage("JOIN queries do not support aggregations on the joined result.");
```
Swiddis: todo: This is a hard-coded copy of the validation's private error message

We should just make it public & reference it like we do with most of the other strings. I've had a few cases now of trying to clarify error messages and getting several pointless test failures for it

legacy/src/test/java/org/opensearch/sql/legacy/unittest/parser/SqlParserTest.java
```diff
@@ -1784,4 +1784,134 @@ private boolean conditionExist(
...
+  }
+
+  @Test
+  public void joinWithGroupByShouldThrowException() throws SqlParseException {
+    expectJoinAggregationException();
```
Swiddis: praise: Nice idea to make the exception expectation a shared function

I probably would have done a bunch of `thrown.expect` && `thrown.expectMessage(ERROR_MSG)` copies, this is nicer

legacy/src/test/java/org/opensearch/sql/legacy/unittest/parser/SqlParserTest.java
```diff
@@ -1784,4 +1784,134 @@ private boolean conditionExist(
...
+            "SELECT COUNT(*) FROM %s/account a "
+                + "JOIN %s/account b ON a.account_number = b.account_number",
+            TestsConstants.TEST_INDEX_ACCOUNT,
+            TestsConstants.TEST_INDEX_ACCOUNT);
+
+    parser.parseJoinSelect((SQLQueryExpr) queryToExpr(query));
+  }
+
+  @Test
+  public void joinWithSumFunctionShouldThrowException() throws SqlParseException {
```
Swiddis: suggestion: Since we're just leaning on `Select.AGGREGATE_FUNCTIONS`, we could reduce all these aggregation cases to a single [table-driven test](https://go.dev/wiki/TableDrivenTests)

legacy/src/main/java/org/opensearch/sql/legacy/parser/SqlParser.java
RyanL1997: good idea. implemented

legacy/src/test/java/org/opensearch/sql/legacy/unittest/parser/SqlParserTest.java
RyanL1997: I've extract it into the util class with the above dynamic versioning doc link url func.

legacy/src/test/java/org/opensearch/sql/legacy/unittest/parser/SqlParserTest.java
RyanL1997: good point, and I was actually thinking the same. Fixed.

Swiddis (approved):

legacy/src/main/java/org/opensearch/sql/legacy/parser/SqlParser.java
RyanL1997: good point. applied the change

legacy/src/main/java/org/opensearch/sql/legacy/domain/JoinSelect.java
```diff
@@ -30,6 +38,73 @@ public JoinSelect() {
...
+   * constructed from a query.
+   *
+   * @param query The select query block to validate
+   * @return A new JoinSelect instance
+   * @throws SqlParseException if the query contains aggregations (GROUP BY or aggregate functions)
+   */
+  public static JoinSelect withValidation(MySqlSelectQueryBlock query) throws SqlParseException {
+    validateJoinWithoutAggregations(query);
+    return new JoinSelect();
+  }
```
dai-chen: I think this new validation should be part of `AntlrSqlParseTreeVisitor`. To simplify, when we visit group by, we can confirm if any Join?

legacy/src/main/java/org/opensearch/sql/legacy/domain/JoinSelect.java
```diff
@@ -30,6 +38,73 @@ public JoinSelect() {
...
+      throw new SqlParseException(errorMessage);
+    }
+
+    if (query.getSelectList() != null) {
+      for (SQLSelectItem selectItem : query.getSelectList()) {
+        if (containsAggregateFunction(selectItem.getExpr())) {
+          throw new SqlParseException(errorMessage);
+        }
+      }
+    }
```
dai-chen: I'm just thinking if we don't support any post-processing after Join operation, should I validate in the opposite way, like above Join operator, the plan tree can only include Project, Limit etc?

legacy/src/main/java/org/opensearch/sql/legacy/utils/Util.java
```diff
@@ -41,6 +41,16 @@ public class Util {
...
 
+  public static final String JOIN_AGGREGATION_ERROR_PREFIX =
+      "JOIN queries do not support aggregations on the joined result.";
+
+  public static final String DOC_REDIRECT_MESSAGE = " For more information, see ";
+
+  public static final String OPENSEARCH_DOC_BASE_URL = "https://docs.opensearch.org/";
+
+  public static final String JOIN_AGGREGATION_DOC_PATH =
+      "/search-plugins/sql/limitation/#join-does-not-support-aggregations-on-the-joined-result";
```
dai-chen: Not sure if any way we can make sure this won't be broken.

RyanL1997: +1 to the concern. However, I'd say to avoid these links in general we actually need some proper link checker for our docs. I know that OSD has some similar framework. - which is kinda off the scope of this task. 

The good thing is that I have implemented the dynamic versioning of the doc link (instead of using the `latest` I was using the current version of OS, and I think for the released version's doc, highly likely we will not change the structure of the url.

Swiddis: If we want to be extra safe we can just remove the `#` anchor and link straight to limitations. The anchor is coupled to the section title and seems likely to change, the overall `limitation` link is probably stable.

legacy/src/main/java/org/opensearch/sql/legacy/domain/JoinSelect.java
RyanL1997: I found the following:

1. Two Parser Systems:
  - ANTLR Parser: `AntlrSqlParseTreeVisitor` - Used for semantic analysis, type checking
  - Druid Parser: `SqlParser.java` - Used for actual query execution (legacy)
2. `JOIN` Processing Flow:
  - OpenSearchActionFactory.createAction() determines if it's a `JOIN` query
  - If `JOIN`: Uses `Druid` parser → `SqlParser.parseJoinSelect()` → Our current validation
  - If not `JOIN`: Uses regular select parsing
3. Current JOIN Validation:
  - Happens in Druid-based `SqlParser` during `parseJoinSelect()`
  - The current implementation - validation is in `JoinSelect.withValidation()`
  
So the ANTLR visitor approach would be for different use cases (like type checking, semantic analysis), but `JOIN` execution uses the Druid parser. But yes, I can try to switch to the Antlr parsing approach and trying to remove the current approach. SInce the the issue will be catched earlier before it reaches to the parser we are working on.

legacy/src/main/java/org/opensearch/sql/legacy/domain/JoinSelect.java
RyanL1997: fixed

legacy/src/main/java/org/opensearch/sql/legacy/domain/JoinSelect.java
RyanL1997: fixed and moved the checking logic to the antlr visitor layer

legacy/src/main/java/org/opensearch/sql/legacy/utils/Util.java
RyanL1997: fixed

legacy/src/main/java/org/opensearch/sql/legacy/antlr/visitor/AntlrSqlParseTreeVisitor.java
```diff
@@ -264,6 +266,93 @@ public T visitFunctionNameBase(FunctionNameBaseContext ctx) {
     return visitor.visitFunctionName(ctx.getText());
   }
 
+  @Override
+  public T visitGroupByItem(OpenSearchLegacySqlParser.GroupByItemContext ctx) {
+    if (hasJoinInQuery(ctx)) {
```
dai-chen: I think the current approach of detecting JOINs is unnecessarily complex. Instead, we can simply do it by a small visitor as below?

```
  @Override
  public T visitGroupByItem(OpenSearchLegacySqlParser.GroupByItemContext ctx) {
      // Get the parent context (typically the FROM clause area)
      ParserRuleContext fromClause = ctx.getParent();

      // Check for JOINs using a focused visitor
      boolean hasJoin = fromClause.accept(new OpenSearchLegacySqlParserBaseVisitor<Boolean>() {
          @Override
          public Boolean visitTableSourceBase(TableSourceBaseContext ctx) {
              return !ctx.joinPart().isEmpty(); // Found a JOIN if joinPart is not empty
          }

          @Override
          protected Boolean defaultResult() {
              return false;
          }

          @Override
          protected Boolean aggregateResult(Boolean aggregate, Boolean nextResult) {
              return aggregate || nextResult;
          }
      });
```

RyanL1997: good call. fixed. can you take a look again?

dai-chen (commented):
Did you remove your test accidentally? I don't see any UT or IT now.

RyanL1997:
Added the new tests in `AntlrSqlParseTreeVisitorTest.java` cc @dai-chen

legacy/src/main/java/org/opensearch/sql/legacy/antlr/visitor/AntlrSqlParseTreeVisitor.java
```diff
@@ -299,6 +282,41 @@ protected Boolean aggregateResult(Boolean aggregate, Boolean nextResult) {
...
+              return true;
+            }
+
+            for (int i = 0; i < ctx.tableSource().size(); i++) {
+              Boolean hasJoinInTableSource = visit(ctx.tableSource(i));
+              if (hasJoinInTableSource != null && hasJoinInTableSource) {
+                return true;
+              }
+            }
+            return false;
```
dai-chen: what's this trying to check?

RyanL1997: this code is checking for explicit `JOINs` within individual table sources -  This loop is iterating through each `tableSource() `and calling `visit()` on it, which will eventually call `visitTableSourceBase()` if the table source has join parts... However, I think we should just do
```java
            return ctx.tableSource().size() > 1;
```
instead.

RyanL1997: transferring some offline communication with @dai-chen here: we should limit the scope of this for only check the explicit join only - so removed this section.

dai-chen (approved):

Swiddis (approved):

RyanL1997:
#4378 is failing at the same

Swiddis:
Failures exist in main and are only with recently added tests, think this is ok to merge

PR #4345, opened 2025-09-22
Implement one-batch lookahead for index enumerators

Swiddis:
### Description
In local benchmarking of merge operations, I saw we were spending a lot of time waiting for synchronous fetching of batches across both indices.

Because of the PIT-based design, we can't parallelize page fetches directly, but one low-hanging fruit here is to start fetching the next batch as soon as we get the current one. This cuts enumerated merge times by ~40%.

Some alternatives for the long-term:
- Implement range-based/adaptive parallel fetching
- Skip paginating with the rest client and just go directly through Lucene
- Core is working on streaming queries: https://github.com/opensearch-project/OpenSearch/issues/18725

In draft pending testing.

### Related Issues
N/A

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
@@ -8,6 +8,8 @@ package org.opensearch.sql.opensearch.storage.scan;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
 import org.apache.calcite.linq4j.Enumerator;
@@ -53,6 +55,9 @@ public class OpenSearchIndexEnumerator implements Enumerator<Object> {
 
   private ExprValue current;
 
+  private CompletableFuture<OpenSearchResponse> nextBatchFuture;
+  private boolean isLastBatch = false;
+
   public OpenSearchIndexEnumerator(
       OpenSearchClient client,
       List<String> fields,
@@ -69,14 +74,29 @@ public class OpenSearchIndexEnumerator implements Enumerator<Object> {
     if (!this.monitor.isHealthy()) {
       throw new NonFallbackCalciteException("insufficient resources to run the query, quit.");
     }
+    this.nextBatchFuture = CompletableFuture.supplyAsync(() -> client.search(request));
   }
 
   private void fetchNextBatch() {
-    OpenSearchResponse response = client.search(request);
-    if (!response.isEmpty()) {
-      iterator = response.iterator();
-    } else if (iterator == null) {
-      iterator = Collections.emptyIterator();
+    try {
+      // Get results from the pre-fetched batch
+      OpenSearchResponse response = nextBatchFuture.get();
+
+      if (!response.isEmpty()) {
+        iterator = response.iterator();
+
+        // If we haven't hit the end, start pre-fetching next batch
+        if (!isLastBatch) {
+          nextBatchFuture = CompletableFuture.supplyAsync(() -> client.search(request));
+        }
+      } else {
+        if (iterator == null) {
+          iterator = Collections.emptyIterator();
+        }
+        isLastBatch = true;
+      }
+    } catch (InterruptedException | ExecutionException e) {
+      throw new NonFallbackCalciteException("Error fetching batch: " + e.getMessage());
     }
   }
 
@@ -98,6 +118,7 @@ public class OpenSearchIndexEnumerator implements Enumerator<Object> {
   @Override
   public boolean moveNext() {
     if (queryCount >= maxResponseSize) {
+      isLastBatch = true;
       return false;
     }
 
@@ -114,17 +135,21 @@ public class OpenSearchIndexEnumerator implements Enumerator<Object> {
       queryCount++;
       return true;
     } else {
+      isLastBatch = true;
       return false;
     }
   }
 
   @Override
   public void reset() {
+    isLastBatch = false;
+    nextBatchFuture = CompletableFuture.supplyAsync(() -> client.search(request));
     OpenSearchResponse response = client.search(request);
     if (!response.isEmpty()) {
       iterator = response.iterator();
     } else {
       iterator = Collections.emptyIterator();
+      isLastBatch = true;
     }
     queryCount = 0;
   }
@@ -133,6 +158,10 @@ public class OpenSearchIndexEnumerator implements Enumerator<Object> {
   public void close() {
     iterator = Collections.emptyIterator();
     queryCount = 0;
+    isLastBatch = true;
+    if (nextBatchFuture != null) {
+      nextBatchFuture.cancel(true);
+    }
     client.cleanup(request);
   }
 }


Swiddis:
Security IT failures are confusing me here -- seems like they're all consistently failing but the changed code doesn't show up anywhere in any of the stack traces

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -277,13 +277,23 @@ public ScheduledJobParser getJobParser() {
...
+    // prevent deadlocks during async fetches on small node counts. Tasks in the background pool
+    // should do no work except I/O to other services.
+    return List.of(
         new FixedExecutorBuilder(
             settings,
-            AsyncRestExecutor.SQL_WORKER_THREAD_POOL_NAME,
+            SQL_WORKER_THREAD_POOL_NAME,
             OpenSearchExecutors.allocatedProcessors(settings),
             1000,
-            null));
```
Swiddis, line 324: This previous null prefix meant that we were literally registering the global YML setting `null` to configure the SQL thread pool:

```yml
null:
  size: 30
```

I updated this to use the proper thread_pool setting prefix with accurate naming. This also meant tweaking the thread pool to be named with an underscore, which should be a safe change. New settings are shown in the docs change.

Swiddis:
Some additional testing info:

I took 5 million records from the big5 benchmarking dataset and compared the current `mainline` with this.

First, as sanity, the results are the same for one of the queries requiring a full index enumeration:

```
source = big5
| eval range_bucket = case(
   `metrics.size` < -10, 'range_1',
   `metrics.size` >= -10 and `metrics.size` < 10, 'range_2',
   `metrics.size` >= 10 and `metrics.size` < 100, 'range_3',
   `metrics.size` >= 100 and `metrics.size` < 1000, 'range_4',
   `metrics.size` >= 1000 and `metrics.size` < 2000, 'range_5',
   `metrics.size` >= 2000, 'range_6')
| stats count() by range_bucket, span(`@timestamp`, 1h) as auto_span
| sort + range_bucket, + auto_span

Current mainline:

fetched rows / total rows = 48/48
+---------+---------------------+--------------+
| count() | auto_span           | range_bucket |
|---------+---------------------+--------------|
| 122464  | 2022-12-31 16:00:00 | range_5      |
| 121585  | 2022-12-31 17:00:00 | range_5      |
| 122052  | 2022-12-31 18:00:00 | range_5      |
| 122220  | 2022-12-31 19:00:00 | range_5      |
| 122163  | 2022-12-31 20:00:00 | range_5      |
| 121840  | 2022-12-31 21:00:00 | range_5      |
| 121606  | 2022-12-31 22:00:00 | range_5      |
| 121889  | 2022-12-31 23:00:00 | range_5      |
| 121088  | 2023-01-01 00:00:00 | range_5      |
| 121943  | 2023-01-01 01:00:00 | range_5      |

After update:

fetched rows / total rows = 48/48
+---------+---------------------+--------------+
| count() | auto_span           | range_bucket |
|---------+---------------------+--------------|
| 122464  | 2022-12-31 16:00:00 | range_5      |
| 121585  | 2022-12-31 17:00:00 | range_5      |
| 122052  | 2022-12-31 18:00:00 | range_5      |
| 122220  | 2022-12-31 19:00:00 | range_5      |
| 122163  | 2022-12-31 20:00:00 | range_5      |
| 121840  | 2022-12-31 21:00:00 | range_5      |
| 121606  | 2022-12-31 22:00:00 | range_5      |
| 121889  | 2022-12-31 23:00:00 | range_5      |
| 121088  | 2023-01-01 00:00:00 | range_5      |
| 121943  | 2023-01-01 01:00:00 | range_5      |
```

Second, I wanted to benchmark and check for impact. I already tested with joins and it's ~40% faster, but for non-joins we potentially pay overhead for nothing.

For the slowest big5 queries (BG fetches on the left, sync fetches on the right), we see slight perf gains:
image

For the fastest ones, the performance is approximately the same (some minor latency and throughput diffs but I'm not confident that this isn't just random variation):
image

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
vamsimanohar: unfortunate.

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -277,13 +277,23 @@ public ScheduledJobParser getJobParser() {
...
+    // The worker pool is the primary pool where most of the work is done. The background thread
+    // pool is a separate queue for asynchronous requests to other nodes. We keep them separate to
+    // prevent deadlocks during async fetches on small node counts. Tasks in the background pool
+    // should do no work except I/O to other services.
+    return List.of(
         new FixedExecutorBuilder(
             settings,
-            AsyncRestExecutor.SQL_WORKER_THREAD_POOL_NAME,
+            SQL_WORKER_THREAD_POOL_NAME,
             OpenSearchExecutors.allocatedProcessors(settings),
```
vamsimanohar, line 326: I am not sure if this is the best number.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
```diff
@@ -45,6 +46,7 @@
 import org.opensearch.sql.planner.logical.LogicalPlan;
 import org.opensearch.sql.planner.physical.PhysicalPlan;
 import org.opensearch.sql.storage.read.TableScanBuilder;
+import org.opensearch.transport.client.node.NodeClient;
 
 /** OpenSearch table (index) implementation. */
 public class OpenSearchIndex extends AbstractOpenSearchTable {
```
vamsimanohar, line 53: This shows that we have good number of use cases dependent on NodeClient.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/BackgroundSearchScanner.java
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Initiates the scanning process. If async operations are available, this will trigger the first
+   * background fetch.
+   *
+   * @param request The OpenSearch request to execute
+   */
+  public void startScanning(OpenSearchRequest request) {
+    if (isAsync()) {
+      nextBatchFuture =
+          CompletableFuture.supplyAsync(() -> client.search(request), backgroundExecutor);
```
vamsimanohar, line 99: I am surprised this is working without copying the thread context  like this: https://github.com/opensearch-project/sql/blob/main/opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchQueryManager.java#L39

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -277,13 +277,23 @@ public ScheduledJobParser getJobParser() {
...
             1000,
-            null));
+            "thread_pool." + SQL_WORKER_THREAD_POOL_NAME),
+        new FixedExecutorBuilder(
```
vamsimanohar, line 329: How would it be different if we use the same thread pool or increase the size of the same pool and submit the task to same.


Threads from all the pools on the node will take turns for the same cores available.

`Now double the number of threads from the sql plugin will be contending for the cpu.`  Just food for thought: is it good or bad?

Now that PPL is becoming front and center in discover I don't mind doing this.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/BackgroundSearchScanner.java
vamsimanohar: What am I missing? Why do we need context copy in other file. Can you print the thread context in this task and see if it has user credentials in case of FGAC?

PPLPermissionsIT can we add a join test here.

vamsimanohar:
Added few comments. Good one 👍 .

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/BackgroundSearchScanner.java
Swiddis: This isn't just used for joins, every query goes through this interface. I realized while benchmarking that there were no queries that weren't hitting the pool. That the security ITs pass means either this works or we don't have security ITs.

I believe it works because we supply the cluster settings during the construction of the executor, so it's built-in to the thread context (as opposed to starting a fresh thread with no executor)

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
Swiddis: The background threads have very light work, it's just serialization costs. They'll take virtually no resources compared to the search and worker threads. Modern threads are cheap so they'll pretty much just sit around in the process table waiting for the network response event.

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
Swiddis: Ideally it should match the number of search threads since that's where all the requests go, maybe I can find where that number is stored and do a lookup.

Swiddis: Updated to pull the search thread pool count if available, otherwise fallback to node processors. This is what it looks like if you limit the search thread pool under heavy load:

image

Intuitively this seems like a pretty informative view of what state the cluster's in regarding SQL queries.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchQueryManager.java
```diff
@@ -21,7 +21,8 @@ public class OpenSearchQueryManager implements QueryManager {
 
   private final NodeClient nodeClient;
 
-  private static final String SQL_WORKER_THREAD_POOL_NAME = "sql-worker";
+  public static final String SQL_WORKER_THREAD_POOL_NAME = "sql_worker";
```
penghuo: [SQL has metrics](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/managedomains-cloudwatchmetrics.html#managedomains-cloudwatchmetrics-sql) depend on sql-worker. Could u double confirm name change with not break the metrics?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
```diff
@@ -66,33 +65,24 @@ public OpenSearchIndexEnumerator(
...
     this.maxResultWindow = maxResultWindow;
     this.monitor = monitor;
-    this.queryCount = 0;
-    this.current = null;
-    if (!this.monitor.isHealthy()) {
-      throw new NonFallbackCalciteException("insufficient resources to run the query, quit.");
-    }
+    this.client = client;
+    this.bgScanner = new BackgroundSearchScanner(client);
+    this.bgScanner.startScanning(request);
```
penghuo, line 78: This constructor only been called once per query? I found comments, Could u double confirm? 
```
  /**
   * This Enumerator may be iterated for multiple times, so we need to create opensearch request for
   * each time to avoid reusing source builder. That's because the source builder has stats like PIT
   * or SearchAfter recorded during previous search.
   */
  @Override
  public Enumerable<@Nullable Object> scan() {
    return new AbstractEnumerable<>() {
      @Override
      public Enumerator<Object> enumerator() {
        OpenSearchRequestBuilder requestBuilder = getOrCreateRequestBuilder();
        return new OpenSearchIndexEnumerator(
            osIndex.getClient(),
            getFieldPath(),
            requestBuilder.getMaxResponseSize(),
            requestBuilder.getMaxResultWindow(),
            osIndex.buildRequest(requestBuilder),
            osIndex.createOpenSearchResourceMonitor());
      }
    };
  }
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/BackgroundSearchScanner.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    Iterator<ExprValue> iterator;
+    if (!response.isEmpty()) {
+      iterator = response.iterator();
+
+      // Pre-fetch next batch if needed
+      if (!stopIteration && isAsync()) {
+        nextBatchFuture =
+            CompletableFuture.supplyAsync(() -> client.search(request), backgroundExecutor);
```
penghuo, line 144: what if fixedThreadPool full, should fallback to sync?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
```diff
@@ -135,6 +125,7 @@ public boolean moveNext() {
 
   @Override
   public void reset() {
+    bgScanner.reset(request);
     OpenSearchResponse response = client.search(request);
```
penghuo: Why use client.search instead of bgScanner.fetchNextBatch in this case?

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchQueryManager.java
Swiddis: Oh, that's annoying... This thread pool is as far as I know the only one in OpenSearch to use a hyphen instead of an underscore :/

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
Swiddis: It's the same as current behavior, right? If you recreate the enumerator with a new client, you erase all of its current state and start a new search. In that snippet it looks like this is deliberately meant to restart the search multiple times

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/BackgroundSearchScanner.java
Swiddis: We just buffer: https://github.com/opensearch-project/sql/pull/4345#discussion_r2383251023

If we fallback to sync, it eliminates the utility of being able to directly view/control the active SQL network requests via the BG thread pool

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
Swiddis: Since it's the same client in both instances, it doesn't make a difference (fetchNextBatch would force a sync fetch). I can change it for consistency

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
penghuo: The concerns is, If scan() metheod is been called multiple times in planning stage, it will invoke startScanning multiple times.

Swiddis: Woah, I wouldn't have expected planning to make a call to scan, seems weird... Can try to find a better way to handle that, but scan intuitively to me means "actually start scanning something"

Swiddis:
Turns out I flipped the benchmark in my head, so this is overall a regression -- going to put back in draft and figure out a better approach

penghuo (approved):

LantaoJin:
> Turns out I flipped the benchmark in my head, so this is overall a regression -- going to put back in draft and figure out a better approach

@Swiddis , do you mean the current implementation has performance regression? So why the PR merged finally? If there is no regression, please backport it to 2.19-dev since backporting of #4884 is blocked by this backporting.

Swiddis:
Couldn't get the regression to reliably repro & the regression was smaller than the gains in more tests, so I wanted to see what the diff was in the OSB benchmarks. I don't see any benchmark diff since merging.

Wasn't planning on backporting this originally since it's still largely experimental, can open the PR at least

Issue #4312, opened 2025-09-17
[FEATURE] Enhance cost computing mechanism in AbstractCalciteIndexScan


qianheng-aws:
**Is your feature request related to a problem?**
The current cost computing mechanism in `AbstractCalciteIndexScan` should be out-of-date after we've introduced so many push down rules for variety cases.

Some bad cases: 
1. Prefer more scripts than less script with additional project operator: https://github.com/opensearch-project/sql/pull/4279#discussion_r2354020132
2. Prefer `scan(agg+project) + agg` than `scan(agg)`
3. ...

The shortcoming of current mechanism:
1. Mix rows estimation with the cost computing in one place
2. Lack of the delicate cost computing for scripts

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

PR #4279, opened 2025-09-11
Push down project operator with non-identity projections into scan

qianheng-aws:
### Description
This PR aims to push down or eliminate the project upon the scan with aggregate pushed down. 

Since we will always append a project operator after aggragtion for adjusting the order of its output columns, this optimization will enhance this common case to have only one `scan` operator, then it is able to leverage our https://github.com/opensearch-project/sql/pull/3853 to skip codegen.

e.g.
PPL:
```
PPL: 
source=opensearch-sql_test_index_bank | stats count() by span(birthdate,1M)
```
Before this PR:
```
FINAL PLAN:
EnumerableCalc(expr#0..1=[{inputs}], count()=[$t1], span(birthdate,1M)=[$t0])
  CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_bank]], PushDownContext=[[AGGREGATION->...
```

With this PR:
```
FINAL PLAN:
CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_bank]], PushDownContext=[[AGGREGATION->..., Project->[count(), span(birthdate,1M)]] ...]
```

### Related Issues
Resolves 
https://github.com/opensearch-project/sql/issues/4230
https://github.com/opensearch-project/sql/issues/4296
### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4353, opened 2025-09-23
Enhance the cost computing mechanism and push down context

qianheng-aws:
### Description
This PR includes changes:
1. Support cacheing OpenSearchRequestBuilder inner PushContext. Otherwise it will be built twice for the cheapest plan.
2. Ehance the cost computing mechanism by:
    - Split method `computeSelfCost` from `estimateRowCount`. In `computeSelfCost`, we've implemented more proper cost computing mechanism rather than relying on rows count only.
    - Support more accurate script cost by involving script count rather than the status of whether containing scripts.
3. Move part of our explain IT from json format to yaml format.

The new cost computing mechanism has advantages in several cases:
- The cost among scans with the same OpenSearch request will get equivilant cost. For example, before this PR, `Scan(PushDownContext[LIMIT, PROJECT])` will get smaller cost than `Scan(PushDownContext[PROJECT, LIMIT])` although they produce the same request. With this PR, they two will get the equivilent cost. And then the optimizer will finally choose the plan with push down sequence more close to the original logical plan.
-  Before this PR, the optimizer will choose plan like `Scan(PushDownContext[AGG, PROJECT]) - Aggregate` although the 2 agg could be merged. With this PR, the optimizer will choose the scan with a merged agg pushed down.
- Before this PR, the cases like `https://github.com/opensearch-project/sql/pull/4279#discussion_r2354020132` with several scripts in our agg builder, the optimizer will still choose it since scirpt count won't affect the cost of the plan. This PR will address it.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/4312

### Check List


core/src/main/java/org/opensearch/sql/utils/YamlFormatter.java
```diff
@@ -27,7 +26,6 @@ public class YamlFormatter {
         YAMLGenerator.Feature.ALWAYS_QUOTE_NUMBERS_AS_STRINGS); // Quote numeric strings
     yamlFactory.enable(YAMLGenerator.Feature.INDENT_ARRAYS_WITH_INDICATOR);
     YAML_MAPPER = new ObjectMapper(yamlFactory);
-    YAML_MAPPER.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
```
qianheng-aws, line 30: Remove this config, otherwise for the plan of v2 which contains multi-children for one level, it will put the children at beginning and the operator name at the end

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PushDownContext.java
```diff
@@ -0,0 +1,343 @@
...
+
+  void add(PushDownType type, Object digest, AbstractAction<?> action) {
+    add(new PushDownOperation(type, digest, action));
+  }
+
+  public boolean containsDigest(Object digest) {
+    return this.stream().anyMatch(action -> action.digest().equals(digest));
+  }
+
+  public synchronized OpenSearchRequestBuilder createOrGetRequestBuilder() {
```
LantaoJin: add java doc

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -109,8 +91,84 @@ protected Integer getQuerySizeLimit() {
...
+   *
+   * <p>While the left operations like project, filter, limit will be ignored in the accumulation
+   * process. But they will also affect the cost of cost-counted operations after them and the final
+   * external cost, which is calculated by `rows count * fields count`.
+   *
+   * <p>In the end, we still need to multiply the total cost by a factor to make the cost cheaper
+   * than non-pushdown operators.
+   */
+  @Override
+  public @Nullable RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
```
LantaoJin, line 133: please add a UT for these two method to ensure the behaviour would be stable.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PushDownContext.java
LantaoJin: sync offline, let' move this method out

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -109,8 +91,84 @@ protected Integer getQuerySizeLimit() {
...
+        case AGGREGATION -> {
+          dRows = mq.getRowCount((RelNode) operation.digest());
+          dCpu += dRows * getAggMultiplier(operation);
+        }
+          // Ignored Project in cost accumulation, but it will affect the external cost
+        case PROJECT -> {}
+        case SORT -> dCpu += dRows;
+          // Refer the org.apache.calcite.rel.metadata.RelMdRowCount.getRowCount(Aggregate rel,...)
+        case COLLAPSE -> {
+          dRows = dRows / 10;
```
LantaoJin, line 146: what the diff with https://github.com/opensearch-project/sql/pull/4353/files#diff-70fd6c948cefa04f6fa28ddd2ccf0e85d6cbce9fa8153b995628e186b1bd918dR109?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -109,8 +91,84 @@ protected Integer getQuerySizeLimit() {
...
+          dCpu +=
+              NumberUtil.multiply(
+                  NumberUtil.multiply(dRows, RelMdUtil.guessSelectivity(filterDigest.condition())),
+                  Math.pow(1.1, filterDigest.scriptCount()));
+        }
+          // Ignore cost the LIMIT but it will affect the rows count.
+          // Try to reduce the rows count by 1 to make the cost cheaper slightly than non-push down.
+          // Because we'd like to push down LIMIT even when the fetch in LIMIT is greater than
+          // dRows.
+        case LIMIT -> dRows = Math.min(dRows, ((LimitDigest) operation.digest()).limit()) - 1;
```
LantaoJin, line 164: could be negative with `-1`?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -109,8 +91,84 @@ protected Integer getQuerySizeLimit() {
...
+            NumberUtil.multiply(
+                dRows, RelMdUtil.guessSelectivity(((FilterDigest) operation.digest()).condition()));
+        case SCRIPT -> {
+          FilterDigest filterDigest = (FilterDigest) operation.digest();
+          // Calculate the cost of script filter by multiplying the selectivity of the filter and
+          // the factor amplified by script count.
+          dCpu +=
+              NumberUtil.multiply(
+                  NumberUtil.multiply(dRows, RelMdUtil.guessSelectivity(filterDigest.condition())),
+                  Math.pow(1.1, filterDigest.scriptCount()));
```
yuancu: Will `Math.pow(1.1, filterDigest.scriptCount() - 1)` be better? It ensures that when there is only one script, the cost of pushed down script filter is smaller than that the same filter in memory ( `dRow * guessedFactor * estimateRowCountFactor < dRow * guessedFactor`, otherwise it's something like `dRow * guessedFactor * 1.1 * estimateRowCountFactor`, which is not necessarily smaller than `dRow * guessedFactor`)

qianheng-aws: The default value of `estimateRowCountFactor`  is 0.9, so the cost of pushed-down filter is smaller.

On the other hand, the biggest reduction of cost from pushed-down filter is not from the cost reduction on filter iteself, but the final output rows count of the scan with filter push down.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
qianheng-aws: It could be but `VolcanoPlanner` in Clacite will replace it with `TINY` cost it's less than 0.

qianheng-aws:
@LantaoJin  There are some enhancement on plan with this PR:
explain_agg_with_sum_enhancement.yaml
explain_timechart_count.yaml
explain_filter_then_limit_push.yaml
explain_agg_counts_by1
explain_agg_counts_by2
explain_filter_push.yaml
explain_filter_push_compare_date_string.yaml
explain_filter_push_compare_time_string.yaml
explain_filter_push_compare_timestamp_string.yaml

And https://github.com/opensearch-project/sql/pull/4377 is no longer needed after this PR. I've removed the code and keep the yaml test only.

yuancu (approved):

Issue #4307, opened 2025-09-16
(Calcite PPL) Spath dynamic columns support


ykmr1224:
Support dynamic columns use case with `spath` PPL command.

Example: following `spath` command will parse `doc` attribute and extract all the attributes as columns.
```
source=idx
| spath index=doc
```
Each item could be any type since JSON attribute could be string/number/array/null

penghuo:
> source=idx
| spath input=doc

@ykmr1224  if we add change requirement of spath as 
* input: mandatory. The field to scan for JSON data.
* output: The destination field that the data will be loaded to. Default: value of path.
  * optional, if path  exist.
  * mandatory, if path does not exist.
* path: The path of the data to load for the object. 
  * optional if output exist.
  * mandatory if output does not exist. 

then PPL does not need to support dynamic columns. does it sounds as workaround?

penghuo:
**Proposal - Remove dynamic-column expansion**
- Add extract-all-to-map mode for `spath` that flattens JSON into a single map field. Remove dynamic-column expansion entirely.

**Behavior**
- Path-based (unchanged): `spath input=f [output=o] path=p` → `o = json_extract(f, p)`; default `output = path`.
- Extract-all: `spath input=f output=o` or `spath input=f` → `o = json_extract_all(f)`; default `o = f`.
- No dynamic columns: `input`-only no longer expands columns; it produces a map instead.

**Flattening & Merging**
- Dotted keys for objects (`a.b = 1`).
- Arrays use `{}` suffix (`list{} = [1,2]`; `list{}.x = [1,2]`).
- Duplicate logical keys merge into arrays (e.g., `a.b: [1,2]`).
- Nulls preserved; key order not guaranteed.

**Types**
- Return `map<string,string>`: stringify all values for consistency now.
- Consider `map<string,any>` as a future enhancement.

**Compatibility**
- Syntax remains compatible.
- Semantic change: `spath input=f` now yields a map in `f` (not dynamic columns). Clearly document removal of dynamic-field expansion and the new default.

**Corner Cases**
- Invalid or empty JSON → null map.
- Large payloads: still runs on coordinator; filter before `spath`.

**Example**
- Input: `{"body": "{\"a\":1, \"b\":2}"}`
  - `| spath input=body output=body` → `body = {"a":"1","b":"2"}`

anasalkouz:
I agree with @penghuo proposal. Let's remove the Dynamic field expansion and use output param. The only concern I have in this proposal is 

[...]

PR #4355, opened 2025-09-23
Support spath command without path param by dynamic columns

ykmr1224:
### Description
- Implement dynamic columns support, which allows commands to generate columns dynamically.
  - Dynamic columns are stored in `_dynamic_columns` column typed as MAP<String, Any>
  - `MAP_MERGE` function is used to merge `_dynamic_columns` for the case `_dynamic_columns` fields already exist.
  - `MAP_GET` function is used to access dynamic columns. (field access will fallback to MAP_GET when fields does not exist in static fields)
- Support `spath` command without `path` param, which extracts all the attributes from JSON and generate columns dynamically
  - `JSON_EXTRACT_ALL` function is defined and used to extract all the attributes from JSON string.
  - Extracted attributes are stored to `_dynamic_columns`
- `fields` command is enhanced to project dynamic columns with wildcard pattern.
- Currently, it has limitation with `join` command, where dynamic columns cannot be used for join condition, and dynamic columns will become unavailable in case both of left and right relation contains dynamic columns. (This will be addressed in a separate PR)

### Review guidance
- [DynamicColumnProcessor.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-37e03640675247004ab204c9a74313887e8dc9f420f865383cfe3bf58e9850c7), [DynamicColumnReferenceDetector.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-90ae58436f059c103e8090272795510a3e5ef19f8a2c894962a5e267b1e6316d) collects most of the major logic related to dynamic columns, and those are called from: [CalciteRexNodeVisitor.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-e5198d773af75bf3173ef25676a2803a0091cb51e32d6ae30241273519d30261) and [CalciteRelNodeVisitor.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-518ddbde8cc139f2be35ffaeccd6dca4468147cc06f24f3458792e05584a61da)
  - Please review if the places where it is handling dynamic columns are right place, and the handling is correct.
- This PR is modifying the handling of wildcard by [DynamicWildcardProcessor.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-6fb9a9e3766b1277240fe3c05c5897a2a6d3eabe487f878fc8df67739456cb18)
  - Please review if it covers necessary cases and the logic is fine.
  - As we cannot know if specific dynamic field exist during planning time, it considers pattern could exist in dynamic fields.
- Please review internal functions
  - [MapGetFunctionImpl.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-efa5b3c6ebfe3dff805d068a3831f7356b9ed5abec48304fa3bea33f79eea319)
  - [MapMergeFunctionImpl.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-6b02f6b163e9a11dc49ac37c1e5f052a07965a4f82a507666a7728a56c94ec17)
  - [JsonExtractAllFunctionImpl.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-b556b68c046a9e77e60422a12936f0f8a921a4b8100449607276ce56b2bf6ce2)
  - [DynamicWildcardExcludeFunction.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-3da9cc922f256e0439e22bf6f225fb817c50a50ce3a24a4af5fb3de99fd464e9)
  - [DynamicWildcardExpandFunction.java](https://github.com/opensearch-project/sql/pull/4355/files#diff-787825f76f65dcc73a0812209b9751fc4ae36016f27929b28ce482cafbd8d222)
- In high level, let me know if there is some special case where dynamic fields could cause issue. (e.x. combination use with specific commands, etc.)

### Related Issues
https://github.com/opensearch-project/sql/issues/4307

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
+++ b/core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
@@ -60,7 +60,8 @@ public class SPath extends UnresolvedPlan {
         this.child,
         AstDSL.let(
             AstDSL.field(DynamicColumnProcessor.DYNAMIC_COLUMNS_FIELD),
-            // Merge if the dynamic columns field already exists
+            // Merge if the dynamic columns field already exists (using coalesce to let optimizer
+            // remove map_merge)
             AstDSL.function(
                 "coalesce",
                 AstDSL.function(
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
@@ -45,8 +45,7 @@ public class DynamicColumnProcessor {
     }
 
     Set<String> dynamicColumnNames = collectDynamicColumnNames(response.getResults());
-    Schema expandedSchema =
-        createExpandedSchema(response.getSchema(), dynamicColumnNames, response.getResults());
+    Schema expandedSchema = createExpandedSchema(response.getSchema(), dynamicColumnNames);
     List<ExprValue> expandedResults =
         expandResultRowsWithDynamicColumns(response.getResults(), dynamicColumnNames);
 
@@ -100,13 +99,8 @@ public class DynamicColumnProcessor {
     for (ExprValue row : results) {
       if (row instanceof ExprTupleValue) {
         ExprValue dynamicColumnsValue = row.tupleValue().get(DYNAMIC_COLUMNS_FIELD);
-        if (dynamicColumnsValue != null
-            && !dynamicColumnsValue.isNull()
-            && !dynamicColumnsValue.isMissing()) {
-          // Extract keys from the MAP
-          Map<String, ExprValue> mapValue = extractMapValue(dynamicColumnsValue);
-          columnNames.addAll(mapValue.keySet());
-        }
+        Map<String, ExprValue> mapValue = extractMapValue(dynamicColumnsValue);
+        columnNames.addAll(mapValue.keySet());
       }
     }
 
@@ -115,7 +109,7 @@ public class DynamicColumnProcessor {
 
   /** It will deduplicate columns if dynamic column name conflicts with original schema */
   private static Schema createExpandedSchema(
-      Schema originalSchema, Set<String> dynamicColumnNames, List<ExprValue> results) {
+      Schema originalSchema, Set<String> dynamicColumnNames) {
     Set<String> originalColumnNames = collectColumnNames(originalSchema);
     List<Column> expandedColumns = getColumnsExcludingDynamic(originalSchema);
 
--- a/docs/user/ppl/cmd/spath.rst
+++ b/docs/user/ppl/cmd/spath.rst
@@ -29,6 +29,7 @@ spath input=<field> [output=<field>] [path=<path>]
 
 * Limitation: The fields extracted by `spath` are always of type `text`. If you need to perform numeric operations on the extracted fields, you must cast them to the appropriate numeric type using the `eval` command.
 * Limitation: When `path` parameter is omitted, extracted fields cannot be used for `join` operations.
+* Limitation: When `path` parameter is omitted, `_dynamic_columns` field is used to store the extracted fields internally, and you cannot use this field name in that query.
 
 Note
 =====


core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
```diff
@@ -246,12 +246,17 @@ public enum BuiltinFunctionName {
...
+  JSON_EXTRACT_ALL(FunctionName.of("json_extract_all")),
   JSON_KEYS(FunctionName.of("json_keys")),
   JSON_SET(FunctionName.of("json_set")),
   JSON_DELETE(FunctionName.of("json_delete")),
   JSON_APPEND(FunctionName.of("json_append")),
   JSON_EXTEND(FunctionName.of("json_extend")),
 
+  /** Map Functions. */
+  MAP_MERGE(FunctionName.of("map_merge")),
+  MAP_GET(FunctionName.of("map_get")),
```
penghuo, line 258: Reuse ITEM function in calcite

core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
```diff
@@ -246,12 +246,17 @@ public enum BuiltinFunctionName {
...
   JSON_EXTRACT(FunctionName.of("json_extract")),
+  JSON_EXTRACT_ALL(FunctionName.of("json_extract_all")),
   JSON_KEYS(FunctionName.of("json_keys")),
   JSON_SET(FunctionName.of("json_set")),
   JSON_DELETE(FunctionName.of("json_delete")),
   JSON_APPEND(FunctionName.of("json_append")),
   JSON_EXTEND(FunctionName.of("json_extend")),
 
+  /** Map Functions. */
+  MAP_MERGE(FunctionName.of("map_merge")),
```
penghuo, line 257: Reuse map_concat function in calcite?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLSpathTest.java
```diff
@@ -45,4 +45,93 @@ public void testEvalWithOutput() {
...
+        "LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5],"
+            + " COMM=[$6], DEPTNO=[$7], _dynamic_columns=[COALESCE(MAP_MERGE(null:(VARCHAR NOT"
+            + " NULL, ANY NOT NULL) MAP, JSON_EXTRACT_ALL($1)), JSON_EXTRACT_ALL($1))])\n"
+            + "  LogicalTableScan(table=[[scott, EMP]])\n";
+    verifyLogical(root, expectedLogical);
+
+    String expectedSparkSql =
+        "SELECT `EMPNO`, `ENAME`, `JOB`, `MGR`, `HIREDATE`, `SAL`, `COMM`, `DEPTNO`,"
+            + " COALESCE(`MAP_MERGE`(NULL, `JSON_EXTRACT_ALL`(`ENAME`)),"
+            + " `JSON_EXTRACT_ALL`(`ENAME`)) `_dynamic_columns`\n"
```
penghuo, line 63: _dynamic_columns could be conflict with existing columns in index / extractd. We should add a section in doc to cover limitation on field name

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLSpathTest.java
```diff
@@ -45,4 +45,93 @@ public void testEvalWithOutput() {
...
+    String expectedLogical =
+        "LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5],"
+            + " COMM=[$6], DEPTNO=[$7], _dynamic_columns=[COALESCE(MAP_MERGE(null:(VARCHAR NOT"
+            + " NULL, ANY NOT NULL) MAP, JSON_EXTRACT_ALL($1)), JSON_EXTRACT_ALL($1))])\n"
+            + "  LogicalTableScan(table=[[scott, EMP]])\n";
+    verifyLogical(root, expectedLogical);
+
+    String expectedSparkSql =
+        "SELECT `EMPNO`, `ENAME`, `JOB`, `MGR`, `HIREDATE`, `SAL`, `COMM`, `DEPTNO`,"
+            + " COALESCE(`MAP_MERGE`(NULL, `JSON_EXTRACT_ALL`(`ENAME`)),"
```
penghuo, line 62: Can we drop the outer COALESCE? map_merge(null, JSON_EXTRACT_ALL(...)) already returns the new map, so COALESCE is redundant.

core/src/test/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessorTest.java
```diff
@@ -0,0 +1,216 @@
...
+    // Verify schema expansion
+    List<Column> expandedColumns = expandedResponse.getSchema().getColumns();
+    assertEquals(4, expandedColumns.size());
+
+    // Check that original columns are preserved (except _dynamic_columns)
+    assertEquals("id", expandedColumns.get(0).getName());
+    assertEquals(ExprCoreType.INTEGER, expandedColumns.get(0).getExprType());
+
+    assertEquals("age", expandedColumns.get(1).getName());
+    assertEquals(ExprCoreType.STRING, expandedColumns.get(1).getExprType());
```
penghuo, line 74: Expect is  STRING not INTEGER?

core/src/test/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessorTest.java
```diff
@@ -0,0 +1,216 @@
...
+import org.opensearch.sql.data.model.ExprValueUtils;
+import org.opensearch.sql.data.type.ExprCoreType;
+import org.opensearch.sql.executor.ExecutionEngine.QueryResponse;
+import org.opensearch.sql.executor.ExecutionEngine.Schema;
+import org.opensearch.sql.executor.ExecutionEngine.Schema.Column;
+
+class DynamicColumnProcessorTest {
+
+  @Test
+  void testExpandDynamicColumnsWithMixedTypes() {
```
penghuo, line 27: Add test cases verify name conflict columns.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
...
+        copiedRow.put(entry.getKey(), entry.getValue());
+      }
+    }
+    return copiedRow;
+  }
+
+  private static Map<String, ExprValue> extractMapValue(ExprValue mapValue) {
+    if (mapValue == null || mapValue.isNull() || mapValue.isMissing()) {
+      return Map.of();
+    }
```
penghuo, line 180: L103-L105 did same test.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
...
+
+    if (mapValue instanceof ExprTupleValue) {
+      return mapValue.tupleValue();
+    }
+
+    throw new IllegalStateException(
```
penghuo, line 186: return Map.of is better than exception?

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
...
+      if (!originalColumnNames.contains(columnName)) {
+        expandedColumns.add(new Column(columnName, null, ExprCoreType.STRING));
+      }
+    }
+
+    return new Schema(expandedColumns);
+  }
+
+  private static Set<String> collectColumnNames(Schema schema) {
+    return schema.getColumns().stream().map(Column::getName).collect(Collectors.toSet());
```
penghuo, line 126: Does columns order keeped?

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  /** It will deduplicate columns if dynamic column name conflicts with original schema */
+  private static Schema createExpandedSchema(
+      Schema originalSchema, Set<String> dynamicColumnNames, List<ExprValue> results) {
+    Set<String> originalColumnNames = collectColumnNames(originalSchema);
+    List<Column> expandedColumns = getColumnsExcludingDynamic(originalSchema);
+
+    for (String columnName : dynamicColumnNames) {
+      if (!originalColumnNames.contains(columnName)) {
```
penghuo, line 117: duplicate columns in dynamicColnumNames is discard? if yes, it is conflict iwht L116 comments.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
...
+        }
+      }
+    }
+
+    return columnNames;
+  }
+
+  /** It will deduplicate columns if dynamic column name conflicts with original schema */
+  private static Schema createExpandedSchema(
+      Schema originalSchema, Set<String> dynamicColumnNames, List<ExprValue> results) {
```
penghuo: results is not been used, remove it.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
...
+    Set<String> originalColumnNames = collectColumnNames(originalSchema);
+    List<Column> expandedColumns = getColumnsExcludingDynamic(originalSchema);
+
+    for (String columnName : dynamicColumnNames) {
+      if (!originalColumnNames.contains(columnName)) {
+        expandedColumns.add(new Column(columnName, null, ExprCoreType.STRING));
```
penghuo, line 118: alias always null?
type always string?

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
penghuo: combine duplicate column values in array is another option?

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
```diff
@@ -0,0 +1,199 @@
...
+        String newPath = currentPath.isEmpty() ? fieldName : currentPath + "." + fieldName;
+        JsonNode value = field.getValue();
+
+        visitValue(value, newPath, resultMap);
+      }
+    } else if (node.isArray()) {
+      String arrayPath = currentPath + ARRAY_SUFFIX;
+      for (JsonNode element : node) {
+        if (element.isObject() || element.isArray()) {
+          extractJsonValueRecursively(element, arrayPath, resultMap);
```
penghuo, line 108: extractJsonValueRecursively recurses without depth/size guard—maliciously deep payloads can overflow. Either cap recursion depth or convert to iterative traversal.

Add a TODO issue.

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
```diff
@@ -0,0 +1,199 @@
...
+   * @param resultMap The map to store extracted key-value pairs
+   */
+  private static void extractJsonValueRecursively(
+      JsonNode node, String currentPath, Map<String, Object> resultMap) {
+    if (node.isObject()) {
+      Iterator<Map.Entry<String, JsonNode>> fields = node.fields();
+      while (fields.hasNext()) {
+        Map.Entry<String, JsonNode> field = fields.next();
+        String fieldName = field.getKey();
+        String newPath = currentPath.isEmpty() ? fieldName : currentPath + "." + fieldName;
```
penghuo, line 99: nit, we should avoid isEmpty check by change extractJsonValueRecursively to extractJsonObject. and call it in eval.
```
eval() {
   Map.Entry<String, JsonNode> field = fields.next();
   extractJsonObject(node, field.getKey, resultMap)
}
```

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    try {
+      JsonNode jsonNode = OBJECT_MAPPER.readTree(jsonStr);
+      if (!jsonNode.isObject()) {
+        return null;
+      }
+
+      Map<String, Object> resultMap = new HashMap<>();
+      extractJsonValueRecursively(jsonNode, "", resultMap);
```
penghuo, line 78: what is the field order in resultMap, DFS/BFS?

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
```diff
@@ -51,6 +51,21 @@ public class CalcitePlanContext {
...
+  @Getter @Setter private boolean dynamicColumnsAvailable = false;
+
+  /**
+   * The flag used to determine whether we are currently resolving fields within a fields command.
+   * This affects how dynamic field resolution is handled - fields command uses aliasing while other
+   * commands use direct MAP access.
+   */
+  @Getter @Setter private boolean inFieldsCommand = false;
+
+  @Getter @Setter private boolean inGroupByContext = false;
```
Swiddis, line 67: [Someone once told me](https://github.com/opensearch-project/opensearch-spark/pull/837#discussion_r1841069546) that doing a bunch of conditional logic on boolean flags was bad practice

ykmr1224: I am looking at if there is better way, but so far it is needed since the branching logic happens in different places due to visitor pattern used to process plan tree. (like name resolution while visiting qualified name should behave differently depending on the context...)
Let me know if you have good idea to solve it in smarter way.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.utils;
```
Swiddis, line 6: chore: I'd prefer if the dynamic column core was in a dedicated package instead of utils

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
...
+    for (ExprValue row : originalResults) {
+      if (row instanceof ExprTupleValue) {
+        Map<String, ExprValue> originalRow = row.tupleValue();
+        Map<String, ExprValue> expandedRow = copyExceptDynamic(originalRow);
+
+        Map<String, ExprValue> dynamicColumns =
+            extractMapValue(originalRow.get(DYNAMIC_COLUMNS_FIELD));
+
+        // This will overwrite existing columns if there's a name conflict
+        for (String columnName : dynamicColumnNames) {
```
Swiddis, line 148: thought: We could probably avoid running the full nested loop for the majority of cases w/ arithmetic

If we have columns `[known1, known2, unknown1]` and in a previous loop we've found `unknown1`, we know there's 2 known columns (defined by the schema), and `dynamicColumnNames.size()` tells us there's 3 total. If the column set contains `unknown1`, we now have all our columns. Since "most" queries will have a fixed number of columns, this should let us skip the loop on most rows.

Maybe better to do this in a future PR, idk the benchmarking characteristics of this

core/src/test/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImplTest.java
```diff
@@ -0,0 +1,335 @@
...
+        "{\"mixed\": [{\"name\": \"Alice\", \"age\": 25}, \"primitive\", {\"name\": \"Bob\"}]}";
+
+    @SuppressWarnings("unchecked")
+    Map<String, Object> result = (Map<String, Object>) JsonExtractAllFunctionImpl.eval(json);
+
+    assertEquals(3, result.size());
+
+    // Arrays are now converted to JSON string representations
+    String nameValues = (String) result.get("mixed{}.name");
+    assertEquals("[\"Alice\",\"Bob\"]", nameValues);
```
penghuo, line 314: nice, merge policy is combine to array, right? if yes, it is should be the policy for combine with outter field also.

Swiddis (commented):
Partial review for now due to time constraints, will review the rest later

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
```diff
@@ -0,0 +1,239 @@
...
+  public static boolean tryResolveDynamicField(
+      String fieldName, List<String> currentFields, CalcitePlanContext context) {
+    // Don't resolve dynamic fields in coalesce function (it has special null handling)
+    if (context.isInCoalesceFunction()) {
+      return false;
+    }
+
+    if (currentFields.contains(fieldName)) {
+      return false;
+    }
```
penghuo, line 204: current solution resolved field in schema first, if not exist, resolve in _dynamic_col MAP[], which have have 2 behaviors,
* enforce-schema if _dynamic_col not exist, 
* schema-less if _dynamic_col exist.

core/src/main/java/org/opensearch/sql/ast/tree/SPath.java
```diff
@@ -45,6 +48,29 @@ public <T, C> T accept(AbstractNodeVisitor<T, C> nodeVisitor, C context) {
...
+  private Eval rewriteAsDynamicColumns() {
+    return AstDSL.eval(
+        this.child,
+        AstDSL.let(
+            AstDSL.field(DynamicColumnProcessor.DYNAMIC_COLUMNS_FIELD),
+            // Merge if the dynamic columns field already exists
+            AstDSL.function(
+                "coalesce",
+                AstDSL.function(
+                    "map_merge",
```
penghuo, line 68: we should merge on all fields, not just spath generate fields. e.g.
```
## query
| eval msg="{\"age\":\"10\", \"b\": 1}"
| eval age=1
| eval b=2
| spath input=msg

## results
age: [1, 10]
b: [2, 1]
```

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
ykmr1224: This result is used only for duplicate check, and don't need to care about the order here.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
ykmr1224: I think this requires some more deep dive, since we need different behavior per commands. Seems `spath` without path should append result to existing field, but with path replace. And other commands basically replace.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
ykmr1224: I think alias won't matter for the response.
Currently, I assume dynamic columns are typed as string.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
ykmr1224: I am not quite sure what you are proposing.
Here, we just populate dynamic column values into result table. The nested look is to go through rows and fields.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
ykmr1224: nice catch!, fixed.

core/src/main/java/org/opensearch/sql/calcite/utils/DynamicColumnProcessor.java
ykmr1224: If we prefer returning incorrect result than returning failure. (This case indicate something unexpected happened during process) What do you thing?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLSpathTest.java
ykmr1224: I modeled MAP_MERGE to return NULL when first parameter is NULL, and that's why COALESCE was needed. And the execution would become simpler since optimizer can know MAP_MERGE become NULL and COALESCE can be eliminated during optimization.

PR #4360, opened 2025-09-23
Fix `ClassCastException` for value-storing aggregates on nested PPL fields

### LINKED ISSUE #4359: [BUG] ClassCastException for value-storing aggregates on nested PPL fields
**What is the bug?**
Value-storing aggregate functions (`first()`, `last()`, `min()`, `max()`, etc.) throw a `ClassCastException` when their argument is a deeply nested field accessed via dotted path wrapped in backticks (e.g., `resource.attributes.telemetry.sdk.language`). Calculation-only aggregates (`count()`, `dc()`, `sum()`, `avg()`) work as expected.

---

1. Use an index matching `logs-otel-v1*` (or any index with similar nested object structure shown below).
2. In PPL, run one of the following (value-storing aggregates with nested fields):

   ```spl
   source=logs-otel-v1* | stats first(`resource.attributes.telemetry.sdk.language`) by severityNumber
   source=logs-otel-v1* | stats last(`resource.attributes.telemetry.sdk.language`) by severityNumber
   source=logs-otel-v1* | stats min(`resource.attributes.telemetry.sdk.language`) by severityNumber
   source=logs-otel-v1* | stats max(`resource.attributes.telemetry.sdk.language`) by severityNumber
   ```
3. Observe a 500 with `ClassCastException` in the response.
4. Compare with working queries:

   ```spl
   -- calculation-only aggregates on the same nested path
   source=logs-otel-v1* | stats count(`resource.attributes.telemetry.sdk.language`) by severityNumber
   source=logs-otel-v1* | stats dc(`resource.attributes.telemetry.sdk.language`) by severityNumber
   -- simple fields work for all aggregates
   source=logs-otel-v1* | stats first(severityNumber)
   source=logs-otel-v1* | stats last(severityNumber)
   source=logs-otel-v1* | stats min(severityNumber)
   source=logs-otel-v1* | stats max(severityNumber)
   ```

---

* `first()`/`last()` return the first/last non-null value of the nested field (e.g., `"java"`).
* `min()`/`max()` return the lexicographic min/max for keyword fields on the nested path.
* All value-storing aggregates should accept nested dotted paths the same as simple fields—without throwing.

---

* **Environment:** \[explore env]
* **Index Pattern:** `logs-otel-v1*` (or similar)
* **Plugins:** OpenSearch SQL (PPL)

**Actual error response**

```json
{
  "error": {
    "details": "class java.util.HashMap cannot be cast to class java.lang.String (java.util.HashMap and java.lang.String are in module java.base of loader 'bootstrap')",
    "type": "ClassCastException"
  },
  "status": 500
}
```

**Function classification**

* **Affected (value-storing):** `first()`, `last()`, `min()`, `max()`, and any aggregate that returns the field value itself.
* **Unaffected (calculation-only):** `count()`, `dc()`, `sum()`, `avg()`.

ahkcs:
## Summary

Value-storing aggregate functions (`first`, `last`, `min`, `max`) failed with a `ClassCastException` when targeting deeply nested field paths such as `resource.attributes.telemetry.sdk.language`.
This PR fixes nested-field value extraction in OpenSearch SQL by handling `Map`/`HashMap` objects that represent nested structures, eliminating the unsafe cast and restoring expected behavior for value-storing aggregates.

---

## Problem

These queries previously failed with `ClassCastException` for value-storing aggregates:

```spl
source=logs-otel-v1* | stats first(`resource.attributes.telemetry.sdk.language`) by severityNumber
source=logs-otel-v1* | stats last(`resource.attributes.telemetry.sdk.language`) by severityNumber
source=logs-otel-v1* | stats min(`resource.attributes.telemetry.sdk.name`) by severityNumber
source=logs-otel-v1* | stats max(`resource.attributes.telemetry.sdk.name`) by severityNumber
```

While simple fields worked:

```spl
source=logs-otel-v1* | stats first(severityNumber)
```

> Note: Calculation-only aggregates like `count()` and `dc()` were unaffected because they don’t store field values.

Observed error:

```json
{
  "error": {
    "details": "class java.util.HashMap cannot be cast to class java.lang.String",
    "type": "ClassCastException"
  }
}
```

---

## Root Cause

Execution-path analysis showed the failure at the final type-conversion layer in value-storing aggregates:

1. **Field resolution** returned partial `Map` structures (e.g., `{attributes={telemetry={sdk={language=java}}}}`) instead of the terminal primitive (`"java"`).
2. **Aggregates** (`first`, `last`, `min`, `max`) stored these `Map` objects in their accumulators.
3. **Type conversion** later attempted `(String) value` in `ObjectContent.stringValue()` where `value` was a `Map`.

Why calculation-only aggregates work: functions like `count()` and `dc()` perform calculations without storing the actual field values, so they never hit this conversion path.

---

## Solution

Implement nested-map handling in `ObjectContent.stringValue()` so value-storing aggregate outputs can be safely converted to strings.

**Key changes**

* Detect `Map` values and recursively extract the terminal primitive via `extractFinalPrimitiveValue(...)`.
* Preserve existing behavior for non-`Map` values (backward compatible).

**Before**

```java
return (String) value;  // throws ClassCastException if value is a Map
```

**After**

```java
if (value instanceof Map) {
  Map<String, Object> map = (Map<String, Object>) value;
  Object finalValue = extractFinalPrimitiveValue(map); // drills down single-value maps
  if (finalValue != null && !(finalValue instanceof Map)) {
    return finalValue.toString();
  }
  return map.toString();
}
return (String) value;
```

**Example extraction**
Input: `{attributes={telemetry={sdk={language=java}}}}`
→ `{telemetry={sdk={language=java}}}`
→ `{sdk={language=java}}`
→ `{language=java}`
→ `"java"`

---

## Verification (now succeeds)

```spl
source=logs-otel-v1* | stats first(`resource.attributes.telemetry.sdk.language`) by severityNumber
source=logs-otel-v1* | stats last(`resource.attributes.telemetry.sdk.language`) by severityNumber
source=logs-otel-v1* | stats min(`resource.attributes.telemetry.sdk.language`) by severityNumber
source=logs-otel-v1* | stats max(`resource.attributes.telemetry.sdk.language`) by severityNumber
```

## Resolves

* Resolves #4359

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAggregationIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAggregationIT.java
@@ -11,6 +11,7 @@ import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_CALCS;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_DATATYPE_NUMERIC;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_DATE_FORMATS;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_LOGS;
+import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_TELEMETRY;
 import static org.opensearch.sql.util.MatcherUtils.assertJsonEquals;
 import static org.opensearch.sql.util.MatcherUtils.rows;
 import static org.opensearch.sql.util.MatcherUtils.schema;
@@ -41,6 +42,7 @@ public class CalcitePPLAggregationIT extends PPLIntegTestCase {
     loadIndex(Index.DATE_FORMATS);
     loadIndex(Index.DATA_TYPE_NUMERIC);
     loadIndex(Index.LOGS);
+    loadIndex(Index.TELEMETRY);
     loadIndex(Index.TIME_TEST_DATA);
   }
 
@@ -1245,5 +1247,55 @@ public class CalcitePPLAggregationIT extends PPLIntegTestCase {
                 "source=%s | stats count() by age | sort -age | head 3", TEST_INDEX_BANK));
     verifySchema(response, schema("count()", "bigint"), schema("age", "int"));
     verifyDataRows(response, rows(1, 39), rows(2, 36), rows(1, 34));
+  }
+
+  @Test
+  public void testFirstLastWithSimpleField() throws IOException {
+    // This should work - testing simple field first
+    JSONObject actual =
+        executeQuery(
+            String.format("source=%s | stats first(severityNumber)", TEST_INDEX_TELEMETRY));
+    verifySchema(actual, schema("first(severityNumber)", "int"));
+    verifyDataRows(actual, rows(9));
+  }
+
+  @Test
+  public void testFirstLastWithDeepNestedField() throws IOException {
+    // This test should now work with the fix for ClassCastException
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | stats first(`resource.attributes.telemetry.sdk.language`)",
+                TEST_INDEX_TELEMETRY));
+    verifySchema(actual, schema("first(`resource.attributes.telemetry.sdk.language`)", "string"));
+    verifyDataRows(actual, rows("java"));
+  }
+
+  @Test
+  public void testLastWithDeepNestedField() throws IOException {
+    // This test should now work with the fix for ClassCastException
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | stats last(`resource.attributes.telemetry.sdk.language`)",
+                TEST_INDEX_TELEMETRY));
+    verifySchema(actual, schema("last(`resource.attributes.telemetry.sdk.language`)", "string"));
+    verifyDataRows(actual, rows("rust"));
+  }
+
+  @Test
+  public void testFirstLastWithDeepNestedFieldByGroup() throws IOException {
+    // This test should now work with the fix for ClassCastException
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | stats first(`resource.attributes.telemetry.sdk.language`) by"
+                    + " severityNumber",
+                TEST_INDEX_TELEMETRY));
+    verifySchema(
+        actual,
+        schema("first(`resource.attributes.telemetry.sdk.language`)", "string"),
+        schema("severityNumber", "int"));
+    verifyDataRows(actual, rows("java", 9), rows("python", 12), rows("go", 16));
   }
 }
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAggregationIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAggregationIT.java
@@ -1356,4 +1356,147 @@ public class CalcitePPLAggregationIT extends PPLIntegTestCase {
     verifySchema(actual, schema("min_lang", "string"), schema("max_lang", "string"));
     verifyDataRows(actual, rows("go", "rust"));
   }
+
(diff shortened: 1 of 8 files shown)


integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAggregationIT.java
```diff
@@ -1246,4 +1248,112 @@ public void testLimitAfterAggregation() throws IOException {
...
+  @Test
+  public void testMinMaxMultipleNestedFields() throws IOException {
+    // Test min/max with multiple nested field aggregations in one query
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                "source=%s | stats min(`resource.attributes.telemetry.sdk.language`) as min_lang,"
+                    + " max(`resource.attributes.telemetry.sdk.language`) as max_lang",
+                TEST_INDEX_TELEMETRY));
+    verifySchema(actual, schema("min_lang", "string"), schema("max_lang", "string"));
```
LantaoJin, line 1356: What if the type of nested `language` is an integer, can the ppl work?

ahkcs: Yes, in ObjectContent.java, first the `extractFinalPrimitiveValue` method will extract the value as Object and then we'll add toString() to it.

LantaoJin: No I mean what if the mapping contains 
```
                    "properties": {
                      "version": {
                        "type": "integer"
                      }
```
and query with `` | stats min(`resource.attributes.telemetry.sdk.version`) ``.
Can you update the mapping and data, and test above query?

ahkcs: Changed the approach - moved the fix from ObjectContent.java to OpenSearchExprValueFactory.construct() for better universal handling. Updated the telemetry mapping to test for types like integer and boolean

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -526,4 +527,26 @@ private ExprValue parseInnerArrayValue(
...
+  @SuppressWarnings("unchecked")
+  private Object extractFinalPrimitiveValue(Object value) {
+    if (value == null || !(value instanceof Map)) {
+      return value;
+    }
+
+    Map<String, Object> map = (Map<String, Object>) value;
+    if (map.size() == 1) {
+      Object singleValue = map.values().iterator().next();
+      return extractFinalPrimitiveValue(singleValue);
```
Swiddis, line 547: Is there risk of a RecursionError here?

If this is only loaded from an index, it's maybe fine (I think you can't index massively nested structures, not sure). But if there's a path where this is executed on user input, it'd be very easy to cause a crash with this.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -526,4 +527,26 @@ private ExprValue parseInnerArrayValue(
...
+   * @param value The value to extract from
+   * @return The extracted primitive value, or the original value if extraction is not possible
+   */
+  @SuppressWarnings("unchecked")
+  private Object extractFinalPrimitiveValue(Object value) {
+    if (value == null || !(value instanceof Map)) {
+      return value;
+    }
+
+    Map<String, Object> map = (Map<String, Object>) value;
```
Swiddis, line 544: Why do this cast at runtime instead of just specifying the type in the signature and pushing the check to the left?

Swiddis (approved):

LantaoJin (approved):
LGTM if the CI rerun passed.

LantaoJin:
@ahkcs please manually backport it, thanks

PR #4379, opened 2025-09-25
[Doc] Enable doctest with Calcite

LantaoJin:
### Description
Followup of #4372 

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4384, opened 2025-09-25
Update bin doctest

ahkcs:
### Description
Update the doctest for bin.rst file
Removed floating point for numeric span

### DIFF
--- a/docs/category.json
+++ b/docs/category.json
@@ -8,7 +8,6 @@
   ],
   "ppl_cli": [
     "user/ppl/cmd/ad.rst",
-    "user/ppl/cmd/bin.rst",
     "user/ppl/cmd/dedup.rst",
     "user/ppl/cmd/describe.rst",
     "user/ppl/cmd/showdatasources.rst",
@@ -63,6 +62,7 @@
     "user/ppl/cmd/rex.rst",
     "user/ppl/cmd/stats.rst",
     "user/ppl/cmd/timechart.rst",
+    "user/ppl/cmd/bin.rst",
     "user/ppl/cmd/search.rst"
   ]
 }
--- a/docs/user/ppl/cmd/bin.rst
+++ b/docs/user/ppl/cmd/bin.rst
@@ -240,14 +240,14 @@ Example 1: Basic numeric span
 
 PPL query::
 
-    ppl> source=accounts | bin age span=10 | fields age | head 3;
+    os> source=accounts | bin age span=10 | fields age | head 3;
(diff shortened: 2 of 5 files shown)


docs/user/ppl/cmd/bin.rst
```diff
@@ -240,186 +240,172 @@ Example 1: Basic numeric span
...
+    os> source=accounts | bin age span=10 | fields age | head 3;
     fetched rows / total rows = 3/3
     +-------+
     | age   |
     |-------|
     | 30-40 |
-    | 35-45 |
-    | 25-35 |
+    | 30-40 |
+    | 20-30 |
```
ykmr1224: Not critical, but these examples might be a bit confusing since it shows duplicate age bins. It is better showing other field to indicate `age` is just showing the bin for each record.

doctest/test_docs.py
```diff
@@ -41,7 +41,9 @@
     'worker': 'worker.json',
     'work_information': 'work_information.json',
     'events': 'events.json',
-    'otellogs': 'otellogs.json'
+    'otellogs': 'otellogs.json',
+    'time_test': 'time_test.json',
+    'bank': 'bank.json'
```
ykmr1224: Let's reuse existing ones if possible. I think bank can be replaced with accounts, at least. Not sure if we have json with timestamp.

docs/category.json
```diff
@@ -63,6 +62,7 @@
     "user/ppl/cmd/rex.rst",
     "user/ppl/cmd/stats.rst",
     "user/ppl/cmd/timechart.rst",
+    "user/ppl/cmd/bin.rst",
```
ykmr1224: Let's put items in alphabetical order to easily identify duplicate record.

doctest/test_docs.py
vamsimanohar: +1 lets reuse unless necessary.

docs/user/ppl/cmd/bin.rst
ahkcs: Updated some tests to use fields command

doctest/test_docs.py
ahkcs: We currently don't have json with timestamp so I kept time_test.json, removed bank.json

docs/category.json
ahkcs: Updated

LantaoJin:
#4379 is to moving all doctest to calcite. But I can rebase that one. Let's merge this first.

PR #4393, opened 2025-09-26
Refactor name resolution in Calcite PPL

ykmr1224:
### Description
- Made full refactoring on name resolution in Calcite PPL.
- Now it supports names containing `.`, and access to struct field.
- It is needed to support `spath` command with dynamic columns. (I will add logic to resolve dynamic columns in separate PRs)
- I found current logic won't verify the table/alias name while resolving field from correlation, and I couldn't find good way to refer the available table/alias name from correlation variable. We might need to introduce abstraction of stack of scopes to better model the name resolution.

### Related Issues
- https://github.com/opensearch-project/sql/issues/4307

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteQueryAnalysisIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteQueryAnalysisIT.java
@@ -24,6 +24,6 @@ public class CalciteQueryAnalysisIT extends QueryAnalysisIT {
             IllegalArgumentException.class,
             () ->
                 executeQuery(String.format("search source=%s | fields name", TEST_INDEX_ACCOUNT)));
-    verifyErrorMessageContains(e, "Field [nonexistent] not found.");
+    verifyErrorMessageContains(e, "Field [name] not found.");
   }
 }


integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLBasicIT.java
```diff
@@ -95,10 +95,7 @@ public void testFieldsShouldBeCaseSensitive() {
     Throwable e =
         assertThrowsWithReplace(
             IllegalStateException.class, () -> executeQuery("source=test | fields NAME"));
-    verifyErrorMessageContains(
-        e,
-        "field [NAME] not found; input fields are: [name, age, _id, _index, _score, _maxscore,"
-            + " _sort, _routing]");
+    verifyErrorMessageContains(e, "Field [NAME] not found.");
```
penghuo, line 98: Why remove the possible fields list? security concern?

penghuo (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLBasicIT.java
ykmr1224: It is because it could be incorrect, since it was not considering multiple inputs (join) nor correlation.

ykmr1224:
> Nit: Keep a debug log at the entry/exit point of resolve(), showing the incoming qualified name and the chosen resolution path. Remove or greatly reduce logging in private helpers unless it covers an edge case that is hard to reproduce or diagnose.
> 
> If you want to retain some logs for future debugging, comment them with a reference to the bug or scenario they address.

Does `debug` level log still matter?

PR #4400, opened 2025-09-28
Pushdown case function in aggregations as range queries

yuancu:
### Description
This PR push down case functions used in aggregations as range queries. For example, the query `source=bank | eval age_range = case (age < 30, 'u30', age < 40, 'u40' else 'u100') | stats avg(balance) by age_range` will be pushed down as the following OpenSearch DSL:
```json
{
  "aggregations": {
    "age_range": {
      "range": {
        "field": "age",
        "ranges": [
          {
            "key": "u30",
            "to": 30
          },
          {
            "key": "u40",
            "from": 30,
            "to": 40
          },
          {
            "key": "u100",
            "from": 40
          }
        ],
        "keyed": true
      },
      "aggregations": {
        "avg(balance)": {
          "avg": {
            "field": "balance"
          }
        }
      }
    }
  }
}
```
 

### Related Issues

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -249,15 +300,15 @@ && isAutoDateSpan(project.getProjects().get(groupList.getFirst()))) {
           }
           return Pair.of(
               Collections.singletonList(aggregationBuilder),
-              new CompositeAggregationParser(metricParserList, countAggNames));
+              new BucketAggregationParser(metricParsers, countAggNames));
```
qianheng-aws: Does it mean we don't need `CompositeAggregationParser` anymore?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -249,15 +300,15 @@ && isAutoDateSpan(project.getProjects().get(groupList.getFirst()))) {
...
         } catch (CompositeAggUnSupportedException e) {
```
qianheng-aws: Is there any logic for handling range bucket when this exception happened? This part of code will construct composite agg unsupported bucket like `auto_date_histogram`.  I'm wondering what will happens if both having `auto_span` + `range_bucket` in our query.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -207,36 +209,85 @@ public static Pair<List<AggregationBuilder>, OpenSearchAggregationResponseParser
...
       // but only count() can apply doc_count optimization in bucket aggregation.
-      boolean countAllOnly = !aggregate.getGroupSet().isEmpty();
+      boolean countAllOnly = !groupList.isEmpty();
       Pair<List<String>, Builder> countAggNameAndBuilderPair =
           removeCountAggregationBuilders(metricBuilder, countAllOnly);
       Builder newMetricBuilder = countAggNameAndBuilderPair.getRight();
       List<String> countAggNames = countAggNameAndBuilderPair.getLeft();
 
-      if (aggregate.getGroupSet().isEmpty()) {
+      Pair<Set<Integer>, AggregationBuilder> caseAggPushedAndRangeBuilder =
```
qianheng-aws: Why should we handle range bucket separately? Can it be handled in `createNestedBuckets` like how we construct auto_date_span?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -207,36 +209,85 @@ public static Pair<List<AggregationBuilder>, OpenSearchAggregationResponseParser
...
         if (newMetricBuilder != null) {
           bucketBuilder.subAggregations(newMetricBuilder);
         }
         return Pair.of(
             Collections.singletonList(bucketBuilder),
-            new BucketAggregationParser(metricParserList, countAggNames));
-      } else {
+            new BucketAggregationParser(metricParsers, countAggNames));
+      }
+      // It has both composite aggregation and range aggregation:
```
qianheng-aws: Not related to this PR but #4413. It seems we have chance to optimize sub-agg by combining part of buckets into composite bucket if they're supported by composite agg. 

e.g. transform `termBucket-termBucket-autoDateSpanBucket` to `compositeBucket - autoDateSpanBucket`.  

We can also do bucket reorder to scale this optimization to more cases

e.g. transform `termBucket-autoDateSpanBucket-termBucket` to `compositeBucket - autoDateSpanBucket`.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: Please add test case like `bin timestamp bins=xxx | eval age_range = case ... | stats count() by timestamp, age_range`

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: Seems composite agg is more proper than constructing multiple sub-agg by `createNestedBuckets`. It looks good to me to keep the current implementation. https://github.com/opensearch-project/sql/pull/4400#discussion_r2425156433.

We should refactor auto_date_span to have similar implementation but also have to  use `createNestedBuckets` if both have range_bucket and auto_date_span.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
yuancu: Yes, I'm thinking of retiring `CompositeAggregationParser` since a composite aggregation is essentially also a bucket aggregation. By keeping only `CompositeAggregationParser`, we can avoid duplicating codes to handle subaggregations,etc.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
yuancu: I haven't handled the case where auto date histogram and range bucket coexists. Currently, it will treat `age_range` as a sub-term aggregation with script.

<details>
<summary>
explain results for `source=time_test | bin timestamp bins=3 | eval value_range = case(value < 7000, 'small' else 'great') | stats bucket_nullable=false avg(value) by timestamp, value_range`

</summary>

```
{
  "calcite": {
    "logical": """LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(avg(value)=[$2], timestamp=[$0], value_range=[$1])
    LogicalAggregate(group=[{0, 1}], avg(value)=[AVG($2)])
      LogicalProject(timestamp=[$9], value_range=[$10], value=[$2])
        LogicalFilter(condition=[IS NOT NULL($9)])
          LogicalProject(@timestamp=[$0], category=[$1], value=[$2], _id=[$4], _index=[$5], _score=[$6], _maxscore=[$7], _sort=[$8], _routing=[$9], timestamp=[WIDTH_BUCKET($3, 3, -(MAX($3) OVER (), MIN($3) OVER ()), MAX($3) OVER ())], value_range=[CASE(<($2, 7000), 'small':VARCHAR, 'great':VARCHAR)])
            CalciteLogicalIndexScan(table=[[OpenSearch, time_test]])
""",
    "physical": """EnumerableLimit(fetch=[10000])
  CalciteEnumerableIndexScan(table=[[OpenSearch, time_test]], PushDownContext=[[AGGREGATION->rel#2751:LogicalAggregate.NONE.[](input=RelSubset#2693,group={1, 2},avg(value)=AVG($0)), PROJECT->[avg(value), timestamp, value_range]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"timestamp":{"auto_date_histogram":{"field":"timestamp","buckets":3,"minimum_interval":null},"aggregations":{"value_range":{"terms":{"script":{"source":"{\"langType\":\"calcite\",\"script\":\"rO0ABXNyABFqYXZhLnV0aWwuQ29sbFNlcleOq7Y6G6gRAwABSQADdGFneHAAAAADdwQAAAAGdAAHcm93VHlwZXQERHsKICAiZmllbGRzIjogWwogICAgewogICAgICAidWR0IjogIkVYUFJfVElNRVNUQU1QIiwKICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICJudWxsYWJsZSI6IHRydWUsCiAgICAgICJwcmVjaXNpb24iOiAtMSwKICAgICAgIm5hbWUiOiAiQHRpbWVzdGFtcCIKICAgIH0sCiAgICB7CiAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAicHJlY2lzaW9uIjogLTEsCiAgICAgICJuYW1lIjogImNhdGVnb3J5IgogICAgfSwKICAgIHsKICAgICAgInR5cGUiOiAiSU5URUdFUiIsCiAgICAgICJudWxsYWJsZSI6IHRydWUsCiAgICAgICJuYW1lIjogInZhbHVlIgogICAgfSwKICAgIHsKICAgICAgInVkdCI6ICJFWFBSX1RJTUVTVEFNUCIsCiAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAicHJlY2lzaW9uIjogLTEsCiAgICAgICJuYW1lIjogInRpbWVzdGFtcCIKICAgIH0sCiAgICB7CiAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAicHJlY2lzaW9uIjogLTEsCiAgICAgICJuYW1lIjogIl9pZCIKICAgIH0sCiAgICB7CiAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAicHJlY2lzaW9uIjogLTEsCiAgICAgICJuYW1lIjogIl9pbmRleCIKICAgIH0sCiAgICB7CiAgICAgICJ0eXBlIjogIlJFQUwiLAogICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAibmFtZSI6ICJfc2NvcmUiCiAgICB9LAogICAgewogICAgICAidHlwZSI6ICJSRUFMIiwKICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgIm5hbWUiOiAiX21heHNjb3JlIgogICAgfSwKICAgIHsKICAgICAgInR5cGUiOiAiQklHSU5UIiwKICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgIm5hbWUiOiAiX3NvcnQiCiAgICB9LAogICAgewogICAgICAidHlwZSI6ICJWQVJDSEFSIiwKICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgInByZWNpc2lvbiI6IC0xLAogICAgICAibmFtZSI6ICJfcm91dGluZyIKICAgIH0KICBdLAogICJudWxsYWJsZSI6IHRydWUKfXQABGV4cHJ0Atp7CiAgIm9wIjogewogICAgIm5hbWUiOiAiQ0FTRSIsCiAgICAia2luZCI6ICJDQVNFIiwKICAgICJzeW50YXgiOiAiU1BFQ0lBTCIKICB9LAogICJvcGVyYW5kcyI6IFsKICAgIHsKICAgICAgIm9wIjogewogICAgICAgICJuYW1lIjogIjwiLAogICAgICAgICJraW5kIjogIkxFU1NfVEhBTiIsCiAgICAgICAgInN5bnRheCI6ICJCSU5BUlkiCiAgICAgIH0sCiAgICAgICJvcGVyYW5kcyI6IFsKICAgICAgICB7CiAgICAgICAgICAiaW5wdXQiOiAyLAogICAgICAgICAgIm5hbWUiOiAiJDIiCiAgICAgICAgfSwKICAgICAgICB7CiAgICAgICAgICAibGl0ZXJhbCI6IDcwMDAsCiAgICAgICAgICAidHlwZSI6IHsKICAgICAgICAgICAgInR5cGUiOiAiSU5URUdFUiIsCiAgICAgICAgICAgICJudWxsYWJsZSI6IGZhbHNlCiAgICAgICAgICB9CiAgICAgICAgfQogICAgICBdCiAgICB9LAogICAgewogICAgICAibGl0ZXJhbCI6ICJzbWFsbCIsCiAgICAgICJ0eXBlIjogewogICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICJudWxsYWJsZSI6IGZhbHNlLAogICAgICAgICJwcmVjaXNpb24iOiAtMQogICAgICB9CiAgICB9LAogICAgewogICAgICAibGl0ZXJhbCI6ICJncmVhdCIsCiAgICAgICJ0eXBlIjogewogICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICJudWxsYWJsZSI6IGZhbHNlLAogICAgICAgICJwcmVjaXNpb24iOiAtMQogICAgICB9CiAgICB9CiAgXQp9dAAKZmllbGRUeXBlc3NyABdqYXZhLnV0aWwuTGlua2VkSGFzaE1hcDTATlwQbMD7AgABWgALYWNjZXNzT3JkZXJ4cgARamF2YS51dGlsLkhhc2hNYXAFB9rBwxZg0QMAAkYACmxvYWRGYWN0b3JJAAl0aHJlc2hvbGR4cD9AAAAAAAAMdwgAAAAQAAAABHQACkB0aW1lc3RhbXBzcgA6b3JnLm9wZW5zZWFyY2guc3FsLm9wZW5zZWFyY2guZGF0YS50eXBlLk9wZW5TZWFyY2hEYXRlVHlwZZ4tUq4QfcqvAgABTAAHZm9ybWF0c3QAEExqYXZhL3V0aWwvTGlzdDt4cgA6b3JnLm9wZW5zZWFyY2guc3FsLm9wZW5zZWFyY2guZGF0YS50eXBlLk9wZW5TZWFyY2hEYXRhVHlwZcJjvMoC+gU1AgADTAAMZXhwckNvcmVUeXBldAArTG9yZy9vcGVuc2VhcmNoL3NxbC9kYXRhL3R5cGUvRXhwckNvcmVUeXBlO0wAC21hcHBpbmdUeXBldABITG9yZy9vcGVuc2VhcmNoL3NxbC9vcGVuc2VhcmNoL2RhdGEvdHlwZS9PcGVuU2VhcmNoRGF0YVR5cGUkTWFwcGluZ1R5cGU7TAAKcHJvcGVydGllc3QAD0xqYXZhL3V0aWwvTWFwO3hwfnIAKW9yZy5vcGVuc2VhcmNoLnNxbC5kYXRhLnR5cGUuRXhwckNvcmVUeXBlAAAAAAAAAAASAAB4cgAOamF2YS5sYW5nLkVudW0AAAAAAAAAABIAAHhwdAAJVElNRVNUQU1QfnIARm9yZy5vcGVuc2VhcmNoLnNxbC5vcGVuc2VhcmNoLmRhdGEudHlwZS5PcGVuU2VhcmNoRGF0YVR5cGUkTWFwcGluZ1R5cGUAAAAAAAAAABIAAHhxAH4AE3QABERhdGVzcgA8c2hhZGVkLmNvbS5nb29nbGUuY29tbW9uLmNvbGxlY3QuSW1tdXRhYmxlTWFwJFNlcmlhbGl6ZWRGb3JtAAAAAAAAAAACAAJMAARrZXlzdAASTGphdmEvbGFuZy9PYmplY3Q7TAAGdmFsdWVzcQB+ABp4cHVyABNbTGphdmEubGFuZy5PYmplY3Q7kM5YnxBzKWwCAAB4cAAAAAB1cQB+ABwAAAAAc3IAE2phdmEudXRpbC5BcnJheUxpc3R4gdIdmcdhnQMAAUkABHNpemV4cAAAAAF3BAAAAAF0ABdkYXRlX2hvdXJfbWludXRlX3NlY29uZHh0AAhjYXRlZ29yeX5xAH4AEnQABlNUUklOR3QABXZhbHVlfnEAfgASdAAHSU5URUdFUnQACXRpbWVzdGFtcHNxAH4AC3EAfgAUcQB+ABdxAH4AG3NxAH4AHwAAAAF3BAAAAAF0ABdkYXRlX2hvdXJfbWludXRlX3NlY29uZHh4AHg=\"}","lang":"opensearch_compounded_script","params":{"utcTimestamp":1760342401776012000}},"size":1000,"min_doc_count":1,"shard_min_doc_count":0,"show_term_doc_count_error":false,"order":{"_key":"asc"}},"aggregations":{"avg(value)":{"avg":{"field":"value"}}}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
"""
  }
}
```

</details>

I'll optimize this case and move the snippets of creating range buckets to `createNestedBuckets`

docs/user/ppl/functions/condition.rst
```diff
@@ -227,6 +227,16 @@ Argument type: all the supported data type, (NOTE : there is no comma before "el
 
 Return type: any
 
+Limitations
+>>>>>>>>>>>
+
+When each condition is a field comparison with a numeric literal and each result expression is a string literal, the query will be optimized as `range aggregations <https://docs.opensearch.org/latest/aggregations/bucket/range>`_ if pushdown optimization is enabled. However, this optimization has the following limitations:
```
LantaoJin, line 233: IMO, it's not a limitation of `case` function, it is just a restricted optimization. We can just call out in what case, the `case` function would be optimized to range DSL. Can we add some optimizable `case` usages in user doc?

docs/user/ppl/functions/condition.rst
```diff
@@ -227,6 +227,16 @@ Argument type: all the supported data type, (NOTE : there is no comma before "el
...
+
+- Null values will not be grouped into any bucket of a range aggregation and will be ignored
+- The default ELSE clause will use the string literal ``"null"`` instead of actual NULL values
+
+To avoid these edge-case limitations, set ``plugins.calcite.pushdown.enabled`` to false.
```
LantaoJin: remove this

docs/user/ppl/functions/condition.rst
yuancu: I think the stated conditions are in the scope of restricted optimizations, but the limitations are not because we will still do the optimization regardless of whether it has null values in its column or whether there is a default NULL range.

The problem is that there is no way to know in advance whether there exists null values in a column. Therefore, if we do this optimization, we always risk the discrepancy in results of with & without push-down.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
yuancu: Fixed with the latest implementation

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
yuancu: Fixed

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
yuancu: Fixed & test case added

docs/user/ppl/functions/condition.rst
yuancu: Removed

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -207,58 +203,81 @@ public static Pair<List<AggregationBuilder>, OpenSearchAggregationResponseParser
...
+      // Push auto date span & case in group-by list into nested aggregations
+      Pair<Set<Integer>, AggregationBuilder> aggPushedAndAggBuilder =
+          createNestedAggregation(groupList, project, subBuilder, helper);
+      Set<Integer> aggPushed = aggPushedAndAggBuilder.getLeft();
+      AggregationBuilder pushedAggBuilder = aggPushedAndAggBuilder.getRight();
+      // The group-by list after removing pushed aggregations
+      groupList = groupList.stream().filter(i -> !aggPushed.contains(i)).toList();
+      if (pushedAggBuilder != null) {
+        subBuilder = new Builder().addAggregator(pushedAggBuilder);
+      }
```
qianheng-aws: [non-blocking] Should this part of code be put in the second branch of the below `if` for better readability? It should have structure like:
```
if (aggregate.getGroupSet().isEmpty()) {
  // no group by
} else {
    // Push auto date span & case in group-by list into nested aggregations
   ...
   ...
   if (groupList.isEmpty()) {
        // No composite aggregation at top-level
        ...
   } else {
        // Composite aggregation at top level
        ...
    }
}
```

It works well  with current code but performing useless operations on some empty collections for no-group-by.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/BucketAggregationParser.java
```diff
@@ -55,6 +60,19 @@ private List<Map<String, Object>> parseBucket(
...
             : parse(aggregations);
+
+    if (bucket instanceof CompositeAggregation.Bucket compositeBucket) {
+      Map<String, Object> common = extract(compositeBucket);
+      for (Map<String, Object> r : results) {
+        r.putAll(common);
+      }
+    } else if (bucket instanceof Range.Bucket) {
+      // return null so that an empty range will be filtered out
+      if (bucket.getDocCount() == 0) {
```
qianheng-aws: Put this at the beginning of this method to skip meaningless operations in advance?

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/BucketAggregationParser.java
```diff
@@ -55,6 +60,19 @@ private List<Map<String, Object>> parseBucket(
...
+        return null;
+      }
+      // the content of the range bucket is extracted with `r.put(name, bucket.getKey())` below
+    }
```
qianheng-aws: Should there be another `else` here? Otherwise it will put the `bucket.getKey` into the results for composite agg as well.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/BucketAggregationParser.java
qianheng-aws: Please include `InternalAutoDateHistogram.Bucket` here as well to skip empty bucket for auto_date_span. 

And also remove the code https://github.com/opensearch-project/sql/blob/1e62fba637e67f2c30de312ea0fa6ca965104b37/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java#L312-L329 here. Previously we add another filter to filter out the empty bucket.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLCaseFunctionIT.java
```diff
@@ -246,4 +252,222 @@ public void testCaseWhenInSubquery() throws IOException {
...
+        actual8,
+        rows(5686.0, 1, "a30", "TN", "M"),
+        rows(16418.0, 1, "a30", "WA", "M"),
+        rows(40540.0, 1, "a30", "PA", "F"),
+        rows(4180.0, 1, "a30", "MD", "M"),
+        rows(32838.0, 1, "u30", "VA", "F"),
+        rows(39225.0, 1, "a30", "IL", "M"),
+        rows(48086.0, 1, "a30", "IN", "F"));
+
+    // 2.4 Composite (2 fields) - Range - Range - Metric (with count)
```
qianheng-aws, line 404: Please add test case  for `composite - auto_date_span - range - metric`. Set bins to a big enough value like 100 to verify whether the empty buckets are filtered out properly.

yuancu: Test added

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
yuancu: Thanks for the suggestion! optimized the logic here

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/BucketAggregationParser.java
yuancu: Refactored as suggested

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/BucketAggregationParser.java
yuancu: Yep. I used to put all contents in `bucket.getKey` to pass unit tests in a wrong way; corrected the behavior now.

qianheng-aws (approved):

LantaoJin (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -1160,4 +1146,127 @@ public void testPushDownMinOrMaxAggOnDerivedField() throws IOException {
...
                 TEST_INDEX_ACCOUNT)));
   }
+
+  @Test
+  public void testCasePushdownAsRangeQueryExplain() throws IOException {
+    // CASE 1: Range - Metric
+    // 1.1 Range - Metric
+    assertYamlEqualsJsonIgnoreId(
+        loadExpectedPlan("agg_range_metric_push.yaml"),
+        explainQueryToString(
```
LantaoJin: should change to `explainQueryYaml`

yuancu: Fixed. Thanks for reminding!

Issue #4366, opened 2025-09-24
[DOC] Add documentation for variable_count_threshold and frequency_threshold_percentage on patterns


paulstn:
**What do you want to do?**

- [X] Request a change to existing documentation
- [ ] Add new documentation
- [ ] Report a technical problem with the documentation
- [ ] Other
  
**Tell us about your request.** Provide a summary of the request.

I would like more documentation on `variable_count_threshold` and `frequency_threshold_percentage` within the patterns command, for when `method=brain`. 

Ideally, I'd like to have enough documentation to where I wouldn't have to look within code to know exactly how these parameters would interact with the patterns command results. i.e. how it would potentially transform my data.  

**Version:** List the OpenSearch version to which this issue applies, e.g. 2.14, 2.12--2.14, or all.

3.1-3.3

**What other resources are available?** Provide links to related issues, POCs, steps for testing, etc.

songkant-aws:
Will do this one

penghuo:
@songkant-aws  could we close this PR with https://github.com/opensearch-project/sql/pull/4402

songkant-aws:
@penghuo Yes, I think we can close it with #4402

Issue #4363, opened 2025-09-24
[BUG] Calcite enabled patterns command aggregation mode on brain not returning all tokens


paulstn:
**What is the bug?**

When calcite engine is enabled, the patterns command does not return all the tokens. 
For example, when running the query:
```
source = opensearch_dashboards_sample_data_logs | patterns `message` method=brain mode=aggregation
```
or the query
```
source = opensearch_dashboards_sample_data_logs | patterns `message` method=brain | stats count() by patterns_field
```

I get a result like this:
```
  "datarows": [
    [
      "<token1> - - [<token2>] \"GET <token3> HTTP/<token4><token5>\" 200 <token6> \"-\" \"Mozilla/<token7><token8> (<token9>; Linux <token10>_<token11>; rv:<token12><token13><token14>) Gecko/<token15> Firefox/<token16><token17><token18>\"",
      4935,
      {
        "<token1>": [
          "223.87.60.27",
          "130.246.123.197",
          "6.138.148.165",
          "218.148.135.12",
          "12.132.31.17",
          "129.138.185.193",
          "159.64.35.129",
          "78.142.163.191",
          "54.190.60.72",
          "221.0.214.157"
        ],
        "<token3>": [
          "/opensearch/opensearch-1.0.0.deb_1",
          "/beats/metricbeat_1",
          "/beats/metricbeat/metricbeat-6.3.2-amd64.deb_1",
          "/beats/filebeat/filebeat-6.3.2-linux-x86_64.tar.gz_1",
          "/opensearch/opensearch-1.0.0.tar.gz_1",
          "/beats_1",
          "/opensearch_dashboards/opensearch_dashboards-1.0.0-linux-x86_64.tar.gz_1",
          "/beats_1",
          "/beats/metricbeat/metricbeat-6.3.2-amd64.deb_1",
          "/opensearch_1"
        ],
        "<token2>": [
          "2018-07-22T00:39:02.912Z",
          "2018-07-22T03:26:21.326Z",
          "2018-07-22T03:57:39.612Z",
          "2018-07-22T04:18:12.345Z",
          "2018-07-22T05:36:25.812Z",
          "2018-07-22T05:39:39.668Z",
          "2018-07-22T06:06:42.742Z",
          "2018-07-22T06:10:08.359Z",
          "2018-07-22T06:15:34.660Z",
          "2018-07-22T06:21:08.365Z"
        ]
      }
    ]
  ],
```

[...]

Issue #4362, opened 2025-09-24
[BUG] patterns command returns type of patterns_field inconsistently when calcite enabled


paulstn:
**What is the bug?**

When calcite is enabled, the patterns command switches between the two types of `patterns_fields` inconsistently.
----
For the same index, the two types of `patterns_fields` have been:
(Calcite engine + brain)
```
<token1> - - [<token2>] "GET <token3> HTTP/<token4><token5>" 200 <token6> "-" "Mozilla/<token7><token8> (<token9>; Linux <token10>_<token11>; rv:<token12><token13><token14>) Gecko/<token15> Firefox/<token16><token17><token18>"
```
(V2 engine + brain)
```
<*IP*> - - [<*DATETIME*>] "GET <*> HTTP/<*><*>" 404 <*> "-" "Mozilla/<*><*> (<*>; Linux <*>_<*>; rv:<*><*><*>) Gecko/<*> Firefox/<*><*><*>"
```

This issue is about how when an aggregation isn't performed on a PPL command with patterns in the calcite engine, it switches from the format seen in calcite to the one seen in V2. 

For example, when running the PPL query:
```
source = opensearch_dashboards_sample_data_logs | patterns `message` method=brain
```
What's returned has the pattern field in the "V2" format.

When running that PPL query with any kind of aggregation (also noticed with a `fields` clause):
```
source = opensearch_dashboards_sample_data_logs | patterns `message` method=brain | stats count() by patterns_field
```
it returns with the token format.

This behavior is problematic because trying to select documents with a `where` clause isn't as easy when the `patterns_field` changes between queries.

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

songkant-aws:
Will make it backward compatible

penghuo:
@songkant-aws  could we close this PR with #4402

songkant-aws:
@penghuo Yes, we can close it with #4402. BTW, I don't have permission to close such issues.

PR #4402, opened 2025-09-29
Fix numbered token bug and make it optional output in patterns command

### LINKED ISSUE #4364: [FEATURE] [ENHANCEMENT] Allow users to specify if they want to see tokens or standard delimiters on calcite patterns command

When on Calcite, the `patterns_field` contains a different format, where every variable element in the pattern string is put in with a numbered token. This behavior isn't always wanted, if I want to use the old type of pattern string where it matched types of the variable element (e.g. `<*IP*>).

Give an option in the patterns command where users can opt in to having their patterns string be delimited by numbered tokens. 

songkant-aws:
### Description
This PR will resolve one feature enhancement and a bug fix.

1. Feature: Since 3.1.0, patterns command for brain algorithm outputs numbered token like `<token1>.<token2>@<token3>` instead of `<*>.<*>@<*>`. But it's not always wanted per customer request. So add a new flag option `show_numbered_token=true|false` in the patterns command to make it optional. By default, it will not show numbered token to be compatible with V2's format.
2. Bug fix: In some cases, tokens are parsed incorrectly. The brain algorithm could generate pattern like: `<*> failed to start for block<*><*><*>`. It has continuous variable placeholder `<*>` but actually they can be merged into one. The parsing logic can't identify continuous variable placeholder. So at the final pattern generation part, we will merge continuous `<*>` placeholders.

### Related Issues
Resolves #4364 , #4363 , #4366, #4362

### Check List


docs/user/ppl/cmd/patterns.rst
```diff
@@ -18,22 +18,27 @@ Description
...
 
 * field: mandatory. The text(string) field to analyze for patterns.
 * byClause: optional. Fields or scalar functions used to group logs for labeling/aggregation.
 * method: optional. Algorithm choice: ``simple_pattern`` (default) or ``brain``. The method is configured by the setting ``plugins.ppl.pattern.method``.
 * mode: optional. Output mode: ``label`` (default) or ``aggregation``. The mode is configured by the setting ``plugins.ppl.pattern.mode``.
 * max_sample_count: optional. Max sample logs returned per pattern in aggregation mode (default: 10). The max_sample_count is configured by the setting ``plugins.ppl.pattern.max.sample.count``.
 * buffer_limit: optional. Safeguard parameter for ``brain`` algorithm to limit internal temporary buffer size (default: 100,000, min: 50,000). The buffer_limit is configured by the setting ``plugins.ppl.pattern.buffer.limit``.
+* show_numbered_token: optional. The flag to turn on numbered token output format (default: false). The show_numbered_token is configured by the setting ``plugins.ppl.pattern.show.numbered.token``.
```
LantaoJin, line 34: Seems when version <3.1.0, the command does not show numbered tokens. 3.1.0 <= version < 3.3.0, this command shows numbered tokens by default. when version >=3.3.0, align with the behaviour of version <3.1.0, right?

Any thoughts to control the default value of `show_numbered_token` by `plugins.ppl.syntax.legacy.preferred` in future?

qianheng-aws (dismissed):

docs/user/ppl/cmd/patterns.rst
```diff
@@ -127,30 +129,34 @@ PPL query::
...
--------
-3.1.0
+Starting 3.1.0, patterns command support aggregation mode. Calcite engine by default labels the variables with '<*>' placeholder.
 
-Configuration
--------------
-New output format requires Calcite enabled.
+PPL query::
+
+    PPL> source=apache | patterns message method=simple_pattern mode=aggregation | fields patterns_field, pattern_count | head 1 ;
```
LantaoJin: `PPL` is meaning the query is not executable in doctest. Now [almost of PPL rst files moved to calcite enabled queue](https://github.com/opensearch-project/sql/pull/4379#discussion_r2387558074) since we have enabled calcite by default in main branch (3.3.0), please rebase and check if all queries can execute with calcite correctly.

LantaoJin: It is controlled by `docs/category.json`

songkant-aws: I changed doctest after merging main. After Calcite is enabled by default for doctest, patterns command output has some format change in this fix. So I changed doctest for patterns.rst. It's expected change.

Grok and Parse doctest are actually not impacted.

docs/user/ppl/cmd/patterns.rst
songkant-aws: Will create the issue for it. Another suggestion from user is we could backport the change to older version. Not sure we need some minor version like 3.1.x.

qianheng-aws (approved):

PR #4413, opened 2025-09-30
Fallback to sub-aggregation if composite aggregation doesn't support

qianheng-aws:
### Description
Fallback to sub-aggregation if composite aggregation doesn't support(e.g. auto_span).

As sub-aggregation includes term agg which doesn't include null bucket, so we only fallback to push the aggregation when bucket_nullable=false. Otherwise, the aggregation will fail to been push down and executs as a Calcite's Enumerable aggregate operator. 

### Related Issues

### Check List


yuancu (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchAggregateIndexScanRule.java
```diff
@@ -34,17 +40,31 @@ protected OpenSearchAggregateIndexScanRule(Config config) {
 
   @Override
   public void onMatch(RelOptRuleCall call) {
-    if (call.rels.length == 3) {
+    if (call.rels.length == 4) {
+      final LogicalAggregate aggregate = call.rel(0);
+      final LogicalFilter filter = call.rel(1);
+      final LogicalProject project = call.rel(2);
+      final CalciteLogicalIndexScan scan = call.rel(3);
```
LantaoJin, line 47:
```
        LogicalAggregate
          LogicalProject
            LogicalFilter(condition=[IS NOT NULL($7)])
```
Shouldn't be following?
```
      final LogicalAggregate aggregate = call.rel(0);
      final LogicalProject project = call.rel(1);
      final LogicalFilter filter = call.rel(2);
      final CalciteLogicalIndexScan scan = call.rel(3);
```

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchAggregateIndexScanRule.java
```diff
@@ -123,12 +143,77 @@ public interface Config extends RelRule.Config {
                                         Predicate.not(OpenSearchIndexScanRule::isLimitPushed)
                                             .and(OpenSearchIndexScanRule::noAggregatePushed))
                                     .noInputs()));
+    // TODO: No need this rule once https://github.com/opensearch-project/sql/issues/4403 is
+    // addressed
+    Config BUCKET_NON_NULL_AGG =
+        ImmutableOpenSearchAggregateIndexScanRule.Config.builder()
+            .build()
+            .withDescription("Agg-Filter-Project-TableScan")
```
LantaoJin, line 151: why not "Agg-Project-Filter-TableScan"? I think `Agg-Project` should be treated as a single unit

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -206,74 +212,54 @@ public static Pair<List<AggregationBuilder>, OpenSearchAggregationResponseParser
...
           // The optimization must require all count aggregations are removed,
           // and they have only one field name
-          List<String> countAggNameList =
-              removedCountAggBuilders.stream()
-                  .map(ValuesSourceAggregationBuilder::getName)
-                  .toList();
           return Pair.of(
-              ImmutableList.copyOf(newMetricBuilder.getAggregatorFactories()),
-              new CountAsTotalHitsParser(countAggNameList));
+              List.of(), new CountAsTotalHitsParser(countAggNameAndBuilderPair.getLeft()));
```
LantaoJin: can you create a named var for `countAggNameAndBuilderPair.getLeft()`?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -289,7 +275,7 @@ && isAutoDateSpan(project.getProjects().get(groupList.getFirst()))) {
    * @param countAllOnly remove count() only, or count(FIELD) will be removed.
    * @return a pair of removed ValueCountAggregationBuilder and updated metric builder
```
LantaoJin: javadoc need to update

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -127,6 +127,12 @@ public static class ExpressionNotAnalyzableException extends Exception {
     }
   }
 
+  public static class CompositeUnSupportedException extends RuntimeException {
```
LantaoJin: can you move it to package `org.opensearch.sql.exception`?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchAggregateIndexScanRule.java
qianheng-aws: The original plan is 
```
LogicalAggregate
      LogicalProject
            LogicalFilter(condition=[IS NOT NULL($7)])
                LogicalProject
                  LogicalIndexScan
```

`ProjectFilterTransposeRule` and `ProjectMergeRule` will produce `agg-filter-project-scan` pattern.

Maybe `FilterProjectTransposeRule` + `ProjectMergeRule` may produce `agg-project-filter-scan`

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchAggregateIndexScanRule.java
qianheng-aws: But `filter(isnotnull)` here is also derived from the `agg`.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchAggregateIndexScanRule.java
qianheng-aws: The pattern `agg-filter-project-scan`  will make it easier to detect whether the filter is derived from the `agg`. The other pattern could work as well while we need to do additional transformation since the ref of field has passed through a `project` operator.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: This exception only happens here and will be caught internally. It doesn't need to be accessed outside of this file

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: done

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: done

Issue #4349, opened 2025-09-22
[RFC] Support permissive mode in PPL


penghuo:
## Problem Statement

OpenSearch PPL is strong-schema and fail-fast. When data is dirty or drifting, queries abort instead of degrading gracefully. At scale, this makes analytics and operations brittle.

* **[Type/data mismatch](https://github.com/opensearch-project/sql/issues/4330).** PPL evaluates expressions on strictly typed fields derived from mappings. Mixed representations—e.g., `"1"` (string), `[1]` (array) trigger cast/class errors that terminate the query and break dashboards. Operators lack TRY/SAFE semantics, so evaluation cannot proceed null-on-error.  *Example:* `source=log* | eval xx=flags+1` fails when any shard holds `"1"` or `[1]` in `flags`; desired behavior is `2` or `NULL` with an error counter.
* **[Schema-missing (`dynamic:false`) fields are invisible to PPL](https://github.com/opensearch-project/sql/issues/3995).** PPL relies on mappings for field addressability and typing. When objects are mapped with `dynamic:false`, their subfields are present in `_source` but unmapped, so PPL cannot reference them, even though the Query DSL can project them from `_source`. This creates a usability gap, forces reindex/mapping changes, or pushes users back to DSL. It also pressures teams toward mapping explosion for discoverability. *Example:* object `event` with `dynamic:false`; `fields event.user` returns 400 in PPL, while DSL with `"_source": ["event.user"]` succeeds.
* [Explict cast required when access extract field](https://github.com/opensearch-project/sql/issues/4356), When using the REX/PARSE/SPATH command to extract numeric values and then performing arithmetic operations on them, the system fails because the extracted values are treated as strings rather than numbers.
* **Cross-index type conflicts.** In observability datasets, schemas drift: the same logical field may be `string` in some indices and `long` in others (often from JSON/text extraction). When a PPL query spans those indices, the engine effectively picks one variant to type the field, which is non-deterministic and error-prone.

### Requirements

* **Coverage:** Handle type/data mismatches, dynamic:false invisibility, and cross-index type conflicts under permissive TRY/SAFE semantics.
* **Performance:** No performance regression.
* **Reusability:** Engine-agnostic design reusable for PPL on other engines, including Spark.

## Expected Behaviour

**Setting**: ppl.query.permissive=true, default is false, true enable permissive mode.
**Scenario:** `intA * stringB` where `intA` is mapped as `integer` and `stringB` is unmapped.
[...]

PR #4421, opened 2025-09-30
PPL `fillnull` command enhancement

### LINKED ISSUE #4419: [RFC] Enhancement for PPL `fillnull` command
## Short-Term Goal

* **Add option-style syntax** for `fillnull`:

  ```
  fillnull [value=<string>] [<field-list>]
  ```
* **`value` is mandatory for now**.
* **Rationale:** Backward-compatible with current queries and ergonomics, while **preserving strict, schema-stable behavior**.

## Long-Term Goal (aligned with permissive mode)

Introduce an **opt-in permissive/schemaless execution mode** so `fillnull` (and related expressions) can **gracefully handle mixed types**

See detailed planning at:
#4349

## Current State

* **Syntax today:**

  * `fillnull with <replacement> [in <field-list>]`
  * `fillnull using <field>=<replacement> [, <field>=<replacement> ...]`
* **Typing today:** Lowered to standard `COALESCE(field, <value>)` (type-strict, returns original field type); single replacement across mixed types errors.

## Proposal

* **Short term:** Add the option-style surface (`fillnull [value=<string>] [<field-list>]`) alongside existing forms; semantics remain strict and schema-preserving; require `value=`.
* **Long term:** Implement the permissive/schemaless mode described above

## Examples (short term behavior)

* Same-typed fields:

  ```
  source=logs | fillnull value=0 response_time, status_code
  ```

  (Equivalent to `fillnull with 0 in response_time, status_code`)

* Apply to all fields (strict typing still applies, all fields must be strings in this case):

  ```
  source=logs | fillnull value="missing"
  ```

* Mixed-type workarounds today:

  ```
  source=logs | fillnull using host="N/A", response_time=0
  -- or
  source=logs | fillnull with "N/A" in host | fillnull with 0 in response_time
  ```

* **Type coercion failure (today, single replacement across mixed types):**

  ```
  source=logs | fillnull value="N/A" host, response_time
  ```

  Representative error:

  ```
  {
    "error": {
      "reason": "Invalid Query",
      "details": "Cannot infer return type for COALESCE; operand types: [INTEGER, VARCHAR]",
    },
    "status": 400
  }
  ```

ahkcs:
### What does this PR do?

Adds an alternative, option-style form for `fillnull` (available **since 3.4**) alongside existing `with/using` syntax. Semantics remain **strict and schema-stable** (no permissive casting in this PR).

**New syntax (option-style):**

```ppl
source=accounts | fillnull value=0              // apply to all fields (subject to type constraints)
source=accounts | fillnull value=0 age balance  // apply to specific fields
```

**Existing syntax (still supported):**

```ppl
source=accounts | fillnull with 0 in age, balance
source=accounts | fillnull using age=0, balance=0
```

---

### Notes / Limitations (intentional in this PR)

* `value` is **mandatory** (no default value yet).
* Mixed-type fills with a single replacement still error by design (strict typing preserved).
* Semantics remain strict and schema-preserving

---

### Future Work

* Track and implement an **opt-in permissive/schemaless mode** for mixed-type scenarios (follow-up: **#4349**).

### Related Issues
Partially resolves #4419 

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFillNullCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFillNullCommandIT.java
@@ -143,4 +143,35 @@ public class CalciteFillNullCommandIT extends FillNullCommandIT {
         "fillnull failed: replacement value type VARCHAR is not compatible with field 'num0' "
             + "(type: DOUBLE). The replacement value type must match the field type.");
   }
+
+  @Test
+  public void testFillNullWithLargeIntegerOnIntField() throws IOException {
+    // fillnull using int0=8589934592 (2^33, larger than Integer.MAX_VALUE)
+    // Tests whether type family equality allows BIGINT value for INTEGER field
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | fields int0 | fillnull using int0=8589934592", TEST_INDEX_CALCS));
+    // If this test passes, it confirms that NUMERIC type family allows INT->BIGINT coercion
+    // The result should show 8589934592 for null int0 values
+    verifyDataRows(
+        result,
+        rows(1),
+        rows(8589934592L),
+        rows(8589934592L),
+        rows(8589934592L),
+        rows(7),
+        rows(3),
+        rows(8),
+        rows(8589934592L),
+        rows(8589934592L),
+        rows(8),
+        rows(4),
+        rows(10),
+        rows(8589934592L),
+        rows(4),
+        rows(11),
+        rows(4),
+        rows(8));
+  }
 }


integ-test/src/test/java/org/opensearch/sql/calcite/CalciteNoPushdownIT.java
```diff
@@ -19,88 +18,87 @@
...
-  CalciteSortCommandIT.class,
-  CalciteStatsCommandIT.class,
-  CalciteSystemFunctionIT.class,
-  CalciteTextFunctionIT.class,
-  CalciteTopCommandIT.class,
-  CalciteTrendlineCommandIT.class,
-  CalciteVisualizationFormatIT.class,
-  CalciteWhereCommandIT.class,
-  CalcitePPLTpchIT.class
+@Suite.SuiteClasses({CalciteExplainIT.class
```
dai-chen: why all comment out?

docs/user/ppl/cmd/fillnull.rst
```diff
@@ -90,7 +102,57 @@ PPL query::
...
+
+**Note:** When applying the same value to all fields, all fields must be the same type. For mixed types, use separate fillnull commands.
+
 Limitations
 ===========
 * The ``fillnull`` command is not rewritten to OpenSearch DSL, it is only executed on the coordination node.
 * Before 3.1.0, at least one field is required.
+* **Type Restrictions**:
+
+  The replacement value type must match ALL field types in the field list. When applying the same value to multiple fields, all fields must be the same type (all strings or all numeric).
```
dai-chen: I didn't find validation code and test. What will happen if users provide replacement value of different type?

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
```diff
@@ -212,4 +212,55 @@ public void testFillNullWithFunctionMultipleCommands() throws IOException {
         rows("sixteen", 16.81),
         rows("unknown", 7.12));
   }
+
+  @Test
+  public void testAggregationSumWithNullValues() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source=%s | stats sum(num0) as total", TEST_INDEX_CALCS));
+    verifyDataRows(result, rows(10));
+  }
```
ykmr1224: Looks like this test case is not for `fillnull`.

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
```diff
@@ -212,4 +212,55 @@ public void testFillNullWithFunctionMultipleCommands() throws IOException {
...
+                TEST_INDEX_CALCS));
+    verifyDataRows(result, rows(10));
+  }
+
+  @Test
+  public void testAggregationCountWithNullValues() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source=%s | stats count(num0) as total", TEST_INDEX_CALCS));
+    verifyDataRows(result, rows(8));
+  }
```
ykmr1224: Same as above.

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
```diff
@@ -212,4 +212,55 @@ public void testFillNullWithFunctionMultipleCommands() throws IOException {
...
+                TEST_INDEX_CALCS));
+    verifyDataRows(result, rows(17));
+  }
+
+  @Test
+  public void testAggregationAvgWithNullValues() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source=%s | stats avg(num0) as average", TEST_INDEX_CALCS));
+    verifyDataRows(result, rows(1.25));
+  }
```
ykmr1224: Same as above.

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
```diff
@@ -212,4 +212,55 @@ public void testFillNullWithFunctionMultipleCommands() throws IOException {
...
+        executeQuery(String.format("source=%s | stats sum(num0) as total", TEST_INDEX_CALCS));
+    verifyDataRows(result, rows(10));
+  }
+
+  @Test
+  public void testAggregationSumAfterFillNull() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | fillnull with 0 in num0 | stats sum(num0) as total",
```
ykmr1224: I don't see much value to test with multiple aggregate functions.

And should we add tests for new syntax?

ppl/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -264,6 +262,7 @@ EARLIEST:                           'EARLIEST';
 LATEST:                             'LATEST';
 TAKE:                               'TAKE';
 LIST:                               'LIST';
+VALUE:                              'VALUE';
```
ykmr1224: I think it does not belong to AGGREGATIONS.

docs/user/ppl/cmd/fillnull.rst
```diff
@@ -90,7 +102,57 @@ PPL query::
...
+    os> source=accounts | fields age, balance | fillnull value=0;
+    fetched rows / total rows = 4/4
+    +-----+---------+
+    | age | balance |
+    |-----+---------|
+    | 32  | 39225   |
+    | 36  | 5686    |
+    | 28  | 32838   |
+    | 33  | 4180    |
+    +-----+---------+
```
ykmr1224: This example does not demonstrate the behavior (no field is filled)

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
```diff
@@ -281,6 +281,18 @@ public void testFillNullWithoutFields() {
...
+  public void testFillNullValueSyntaxWithFields() {
+    assertEquals(
+        "source=table | fillnull with *** in identifier, identifier",
+        anonymize("source=t | fillnull value=0 f1 f2"));
+  }
+
+  @Test
+  public void testFillNullValueSyntaxAllFields() {
+    assertEquals("source=table | fillnull with ***", anonymize("source=t | fillnull value=0"));
+  }
```
ykmr1224, line 294: Why does it anonymize to different syntax?

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
ahkcs: Hi Tomo, I added these test case to track how `fillnull` interacts with aggregate functions

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
ahkcs: Hi Tomo, the new syntax is tested in `CalciteFillNullCommandIT`
These tests are added to test how `fillnull` command interacts with aggregate functions, I think they are optional but it can show how aggregate functions changed after using fillnull

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
ahkcs: I updated to remove these tests as they are not directly related to `fillnull` command

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
ahkcs: Removed

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
ahkcs: Removed

integ-test/src/test/java/org/opensearch/sql/ppl/FillNullCommandIT.java
ahkcs: I updated to remove these tests as they are not directly related to fillnull command

ppl/src/main/antlr/OpenSearchPPLLexer.g4
ahkcs: Moved Value to COMMAND ASSIST KEYWORDS

docs/user/ppl/cmd/fillnull.rst
ahkcs: Updated

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
ahkcs: It was reusing the existing logic to anonymize.
The existing logic distinguishes between two syntaxes based on semantic behavior:
  - Same values → with...in
  - Different values → using

However, value= syntax also applies the same value to all fields (like with...in), so we can't rely on this check alone to distinguish them. We need to track which syntax was originally used.

  Added a useValueSyntax boolean flag to the FillNull AST node:
  - Set to true when parsing fillnull value=...
  - Set to false for with...in and using (default)
  
This change might introduce some changes that is not clean so I think maybe it is also okay to keep using the existing logic since it only affects anonymizer test now. 

I updated the implementation to have the same syntax now for anonymizer test

integ-test/src/test/java/org/opensearch/sql/calcite/CalciteNoPushdownIT.java
ahkcs: Hi Chen, I forgot to put it back after testing, restored

docs/user/ppl/cmd/fillnull.rst
ahkcs: Hi Chen, I added test cases in `CalciteFillNullCommandIT` to show the error message when users provide replacement value of different type

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFillNullCommandIT.java
```diff
@@ -13,4 +21,120 @@ public void init() throws Exception {
...
+    // fillnull value=0 on mixed type fields without explicit field list
+    // When no fields specified, all fields must be same type
+    Throwable t =
+        assertThrowsWithReplace(
+            RuntimeException.class,
+            () ->
+                executeQuery(
+                    String.format(
+                        "source=%s | fields str2, int0 | fillnull value=0", TEST_INDEX_CALCS)));
+    verifyErrorMessageContains(t, "Cannot infer return type for COALESCE");
```
dai-chen: I feel this error message is confusing to users? Should we validate in analyzer layer and return clear error message like 'the replacement value type is not compatible ..."?

core/src/main/java/org/opensearch/sql/ast/tree/FillNull.java
```diff
@@ -42,8 +47,14 @@ public static FillNull ofSameValue(UnresolvedExpression replacement, List<Field>
 
   private final List<Pair<Field, UnresolvedExpression>> replacementPairs;
 
-  FillNull(List<Pair<Field, UnresolvedExpression>> replacementPairs) {
+  // Track if value= syntax was used (added in 3.4). Only needed to distinguish from with...in
+  // since both apply same value to all fields. using syntax is detected by checking if all
+  // replacement values are the same.
+  private final boolean useValueSyntax;
```
dai-chen, line 53: Is this only for anonymizer?

ahkcs: Yes, I think this change is debatable, you can refer to this comment: https://github.com/opensearch-project/sql/pull/4421#discussion_r2395563594

docs/user/ppl/cmd/fillnull.rst
```diff
@@ -90,7 +102,57 @@ PPL query::
     +-----------------------+---------------+
 
 
+Example 5: replace null with specified value on specific fields (value= syntax)
+===============================================================================
+
+Since 3.4.
```
RyanL1997: is this the new format of unified rst doc for ppl? cc @ritvibhatt

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFillNullCommandIT.java
```diff
@@ -13,4 +21,120 @@ public void init() throws Exception {
     super.init();
     enableCalcite();
   }
+
+  // value= syntax tests
```
RyanL1997: what is this? is this referencing to the following test case?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFillNullCommandIT.java
RyanL1997: +1

docs/user/ppl/cmd/fillnull.rst
```diff
@@ -90,7 +102,57 @@ PPL query::
...
+    | amberduke@pyrami.com  | Pyrami      |
+    | hattiebond@netagy.com | Netagy      |
+    | <not found>           | Quility     |
+    | daleadams@boink.com   | <not found> |
+    +-----------------------+-------------+
+
+Example 6: replace null with specified value on all fields (value= syntax)
+==========================================================================
+
+Since 3.4. This example only works when Calcite enabled. When no field list is specified, the replacement applies to all fields in the result.
```
RyanL1997: question: since calcite is enabled by default, do we still need to emphasize this? cc @ritvibhatt

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFillNullCommandIT.java
ahkcs: Yes, this is for testing the new value= syntax

RyanL1997: nit: we can remove it, since the test name itself should explain the purpose

ahkcs: Updated to remove this

docs/user/ppl/cmd/fillnull.rst
ahkcs: Update doc to remove the mentioning of calcite

docs/user/ppl/cmd/fillnull.rst
ahkcs: Updated doc to not mention version info

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFillNullCommandIT.java
ahkcs: Updated the error message handling to make it clearer to users, modified IT tests

core/src/main/java/org/opensearch/sql/ast/tree/FillNull.java
Swiddis: I agree with you, we should just have the anonymizer output a consistent format if the underlying parse rule is irrecoverable without special tracking.

Personally I'd actually prefer the anonymizer defaulted to `value=` syntax, since it's consistent and more in line with other command syntax.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1493,6 +1494,30 @@ public RelNode visitWindow(Window node, CalcitePlanContext context) {
...
+   * SemanticCheckException if types are incompatible.
+   */
+  private void validateFillNullTypeCompatibility(
+      RexNode replacement, RexNode fieldRef, String fieldName) {
+    RelDataTypeFamily replacementFamily = replacement.getType().getFamily();
+    RelDataTypeFamily fieldFamily = fieldRef.getType().getFamily();
+
+    // Check if the replacement type is compatible with the field type
+    // Allow NULL type family as it's compatible with any type
+    if (fieldFamily != replacementFamily
```
Swiddis, line 1508: question: what are the semantics of family equality? Will it automatically down-cast?

Wondering what would happen if you tried to fillnull an int with 8589934592 (2^33)

ahkcs: ~~if the replacement value is 2^33, then the replacement value would be LONG, which will throw error if the field type is INT~~
Let me add an IT test for this case

ahkcs: The type family equality check allows it. Both INT and BIGINT belong to SqlTypeFamily.NUMERIC, so the validation passes. Calcite's COALESCE function then handles the type widening automatically - it will widen the INT field to BIGINT to accommodate the larger value. No error is thrown; The result type becomes BIGINT.

I added an IT test (testFillNullWithLargeIntegerOnIntField) that confirms this behavior works correctly at runtime.

dai-chen (approved):
Thanks for the changes!

core/src/main/java/org/opensearch/sql/ast/tree/FillNull.java
ykmr1224: I don't have strong preference on this. I think anonimized query log is needed for operator to check the input query in case of trouble shooting(I suppose we don't use it for analyzing use case statistics, but not quite sure), while we don't want to log sensitive information.

RyanL1997 (approved):

PR #4434, opened 2025-10-03
[Enhancement] Error handling for illegal character usage in java regex named capture group

### LINKED ISSUE #3944: [FEATURE] Support `_/-` as parsed field name

The following query failed with SemanticCheckException.
```
source=test001 | parse message 'endpoint=/(?<action_code>\\w+)' | fields action_code

{
  "error": {
    "reason": "Invalid Query",
    "details": "can't resolve Symbol(namespace=FIELD_NAME, name=action_code) in type env",
    "type": "SemanticCheckException"
  },
  "status": 400
}
```

The query should works successfully.

n/a

RyanL1997:
### Description
[Enhancement] Error handling for underscore usage in rex regex

- Better handling for invalid named capture group
- Updated the doc
- Add ITs

### Related Issue
* Resolve #3944
* Resolve https://github.com/opensearch-project/sql/issues/4467
* Relate to all the named capture group extraction related command (`parse`/`rex`/`replace`...)

### Behavior after enhancement
```bash
# rex command
    -d '{
      "query": "source=accounts | rex field=email \".+@(?<domain_name>.+)\" | fields email, domain_name"
}'

{
  "error": {
    "reason": "Invalid Query",
    "details": "Invalid capture group name 'domain_name'. Java regex group names must start with a letter and contain only letters and digits.",
  },
  "status": 400
}


# parse command
  -d '{
    "query": "source=accounts | parse email \".+@(?<domain-name>.+)\" | fields email, domain-name | head 5"
}'

{
  "error": {
    "reason": "Invalid Query",
    "details": "Invalid capture group name 'domain-name'. Java regex group names must start with a letter and contain only letters and digits.",
  },
  "status": 400
}% 
```

### Check List

### DIFF
--- a/docs/user/ppl/cmd/parse.rst
+++ b/docs/user/ppl/cmd/parse.rst
@@ -114,3 +114,7 @@ There are a few limitations with parse command:
   For example, the following query will not display the parsed fields ``host`` unless the source field ``email`` is also explicitly included::
 
     source=accounts | parse email '.+@(?<host>.+)' | fields email, host ;
+
+- Named capture group must start with a letter and contain only letters and digits.
+
+  For detailed Java regex pattern syntax and usage, refer to the `official Java Pattern documentation <https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html>`_
--- a/docs/user/ppl/cmd/rex.rst
+++ b/docs/user/ppl/cmd/rex.rst
@@ -226,7 +226,6 @@ There are several important limitations with the rex command:
 
 **Named Capture Group Naming:**
 
-- Named capture groups cannot contain underscores or the other illegal special characters due to Java regex limitations
 - Group names must start with a letter and contain only letters and digits
 - For detailed Java regex pattern syntax and usage, refer to the `official Java Pattern documentation <https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html>`_
 


core/src/main/java/org/opensearch/sql/expression/parse/RegexCommonUtils.java
```diff
@@ -20,7 +20,7 @@
 public class RegexCommonUtils {
 
   private static final Pattern NAMED_GROUP_PATTERN =
-      Pattern.compile("\\(\\?<([a-zA-Z][a-zA-Z0-9]*)>");
+      Pattern.compile("\\(\\?<([a-zA-Z_][a-zA-Z0-9_]*)>");
```
Swiddis: This is fine for now because we only use it in one place, but I'm concerned if any functionality comes up that uses this, they might not know they have to validate. It's easy to miss.

Ideally we should have the validation happen directly as part of getting named group candidates, or even just add a second path for if the main group find returns 0, we check against the underscore version and raise the exception then.

RyanL1997: thats a good idea. Applied the improvement.

RyanL1997:
> I think same as issue https://github.com/opensearch-project/sql/issues/3944
If yes, add a yamlRestTest to verify the issued fixed.

Hi @penghuo, good callout. I was actually taking a look for the `-` and ended up that is another limitation from java regex library. So rechecked and in summary:

### Java Regex Named Group Character Rules

According to Java's Pattern class documentation, named capture groups follow these rules:

#### Valid Characters:

- Must start with: Letter (a-z, A-Z)
- Can contain: Letters (a-z, A-Z) and digits (0-9)

#### Invalid Characters (will cause PatternSyntaxException):

- Underscore (_) - This is what we're handling with our validation
- Hyphen/Dash (-) - This is what you tested with curl
- Period/Dot (.)
- Space ( )
- Special characters (@, #, $, %, ^, &, *, etc.)
- Unicode characters beyond ASCII
- Cannot start with digit (0-9)

### Java Documentation:

#### The official Java documentation for this is in:
- Java Pattern class: java.util.regex.Pattern
- Specific section: "Named-capturing group" in the Pattern syntax documentation
- Rule: Named groups use the syntax (?<name>X) where name must be a valid Java identifier but with more restrictions

#### The exact Java rule is:

```
  // From Java's Pattern class source code:
  // Named group name must match: [a-zA-Z][a-zA-Z0-9]*
```

#### Quick Test Examples:

```
# These will FAIL in Java regex:
(?<domain_name>...)  # underscore
(?<domain-name>...)  # hyphen  
(?<domain.name>...)  # dot
(?<domain name>...)  # space
(?<@domain>...)      # special char
(?<1domain>...)      # starts with digit

# These will WORK in Java regex:
(?<domain>...)       # letters only
(?<domainName>...)   # camelCase
(?<domain1>...)      # letters + numbers
(?<DOMAIN>...)       # uppercase
```

RyanL1997:
According to the above, since the limitation is coming from the upstream library we are using, I just unified the error msg to handle all the illegal character cases in extraction commands like `rex` and `parse` and for the above `parse` related issue I also left a comment of work around by using `rename` command: https://github.com/opensearch-project/sql/issues/3944#issuecomment-3383678537. 

cc @penghuo

core/src/main/java/org/opensearch/sql/expression/parse/RegexCommonUtils.java
```diff
@@ -50,20 +53,65 @@ public static Pattern getCompiledPattern(String regex) {
...
+    }
+
+    for (int i = 0; i < groupName.length(); i++) {
+      char c = groupName.charAt(i);
+      if (!Character.isLetterOrDigit(c)) {
+        return false;
+      }
+    }
+
+    return true;
```
penghuo: Simpliy use REGEX?

```
return groupName.matches("^[A-Za-z][A-Za-z0-9_]*$");
```

RyanL1997: fixed, forgot 1 commit 🫠

ykmr1224:
Technically we should be able to support the invalid characters by rewriting regex and map extracted values back to original name like:

1. original pattern: `.+@(?<domain_name>.+)`
2. rewritten pattern: `.+@(?<domainname>.+)`   (maintain mapping `domainname` -> `domain_name` internally)
3. assign extracted value: domain_name = extracted['domainname']

This could cause name overwrap, but we can workaround by adding suffix, etc. to come up with unique name.

RyanL1997:
Hi @ykmr1224 , thanks for the input. And that was an interesting approach and I was actually did a PoC of your suggestion on my local. However, I do find some limitations of this approach:
- If the renaming of the pattern is just removing the unsupported character, what if the renamed pattern after removal is duplicate with a existing field? (e.g. the index field contains a `username`, user would like to extract the alias of the email (`xyz@email.com` -> `xyz`) as a column called `user_name` and in the above logic it will be renamed as `username` which is conflict with the index field.
- Stack traces and error messages refer to rewritten names (e.g. in `explain`) , which I think will cause the confusions for users.
- I haven't try it on any of the benchmarking dataset yet for the re-writing approach, but from the time complexity wise
  - Current approach: `O(1)` per validation
  - Rewriting approach: `O(k)` where k=number of named groups (for mapping tables as we are re-writing the column names)
 
According to the above, I think we can keep as the current limitation handling approach for now, if there are some feature request coming from the users we can think about actually supporting these characters as a new feature requests. But if that day comes, I truly think the correct approach is to find a PCRE style library for regex patterns.

ykmr1224:
@RyanL1997
Thanks for PoCing.
In case of name overwrap, we can add suffix like `username1`, and increment in case it still overwrap. (it will eventually find unique name)
i.e: `(?<user_name>.+)(?<username>.+)(?<username1>.+)` => `(?<username2>.+)(?<username>(?<username1>.+)`, `mapping = {username2 => user_name, username => username, username1 => username1}`

Computation cost for regex rewrite should be ignorable since this happens only once during query analysis. And rename after extraction should be just an alias or projection (should not affect performance in my understanding).

It can be separate task from this PR, btw.
And I think we want to quickly check the possible usage of this use case to see the priority.

RyanL1997:
> It can be separate task from this PR, btw.
And I think we want to quickly check the possible usage of this use case to see the priority.

Transferring some internal communication over here - since the above issue is not blocking this change. This change is ready to be merged and I have created a issue https://github.com/opensearch-project/sql/issues/4549 to discover the feasibility of the above suggestion.

Issue #4220, opened 2025-09-03
[RFC] Standardizing New Feature Documentation in OpenSearch PPL


anasalkouz:
## Problem Statement
Currently, while new features are highlighted in release notes, there's no standardized way to indicate new features or enhancements in the OpenSearch PPL language documentation. This makes it difficult for users to:
1. Identify which features are available in their version.
2. Track when specific enhancements were introduced.
3. Understand version compatibility requirements.

### Current PPL Documentation Examples
1. New Commands (e.g., [regex](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/regex.rst#version) command):
```rst
Version
-------
This command was introduced in OpenSearch 3.3.
```

2. Enhanced Features (e.g., [fields](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/fields.rst#enhanced-features-version-330) command):
```rst
Enhanced Features (Version 3.3.0)
--------------------------------
```

### 1. Standardized Version Notation
I am proposing a format that is commonly used in other documentation softwares like [Sphinx](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded) and [Python](https://devguide.python.org/documentation/markup/#paragraph-level-markup).

For all PPL commands and functions, we should introduce a consistent version information and command status block as following:

```rst
Command Name
===========

Version Information
-----------------
.. versionadded:: 3.3.0
    Initial command implementation

.. versionchanged:: 3.4.0
    Added support for feature X

Status
------
[Stable|Experimental]

Description
----------
[Command description]

Syntax
------
[Command syntax]

Parameters
---------
[Parameter details]

Examples
--------
[Usage examples]
```

### 2. Version Compatibility Matrix
Add a version compatibility matrix in the [main PPL index documentation](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/index.rst):

| Command | Added In | Modified | Status |
|---------|----------|----------|--------|
| [regex](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/regex.rst)   | 3.3.0    | -        | Stable |
| [fields](https://github.com/opensearch-project/sql/blob/1.0/docs/user/ppl/cmd/fields.rst)  | 1.0.0    | 3.3.0    | Stable |

Swiddis:
I like the `versionadded` and `versionchanged` tags, though we'll need to implement handling for them when we host the docs on the web.

[...]

PR #4438, opened 2025-10-04
(Test) Add mvappend function for Calcite PPL

ykmr1224:
### Description

### Related Issues

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVAppendFunctionImpl.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVAppendFunctionImpl.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.expression.function.CollectionUDF;
+
+import static org.apache.calcite.sql.type.SqlTypeUtil.createArrayType;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.calcite.adapter.enumerable.NotNullImplementor;
+import org.apache.calcite.adapter.enumerable.NullPolicy;
+import org.apache.calcite.adapter.enumerable.RexToLixTranslator;
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.linq4j.tree.Expressions;
+import org.apache.calcite.linq4j.tree.Types;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.sql.SqlOperatorBinding;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+import org.opensearch.sql.expression.function.UDFOperandMetadata;
+
+/**
+ * MVAppend function that appends all elements from arguments to create an array. Always returns an
+ * array or null for consistent type behavior.
+ */
+public class MVAppendFunctionImpl extends ImplementorUDF {
+
+  public MVAppendFunctionImpl() {
+    super(new MVAppendImplementor(), NullPolicy.ANY);
+  }
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return sqlOperatorBinding -> {
+      RelDataTypeFactory typeFactory = sqlOperatorBinding.getTypeFactory();
+
+      if (sqlOperatorBinding.getOperandCount() == 0) {
+        return typeFactory.createSqlType(SqlTypeName.NULL);
+      }
+
+      RelDataType elementType = determineElementType(sqlOperatorBinding, typeFactory);
+      return createArrayType(
+          typeFactory, typeFactory.createTypeWithNullability(elementType, true), true);
+    };
+  }
+
+  @Override
+  public UDFOperandMetadata getOperandMetadata() {
+    return null;
+  }
+
+  private static RelDataType determineElementType(
+      SqlOperatorBinding sqlOperatorBinding, RelDataTypeFactory typeFactory) {
+    RelDataType mostGeneralType = null;
+    boolean hasStringType = false;
+    boolean hasNumericType = false;
+
+    for (int i = 0; i < sqlOperatorBinding.getOperandCount(); i++) {
+      RelDataType operandType = getComponentType(sqlOperatorBinding.getOperandType(i));
+
+      if (isStringType(operandType)) {
+        hasStringType = true;
+      } else if (isNumericType(operandType)) {
+        hasNumericType = true;
+      }
+
+      mostGeneralType = updateMostGeneralType(mostGeneralType, operandType, typeFactory);
+    }
+
+    if (hasStringType && hasNumericType) {
+      return typeFactory.createSqlType(SqlTypeName.VARCHAR);
+    }
+
+    return mostGeneralType != null
+        ? mostGeneralType
+        : typeFactory.createSqlType(SqlTypeName.VARCHAR);
+  }
+
+  private static RelDataType getComponentType(RelDataType operandType) {
+    if (!operandType.isStruct() && operandType.getComponentType() != null) {
+      return operandType.getComponentType();
+    }
+    return operandType;
+  }
+
+  private static boolean isStringType(RelDataType type) {
+    SqlTypeName typeName = type.getSqlTypeName();
+    return typeName == SqlTypeName.VARCHAR || typeName == SqlTypeName.CHAR;
+  }
+
+  private static boolean isNumericType(RelDataType type) {
+    return SqlTypeName.NUMERIC_TYPES.contains(type.getSqlTypeName());
+  }
+
+  private static RelDataType updateMostGeneralType(
+      RelDataType current, RelDataType candidate, RelDataTypeFactory typeFactory) {
+    if (current == null) {
+      return candidate;
+    }
+
+    RelDataType leastRestrictive = typeFactory.leastRestrictive(List.of(current, candidate));
+    return leastRestrictive != null ? leastRestrictive : current;
+  }
+
+  public static class MVAppendImplementor implements NotNullImplementor {
+    @Override
+    public Expression implement(
+        RexToLixTranslator translator, RexCall call, List<Expression> translatedOperands) {
+      return Expressions.call(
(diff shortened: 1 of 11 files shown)


ahkcs:
>   * Each argument could be an value, an array, or null.

How does the current logic handle primitive arrays?
I tried this test case in `MVAppendFunctionImplTest.java` and it returns `ClassCastException` error

```
  @Test
  public void testMvappendWithPrimitiveIntArray() {
    int[] primitiveArray = {1, 2, 3};
    Object result = MVAppendFunctionImpl.mvappend(primitiveArray, 4);
    assertEquals(Arrays.asList(1, 2, 3, 4), result);
  }
```

docs/user/ppl/functions/collection.rst
```diff
@@ -216,19 +216,74 @@ Return type: STRING
...
+    | Amber, Duke |
+    +-------------+
+
+MVAPPEND
+--------
+
+Description
+>>>>>>>>>>>
+
+Version: 3.4.0
```
ahkcs: I think we can remove the mentioning of version based on the new doc template
https://github.com/opensearch-project/sql/issues/4220#issuecomment-3353061607
@ritvibhatt

ahkcs:
How does the current implementation handle type coercion?
I tested this in `CalciteMVAppendFunctionIT.java` and it passes:
```
@Test
  public void testMvappendWithIntegerAndDouble() throws IOException {
    JSONObject actual =
        executeQuery(
            source(
                TEST_INDEX_BANK,
                "eval result = mvappend(1, 2.5, 3, 4.7) | head 1 | fields result"));

    verifySchema(actual, schema("result", "array"));
    verifyDataRows(actual, rows(List.of(0, 2.5, 0, 4.7)));
  }
  ```

Is it expected the return 0 for integer value when the array has both integer and double? Or do we need some additional handling logic here

File: integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMVAppendFunctionIT.java [hunk 2]
```diff
@@ -0,0 +1,219 @@
...
+    assertTrue(dataRows.length() > 0);
+    var firstRow = dataRows.getJSONArray(0);
+    var resultArray = firstRow.getJSONArray(2);
+    assertEquals(2, resultArray.length());
+    assertEquals(firstRow.getString(0), resultArray.getString(0));
+    assertEquals(firstRow.getString(1), resultArray.getString(1));
+  }
+
+  @Test
+  public void testMvappendWithFieldsAndLiterals() throws IOException {
```
ahkcs, line 127: Can we change these test cases to use `verifyDataRows`? Test result should be stable after testing on my local

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMVAppendFunctionIT.java
```diff
@@ -0,0 +1,219 @@
...
+            source(
+                TEST_INDEX_BANK,
+                "eval result = mvappend(1, 'text', 2.5) | head 1 | fields result"));
+
+    verifySchema(actual, schema("result", "array"));
+    verifyDataRows(actual, rows(List.of("1", "text", "2.5")));
+  }
+
+  @Test
+  public void testMvappendWithRealFields() throws IOException {
```
ahkcs, line 107: ditto

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMVAppendFunctionIT.java
```diff
@@ -0,0 +1,219 @@
...
+        schema("combined", "array"));
+
+    var dataRows = actual.getJSONArray("datarows");
+    assertTrue(dataRows.length() > 0);
+    var firstRow = dataRows.getJSONArray(0);
+    assertEquals(2, firstRow.getJSONArray(2).length());
+  }
+
+  @Test
+  public void testMvappendWithComplexExpression() throws IOException {
```
ahkcs, line 198: ditto

docs/user/ppl/functions/collection.rst
```diff
@@ -216,19 +216,74 @@ Return type: STRING
...
-    | result                                   |
-    |------------------------------------------|
-    | "Amber, Duke"                            |
-    +------------------------------------------+
-    
+    +--------------+
+    | result       |
+    |--------------|
+    | [1,text,2.5] |
+    +--------------+
```
ahkcs, line 303: Can we add more doctests for other type coercion such as Integer and double?

docs/user/ppl/functions/collection.rst
```diff
@@ -216,19 +216,74 @@ Return type: STRING
...
+
+MVAPPEND
+--------
+
+Description
+>>>>>>>>>>>
+
+Version: 3.4.0
+
+Usage: mvappend(value1, value2, value3...) appends all elements from arguments to create an array. Flattens array arguments and collects all individual elements. Always returns an array or null for consistent type behavior.
```
penghuo, line 225: In example, add following use cases
1. Handle dupldate values. mvappend(1, 1)
2. Handle null values. mvappend(1, null)
3. Nested use case, mvapend(1, mvappend(2, 3))

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVAppendFunctionImpl.java
```diff
@@ -0,0 +1,160 @@
...
+      } else if (isNumericType(operandType)) {
+        hasNumericType = true;
+      }
+
+      mostGeneralType = updateMostGeneralType(mostGeneralType, operandType, typeFactory);
+    }
+
+    if (hasStringType && hasNumericType) {
+      return typeFactory.createSqlType(SqlTypeName.VARCHAR);
+    }
```
penghuo: 3 questions
1. Could explain mostGeneralType rule? 
2. if (hasStringType && hasNumericType) it should return ANY?
3. can we leverage typeFactory.leastRestrictive(List.of(...))?

docs/user/ppl/functions/collection.rst
```diff
@@ -216,19 +216,74 @@ Return type: STRING
...
+
+    os> source=accounts | eval names_array = array(firstname, lastname) | eval result = mvjoin(names_array, ', ') | fields result | head 1
+    fetched rows / total rows = 1/1
+    +-------------+
+    | result      |
+    |-------------|
+    | Amber, Duke |
+    +-------------+
+
+MVAPPEND
```
penghuo, line 219: What is difference with ARRAY?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVAppendFunctionImpl.java
```diff
@@ -0,0 +1,160 @@
...
+    for (Object arg : args) {
+      if (arg == null) {
+        continue;
+      }
+
+      if (arg instanceof List) {
+        addListElements((List<?>) arg, elements);
+      } else if (arg.getClass().isArray()) {
+        addArrayElements((Object[]) arg, elements);
+      } else {
```
penghuo: Not sure this code path been used or not. If yes, it will throw exception

class [I cannot be cast to class [Ljava.lang.Object; ([I and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class [I cannot be cast to class [Ljava.lang.Object; ([I and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')

for example

```
  @Test
  public void test() {
    Object arg = new int[] {1, 2, 3};
    List<Object> objects = new ArrayList<>();

    if (arg.getClass().isArray()) {
      addArrayElements((Object[]) arg, objects);
    }
  }

  private static void addArrayElements(Object[] array, List<Object> elements) {
    for (Object item : array) {
      if (item != null) {
        elements.add(item);
      }
    }
  }
```

penghuo (changes requested):

docs/user/ppl/functions/collection.rst
ykmr1224: Added more examples. A bit unsure if we should put so many examples...

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVAppendFunctionImpl.java
ykmr1224: Fixed the logic to simply return ANY when any incompatible type is detected. typeFactory.leastRestrictive does not do much job for incompatible types (It only works between structs)

docs/user/ppl/functions/collection.rst
ykmr1224: ARRAY cannot take array as input. It cannot handle null as well.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVAppendFunctionImpl.java
ykmr1224: Thanks, it is fixed.

ykmr1224:
Fixed logic to simply return ANY in case of mixed types. This spec might be debatable and welcome your opinion, but the idea is to leave users to cast to their preferred type. (Anyway, `spath` would extract values as `ANY` from schema perspective.)

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVAppendFunctionImpl.java
penghuo: nit: Does AbstractTypeCoercion.getTightestCommonType() works?

ykmr1224: The basic idea of `AbstractTypeCoercion.getTightestCommonType()` is returning widened type with no precision loss, and works only for nullability difference and number precision difference. Is it mainly used for binary operator (or function) to align the input types to execute operation, in my understanding.

docs/user/ppl/functions/collection.rst
```diff
@@ -216,19 +200,104 @@ Return type: STRING
...
-    | "a,b,c"                            |
-    +------------------------------------+
+    +--------------+
+    | result       |
+    |--------------|
+    | [Amber,Duke] |
+    +--------------+
 
-    PPL> source=accounts | eval names_array = array(firstname, lastname) | eval result = mvjoin(names_array, ', ') | fields result | head 1
+    os> source=people | eval result = mvappend(1, 'text', 2.5) | fields result | head 1
```
dai-chen, line 297: This is the main reason to add a UDF rather than reusing `array_append`?

ykmr1224: Main motivation is to handle null/value/array seemlessly, so that value merge operation won't become too complicated.

PR #4451, opened 2025-10-07
Add replace command with Calcite

### LINKED ISSUE #3975: [RFC] Add replace Command to PPL (Calcite Engine)

ahkcs:
## Description

Implement the `replace` command in PPL to replace text patterns in specified fields. This PR includes the grammar implementation and basic replacement functionality. This implementation reuses the existing replace core logic. Missing / new features for regex support will be added in a separate PR.

Original PR: #4248 

---

## Syntax

```
<source> | replace '<pattern>' WITH '<replacement>' IN field_list
```

* **pattern**: The text pattern to search for (case-sensitive)
* **replacement**: The text to replace matches with
* **field_list**: Comma-separated list of fields to perform replacement in
  The `replace` command modifies the specified fields in-place with the replaced text.

---

### Expected Behavior

* **Action**: Modifies specified fields in-place with replaced text values
* **Scope**: Operates only on the specified fields
* **Data Modification**: Specified fields are updated with replacement text
* **Case Sensitivity**: Text literal matching is case-sensitive
* **Pattern Type**: Only supports literal string patterns (wildcards or regex support added later)

### Implementation Approach

* Modifies specified fields in-place using string replacement
* Performs literal string replacement in the specified fields
* Other fields remain unchanged
* Validates the existence of fields and correctness of pattern / replacement parameters

---

## Example Queries

```sql
-- Replace in a single field  
source=logs | replace 'error' WITH 'ERROR' IN message  

-- Replace in multiple fields  
source=logs | replace 'USA' WITH 'United States' IN country, state  

-- Combine with other commands  
source=logs  
  | where level = 'error'  
  | replace 'error' WITH 'ERROR' IN message  
  | sort @timestamp  

-- Replace and select specific fields  
source=logs | replace 'error' WITH 'ERROR' IN message | fields message, level  
```

---

## Output Schema

For each field specified in the `IN` clause, the field is modified in-place:

* **Specified fields**: modified with replacement text
* **Other fields**: remain unchanged

**Example**
Input:

```
message = "error occurred", level = "error"
```

After `replace 'error' WITH 'ERROR' IN message, level` →

```
message = "ERROR occurred"  
level = "ERROR"
```

---

## Resolves
#3975

---

## Check List

* [x] New functionality includes tests
* [x] New functionality has documentation
* [x] Javadoc added for new components
* [x] User manual / user documentation updated
* [x] PPL command checklist confirmed
* [x] Companion API changes PR created (if applicable)
* [x] Commits are signed per DCO (`--signoff` or `-s`)
* [x] Public documentation / issue or PR created

By submitting this pull request, I confirm that my contribution is made under the terms of the **Apache 2.0 license**.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
+++ b/core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
@@ -6,7 +6,6 @@
 package org.opensearch.sql.ast.tree;
 
 import com.google.common.collect.ImmutableList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import lombok.EqualsAndHashCode;
@@ -16,8 +15,6 @@ import lombok.ToString;
 import org.jetbrains.annotations.Nullable;
 import org.opensearch.sql.ast.AbstractNodeVisitor;
 import org.opensearch.sql.ast.expression.Field;
-import org.opensearch.sql.ast.expression.Literal;
-import org.opensearch.sql.ast.expression.UnresolvedExpression;
 
 @Getter
 @Setter
@@ -37,53 +34,6 @@ public class Replace extends UnresolvedPlan {
   public Replace(List<ReplacePair> replacePairs, Set<Field> fieldList) {
     this.replacePairs = replacePairs;
     this.fieldList = fieldList;
-  }
-
-  /**
-   * Backward-compatible constructor with single pattern/replacement pair.
-   *
-   * @param pattern Pattern literal
-   * @param replacement Replacement literal
-   * @param fieldList Set of fields to apply replacement to
-   */
-  public Replace(
-      UnresolvedExpression pattern, UnresolvedExpression replacement, Set<Field> fieldList) {
-    // Convert single pair to list for internal consistency
-    if (!(pattern instanceof Literal) || !(replacement instanceof Literal)) {
-      throw new IllegalArgumentException(
-          "Pattern and replacement must be literals in Replace command");
-    }
-    this.replacePairs =
-        Collections.singletonList(new ReplacePair((Literal) pattern, (Literal) replacement));
-    this.fieldList = fieldList;
-  }
-
-  /**
-   * Get the pattern from the first replacement pair (for backward compatibility).
-   *
-   * @return Pattern expression
-   * @deprecated Use {@link #getReplacePairs()} instead
-   */
-  @Deprecated
-  public UnresolvedExpression getPattern() {
-    if (replacePairs.isEmpty()) {
-      throw new IllegalStateException("No replacement pairs available");
-    }
-    return replacePairs.get(0).getPattern();
-  }
-
-  /**
-   * Get the replacement from the first replacement pair (for backward compatibility).
-   *
-   * @return Replacement expression
-   * @deprecated Use {@link #getReplacePairs()} instead
-   */
-  @Deprecated
-  public UnresolvedExpression getReplacement() {
-    if (replacePairs.isEmpty()) {
-      throw new IllegalStateException("No replacement pairs available");
-    }
-    return replacePairs.get(0).getReplacement();
   }
 
   @Override
--- a/ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
+++ b/ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
@@ -284,9 +284,15 @@ public class PPLQueryDataAnonymizer extends AbstractNodeVisitor<String, String> 
     // Get the child query string
     String child = node.getChild().get(0).accept(this, context);
 
-    // Get pattern and replacement expressions
-    String pattern = visitExpression(node.getPattern());
-    String replacement = visitExpression(node.getReplacement());
+    // Build pattern/replacement pairs string
+    String pairs =
+        node.getReplacePairs().stream()
+            .map(
+                pair ->
+                    StringUtils.format(
+                        "%s WITH %s",
+                        visitExpression(pair.getPattern()), visitExpression(pair.getReplacement())))
+            .collect(Collectors.joining(", "));
 
     // Get field list
     String fieldListStr =
@@ -294,8 +300,7 @@ public class PPLQueryDataAnonymizer extends AbstractNodeVisitor<String, String> 
             + node.getFieldList().stream().map(Field::toString).collect(Collectors.joining(", "));
 
     // Build the replace command string
-    return StringUtils.format(
-        "%s | replace %s WITH %s%s", child, pattern, replacement, fieldListStr);
+    return StringUtils.format("%s | replace %s%s", child, pairs, fieldListStr);
   }
 
   /** Build {@link LogicalAggregation}. */


core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,97 @@
...
+import org.opensearch.sql.ast.expression.Field;
+import org.opensearch.sql.ast.expression.Literal;
+import org.opensearch.sql.ast.expression.UnresolvedExpression;
+
+@Getter
+@Setter
+@ToString
+@EqualsAndHashCode(callSuper = false)
+public class Replace extends UnresolvedPlan {
+  private final UnresolvedExpression pattern;
```
Swiddis: suggestion: We should annotate these with `@NotNull`.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -381,6 +382,25 @@ public UnresolvedPlan visitRenameCommand(RenameCommandContext ctx) {
             .collect(Collectors.toList()));
   }
 
+  /** Replace command. */
+  @Override
+  public UnresolvedPlan visitReplaceCommand(OpenSearchPPLParser.ReplaceCommandContext ctx) {
+    UnresolvedExpression pattern = internalVisitExpression(ctx.pattern);
+    UnresolvedExpression replacement = internalVisitExpression(ctx.replacement);
+
+    List<Field> fieldList =
```
Swiddis: suggestion: Should we have a global method for turning this into a LinkedHashSet?

This is currently checked in `validate()`, but I don't think this is the only command to rely on a fieldList. If we make Replace take a Set instead of a List, we simplify the validation logic there, and we can push the responsibility of deduplicating fieldLists to shared parsing code.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReplaceTest.java
```diff
@@ -0,0 +1,266 @@
...
+        "SELECT `EMPNO`, `ENAME`, `JOB`, `MGR`, `HIREDATE`, `SAL`, `COMM`, `DEPTNO`, "
+            + "REPLACE(`JOB`, 'CLERK', 'EMPLOYEE') `new_JOB`\n"
+            + "FROM `scott`.`EMP`\n"
+            + "WHERE `JOB` = 'CLERK'\n"
+            + "ORDER BY `SAL`";
+
+    verifyPPLToSparkSQL(root, expectedSparkSql);
+  }
+
+  @Test(expected = Exception.class)
```
Swiddis: nit: Catch a more specific Exception

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
ahkcs: Changed to `@Nullable`, received similar comment before: https://github.com/opensearch-project/sql/pull/3878#discussion_r2297019972

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
ahkcs: Updated to accept Set<Field> instead of List<Field>, removing duplicate validation logic

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLReplaceTest.java
ahkcs: updated to be more specific

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
Swiddis: Fwiw, I also prefer using optional in fields, despite the warning. I'm not sure why that warning exists. :shrug: 

But why mark it nullable if we validate that they can't be null?

ahkcs: We marked child as @Nullable, which we didn't include in the validate logic

ahkcs:
I updated the implementation to change from creating a new column `new_*` to in-place replacement for fields

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -155,6 +156,7 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
   private final CalciteRexNodeVisitor rexVisitor;
   private final CalciteAggCallVisitor aggVisitor;
   private final DataSourceService dataSourceService;
+  private static final String NEW_FIELD_PREFIX = "new_";
```
ykmr1224: Is this used?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2397,6 +2399,51 @@ public RelNode visitValues(Values values, CalcitePlanContext context) {
...
+    for (String fieldToReplace : fieldsToReplace) {
+      if (!availableFields.contains(fieldToReplace)) {
+        throw new IllegalArgumentException(
+            String.format(
+                "field [%s] not found; input fields are: %s", fieldToReplace, fieldNames));
+      }
+    }
+
+    List<RexNode> projectList = new ArrayList<>();
+    List<String> newFieldNames = new ArrayList<>();
```
ykmr1224: Not sure why we need this list. Looks like it becomes the copy of fieldNames.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2397,6 +2399,51 @@ public RelNode visitValues(Values values, CalcitePlanContext context) {
...
+
+    // Validate that all fields to replace exist in the current schema
+    Set<String> availableFields = new HashSet<>(fieldNames);
+    for (String fieldToReplace : fieldsToReplace) {
+      if (!availableFields.contains(fieldToReplace)) {
+        throw new IllegalArgumentException(
+            String.format(
+                "field [%s] not found; input fields are: %s", fieldToReplace, fieldNames));
+      }
+    }
```
ykmr1224: Let's extract this.

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -204,6 +206,10 @@ renameCommand
    : RENAME renameClasue (COMMA? renameClasue)*
    ;
 
+replaceCommand
+   : REPLACE pattern=stringLiteral WITH replacement=stringLiteral IN fieldList
```
ykmr1224: Will we enable multiple patterns like `REPLACE a WITH a1, b with b1 IN ...`

ahkcs:
> A high level QQ: any existing command can be reused instead of duplicating same/similar logic? e.g., make this subset of what rex command can support or rewrite to string replace function?

Thanks for the suggestion @dai-chen! For our replace command, we are not adopting regex support so our existing logic(such as replace eval function, rex sed mode) does not work here

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Removed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Extracted

ppl/src/main/antlr/OpenSearchPPLParser.g4
ahkcs: That's a good point! I have updated the grammar. The grammar now supports comma-separated replacement pairs, also added test cases

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,144 @@
...
+   * @return Replacement expression
+   * @deprecated Use {@link #getReplacePairs()} instead
+   */
+  @Deprecated
+  public UnresolvedExpression getReplacement() {
+    if (replacePairs.isEmpty()) {
+      throw new IllegalStateException("No replacement pairs available");
+    }
+    return replacePairs.get(0).getReplacement();
+  }
```
ykmr1224: Let's delete instead of deprecate.

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,144 @@
...
+   * @return Pattern expression
+   * @deprecated Use {@link #getReplacePairs()} instead
+   */
+  @Deprecated
+  public UnresolvedExpression getPattern() {
+    if (replacePairs.isEmpty()) {
+      throw new IllegalStateException("No replacement pairs available");
+    }
+    return replacePairs.get(0).getPattern();
+  }
```
ykmr1224: Let's delete instead of deprecate.

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Backward-compatible constructor with single pattern/replacement pair.
+   *
+   * @param pattern Pattern literal
+   * @param replacement Replacement literal
+   * @param fieldList Set of fields to apply replacement to
+   */
+  public Replace(
+      UnresolvedExpression pattern, UnresolvedExpression replacement, Set<Field> fieldList) {
```
ykmr1224: Is this used?

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  public void validate() {
+    if (replacePairs == null || replacePairs.isEmpty()) {
+      throw new IllegalArgumentException(
+          "At least one pattern/replacement pair is required in Replace command");
+    }
+
+    // Validate each pair
+    for (ReplacePair pair : replacePairs) {
```
dai-chen: Is all or most of these validation already guaranteed by ANTLR grammar?

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,109 @@
...
+ |----------+----------------------|
+ | Illinois | 880 Holmes Lane      |
+ | TN       | 671 Bristol Street   |
+ | VA       | 789 Madison Street   |
+ | MD       | 467 Hutchinson Court |
+ +----------+----------------------+
+
+
+Example 3: Replace with IN clause and other commands
+------------------------------------
```
dai-chen: This example is not necessary as all examples have IN clause?
Meanwhile there is no example for multiple replace pairs?

docs/user/ppl/cmd/replace.rst
```diff
@@ -0,0 +1,109 @@
...
+.. contents::
+ :local:
+ :depth: 2
+
+
+Description
+============
+Using ``replace`` command to replace text in one or more fields in the search result.
+
+Note: This command is only available when Calcite engine is enabled.
```
dai-chen, line 16: Shall we mention version since? I forgot what's the standard way we proposed.. @ritvibhatt

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2400,6 +2402,57 @@ public RelNode visitValues(Values values, CalcitePlanContext context) {
...
+    visitChildren(node, context);
+
+    List<String> fieldNames = context.relBuilder.peek().getRowType().getFieldNames();
+
+    // Create a set of field names to replace for quick lookup
+    Set<String> fieldsToReplace =
+        node.getFieldList().stream().map(f -> f.getField().toString()).collect(Collectors.toSet());
+
+    // Validate that all fields to replace exist in the current schema
+    validateFieldsExist(fieldsToReplace, fieldNames);
```
dai-chen: visitField behind line 2424 already cover this?

docs/user/ppl/cmd/replace.rst
ahkcs: I think for the current proposal, we don't need to add version info
Reference: https://github.com/opensearch-project/sql/issues/4220#issuecomment-3353061607

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Without the explicit validation logic, the invalid field would be silently ignored - no error, no replacement
 The validation logic is used to throw clear error message when users specify invalid fields

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
ahkcs: Removed redundant validation

docs/user/ppl/cmd/replace.rst
ahkcs: That's a good point! Removed unnecessary example and added example for multiple replace pairs

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
dai-chen: I'm a little confused. You mean each command needs to validate field by itself.
Here is the `relbuilder.field()` source code I'm looking at:

```
  public RexInputRef field(int inputCount, int inputOrdinal, String fieldName) {
    final Frame frame = peek_(inputCount, inputOrdinal);
    final List<String> fieldNames = Pair.left(frame.fields());
    int i = fieldNames.indexOf(fieldName);
    if (i >= 0) {
      return field(inputCount, inputOrdinal, i);
    } else {
      throw new IllegalArgumentException("field [" + fieldName
          + "] not found; input fields are: " + fieldNames);
    }
  }
```

ahkcs: In the implementation, we iterate over fieldNames (the actual schema fields) and only call field() for fields that exist, relBuilder.field() never gets called with invalid fields - so they're silently skipped.

I think it's a good catch that you're suggesting we should leverage relBuilder.field()'s built-in validation instead of duplicating it. I have updated the logic to remove this validation logic and use the build-in validation instead by iterating all the fields instead of the actual schema fields.

dai-chen (approved):
Thanks for the changes!

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
ahkcs: Updated

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
ahkcs: Updated

core/src/main/java/org/opensearch/sql/ast/tree/Replace.java
ahkcs: Updated to remove unused parameter

core/src/main/java/org/opensearch/sql/ast/tree/ReplacePair.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import org.opensearch.sql.ast.expression.Literal;
+
+/** A pair of pattern and replacement literals for the Replace command. */
+@Getter
+@AllArgsConstructor
+@EqualsAndHashCode
+@ToString
```
ykmr1224, line 18: optional: consider using `@Value`

PR #4456, opened 2025-10-08
Support Regex for replace eval function

ahkcs:
**Description**
Enhances the `replace()` function to support regular expressions, enabling advanced text manipulation in PPL queries.

---

**Examples**

Before (literal only):

```sql
source=logs | eval cleaned = replace(message, "ERROR", "WARN")
```

After (full regex):

```sql
# Remove digits
source=logs | eval cleaned = replace(message, '\d+', '')

# Swap date format (MM/DD/YYYY → DD/MM/YYYY)
source=logs | eval reformatted = replace(date, '^(\d{2})/(\d{2})/(\d{4})$', '\2/\1/\3')

# Extract domain from email
source=users | eval domain = replace(email, '.*@(.+)', '\1')
```

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
@@ -239,6 +239,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.StringJoiner;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
@@ -688,6 +690,24 @@ public class PPLFuncImpTable {
       register(
           REPLACE,
           (RexBuilder builder, RexNode... args) -> {
+            // Validate regex pattern at query planning time
+            if (args.length >= 2 && args[1] instanceof RexLiteral) {
+              RexLiteral patternLiteral = (RexLiteral) args[1];
+              String pattern = patternLiteral.getValueAs(String.class);
+              if (pattern != null) {
+                try {
+                  // Compile pattern to validate it - this will throw PatternSyntaxException if
+                  // invalid
+                  Pattern.compile(pattern);
+                } catch (PatternSyntaxException e) {
+                  // Convert to IllegalArgumentException so it's treated as a client error (400)
+                  throw new IllegalArgumentException(
+                      String.format("Invalid regex pattern '%s': %s", pattern, e.getDescription()),
+                      e);
+                }
+              }
+            }
+
             if (args.length == 3 && args[2] instanceof RexLiteral) {
               RexLiteral literal = (RexLiteral) args[2];
               String replacement = literal.getValueAs(String.class);
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLStringBuiltinFunctionIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLStringBuiltinFunctionIT.java
@@ -399,6 +399,51 @@ public class CalcitePPLStringBuiltinFunctionIT extends PPLIntegTestCase {
     verifyDataRows(actual, rows("Jane", 20));
   }
 
+  @Test
+  public void testReplaceWithInvalidRegexPattern() {
+    // Test invalid regex pattern - unclosed character class
+    Throwable e1 =
+        assertThrowsWithReplace(
+            Exception.class,
+            () ->
+                executeQuery(
+                    String.format(
+                        "source=%s | eval result = replace(firstname, '[unclosed', 'X') | fields"
+                            + " firstname, result",
+                        TEST_INDEX_ACCOUNT)));
+    verifyErrorMessageContains(e1, "Invalid regex pattern");
+    verifyErrorMessageContains(e1, "Unclosed character class");
+    verifyErrorMessageContains(e1, "400 Bad Request");
+
+    // Test invalid regex pattern - unclosed group
+    Throwable e2 =
+        assertThrowsWithReplace(
+            Exception.class,
+            () ->
+                executeQuery(
+                    String.format(
+                        "source=%s | eval result = replace(firstname, '(invalid', 'X') | fields"
+                            + " firstname, result",
+                        TEST_INDEX_ACCOUNT)));
+    verifyErrorMessageContains(e2, "Invalid regex pattern");
+    verifyErrorMessageContains(e2, "Unclosed group");
+    verifyErrorMessageContains(e2, "400 Bad Request");
+
+    // Test invalid regex pattern - dangling metacharacter
+    Throwable e3 =
+        assertThrowsWithReplace(
+            Exception.class,
+            () ->
+                executeQuery(
+                    String.format(
+                        "source=%s | eval result = replace(firstname, '?invalid', 'X') | fields"
+                            + " firstname, result",
+                        TEST_INDEX_ACCOUNT)));
+    verifyErrorMessageContains(e3, "Invalid regex pattern");
+    verifyErrorMessageContains(e3, "Dangling meta character");
+    verifyErrorMessageContains(e3, "400 Bad Request");
+  }
+
   private void prepareTrim() throws IOException {
     Request request1 =
         new Request("PUT", "/opensearch-sql_test_index_state_country/_doc/5?refresh=true");
--- a/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
@@ -14,7 +14,6 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.regex.PatternSyntaxException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchException;
@@ -60,9 +59,7 @@ public class RestPPLQueryAction extends BaseRestHandler {
         || e instanceof QueryEngineException
         || e instanceof SyntaxCheckException
         || e instanceof DataSourceClientException
-        || e instanceof IllegalAccessException
-        || e instanceof PatternSyntaxException
-        || (e.getCause() instanceof PatternSyntaxException);
+        || e instanceof IllegalAccessException;
   }
 
   @Override


core/src/test/java/org/opensearch/sql/expression/function/udf/ReplaceFunctionTest.java
RyanL1997, line 1: Instead of creating the new UDF we should re-use the existing supported sql opperator. You can reference to my previous implementation at `sed` in `rex` that I was doing the same thing:
- https://github.com/opensearch-project/sql/pull/4241

For more details you can also reference to: https://github.com/opensearch-project/sql/pull/4109#discussion_r2305922467
```
REX_SED - Complete native Calcite optimization
REGEXP_REPLACE_PG_3 for basic substitution
REGEXP_REPLACE_PG_4 for flagged substitution (g, i)
REGEXP_REPLACE_5 for nth occurrence
TRANSLATE3 for transliteration (y/from/to/)
```
As you can see there are various of options that available for regex replace and I think you can choose the one you need.

RyanL1997: More specifically, you may need to use this: 
```java
BuiltinFunctionName.INTERNAL_REGEXP_REPLACE_3,
```

ahkcs: Hi @RyanL1997 , thanks for the suggestion, I have updated the implementation to remove the custom UDF and leverage the existing sql operator

RyanL1997:
@ahkcs , also I was wondering for the usage of the regex pattern in this function, will it support captured name group? If that is the case, there are some known limitation of java regex library. see details at https://github.com/opensearch-project/sql/pull/4434#issuecomment-3383457347, and lets see if we need to call this out in the related docs.

ahkcs:
> @ahkcs , also I was wondering for the usage of the regex pattern in this function, will it support captured name group? If that is the case, there are some known limitation of java regex library. see details at [#4434 (comment)](https://github.com/opensearch-project/sql/pull/4434#issuecomment-3383457347), and lets say if we need to call this out in the related docs.

@RyanL1997 , thanks for calling it out,  named groups like (?<domain_name>.+) are used by the rex command, which does have the Java regex limitation you mentioned (no underscores allowed in group names). The replace() function avoids this limitation entirely by using numbered references.

  For example:
  -  Rex fails: rex field=email ".+@(?<domain_name>.+)" (underscore not allowed)
  -  Replace works: eval domain_name = replace(email, '.+@(.+)', '\1') (field name defined in eval, not in regex)

 So the Java named group limitations in #4434 don't apply to replace() function. The field naming is controlled by the eval command, bypassing Java identifier restrictions.

RyanL1997:
> > @ahkcs , also I was wondering for the usage of the regex pattern in this function, will it support captured name group? If that is the case, there are some known limitation of java regex library. see details at [#4434 (comment)](https://github.com/opensearch-project/sql/pull/4434#issuecomment-3383457347), and lets say if we need to call this out in the related docs.
> 
> @RyanL1997 , thanks for calling it out, named groups like (?<domain_name>.+) are used by the rex command, which does have the Java regex limitation you mentioned (no underscores allowed in group names). The replace() function avoids this limitation entirely by using numbered references.
> 
> For example:
> 
> * Rex fails: rex field=email ".+@(?<domain_name>.+)" (underscore not allowed)
> * Replace works: eval domain_name = replace(email, '.+@(.+)', '\1') (field name defined in eval, not in regex)
> 
> So the Java named group limitations in #4434 don't apply to replace() function. The field naming is controlled by the eval command, bypassing Java identifier restrictions.

Make sense, since in the case of this PR, the extraction should be done by `eval` so it won't triggered the java regex pattern limitation for named capture group.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -668,7 +668,7 @@ void populate() {
       registerOperator(LOWER, SqlStdOperatorTable.LOWER);
       registerOperator(POSITION, SqlStdOperatorTable.POSITION);
       registerOperator(LOCATE, SqlStdOperatorTable.POSITION);
-      registerOperator(REPLACE, SqlStdOperatorTable.REPLACE);
+      registerOperator(REPLACE, SqlLibraryOperators.REGEXP_REPLACE_3);
```
RyanL1997: nice.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLStringBuiltinFunctionIT.java
```diff
@@ -24,6 +25,7 @@ public void init() throws Exception {
 
     loadIndex(Index.STATE_COUNTRY);
     loadIndex(Index.STATE_COUNTRY_WITH_NULL);
+    loadIndex(Index.ACCOUNT); // For regex replacement tests
```
RyanL1997: lets remove the verbose comments

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -508,6 +509,27 @@ public RexNode visitFunction(Function node, CalcitePlanContext context) {
...
     }
 
+    // Convert PCRE-style backreferences (\1, \2) to Java-style ($1, $2) for replace function
+    if ("replace".equalsIgnoreCase(node.getFuncName()) && arguments.size() == 3) {
+      RexNode replacementArg = arguments.get(2);
+      if (replacementArg instanceof RexLiteral) {
+        RexLiteral literal = (RexLiteral) replacementArg;
+        String replacement = literal.getValueAs(String.class);
+        if (replacement != null) {
+          // Convert sed backreferences (\1, \2) to Java style ($1, $2)
```
RyanL1997: Is this because the requirement requires java style? What is the expected behavior over here?

ahkcs: Because we use java regex internally but we want to support PCRE style syntax, so we add this conversion logic to translate PCRE syntax to java regex

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLStringBuiltinFunctionIT.java
ahkcs: removed

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
RyanL1997: non-blocking: I remember I do have the same logic for `sed`, maybe we can extract them into a certain util.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -508,6 +509,27 @@ public RexNode visitFunction(Function node, CalcitePlanContext context) {
       }
     }
 
+    // Convert PCRE-style backreferences (\1, \2) to Java-style ($1, $2) for replace function
+    if ("replace".equalsIgnoreCase(node.getFuncName()) && arguments.size() == 3) {
```
ykmr1224: Do we need to do this in `visitFunction`?
Wondering if we can move this logic to the function side.
If that is difficult, let's extract it to an method

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -508,6 +509,27 @@ public RexNode visitFunction(Function node, CalcitePlanContext context) {
...
+      RexNode replacementArg = arguments.get(2);
```
ykmr1224: What happens if regex is invalid? Should we check validity and raise exception?

docs/user/ppl/functions/string.rst
```diff
@@ -219,6 +225,41 @@ Example::
...
+    +-------------+---------------+------------+
+    | Clean phone | Remove vowels | Add prefix |
+    |-------------+---------------+------------|
+    | 5551234567  | hll wrld      | pre_test   |
+    +-------------+---------------+------------+
+
+**Note**: When using regex patterns in PPL queries:
+
+* Backslashes must be escaped (use ``\\`` instead of ``\``) - e.g., ``\\d`` for digit pattern, ``\\w+`` for word characters
+* Backreferences support both PCRE-style (``\1``, ``\2``, etc.) and Java-style (``$1``, ``$2``, etc.) syntax. PCRE-style backreferences are automatically converted to Java-style internally.
```
ykmr1224, line 261: Does it accept $0 to refer whole match?

What happens when reference does not exist? (like when pattern define only 2 groups, but replacement refer $3)

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ahkcs: Moved from CalciteRexNodeVisitor to PPLFuncImpTable

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ahkcs: Example:
  source=test | eval result=replace(field, "(invalid[", "\1")
                                                                         ↑
                                                                  invalid regex

  What happens:

  1. Our code transforms the replacement string: "\1" → "$1" (this always succeeds)
  2. Calcite builds the plan: REGEXP_REPLACE(field, "(invalid[", "$1")
  3. At execution time, Java's Pattern.compile("(invalid[") throws exception

So I think we don't need to add additional error handling here

docs/user/ppl/functions/string.rst
ahkcs: >Does it accept $0 to refer whole match?

Yes, Our conversion code transforms \0 → $0, and Java regex supports $0 to refer to the entire matched string

>What happens when reference does not exist? (like when pattern define only 2 groups, but replacement refer $3)

It will throw exception: ` IndexOutOfBoundsException: No group 3`

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ykmr1224: 3 -> will that exception be translated to 4xx error and user can identify the issue?

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ykmr1224: Nice!

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
ahkcs: Updated the implementation to have 4xx code for users to understand,
Updated error message:
```
Debugging error message: method [POST], host [http://127.0.0.1:56271], URI [/_plugins/_ppl], status line [HTTP/1.1 400 Bad Request]
{
  "error": {
    "reason": "Invalid Query",
    "details": "java.util.regex.PatternSyntaxException: Unclosed character class near index 8\n[unclosed\n        ^",
    "type": "UncheckedExecutionException"
  },
  "status": 400
}
```

ykmr1224: Hmm, adding PatternSyntaxException to client error could potentially hide our bugs. (Even if we make some mistake in somewhere and generate invalid pattern within our logic, the error would be reported as 400)
Can't we simply compile pattern once and raise IllegalArgumentException?

ahkcs: It makes sense, updated the implementation to handle error in PPLFuncImpTable as well, new error message:
```
  Error message: method [POST], host [http://127.0.0.1:61357], URI [/_plugins/_ppl], status line [HTTP/1.1 400 Bad Request]
{
  "error": {
    "reason": "Invalid Query",
    "details": "Cannot resolve function: REPLACE, arguments: [STRING,STRING,STRING], caused by: Invalid regex pattern '[unclosed': Unclosed character class",
  },
  "status": 400
}
```

Also added IT test for this

Issue #4350, opened 2025-09-22
[FEATURE] PPL `per_*` aggregation function support


dai-chen:
**Is your feature request related to a problem?**

PPL currently lacks support for `per_*` aggregation functions (`per_second`, `per_minute`, `per_hour`, `per_day`). These functions calculate rate-based metrics by normalizing aggregated values to specific time units, converting raw counts into meaningful per-unit rates.

Without these functions, users cannot easily perform rate calculations that are common in performance monitoring scenarios, such as calculating packets per second, requests per minute when using the `timechart` command.

**What solution would you like?**

Implement the four `per_*` aggregation functions in PPL:

- `per_second(<value>)` - Returns values normalized to per-second rate
- `per_minute(<value>)` - Returns values normalized to per-minute rate
- `per_hour(<value>)` - Returns values normalized to per-hour rate
- `per_day(<value>)` - Returns values normalized to per-day rate

These functions should work exclusively with the `timechart` command (due to implicit timestamp field dependency):

```
# Sample data
  {\"_time\":\"2025-09-08T10:00:00\", \"packets\":10},
  {\"_time\":\"2025-09-08T10:00:05\", \"packets\":60},
  {\"_time\":\"2025-09-08T10:00:30\", \"packets\":20},
  {\"_time\":\"2025-09-08T10:00:50\", \"packets\":30}

# Example 1
...
| eval _time=strptime(_time, "%Y-%m-%dT%H:%M:%S")
| timechart per_second(packets) span=1m

----------------------|--------------------
_time                 | per_second(packets)
----------------------|--------------------
2025-09-08T10:00:00   | 2                   # (10+60+20+30)/60s

# Example 2
timechart per_second(packets) span=20s

----------------------|--------------------
_time                 | per_second(packets)
----------------------|--------------------
2025-09-08T10:00:00   | 3.5                  # (10+60)/20s
2025-09-08T10:00:20   | 1                    # (20)/20s
2025-09-08T10:00:40   | 1.5                  # (30)/20s

# Example 3
timechart per_minute(packets), per_hour(packets), per_day(packets) span=1m

----------------------|--------------------|--------------------|--------------------
_time                 | per_minute(packets)| per_hour(packets)  | per_day(packets)
----------------------|--------------------|--------------------|--------------------
2025-09-08T10:00:00   | 120                | 7200               | 172800 
```

**What alternatives have you considered?**

[...]

PR #4464, opened 2025-10-08
Add `per_second` function support for `timechart` command

dai-chen:
### Description

This PR adds `per_second()` support to the `timechart` command by implementing **Option 3 (Eval Transformation)**. As a short-term solution, it extends the `Timechart` AST to recognize the per_second function and rewrites into the equivalent math formula shown in the examples below. See issue #4350 for background and alternatives.

#### Examples

<pre>
-- Original
source=events
| timechart span=5m per_second(packets)

-- Rewritten
source=logs
| timechart span=5m <b>sum(packets)</b> as `per_second(packets)`
| eval `per_second(packets)` = `per_second(packets)` / <b>300</b>   -- 5 minutes = 300 seconds
</pre>

For spans whose length varies with the calendar (month/quarter/year), the number of seconds depends on actual start/end timestamps (e.g., September has 30 days; October has 31 days; February in a leap year has 29 days). To ensure correctness, the rewrite computes the exact bucket length dynamically:

<pre>
-- Example: span=2mon
source=logs
| timechart span=2mon sum(packets) as "per_second(packets)"
| eval `per_second(packets)` = `per_second(packets)` /
    <b>timestampdiff(
      SECOND,
      @timestamp,                                  -- span start
      timestampadd(MONTH, 2, @timestamp)           -- span end   (start + 2 months)
    )</b>
</pre>

### Related Issues
Resolves partially #4350.

### Check List


core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
```diff
@@ -62,4 +92,112 @@ public List<UnresolvedPlan> getChild() {
...
+
+    PerFunction perFunc = perFuncOpt.get();
+    Span span = (Span) this.binExpression;
+    Field spanStartTime = AstDSL.field(IMPLICIT_FIELD_TIMESTAMP);
+    Function spanEndTime = timestampadd(span.getUnit(), span.getValue(), spanStartTime);
+    Function spanSeconds = timestampdiff(SECOND, spanStartTime, spanEndTime);
+
+    return eval(
+        timechart(AstDSL.alias(perFunc.aggName, sum(perFunc.aggArg))),
+        let(perFunc.aggName).multiply(perFunc.seconds).dividedBy(spanSeconds));
```
RyanL1997, line 119: What happens if the span evaluates to 0 seconds (e.g., with millisecond spans or edge cases)? Should there be validation or error handling for division by zero?

core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
```diff
@@ -49,8 +78,9 @@ public Timechart useOther(Boolean useOther) {
   }
 
   @Override
-  public Timechart attach(UnresolvedPlan child) {
-    return toBuilder().child(child).build();
+  public UnresolvedPlan attach(UnresolvedPlan child) {
```
RyanL1997, line 81: Just for my knowledge: This `attach()` cannot be called multiple times in a way that causes double-transformation right?

I believe the current implementation is safe because `transformPerFunction() `checks if the aggregate function is a `per_*` function before transforming, and after the first transformation it becomes `sum()`. But still, I'm just trying to understand the logic better.

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -74,6 +74,22 @@ static SpanUnit intervalUnitToSpanUnit(IntervalUnit unit) {
...
+      case SECOND, SECONDS, SEC, SECS, S -> IntervalUnit.SECOND;
+      case MINUTE, MINUTES, MIN, MINS, m -> IntervalUnit.MINUTE;
+      case HOUR, HOURS, HR, HRS, H -> IntervalUnit.HOUR;
+      case DAY, DAYS, D -> IntervalUnit.DAY;
+      case WEEK, WEEKS, W -> IntervalUnit.WEEK;
+      case MONTH, MONTHS, MON, M -> IntervalUnit.MONTH;
+      case QUARTER, QUARTERS, QTR, QTRS, Q -> IntervalUnit.QUARTER;
+      case YEAR, YEARS, Y -> IntervalUnit.YEAR;
+      case UNKNOWN -> IntervalUnit.UNKNOWN;
+      default -> throw new UnsupportedOperationException("Unsupported span unit: " + unit);
```
ykmr1224: I think we want to prefer Java 11 syntax as backport will break it.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTimechartPerFunctionIT.java
```diff
@@ -0,0 +1,89 @@
...
+        schema("@timestamp", "timestamp"),
+        schema("host", "string"),
+        schema("per_second(packets)", "double"));
+    verifyDataRows(
+        result,
+        rows("2025-09-08 10:00:00", "server1", 1.5),
+        rows("2025-09-08 10:02:00", "server1", 0.5),
+        rows("2025-09-08 10:02:00", "OTHER", 1.5));
+  }
+}
```
ahkcs, line 111: Can we add more integration tests here for queries like:
```
source=events_traffic | timechart span=1M per_second(packets)
```
to make it clear on how it handles variable month lengths (28-31 days)

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
dai-chen: Yes, I tried to avoid new feature elsewhere. But this one is following the intervalUntiToSpanUnit above it. Let me change it back to old switch if concerned. Thanks!

core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
dai-chen: Good catch. Let me check if `timechart` has the validation or not. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTimechartPerFunctionIT.java
dai-chen: Sure, let me check if I can add more. Thanks!

dai-chen: Addressed in https://github.com/opensearch-project/sql/pull/4464/commits/995171e7da73d18e63d482f3cdc94241ddadd1ce.

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
dai-chen: Addressed in https://github.com/opensearch-project/sql/pull/4464/commits/995171e7da73d18e63d482f3cdc94241ddadd1ce.

core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
dai-chen: Yes, because of the transformation, any further call (if any) will be no-op. Ideally I think we should have a new AST rewriter abstraction. However, because this is a short-term solution, I finally insert the logic here.

core/src/main/java/org/opensearch/sql/ast/tree/Timechart.java
dai-chen: I quick tested `timechart` command and found negative span value is not allowed in our grammar. But zero-span validation is missing. Let me create an issue and raise separate PR to fix.

<pre>
opensearchsql> source=test_data_2023 | timechart <b>span=0m</b> per_second(packets);
TransportError(500, 'SearchPhaseExecutionException', {'error': 
{'reason':'Error occurred in OpenSearch engine: all shards failed', 'details': 'Shard[0]:
  java.lang.IllegalArgumentException: Zero or negative time interval not supported\n\n
    For more details, please send request for Json format to see the raw response from OpenSearch engine.',
      'type': 'SearchPhaseExecutionException'}, 'status': 400})

opensearchsql> source=test_data_2023 | timechart <b>span=-1m</b> per_second(packets);
{'reason': 'Invalid Query', 'details': "[-] is not a valid term at this part of the query: '...23
  | timechart span=-' <-- HERE. extraneous input '-' expecting {SPANLENGTH, INTEGER_LITERAL}",
    'type': 'SyntaxCheckException'}
</pre>

dai-chen: Created issue: https://github.com/opensearch-project/sql/issues/4527. Thanks!

PR #4475, opened 2025-10-09
Fix the bug of explicit makeNullLiteral for UDT fields

### LINKED ISSUE #4383: [RFC] Fix Type Conflict Schema Merging for Cross-Index Fields in Append and Multisearch Commands
## Problem Statement

Both append and multisearch commands currently handle schema merging with type conflicts using a field renaming workaround instead of implementing intelligent type resolution. While both commands now work functionally, they create confusing user experiences with duplicated semantic fields.

## Current Behavior Analysis

Both append and multisearch commands now work around type conflicts by **renaming conflicting fields**:

**Example with Type Conflict:**
```ppl
// Append syntax:
source=account | fields firstname, age, balance | head 2 |
append [ source=locations | fields description, age, place_id | head 2 ]

// Multisearch syntax:
| multisearch
  [search source=account | fields firstname, age, balance | head 2]
  [search source=locations | fields description, age, place_id | head 2]
```

**Current Result Schema (Both Commands):**
- `firstname` (string) - from account
- `age` (bigint) - from account (original)
- `balance` (bigint) - from account
- `description` (string) - from locations
- `age0` (string) - from locations (renamed to avoid conflict)
- `place_id` (int) - from locations

**Problem:** This creates confusing column names and duplicated semantic fields (`age` and `age0` both represent the same logical concept).

## Test Index Schemas Used

**ACCOUNT Index:**
- `firstname` (string), `age` (bigint), `balance` (bigint), `account_number` (bigint)

**LOCATIONS_TYPE_CONFLICT Index:**
- `description` (string), `age` (string), `place_id` (integer)

## Long-Term Goals

**Expected Behavior**: Both commands should intelligently merge schemas with type conflicts by:

1. **No Type Conflict**: Automatically resolve type conflicts
   - `age` (bigint) + `age` (string) → `age`
   - `balance` (int) + `balance` (double) → `balance`

2. **Consistent Behavior**: Both append and multisearch should use identical schema merging logic

## Potential solution

https://github.com/opensearch-project/sql/issues/4349

songkant-aws:
### Description
Fix the bug that Calcite will error out when CalciteRelBuilder explicitly calling `makeNullLiteral` for UDT type fields. 

UDT type cast is handled by UDF. It's totally fine for `CAST` function, regardless whether the field value is NULL or non NULL. But for some cases in building RelNode, it has the problem of explicitly casting NULL to UDT type by `RexBuilder.makeNullLiteral` method. UDT cast is a RexCall of calling UDF, but `makeNullLiteral` internally expects the returned node to be RexLiteral of NULL. Thus, Calcite throws a hard error due to Java class casting failure.

This PR fixes this issue by adding a simple check and calling super method in case the RexNode to be cast is literally NULL.

### Related Issues
Resolves #4383 

### Check List

### DIFF
--- a/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4383.yml
+++ b/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4383.yml
@@ -1,0 +1,62 @@
+setup:
+  - do:
+      query.settings:
+        body:
+          transient:
+            plugins.calcite.enabled : true
+
+---
+teardown:
+  - do:
+      query.settings:
+        body:
+          transient:
+            plugins.calcite.enabled : false
+
+---
+"Append UDT fields should merge schema successfully":
+  - skip:
+      features:
+        - headers
+        - allowed_warnings
+  - do:
+      indices.create:
+        index: log-test1
+        body:
+          mappings:
+            properties:
+              "timestamp":
+                type: date
+  - do:
+      indices.create:
+        index: log-test2
+        body:
+          mappings:
+            properties:
+              "host":
+                type: ip
+
+  - do:
+      bulk:
+        index: log-test1
+        refresh: true
+        body:
+          - '{"index": {}}'
+          - '{ "timestamp" : "2025-09-04T16:15:00.000Z" }'
+  - do:
+      bulk:
+        index: log-test2
+        refresh: true
+        body:
+          - '{"index": {}}'
+          - '{ "host" : "0.0.0.2" }'
+
+  - do:
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: source=log-test1 | append [ source=log-test2 ]
+
+  - match: { total: 2 }
+  - match: { datarows: [["2025-09-04 16:15:00", null], [null, "0.0.0.2"]] }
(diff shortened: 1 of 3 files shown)


songkant-aws:
cc @ahkcs @yuancu Please let me know if this simple fix can cover the issue.

ahkcs:
Thanks for the fix! I think for #4383, it's more like an issue for long-term goals for `multisearch` command(can be applied to `append` command as well), so this PR might only partially resolve the issue

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4383.yml
```diff
@@ -0,0 +1,62 @@
...
+        body:
+          - '{"index": {}}'
+          - '{ "host" : "0.0.0.2" }'
+
+  - do:
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: source=log-test1 | append [ source=log-test2 ]
```
LantaoJin, line 59: will `source=log-test1 | append [ source=log-test2 ] | fields timestamp` return `- match: { datarows: [["2025-09-04 16:15:00"], [null]] }`?

LantaoJin:
> Thanks for the fix! I think for #4383, it's more like an issue for long-term goals for `multisearch` command(can be applied to `append` command as well), so this PR might only partially resolve the issue

Is `multisearch` available? can you test that after this merged.

songkant-aws:
@ahkcs Makes sense. It's a short-term bug fix to not fail the query. 4383 wants to resolve another issue. I can create another report bug issue to avoid confusion

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4383.yml
songkant-aws: Synced offline. It will append unique name in case of type conflicts and select fields are also successful

songkant-aws:
@LantaoJin Tried the `multisearch` command. It also fixes the same issue in `multisearch` command.

Issue #4433, opened 2025-10-03
[RFC] Dynamic Fields in PPL Calcite Implementation


ykmr1224:
# RFC: Dynamic Fields in PPL Calcite Implementation

## Abstract

This RFC proposes an implementation approach for dynamic fields support in PPL Calcite integration and evaluates the MAP-based approach against alternative solutions. Dynamic fields enable PPL commands like `spath` and `timechart` to handle semi-structured data where column names and types are determined at runtime rather than query planning time.

Related issue: https://github.com/opensearch-project/sql/issues/4112

### Problem Statement

OpenSearch documents often contain nested JSON structures and semi-structured data where field names and types are not known until query execution. Traditional SQL systems assume fixed schemas at planning time, creating challenges for:

1. **spath Command**: Extracting arbitrary JSON paths from documents
2. **timechart Command**: Creating pivot-style results where fields depend on data values
3. **Schema Evolution**: Handling documents with varying structures
4. **Type Preservation**: Maintaining original JSON types (int, boolean, double)

### Requirements

- **Runtime field Generation**: Support fields determined by data content
- **Calcite Integration**: Work within Apache Calcite's type system and optimization framework
- **Compatibility**: Works with existing (and future) PPL commands
- **Performance**: Minimal overhead compared to static field access
- **Extensibility**: Foundation for additional dynamic fields commands
- **Type Preservation**: Maintain original JSON data types

### Architecture Overview

The proposed solution uses a MAP-based storage approach to handle dynamic fields in PPL commands. Instead of trying to predict unknown field names at planning time, this approach stores dynamic fields in a structured MAP container that can be accessed efficiently during query execution.

#### 1. Structured Dynamic Storage
Dynamic fields are stored in a MAP<STRING, ANY> structure where:
- **Keys**: Field names determined at runtime (e.g., JSON paths, pivot column names)
- **Values**: Actual field values with preserved types
- **Container**: Well-defined MAP type that Calcite can optimize

#### 2. Type System Integration
- **MAP Type Support**: Leverages Calcite's native MAP type support
- **Type Preservation**: Uses ANY type to maintain original JSON data types
- **Schema Visibility**: MAP field appears in schema metadata for tools and optimization

#### 3. Two-Phase Field Resolution
- **Static Fields**: Resolved normally through existing schema
- **Dynamic Fields**: Resolved to MAP access operations when field not found in static schema
- **Fallback Strategy**: Clear error handling when fields don't exist in either location

#### Schema Extension
[...]

PR #4477, opened 2025-10-09
Add MAP_CONCAT internal function

ykmr1224:
### Description
- Add MAP_CONCAT internal function for dynamic fields implementation
  - It will be used to merge dynamic fields map among inputs (mainly join), where same key will be overwritten by later input.
- It is referring Spark library implementation
- Added integration test to verify its behavior and access to the function (manual RelNode generation to avoid exposing function in PPL syntax)

### Related Issues
- https://github.com/opensearch-project/sql/issues/4307

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
@@ -62,7 +62,7 @@ public enum BuiltinFunctionName {
   /** Collection functions */
   ARRAY(FunctionName.of("array")),
   ARRAY_LENGTH(FunctionName.of("array_length")),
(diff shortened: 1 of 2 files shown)


ykmr1224:
> Not sure. Do we need map_concat or map_zip_with?
> 
> ```
> spark-sql (default)> SELECT map_zip_with(map(1, 'a', 2, 'b'), map(1, 'x', 2, 'y'), (k, v1, v2) -> array(v1, v2)) as res;
> res
> {1:["a","x"],2:["b","y"]}
> ```

That sounds more related to MAP_APPEND which I am planning to implement, and that will be used to merge dynamic fields when `spath` command is executed (because `spath` command needs to append instead of replace). 
MAP_CONAT is different because it will simply overwrite value in case of key conflict. This is needed mainly for join operation where we don't merge values.

Issue #4112, opened 2025-08-22
[FEATURE] Support dynamic columns in PPL


penghuo:
**Is your feature request related to a problem?**
Certain PPL operators naturally require **data-driven (dynamic) columns** whose names and counts are not known at plan time. It is the blocker to support timechart and spath command

### Example — `timechart` pivot
```ppl
source=logs
| timechart span=1h count() by status limit=3
```
**Problem:** The set of `status` labels (e.g., `ERROR`, `OK`, `TIMEOUT`) is only known at runtime and can vary by dataset. Producing a wide, pivoted table requires **dynamic columns**, which are not available under a fixed per-operator schema.

### Example — `spath` with object keys
```ppl
source=idx
| spath input=doc path=event.tags
```
**Problem:** `event.tags` is a JSON object whose keys differ across documents. The extracted keys are unknown until read, so the output columns are inherently dynamic.

PR #4489, opened 2025-10-10
Add JSON_EXTRACT_ALL internal function for Calcite PPL

ykmr1224:
### Description
- Add JSON_EXTRACT_ALL internal function for Calcite PPL
  - This function will be used by `spath` command without `path` param implementation.
  - Extracts all the values from input JSON string, and store them into MAP with JSON path as the key.
  - Example: input=`{"a": 1, "b": [2, 3, 4], "c": [{"e": 5}, {"e": 6}, {"f": 7}]}` output=`Map("a"=> 1, "b{}"=> [2,3,4], "c{}.e"=> [5, 6], "c{}.f"=> 7)`
    - `{}` is a special notation to indicate array in the path.
    - A value could be array or single value regardless the key ends with or contains `{}`. (see `c{}.e` and `c{}.f` in above example)
  - Please check unit tests for detailed spec.
- We might extend it to allow specifying paths to collect only needed values for optimization.

### Related Issues
- https://github.com/opensearch-project/sql/issues/4112

### Check List

### DIFF
--- a/core/src/test/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImplTest.java
+++ b/core/src/test/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImplTest.java
@@ -12,7 +12,6 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 import org.junit.jupiter.api.Test;
 
 public class JsonExtractAllFunctionImplTest {
@@ -52,16 +51,7 @@ public class JsonExtractAllFunctionImplTest {
 
   private Map<String, Object> eval(String json) {
     Object result = JsonExtractAllFunctionImpl.eval(json);
-    return debug(assertValidMapResult(result));
-  }
-
-  private Map<String, Object> debug(Map<String, Object> map) {
-    System.out.println(
-        "map: "
-            + map.entrySet().stream()
-                .map(entry -> entry.getKey() + "=" + entry.getValue())
-                .collect(Collectors.joining(", ")));
-    return map;
+    return assertValidMapResult(result);
   }
 
   @Test


core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        return parser.getValueAsString();
+    }
+  }
+
+  private static Object getIntOrLong(JsonParser parser) throws IOException {
+    try {
+      return parser.getIntValue();
+    } catch (Exception e) {
+      return parser.getLongValue();
+    }
```
penghuo: Avoid exception, User parser.getNumberType?

core/src/test/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImplTest.java
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  private Map<String, Object> debug(Map<String, Object> map) {
+    System.out.println(
+        "map: "
+            + map.entrySet().stream()
+                .map(entry -> entry.getKey() + "=" + entry.getValue())
+                .collect(Collectors.joining(", ")));
+    return map;
+  }
```
penghuo: remove debug

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
ykmr1224: Thanks, fixed!

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
```diff
@@ -0,0 +1,218 @@
...
+
+  @SuppressWarnings("unchecked")
+  private static void appendValue(Map<String, Object> resultMap, String path, Object value) {
+    Object existingValue = resultMap.get(path);
+    if (existingValue == null) {
+      resultMap.put(path, value);
+    } else if (existingValue instanceof List) {
+      ((List<Object>) existingValue).add(value);
+    } else {
+      resultMap.put(path, list(existingValue, value));
```
dai-chen, line 158: Is this list copy intentional?

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  private static Map<String, Object> parseJson(String jsonStr) {
+    Map<String, Object> resultMap = new HashMap<>();
+    Stack<String> pathStack = new Stack<>();
+
+    try (JsonParser parser = JSON_FACTORY.createParser(jsonStr)) {
+      JsonToken token;
+
+      while ((token = parser.nextToken()) != null) {
```
dai-chen, line 95: Do we need to limit the depth of stack?

ykmr1224: This does not use recursion and I don't see much risk to allow any depth. (It would raise exception when it fails to allocate memory, though)

Let me know if you see some specific risk.

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
ykmr1224: The `else` clause is for when existingValue is not a list, and will create new list in that case.

Issue #4430, opened 2025-10-02
[BUG] PPL Inner Join crash OpenSearch with OOM


penghuo:
## Query Information
**PPL Command/Query:**
```ppl
POST {{baseUrl}}/_plugins/_ppl/
Content-Type: application/x-ndjson

{
    "query": "source=sso_logs-nginx-prod-2024.05.08 | where ISPRESENT(code) | join type=inner left=SUCCESS right=FAILED on SUCCESS.remote = FAILED.ClientIP [ source=sso_logs-nginx-prod-2024.05.08 | where ISPRESENT(log) | parse log '.+client: (?<ClientIP>.+?),.*'] | dedup ClientIP, agent"
}
```
**Actual Result:**
```
OpenSearch crashed with OOM

=== Standard error of node `node{:opensearch-sql-plugin:integTest-0}` ===
»   ↓ last 40 non error or warning messages from /Users/penghuo/oss/os-sql/plugin/build/testclusters/integTest-0/logs/opensearch.stderr.log ↓
» WARNING: Using incubator modules: jdk.incubator.vector
»  WARNING: Unknown module: org.apache.arrow.memory.core specified to --add-opens
»  fatal error in thread [opensearch[integTest-0][scheduler][T#1]], exiting
»  java.lang.OutOfMemoryError: Java heap space
»       at org.opensearch.common.util.concurrent.ThreadContext.preserveContext(ThreadContext.java:558)
»       at org.opensearch.threadpool.ThreadPool.schedule(ThreadPool.java:602)
»       at org.opensearch.threadpool.Scheduler$ReschedulingRunnable.onAfter(Scheduler.java:266)
»       at org.opensearch.common.util.concurrent.ThreadContext$ContextPreservingAbstractRunnable.onAfter(ThreadContext.java:953)
»       at org.opensearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:56)
»       at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
»       at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
»       at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
»       at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
[...]

PR #4501, opened 2025-10-10
Add configurable sytem limitations for `subsearch` and `join` command

### LINKED ISSUE #3731: [ENHANCEMENT] Set operator limitation for high cost operators

LantaoJin:
### Description
Add two configurable limitations for PPL.
1. plugins.ppl.subsearch.maxout (default value 10000, similar to `maxout` in [subsearch], [ref](https://help.splunk.com/en/splunk-enterprise/administer/admin-manual/10.0/configuration-file-reference/10.0.1-configuration-file-reference/limits.conf#subsearch-0))
2. plugins.ppl.join.subsearch_maxout (default value 50000, similar to `subsearch_maxout` in [join], [ref](https://help.splunk.com/en/splunk-enterprise/administer/admin-manual/10.0/configuration-file-reference/10.0.1-configuration-file-reference/limits.conf#join-0))

### Related Issues
Resolves #3731 and #4430

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
```diff
@@ -205,7 +205,8 @@ public void execute(
                     () -> {
                       try (PreparedStatement statement = OpenSearchRelRunners.run(context, rel)) {
                         ResultSet result = statement.executeQuery();
-                        buildResultSet(result, rel.getRowType(), context.querySizeLimit, listener);
+                        buildResultSet(
+                            result, rel.getRowType(), context.sysLimit.querySizeLimit(), listener);
```
penghuo, line 209: nit, revert unncessary format change.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1136,6 +1150,15 @@ private Optional<RexLiteral> extractAliasLiteral(RexNode node) {
   public RelNode visitJoin(Join node, CalcitePlanContext context) {
     List<UnresolvedPlan> children = node.getChildren();
     children.forEach(c -> analyze(c, context));
+    // add join.subsearch_maxout limit to subsearch side
+    if (context.sysLimit.joinSubsearchLimit() >= 0) {
+      replaceTop(
```
penghuo: nit, is it possible to avoid access private method?

2cents, Add a frame in CalcitePlanContext, frame is boundary of subsearch, and define limit on frame. When visit subsearch, append LogicalSystemLimit to subsearch on each frame.

docs/user/ppl/admin/settings.rst
```diff
@@ -265,3 +308,95 @@ PPL query::
...
+Rollback to default value::
+
+    sh$ curl -sS -H 'Content-Type: application/json' \
+    ... -X PUT localhost:9200/_plugins/_query/settings \
+    ... -d '{"persistent" : {"plugins.ppl.subsearch.maxout" : null}}'
+    {
+      "acknowledged": true,
+      "persistent": {},
+      "transient": {}
+    }
```
penghuo: Remove Rollback, https://github.com/opensearch-project/sql/pull/4449 fix it.

docs/user/ppl/admin/settings.rst
```diff
@@ -265,3 +308,95 @@ PPL query::
...
     }
+
+
+plugins.ppl.subsearch.maxout
+============================
+
+Description
+-----------
+
+The size configures the maximum of rows to return from subsearch. The default value is: ``10000``. A value of ``-1`` indicates that the restriction is unlimited.
```
penghuo, line 293: What if set to 0? Join/Subquery will be optimzied by Calcite?

LantaoJin: For now, I followed the standard behaviour in database:
```
select * from t_outer where exists (select 1 from t_inner where t_outer.id = t_inner.id limit 0);
select * from t_outer where id in (select id from t_inner limit 0);
select * from t_outer where id = (select id from t_inner limit 0);
select * from t_outer where id = (select count(*) from t_inner limit 0);
```
All above queries return empty in SQL (postgresql).

The implementation is here https://github.com/opensearch-project/sql/pull/4501/files#diff-e5198d773af75bf3173ef25676a2803a0091cb51e32d6ae30241273519d30261R601-R605

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: > nit, is it possible to avoid access private method?

I don't think so.

When visit the subsearch side (right in join for example), the right plan was pushed to stack.
```
public RelNode analyze(UnresolvedPlan unresolved, CalcitePlanContext context) {
    return unresolved.accept(this, context);
  }
```
`RelBuilder.pop()` is private either. So we don't have a way to replace it.

Here was my previous try code for join
```
  public RelNode visitJoin(Join node, CalcitePlanContext context) {
    // visit the main side
    analyze(node.getLeft(), context);
    if (context.sysLimit.joinSubsearchLimit() >= 0) {
      // add join.subsearch_maxout limit to subsearch side
      RelNode withLimit = context.relBuilder.with(
          analyze(node.getRight(), context),
          r -> LogicalSystemLimit.create(
            SystemLimitType.JOIN_SUBSEARCH_MAXOUT,
            r.peek(),
            r.literal(context.sysLimit.joinSubsearchLimit())));
      context.relBuilder.push(withLimit); // push the new subsearch plan
    } else {
      // visit the subsearch side
      analyze(node.getRight(), context);
    }
```

The code use `relBuilder.with()`, but the first parameter `analyze(node.getRight(), context)` will push the subsearch to stack, and the `with()` method push it twice. 
```
  /** Evaluates an expression with a relational expression temporarily on the
   * stack. */
  public <E> E with(RelNode r, Function<RelBuilder, E> fn) {
    try {
      push(r);
      return fn.apply(this);
    } finally {
      stack.pop();
    }
  }
```
1. push left plan by `analyze(node.getLeft(), context)`, stack size is 1
2. push right plan by the first parameter of `with`(`analyze(node.getRight(), context)`), stack size is 2
3. push duplicated right plan by `push` in `with`, stack size is 3
4. pop duplicated right plan by `pop` in `with`, stack size is 2
5. push new right plan by `context.relBuilder.push(withLimit)`, stack size is 3 (incorrect)

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
LantaoJin: Didn't get the point. This change is necessary, from `context.querySizeLimit` to `context.sysLimit.querySizeLimit()`.

docs/user/ppl/admin/settings.rst
LantaoJin: done

docs/user/ppl/admin/settings.rst
LantaoJin: What's your thoughts, set both `0` and negative value to unlimited? @penghuo

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
qianheng-aws: Does it work by using `relbuilder.build() + relbuilder.push(newTop)`? `relbuilder.build()` will do pop while public.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -496,21 +505,104 @@ public RexNode visitExistsSubquery(ExistsSubquery node, CalcitePlanContext conte
...
+                        if (newFilter != filter) {
+                          return newFilter;
+                        }
+                        return visit((RelNode) filter);
+                      }
+
+                      @Override
+                      public RelNode visit(RelNode other) {
+                        RelNode newInput =
+                            other.getInputs().isEmpty() ? null : other.getInput(0).accept(this);
```
qianheng-aws: [question]Will there be case that there is join or union in subsearch? In those case there will be more than 1 input for the specific operators? If so, the current code will construct incorrect plan.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -496,21 +505,104 @@ public RexNode visitExistsSubquery(ExistsSubquery node, CalcitePlanContext conte
...
+                        if (newInput == null || newInput == other.getInput(0)) {
+                          return other;
+                        }
+                        return other.copy(other.getTraitSet(), Collections.singletonList(newInput));
+                      }
+                    });
+        planVisitor.replaceTop(context.relBuilder, replacement);
+      }
+      if (subqueryExpression instanceof InSubquery) {
```
qianheng-aws: SQL support correlate condition for in subquery or scalar subquery. So Calcite should support them as well.
e.g.
```
SELECT * FROM EMPLOYEE WHERE location in (select location from DEPART where EMPLOYEE.dept = DEPART.name) limit 1
```

If there is correlate condition for in or scalar subsearch, shall we do similar operation like above?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -99,8 +99,8 @@ public void testJoinWithFieldList() throws IOException {
         "source=opensearch-sql_test_index_bank | join type=outer account_number"
             + " opensearch-sql_test_index_bank";
```
yuancu, line 104: Should there also be explaining tests for subsearch?

opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
```diff
@@ -125,6 +125,22 @@ public class OpenSearchSettings extends Settings {
...
+          10000,
+          -1,
+          Setting.Property.NodeScope,
+          Setting.Property.Dynamic);
+
+  public static final Setting<?> PPL_JOIN_SUBSEARCH_MAXOUT_SETTING =
+      Setting.intSetting(
+          Key.PPL_JOIN_SUBSEARCH_MAXOUT.getKeyValue(),
+          50000,
           0,
```
yuancu: Should the minimum be set to `-1` instead?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
LantaoJin: added

opensearch/src/main/java/org/opensearch/sql/opensearch/setting/OpenSearchSettings.java
LantaoJin: fixed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: Sync offline. We still cannot use `relbuilder.build() + relbuilder.push(newTop)` since it will empty the `fields of Frame`.
```
  private void replaceTop(RelNode node) {
    final Frame frame = stack.pop();
    stack.push(new Frame(node, frame.fields));    // <--- frame.fields will be kept all the time
  }
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
LantaoJin: Fixed in latest commit. For BiRel or SetOp, just return.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
LantaoJin: Added the same logic for correlated in-subquery. For correlated scalar-subquery, since there is always an aggregation will be perform in subquery, sysLimit is not necessary.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
penghuo: How does SQL Join translate to RelNode? It use the private method?

docs/user/ppl/admin/settings.rst
penghuo: discussed offline, 0 and -1 means unlimited.

LantaoJin: > discussed offline, 0 and -1 means unlimited.

sure, let me update the code and doc to
`0` means unlimited, and minValue=0 in Settings

PR #4511, opened 2025-10-10
Add MAP_REMOVE internal function for Calcite PPL

ykmr1224:
### Description
- Add internal MAP_REMOVE function for Calcite PPL
  - This function takes map and multiple keys to eliminate keys in single call.
  - This will be used for dynamic fields implementation, where we need to eliminate duplicate fields from `_MAP` to keep static/dynamic fields consistent.

### Related Issues
- https://github.com/opensearch-project/sql/issues/4112

### Check List


penghuo:
could we marked field as removed instead of really remove it from map? we can evaluate it during perf test stage.

ykmr1224:
> could we marked field as removed instead of really remove it from map? we can evaluate it during perf test stage.

It should be possible, but it would complicate the map related operations, since we need several operations to consider the removed fields.

penghuo (approved):

PR #4512, opened 2025-10-10
Fixes for `Multisearch` and `Append` command

ahkcs:
### Description
1. Type Conflict Handling 
  - Changed behavior: Type conflicts now throw IllegalArgumentException instead of auto-renaming fields
  - Modified: SchemaUnifier.java - Removed automatic field renaming logic (e.g., age → age0)
  - Updated Tests:
    - CalciteMultisearchCommandIT.testMultisearchWithDirectTypeConflict - Now expects exception
    - CalcitePPLAppendCommandIT.testAppendWithConflictTypeColumn - Now expects exception
  - Documentation: Updated multisearch.rst and append.rst to reflect new behavior and add Limitations section

  2. Timestamp Interleaving
  - Modified: CalciteRelNodeVisitor.findTimestampField() - Now only detects @timestamp field for timestamp interleaving
  - Other timestamp fields (_time, timestamp, time) are no longer used for interleaving

  3. Documentation Fixes
  - multisearch.rst:
    - Added Limitations section
    - Removed redundant Example 4 (Handling Empty Results)
    - Removed Example 6 (type-conflict resolution)
  - append.rst:
    - Added Limitations section
    - Removed Example 3 (type conflict example no longer valid)

### Related PRs

#4332 #4123

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/SchemaUnifier.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/SchemaUnifier.java
@@ -82,8 +82,8 @@ public class SchemaUnifier {
           // Same field name but different type - throw exception
           throw new IllegalArgumentException(
               String.format(
-                  "Schema unification failed: field '%s' has conflicting types across subsearches",
-                  fieldName));
+                  "Unable to process column '%s' due to incompatible types: '%s' and '%s'",
+                  fieldName, existingType.getSqlTypeName(), fieldType.getSqlTypeName()));
         }
         // If we've seen this exact (name, type) combination, skip it
       }
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
@@ -316,7 +316,7 @@ public class CalciteMultisearchCommandIT extends PPLIntegTestCase {
         "Error message should indicate type conflict",
         exception
             .getMessage()
-            .contains("Schema unification failed: field 'age' has conflicting types"));
+            .contains("Unable to process column 'age' due to incompatible types:"));
   }
 
   @Test
@@ -362,6 +362,6 @@ public class CalciteMultisearchCommandIT extends PPLIntegTestCase {
         "Error message should indicate type conflict",
         exception
             .getMessage()
-            .contains("Schema unification failed: field 'age' has conflicting types"));
+            .contains("Unable to process column 'age' due to incompatible types:"));
   }
 }
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAppendCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAppendCommandIT.java
@@ -234,7 +234,7 @@ public class CalcitePPLAppendCommandIT extends PPLIntegTestCase {
         "Error message should indicate type conflict",
         exception
             .getMessage()
-            .contains("Schema unification failed: field 'sum' has conflicting types"));
+            .contains("Unable to process column 'sum' due to incompatible types:"));
   }
 
   @Test
--- a/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAppendTest.java
+++ b/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAppendTest.java
@@ -234,6 +234,6 @@ public class CalcitePPLAppendTest extends CalcitePPLAbstractTest {
     Exception exception =
         Assert.assertThrows(IllegalArgumentException.class, () -> getRelNode(ppl));
     verifyErrorMessageContains(
-        exception, "Schema unification failed: field 'DEPTNO' has conflicting types");
+        exception, "Unable to process column 'DEPTNO' due to incompatible types:");
   }
 }


core/src/main/java/org/opensearch/sql/calcite/SchemaUnifier.java
ykmr1224: I think we should rather raise error from here. (Then we don't need to check again)
Later we would implement a logic to generalize types for the same field name here.

ahkcs: Thanks for the suggestion! I think it makes sense, moved the location to raise error

core/src/main/java/org/opensearch/sql/calcite/SchemaUnifier.java
```diff
@@ -55,47 +53,37 @@ public static List<RelNode> buildUnifiedSchemaWithConflictResolution(
...
       for (RelDataTypeField field : node.getRowType().getFieldList()) {
         String fieldName = field.getName();
         RelDataType fieldType = field.getType();
 
-        // Track which (name, type) combinations we've seen
-        Set<RelDataType> typesForName = seenFields.computeIfAbsent(fieldName, k -> new HashSet<>());
-
-        if (!typesForName.contains(fieldType)) {
-          // New field or same name with different type - add to schema
+        RelDataType existingType = seenFields.get(fieldName);
```
songkant-aws, line 76: We recently find another issue of type conflicts here. RelDataType evaluates the hash equality by its digested string as well. For example, "INTEGER" is not equal to "INTEGER NOT NULL". A quick fix would be aligning the same SqlType to be nullable. Ideally it won't affect the data type resolution while execution. cc @xinyual

songkant-aws: We can consider to allow same SqlTypeName but with different nullability to be merged here.

docs/user/ppl/cmd/append.rst
```diff
@@ -24,6 +24,11 @@ append <sub-search>
 
 * sub-search: mandatory. Executes PPL commands as a secondary search.
 
+Limitations
+===========
+
+* **Schema Compatibility**: When fields with the same name exist between the main search and sub-search but have incompatible types, the query will fail with an error. To avoid type conflicts, ensure that fields with the same name have the same data type, or use different field names (e.g., by renaming with ``eval`` or using ``fields`` to select non-conflicting columns).
```
songkant-aws, line 30: Understand the intention here. Strong schema engine like SQL restricts the type to be the same. Some weak schema engine resolves types at runtime and doesn't care the data type. I think it's not easy to make it compatible.

Not sure what's better user experience and customer expectation here. Does user accept this behavior or expect to union anyway? cc @LantaoJin

ahkcs: We are planning to enable permissive mode in the future: #4349 to support schema merging with type conflicts, in order to avoid breaking changes in the future, we are marking this as a limitation now instead of using a workaround. cc @penghuo

ykmr1224 (dismissed):

core/src/main/java/org/opensearch/sql/calcite/SchemaUnifier.java
ahkcs: Thanks for the suggestion! I have updated the implementation to allow same SqlTypeName but with different nullability to be merged

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1809,18 +1809,16 @@ public RelNode visitMultisearch(Multisearch node, CalcitePlanContext context) {
...
 
   /**
-   * Finds the timestamp field for multisearch ordering.
+   * Finds the @timestamp field for multisearch ordering. Only @timestamp field is used for
+   * timestamp interleaving. Other timestamp-like fields are ignored.
    *
-   * @param rowType The row type to search for timestamp fields
-   * @return The name of the timestamp field, or null if not found
+   * @param rowType The row type to search for @timestamp field
+   * @return "@timestamp" if the field exists, or null if not found
```
Swiddis, line 1816: issue (non-blocking): Should we think about indices with different timestamp field names?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
```diff
@@ -352,18 +345,23 @@ public void testMultisearchCrossIndexFieldSelection() throws IOException {
...
+                            + "[search source=%s | fields age] "
+                            + "| stats count() as total",
+                        TEST_INDEX_ACCOUNT, TEST_INDEX_LOCATIONS_TYPE_CONFLICT)));
 
-    verifyDataRows(result, rows(1010L));
+    assertTrue(
+        "Error message should indicate type conflict",
+        exception
+            .getMessage()
+            .contains("Schema unification failed: field 'age' has conflicting types"));
```
Swiddis: suggestion: This error message isn't very clear

Users might not know what schema unification means, and we don't say the nature of the conflict. I'd like to see something more like:

```
Unable to process column 'age' due to incompatible types: 'integer' and 'text'.
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Currently it is set as a limitation: we want to only support @timestamp

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
ahkcs: Updated error message

docs/user/ppl/cmd/append.rst
songkant-aws: Got it. Thanks for the change.

core/src/main/java/org/opensearch/sql/calcite/SchemaUnifier.java
```diff
@@ -104,6 +92,10 @@ private static List<SchemaField> buildUnifiedSchema(List<RelNode> nodes) {
     return schema;
   }
 
+  private static boolean areTypesCompatible(RelDataType type1, RelDataType type2) {
```
songkant-aws, line 95: There is another concern of using this method to allow type merge. If index A's RelDataType 'INTEGER NOT NULL' is put to unified schema, index B's same name RelDataType 'INTEGER' will be merged silently. Index B's column values could contain NULL values. 

The generated code could ignore the null check because the merged unified schema has 'iNTEGER NOT NULL'. It will probably throw NPE when merging index B's NULL values. We could write some query to double check if we can reproduce this scenario.

songkant-aws: I tried some queries in my local test. Haven't seen such NPE errors yet. Not sure if there is edge case. But for now we can leave it there until needed fix.

PR #4560, opened 2025-10-14
Extract ProjectHandler from CalciteRelNodeVisitor

ykmr1224:
### Description
- Extract ProjectHandler from CalciteRelNodeVisitor
- This is pure refactoring, and no logic change.
- I will raise some more PRs to reduce the size of CalciteRelNodeVisitor

### Related Issues
n/a

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -39,8 +39,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.apache.calcite.plan.RelOptTable;
-import org.apache.calcite.plan.ViewExpanders;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.Aggregate;
 import org.apache.calcite.rel.core.JoinRelType;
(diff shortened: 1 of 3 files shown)


LantaoJin:
> I will raise some more PRs to reduce the size of CalciteRelNodeVisitor

LGTM for Project. But I don't think the the size of `CalciteRelNodeVisitor` is a problem. Some methods can be shared in `CalciteRelNodeVisitor`. No need to create one handler per visit(). And small method sometimes is bad smell code. For example:
```
    if (isSingleAllFieldsProject(node)) {
      return handleAllFieldsProject(node, context);
    }
```
is confused. 

```
    if ( node.getProjectList().size() == 1
        && node.getProjectList().getFirst() instanceof AllFields ) {
      return handleAllFieldsProject(node, context);
    }
```
looks better.

ykmr1224:
> > I will raise some more PRs to reduce the size of CalciteRelNodeVisitor
> 
> LGTM for Project. But I don't think the the size of `CalciteRelNodeVisitor` is a problem. Some methods can be shared in `CalciteRelNodeVisitor`. No need to create one handler per visit(). And small method sometimes is bad smell code. For example:

I am not planning to extract all visitor to separate handlers. Just a few of them are huge and better extracted.
My main motivation is to control the context size for AI agent. It tend to fill up the context when touching CalciteRelNodeVisitor. Smaller context would improve AI agent's quality, even when it won't fill the context.

> And small method sometimes is bad smell code. 

`isSingleAllFieldsProject` is from before this change. And I think this kind of extraction is normally good one for readability.

penghuo:
What is the target state of refactor? Does it simplify how to compose relNode? Could u give an example?

PR #4378, opened 2025-09-25
Change the default search sort tiebreaker to `_shard_doc` for PIT search

LantaoJin:
### Description
Change the default search sort tiebreaker of PIT search requests to `_shard_doc` after https://github.com/opensearch-project/OpenSearch/pull/18924 being merged.

Main changes:
1. Remove sort field `_doc` in pushDownFilter(). (`_doc` is a default hits ordering if no specific orders are set)
2. Explicitly add `_shard_doc` as sort tiebreaker of PIT search (~~not 100% sure the current implementation of `_shard_doc` is an implicit sort in PIT search, see~~ [question](https://github.com/opensearch-project/OpenSearch/pull/18924#issuecomment-3342365950)).
3. Keep the legacy engine unchanged.

### Related Issues
Resolves #3064 

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4569, opened 2025-10-15
Use `_doc` + `_shard_doc` as sort tiebreaker to get better performance

LantaoJin:
### Description
Before https://github.com/opensearch-project/sql/pull/4378, the sort in PIT search is
case 1: if no sort field specified, sort by `_doc` + `_id` (+ means "then"). (❎ could cause high memory issue)
case 2: if sort fields specified, sort by `fields`. (❎ paged results could miss or duplicate hits)
case 3: if sort fields specified and query contains a filter, sort by `_doc`. (❎ paged results could miss or duplicate hits)

https://github.com/opensearch-project/sql/pull/4378 added the `_shard_doc` as sort tiebreaker with
case 1: if no sort field specified, sort by `_shard_doc`. (❎ performance regression)
case 2: if sort fields specified, sort by `fields` + `_shard_doc`.(❎ lower performance on low cardinality field)

https://github.com/opensearch-project/sql/pull/4435 found performance regression in case 1 and partially revert the changes to
case 1: if no sort field specified, sort by `_doc` + `_id`. (❎ could cause high memory issue)
case 2: if sort fields specified, sort by `fields`. (❎ paged results could miss or duplicate hits)

After this PR, we change the sort in PIT search to
case 1: if no sort field specified, sort by `_doc` + `_shard_doc`. ✅
case 2: if sort fields specified, sort by `fields` + `_doc` + `_shard_doc`.✅

**RCA of performance regression**:
`_shard_doc` is not a stored field in index which will be generated in runtime when comparison. Computing `_shard_doc` per document is a high cost operation. But sorting by `_doc` then `_shard_doc` only generates `_shard_doc` when the `_doc` values are conflicted.
Even in the case of user specified sort fields, we should sort by `fields` then `_doc` then `_shard_doc` to reduce the computing of `_shard_doc`. For example, if the sort field is a low cardinality field, e.g. `gender`, sorting by `gender` then `_doc` then `_shard_doc` generates `_shard_doc` for comparison only if values of `gender` and `_doc` are both conflicted.

**This PR is no needed to backport to 2.19-dev since `shard_doc` feature is only available since OS 3.3.0**

### Related Issues

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
@@ -30,6 +30,7 @@ import org.opensearch.search.SearchHits;
 import org.opensearch.search.SearchModule;
 import org.opensearch.search.builder.PointInTimeBuilder;
 import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.sort.FieldSortBuilder;
 import org.opensearch.search.sort.ShardDocSortBuilder;
 import org.opensearch.search.sort.SortBuilders;
 import org.opensearch.sql.opensearch.data.value.OpenSearchExprValueFactory;
@@ -52,7 +53,7 @@ public class OpenSearchQueryRequest implements OpenSearchRequest {
   private final IndexName indexName;
 
   /** Search request source builder. */
-  private SearchSourceBuilder sourceBuilder;
+  private final SearchSourceBuilder sourceBuilder;
 
   /** OpenSearchExprValueFactory. */
   @EqualsAndHashCode.Exclude @ToString.Exclude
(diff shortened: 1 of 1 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
```diff
@@ -203,12 +205,23 @@ public OpenSearchResponse searchWithPIT(Function<SearchRequest, SearchResponse>
...
+      // Add sort tiebreaker for PIT search.
+      // We cannot remove it since `_shard_doc` is not added implicitly in PIT now.
+      // Ref https://github.com/opensearch-project/OpenSearch/pull/18924#issuecomment-3342365950
+      if (this.sourceBuilder.sorts() == null || this.sourceBuilder.sorts().isEmpty()) {
+        // If no sort field specified, sort by `_doc` + `_shard_doc`to get better performance
         this.sourceBuilder.sort(DOC_FIELD_NAME, ASC);
-        // Workaround to preserve sort location more exactly,
-        // see https://github.com/opensearch-project/sql/pull/3061
-        this.sourceBuilder.sort(METADATA_FIELD_ID, ASC);
+        this.sourceBuilder.sort(SortBuilders.shardDocSort());
```
Swiddis, line 214: Does it matter if we duplicate fields in the sorting list? We could simplify/remove the below `else` logic by just always appending this, I would expect Lucene to optimize it in the background but I haven't measured it.

Swiddis (approved):

anasalkouz:
Can you share the performance benchmark for the 3 approaches?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
LantaoJin: not sure will Lucene optimize duplicated `fields` or `_doc` in sorting, but for sure the duplicated `_shard_doc` is not allowed in OpenSearch Core. It is no harmful for restricted checker here.

LantaoJin:
> Can you share the performance benchmark for the 3 approaches?

I haven't run the benchmark, the RCA was made by reading the code of Luence and OS `shard_doc` feature.

The performance of `_doc` then `_shard_doc` is same as `_doc` then `_id`, provided by @ahkcs on Oct 1st. (the case 1)

For case 2, the current `fields` + `_doc` + `_shard_doc` is an further optimization upon `fields` + `_shard_doc` based on above benchmark result with inference.

Will rerun some benchmark to double confirm.

PR #4579, opened 2025-10-16
Support `chart` command in PPL

### LINKED ISSUE #399: [RFC] PPL Chart Command
## Problem Statements
In observability, user usually try to ask TopN question to find root cause. For example, user want to find out, on last two week, top 5 high CPU utilization host and visualize CPU utilization of hosts on each day. Currently, PPL does not support this feature in different ways
* No single command, using `stats max(cpu) maxcpu by host | sort 5 maxcpu ` could return top5 host which have high cpu utilization. If user want to visualize each day's max cpu utilization of these 5 hosts, user must compose another query, e.g. `where host in (h1, h2, h3, h4, h5) | status max(cpu) maxcpu by span(timestamp) host`. 
* Response Format, PPL current response is row format, but visualization library prefer series data. e.g. Using series data, visualization library could use first column as x-axis, and other columns as series. 
```
# row format
day  host  cpu
1     h1    30
1     h2    20
2     h1    40
3     h2    30

# series format
day  h1    h2
1    30    20
2    40    30
```

#### Overview

Create a chart with a corresponding table of statistics.

#### Syntax

```
chart <aggregation> OVER <row-split> [BY <column-split> [limit=(top|bottom)<int>]] 
```

* aggregation: A statistical aggregation function.
* limit: Specifies a limit for the number of distinct values of column-split field to return, Default=top 5. 
* row-split: The field that you specify becomes the first column in the results table. The field values become the row labels in the results table. In a chart, the field name is used to label the X-axis. The field values become the X-axis values.
* column-split: Specifies a field to use as the columns in the result table. By default, when the result are visualized, the columns become the data series in the chart.

#### Example

* Example 1. For each response status, visualize top2 host’s request count. Top2 means, in a period, Top2 host has most request count.

>Note: One aggregation, one row-split field, one column-split field. 

```
chart sum(req) OVER status BY host limit top=2

# sample response
status host1 host2
200    100   200
404    10    20
503    1     2
```

* Example 2. For each day, visualize top2 host’s cpu utilization. Top2 means, in a period, Top2 host has highest cpu utilization.

>Note: One aggregation, one span row-split field, one column-split field. 

```
chart max(cpu) OVER span(timestamp, 1d) BY host limit top=2

# sample response
timestamp host3 host4
Day1      30    20
Day2      20    70
Day3      50    50
```

* Example 3. For each day, visualize top2 host cpu utilization and request count. Top2 means, in a period, Top2 host has highest cpu utilization.

>Note: Two aggregations, one span row-split field, one column-split field. 

```
chart max(cpu), sum(req) OVER span(timestamp, 1d) BY host limit top=2

# sample response
max(cpu)
timestamp host3 host4
Day1      30    20
Day2      20    70
Day3      50    50

sum(req)
timestamp host3 host4
Day1      200   100
Day2      100   900
Day3      800   500
```

* Example 4. For each response status, total request count

```
chart sum(req) OVER status

# sample response
status sum
200    300
404    30
503    3
```

yuancu:
### Description

The chart command returns an aggregation result that can be easily pivoted to a two-dimension table format.

**Work items**
- [x] support span
- [x] support `limit`, `limit=top x`, `limit=bottom x`
- [x] support `useother`, `otherstr`
- [x] correct limit behavior with non-accumulative aggregation functions (min, max, avg, etc) // fixed in #4594
- [x] support `usenull`, `nullstr`
- [x] support non-string fields as column split
- [x] add integration tests
- [x] add explain tests
- [x] add a doc
- [x] Add a brief walk-through of the implementation
- [x] Anonymizer & test

**Difference between stats, timechart, and chart**

- with `stats`
    `chart` is conceptually similar to `stats` in that they all compute an aggregation value and then group them by a given criterion. The main differences lies in their output format.
    
    For example, for the query in introduction, we can rewrite it with stats: `... | stats count BY status, host`. It gives the following result:
    
    status | host | count
    -- | -- | --
    200 | www1 | 11835
    200 | www2 | 11186
    200 | www3 | 11261
    400 | www1 | 233
    400 | www2 | 257
    400 | www3 | 211
    403 | www2 | 228
    404 | www1 | 244
    404 | www2 | 209
    404 | www3 | 237

    Each field specified in BY clause becomes a separate column in the result table; each row is a unique combination of them. Whereas in `chart`, each unique status becomes a row; each individual value of host becomes a column. This format makes it easier to view and visualize the results.  
  
    Similar to the syntax of `stats`, an equivalent expression of `... | CHART count OVER status BY host` is `... | CHART count BY status, host`.

  - It is worth noticing that `chart` will ignore documents with NULL in row split and in aggregation results, while `stats` will keep them.

- with `timechart`

    The key difference between `chart` and `timechart` is that `timechart` leverages a default `@timestamp` field as the field to perform aggregation on. `... | timechart agg BY field` is conceptually equivalent to `... | CHART agg OVER _time BY field`  
      
    The following table summarizes the differences between the three commands.  
    
    |Feature|chart|stats|timechart|
    |---|---|---|---|
    |BY clause fields|Limited to two (row-split, column-split)|Multiple (3+ possible)|Always uses _time + one optional field|
    |Primary purpose|Consolidated visualizations|Detailed statistical calculations|Time-based analysis|
    |Output format|Table format optimized for visualization (Not implemented) |Row-based results|Table format time series visualization (Not implemented)|
    |Best use case|Data comparisons across categories|Detailed data analysis|Trend analysis over time|
    |X-axis control|Any field|N/A|Always _time|
    
    
### Related Issues
Resolves #399

### Implementation Walk-through

Ideally, chart should pivot the result into a 2-dimension table. E.g. for the following table:

|  a |  b |  val |
|---|---|---|
|  m | x  | 3  |
|  m |  y |  4 | 

`| chart avg(val) by a, b` should make it a table like this:

|  a |  x |  y |
|---|---|---|
|  m | 3  | 4  |

However, it seems dynamic pivoting is not supported in SQL/Calcite (see original discussion in https://github.com/opensearch-project/sql/issues/3965#issue-3281969676). Therefore, the result table for the implemented`chart` is like:

|  a |  b |  avg(val) |
|---|---|---|
|  m | x  | 3  |
|  m |  y |  4 | 

The pivoting can be performed in the front-end.

The above operation is equivalent to `stats avg(val) by a, b` -- this is the case when parameters like `usenull`, `useother`, and `limit` is not involved in the result.

When these parameters are involved, `chart` command will find the top-N categories of `b`, aggregating the rest to an `OTHER` category, and aggregating those whose `b` is null to a "NULL" category. This leads to the following implementation:

1. normal aggregation based on a, b (equivalent to `stats agg_func by a, b`)
2. find out the top-N categories (unique values of column b) by aggregating on the above aggregation results
    1. aggregate on b
    2. sort on aggregation results
    3. number the rows
3. left join the ranked results with the original aggregation
4. keep rows whose row number is no greater than the limit, categorizing the rest to OTHER or NULL
5. Aggregate again because values categorized into OTHER or NULL need to be merged

**Note**:

This implementation did not reuse the implementation of timechart to circumvent some existing bugs. A following PR will merge their implementation as chart essentially is a superset of timechart in terms of functionality.

**Future work items**
- [ ] support multiple aggregation functions (Left as a TODO in the future: the output will be messy when multiple aggregations are involved because the results are not pivoted.)
- [ ] unify implementation of `timechart` and `chart`
- [ ] support more bin options like `bins` (after #4612 )

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -2415,9 +2415,9 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
   public RelNode visitChart(Chart node, CalcitePlanContext context) {
     visitChildren(node, context);
     ArgumentMap argMap = ArgumentMap.of(node.getArguments());
+    ChartConfig config = ChartConfig.fromArguments(argMap);
     List<UnresolvedExpression> groupExprList =
         Stream.of(node.getRowSplit(), node.getColumnSplit()).filter(Objects::nonNull).toList();
-    ChartConfig config = ChartConfig.fromArguments(argMap);
     Aggregation aggregation =
         new Aggregation(
             List.of(node.getAggregationFunction()), List.of(), groupExprList, null, List.of());
@@ -2441,15 +2441,6 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
       return relBuilder.peek();
     }
 
-    String aggFunctionName = getAggFunctionName(node.getAggregationFunction());
-    BuiltinFunctionName aggFunction =
-        BuiltinFunctionName.of(aggFunctionName)
-            .orElseThrow(
-                () ->
-                    new IllegalArgumentException(
-                        StringUtils.format(
-                            "Unrecognized aggregation function: %s", aggFunctionName)));
-
     // Convert the column split to string if necessary: column split was supposed to be pivoted to
     // column names. This guarantees that its type compatibility with useother and usenull
     RexNode colSplit = relBuilder.field(1);
@@ -2463,34 +2454,8 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     }
     relBuilder.project(relBuilder.field(0), colSplit, relBuilder.field(2));
     RelNode aggregated = relBuilder.peek();
-
     // 1: column-split, 2: agg
-    relBuilder.project(relBuilder.field(1), relBuilder.field(2));
-    // Make sure that rows who don't have a column split not interfere grand total calculation
-    relBuilder.filter(relBuilder.isNotNull(relBuilder.field(0)));
-    final String GRAND_TOTAL_COL = "__grand_total__";
-    relBuilder.aggregate(
-        relBuilder.groupKey(relBuilder.field(0)),
-        // Top-K semantic: Retain categories whose summed values are among the greatest
-        relBuilder.sum(relBuilder.field(1)).as(GRAND_TOTAL_COL)); // results: group key, agg calls
-    RexNode grandTotal = relBuilder.field(GRAND_TOTAL_COL);
-    // Apply sorting: keep the max values if top is set
-    if (config.top) {
-      grandTotal = relBuilder.desc(grandTotal);
-    }
-    // Always set it to null last so that nulls don't interfere with top / bottom calculation
-    grandTotal = relBuilder.nullsLast(grandTotal);
-    RexNode rowNum =
-        PlanUtils.makeOver(
-            context,
-            BuiltinFunctionName.ROW_NUMBER,
-            relBuilder.literal(1), // dummy expression for row number calculation
-            List.of(),
-            List.of(),
-            List.of(grandTotal),
-            WindowFrame.toCurrentRow());
-    relBuilder.projectPlus(relBuilder.alias(rowNum, PlanUtils.ROW_NUMBER_COLUMN_FOR_CHART));
-    RelNode ranked = relBuilder.build();
+    RelNode ranked = rankByColumnSplit(context, 1, 2, config.top);
 
     relBuilder.push(aggregated);
     relBuilder.push(ranked);
@@ -2534,12 +2499,56 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
         relBuilder.field(0),
         relBuilder.alias(columnSplitExpr, columnSplitName),
         relBuilder.field(2));
+    String aggFunctionName = getAggFunctionName(node.getAggregationFunction());
+    BuiltinFunctionName aggFunction =
+        BuiltinFunctionName.of(aggFunctionName)
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        StringUtils.format(
+                            "Unrecognized aggregation function: %s", aggFunctionName)));
     relBuilder.aggregate(
         relBuilder.groupKey(relBuilder.field(0), relBuilder.field(1)),
         buildAggCall(context.relBuilder, aggFunction, relBuilder.field(2)).as(aggFieldName));
     // The output of chart is expected to be ordered by row and column split names
     relBuilder.sort(relBuilder.field(0), relBuilder.field(1));
     return relBuilder.peek();
+  }
+
+  /**
+   * Aggregate by column split then rank by grand total (summed value of each category). The output
+   * is <code>[col-split, grand-total, row-number]</code>
+   */
+  private RelNode rankByColumnSplit(
+      CalcitePlanContext context, int columnSplitOrdinal, int aggOrdinal, boolean top) {
+    RelBuilder relBuilder = context.relBuilder;
+
+    relBuilder.project(relBuilder.field(columnSplitOrdinal), relBuilder.field(aggOrdinal));
+    // Make sure that rows who don't have a column split not interfere grand total calculation
+    relBuilder.filter(relBuilder.isNotNull(relBuilder.field(0)));
+    final String GRAND_TOTAL_COL = "__grand_total__";
+    relBuilder.aggregate(
+        relBuilder.groupKey(relBuilder.field(0)),
+        // Top-K semantic: Retain categories whose summed values are among the greatest
+        relBuilder.sum(relBuilder.field(1)).as(GRAND_TOTAL_COL)); // results: group key, agg calls
+    RexNode grandTotal = relBuilder.field(GRAND_TOTAL_COL);
+    // Apply sorting: keep the max values if top is set
+    if (top) {
+      grandTotal = relBuilder.desc(grandTotal);
+    }
+    // Always set it to null last so that nulls don't interfere with top / bottom calculation
+    grandTotal = relBuilder.nullsLast(grandTotal);
+    RexNode rowNum =
+        PlanUtils.makeOver(
+            context,
+            BuiltinFunctionName.ROW_NUMBER,
+            relBuilder.literal(1), // dummy expression for row number calculation
+            List.of(),
+            List.of(),
+            List.of(grandTotal),
+            WindowFrame.toCurrentRow());
+    relBuilder.projectPlus(relBuilder.alias(rowNum, PlanUtils.ROW_NUMBER_COLUMN_FOR_CHART));
+    return relBuilder.build();
   }
 
   @AllArgsConstructor


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2019,6 +2040,153 @@ private String getValueFunctionName(UnresolvedExpression aggregateFunction) {
...
+        BuiltinFunctionName.of(aggFunctionName)
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        StringUtils.format(
+                            "Unrecognized aggregation function: %s", aggFunctionName)));
+
+    // Convert the column split to string if necessary: column split was supposed to be pivoted to
+    // column names. This guarantees that its type compatibility with useother and usenull
+    RexNode colSplit = relBuilder.field(1);
```
yuancu, line 2446: The fields are `[row-split, col-split, aggregation]` now

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2019,6 +2040,153 @@ private String getValueFunctionName(UnresolvedExpression aggregateFunction) {
...
+    String columSplitName = relBuilder.peek().getRowType().getFieldNames().get(1);
+    if (!SqlTypeUtil.isCharacter(colSplit.getType())) {
+      colSplit =
+          relBuilder.alias(
+              context.rexBuilder.makeCast(
+                  UserDefinedFunctionUtils.NULLABLE_STRING, colSplit, true, true),
+              columSplitName);
+    }
```
yuancu, line 2454: Convert the column split to string so that they can be labels of columns once pivoted. This also guarantees that its type is compatible with `nullstr` and `otherstr`.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2019,6 +2040,153 @@ private String getValueFunctionName(UnresolvedExpression aggregateFunction) {
...
+    relBuilder.project(relBuilder.field(0), colSplit, relBuilder.field(2));
+    aggregated = relBuilder.peek();
+
+    // 1: column-split, 2: agg
+    relBuilder.project(relBuilder.field(1), relBuilder.field(2));
+    // Make sure that rows who don't have a column split not interfere grand total calculation
+    relBuilder.filter(relBuilder.isNotNull(relBuilder.field(0)));
```
yuancu: `testChartWithNullAndLimit` covers this case. Without this line, it will number rows who don't have a column split if their aggregation result is great.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2019,6 +2040,153 @@ private String getValueFunctionName(UnresolvedExpression aggregateFunction) {
...
+        relBuilder.groupKey(relBuilder.field(0)),
+        buildAggCall(context.relBuilder, aggFunction, relBuilder.field(1))
+            .as(GRAND_TOTAL_COL)); // results: group key, agg calls
+    RexNode grandTotal = relBuilder.field(GRAND_TOTAL_COL);
+    // Apply sorting: for MIN/EARLIEST, reverse the top/bottom logic
+    boolean smallestFirst =
+        aggFunction == BuiltinFunctionName.MIN || aggFunction == BuiltinFunctionName.EARLIEST;
+    if (config.top != smallestFirst) {
+      grandTotal = relBuilder.desc(grandTotal);
+    }
```
yuancu: See explanations in #4594

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2019,6 +2040,153 @@ private String getValueFunctionName(UnresolvedExpression aggregateFunction) {
...
+            WindowFrame.toCurrentRow());
+    relBuilder.projectPlus(relBuilder.alias(rowNum, ROW_NUM_COL));
+    RelNode ranked = relBuilder.build();
+
+    relBuilder.push(aggregated);
+    relBuilder.push(ranked);
+
+    // on column-split = group key
+    relBuilder.join(
+        JoinRelType.LEFT, relBuilder.equals(relBuilder.field(2, 0, 1), relBuilder.field(2, 1, 0)));
```
yuancu, line 2465: aggregated: `[row-split, col-split, aggregation]`
ranked: `[col-split, grand-total, row-number]`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2019,6 +2040,153 @@ private String getValueFunctionName(UnresolvedExpression aggregateFunction) {
...
+    }
+
+    String aggFieldName = relBuilder.peek().getRowType().getFieldNames().get(2);
+    relBuilder.project(
+        relBuilder.field(0),
+        relBuilder.alias(columnSplitExpr, columnSplitName),
+        relBuilder.field(2));
+    relBuilder.aggregate(
+        relBuilder.groupKey(relBuilder.field(0), relBuilder.field(1)),
+        buildAggCall(context.relBuilder, aggFunction, relBuilder.field(2)).as(aggFieldName));
```
yuancu, line 2512: Final aggregation: to merge values in the `OTHER` categories.

docs/user/ppl/cmd/chart.rst
```diff
@@ -0,0 +1,208 @@
...
+   [limit=(top|bottom) <number>] [useother=<boolean>] [usenull=<boolean>] [nullstr=<string>] [otherstr=<string>]
+   <aggregation_function>
+   [ by <row_split> <column_split> ] | [over <row_split> ] [ by <column_split>]
+
+**Parameters:**
+
+* **limit**: optional. Specifies the number of distinct values to display when using column split.
+
+  * Default: 10
+  * Syntax: ``limit=(top|bottom) <number>`` or ``limit=<number>`` (defaults to top)
```
penghuo: Explain definition of top in doc,  is it `stats dc by col | sort -dc, +col`?

docs/user/ppl/cmd/chart.rst
```diff
@@ -0,0 +1,208 @@
...
+  * When there are more distinct column split values than the limit, the additional values are grouped into an "OTHER" category if ``useother`` is not set to false.
+  * Set to 0 to show all distinct values without any limit.
+  * Only applies when column split presents (by 2 fields or over...by... coexists).
+
+* **useother**: optional. Controls whether to create an "OTHER" category for distinct column values beyond the limit.
```
penghuo: column -> column_split

docs/user/ppl/cmd/chart.rst
```diff
@@ -0,0 +1,208 @@
...
+
+  * Default: true
+  * When set to false, only the top/bottom N distinct values (based on limit) are shown without an "OTHER" category.
+  * When set to true, distinct values beyond the limit are grouped into an "OTHER" category.
+  * Only applies when using column split and when there are more distinct column values than the limit.
+
+* **usenull**: optional. Controls whether to include null values as a separate category.
```
penghuo: change doc, make it clearly.
* usenull=true only applie to column_split
* row_split should always be non-null value.

docs/user/ppl/cmd/chart.rst
```diff
@@ -0,0 +1,208 @@
...
+
+* **over...by...**: optional. Alternative syntax for grouping by multiple fields.
+
+  * ``over <row_split> by <column_split>`` groups the results by both fields.
+  * Using ``over`` alone on one field is equivalent to ``by <row-split>``
+
+Notes
+=====
+
+* The column split field in the result will become strings so that they are compatible with ``nullstr`` and ``otherstr`` and can be used as column names once pivoted.
```
penghuo: The column split field in the result will become strings ->
The fields generated by column splitting are converted to strings

docs/user/ppl/cmd/chart.rst
```diff
@@ -0,0 +1,208 @@
...
+
+    os> source=accounts | chart limit=1 count() over gender by age
+    fetched rows / total rows = 3/3
+    +--------+-------+---------+
+    | gender | age   | count() |
+    |--------+-------+---------|
+    | M      | OTHER | 2       |
+    | M      | 33    | 1       |
+    | F      | OTHER | 1       |
+    +--------+-------+---------+
```
penghuo, line 157: I expect result should another row?
F 33 0

then, pivot table will be
gender 33 OTHER
M,1,2
F,1,0

docs/user/ppl/cmd/chart.rst
```diff
@@ -0,0 +1,208 @@
...
+
+Example 5: Using limit with other parameters
+--------------------------------------------
+
+This example shows using limit with useother and custom otherstr parameters.
+
+PPL query::
+
+    os> source=accounts | chart limit=top 1 useother=true otherstr='minor_gender' count() over state by gender
```
penghuo: syntax should be `limit=top10`

docs/user/ppl/cmd/chart.rst
```diff
@@ -0,0 +1,208 @@
...
+    +-------+--------------+---------+
+
+Example 6: Using null parameters
+--------------------------------
+
+This example shows using limit with usenull and custom nullstr parameters.
+
+PPL query::
+
+    os> source=accounts |  chart usenull=true nullstr='employer not specified' count() over firstname by employer
```
penghuo, line 184: add an example to demo convert column_split to string.

docs/user/ppl/cmd/chart.rst
yuancu: Since we are not really doing pivoting, I think it's better to omit empty groups? This avoids a large sparse response and reduces traffic. Besides, other aggregations also don't return results for empty buckets.

`timechart` also claims to omit those buckets:

> Only combinations with actual data are included in the results - empty combinations are omitted rather than showing null or zero values.

If front-end wants to add it back, they can easily fill null or 0 to those missing groups.

docs/user/ppl/cmd/chart.rst
yuancu: It's to keep the top-K categories (distinct column splits).

E.g.
- `chart limit=1 count() by a b` keeps the top 1 b categories with most rows
- `chart limit=bottom3 sum(value) by a b` keeps the one b categories with minimum sum of values in its category.
- `chart limit=top2 min(value) by a b` keeps 2 b categories whose minimum value within its group are the smallest 2.
- `chart limit=bottom2 min(value) by a b` keep 2 b categories whose minimum value within its group are the largest 2.

docs/user/ppl/cmd/chart.rst
yuancu: Fixed

docs/user/ppl/cmd/chart.rst
yuancu: Fixed. row_split can actually contain `null`; it will be handled in the same manner as normal aggregations like `stats count() by a, b` where there exists null values in column a.

docs/user/ppl/cmd/chart.rst
yuancu: Fixed. Thanks for the suggestion!

docs/user/ppl/cmd/chart.rst
yuancu: Fixed. Thanks for double checking

docs/user/ppl/cmd/chart.rst
yuancu: It's actually covered by example 3 and 4. Updated their descriptions.

integ-test/src/test/resources/expectedOutput/calcite/explain_chart_multiple_group_keys.yaml
```diff
@@ -0,0 +1,33 @@
+calcite:
+  logical: |
+    LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
+      LogicalAggregate(group=[{0, 1}], avg(balance)=[AVG($2)])
```
qianheng-aws: `timechart` has another sort operator upon the final aggregate operator to ensure the sequence. Does `chart` has to keep the same behavior?

docs/user/ppl/cmd/chart.rst
qianheng-aws:
```
 // Use zero-filling for count aggregations, standard result for others
    if (valueFunctionName.equals("count")) {
      return buildZeroFilledResult(
          completeResults, topCategories, byFieldName, valueFunctionName, useOther, limit, context);
    } else {
      return buildStandardResult(
          completeResults, topCategories, byFieldName, valueFunctionName, useOther, context);
    }
```

`timechart` fills zero specially only for `count` agg and that's what makes the plan so complex for `count`. I cannot recall the reason why we do this, maybe just want to keep align with SPL?

yuancu: Yes, her aim was to align with SPL. In timechart's issue (#3965), it states:
> Post-Processing: The formatter layer transforms this "long" format into "wide" format by:
> - Filling missing values with 0 for hosts with no data in a time bucket

However, I don't find filling 0 only for count is necessary considering that we are not really converting the output to a *wide* format. This is also inconsistent with the output of other aggregation functions, where the values for missing group combinations (`{F-33}` for the above case) are not filled. If front-end side is going to pivoting the result, it's better to let it fill the missing value as well; if they don't, it does not make much sense to manually create the empty group combinations just for count.

integ-test/src/test/resources/expectedOutput/calcite/explain_chart_multiple_group_keys.yaml
yuancu: Good catch. I believe the added order is to align with SPL's behavior. The row and column names are indeed sorted in SPL.

Yet I'm a little unsure whether it is necessary as SPL's doc did not explicit or guarantee any order of its output for chart and timechart.

yuancu: I added the final reorder. I'll revert it if it's unnecessary.

docs/user/ppl/cmd/chart.rst
penghuo: > Null values in the row split are handled in the same way as normal aggregations.

null value in row split should be ignored.

docs/user/ppl/cmd/chart.rst
penghuo: I suspect topk is not distinct columns splits.  e.g.  depend on metrics agg
* case-1, distinct columns
```
| chart limit=1 count by id, depto

## case-1, distinct columns
| stats dc by split_cols
```

* case-2, remove invalide rows, then distinct columns
```
| chart limit=1 sum(age) as sum by id, depto

## case-2, remove invalide rows, then distinct columns
| where isnumber(age)
| stats dc by split_cols
```

docs/user/ppl/cmd/chart.rst
penghuo: chart and timechart should have same behavior on filling. 
@yuancu could u sync with yulong to double confirm the frontend can fill results when pivoting.

docs/user/ppl/cmd/chart.rst
yuancu: Sorry for missing the point. Double confirmed SPL's behavior, null values in row splits are indeed ignored.

**Update**: Fixed; ignores events without a row split now

docs/user/ppl/cmd/chart.rst
yuancu: I discussed with yulong, filling 0 for empty group for count in front-end is not a problem.

I'll leave the empty groups unfilled for now.

docs/user/ppl/cmd/chart.rst
yuancu: The top-k columns seems to be defined as *the categories whose summed values are among the greatest-K*.

E.g. for the following data

```
[
{ "age":10, "group":1, "depto":"a"},
{"age":20,  "group":2, "depto":"a"},
{"age":22, "group":1, "depto":"b"} ,
{"age":6, "group":2, "depto":"b"}
]
```

`| chart limit=1 min(age) by group, depto` has the following result:

| group | a  | OTHER |
|-------|----|----- |
| 1     | 10 | 22    |
| 2     | 20 | 6     |

Not the category with minimum value (`[2, b]` has 6 as minimum value), nor the category with minimum sum (category b has the sum of 28, which is smaller that 30) is retained, but the column with maximum sum is kept. 

**Update**: Updated the implementation to mimic the behavior

docs/user/ppl/cmd/chart.rst
```diff
@@ -0,0 +1,212 @@
...
+
+**Parameters:**
+
+* **limit**: optional. Specifies the number of categories to display when using column split. Each unique value in the column split field represents a category.
+
+  * Default: top10
+  * Syntax: ``limit=(top|bottom)<number>`` or ``limit=<number>`` (defaults to top)
+  * When ``limit=K`` is set, the top or bottom K categories from the column split field are retained; the remaining categories are grouped into an "OTHER" category if ``useother`` is not set to false.
+  * Set limit to 0 to show all categories without any limit.
+  * Use ``limit=topK`` or ``limit=bottomK`` to specify whether to retain the top or bottom K column categories. The ranking is based on the aggregated values for each category.  For example, ``chart limit=top3 count() by a b`` retains the 3 most common b categories; ``chart limit=top5 min(value) by a b`` selects the 5 b categories that contain the smallest aggregated values. If not specified, top is used by default.
```
penghuo: Update example, the topk is define is
```
| stats aggregation_function() as m by row_split, col_split 
| stats sum(m) as m by col_split
| sort -m
| head k
```

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -287,6 +288,28 @@ reverseCommand
    : REVERSE
    ;
 
+chartCommand
+  : CHART chartOptions* statsAggTerm (OVER rowSplit)? (BY columnSplit)?
+  | CHART chartOptions* statsAggTerm BY rowSplit (COMMA)? columnSplit
```
penghuo, line 293: Not a blocker, but we should support chartOptions at beginning and at the end. This will allow users to modify queries easily without changing the aggregation and group-by components.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteChartCommandIT.java
```diff
@@ -0,0 +1,335 @@
...
+        result,
+        schema("gender", "string"),
+        schema("age", "string"),
+        schema("avg(balance)", "double"));
+    verifyDataRows(
+        result,
+        rows("M", "30", 21702.5),
+        rows("F", "30", 48086.0),
+        rows("F", "20", 32838.0),
+        rows("F", "nil", null));
```
penghuo: since avg(balance) is null, this row should be ignored.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -1414,6 +1406,65 @@ public void testPushDownMinOrMaxAggOnDerivedField() throws IOException {
...
+        loadExpectedPlan("explain_chart_single_group_key.yaml"),
+        explainQueryYaml(
+            String.format("source=%s | chart avg(balance) by gender", TEST_INDEX_BANK)));
+
+    assertYamlEqualsIgnoreId(
+        loadExpectedPlan("explain_chart_with_span.yaml"),
+        explainQueryYaml(
+            String.format("source=%s | chart max(balance) by age span=10", TEST_INDEX_BANK)));
+
+    assertYamlEqualsIgnoreId(
```
penghuo, line 1455: query include row_split and col_split, should move to testExplainChartWithMultipleGroupKeys?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
penghuo: non-blocker, can we rely on type-corecetion? if(row_number<topK, row, "OTHER")

integ-test/src/test/resources/expectedOutput/calcite/explain_chart_with_limit.yaml
```diff
@@ -0,0 +1,10 @@
+calcite:
+  logical: |
+    LogicalSystemLimit(sort0=[$0], dir0=[ASC], fetch=[10000], type=[QUERY_SIZE_LIMIT])
+      LogicalSort(sort0=[$0], dir0=[ASC])
+        LogicalAggregate(group=[{0, 1}], avg(balance)=[AVG($2)])
```
penghuo: avg(balance) should be non-null value?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2375,6 +2394,152 @@ private String getValueFunctionName(UnresolvedExpression aggregateFunction) {
...
+    // The output of chart is expected to be ordered by row and column split names
+    relBuilder.sort(relBuilder.field(0), relBuilder.field(1));
+    return relBuilder.peek();
+  }
+
+  private record ChartConfig(
```
penghuo: non-blocker, As we need to backport PR to 2.19-dev, Try to avoid record.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2375,6 +2394,152 @@ private String getValueFunctionName(UnresolvedExpression aggregateFunction) {
...
+    if (config.useNull) {
+      nonNullGroupMask.set(0);
+    } else {
+      nonNullGroupMask.set(0, groupExprList.size());
+    }
+    visitAggregation(aggregation, context, false, nonNullGroupMask);
+    RelBuilder relBuilder = context.relBuilder;
+
+    // If row or column split does not present or limit equals 0, this is the same as `stats agg
+    // [group by col]` because all truncating is performed on the column split
```
penghuo: row or column split does not present or limit equals 0, chart and stats produce different results, 

```
[
    {"_time":1755622404, "age":"10", "id":"1", "deptno":"a"},
    {"_time":1755622405, "age":10, "id":"2", "deptno":"a"},
    {"_time":1755622408, "age":"str", "id":"1", "deptno":"b"},
    {"_time":1755622409, "age":"str", "id":"2", "deptno":"b"}
]
```
* chart should ignore null value
```
| chart limit=0 sum(age) by id, deptno

id, deptno, sum
1, a, 10
2, a, 10
```

* stats keep null value
```
| stats sum(age) by id, depto

id, deptno, sum
1, a, 10
1, b, null
2, a, 10
2, b, null
```

ppl/src/main/antlr/OpenSearchPPLParser.g4
yuancu: I'm a little confused on what do you mean by *support chartOptions eventually*. Do you mean that we should support more chart options or we should support placing them not only before `statsAggTerm`, etc?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteChartCommandIT.java
yuancu: Do you mean that we should add a special nonnull filter for chart post its aggregation?

The query is `source=%s | chart nullstr='nil' avg(balance) over gender by age span=10`. Based on the following criteria:

>  * ``usenull`` only applies to column split.
>  * Row split should always be non-null value.

Documents whose `age` is null will be kept; documents whose `gender` is null will be ignored.

 This leads to the group `["F", "nil", null]` in the result, which is derived from the document `{"gender": "F", "age": null, "balance": null}`. It's `balance` happens to be `null`, therefore, the result aggregation is `null`.

------

This aligns with SPL's behavior indeed.

After a second thought, I think the non-null filter should be applied directly to the original field. I.e. apply it on age **before** the aggregation instead of apply it on `avg(age) after aggregation.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
yuancu: Moved. Thanks for pointing out.

integ-test/src/test/resources/expectedOutput/calcite/explain_chart_with_limit.yaml
yuancu: Fixed by applying a non-null filter on fields referred by aggregations.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteChartCommandIT.java
yuancu: Fixed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: Fixed. Thanks for catching the problem

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: It faces 2 challenges to apply type-coercion here:

1. Existing coercion between number and string implemented in #4599 supports only **string -> double**, all the rest combination is not supported. Therefore, this case won't work: `if(row_number<topK, number_field, "OTHER")` because number -> string coercion is not supported.
2. The current type checking does not support `BOOLEAN_SAME_SAME` well. Its type checker is defined as `PPLTypeChecker.family(SqlTypeFamily.BOOLEAN, SqlTypeFamily.ANY, SqlTypeFamily.ANY)`. `[boolean, integer, string]` can pass the check. An error will be thrown if later when it can not find a least restrictive type between integer and string as the return type. But it will not perform coercion by then. A not-so-elegant workaround is to define its type checker as something like
    ```java
    PPLTypeChecker.wrapComposite((CompositeOperandTypeChecker) 
    OperandTypes.family(SqlTypeFamily.BOOLEAN, SqlTypeFamily.INTEGER, SqlTypeFamily.INTEGER)
      .or(OperandTypes.family(SqlTypeFamily.BOOLEAN, SqlTypeFamily.BOOLEAN, SqlTypeFamily.BOOLEAN))
      .or(OperandTypes.family(SqlTypeFamily.BOOLEAN, SqlTypeFamily.CHARACTER, SqlTypeFamily.CHARACTER))
      .or(...).or(....)
    , false));
    ```

Therefore, I think it would be better to leave it as is until coercion is refactored in the future.

docs/user/ppl/cmd/chart.rst
yuancu: Doc updated

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: Replaced with normal class

ppl/src/main/antlr/OpenSearchPPLParser.g4
penghuo: my bad, we should support chartOptions at beginning and at the end

penghuo: e.g. chart count by row over col limit=10

The reason is that PPL discovery customer are typically keyboard-focused, making it easier for them to modify options at the end of the query rather than at the beginning.

yuancu: I get it. I'll improve the grammar in the following PR unifying timechart & chart

qianheng-aws (approved):

yuancu:
> Thanks for the change! Please follow up on unifiy chart and timechart implementation in following PRs.

Thanks @penghuo ! Raisee #4755 for the unification

PR #4586, opened 2025-10-16
Allow renaming group-by fields to existing field names

### LINKED ISSUE #4580: [BUG] Group-by fields can not be aliased to the original field names with Calcite
**What is the bug?**

Queries that rename the group-by fields to their original field names will report errors when Calcite is enabled:

```
source=time_test | stats count() by span(value, 2000) as value
source=time_test | stats count() by span(timestamp, 1h) as timestamp
```

The error reports something like `field [field_name]` not found, input fields are `[field_name0, ...]`
```json
{
  "error": {
    "reason": "Invalid Query",
    "details": "field [value] not found; input fields are: [value0, count()]",
  },
  "status": 400
}
```

This prohibits queries like `by field_name span=xxx`, where the output field name remains the same with the original field (`field_name` in this case). 

```
POST /_plugins/_ppl
{
  "query" : "source=time_test | stats count() by span(value, 2000) as value"
}
```

The query can be executed when Calcite is disabled:

Output:

```json
{
  "schema": [
    {
      "name": "count()",
      "type": "bigint"
    },
    {
      "name": "value",
      "type": "int"
    }
  ],
  "datarows": [
    [
      49,
      6000
    ],
    [
      51,
      8000
    ]
  ],
  "total": 2,
  "size": 2
}
```

 - OS: macOS
 - Version: 3.3.0
 - Plugins

yuancu:
### Description

This PR fixes a bug in Calcite-enabled PPL queries where group-by fields cannot be aliased to their original field names, causing queries to fail with "field not found" errors.

When Calcite is enabled, PPL queries that use span functions with aliases matching the original field names fail with errors like: `field [value] not found; input fields are: [value0, count()]`

**Affected Query Patterns:**
- `source=time_test | stats count() by span(value, 2000) as value`
- `source=time_test | stats count() by span(timestamp, 1h) as timestamp`

**Root Cause Analysis**

The issue occurs during Calcite's aggregation processing:

1. Input Dependencies: Calcite identifies that both the input field value and output alias value have the same name
2. Name Collision Resolution: To avoid conflicts, Calcite automatically renames fields by adding numeric suffixes (value → value0)
3. Reference Resolution Failure: Subsequent query processing still references the original name value, but Calcite has internally renamed it to value0, causing a "field not found" error

**Solution Implementation**

This PR implements a post-aggregation field renaming strategy that preserves intended aliases.

### Related Issues
Resolves #4580

### Check List

### DIFF
--- a/.github/workflows/maven-publish.yml
+++ b/.github/workflows/maven-publish.yml
@@ -8,9 +8,6 @@ on:
       - '[0-9]+.[0-9]+'
       - '[0-9]+.x'
 
-env:
-  SNAPSHOT_REPO_URL: https://ci.opensearch.org/ci/dbc/snapshots/maven/
-
 jobs:
   build-and-publish-snapshots:
     strategy:
--- a/.github/workflows/publish-async-query-core.yml
+++ b/.github/workflows/publish-async-query-core.yml
@@ -5,8 +5,8 @@ on:
   push:
     branches:
       - main
-      - 1.*
-      - 2.*
+      - '[0-9]+.[0-9]+'
+      - '[0-9]+.x'
     paths:
       - 'async-query-core/**'
       - '.github/workflows/publish-async-query-core.yml'
@@ -18,7 +18,6 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  SNAPSHOT_REPO_URL: https://central.sonatype.com/repository/maven-snapshots/
   COMMIT_MAP_FILENAME: commit-history-async-query-core.json
 
 jobs:
@@ -47,8 +46,19 @@ jobs:
           export-env: true
         env:
           OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
-          SONATYPE_USERNAME: op://opensearch-infra-secrets/maven-central-portal-credentials/username
-          SONATYPE_PASSWORD: op://opensearch-infra-secrets/maven-central-portal-credentials/password
+          MAVEN_SNAPSHOTS_S3_REPO: op://opensearch-infra-secrets/maven-snapshots-s3/repo
+          MAVEN_SNAPSHOTS_S3_ROLE: op://opensearch-infra-secrets/maven-snapshots-s3/role
+
+      - name: Export SNAPSHOT_REPO_URL
+        run: |
+          snapshot_repo_url=${{ env.MAVEN_SNAPSHOTS_S3_REPO }}
+          echo "SNAPSHOT_REPO_URL=$snapshot_repo_url" >> $GITHUB_ENV
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v5
+        with:
+          role-to-assume: ${{ env.MAVEN_SNAPSHOTS_S3_ROLE }}
+          aws-region: us-east-1
 
       - name: Set commit ID
         id: set_commit
--- a/.github/workflows/publish-grammar-files.yml
+++ b/.github/workflows/publish-grammar-files.yml
@@ -5,8 +5,8 @@ on:
   push:
     branches:
       - main
-      - 1.*
-      - 2.*
+      - '[0-9]+.[0-9]+'
+      - '[0-9]+.x'
     paths:
       - 'language-grammar/src/main/antlr4/**'
       - 'language-grammar/build.gradle'
@@ -19,7 +19,6 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  SNAPSHOT_REPO_URL: https://central.sonatype.com/repository/maven-snapshots/
   COMMIT_MAP_FILENAME: commit-history-language-grammar.json
 
 jobs:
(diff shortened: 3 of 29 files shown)


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -997,13 +997,54 @@ private Pair<List<RexNode>, List<AggCall>> aggregateWithTrimming(
...
+  /** Whether a rex node is an aliased input reference */
+  private boolean isInputRef(RexNode node) {
+    return switch (node.getKind()) {
+      case AS, DESCENDING, NULLS_FIRST, NULLS_LAST -> {
+        final List<RexNode> operands = ((RexCall) node).operands;
+        yield isInputRef(operands.getFirst());
+      }
+      default -> node instanceof RexInputRef;
+    };
+  }
```
LantaoJin, line 1051: can the `PlanUtil.getInputRefs` be used to replace this?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -997,13 +997,54 @@ private Pair<List<RexNode>, List<AggCall>> aggregateWithTrimming(
...
         resolveAttributesForAggregation(groupExprList, aggExprList, context);
-
+    List<String> names = getGroupKeyNamesAfterAggregation(reResolved.getLeft());
     context.relBuilder.aggregate(
         context.relBuilder.groupKey(reResolved.getLeft()), reResolved.getRight());
-
+    // During aggregation, Calcite projects both input dependencies and output group-by fields.
+    // When names conflict, Calcite adds numeric suffixes (e.g., "value0").
+    // Apply explicit renaming to restore the intended aliases.
+    if (names.size() == reResolved.getLeft().size()) {
```
LantaoJin: when the `names.size` not equals to `reResolved.getLeft().size()`? seems the condition is always true

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -997,13 +997,54 @@ private Pair<List<RexNode>, List<AggCall>> aggregateWithTrimming(
     // because that Mapping only works for RexNode, but we need both AggCall and RexNode list.
     Pair<List<RexNode>, List<AggCall>> reResolved =
         resolveAttributesForAggregation(groupExprList, aggExprList, context);
-
+    List<String> names = getGroupKeyNamesAfterAggregation(reResolved.getLeft());
```
LantaoJin: can you rename the var `names` to make it more meaningful

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: The lengths do not equal when a group key is not aliased -- under which circumstance `extractAliasLiteral` will return empty:
```java
private Optional<RexLiteral> extractAliasLiteral(RexNode node) {
  if (node == null) {
    return Optional.empty();
  } else if (node.getKind() == AS) {
    return Optional.of((RexLiteral) ((RexCall) node).getOperands().get(1));
  } else {
    return Optional.empty();
  }
```

Although it seems that all group keys are aliased in practice, this defense check was to prevent unintended future changes to avoid in-correspondent renaming. Should I remove it?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: Renamed

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: I think they serve different purposes. `PlanUtil.getInputRefs` returns all referred input refs. Besides, if a node refers multiple inputs, it will return all of them. Yet here I just want to check whether a node is an input ref (optionally aliased), keeping the node as is.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -998,12 +998,52 @@ private Pair<List<RexNode>, List<AggCall>> aggregateWithTrimming(
...
+    context.relBuilder.rename(intendedGroupKeyAliases);
 
     return Pair.of(reResolved.getLeft(), reResolved.getRight());
   }
 
+  /**
+   * Imitates {@code Registrar.registerExpression} of {@link RelBuilder} to derive the output order
+   * of group-by keys after aggregation.
+   *
+   * <p>The projected input reference comes first, while any other computed expression follows.
```
qianheng-aws, line 1018: In `Registrar.registerExpression`, seems the other computed expression won't promise following the original order if there is expression duplication.

But since our PPL only allow `span` expr in our group by and it cannot be combined with other `span` expr. This logic may be right and I cannot find any bad case so far.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -998,12 +998,52 @@ private Pair<List<RexNode>, List<AggCall>> aggregateWithTrimming(
...
+        .map(this::extractAliasLiteral)
+        .flatMap(Optional::stream)
+        .map(RexLiteral::stringValue)
+        .toList();
+  }
+
+  /** Whether a rex node is an aliased input reference */
+  private boolean isInputRef(RexNode node) {
+    return switch (node.getKind()) {
+      case AS, DESCENDING, NULLS_FIRST, NULLS_LAST -> {
```
qianheng-aws, line 1045: Is there any case that we have `DESCENDING, NULLS_FIRST, NULLS_LAST` in our `stats .. by ...` command

yuancu: No, I didn't manage to create any. It seems there is always a projection after sorting and before aggregation. 

E.g.
```
LogicalAggregate(group=[{0}], count()=[COUNT()])
  LogicalProject(value=[$2])
    LogicalSort(sort0=[$2], dir0=[DESC-nulls-last])
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: I found a bad case: `stats count() by value, value, @timestamp`. I'll fix it.

**Update**: Fixed by checking duplication

LantaoJin (approved):

qianheng-aws (approved):

PR #4599, opened 2025-10-18
Support Automatic Type Conversion for REX/SPATH/PARSE Command Extractions

### LINKED ISSUE #4356: [FEATURE] Automatic Type Conversion for REX/SPATH/PARSE Command Extractions
Yes. When using the REX command to extract numeric values and then performing arithmetic operations on them, the system fails because the extracted values are treated as strings rather than numbers.

For example, the following query fails:
```
source=log00001 | rex field=v "value=(?<digits>\d*)" | eval multi=digits * 10
```

With the error:
```
{
  "error": {
    "reason": "Invalid Query",
    "details": "MULTIPLY function expects {[INTEGER,INTEGER]|[INTEGER,DOUBLE]|[DOUBLE,INTEGER]|[DOUBLE,DOUBLE]|[INTERVAL,INTEGER]|[INTERVAL,DOUBLE]|[INTEGER,INTERVAL]|[DOUBLE,INTERVAL]}, but got [STRING,INTEGER]",
  },
  "status": 400
}
```

The issue is that the `rex` command extracts patterns as strings, but users often need to perform numeric operations on these extracted values.

**Automatic type inference**: The system should attempt to automatically convert extracted values to appropriate types when used in operations. For example, if a string can be converted as a number and used in a numeric operation, it should be automatically converted.

1. **Explicit conversion function** - Current workaround
   ```
   source=log00001 | rex field=v \"value=(?<digits>\\d*)\" | eval multi=cast(digits as int) * 10
   ```

2. **Type specification in rex command**: Allow users to specify the expected type of the extracted field directly in the rex command:
   ```
   source=log00001 | rex field=v "value=(?<digits:int>\d*)" | eval multi=digits * 10
   ```

This feature would significantly improve the user experience when working with log data, where numeric values are often embedded within text strings. Similar functionality exists in other log processing languages, which automatically handles type conversions in many contexts.

The ability to seamlessly extract and perform operations on numeric values would reduce query complexity and make the system more intuitive for users coming from other log analysis platforms.

**Test dataset**
```
###
PUT {{baseUrl}}/log00001

{
  "mappings": {
    "properties": {
      "v": {
        "type": "text"
      }
    }
  }
}

###

{"index": { "_id": 1 }}
{"v": "value=1000"}
{"index": { "_id": 2 }}
{"v": "value=2000"}
{"index": { "_id": 3 }}
{"v": "value=3000"}
{"index": { "_id": 4 }}
{"v": "value=4000"}
```

penghuo:
### Description
* Expand the Calcite coercion logic, so functions and aggregation operators automatically cast compatible strings to numeric or boolean overloads, and return the coerced Rex nodes to the execution pipeline.
* Test cover in [4356.yaml](https://github.com/penghuo/os-sql/blob/1066422cd23a2ba05ef74b5900a226b096a63dde/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4356.yml).
* [Document implicit conversion rules for PPL](https://github.com/opensearch-project/sql/pull/4599/files#diff-2cd6bff3d6c2cac7daebd7d067ebba2c545feb4c8332201f6bf78cede14b0b52) 
* To revieiwer
  * Add TypeChecker for SUBTRACT, CEIL/CEILING, FLOOR, and ROUND. It is a workaround to support sequence CompositeOperandTypeChecker.
  * To migrate user issue, current CoercionUtils change is hack, refactor the code in following PRs to include leverage TypeCoercionImpl

### Related Issues
https://github.com/opensearch-project/sql/issues/4634
Resolves #4356 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
@@ -1397,7 +1397,7 @@ public class PPLFuncImpTable {
       try {
         pplTypeChecker = PPLTypeChecker.wrapComposite(compositeTypeChecker, !isUserDefinedFunction);
       } catch (IllegalArgumentException | UnsupportedOperationException e) {
-        logger.warn(
+        logger.debug(
             String.format(
                 "Failed to create composite type checker for operator: %s. Will skip its type"
                     + " checking",


core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
```diff
@@ -136,4 +154,122 @@ private static boolean areDateAndTime(ExprType type1, ExprType type2) {
...
+  private static int distance(ExprType type1, ExprType type2) {
+    return distance(type1, type2, TYPE_EQUAL);
+  }
+
+  private static int distance(ExprType type1, ExprType type2, int distance) {
+    if (type1 == type2) {
+      return distance;
+    } else if (type1 == UNKNOWN) {
+      return IMPOSSIBLE_WIDENING;
+    } else if (type1 == ExprCoreType.STRING && type2 == ExprCoreType.DOUBLE) {
```
yuancu, line 225: Does this mean that the only viable cast between string and numbers are string -> double? Are all the rest like number -> string, string -> integer not possible?

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
```diff
@@ -90,9 +109,12 @@ public class CoercionUtils {
...
     }
-
-    if (WideningTypeRule.distance(argType, targetType) != WideningTypeRule.IMPOSSIBLE_WIDENING) {
-      return builder.makeCast(OpenSearchTypeFactory.convertExprTypeToRelDataType(targetType), arg);
+    if (distance(argType, targetType) != IMPOSSIBLE_WIDENING) {
+      return builder.makeCast(
+          OpenSearchTypeFactory.convertExprTypeToRelDataType(targetType), arg, true, true);
+    } else if (argType == ExprCoreType.STRING && NUMBER_TYPES.contains(targetType)) {
+      return builder.makeCast(
+          OpenSearchTypeFactory.convertExprTypeToRelDataType(ExprCoreType.DOUBLE), arg, true, true);
```
yuancu: I'm wondering why is here a special handling for string and number types. If I got it correctly, the computed targetType should already be a double if the arugment is a string and there exists a signature that accepts double.

I'm worried that it will cast string to double for cases where there actually only accepts integer: e.g. will the `"1"` in `week('2020-12-11', '1')` be casted to double 1?

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
```diff
@@ -136,4 +154,122 @@ private static boolean areDateAndTime(ExprType type1, ExprType type2) {
...
+  /**
+   * The max type among two types. The max is defined as follow if type1 could widen to type2, then
+   * max is type2, vice versa if type1 couldn't widen to type2 and type2 could't widen to type1,
+   * then throw {@link ExpressionEvaluationException}.
+   *
+   * @param type1 type1
+   * @param type2 type2
+   * @return the max type among two types.
+   */
+  public static ExprType max(ExprType type1, ExprType type2) {
```
yuancu, line 244: Do we rewrite `WideningTypeRule.max` and `WideningTypeRule.distance` to avoid introducing breaking changes to v2?

penghuo: Yes.
The straitforward solution is add double as parent of string, but it will breaking existing V2 behavior.
I plan to decouple coercion logic from V2 type system. what do u think?

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
penghuo: yes, only string -> double (to the most widen numeric type). the reasosn is to handle most of use case, e.g. "1.1"
number -> string, maybe. not the [major use case](https://github.com/opensearch-project/sql/issues/4356) to solve.

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
penghuo: currently, distance function is measure direct relation of two data types.  for example, [string, int], will return IMPOSSIBLE, but we expected return [safe_cast(string as double), cast(int as double)].

Agree, code looks tricky, i am working on better way to handle this.

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
```diff
@@ -136,4 +154,122 @@ private static boolean areDateAndTime(ExprType type1, ExprType type2) {
...
+    static CoercionRule of(
+        BiPredicate<ExprType, ExprType> predicate, BinaryOperator<ExprType> resolver) {
+      return new CoercionRule(predicate, resolver);
+    }
+  }
+
+  private static final int IMPOSSIBLE_WIDENING = Integer.MAX_VALUE;
+  private static final int TYPE_EQUAL = 0;
+
+  private static int distance(ExprType type1, ExprType type2) {
```
penghuo, line 216: Copy distance max function WidenRule to avoid impact v2 behavior.

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
```diff
@@ -31,15 +39,26 @@ public class CoercionUtils {
       RexBuilder builder, PPLTypeChecker typeChecker, List<RexNode> arguments) {
     List<List<ExprType>> paramTypeCombinations = typeChecker.getParameterTypes();
```
qianheng-aws, line 40: As used to be mentioned in the original PR of type coercion, https://github.com/opensearch-project/sql/pull/3914#discussion_r2244919778, we may have chance to do type coercion based on Calcite's type system like `RelDataType` and SqlTypeName. Then we should be able to avoid reusing v2's coercion logic.

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
qianheng-aws: Shouldn't `findWidestType(arguments)` return `double` as the `targetType` for [string, int]? I see `COMMON_COERCION_RULES` has contained such a rule.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -445,9 +445,26 @@ public RexNode visitWindowFunction(WindowFunction node, CalcitePlanContext conte
...
-                  context, functionName, field, args, partitions, List.of(), node.getWindowFrame());
+              List<RexNode> nodes =
+                  PPLFuncImpTable.INSTANCE.validateAggFunctionSignature(
+                      functionName, field, args, context.rexBuilder);
+              return nodes != null
+                  ? PlanUtils.makeOver(
+                      context,
+                      functionName,
+                      nodes.getFirst(),
+                      args.subList(0, nodes.size()),
```
qianheng-aws: Shouldn't this be `nodes.subList(1, nodes.size())`?

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
penghuo: +1. the next step is refactor TypeCoercion logic, and try to leverage calcite TypeCoercionImpl interface.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
penghuo: Thanks!
https://github.com/opensearch-project/sql/pull/4599/commits/cde2238291a01016c22dcc4cb4573ba8177ce755

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -432,17 +436,21 @@ public RelBuilder.AggCall resolveAgg(
...
 
     var handler = implementation.getValue();
-    return handler.apply(distinct, field, argList, context);
+    return nodes != null
+        ? handler.apply(distinct, nodes.getFirst(), nodes.subList(1, nodes.size()), context)
+        : handler.apply(distinct, field, argList, context);
   }
 
-  static void validateFunctionArgs(
+  static List<RexNode> validateFunctionArgs(
```
penghuo, line 448: To reviewer:  validateFunctionArgs return coercionNode if possible, if not, throw ExpressionEvaluationException (current behavior).

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
penghuo: fixed. reuse resolveCommonType.

core/src/main/java/org/opensearch/sql/expression/function/CoercionUtils.java
yuancu: Let's keep it this way then. As mentioned in [another comment](https://github.com/opensearch-project/sql/pull/4599#discussion_r2452503624), we can refactor later when we adopt `RelDataType` for type coercion.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -455,19 +463,29 @@ static void validateFunctionArgs(
...
-              errorMessagePattern,
-              functionName,
-              signature.typeChecker().getAllowedSignatures(),
-              PlanUtils.getActualSignature(argTypes)));
+      List<RexNode> fields = new ArrayList<>();
+      fields.add(field);
+      fields.addAll(argList);
+      if (CoercionUtils.hasString(fields)) {
+        coercionNodes = CoercionUtils.castArguments(rexBuilder, signature.typeChecker(), fields);
+      }
```
yuancu, line 473: Why is the cast necessary when there are string in fields?

Update: I get it. It is to apply string->number coercions if necessary.

qianheng-aws (approved):

PR #4602, opened 2025-10-20
Support `appendpipe`command in PPL

xinyual:
### Description
Add appendpipe command. 
`Appendpipe` appends the result of the subpipeline to the search results. Unlike a subsearch, the subpipeline is not run first. The subpipeline is run when the search reaches the appendpipe command.

### Related Issues

### Check List


ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -212,6 +223,12 @@ public UnresolvedPlan visitWhereCommand(WhereCommandContext ctx) {
     return new Filter(internalVisitExpression(ctx.logicalExpression()));
   }
 
+  @Override
+  public UnresolvedPlan visitAppendPipeCommand(OpenSearchPPLParser.AppendPipeCommandContext ctx) {
+    UnresolvedPlan plan = visit(ctx.subPipeline());
+    return new AppendPipe(plan);
+  }
```
yuancu, line 254: Can you add a unit test in AstBuilderTest?

yuancu (commented):
Review in progress

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
xinyual: Already add one.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLAppendPipeCommandIT.java
```diff
@@ -0,0 +1,94 @@
...
+    JSONObject actual =
+        executeQuery(
+            String.format(
+                Locale.ROOT,
+                "source=%s | stats sum(age) as sum by gender |"
+                    + " appendpipe [  stats sum(sum) as sum ] | head 5",
+                TEST_INDEX_ACCOUNT,
+                TEST_INDEX_ACCOUNT));
+    verifySchemaInOrder(actual, schema("sum", "bigint"), schema("gender", "string"));
+    verifyDataRows(actual, rows(14947, "F"), rows(15224, "M"), rows(30171, null));
```
yuancu, line 73: I found a case that's slightly different from expectation: `source=opensearch-sql_test_index_account | stats count() by state, span(age, 10) as span | appendpipe [ stats count() by span | eval state='nomadland'] | sort span"`

The result is not merged although both `state` in the main- and sub-query are of string types:
```json
{
  "schema": [
    {
      "name": "count()",
      "type": "bigint"
    },
    {
      "name": "span",
      "type": "bigint"
    },
    {
      "name": "state",
      "type": "string"
    },
    {
      "name": "state0",
      "type": "string"
    }
  ]
}
```


This may originates from the fact that the `state` in the main query is nullable, while the type of `state` in the subquery is not. IMO, they can be merged, but I'm also not quite sure whether this should be a concern.

xinyual: I think it's because I don't merge the latest code. This problem is already fixed by #4512

xinyual:
@Swiddis could you please approve again?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -239,6 +240,25 @@ public RelNode visitFilter(Filter node, CalcitePlanContext context) {
...
 
+  @Override
+  public RelNode visitAppendPipe(AppendPipe node, CalcitePlanContext context) {
+    visitChildren(node, context);
+    UnresolvedPlan subqueryPlan = node.getSubQuery();
+    UnresolvedPlan childNode = subqueryPlan;
+    while (childNode.getChild() != null
+        && !childNode.getChild().isEmpty()
+        && !(childNode.getChild().getFirst() instanceof Values)) {
+      childNode = (UnresolvedPlan) childNode.getChild().getFirst();
```
qianheng-aws, line 261: Will there be more than 1 children in the subquery? Maybe adding assertion and throw exception to avoid this case if we don't support.

yuancu (approved):

PR #4603, opened 2025-10-20
Pushdown sort aggregate metrics

### LINKED ISSUE #4282: [FEATURE] Push down bucket aggregation with sorting by metrics to `multi-terms` aggregation

LantaoJin:
### Description
Pushdown sort aggregate metrics:
- Add `SORT_AGG_METRICS` after `AGGREGATION` in pushdown context
- Convert `composite` bucket agg to `terms/histo/date-histo` agg if sort one metric on `single` bucket
- Convert `composite` bucket agg (all terms source) to `multi-terms` agg if sort one metric on `multiple` buckets

### Related Issues
Resolves #4282 

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
@@ -296,7 +296,7 @@ public class CalciteLogicalIndexScan extends AbstractCalciteIndexScan {
         return null;
       }
       List<String> collationNames = getCollationNames(sort.getCollation().getFieldCollations());
-      if (!hasAggregatorInSortBy(collationNames)) {
+      if (!isAllCollationNamesEqualAggregators(collationNames)) {
         return null;
       }
       AbstractAction<?> newAction =
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
@@ -1113,6 +1113,24 @@ public class CalciteExplainIT extends ExplainIT {
   }
 
   @Test
+  public void testExplainMultipleAggregatorsWithSortOnOneMetricNotPushDown() throws IOException {
+    enabledOnlyWhenPushdownIsEnabled();
+    String expected =
+        loadExpectedPlan("explain_multiple_agg_with_sort_on_one_metric_not_push1.yaml");
+    assertYamlEqualsIgnoreId(
+        expected,
+        explainQueryYaml(
+            "source=opensearch-sql_test_index_account | stats bucket_nullable=false count() as c,"
+                + " sum(balance) as s by state | sort c"));
+    expected = loadExpectedPlan("explain_multiple_agg_with_sort_on_one_metric_not_push2.yaml");
+    assertYamlEqualsIgnoreId(
+        expected,
+        explainQueryYaml(
+            "source=opensearch-sql_test_index_account | stats bucket_nullable=false count() as c,"
+                + " sum(balance) as s by state | sort c, s"));
+  }
+
+  @Test
   public void testExplainEvalMax() throws IOException {
     String expected = loadExpectedPlan("explain_eval_max.json");
     assertJsonEqualsIgnoreId(
--- a/integ-test/src/test/resources/expectedOutput/calcite/explain_multiple_agg_with_sort_on_one_metric_not_push1.yaml
+++ b/integ-test/src/test/resources/expectedOutput/calcite/explain_multiple_agg_with_sort_on_one_metric_not_push1.yaml
@@ -1,0 +1,13 @@
+calcite:
+  logical: |
+    LogicalSystemLimit(sort0=[$0], dir0=[ASC-nulls-first], fetch=[10000], type=[QUERY_SIZE_LIMIT])
+      LogicalSort(sort0=[$0], dir0=[ASC-nulls-first])
+        LogicalProject(c=[$1], s=[$2], state=[$0])
+          LogicalAggregate(group=[{0}], c=[COUNT()], s=[SUM($1)])
+            LogicalProject(state=[$7], balance=[$3])
+              LogicalFilter(condition=[IS NOT NULL($7)])
+                CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])
+  physical: |
+    EnumerableLimit(fetch=[10000])
+      EnumerableSort(sort0=[$0], dir0=[ASC-nulls-first])
+        CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={1},c=COUNT(),s=SUM($0)), PROJECT->[c, s, state]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"composite_buckets":{"composite":{"size":1000,"sources":[{"state":{"terms":{"field":"state.keyword","missing_bucket":false,"order":"asc"}}}]},"aggregations":{"s":{"sum":{"field":"balance"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
--- a/integ-test/src/test/resources/expectedOutput/calcite/explain_multiple_agg_with_sort_on_one_metric_not_push2.yaml
+++ b/integ-test/src/test/resources/expectedOutput/calcite/explain_multiple_agg_with_sort_on_one_metric_not_push2.yaml
@@ -1,0 +1,13 @@
+calcite:
+  logical: |
+    LogicalSystemLimit(sort0=[$0], sort1=[$1], dir0=[ASC-nulls-first], dir1=[ASC-nulls-first], fetch=[10000], type=[QUERY_SIZE_LIMIT])
+      LogicalSort(sort0=[$0], sort1=[$1], dir0=[ASC-nulls-first], dir1=[ASC-nulls-first])
+        LogicalProject(c=[$1], s=[$2], state=[$0])
+          LogicalAggregate(group=[{0}], c=[COUNT()], s=[SUM($1)])
+            LogicalProject(state=[$7], balance=[$3])
+              LogicalFilter(condition=[IS NOT NULL($7)])
+                CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])
+  physical: |
+    EnumerableLimit(fetch=[10000])
+      EnumerableSort(sort0=[$0], sort1=[$1], dir0=[ASC-nulls-first], dir1=[ASC-nulls-first])
+        CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={1},c=COUNT(),s=SUM($0)), PROJECT->[c, s, state]], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"composite_buckets":{"composite":{"size":1000,"sources":[{"state":{"terms":{"field":"state.keyword","missing_bucket":false,"order":"asc"}}}]},"aggregations":{"s":{"sum":{"field":"balance"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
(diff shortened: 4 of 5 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/SortAggregationMetricsRule.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    final LogicalSort sort = call.rel(0);
+    final CalciteLogicalIndexScan scan = call.rel(1);
+    // Only support single metric sort
+    if (sort.getCollation().getFieldCollations().size() != 1) {
+      return;
+    }
```
yuancu: nit: can be moved to `PlanUtils`, then used in rule config as a predicate

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AbstractAction.java
```diff
@@ -0,0 +1,12 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.storage.scan.context;
+
+public interface AbstractAction<T> {
+  void apply(T target);
+
+  void transform(PushDownContext context, PushDownOperation operation);
+}
```
yuancu, line 23: javadoc may help future developers using this interface

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
```diff
@@ -232,6 +69,110 @@ public void apply(OpenSearchRequestBuilder requestBuilder) {
...
+    if (aggregationBuilder.getLeft().isEmpty()) return;
+    AggregationBuilder builder = aggregationBuilder.getLeft().getFirst();
+    if (builder instanceof CompositeAggregationBuilder composite) {
+      String path = getAggregationPath(collations, fieldNames, composite);
+      BucketOrder bucketOrder =
+          collations.get(0).getDirection() == RelFieldCollation.Direction.ASCENDING
+              ? BucketOrder.aggregation(path, true)
+              : BucketOrder.aggregation(path, false);
+
+      if (composite.sources().size() == 1) {
```
yuancu, line 97: [non-blocking] I'm thinking is there a way to construct the aggregation upfront instead of decomposing a composite aggregation later. E.g. create term aggregation separately in aggregate analyzer, then only pushdown sort as bucket order here. This may simplify the implementation.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/SortAggregationMetricsRule.java
LantaoJin: IMO, not worth to move this to util class. This could be changed in following up PRs.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AbstractAction.java
LantaoJin: added

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
LantaoJin: composite aggregation is better than terms aggregation on performance even for single bucket

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/SortAggregationMetricsRule.java
qianheng-aws: Could change to be an anonymous function in predication if not common used.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
```diff
@@ -232,6 +70,110 @@ public void apply(OpenSearchRequestBuilder requestBuilder) {
...
+  private String getAggregationPath(
+      List<RelFieldCollation> collations,
+      List<String> fieldNames,
+      CompositeAggregationBuilder composite) {
+    String path;
+    if (composite.getSubAggregations().isEmpty()) {
+      // count agg optimized, get the path name from field names
+      path = fieldNames.get(collations.get(0).getFieldIndex());
+    } else {
+      path = composite.getSubAggregations().stream().toList().get(0).getName();
```
qianheng-aws: Is `path` refer to the field name of agg call here? Anyway, composite's sub agg could be other bucket builder than only metric builder here after we introduce auto_span or range bucket. It seems incorrect here in that case.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -268,6 +284,37 @@ private RelTraitSet reIndexCollations(List<Integer> selectedColumns) {
     return newTraitSet;
   }
 
+  public CalciteLogicalIndexScan pushDownSortAggregateMetrics(Sort sort) {
```
qianheng-aws, line 287: We can only do this operation when bucket_nullable=false, right? Any place to check this parameter?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
LantaoJin: do you mean composite + auto_span?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
LantaoJin: [AggPushDownAction.L97](https://github.com/opensearch-project/sql/pull/4603/files#diff-f1d75dafda6749c7779bd32c977296f79215fcd034e06d9d1e964461383dbed3R97) and L126

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
LantaoJin: addressed, add more explain tests

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
qianheng-aws: [nit] Will it be better to put these checks in one place if possible? Now the prerequisites are split both here and in `aggAction.pushDownSortAggMetrics`, although they are both `aggregationBuilders` and some checking seems overlapped.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
```diff
@@ -232,6 +70,116 @@ public void apply(OpenSearchRequestBuilder requestBuilder) {
...
+                  Collections.singletonList(histoBuilder),
+                  convertTo(aggregationBuilder.getRight()));
+          return;
+        }
+      } else {
+        if (composite.sources().stream()
+            .allMatch(
+                src -> src instanceof TermsValuesSourceBuilder terms && !terms.missingBucket())) {
+          // multi-term agg
+          return;
```
qianheng-aws, line 174: Shouldn't throw exception here for now? Otherwise the sort is pushed down without any transformation. 

Could you please add a case for multi-terms? The plan should get unchanged since we don't implement muti-term agg now until the following-up PR.

LantaoJin: Will fix the part in a followup PR soon when this PR merged.

LantaoJin: Or I will submit the multi-terms in the same PR later today, what do you think? @qianheng-aws . just don't want make a single PR too large.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
LantaoJin: Cannot get the bucket_nullable information here. Aggregation has already pushed down. So [AggPushDownAction.L97](https://github.com/opensearch-project/sql/pull/4603/files#diff-f1d75dafda6749c7779bd32c977296f79215fcd034e06d9d1e964461383dbed3R97) and L126 would be better IMO.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
LantaoJin: Submitted a new commit and updated the PR description. Now both single bucket or multiple buckets are supported.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/SortAggregationMetricsRule.java
LantaoJin: fixed.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
qianheng-aws: But I see you've already get the pushed aggregation builder here.
```
 List<AggregationBuilder> aggregationBuilders =
          pushDownContext.getAggPushDownAction().getAggregationBuilder().getLeft();
```

Or otherwise L290-310 could all be moved into `aggAction.pushDownSortAggMetrics`.

It's also fine to me if not making change since the logic is correct anyway.

LantaoJin: I see, but I don't want always to check CompositeValuesSourceBuilder.missingBucket() here since not all ValuesSourceBuilders are impacted by `bucket_nullable`, for example date histogram. So moving to here has to check types, sounds not worth to check types twice.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
```diff
@@ -0,0 +1,322 @@
...
+      List<RelFieldCollation> collations,
+      List<String> fieldNames,
+      CompositeAggregationBuilder composite) {
+    String path;
+    AggregationBuilder metric = composite.getSubAggregations().stream().findFirst().orElse(null);
+    if (metric == null) {
+      // count agg optimized, get the path name from field names
+      path = fieldNames.get(collations.get(0).getFieldIndex());
+    } else if (metric instanceof ValuesSourceAggregationBuilder.LeafOnly) {
+      path = metric.getName();
```
qianheng-aws, line 192: [nit, non-blocking] Could this be combined with above branch? i.e. Does `path = fieldNames.get(collations.get(0).getFieldIndex());` also works for `metric instanceof ValuesSourceAggregationBuilder.LeafOnly`?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
```diff
@@ -0,0 +1,322 @@
...
+          "Cannot pushdown sort aggregate metrics");
+    }
+  }
+
+  private String getAggregationPath(
+      List<RelFieldCollation> collations,
+      List<String> fieldNames,
+      CompositeAggregationBuilder composite) {
+    String path;
+    AggregationBuilder metric = composite.getSubAggregations().stream().findFirst().orElse(null);
```
qianheng-aws, line 187: Do we only support only 1 agg metric for this enhancement? And anywhere to check this?

LantaoJin: > Do we only support only 1 agg metric for this enhancement? And anywhere to check this?

https://github.com/opensearch-project/sql/pull/4603/files#diff-b4a250f8c0175f824789fb6cd403e69f4de986c912461daf766d39442a7f530bR40

LantaoJin: good catching. added restricted checking.

PR #4606, opened 2025-10-20
Fix bin nested fields issue

### LINKED ISSUE #4482: [BUG] PPL `bin` command fails to produce output for nested/struct fields
## Query Information
**PPL Command/Query:**
```ppl
source=kube-state-metrics | bin `monitoring.metrics.beat.handles.open` bins=5
source=kube-state-metrics | bin `monitoring.metrics.beat.cpu.system.ticks` span=1000
```

**Expected Result:**
The `bin` command should create binned ranges for the nested field values and include them in the output, similar to how it works with flat (non-nested) fields. The binned values should appear in the result set either as a new column or by replacing the original field value.

**Actual Result:**
When using the `bin` command on nested/struct fields without an explicit `fields` projection, the binned values are not included in the output. The query returns the original nested field values unchanged, making it appear as if the `bin` command is being ignored entirely.

However, when adding `| fields <nested.field>` after the bin command, the binned values do appear correctly.

## Dataset Information
**Dataset/Schema Type**
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "monitoring": {
        "properties": {
          "metrics": {
            "properties": {
              "beat": {
                "properties": {
                  "handles": {
                    "properties": {
                      "open": { "type": "long" }
                    }
                  },
                  "cpu": {
                    "properties": {
                      "system": {
                        "properties": {
                          "ticks": { "type": "long" }
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

**Sample Data**
```json
{
  "monitoring": {
    "metrics": {
      "beat": {
        "handles": {
          "open": 66
        },
        "cpu": {
          "system": {
            "ticks": 11403
          }
        }
      }
    }
  }
}
```

## Bug Description
**Issue Summary:**
The PPL `bin` command fails to produce binned output for nested/struct fields when used without an explicit `fields` projection. The binned expression is not properly integrated into the query result, causing the original field values to be returned instead of the binned ranges.

**Steps to Reproduce:**
1. Create an index with nested field structure (e.g., `nested.value` where `nested` is an object with a `value` field)
2. Insert documents with numeric values in the nested field
3. Run a PPL query: `source=<index> | bin nested.value span=10`
4. Observe that the output contains the original `nested.value` values, not binned ranges
5. Compare with flat field: `source=<index> | bin flat_value span=10` - this works correctly and shows binned ranges

**Comparison:**

*Flat field (WORKS):*
```ppl
source=test-bin-nested | bin flat_value span=10
```
Result includes binned column: `["0-10", "10-20", "20-30", ...]`

*Nested field (BROKEN):*
```ppl
source=test-bin-nested | bin nested.value span=10
```
Result shows original values: `[5, 15, 25, ...]` - no binned output

*Nested field with workaround (WORKS):*
```ppl
source=test-bin-nested | bin nested.value span=10 | fields nested.value
```
Result shows binned values: `["0-10", "10-20", "20-30", ...]`

**Impact:**
This bug affects users working with structured datasets that use nested field mappings. The `bin` command is essential for creating histograms and time-series visualizations, and this limitation forces users to either:
- Flatten their data structure (losing semantic organization)
- Use the workaround of explicitly projecting fields (which is non-intuitive and undocumented)
- Avoid using `bin` with nested fields entirely

## Environment Information
**OpenSearch Version:**
OpenSearch 3.3.0-SNAPSHOT (reproduced on development build)

**Additional Details:**
The issue was confirmed through execution plan analysis using `POST _plugins/_ppl/_explain`:
- For flat fields: The logical plan correctly includes `SPAN_BUCKET($field, span)` function
- For nested fields: The logical plan is missing the `SPAN_BUCKET` function entirely, showing only the original field projection

The root cause appears to be in the `projectPlusOverriding` method in `CalciteRelNodeVisitor.java` (lines 844-867). This method is responsible for adding the binned expression to the query plan while handling field name conflicts.

The issue occurs at lines 847-851:
```java
List<String> originalFieldNames = context.relBuilder.peek().getRowType().getFieldNames();
List<RexNode> toOverrideList =
    originalFieldNames.stream()
        .filter(newNames::contains)
        .map(a -> (RexNode) context.relBuilder.field(a))
        .toList();
```

For nested fields like `nested.value`:
- The `originalFieldNames` list contains only top-level field names (e.g., `["flat_value", "nested"]`)
- The `newNames` list contains the full nested path (e.g., `["nested.value"]`)
- The filter `newNames::contains` never matches because `"nested"` ≠ `"nested.value"`
- As a result, `toOverrideList` is empty, and the binned expression is added as a new field but never properly integrated into the output

This explains why:
1. The bin command appears to do nothing for nested fields (the binned expression exists but isn't visible)
2. Using `| fields nested.value` works (it explicitly projects the binned expression)
3. Flat fields work correctly (the field name matches exactly)

## Tentative Proposed Fix

The fix should modify the `projectPlusOverriding` method to handle nested field names correctly. One approach:

1. **Extract the top-level field name** from nested paths (e.g., `"nested.value"` → `"nested"`)
2. **Check if the top-level field exists** in `originalFieldNames`
3. **If it exists, mark it for override** to ensure the binned expression replaces the nested field value

Alternatively, the field resolution logic in `BinUtils.extractFieldName()` could be enhanced to work with Calcite's field access patterns for nested structures.

A more robust solution might involve:
- Modifying how nested fields are represented in the Calcite plan
- Ensuring that `context.relBuilder.field(a)` can properly resolve nested field paths
- Updating the rename logic to handle dotted field names

The fix should be validated against:
- Simple nested fields (one level: `nested.value`)
- Deeply nested fields (multiple levels: `a.b.c.value`)
- Mixed scenarios (some flat, some nested fields in the same query)
- All bin variants (span, bins, minspan, range)

## Workaround

Add an explicit `fields` projection after the `bin` command:

```ppl
source=kube-state-metrics | bin `monitoring.metrics.beat.handles.open` bins=5 | fields `monitoring.metrics.beat.handles.open`
```

This forces the binned expression to be projected into the output, making the binned ranges visible.

ahkcs:
### Description
Fixed bin command failing on nested fields (e.g., resource.attributes.telemetry.sdk.version). 
Updated `CalciteRelNodeVisitor.projectPlusOverriding()` to use prefix matching instead of exact matching for nested field names.
### Related Issues
* Resolve #4482

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteBinCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteBinCommandIT.java
@@ -1037,4 +1037,33 @@ public class CalciteBinCommandIT extends PPLIntegTestCase {
     verifyDataRows(
         result, rows("10-12"), rows("10-12"), rows("12-14"), rows("12-14"), rows("14-16"));
   }
+
+  @Test
+  public void testBinWithEvalCreatedDottedFieldName() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | eval `resource.temp` = 1 | bin"
+                    + " `resource.attributes.telemetry.sdk.version` span=2 | sort"
+                    + " `resource.attributes.telemetry.sdk.version`",
+                TEST_INDEX_TELEMETRY));
+
+    verifySchema(
+        result,
+        schema("resource.attributes.telemetry.sdk.enabled", null, "boolean"),
+        schema("resource.attributes.telemetry.sdk.language", null, "string"),
+        schema("resource.attributes.telemetry.sdk.name", null, "string"),
+        schema("resource.temp", null, "int"),
+        schema("severityNumber", null, "int"),
+        schema("resource.attributes.telemetry.sdk.version", null, "string"));
+
+    // Data column order: enabled, language, name, severityNumber, resource.temp, version
+    verifyDataRows(
+        result,
+        rows(true, "java", "opentelemetry", 9, 1, "10-12"),
+        rows(false, "python", "opentelemetry", 12, 1, "10-12"),
+        rows(true, "javascript", "opentelemetry", 9, 1, "12-14"),
+        rows(false, "go", "opentelemetry", 16, 1, "12-14"),
+        rows(true, "rust", "opentelemetry", 12, 1, "14-16"));
+  }
 }


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -850,7 +850,16 @@ private void projectPlusOverriding(
     List<String> originalFieldNames = context.relBuilder.peek().getRowType().getFieldNames();
     List<RexNode> toOverrideList =
         originalFieldNames.stream()
-            .filter(newNames::contains)
+            .filter(
+                originalName ->
```
ykmr1224: Let's extract for readability / maintainability.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -850,7 +850,16 @@ private void projectPlusOverriding(
...
+                    newNames.stream()
+                        .anyMatch(
+                            newName ->
+                                // Match exact field names (e.g., "age" == "age")
+                                // OR nested paths (e.g., "resource.attributes..." starts with
+                                // "resource")
+                                newName.equals(originalName)
+                                    || newName.startsWith(originalName + ".")))
```
ykmr1224: Does this mean it could override all the original fields start with `originalName + "."`? (like when original fields include `resource.a`, `resource.b`, etc. then all of them will be removed?

ahkcs: Thanks for the question! No - originalFieldNames only contains top-level field names like "resource", not nested paths like "resource.a" or "resource.b".

So when binning resource.c.nested, the filter only matches the top-level "resource" field (via startsWith("resource.")). This causes the entire resource struct to be removed and flattened, with the binned field replacing the original nested value.

Only one top-level struct gets matched and processed, regardless of how many nested fields it contains.

You can refer to this IT test I added: `testBinWithNestedFieldWithoutExplicitProjection`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Extracted

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: Field name could contain `.` and `newName.startsWith("resource.")` could match with multiple fields in my understanding.
I am unsure how much we currently support field name containing `.`, but [`QualifiedNameResolver`](https://github.com/opensearch-project/sql/blob/c30d5d01c36b7c959d7b2a290c319c4a8f0e8367/core/src/main/java/org/opensearch/sql/calcite/QualifiedNameResolver.java#L105-L114) implements longest match logic to decide the referred field in case qualified name contains multiple dots.

ahkcs: I think it would be an unusual case for field name to contain mutliple `.`?
Looking at the end-to-end flow, originalFieldNames comes from:
context.relBuilder.peek().getRowType().getFieldNames()

This returns only the direct fields of the current RelNode's row type. For a schema with nested structures, the row type would have:
  - Top-level field: "resource" (type: STRUCT)
  - Top-level field: "severityNumber" (type: INT)

It would not contain "resource.a", "resource.b", etc. as separate top-level fields - those only exist as nested fields within the resource struct definition.

The scenario where multiple top-level fields share a prefix (like resource, resource.a, resource.b all being top-level) would require an unusual schema design(I think we can rarely meet this?)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -870,6 +870,17 @@ private void projectPlusOverriding(
...
 
+  private boolean shouldOverrideField(String originalName, List<String> newNames) {
+    return newNames.stream()
+        .anyMatch(
+            newName ->
+                // Match exact field names (e.g., "age" == "age") for flat fields
+                newName.equals(originalName)
+                    // OR match nested paths (e.g., "resource.attributes..." starts with
+                    // "resource.")
+                    || newName.startsWith(originalName + "."));
```
RyanL1997, line 881: is there any other place using this pattern of detection? Or if there is a better way to detect if it is a nested field or not?

ahkcs: In CalciteRelNodeVisitor.java, the visitFlatten() method uses the same startsWith() pattern:
```
  public RelNode visitFlatten(Flatten node, CalcitePlanContext context) {
    visitChildren(node, context);
    RelBuilder relBuilder = context.relBuilder;
    String fieldName = node.getField().getField().toString();
    // Match the sub-field names with "field.*"
    List<RelDataTypeField> fieldsToExpand =
        relBuilder.peek().getRowType().getFieldList().stream()
            .filter(f -> f.getName().startsWith(fieldName + "."))  // ← Same pattern!
            .toList();
    // ... rest of method
  }
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: OpenSearch mapping key could include `.`, and also we can easily introduce field which contains `.` with `eval` command.
Why can't we directly match with the actual field we used?
`projectPlusOverriding` is used for multiple commands, and could have side effect to other commands.

ahkcs: >Why can't we directly match with the actual field we used?

We DO directly match. The logic uses both:
  1. Exact match: newName.equals(originalName)
  2. Struct match: newName.startsWith(originalName + ".")

 This exact logic is already used by eval. Example:
  source=telemetry | eval `resource.temp` = 1 | bin `resource.attributes.telemetry.sdk.version` span=2

  Step 1 - Eval:
  - "resource.temp".startsWith("resource.") → true → flattens struct
  - Schema becomes: ["resource.attributes.telemetry.sdk.enabled", "resource.attributes.telemetry.sdk.language", ..., "resource.temp"]

  Step 2 - Bin:
  - Schema already flattened, exact match works: "resource.attributes.telemetry.sdk.version".equals("resource.attributes.telemetry.sdk.version") → true

Both commands use the same logic safely. Added test (testBinWithEvalCreatedDottedFieldName) to confirm no edge case issues

yuancu: I think @ykmr1224 means if we can directly know the sub field of a struct and project it, instead of relying on the crispy name matching. E.g. if a struct is `info {a: string, b: int}`, we can directly project `info, info.a, info.b` instead of relying on pattern `info.*` to match the names, because there may be another schema `info: string, info.a integer`, where `info.a` is not a sub-field of `info`.

ahkcs: @yuancu Thanks for the clarification! However, I think the current logic handles this situation, you can refer to this newly added integration test using eval command to test: `testBinWithEvalCreatedDottedFieldName`

For the edge case you mentioned, I think it would be problematic if we have two `info.a` with different type because that would cause confusion for the commands. For example, when we are using stats count(info.a), which `info.a` is it referring to? For other cases where there's no naming overlap, I think you can refer to the eval test case

PR #4611, opened 2025-10-21
Add permissive mode setting

ykmr1224:
### Description
- Add permissive mode setting, and enable **only in standalone integ-test**
  - When permissive mode is enabled, the fields not found in the index mapping will be collected and stored in `_MAP` field typed as `MAP<String, Object>` (dynamic fields).
- This PR adds only most primitive support to handle dynamic fields.
  - I will raise separate PRs to provide wider support.

### Related Issues

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLIntegTestCase.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLIntegTestCase.java
@@ -252,7 +252,7 @@ public abstract class CalcitePPLIntegTestCase extends PPLIntegTestCase {
       @Override
       public void onResponse(T response) {
         String responseContent = formatter.format(response);
-        actual.set(responseContent.replace("\\r\\n", "\\n"));
+        actual.set(responseContent.replace("\r\n", "\n"));
       }
 
       @Override


core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -181,7 +181,7 @@ public static RelDataType convertExprTypeToRelDataType(ExprType fieldType, boole
           // https://github.com/opensearch-project/sql/issues/3459
           final RelDataType relKey = TYPE_FACTORY.createSqlType(SqlTypeName.VARCHAR);
           return TYPE_FACTORY.createMapType(
-              relKey, TYPE_FACTORY.createSqlType(SqlTypeName.BINARY), nullable);
+              relKey, TYPE_FACTORY.createSqlType(SqlTypeName.ANY), nullable);
```
penghuo, line 184: Does https://github.com/opensearch-project/sql/issues/3459 solved?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsCommandIT.java
```diff
@@ -0,0 +1,98 @@
...
+        executeQuery(
+            source(TEST_INDEX_DYNAMIC, "fields firstname, lastname, department, salary | head 1"));
+
+    verifySchema(
+        result,
+        schema("firstname", "string"),
+        schema("lastname", "string"),
+        schema("department", "string"),
+        schema("salary", "int"));
+    verifyDataRows(result, rows("John", "Doe", "Engineering", 75000));
```
penghuo, line 59: What is explain results?

core/src/main/java/org/opensearch/sql/storage/StorageEngine.java
```diff
@@ -16,6 +16,12 @@ public interface StorageEngine {
   /** Get {@link Table} from storage engine. */
   Table getTable(DataSourceSchemaName dataSourceSchemaName, String tableName);
 
+  /** Get {@link Table} which allows permissive mode or not. */
+  default Table getPermissiveAwareTable(
```
penghuo, line 20: getPermissiveAwareTable is temporay change? if yes, add TODO.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
```diff
@@ -107,9 +111,29 @@ public Object current() {
...
+      return collectDynamicFields(value);
+    }
     return ExprValueUtils.resolveRefPaths(value, List.of(rawPath.split("\\."))).valueForCalcite();
   }
 
+  private Object collectDynamicFields(ExprValue value) {
+    Map<String, Object> dynamicFields = new HashMap<>();
+
+    try {
+      for (Map.Entry<String, ExprValue> entry : value.tupleValue().entrySet()) {
```
penghuo, line 124: only the top level <k, v> are extracted? what if field name is field.f1.f2?

core/src/main/java/org/opensearch/sql/storage/StorageEngine.java
ykmr1224: This will be kept to avoid permissive is applied to SQL.
(Only Calcite PPL implementation will use this method to allow permissive mode)

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
ykmr1224: Hmm, do we want to flatten nested data?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsCommandIT.java
ykmr1224: Add explain validation in the tests.

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
ykmr1224: #3459 is not solved yet. I think #3459 is not directly related to this type mapping.

This change is needed to set `_MAP` as Map<String, Any>.
It seems that currently object is converted to array of map when read like: `[{"attr1": 1, "attr2": "v"}], "otherValue"`.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/PermissiveOpenSearchIndex.java
```diff
@@ -0,0 +1,28 @@
...
+public class PermissiveOpenSearchIndex extends OpenSearchIndex {
+
+  public PermissiveOpenSearchIndex(OpenSearchClient client, Settings settings, String indexName) {
+    super(client, settings, indexName);
+  }
+
+  @Override
+  public Map<String, ExprType> getReservedFieldTypes() {
+    Map<String, ExprType> extendedMap = new LinkedHashMap<>(METADATAFIELD_TYPE_MAP);
+    extendedMap.put(DynamicFieldsConstants.DYNAMIC_FIELDS_MAP, ExprCoreType.STRUCT);
```
dai-chen, line 25: Is this line the only reason to add this new class or more future changes? Because I see `OpenSearchIndex` already has `Settings`.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLPermissiveIntegTestCase.java
```diff
@@ -0,0 +1,52 @@
...
+import org.opensearch.sql.common.setting.Settings;
+import org.opensearch.sql.common.setting.Settings.Key;
+import org.opensearch.sql.datasource.model.DataSource;
+import org.opensearch.sql.datasource.model.DataSourceMetadata;
+import org.opensearch.sql.datasource.model.DataSourceType;
+import org.opensearch.sql.opensearch.client.OpenSearchClient;
+import org.opensearch.sql.opensearch.storage.OpenSearchDataSourceFactory;
+import org.opensearch.sql.opensearch.storage.OpenSearchStorageEngine;
+
+public abstract class CalcitePPLPermissiveIntegTestCase extends CalcitePPLIntegTestCase {
```
dai-chen, line 17: Instead of a base class, just wondering should we have a test suite similar as `CalciteNoPushdownIT` to run all IT w/wo permissive mode?

core/src/main/java/org/opensearch/sql/calcite/plan/DynamicFieldsConstants.java
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.plan;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class DynamicFieldsConstants {
+  public static String DYNAMIC_FIELDS_MAP = "_MAP";
```
dai-chen, line 12: np: any more constants later?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsCommandIT.java
```diff
@@ -0,0 +1,133 @@
...
+import static org.opensearch.sql.util.MatcherUtils.schema;
+import static org.opensearch.sql.util.MatcherUtils.verifyDataRows;
+import static org.opensearch.sql.util.MatcherUtils.verifySchema;
+
+import java.io.IOException;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+import org.opensearch.client.Request;
+
+public class CalciteDynamicFieldsCommandIT extends CalcitePPLPermissiveIntegTestCase {
```
dai-chen, line 20: Could you add test/assertion for SparkSQL now or later? Just want to make sure this feature is portable to other engine.

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
```diff
@@ -35,6 +35,7 @@ public enum Key {
     PPL_SYNTAX_LEGACY_PREFERRED("plugins.ppl.syntax.legacy.preferred"),
     PPL_SUBSEARCH_MAXOUT("plugins.ppl.subsearch.maxout"),
     PPL_JOIN_SUBSEARCH_MAXOUT("plugins.ppl.join.subsearch_maxout"),
+    PPL_QUERY_PERMISSIVE("plugins.ppl.query.permissive"),
```
Swiddis, line 38: thought: I'd like if this could be configured per-query instead of cluster-global

If 99.9% of my queries are static, I don't want to slow all of them down for one dynamic query

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -301,6 +301,9 @@ public static ExprValue getExprValueByExprType(ExprType type, Object value) {
         return ExprValueUtils.collectionValue((List<Object>) value);
       case STRUCT:
         return ExprValueUtils.tupleValue((Map<String, Object>) value);
+      case UNKNOWN:
```
dai-chen, line 304: Is this change already covered by test?

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
dai-chen: Previously it's `Map<String, Binary>`? No test change needed for this?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -404,7 +404,12 @@ private List<RexNode> expandProjectFields(
             }
             matchingFields.forEach(f -> expandedFields.add(context.relBuilder.field(f)));
           } else if (addedFields.add(fieldName)) {
-            expandedFields.add(rexVisitor.analyze(field, context));
+            RexNode analyzed = rexVisitor.analyze(field, context);
```
dai-chen, line 407: Is this change related to some other issue or required by dynamic field?

ykmr1224: This is needed for dynamic fields. Dynamic fields require alias to expose the item name as field name into result record schema.

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
ykmr1224: Can you give your inputs here? https://github.com/opensearch-project/sql/issues/4349
I think we can introduce per query or per index configuration as needed later.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/PermissiveOpenSearchIndex.java
ykmr1224: I made this class to avoid the setting is applied to SQL query as well, since OpenSearchIndex is shared.
Once we enable permissive mode to SQL (and deprecate legacy engines), we can unify it.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLPermissiveIntegTestCase.java
ykmr1224: Yes, we will definitely need them, but for now many tests will fail since this change cover very limited case. (Once we establish good foundation, I will introduce such IT)

core/src/main/java/org/opensearch/sql/calcite/plan/DynamicFieldsConstants.java
ykmr1224: I don't expect other constant. (I didn't find good place to put this, and made this class)

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
ykmr1224: Yes, it is covered in the test when dynamic field is selected in the result.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsCommandIT.java
ykmr1224: Sure, let me add in the following PRs.

ykmr1224:
> After introducing map functions, we're beginning to touch the "core" logic. Shall we consider using feature branch since there is no `.x` branch today?

Very good point. I was on the fence as it could lead to huge conflict when merging.
But made feature branch and switched the merge destination.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/PermissiveOpenSearchIndex.java
dai-chen: I see. Our SQL and PPL shared both the index class and cluster setting.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLPermissiveIntegTestCase.java
dai-chen: Np. But I'm thinking only add IT that can pass to suite classes list.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
dai-chen: We may add a IT to clarify this either a limitation or by design?

Swiddis (approved):

Issue #4578, opened 2025-10-16
[BUG] PPL `bin` command fails with auto date span


yuancu:
**What is the bug?**

When performing `bin` command on a timestamp field with `bins` parameter specified, it fails to prepare the plan.

**How can one reproduce the bug?**

1. disable pushdown

```
PUT _cluster/settings
{
   "persistent":{
      "plugins.calcite.enabled": true,
      "plugins.calcite.fallback.allowed": false,
      "plugins.calcite.pushdown.enabled": true
   }
}
```

2. execute the query
```
POST /_plugins/_ppl
{
  "query" : "source=time_test |  bin @timestamp bins=2"
}
```

The `time_test` index can be substituted with any index that contains a `@timestamp` field. It throws the following error:

```json
{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "java.sql.SQLException: Error while preparing plan [LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n  LogicalProject(category=[$1], value=[$2], timestamp=[$3], @timestamp=[WIDTH_BUCKET($0, 2, -(MAX($0) OVER (), MIN($0) OVER ()), MAX($0) OVER ())])\n    CalciteLogicalIndexScan(table=[[OpenSearch, time_test]])\n]",
    "type": "RuntimeException"
  },
  "status": 500
}
```

<details>
<summary> complete stacktrace </summary>

```
[2025-10-16T10:14:29,805][ERROR][o.o.s.p.r.RestPPLQueryAction] [7cf34de73b85] Error happened during query handling
java.lang.RuntimeException: java.sql.SQLException: Error while preparing plan [LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(category=[$1], value=[$2], timestamp=[$3], @timestamp=[WIDTH_BUCKET($0, 2, -(MAX($0) OVER (), MIN($0) OVER ()), MAX($0) OVER ())])
    CalciteLogicalIndexScan(table=[[OpenSearch, time_test]])
]
        at org.apache.calcite.util.Util.throwAsRuntime(Util.java:956) ~[?:?]
        at org.opensearch.sql.calcite.utils.CalciteToolsHelper$OpenSearchRelRunners.run(CalciteToolsHelper.java:358) ~[?:?]
[...]

PR #4612, opened 2025-10-21
Fix `bins` on time-related fields

ahkcs:
### Description

---

**Summary**

This PR extends the `bin` command to support time-related fields (`timestamp`, `date`, `time`) in non-aggregation scenarios using the `WIDTH_BUCKET` UDF.
Previously, `bins` on time fields only worked with aggregations via `auto_date_histogram` pushdown.
Now, users can apply binning on timestamp fields in projection queries without aggregation.

### Related Issues
Resolves #https://github.com/opensearch-project/sql/issues/4578


core/src/main/java/org/opensearch/sql/calcite/utils/binning/handlers/CountBinHandler.java
```diff
@@ -30,20 +29,21 @@ public RexNode createExpression(
       requestedBins = BinConstants.DEFAULT_BINS;
     }
 
-    // Calculate data range using window functions
-    RexNode minValue = context.relBuilder.min(fieldExpr).over().toRex();
+    // Calculate MIN and MAX using window functions
     RexNode maxValue = context.relBuilder.max(fieldExpr).over().toRex();
-    RexNode dataRange = context.relBuilder.call(SqlStdOperatorTable.MINUS, maxValue, minValue);
+    RexNode minValue = context.relBuilder.min(fieldExpr).over().toRex();
```
ykmr1224: nitnit: more intuitive to start with minValue

core/src/main/java/org/opensearch/sql/expression/function/udf/binning/WidthBucketFunction.java
```diff
@@ -76,35 +80,76 @@ public Expression implement(
...
           Expressions.convert_(numBins, Number.class),
-          Expressions.convert_(dataRange, Number.class),
-          Expressions.convert_(maxValue, Number.class));
+          minValueExpr,
+          maxValueExpr,
+          isTimestamp);
     }
 
     /** Width bucket calculation using nice number algorithm. */
     public static String calculateWidthBucket(
```
ykmr1224, line 111: I think it is cleaner to have separate method for timestamp.

integ-test/src/test/resources/expectedOutput/calcite/explain_bin_bins.yaml
```diff
@@ -0,0 +1 @@
...stname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], email=[$9], lastname=[$10], _id=[$11], _index=[$12], _score=[$13], _maxscore=[$14], _sort=[$15], _routing=[$16], age=[WIDTH_BUCKET($8, 3, MIN($8) OVER (), MAX($8) OVER ())])\n        CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n","physical":"EnumerableLimit(fetch=[10000])\n  EnumerableCalc(expr#0..12=[{inputs}], expr#13=[3], expr#14=[WIDTH_BUCKET($t8, $t13, $t11, $t12)], proj#0..7=[{exprs}], email=[$t9], lastname=[$t10], age=[$t14])\n    EnumerableLimit(fetch=[5])\n      EnumerableWindow(window#0=[window(aggs [MIN($8), MAX($8)])])\n        CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname]], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"_source\":{\"includes\":[\"account_number\",\"firstname\",\"address\",\"balance\",\"gender\",\"city\",\"employer\",\"state\",\"age\",\"email\",\"lastname\"],\"excludes\":[]}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])\n"}}
```
ykmr1224: It looks like JSON file instead of YAML.

core/src/main/java/org/opensearch/sql/expression/function/udf/binning/WidthBucketFunction.java
```diff
@@ -152,6 +197,93 @@ private static double calculateOptimalWidth(
...
+    private static long convertTimestampToMillis(Object timestamp) {
+      if (timestamp instanceof Number) {
+        return ((Number) timestamp).longValue();
+      } else if (timestamp instanceof String) {
+        // Parse timestamp string "yyyy-MM-dd HH:mm:ss" to milliseconds
+        // Use LocalDateTime to parse without timezone, then convert to UTC
+        String timestampStr = (String) timestamp;
+        java.time.LocalDateTime localDateTime =
+            java.time.LocalDateTime.parse(
+                timestampStr, DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"));
```
ykmr1224: Why this pattern is different from something used in time_bins_test_data.json file?

And I was not quite sure if we can always assume this format.

core/src/main/java/org/opensearch/sql/expression/function/udf/binning/WidthBucketFunction.java
```diff
@@ -152,6 +197,93 @@ private static double calculateOptimalWidth(
...
+
+      // If target is larger than all intervals, use the largest
+      return intervals[intervals.length - 1];
+    }
+
+    /** Format timestamp in milliseconds as ISO 8601 string. Format: "yyyy-MM-dd HH:mm:ss" */
+    private static String formatTimestamp(long timestampMillis) {
+      Instant instant = Instant.ofEpochMilli(timestampMillis);
+      ZonedDateTime zdt = instant.atZone(ZoneOffset.UTC);
+      return zdt.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"));
```
ykmr1224: Should we make DateTimeFormatter instance a constant? (it is used above as well)

integ-test/src/test/resources/expectedOutput/calcite/explain_bin_bins.yaml
ahkcs: Updated

core/src/main/java/org/opensearch/sql/expression/function/udf/binning/WidthBucketFunction.java
ahkcs: Separated

core/src/main/java/org/opensearch/sql/expression/function/udf/binning/WidthBucketFunction.java
ahkcs: The test data uses ISO 8601 format (2025-07-28T10:00:00 with 'T') for ingestion into OpenSearch, but when Calcite retrieves timestamp values and converts them to strings, it uses the SQL literal format (yyyy-MM-dd HH:mm:ss with space).

core/src/main/java/org/opensearch/sql/expression/function/udf/binning/WidthBucketFunction.java
ahkcs: Great suggestion!
I updated the following:
For parsing (input) - uses DATE_TIMESTAMP_FORMATTER
For formatting (output) - uses SQL_LITERAL_DATE_TIME_FORMAT

core/src/main/java/org/opensearch/sql/calcite/utils/binning/handlers/CountBinHandler.java
ahkcs: switched order

Issue #4466, opened 2025-10-09
[BUG] `rex` command off-by-one error with nested capture groups in named groups


alexey-temnikov:
## Query Information
**PPL Command/Query:**
```ppl
source=opensearch_dashboards_sample_data_logs
| rex field=message '.+"(?<RestMethod>(GET|POST)) .+?(?<HttpVersion>HTTP/[0-9\.]+)" (?<HttpStatus>\d+) (?<BytesSent>.+?) .+'
| fields RestMethod, HttpVersion, HttpStatus, BytesSent, message
| head 1
```

**Expected Result:**
```json
{
  "datarows": [
    [
      "GET",
      "HTTP/1.1",
      "200",
      "6219",
      "223.87.60.27 - - [2018-07-22T00:39:02.912Z] \"GET /opensearch/opensearch-1.0.0.deb_1 HTTP/1.1\" 200 6219 \"-\" \"Mozilla/5.0 (X11; Linux x86_64; rv:6.0a1) Gecko/20110421 Firefox/6.0a1\""
    ]
  ]
}
```

**Actual Result:**
```json
{
  "datarows": [
    [
      "GET",
      "GET",
      "HTTP/1.1",
      "200",
      "223.87.60.27 - - [2018-07-22T00:39:02.912Z] \"GET /opensearch/opensearch-1.0.0.deb_1 HTTP/1.1\" 200 6219 \"-\" \"Mozilla/5.0 (X11; Linux x86_64; rv:6.0a1) Gecko/20110421 Firefox/6.0a1\""
    ]
  ]
}
```

The values are shifted by one position:
- `RestMethod` column shows "GET" (correct value, but appears twice)
- `HttpVersion` column shows "GET" (should be "HTTP/1.1")
- `HttpStatus` column shows "HTTP/1.1" (should be "200")
- `BytesSent` column shows "200" (should be "6219")

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [ ] Simple Schema for Observability (SS4O)
- [ ] Open Cybersecurity Schema Framework (OCSF)
- [x] Custom (OpenSearch Dashboards sample data)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "message": {
        "type": "text",
        "fields": {
          "keyword": {
            "type": "keyword",
            "ignore_above": 256
          }
        }
      },
      "timestamp": {
        "type": "date"
      },
      "response": {
        "type": "text"
      },
[...]

PR #4641, opened 2025-10-22
[BugFix] Fix unexpected shift of extraction for `rex` with nested capture groups in named groups 

RyanL1997:
### Description
Fix unexpected shift of extraction for `rex` with nested capture groups in named groups 

The rex command in PPL had a critical bug when using named capture groups that contained nested unnamed groups. This caused extracted field values to shift by one position, producing incorrect results.
- **Root Cause:** Code used sequential indices `1`, `2`, `3`... but nested groups create non-sequential indices `1`, `3`, `5`...
- **Solution:** Bypass index calculation entirely by using Java's native named group extraction (`matcher.group(groupName)`)

### Example of the Bug

Query:
```bash
    -d '{
      "query": "source=accounts | rex field=email \"(?<user>(amber|hattie|nanette)[a-z]*)@(?<domain>(pyrami|netagy|quility))\\.(?<tld>(com|org))\" | fields user, domain, tld | head 1"
    }'
```

**Expected Result (correct):**
```bash
["amberduke", "pyrami", "com"]
```

**Actual Result (wrong):**
```bash
["amberduke", "amber", "pyrami"]

```

### Root Cause

When Java's regex engine processes the pattern `(?<user>(amber|hattie))[a-z]*`, it assigns group numbers to ALL capture groups (named and unnamed):

Pattern: `(?<user>(amber|hattie))[a-z]*@(?<domain>(pyrami|netagy))\.(?<tld>(com|org))`

Group Assignment:
- Group 0: Entire match
- Group 1: `(?<user>...)`     ← Named group "user"
- Group 2: `(amber|hattie)`   ← Unnamed nested group
- Group 3: `(?<domain>...)`   ← Named group "domain" 
- Group 4: `(pyrami|netagy)`  ← Unnamed nested group
- Group 5: `(?<tld>...)`      ← Named group "tld"
- Group 6: `(com|org)`        ← Unnamed nested group

Before the fix, the bug is in [`CalciteRelNodeVisitor.java` (lines 265-321)](https://github.com/opensearch-project/sql/blob/c6a3a70ac0d0e3ae3aa5eaefba88dab000bc70d5/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java#L265-L321). The code does:
```java
  List<String> namedGroups = RegexCommonUtils.getNamedGroupCandidates(patternStr);
  // namedGroups = ["user", "domain", "tld"]

  for (int i = 0; i < namedGroups.size(); i++) {
      extractCall = PPLFuncImpTable.INSTANCE.resolve(
          context.rexBuilder,
          BuiltinFunctionName.REX_EXTRACT,
          fieldRex,
          context.rexBuilder.makeLiteral(patternStr),
          context.relBuilder.literal(i + 1));  // ← WRONG: Assumes sequential named groups
      // ...
  }
```
The code assumes named groups are at indices `1`, `2`, `3`, ... but the actual indices are `1`, `3`, `5`, ... due to the unnamed nested groups.

With the above buggy logic:
- `REX_EXTRACT(field, pattern, 1)` → Gets Group 1 `(?<user>...)` = "amberduke" → CORRECT
- `REX_EXTRACT(field, pattern, 2)` → Gets Group 2 `(amber|hattie)` = "amber" → WRONG
- `REX_EXTRACT(field, pattern, 3)` → Gets Group 3 `(?<domain>...)` = "pyrami" → WRONG

The second and third extractions are off by one group because they hit the unnamed nested groups.

```bash
LogicalProject(
user=[REX_EXTRACT($7, '(?<user>(amber|hattie))[a-z]*@(?<domain>(pyrami|netagy))\.(?<tld>(com|org))', 1)],
domain=[REX_EXTRACT($7, '(?<user>(amber|hattie))[a-z]*@(?<domain>(pyrami|netagy))\.(?<tld>(com|org))', 2)],  -- Wrong!
    tld=[REX_EXTRACT($7, '(?<user>(amber|hattie))[a-z]*@(?<domain>(pyrami|netagy))\.(?<tld>(com|org))', 3)]     -- Wrong!
)
```

### Related Issues
* Resolve https://github.com/opensearch-project/sql/issues/4466

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -282,6 +282,9 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
           "Rex pattern must contain at least one named capture group");
     }
 
+    // TODO: Once JDK 20+ is supported, consider using Pattern.namedGroups() API for more efficient
+    // named group handling instead of manual parsing in RegexCommonUtils
+
     List<RexNode> newFields = new ArrayList<>();
     List<String> newFieldNames = new ArrayList<>();
 


core/src/main/java/org/opensearch/sql/expression/function/udf/RexExtractMultiFunction.java
```diff
@@ -50,35 +64,105 @@ public Expression implement(
...
-            matches.add(match);
-            matchCount++;
-          }
+        String match = extractor.apply(matcher);
+        if (match != null) {
+          matches.add(match);
+          matchCount++;
+        } else {
+          // If extractor returns null, it might indicate an error (like invalid group name)
+          // Stop processing to avoid infinite loop
```
RyanL1997, line 164: I'm currently thinking about adding an error handling here

Swiddis (approved):

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -294,7 +294,7 @@ public RelNode visitRex(Rex node, CalcitePlanContext context) {
                 BuiltinFunctionName.REX_EXTRACT_MULTI,
                 fieldRex,
                 context.rexBuilder.makeLiteral(patternStr),
-                context.relBuilder.literal(i + 1),
+                context.rexBuilder.makeLiteral(groupName),
```
dai-chen, line 300: I found there is a `namedGroups()` API in `Matcher` (since JDK 20?). If we can get correct index here, we don't need to modify the UDFs below? Alternatively we can move capture name -> index logic here from UDFs?

RyanL1997: That is correct - the core issue is matching named groups to their correct indices, and `Pattern.namedGroups()` would be the perfect solution. However, I discovered that we're blocked by a
compatibility constraint:

- `Pattern.namedGroups()` was introduced in JDK 20
- We need backward compatibility with JDK 11/17 - for `2.19-dev`

I agree that directly leveraging the `Pattern.namedGroups()` is the right architectural approach - we should definitely migrate to it when we fully upgrade to JDK 20+. At that point, it would be a simple one-line change in `CalciteRelNodeVisitor`.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -282,10 +282,13 @@ public RelNode visitRex(Rex node, CalcitePlanContext context) {
           "Rex pattern must contain at least one named capture group");
     }
 
+    // TODO: Once JDK 20+ is supported, consider using Pattern.namedGroups() API for more efficient
```
RyanL1997, line 285: as for now, I added a `TODO` here @dai-chen

dai-chen (approved):

PR #4629, opened 2025-10-22
Return comparable LinkedHashMap in `valueForCalcite()` of ExprTupleValue

LantaoJin:
### Description
Fix `class java.util.LinkedHashMap cannot be cast to class java.lang.Comparable` in `dedup` struct field

### Related Issues
Resolves #4339

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4648, opened 2025-10-23
Followup: Change ComparableLinkedHashMap to compare Key than Value

### LINKED ISSUE #4339: [BUG] dedup raise 500 error when object attribute is specified
**What is the bug?**
`dedup` command raise 500 error when object attribute is specified.
Sample query:
```
source=logs-* | head 10 | dedup attributes
```
where attributes is struct like:
```
      {
        "cluster": {
          "name": "xyz-cluster1-ci-prod-us-east-1",
          "region": "us-east-1",
          "env": "prod"
        },
        "log": {
          "file": {
            "path": "/var/log/xyz/abc.log"
          }
        },
        "obs_body_length": 146
      },
```

error:
```
{
  "error": {
    "details": "java.sql.SQLException: exception while executing query: class java.util.LinkedHashMap cannot be cast to class java.lang.Comparable (java.util.LinkedHashMap and java.lang.Comparable are in module java.base of loader 'bootstrap')",
    "type": "RuntimeException"
  },
  "status": 500
}
```

run ppl query: `source=logs-* | head 10 | dedup attributes`

It should return invalid query error with 4xx status code.

 - Version [e.g. 22] 3.1
 - Plugins

LantaoJin:
### Description
Followup of https://github.com/opensearch-project/sql/pull/4629: `ComparableLinkedHashMap` as an util class, better to change it to compare key than value instead of values only.

### Related Issues
Resolves #4339 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/data/utils/ComparableLinkedHashMap.java
+++ b/core/src/main/java/org/opensearch/sql/data/utils/ComparableLinkedHashMap.java
@@ -42,15 +42,29 @@ public class ComparableLinkedHashMap<K, V> extends LinkedHashMap<K, V>
     if (!thisHasNext) return -1;
     if (!otherHasNext) return 1;
 
-    V thisValue = thisIterator.next().getValue();
-    V otherValue = otherIterator.next().getValue();
-    int comparison = compareValues(thisValue, otherValue);
+    Map.Entry<K, V> thisEntry = thisIterator.next();
+    Map.Entry<K, V> otherEntry = otherIterator.next();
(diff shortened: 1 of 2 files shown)


qianheng-aws (commented):
Can you elaborate more on the scenario of comparing key? 

I'm guessing whether it's for the case that different rows get different schema, like 
```
row1: {"c": 1, "a": 1, "b": 1}
row2: {"b": "2", "a": 1}
```

However, for this case, shouldn't we sort the map by their key first separately and then do comparison between them?

PR #4651, opened 2025-10-23
Fix asc/desc keyword behavior for sort command

ritvibhatt:
### Description
Fixed PPL sort command so that asc/desc keywords specify sort direction for individual fields instead of applying all fields in the sort command. (keywords work on individual fields ex ```sort field1 asc, field2 desc``` means field1 is ascending and field2 is descending) while preventing mixing of prefix (+/-) and suffix (asc/desc) syntax, and updated integration tests to match this behavior.

Files Changed

- ppl/src/main/antlr/OpenSearchPPLParser.g4: Updated grammar rules for sort command to support individual field ASC/DESC keywords
- ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java: Updated visitSortCommand() method to handle new ASC/DESC field syntax and prevent mixing of +/-and asc/desc sort syntax
- ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java: Modified expression building logic for sort field parsing and ensure consistent sort direction syntax usage across all sort fields
- ppl/src/main/java/org/opensearch/sql/ppl/utils/ArgumentFactory.java: Updated getSortArguments() method to process new field-level ASC/DESC syntax

### Related Issues

### Check List

### DIFF
--- a/ppl/src/main/java/org/opensearch/sql/ppl/utils/ArgumentFactory.java
+++ b/ppl/src/main/java/org/opensearch/sql/ppl/utils/ArgumentFactory.java
@@ -26,6 +26,7 @@ import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.FieldsCommandCont
 import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.IntegerLiteralContext;
 import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.PrefixSortFieldContext;
 import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.RareCommandContext;
+import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.SortFieldContext;
 import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.SuffixSortFieldContext;
 import org.opensearch.sql.ppl.antlr.parser.OpenSearchPPLParser.TopCommandContext;
 
@@ -108,6 +109,22 @@ public class ArgumentFactory {
   }
 
   /**
+   * Get list of {@link Argument}.
+   *
+   * @param ctx SortFieldContext instance
+   * @return the list of arguments fetched from the sort field in sort command
+   */
+  public static List<Argument> getArgumentList(SortFieldContext ctx) {
+    if (ctx instanceof PrefixSortFieldContext) {
+      return getArgumentList((PrefixSortFieldContext) ctx);
+    } else if (ctx instanceof SuffixSortFieldContext) {
+      return getArgumentList((SuffixSortFieldContext) ctx);
+    } else {
+      return getArgumentList((DefaultSortFieldContext) ctx);
+    }
+  }
+
+  /**
    * Get list of {@link Argument} for prefix sort field (+/- syntax).
    *
    * @param ctx PrefixSortFieldContext instance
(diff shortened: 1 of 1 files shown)


ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -819,7 +819,10 @@ fieldList
    ;
 
 sortField
-   : (PLUS | MINUS)? sortFieldExpression
+   : (PLUS | MINUS) sortFieldExpression (ASC | A | DESC | D)  # invalidMixedSortField
```
ykmr1224, line 822: Why do we need to define invalid syntax? wouldn't it fail parse without this?

docs/user/ppl/cmd/sort.rst
```diff
@@ -16,13 +16,16 @@ Description
...
 
 
 * count (Since 3.3): optional. The number of results to return. **Default:** returns all results. Specifying a count of 0 or less than 0 also returns all results.
 * [+|-]: optional. The plus [+] stands for ascending order and NULL/MISSING first and a minus [-] stands for descending order and NULL/MISSING last. **Default:** ascending order and NULL/MISSING first.
+* [asc|a|desc|d]: optional. asc/a stands for ascending order and NULL/MISSING first. desc/d stands for descending order and NULL/MISSING last. **Default:** ascending order and NULL/MISSING first.
 * sort-field: mandatory. The field used to sort. Can use ``auto(field)``, ``str(field)``, ``ip(field)``, or ``num(field)`` to specify how to interpret field values.
-* [asc|a|desc|d] (Since 3.3): optional. asc/a keeps the sort order as specified. desc/d reverses the sort results. If multiple fields are specified with desc/d, reverses order of the first field then for all duplicate values of the first field, reverses the order of the values of the second field and so on. **Default:** asc.
+
+.. note::
+   You cannot mix +/- and asc/desc in the same sort command. Choose one approach for all fields in a single sort command.
```
ykmr1224, line 28: Do we need to prohibit this? Any concern to allow the mixture?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -226,20 +226,54 @@ public UnresolvedExpression visitRenameFieldExpression(RenameFieldExpressionCont
...
+            suffixKeyword,
+            prefixOperator,
+            ctx.sortFieldExpression().getText(),
+            ctx.sortFieldExpression().getText(),
+            suffixKeyword));
+  }
 
-    UnresolvedExpression fieldExpression =
-        visit(ctx.sortFieldExpression().fieldExpression().qualifiedName());
+  private Field buildSortField(
```
ykmr1224, line 263: It is not critical, but please improve if you have capacity.
Following logics looks weird to me since visitPrefixSortField / visitSuffixSortField / visitDefaultSortField already know their type, but come to the same place and branch based on the context class.
I don't have clear idea to improve it, but I think there are better way to organize the code.

ppl/src/main/antlr/OpenSearchPPLParser.g4
ritvibhatt: It would fail the parse but added this to have a more clear error for when the two syntaxes are mixed

docs/user/ppl/cmd/sort.rst
ritvibhatt: Added because the syntax gets confusing if there's a lot of mixing of the 2 different syntaxes

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ritvibhatt: Yeah it is a bit confusing, I can take this up as a task in a follow up PR to improve the logic here

docs/user/ppl/cmd/sort.rst
```diff
@@ -81,10 +84,28 @@ PPL query::
     | 13             | 28  |
     +----------------+-----+
 
-Example 4: Sort by multiple field
-=============================
+Example 4: Sort by one field in descending order (using desc)
+==============================================================
 
-The example show sort all the document with gender field in ascending order and age field in descending.
+The example show sort all the document with age field in descending order using the desc keyword.
```
RyanL1997, line 90: nit: show -> shows

ppl/src/main/java/org/opensearch/sql/ppl/utils/ArgumentFactory.java
```diff
@@ -112,19 +115,68 @@ public static List<Argument> getArgumentList(DedupCommandContext ctx) {
...
+      return new Argument("type", new Literal("auto", DataType.STRING));
+    } else if (ctx.IP() != null) {
+      return new Argument("type", new Literal("ip", DataType.STRING));
+    } else if (ctx.NUM() != null) {
+      return new Argument("type", new Literal("num", DataType.STRING));
+    } else if (ctx.STR() != null) {
+      return new Argument("type", new Literal("str", DataType.STRING));
+    } else {
+      return new Argument("type", new Literal(null, DataType.NULL));
+    }
```
RyanL1997, line 179:
```suggestion
private static Argument getTypeArgument(OpenSearchPPLParser.SortFieldExpressionContext ctx) {
    if (ctx.AUTO() != null) return createTypeArgument("auto");
    if (ctx.IP() != null) return createTypeArgument("ip");
    if (ctx.NUM() != null) return createTypeArgument("num");
    if (ctx.STR() != null) return createTypeArgument("str");
    return createTypeArgument(null);
}

private static Argument createTypeArgument(String value) {
    DataType dataType = value != null ? DataType.STRING : DataType.NULL;
    return new Argument("type", new Literal(value, dataType));
}
```

RyanL1997 (approved):

RyanL1997 (approved):

Issue #4570, opened 2025-10-15
[BUG] system limit from `plugins.ppl.join.subsearch_maxout` prevents sort push down for sort merge join


qianheng-aws:
**What is the bug?**
system limit from `plugins.ppl.join.subsearch_maxout` prevents sort push down for sort merge join

**How can one reproduce the bug?**
1. Create a index, e.g.:
```
PUT localhost:9200/test
{
    "mappings": {
        "properties": {
            "id": {
                "type": "text"
            },
            "value": {
                "type": "integer"
            }
        }
    }
}
```

2. explain the physical plan for a join query:
```
POST localhost:9200/_plugins/_ppl/_explain?format=extended

source=test | join value test 
```
and get final plan:
```
...
EnumerableCalc(expr#0..2=[{inputs}], value=[$t1], id=[$t2])
  EnumerableLimit(fetch=[10000])
    EnumerableMergeJoin(condition=[=($0, $1)], joinType=[inner])
      CalciteEnumerableIndexScan(table=[[OpenSearch, test]], PushDownContext=[[PROJECT->[value], SORT->[{\"value\" : {\"order\" : \"asc\",\"missing\" : \"_last\"}]], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"timeout\":\"1m\",\"_source\":{\"includes\":[\"value\"],\"excludes\":[]},\"sort\":[{\"value\":{\"order\":\"asc\",\"missing\":\"_last\"}}]}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
      EnumerableSort(sort0=[$0], dir0=[ASC])
        CalciteEnumerableIndexScan(table=[[OpenSearch, test]], PushDownContext=[[PROJECT->[value, id], LIMIT->50000], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"size\":50000,\"timeout\":\"1m\",\"_source\":{\"includes\":[\"value\",\"id\"],\"excludes\":[]}}, requestedTotalSize=50000, pageSize=null, startFrom=0)])
...
```

**What is the expected behavior?**
The `EnumerableSort` operator on the right child of join should be pushed down into scan. Although there is limit pushed down before the sort, that limit is a system limit from a guard rail configuration which should not prevent sort. 

**What is your host/environment?**
 - OS: macos
 - Version 3.3.0-snapshot
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

PR #4657, opened 2025-10-24
Support push down sort after limit

qianheng-aws:
### Description
Support push down sort after limit if there is only limit pushed, since PPL or other database all don't  promise sequence for only limit operator. Thus it should be acceptable to transform `limit + sort` to `sort + limit`.

However, we should avoid pushing down `sort`, if there is existing `sort` before `limit`. In such case, users intense to retrieve the Top-K values on the first sort fields from our index. The order will be overridden by the second `sort` if we keep pushing down it. So this PR prevents this case by detecting whether `top-k` is pushed down already.

Since https://github.com/opensearch-project/sql/pull/4501 has always introduced a limit before sort for join or subsearch, which will block sort push down. This PR will also enhance these scenarios, especially for left join -- both sides will have limit and sort pushed down then. See `CalciteExplainIT::testExplainScalarCorrelatedSubqueryInSelect`

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/4570

### Check List


integ-test/src/test/java/org/opensearch/sql/ppl/SortCommandIT.java
```diff
@@ -253,4 +253,11 @@ public void testSortWithAscMultipleFields() throws IOException {
         rows(36, 20),
         rows(39, 25));
   }
+
+  @Test
+  public void testHeadThenSort() throws IOException {
```
songkant-aws, line 258: [nit] Better to check NoPushDownIT case. We may need different branch for these two test cases

LantaoJin:
High level question: At the beginning, I thought you are targeting to push down a `sort` through `sysLimit`, so I renamed the PR title. But from the description, it seems a general purpose to push down the `sort` through `limit`?

qianheng-aws:
> High level question: At the beginning, I thought you are targeting to push down a `sort` through `sysLimit`, so I renamed the PR title. But from the description, it seems a general purpose to push down the `sort` through `limit`?

Yes, this a general PR for enhancement not only bug fix.

In the process of finding solution to the issue https://github.com/opensearch-project/sql/issues/4570, I found it hard to only fix that specific issue because `syslimit` has already been replaced with calcite's `logicalsort`  before optimizing. So we're not able to do anything with the `systemlimit` operator we added.

But with this general enhancement, the original issue can also been addressed.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/physical/OpenSearchSortIndexScanRule.java
```diff
@@ -48,11 +48,12 @@ public interface Config extends RelRule.Config {
...
-                                    // Skip the rule if a limit has already been pushed down
-                                    // because pushing down a sort after a limit will be treated
-                                    // as sort-then-limit by OpenSearch DSL.
+                                    // Skip the rule if Top-K(i.e. sort + limit) has already been
+                                    // pushed down. Otherwise,
+                                    // Continue to push down sort although limit has already been
+                                    // pushed down since we don't promise collation with only limit.
                                     .predicate(
-                                        Predicate.not(AbstractCalciteIndexScan::isLimitPushed)
+                                        Predicate.not(AbstractCalciteIndexScan::isTopKPushed)
```
LantaoJin, line 56: can we add a case for below ppl?
```
source = t | head 100 | stats count() as cnt | sort cnt
```
The `sort cnt` must not be pushed down through `head 100`.

Cannot for following ppl either.
```
source = t | head 100 | eval rand = rand() | sort rand
```

LantaoJin: Any `sort` expression evaluated after `limit` cannot pushed through `limit`.

qianheng-aws: Both cases cannot push the sort and it's not related to whether there is limit in the PPL query. Their no push down reason are:
1. We cannot push down metric sort into agg unless the agg is nullable=false
2. We don't support script push down for sort currently.

LantaoJin: could you add these cases in explain IT:
1. It's not the case of pushdown sort agg metrics due to there is no bucket.
2. got it. but better to add a explain IT.

LantaoJin (approved):

PR #3895, opened 2025-07-18
CVE-2025-48924: upgrade commons-lang3 to 3.18.0

LantaoJin:
### Description
Fix [CVE-2025-48924](https://github.com/advisories/GHSA-j288-q9x7-2f5v)

https://github.com/opensearch-project/opensearch-build/issues/5637

- upgrade commons-lang3 to 3.18.0
- remove commons-lang (introduced by net.hydromatic:aggdesigner-algorithm:6.0 which is not used in our code. aggdesigner is used when the schema contains Lattice entities)

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4665, opened 2025-10-25
[Maintenance] Fix CVE-2025-48924

RyanL1997:
### Description
Fix CVE-2025-48924

The problematic dependency happens to be in a diamond dependency situation:

```bash
  ├── commons-lang3:3.18.0  <- correct version already
  └── calcite-core:1.38.0
      └── aggdesigner-algorithm:6.0
          └── commons-lang:2.4  <- reported version
```

### Related Issues
* Resolve https://advisories.opensearch.org/advisory/CVE-2025-48924

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -149,6 +149,9 @@ allprojects {
         resolutionStrategy.force 'org.apache.commons:commons-text:1.11.0'
         resolutionStrategy.force 'commons-io:commons-io:2.15.0'
         resolutionStrategy.force 'org.yaml:snakeyaml:2.2'
+        resolutionStrategy.dependencySubstitution {
+            substitute module('commons-lang:commons-lang') using module('org.apache.commons:commons-lang3:3.18.0') because 'CVE-2025-48924: commons-lang 2.x vulnerable to StackOverflowError'
+        }
     }
 }
 


RyanL1997:
> Both common-lang3 3.18 and common-lang 2.6+ can fix the CVE right? Is current change enforcing common-lang3 which rename its root package?

Hi @dai-chen ,

You're correct that there are no patched versions in the commons-lang 2.x series (the advisory shows "Patched versions: None" for `commons-lang 2.0-2.6`). The only fix is upgrading to `commons-lang3 3.18.0+`. 

Yes, our dependency substitution does enforce `commons-lang3` which uses the renamed `org.apache.commons.lang3.*` package structure, but our testing confirms this works correctly. Since there's no way to patch the old `commons-lang 2.x` and we can't control what `aggdesigner-algorithm` depends on, dependency substitution to `commons-lang3 3.18.0` is the approach to resolve this.

LantaoJin:
@RyanL1997 can you explain how this problem happens again? I had excluded the commons-lang in https://github.com/opensearch-project/sql/pull/3895/files

RyanL1997:
> @RyanL1997 can you explain how this problem happens again? I had excluded the commons-lang in [#3895 (files)](https://github.com/opensearch-project/sql/pull/3895/files)

Hi @LantaoJin, in the PR you mentioned:
```gradle
// Force commons-lang3 version when requested
resolutionStrategy.force 'org.apache.commons:commons-lang3:3.18.0'

// Exclude commons-lang from calcite-core specifically  
exclude group: 'commons-lang', module: 'commons-lang'
```

The fix you did works for direct dependencies and the calcite-core case you identified. However, `resolutionStrategy.force` only applies when `commons-lang3` is specifically requested, and the `exclude` only removes `commons-lang` from that one dependency. (In other word, the case I mentioned in this PR description will not be covered.)

Based on your previous fix, mine provides global substitution - ANY request for `commons-lang:commons-lang` from anywhere in the transitive dependency tree gets automatically replaced with the safe version of `commons-lang3:3.18.0`.

So your fix handled the known cases, and the substitution acts as a safety net for any other transitive dependencies that might still pull in the vulnerable 2.x version. Both pieces were needed for this complete fix.

PR #4672, opened 2025-10-27
Support millisecond span

### LINKED ISSUE #4550: [BUG] Span millisecond incorrectly converted to microsecond interval
**What is the bug?**

When using span=X milliseconds in `timechart` commands, the span unit is incorrectly converted to `IntervalUnit.MICROSECOND` by `PlanUtils.spanUnitToIntervalUnit()`. The root cause is that the `IntervalUnit` enum does not include a `MILLISECOND` value, only `MICROSECOND`.

1. Execute a PPL query with timechart using millisecond span:
  `source=logs | timechart span=500ms count()`
2. The span calculation will use microsecond interval unit instead of millisecond, causing incorrect time bucketing.

- IntervalUnit enum should include a `MILLISECOND` value
- `PlanUtils` should correctly convert between `SpanUnit.MILLISECOND` and `IntervalUnit.MILLISECOND`

**Impact**: This bug also blocks the implementation of millisecond support in the timechart `per_*` functions (per_second, per_minute, per_hour, per_day). The `per_*` function transformation logic calculates span width in seconds instead of millisecond, leading to incorrect results when a millisecond span is used.

```
source=test_data_2023 | timechart span=1ms per_second(packets);
fetched rows / total rows = 5/5
+---------------------+-----------------------+
| @timestamp          | per_second(packets)   |
|---------------------+-----------------------|
| 2023-10-08 10:00:00 | null                  |
| 2023-10-08 10:00:05 | null                  |
| 2023-10-08 10:00:30 | null                  |
| 2023-10-08 10:00:40 | null                  |
| 2023-10-08 10:00:50 | null                  |
+---------------------+-----------------------+
```

yuancu:
### Description

Milliseconds had been improperly converted to microseconds. This PR fixes the issue by incorporating support to milliseconds.

Additionally, this PR correct a minor flaw of per function implementations so that they work properly with milliseconds.

### Related Issues
Resolves #4550

### Check List

### DIFF
--- a/ppl/src/test/java/org/opensearch/sql/ppl/parser/AstBuilderTest.java
+++ b/ppl/src/test/java/org/opensearch/sql/ppl/parser/AstBuilderTest.java
@@ -1104,10 +1104,10 @@ public class AstBuilderTest {
                 field("per_second(a)"),
                 function(
                     "/",
-                    function("*", field("per_second(a)"), doubleLiteral(1.0)),
+                    function("*", field("per_second(a)"), doubleLiteral(1000.0)),
                     function(
                         "timestampdiff",
-                        stringLiteral("SECOND"),
+                        stringLiteral("MILLISECOND"),
                         field("@timestamp"),
                         function(
                             "timestampadd",
@@ -1129,10 +1129,10 @@ public class AstBuilderTest {
                 field("per_minute(a)"),
                 function(
                     "/",
-                    function("*", field("per_minute(a)"), doubleLiteral(60.0)),
+                    function("*", field("per_minute(a)"), doubleLiteral(60000.0)),
                     function(
                         "timestampdiff",
-                        stringLiteral("SECOND"),
+                        stringLiteral("MILLISECOND"),
                         field("@timestamp"),
                         function(
                             "timestampadd",
@@ -1154,10 +1154,10 @@ public class AstBuilderTest {
                 field("per_hour(a)"),
                 function(
                     "/",
-                    function("*", field("per_hour(a)"), doubleLiteral(3600.0)),
+                    function("*", field("per_hour(a)"), doubleLiteral(3600000.0)),
                     function(
                         "timestampdiff",
-                        stringLiteral("SECOND"),
+                        stringLiteral("MILLISECOND"),
                         field("@timestamp"),
                         function(
                             "timestampadd",
@@ -1179,10 +1179,10 @@ public class AstBuilderTest {
                 field("per_day(a)"),
                 function(
                     "/",
-                    function("*", field("per_day(a)"), doubleLiteral(86400.0)),
+                    function("*", field("per_day(a)"), doubleLiteral(8.64E7)),
                     function(
                         "timestampdiff",
-                        stringLiteral("SECOND"),
+                        stringLiteral("MILLISECOND"),
                         field("@timestamp"),
                         function(
                             "timestampadd",
--- a/core/src/test/java/org/opensearch/sql/ast/tree/TimechartTest.java
+++ b/core/src/test/java/org/opensearch/sql/ast/tree/TimechartTest.java
@@ -54,9 +54,9 @@ class TimechartTest {
                 let(
                     "per_second(bytes)",
                     divide(
-                        multiply("per_second(bytes)", 1.0),
+                        multiply("per_second(bytes)", 1000.0),
                         timestampdiff(
-                            "SECOND",
+                            "MILLISECOND",
                             "@timestamp",
                             timestampadd(expectedIntervalUnit, spanValue, "@timestamp")))),
                 timechart(span(spanValue, spanUnit), alias("per_second(bytes)", sum("bytes")))));
@@ -73,9 +73,9 @@ class TimechartTest {
                 let(
                     "per_minute(bytes)",
                     divide(
-                        multiply("per_minute(bytes)", 60.0),
+                        multiply("per_minute(bytes)", 60000.0),
                         timestampdiff(
-                            "SECOND",
+                            "MILLISECOND",
                             "@timestamp",
                             timestampadd(expectedIntervalUnit, spanValue, "@timestamp")))),
                 timechart(span(spanValue, spanUnit), alias("per_minute(bytes)", sum("bytes")))));
@@ -92,9 +92,9 @@ class TimechartTest {
                 let(
                     "per_hour(bytes)",
                     divide(
-                        multiply("per_hour(bytes)", 3600.0),
+                        multiply("per_hour(bytes)", 3600000.0),
                         timestampdiff(
-                            "SECOND",
+                            "MILLISECOND",
                             "@timestamp",
                             timestampadd(expectedIntervalUnit, spanValue, "@timestamp")))),
                 timechart(span(spanValue, spanUnit), alias("per_hour(bytes)", sum("bytes")))));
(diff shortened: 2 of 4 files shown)


ppl/src/test/java/org/opensearch/sql/ppl/parser/AstBuilderTest.java
```diff
@@ -1104,10 +1104,10 @@ public void testTimechartWithPerSecondFunction() {
                 field("per_second(a)"),
                 function(
                     "/",
-                    function("*", field("per_second(a)"), doubleLiteral(1.0)),
+                    function("*", field("per_second(a)"), doubleLiteral(1000.0)),
                     function(
                         "timestampdiff",
-                        stringLiteral("SECOND"),
+                        stringLiteral("MILLISECOND"),
```
LantaoJin, line 1110: Can you add some tests for `MICROSECOND`

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4550.yml
```diff
@@ -0,0 +1,93 @@
...
+  - skip:
+      features:
+        - headers
+        - allowed_warnings
+  - do:
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: source=test_data_2023 | timechart span=500ms count()
```
LantaoJin, line 39: can you add a test for `MICROSECOND`:
```
source=test_data_2023 | timechart span=500000us count()
```

yuancu: It seems the minimum date time unit supportted in span function is millisecond:

- [SpanFunction.java](https://github.com/opensearch-project/sql/blob/484f49ebaa44264b421f54225dc19e2623ae9f85/core/src/main/java/org/opensearch/sql/expression/function/udf/SpanFunction.java)
  ```java
  public static Object evalTimestamp(
      @Parameter(name = "value") String value,
      @Parameter(name = "interval") int interval,
      @Parameter(name = "unit") String unit) {
    ExprValue exprInterval = ExprValueUtils.fromObjectValue(interval, ExprCoreType.INTEGER);
    ExprValue exprValue = ExprValueUtils.fromObjectValue(value, ExprCoreType.TIMESTAMP);
    Rounding<?> rounding = new TimestampRounding(exprInterval, unit);
    return rounding.round(exprValue).valueForCalcite();
  }
  ```
- then in [Rounding.java](https://github.com/opensearch-project/sql/blob/e4c299a73454531a05656c9fc38ab6e6681b8e4f/core/src/main/java/org/opensearch/sql/planner/physical/collector/Rounding.java)
  ```java
  DateTimeUnit.resolve(unit)

  ...

  public enum DateTimeUnit {
      @Getter private final int id;
      @Getter private final String name;
      protected final boolean isMillisBased;
      protected final long ratio;

      MILLISECOND(1, "ms", true, ChronoField.MILLI_OF_SECOND.getBaseUnit().getDuration().toMillis()) {
        @Override
        public long round(long utcMillis, int interval) {
          return DateTimeUtils.roundFloor(utcMillis, ratio * interval);
        }
      },
  
      SECOND(2, "s", true, ChronoField.SECOND_OF_MINUTE.getBaseUnit().getDuration().toMillis()) {
        @Override
        public long round(long utcMillis, int interval) {
          return DateTimeUtils.roundFloor(utcMillis, ratio * interval);
        }
      },
  ...
  }
  ```

`DateTimeUnit` is built based on milliseconds. Its `ration` attribute if of type `long`, thus can not represent a fraction of millisecond.

Should I refactor this part to extend its support to microsecond and nanosecond?

LantaoJin: What does the current user doc say?

ppl/src/main/antlr/OpenSearchPPLLexer.g4
```diff
@@ -502,7 +503,8 @@ ALIGNTIME:                          'ALIGNTIME';
 PERCENTILE_SHORTCUT:                PERC(INTEGER_LITERAL | DECIMAL_LITERAL) | 'P'(INTEGER_LITERAL | DECIMAL_LITERAL);
 
 SPANLENGTH: [0-9]+ (
-    'US'|'MS'|'CS'|'DS'
+    'US' |'CS'|'DS'
+    |'MS'|'MILLISECOND'|'MILLISECONDS'
```
dai-chen, line 507: If we want to expose microsecond span support, we need to update all docs related? Also we have to change `per_*` function to microsecond-based, otherwise `timechart span=1us per_second...` will break?

I'm thinking shall we separate the PR, e.g., fix mismatch of microsecond and millisecond in this PR and decide whether to expose all subsecond to PPL span clause later?

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4550.yml
yuancu: In [the doc for stats command](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/stats.rst), it claims that up to milliseconds is supported:


| Span Interval Units        |
|----------------------|
| millisecond (ms)           |
| second (s)                 |
| minute (m, case sensitive) |
| hour (h)                   |
| day (d)                    |
| week (w)                   |
| month (M, case sensitive)  |
| quarter (q)                |
| year (y)                   |

LantaoJin: Ok, we can file a new PR to support `MICROSECOND` , as well as `source=test_data_2023 | timechart span=500000us count()`

LantaoJin (approved):

ppl/src/main/antlr/OpenSearchPPLLexer.g4
yuancu: It seems microseconds are not supported by span yet.

I'm wondering if we are going to expand the support to microseconds, as nanoseconds are also supported in PPL, should we also do it for nanoseconds?

core/src/main/java/org/opensearch/sql/ast/expression/SpanUnit.java
```diff
@@ -15,6 +15,8 @@
 public enum SpanUnit {
   UNKNOWN("unknown"),
   NONE(""),
+  MICROSECOND("us"),
+  US("us"),
```
Swiddis, line 19: question: `μs`? Why fear unicode in the strings?

ppl/src/main/antlr/OpenSearchPPLLexer.g4
dai-chen: I'm under the same impression for `span` with `timechart` command. Probably we can track this as separate task and don't need to make more changes for span and per_* function here.

core/src/main/java/org/opensearch/sql/ast/expression/SpanUnit.java
yuancu: That's a good point. I didn't think of it. I used `us` because it was already used to represent microseconds in the grammar file:
https://github.com/opensearch-project/sql/blob/fcff083fd1a8b15686c22cf7decb80ec6fd620fd/ppl/src/main/antlr/OpenSearchPPLLexer.g4#L505

Besides, `us` has the merit in its convenience of typing; SPL also adopts `us` for microseconds: 
```
Error in 'timechart' command: The value for option span (1μs) is invalid. When span is expressed using a sub-second unit (ds, cs, ms, us), the span value needs to be < 1 second, and 1 second must be evenly divisible by the span value. 
```

Therefore, I think it's acceptable to adopt `us` for our use case. But of course, we can extend the support to `μs`.

yuancu: It comes to me that `cs` and `ds` are not supported well in span as well. I'll raise another issue for it.

PR #4562, opened 2025-10-15
Update PPL Command Documentation

ritvibhatt:
This PR standardizes the structure and content of PPL command documentation files to improve consistency and user experience.

### Documentation Structure Standardization
  Implemented consistent section ordering across all PPL command files:
  1. **Description** - Clear explanation of command functionality
  2. **Syntax** - Command syntax with parameter details
  3. **Optional sections** - Behavior notes, configuration details, or usage guidance as needed
  4. **Examples** - Practical usage examples with expected output
  5. **Limitations** - Known constraints or considerations

### Content Cleanup and Modernization
  - **Removed version information** - Eliminated outdated version references (e.g., "Version:
  3.3.0") to keep documentation current and reduce maintenance overhead
  - **Extracted aggregation functions** - Moved aggregation function documentation to dedicated
  functions file (`/functions/aggregations.rst`) for better organization

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]
#4220 

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4675, opened 2025-10-27
Mvexpand feature

### LINKED ISSUE #4439: Add native support for mvexpand command in PPL
Currently, OpenSearch SQL does not provide a native way to expand multivalue fields into separate rows/events, similar to Splunk's `mvexpand` command. This limits the ability to efficiently work with and analyze multivalue fields directly within queries, which is a common use case in log analytics and data exploration.

Introduce an `mvexpand` command (or equivalent SQL function/operator) in OpenSearch SQL that:
- Expands a multivalue field into separate rows, one per value in the field, for each input event/document.
- Supports an optional `limit` parameter to restrict how many values per event/document are expanded (similar to `mvexpand <field> [limit=<int>]` in Splunk).
- Handles memory/resource limits similar to Splunk's RAM usage (with configurable limits and clear error messaging if limits are exceeded).
- Works as a streaming/distributable command for performance at scale.

- Manual preprocessing of documents to flatten multivalue fields before indexing, which is inflexible and not always possible.
- Using scripting or custom plugins, which increases complexity for users and maintainers.
- Workarounds using nested queries or Painless scripts, which are less intuitive and harder for users familiar with Splunk/SQL semantics.

This feature would help users migrating from Splunk or other analytics platforms that rely on multivalue field expansion. For reference, see Splunk's `mvexpand` documentation:

> Expands the values of a multivalue field into separate events, one event for each value in the multivalue field. For each result, the mvexpand command creates a new result for every multivalue field.

Example usage in Splunk:
```
... | mvexpand foo
... | mvexpand foo limit=100
```

It would be valuable to have similar functionality in OpenSearch SQL to facilitate analytics, dashboarding, and data preparation tasks involving arrays or multivalue fields.

---

**Examples and Edge Cases:**

1. **Basic Example**
   - Input: A document with a field `tags: ["error", "warning", "info"]`
   - Query: `SELECT mvexpand(tags) FROM logs`
   - Output:
     | tags    |
     |---------|
     | error   |
     | warning |
     | info    |

2. **With Limit Option**
   - Input: A document with a field `ids: [1, 2, 3, 4, 5]`
   - Query: `SELECT mvexpand(ids, limit=3) FROM docs`
   - Output:
     | ids |
     |-----|
     | 1   |
     | 2   |
     | 3   |

3. **Multiple Multivalue Fields (Edge Case)**
   - Input: Document with `a: [10, 20]`, `b: [100, 200]`
   - Query: Should clarify whether expansion is supported on more than one field at a time. If not, document the limitation.

4. **Empty or Null Arrays (Edge Case)**
   - Input: `tags: []` or `tags: null`
   - Output: No rows or a row with null value, depending on desired behavior (should be documented).

5. **Large Arrays / Memory Limits (Edge Case)**
   - Input: Document with a field containing thousands of values
   - Behavior: Should trigger a warning or error if memory/resource usage exceeds limits, and document how this is handled.

6. **Non-Array Fields (Edge Case)**
   - Input: `tags: "error"` (not an array)
   - Output: Should document if the field is left as-is, converted to a single-value array, or an error is thrown.

srikanthpadakanti:
### Description
This pull request adds native support for the mvexpand command in PPL to OpenSearch SQL, enabling users to expand multivalue fields (arrays) into separate rows directly within queries. This functionality is analogous to Splunk's mvexpand command and streamlines analytics, dashboarding, and data preparation involving arrays or multivalue fields.

Key features introduced:

Native mvexpand command for PPL queries to expand array fields into separate rows/events.
Optional limit parameter to restrict the number of expanded values per event/document.
Robust handling of empty/null arrays, large arrays (with memory/resource limits), and non-array fields.
Streaming/distributable execution for performance and scalability.
Comprehensive documentation and edge case coverage.
This feature makes OpenSearch SQL more powerful and user-friendly for log analytics, data exploration, and migration from platforms like Splunk.

### Related Issues
Resolves #4439
https://github.com/opensearch-project/sql/issues/4439

### Check List
- [ X] New functionality includes testing.
- [ X] New functionality has been documented.
 - [ X] New functionality has javadoc added.
 - [ X] New functionality has a user manual doc added.
- [ X] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ X] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ X] Commits are signed per the DCO using `--signoff` or `-s`.

### DIFF
--- a/docs/user/ppl/cmd/stats.md
+++ b/docs/user/ppl/cmd/stats.md
@@ -48,6 +48,36 @@ The stats command supports the following aggregation functions:
 * VALUES: Collect unique values into sorted array  
   
 For detailed documentation of each function, see [Aggregation Functions](../functions/aggregations.md).
+
+## Limitations
+
+### Bucket aggregation result may be approximate in large dataset
+
+In OpenSearch, `doc_count` values for a terms bucket aggregation may be approximate. As a result, any aggregations (such as `sum` and `avg`) on the terms bucket aggregation may also be approximate.
+For example, the following PPL query (find the top 10 URLs) may return an approximate result if the cardinality of `URL` is high.
+
+```ppl ignore
+source=hits
+| stats bucket_nullable=false count() as c by URL
+| sort - c
+| head 10
+```
+
+This query is pushed down to a terms bucket aggregation DSL query with `"order": { "_count": "desc" }`. In OpenSearch, this terms aggregation may throw away some buckets.
+
+### Sorting by ascending doc_count may produce inaccurate results
+
+Similar to above PPL query, the following query (find the rare 10 URLs) often produces inaccurate results.
+
+```ppl ignore
+source=hits
+| stats bucket_nullable=false count() as c by URL
+| sort + c
+| head 10
+```
+
+A term that is globally infrequent might not appear as infrequent on every individual shard or might be entirely absent from the least frequent results returned by some shards. Conversely, a term that appears infrequently on one shard might be common on another. In both scenarios, rare terms can be missed during shard-level aggregation, resulting in incorrect overall results.
+
 ## Example 1: Calculate the count of events  
 
 This example shows calculating the count of events in the accounts.
--- a/docs/user/ppl/functions/aggregations.md
+++ b/docs/user/ppl/functions/aggregations.md
@@ -2,7 +2,7 @@
 
 ## Description  
 
-Aggregation functions perform calculations across multiple rows to return a single result value. These functions are used with `stats` and `eventstats` commands to analyze and summarize data.
+Aggregation functions perform calculations across multiple rows to return a single result value. These functions are used with `stats`, `eventstats` and `streamstats` commands to analyze and summarize data.
 The following table shows how NULL/MISSING values are handled by aggregation functions:
   
 | Function | NULL | MISSING |
--- a/docs/user/ppl/limitations/limitations.md
+++ b/docs/user/ppl/limitations/limitations.md
@@ -87,3 +87,30 @@ If a document contains malformed field names inside an object field, PPL ignores
 When ``log`` is an object field with ``enabled: false``, subfields with malformed names are ignored.
 
 **Recommendation:** Avoid using field names that contain leading dots, trailing dots, consecutive dots, or consist only of dots. This aligns with OpenSearch's default field naming requirements.
+
+## Bucket aggregation result may be approximate in large dataset
+
+In OpenSearch, `doc_count` values for a terms bucket aggregation may be approximate. As a result, any aggregations (such as `sum` and `avg`) on the terms bucket aggregation may also be approximate.
+For example, the following PPL query (find the top 10 URLs) may return an approximate result if the cardinality of `URL` is high.
+
+```ppl ignore
+source=hits
+| stats bucket_nullable=false count() as c by URL
+| sort - c
+| head 10
+```
+
+This query is pushed down to a terms bucket aggregation DSL query with `"order": { "_count": "desc" }`. In OpenSearch, this terms aggregation may throw away some buckets.
+
+## Sorting by ascending doc_count may produce inaccurate results
+
+Similar to above PPL query, the following query (find the rare 10 URLs) often produces inaccurate results.
+
+```ppl ignore
+source=hits
+| stats bucket_nullable=false count() as c by URL
+| sort + c
+| head 10
+```
+
+A term that is globally infrequent might not appear as infrequent on every individual shard or might be entirely absent from the least frequent results returned by some shards. Conversely, a term that appears infrequently on one shard might be common on another. In both scenarios, rare terms can be missed during shard-level aggregation, resulting in incorrect overall results.


srikanthpadakanti:
Attaching the manual test cases and its output 
[mvexpand_manual_test_results.md](https://github.com/user-attachments/files/23172416/mvexpand_manual_test_results.md)

srikanthpadakanti:
Hi maintainers @penghuo , the "Enforce PR labels" check is failing because I can't add labels. Could you please add the required labels to this PR so the checks pass? Thank you!

core/src/main/java/org/opensearch/sql/ast/dsl/AstDSL.java
```diff
@@ -135,6 +136,10 @@ public Expand expand(UnresolvedPlan input, Field field, String alias) {
     return new Expand(field, alias).attach(input);
   }
 
+  public static UnresolvedPlan mvexpand(UnresolvedPlan input, Field field, Integer limit) {
+    return new MvExpand(field, limit);
```
dai-chen: `input` is useless?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,428 @@
...
+    return Frameworks.newConfigBuilder()
+        .parserConfig(SqlParser.Config.DEFAULT)
+        .defaultSchema(schema)
+        .traitDefs((List<RelTraitDef>) null)
+        .programs(Programs.heuristicJoinOrder(Programs.RULE_SET, true, 2));
+  }
+
+  // Option 2: Assert specific logical plan (as per the main edge/typical cases)
+  @Test
+  public void testMvExpandBasic() {
```
dai-chen: Could you add verifyPPLToSparkSQL assertion for each test?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -458,6 +459,10 @@ expandCommand
     : EXPAND fieldExpression (AS alias = qualifiedName)?
     ;
 
+mvexpandCommand
+    : MVEXPAND fieldExpression (LIMIT INTEGER_LITERAL)?
```
dai-chen: `limit int` or `limit=int`?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1553,6 +1554,92 @@ private static void buildDedupNotNull(
     context.relBuilder.projectExcept(_row_number_dedup_);
   }
 
+  private void buildMvExpandRelNode(
```
dai-chen: Could you help understand the complexity here? I thought mvexpand is essentially SQL unnest/lateral join. Any logic here can be delegated to other visit method recursively?

ykmr1224: It should be very similar to `expand` command, and suppose we can unify most of the logic.

core/src/main/java/org/opensearch/sql/ast/tree/MvExpand.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public Field getField() {
+    return field;
+  }
+
+  @Nullable
+  public Integer getLimit() {
+    return limit;
+  }
```
ykmr1224: Is it needed them even though we have `@Getter` annotation?

docs/user/ppl/cmd/mvexpand.rst
ykmr1224, line 1: Please add it to `docs/category.json` so doctest tests it.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,134 @@
...
+        "2|{\"username\":\"single\", \"skills\":[{\"name\":\"go\"}]}",
+        "3|{\"username\":\"empty\", \"skills\":[]}",
+        "4|{\"username\":\"nullskills\", \"skills\":null}",
+        "5|{\"username\":\"noskills\"}",
+        "6|{\"username\":\"missingattr\", \"skills\":[{\"name\":\"c\"},{\"level\":\"advanced\"}]}",
+        "7|{\"username\":\"complex\","
+            + " \"skills\":[{\"name\":\"ml\",\"level\":\"expert\"},{\"name\":\"ai\"},{\"level\":\"novice\"}]}",
+        "8|{\"username\":\"duplicate\", \"skills\":[{\"name\":\"dup\"},{\"name\":\"dup\"}]}",
+        "9|{\"username\":\"large\","
+            + " \"skills\":[{\"name\":\"s1\"},{\"name\":\"s2\"},{\"name\":\"s3\"},{\"name\":\"s4\"},{\"name\":\"s5\"},{\"name\":\"s6\"},{\"name\":\"s7\"},{\"name\":\"s8\"},{\"name\":\"s9\"},{\"name\":\"s10\"}]}");
```
ykmr1224: Not sure if we need to encode the id into the string. It should be simpler to increment id in `bulkInsert` method.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,134 @@
...
+public class CalciteMvExpandCommandIT extends PPLIntegTestCase {
+
+  private static final String INDEX = "mvexpand_edge_cases";
+
+  @Override
+  public void init() throws Exception {
+    super.init();
+    enableCalcite();
+    deleteIndexIfExists(INDEX);
+    createIndex(
```
ykmr1224: Should we rather use Index.MVEXPAND_EDGE_CASES? (Is it duplicate?)

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,428 @@
...
+  public CalcitePPLMvExpandTest() {
+    super(CalciteAssert.SchemaSpec.SCOTT_WITH_TEMPORAL);
+  }
+
+  @Override
+  protected Frameworks.ConfigBuilder config(CalciteAssert.SchemaSpec... schemaSpecs) {
+    final SchemaPlus rootSchema = Frameworks.createRootSchema(true);
+    final SchemaPlus schema = CalciteAssert.addSchema(rootSchema, schemaSpecs);
+
+    ImmutableList<Object[]> users =
```
ykmr1224: Do we need such a complicated data for this test? (I think the tests won't use the data)

core/src/main/java/org/opensearch/sql/ast/dsl/AstDSL.java
srikanthpadakanti: fixed.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: buildExpandRelNode assumes the left-side rowType / relBuilder stack are unchanged, so it uses existing RexInputRefs (indexes). If a prior project/rename changed field positions, those indexes become invalid and correlate’s required-columns can point at the wrong column. 

For mvexpand I capture leftBuilt and its rowType, resolve the array field index by name (fallback to original), build the correlated Uncollect with rexBuilder.makeCorrel(...), and create requiredColumnRef with rexBuilder.makeInputRef(leftRowType, index) before correlating.
That sequence guarantees the required-column refs are created against the actual left rowType and avoids subtle planner errors.
buildExpandRelNode still works for the simple/no-projection case, but the above is needed for correctness when renames/projections exist.
I can add a tiny unit test that renames the array field before mvexpand to demonstrate the failure of the naive approach if you want. 

Example that breaks the naive approach:
Query: source=USERS | fields USERNAME, skills as skills_renamed | mvexpand skills_renamed
Naive buildExpandRelNode uses the RexInputRef/index computed before the rename; after the projection/rename that numeric index no longer points to the array column, so correlate's required-columns target the wrong column → wrong projection/removal or planner errors.

What I did and why it fixes it:
I capture leftBuilt and its rowType before building the right side, resolve the array field index by name against leftBuilt.getRowType() (fallback to original index), then build the correlated Uncollect using rexBuilder.makeCorrel(...) and create requiredColumnRef with rexBuilder.makeInputRef(leftRowType, index).
This guarantees the required-column refs are created against the actual left rowType (the one used in correlate), so indexes never drift after projections/renames.
Bottom line: buildExpandRelNode works for simple/no-projection cases, but the extra steps are required to be correct and robust when prior projections/renames change field positions.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
srikanthpadakanti: Done

ppl/src/main/antlr/OpenSearchPPLParser.g4
srikanthpadakanti: limit=int. Fixed it

core/src/main/java/org/opensearch/sql/ast/tree/MvExpand.java
srikanthpadakanti: That was sloppy. Fixed.

docs/user/ppl/cmd/mvexpand.rst
srikanthpadakanti: Added to docs/category.json

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
dai-chen: I don't see it. Did I miss it somewhere?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: Fixed.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: done.

srikanthpadakanti:
Hello @jed326 @ykmr1224 
I have addressed the PR comments and made the changes accordingly. Please do review.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
srikanthpadakanti: I did pushed all the changes at once. You must find them now.

docs/category.json
```diff
@@ -50,7 +50,6 @@
     "user/ppl/cmd/subquery.rst",
     "user/ppl/cmd/syntax.rst",
     "user/ppl/cmd/timechart.rst",
-    "user/ppl/cmd/search.rst",
```
srikanthpadakanti: I deleted the duplicate entry. Which is not part of my change. Just FYI

srikanthpadakanti:
Hello @dai-chen There are some tests failed in Security Plugin IT - unrelated to my mvexpand changes. 
@ykmr1224

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -459,6 +460,10 @@ expandCommand
     : EXPAND fieldExpression (AS alias = qualifiedName)?
     ;
 
+mvexpandCommand
+    : MVEXPAND fieldExpression (LIMIT '=' INTEGER_LITERAL)?
```
dai-chen: Use token `EQUAL`?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,455 @@
...
+  /**
+   * Verify PPL -> SparkSQL conversion for an already-built RelNode.
+   *
+   * <p>Strategy: 1) Try to instantiate org.opensearch.sql.calcite.OpenSearchSparkSqlDialect.DEFAULT
+   * via reflection and run a typed RelToSqlConverter with it (best effort - exercises same path
+   * used in other tests). 2) Fallback: use the private 'converter' instance from
+   * CalcitePPLAbstractTest via reflection and call its visitRoot(...) method; assert it produced a
+   * non-null statement object.
+   */
+  private void verifyPPLToSparkSQL(RelNode root) {
```
dai-chen: Could clarify why we need to do this? Other test just do `verifyPPLToSparkSQL` with expected SparkSQL text. I think the purpose of this assertion is 1) show equivalent SQL query to help understand the new command; 2) make sure new PPL command can be ported to other engine via SQL representation.

docs/user/ppl/cmd/mvexpand.rst
```diff
@@ -0,0 +1,201 @@
...
+
+| Key features of ``mvexpand``:
+- Expands array fields into multiple rows, one per value.
+- Supports an optional ``limit`` parameter to restrict the number of expanded values per document.
+- Handles empty, null, and non-array fields gracefully.
+- Works as a streaming/distributable command for performance and scalability.
+
+Version
+=======
+3.3.0
```
dai-chen: FYI, we're trying to standardize PPL documentation and explicit version is not needed. Ref: https://github.com/opensearch-project/sql/pull/4562

docs/user/ppl/cmd/mvexpand.rst
```diff
@@ -0,0 +1,201 @@
...
+
+Syntax
+======
+mvexpand <field> [limit=<int>]
+
+* **field**: The multivalue (array) field to expand. (Required)
+* **limit**: Maximum number of values per document to expand. (Optional)
```
dai-chen: Clarify the behavior without limit specified? Either default limit or no limit. Is returned values in order or randomly?

docs/user/ppl/cmd/mvexpand.rst
```diff
@@ -0,0 +1,201 @@
...
+    fetched rows / total rows = 1/1
+    +-------+
+    | tags  |
+    +-------+
+    | error |
+    +-------+
+
+Example 5: Large Arrays and Memory Limits
+----------------------------------------
+If an array exceeds configured memory/resource limits, mvexpand returns an error.
```
dai-chen: Is this config already supported?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1547,6 +1548,20 @@ private static void buildDedupNotNull(
     context.relBuilder.projectExcept(_row_number_dedup_);
   }
 
+  @Override
+  public RelNode visitMvExpand(MvExpand node, CalcitePlanContext context) {
```
dai-chen: Is `limit` option the only difference between `expand` and `mvexpand`? I see `expand` is translated to SQL below. If so, we just need to add a `LIMIT` clause in inner query? Ref: https://github.com/opensearch-project/sql/blob/main/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLExpandTest.java#L105

cc: @yuancu

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
srikanthpadakanti: Good point — removed the reflective PPL->SQL helper and its calls to keep tests consistent with other planner tests. If we want SQL assertions we can add explicit expectedSparkSql checks per test.

ppl/src/main/antlr/OpenSearchPPLParser.g4
srikanthpadakanti: Ack.

docs/user/ppl/cmd/mvexpand.rst
srikanthpadakanti: Ack.

docs/user/ppl/cmd/mvexpand.rst
srikanthpadakanti: If no `limit` is specified, mvexpand expands all elements in the array (there is no implicit per-document cap). Elements are emitted in the same order they appear in the array (array iteration order). If the underlying field does not provide a defined order, the output order is undefined. Use `limit` to bound the number of expanded rows per document and to avoid resource issues on very large arrays. 

Let me add the same to .rst file.

docs/user/ppl/cmd/mvexpand.rst
srikanthpadakanti: There is no mvexpand-specific config. Resource limits are enforced by the engine/cluster (e.g. circuit breakers and SQL/PPL query size/timeouts). Use mvexpand's limit data to avoid hitting those limits; if you want I can add references to the exact cluster/SQL settings for our release. 
 
 Let me add the same to the documentation

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: I have added the limit clause in the inner query.

core/src/main/java/org/opensearch/sql/ast/tree/MvExpand.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Override
+  public MvExpand attach(UnresolvedPlan child) {
+    this.child = child;
+    return this;
+  }
+
+  @Nullable
+  public Integer getLimit() {
```
ykmr1224: Is this needed? (`@Getter` should be enough)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1549,6 +1550,25 @@ private static void buildDedupNotNull(
...
+    visitChildren(node, context);
+    Field arrayField = node.getField();
+    RexInputRef arrayFieldRex = (RexInputRef) rexVisitor.analyze(arrayField, context);
+
+    // buildMvExpandRelNode(arrayFieldRex, arrayField.getField().toString(), null, context);
```
ykmr1224: should delete?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1549,6 +1550,25 @@ private static void buildDedupNotNull(
...
+
+    // pass the per-document limit into the builder so it can be applied inside the UNNEST inner
+    // query
+    buildMvExpandRelNode(
+        arrayFieldRex, arrayField.getField().toString(), null, node.getLimit(), context);
+
+    //    if (node.getLimit() != null) {
+    //      context.relBuilder.limit(0, node.getLimit());
+    //    }
```
ykmr1224: should delete?

ykmr1224:
https://github.com/opensearch-project/sql/pull/4675#discussion_r2479252548
From this comment, it looks like an improvement for existing `expand` command.
Why don't we improve `expand` at the same time, and consider `mvexpand` as and alias of `expand`?

docs/user/ppl/cmd/mvexpand.rst
dai-chen: Do you mean SQL/PPL's circuit breaker or OpenSearch DSL's? I think our circuit breaker today only protect scan operator.

srikanthpadakanti: I clarified in the mvexpand docs that resource limits are enforced by OpenSearch node/SQL-PPL limits (no mvexpand-specific config) and documented using limit + prefiltering as the recommended mitigation; 

https://docs.opensearch.org/1.0/search-plugins/ppl/settings/ 

Referring to this: 
plugins.query.memory_limit | Set heap memory usage limit. If a query crosses this limit, it’s terminated.

core/src/main/java/org/opensearch/sql/ast/tree/MvExpand.java
srikanthpadakanti: Done

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: done

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: done

srikanthpadakanti:
> [#4675 (comment)](https://github.com/opensearch-project/sql/pull/4675#discussion_r2479252548) From this comment, it looks like an improvement for existing `expand` command. Why don't we improve `expand` at the same time, and consider `mvexpand` as and alias of `expand`?

Refactored as requested. 
Improved the "expand" command and aliased the - "mvexpand" to it.

docs/user/ppl/cmd/mvexpand.rst
dai-chen: I see. I recall in our V3 (Calcite-based) engine, only index scan operator is protected. Circuit breaking other operators are under research. I think probably we just keep "To avoid failures when expanding large arrays:" section which is very helpful.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3315,46 +3335,51 @@ private void flattenParsedPattern(
...
   }
 
-  private void buildExpandRelNode(
-      RexInputRef arrayFieldRex, String arrayFieldName, String alias, CalcitePlanContext context) {
-    // 3. Capture the outer row in a CorrelationId
-    Holder<RexCorrelVariable> correlVariable = Holder.empty();
-    context.relBuilder.variable(correlVariable::set);
+  // New generic helper: builds Uncollect + Correlate using a provided left node (so caller
+  // can ensure left rowType is fixed).
+  private void buildUnnestForLeft(
```
ykmr1224: I think we no longer need this method as the logic is unified. We might want to extract some portion of `buildExpandRelNode`, but we should extract with fewer parameters in that case.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1589,6 +1595,20 @@ private static void buildDedupNotNull(
     context.relBuilder.projectExcept(_row_number_dedup_);
   }
 
+  @Override
+  public RelNode visitMvExpand(MvExpand node, CalcitePlanContext context) {
```
ykmr1224: nit: Let's move this method next to `visitExpand` as it is closely related.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3363,6 +3388,74 @@ private void buildExpandRelNode(
...
+    if (arrayFieldRexNode instanceof RexInputRef) {
+      arrayFieldRex = (RexInputRef) arrayFieldRexNode;
+    } else {
+      RelDataType currentRowType = context.relBuilder.peek().getRowType();
+      RelDataTypeField fld = currentRowType.getField(arrayFieldName, false, false);
+      if (fld != null) {
+        arrayFieldRex = context.rexBuilder.makeInputRef(currentRowType, fld.getIndex());
+      } else {
+        throw new IllegalArgumentException(
+            "buildExpandRelNode: expected RexInputRef or resolvable field name: " + arrayFieldName);
```
ykmr1224: Is this expected to be visible to the user? In that case, error message should be meaningful for the user.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3363,6 +3388,74 @@ private void buildExpandRelNode(
...
+      Integer mvExpandLimit,
+      CalcitePlanContext context) {
+
+    // Convert incoming RexNode to RexInputRef when possible; otherwise resolve by field name.
+    RexInputRef arrayFieldRex;
+    if (arrayFieldRexNode instanceof RexInputRef) {
+      arrayFieldRex = (RexInputRef) arrayFieldRexNode;
+    } else {
+      RelDataType currentRowType = context.relBuilder.peek().getRowType();
+      RelDataTypeField fld = currentRowType.getField(arrayFieldName, false, false);
```
ykmr1224: What kind of case falls into this branch? Which test case associated with this branch?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,148 @@
...
+        INDEX,
+        "{\"username\":\"happy\",\"skills\":[{\"name\":\"python\"},{\"name\":\"java\"},{\"name\":\"sql\"}]}",
+        "{\"username\":\"single\",\"skills\":[{\"name\":\"go\"}]}",
+        "{\"username\":\"empty\",\"skills\":[]}",
+        "{\"username\":\"nullskills\",\"skills\":null}",
+        "{\"username\":\"noskills\"}",
+        "{\"username\":\"missingattr\",\"skills\":[{\"name\":\"c\"},{\"level\":\"advanced\"}]}",
+        "{\"username\":\"complex\",\"skills\":[{\"name\":\"ml\",\"level\":\"expert\"},{\"name\":\"ai\"},{\"level\":\"novice\"}]}",
+        "{\"username\":\"duplicate\",\"skills\":[{\"name\":\"dup\"},{\"name\":\"dup\"}]}",
+        "{\"username\":\"large\",\"skills\":[{\"name\":\"s1\"},{\"name\":\"s2\"},{\"name\":\"s3\"},{\"name\":\"s4\"},{\"name\":\"s5\"},{\"name\":\"s6\"},{\"name\":\"s7\"},{\"name\":\"s8\"},{\"name\":\"s9\"},{\"name\":\"s10\"}]}");
```
ykmr1224: Looks like some records are not used in tests. Should we add test cases or remove unused records?

docs/user/ppl/cmd/mvexpand.rst
srikanthpadakanti: I updated the docs to clarify that there is no mvexpand-specific configuration. Resource limits that can terminate queries are enforced by the OpenSearch node (e.g. query/heap limits) and SQL/PPL query controls. I also added a note that, in the current Calcite-based engine, circuit breaking protections are applied primarily to the index-scan operator and protections for other operators are still under research. The docs now recommend using mvexpand's limit + prefiltering as the primary mitigation and link to the relevant PPL/engine settings.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Addressed: removed buildUnnestForLeft and inlined its logic into buildExpandRelNode. buildMvExpandRelNode now delegates to buildExpandRelNode and passes the per-document limit through. Also added JavaDoc for visitMvExpand to match visitExpand. Functionality is unchanged

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: done

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Good catch — I updated the method to throw SemanticCheckException with clear, user-facing messages ("field not found" / "expected ARRAY but found X") and to validate type both at resolve-time and after materialization (to catch rename/alias cases)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: The else branch is hit when buildExpandRelNode receives a RexNode that is not a RexInputRef (e.g. calls like context.relBuilder.field(node.getAlias()) from the Patterns aggregation). I added unit tests covering the two negative, user-facing validation cases (missing field and non-ARRAY type).

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: Removed the unused records.

dai-chen:
Please check CI failure.

srikanthpadakanti:
> Please check CI failure.

@dai-chen  The CI failures are not related to mvexpand. I merged the upstream changes to. my main. CI failure is pointing to - 
testMvindexSingleElementPositive, testMvindexSingleElementNegative, testMvindexRangePositive, testMvindexRangeNegative in CalcitePPLArrayFunctionTest.
The assertion failures are string-equality checks of generated SQL (PPL → SparkSQL text). The RelNode / logical plan printed in the local run shows ITEM(...) and ARRAY_SLICE(...) expressions — i.e. the array functions are present and being produced, but the exact string formatting (parentheses, unary +/-, expression grouping) differs from the test expectation used in CI.

My local run shows those tests pass. But, CI failed. That implies either:
A difference in environment (JDK/OS/Calcite pretty-printer) between my local and CI, or
Different code in CI (different branch/commit or cached build) than my local, or
Tests were updated in one place and translator/pretty printing in another.method in CalcitePPLArrayFunctionTest. @ahkcs Can you please confirm

core/src/main/java/org/opensearch/sql/ast/dsl/AstDSL.java
```diff
@@ -136,6 +137,11 @@ public Expand expand(UnresolvedPlan input, Field field, String alias) {
     return new Expand(field, alias).attach(input);
   }
 
+  public static UnresolvedPlan mvexpand(UnresolvedPlan input, Field field, Integer limit) {
+    // attach the incoming child plan so the AST contains the pipeline link
```
ykmr1224: nit: unneeded comment.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3056,45 +3076,114 @@ private void flattenParsedPattern(
...
-            arrayFieldRex.getIndex());
-    RelNode leftNode = context.relBuilder.build();
+            context.rexBuilder.makeCorrel(leftRowType, correlVariable.get().id),
+            arrayFieldIndexInLeft);
+
+    // Materialize leftBuilt
+    RelNode leftBuilt = context.relBuilder.build();
 
-    // 5. Build join right node and expand the array field using uncollect
+    // Build the right (uncollect) using the small helper
```
ykmr1224: nit: please remove redundant comment (this comment does not give any additional information)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3056,45 +3076,114 @@ private void flattenParsedPattern(
...
+    RexNode toRemove;
+    try {
+      toRemove = context.relBuilder.field(arrayFieldName);
+    } catch (Exception e) {
+      // Fallback in case name lookup fails
+      toRemove = requiredColumnRef;
+    }
+    context.relBuilder.projectExcept(toRemove);
 
+    // Optional rename into alias (preserve the original logic)
```
ykmr1224: nit: please rephrase. `preserve the original logic` won't make any sense for code reader.

core/src/test/java/org/opensearch/sql/calcite/CalciteRelNodeVisitorExpandTest.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /**
+   * Negative: requested field does not exist in current row type -> SemanticCheckException
+   *
+   * <p>This exercises the resolve-by-name branch early validation that throws when the named field
+   * is not found in the current row type.
+   */
+  @Test
+  public void expand_on_nonexistent_field_should_throw_user_friendly_error() throws Exception {
```
ykmr1224: It does not make much sense to directly call private instance method via reflection to test it.
Can't we test it in IT? (otherwise, is the condition realistic?)

core/src/test/java/org/opensearch/sql/calcite/CalciteRelNodeVisitorExpandTest.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /**
+   * Negative: requested field exists but is not an ARRAY -> SemanticCheckException
+   *
+   * <p>This exercises the resolve-by-name branch early validation that throws when the named field
+   * exists but its type is not ArraySqlType.
+   */
+  @Test
+  public void expand_on_non_array_field_should_throw_expected_array_message() throws Exception {
```
ykmr1224: ditto

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
ykmr1224: Is it addressed?
I thinks we should add test case for `happy` at least.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
ykmr1224: Looks it is not fixed.

ykmr1224 (changes requested):

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: @ykmr1224  While building the happy-path test, I noticed the operator is repeating the first element rather than iterating through all values. Using a simple nested‐field index (mvexpand_edge_cases) with skills = [java, python, sql], the PPL query:

source=mvexpand_edge_cases | mvexpand skills | fields username, skills.name


returns three rows of "java" instead of java, python, sql. This indicates the iterator is not advancing through the array and repeatedly emits the first element. 


So I had to rewrite the logic - so that mvexpand works as it supposed to. Please review it. Thanks in advance.

core/src/main/java/org/opensearch/sql/ast/dsl/AstDSL.java
srikanthpadakanti: done

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Deleted

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Added detailed information

core/src/test/java/org/opensearch/sql/calcite/CalciteRelNodeVisitorExpandTest.java
srikanthpadakanti: Moved this to IT and removed this unit test

core/src/test/java/org/opensearch/sql/calcite/CalciteRelNodeVisitorExpandTest.java
srikanthpadakanti: Moved this to IT and removed this unit test.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: changed to  - Index.MVEXPAND_EDGE_CASES.getName()

dai-chen (commented):
I found this in expand command code: https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExpandCommandIT.java#L141.

Could you clarify why not make `mvexpand` an alias of `expand` command with enhancements? If we can do this, I think many new changes can be avoided.

ahkcs:
>My local run shows those tests pass. But, CI failed. That implies either:
A difference in environment (JDK/OS/Calcite pretty-printer) between my local and CI, or
Different code in CI (different branch/commit or cached build) than my local

Can you try rebasing to current main and make sure your local jdk version is the same as CI? Currently CI is failing

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,344 @@
...
+    // followed by an inner uncollect prescription. Update expected logical plan accordingly.
+    String expectedLogical =
+        "LogicalProject(USERNAME=[$0], skills.name=[$2])\n"
+            + "  LogicalCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{1}])\n"
+            + "    LogicalTableScan(table=[[scott, USERS]])\n"
+            + "    LogicalProject(skills.name=[$0])\n"
+            + "      Uncollect\n"
+            + "        LogicalProject(skills=[$cor0.skills])\n"
+            + "          LogicalValues(tuples=[[{ 0 }]])\n";
+    verifyLogical(root, expectedLogical);
```
ahkcs: Can you also add verifyPPLtoSparkSQL to verify expectedSparkSql?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,344 @@
...
+    String expectedLogical =
+        "LogicalSort(fetch=[1])\n"
+            + "  LogicalProject(USERNAME=[$0], skills.name=[$2])\n"
+            + "    LogicalCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{1}])\n"
+            + "      LogicalTableScan(table=[[scott, USERS]])\n"
+            + "      LogicalProject(skills.name=[$0])\n"
+            + "        Uncollect\n"
+            + "          LogicalProject(skills=[$cor0.skills])\n"
+            + "            LogicalValues(tuples=[[{ 0 }]])\n";
+    verifyLogical(root, expectedLogical);
```
ahkcs: ditto for other UTs

docs/user/ppl/cmd/mvexpand.rst
```diff
@@ -0,0 +1,201 @@
...
+
+Input document::
+
+    { "tags": ["error", "warning", "info"] }
+
+PPL query::
+
+    source=logs | mvexpand tags
+
+Output (example)::
```
ahkcs: For the doc examples, can you turn them into doctests? doctest is an integration test that makes sure the examples in the doc works

srikanthpadakanti:
> I found this in expand command code: https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExpandCommandIT.java#L141.
> 
> Could you clarify why not make `mvexpand` an alias of `expand` command with enhancements? If we can do this, I think many new changes can be avoided.


@dai-chen 
mvexpand and expand solve different problems and operate on different data shapes.
expand historically works on nested/struct fields, projecting the struct into rows.
mvexpand works on multivalue/array fields, emitting one row per array element, which requires different type validation, planner rules, and runtime behavior.
mvexpand also needs to handle arrays containing primitives, objects, nulls, and mixed shapes, and must reliably produce per-element rows (including null elements). expand does not extract sub-fields from array elements and its semantics around aliasing and struct projection are different.
OpenSearch mappings complicate this further: a field typed as keyword may still contain arrays at runtime. mvexpand therefore needs stricter semantic checks and fallbacks that don’t align with the legacy expand contract. Reusing expand directly would break backward-compatible behavior expected by existing queries and tests.

**Problems in alias attempt:**
The alias attempt produced repeated values (java, java, java) instead of (java, python, sql) from our IT example. 
The repeated output indicates the plan was projecting the same input reference for every row instead of the per-element value. This typically happens when:
field resolution occurs before the uncollect/correlate materializes element values, the projection uses the outer input ref instead of the element ref, or

column indices (RexInputRef) drift when reusing expand logic, causing all projected rows to read from the same column.

In short, simply delegating to expand bypassed the per-element extraction logic, so the projection kept resolving to the same value on every row. 

--------- 
Refactor: both EXPAND and MVEXPAND delegate to buildExpandCore(...), which implements the // shared correlate + uncollect + projection logic. visitExpand applies EXPAND semantics // (remove original array field, optional alias rename). visitMvExpand handles its two special // cases — missing field (returns an empty VALUES row with a nullable placeholder) and // limit=<n> — then delegates to the shared helper.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
srikanthpadakanti: done

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
srikanthpadakanti: done

docs/user/ppl/cmd/mvexpand.rst
srikanthpadakanti: done

srikanthpadakanti:
Please approve the workflow @dai-chen @penghuo

core/src/main/java/org/opensearch/sql/ast/dsl/AstDSL.java
ykmr1224: It is still there.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1054,6 +1054,14 @@ void populate() {
...
                       .or(OperandTypes.family(SqlTypeFamily.MAP, SqlTypeFamily.ANY)),
               false));
+      // Allow using INTERNAL_ITEM when the element type is unknown/undefined at planning time.
+      // Some datasets (or Calcite's type inference) may give the element an UNDEFINED type.
+      // Accept a "ignore" first-argument family so INTERNAL_ITEM(elem, 'key') can still be planned
+      // and resolved at runtime (fallback semantics handled at execution side). - Used in MVEXPAND
+      registerOperator(
+          INTERNAL_ITEM,
+          SqlStdOperatorTable.ITEM,
+          PPLTypeChecker.family(SqlTypeFamily.IGNORE, SqlTypeFamily.CHARACTER));
```
ykmr1224: Why do we need to define separate one from above one? (can we add `or` to above definition?)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2812,30 +2814,629 @@ private RexNode buildWmaRexNode(
   }
 
   /**
-   * Expand command visitor to handle array field expansion. 1. Unnest 2. Join with the original
-   * table to get all fields
+   * Portions of CalciteRelNodeVisitor related to EXPAND / MVEXPAND.
```
ykmr1224: Looks there are very long code for expand/mvexpand. Please extract to a class instead of using code section.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2812,30 +2814,629 @@ private RexNode buildWmaRexNode(
...
+   * Thin delegator for EXPAND that preserves the original external signature.
+   *
+   * <p>This method intentionally remains a trivial wrapper so external call sites and semantics
+   * remain stable.
+   *
+   * @param arrayFieldRex the resolved input reference of the array/collection field
+   * @param arrayFieldName the logical name of the field
+   * @param alias optional alias to rename expanded prefix to
+   * @param context calcite plan context
+   */
```
ykmr1224: Please don't put too many comments. That would just be noise for code reader (and also coding agent).
Please just add enough comment. Self explanatory code which doesn't require comment is the best.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2812,30 +2814,629 @@ private RexNode buildWmaRexNode(
...
+  private void buildExpandRelNode(
+      RexInputRef arrayFieldRex, String arrayFieldName, String alias, CalcitePlanContext context) {
+    buildExpandCore(false, arrayFieldRex, arrayFieldName, alias, /* mvLimit= */ null, context);
+  }
+
+  /**
+   * Thin delegator for MVEXPAND that preserves the original external signature.
```
ykmr1224: ditto

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2812,30 +2814,629 @@ private RexNode buildWmaRexNode(
...
+        newNames.add(alias + fn.substring(arrayFieldName.length()));
+      } else {
+        newNames.add(fn);
+      }
+    }
+    context.relBuilder.rename(newNames);
+  }
+
+  /**
+   * Single core used by both EXPAND and MVEXPAND flows.
```
ykmr1224: ditto

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2812,30 +2814,629 @@ private RexNode buildWmaRexNode(
...
+        return List.of();
+      }
+    } catch (Exception ignored) {
+      // Any unexpected failure: be defensive and return empty list.
+      return List.of();
+    }
+  }
+
+  /**
+   * Build the right-hand RelNode that projects element sub-fields from the rawRight input.
```
ykmr1224: ditto

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2812,30 +2814,629 @@ private RexNode buildWmaRexNode(
...
+        leftBuilt,
+        rawRight,
+        arrayFieldRex,
+        arrayFieldIndexInLeft,
+        correlVariable.get(),
+        preDerivedSubFields);
+  }
+
+  /**
+   * Safe helper to extract struct field descriptors from an element RexNode.
```
ykmr1224: ditto

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,274 @@
...
+        .defaultSchema(schema)
+        .traitDefs((List<RelTraitDef>) null)
+        .programs(Programs.heuristicJoinOrder(Programs.RULE_SET, true, 2));
+  }
+
+  @Test
+  public void testMvExpandBasic() {
+    String ppl = "source=USERS | mvexpand skills";
+    RelNode root;
+    try {
```
ykmr1224: I think we don't need try/catch. Test case would automatically fail when exception is thrown.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,274 @@
...
+        "SELECT `$cor0`.`USERNAME`, `t1`.`skills.name`\n"
+            + "FROM `scott`.`USERS` `$cor0`,\n"
+            + "LATERAL (SELECT `name` `skills.name`\n"
+            + "FROM UNNEST((SELECT `$cor0`.`skills`\n"
+            + "FROM (VALUES (0)) `t` (`ZERO`))) `t0` (`name`, `level`)) `t1`";
+    verifyPPLToSparkSQL(root, expectedSparkSql);
+  }
+
+  @Test
+  public void testMvExpandEmptyOrNullArray() {
```
ykmr1224: Do we need this test case if we don't verify result?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,274 @@
...
+    try {
+      RelNode root = getRelNode(ppl);
+      assertNotNull(root);
+    } catch (Exception e) {
+      fail("mvexpand on empty/null array should not throw, but got: " + e.getMessage());
+    }
+  }
+
+  @Test
+  public void testMvExpandNoArrayField() {
```
ykmr1224: ditto

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,274 @@
...
+    try {
+      RelNode root = getRelNode(ppl);
+      assertNotNull(root);
+    } catch (Exception e) {
+      fail("mvexpand on missing array field should not throw, but got: " + e.getMessage());
+    }
+  }
+
+  @Test
+  public void testMvExpandWithDuplicates() {
```
ykmr1224: ditto

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,274 @@
...
+    try {
+      RelNode root = getRelNode(ppl);
+      assertNotNull(root);
+    } catch (Exception e) {
+      fail("mvexpand with duplicates should not throw, but got: " + e.getMessage());
+    }
+  }
+
+  @Test
+  public void testMvExpandLargeArray() {
```
ykmr1224: ditto

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,274 @@
...
+    try {
+      RelNode root = getRelNode(ppl);
+      assertNotNull(root);
+    } catch (Exception e) {
+      fail("mvexpand on large array should not throw, but got: " + e.getMessage());
+    }
+  }
+
+  @Test
+  public void testMvExpandPrimitiveArray() {
```
ykmr1224: ditto.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,432 @@
...
+import org.opensearch.sql.ppl.PPLIntegTestCase;
+
+/**
+ * Integration tests for mvexpand behavior via Calcite translation.
+ *
+ * <p>This test follows the layout and style of CalciteExpandCommandIT but targets the mvexpand
+ * command semantics and edge cases. The canonical fixture created in init() contains documents used
+ * by the tests. Per-test temporary indices are created for mapping-specific edge cases to keep
+ * tests deterministic and isolated.
+ */
```
ykmr1224: This comment seems not needed.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,432 @@
...
+        // complex: elements where some have both fields, some missing, used to assert flattening
+        "{\"username\":\"complex\",\"skills\":[{\"name\":\"ml\",\"level\":\"expert\"},{\"name\":\"ai\"},{\"level\":\"novice\"}]}",
+        // large: many elements to exercise multiple rows generation
+        "{\"username\":\"large\",\"skills\":["
+            + "{\"name\":\"s1\"},{\"name\":\"s2\"},{\"name\":\"s3\"},{\"name\":\"s4\"},{\"name\":\"s5\"},"
+            + "{\"name\":\"s6\"},{\"name\":\"s7\"},{\"name\":\"s8\"},{\"name\":\"s9\"},{\"name\":\"s10\"}"
+            + "]}",
+        // hetero_types: same sub-field 'level' as number and string to check type inference edge
+        // case
+        "{\"username\":\"hetero_types\",\"skills\":[{\"level\":\"senior\"},{\"level\":3}]}");
```
ykmr1224: Do we need these comments?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,432 @@
...
+
+    // Make indexed documents available for search
```
ykmr1224: ditto

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  @AfterEach
+  public void cleanupAfterEach() throws Exception {
+    // best-effort cleanup for test-local indices
+    try {
+      deleteIndexIfExists(INDEX + "_not_array");
+      deleteIndexIfExists(INDEX + "_missing_field");
+      deleteIndexIfExists(INDEX + "_limit_test");
+      deleteIndexIfExists(INDEX + "_int_field");
```
ykmr1224: Why do we need different indices? Single index should be enough.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,432 @@
...
+            INDEX);
+    JSONObject result = executeQuery(query);
+    verifyDataRows(result, rows("single", "go"));
+  }
+
+  @Test
+  public void testMvexpandEmptyArray() throws Exception {
+    String query =
+        String.format(
+            "source=%s | mvexpand skills | where username='empty' | fields username, skills.name",
```
ykmr1224: It's better doing `where` before `mvexpand` to make the purpose clear. Same for other test cases.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,432 @@
...
+      bulkInsert(idx, "{\"username\":\"u_int\",\"skills\":5}");
+      refreshIndex(idx);
+
+      String query =
+          String.format(
+              "source=%s | mvexpand skills | where username='u_int' | fields username, skills",
+              idx);
+
+      ResponseException ex =
+          org.junit.jupiter.api.Assertions.assertThrows(
```
ykmr1224: Can't we import?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,432 @@
...
+              ResponseException.class, () -> executeQuery(query));
+      String msg = ex.getMessage();
+      org.junit.jupiter.api.Assertions.assertTrue(
+          msg.contains("Cannot expand field 'skills': expected ARRAY type but found VARCHAR"),
+          "Expected SemanticCheckException about non-array field, got: " + msg);
+    } finally {
+      deleteIndexIfExists(idx);
```
ykmr1224: Can't we do it in `@AfterEach` method?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,432 @@
...
+
+      String query =
+          String.format(
+              "source=%s | mvexpand skills | where username='u_missing' | fields username, skills",
+              idx);
+
+      JSONObject result = executeQuery(query);
+      verifyDataRows(result); // expect empty result set for missing field
+    } finally {
+      deleteIndexIfExists(idx);
```
ykmr1224: ditto

integ-test/src/test/java/org/opensearch/sql/legacy/SQLIntegTestCase.java
```diff
@@ -686,6 +686,11 @@ public enum Index {
         "_doc",
         getNestedSimpleIndexMapping(),
         "src/test/resources/nested_simple.json"),
+    MVEXPAND_EDGE_CASES(
+        "mvexpand_edge_cases",
+        "mvexpand_edge_cases",
+        getMappingFile("mvexpand_edge_cases_mapping.json"),
+        "src/test/resources/mvexpand_edge_cases.json"),
```
ykmr1224: I suspect this (and json) is really needed. It seems used only in CalciteExplainIT, but the data does matter for ExplainIT.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAbstractTest.java
```diff
@@ -64,6 +65,21 @@ public CalcitePPLAbstractTest(CalciteAssert.SchemaSpec... schemaSpecs) {
     this.settings = mock(Settings.class);
   }
 
+  public CalcitePPLAbstractTest(Schema customSchema) {
```
ykmr1224: What is this for?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: I was just following the existing pattern for command and function javadcos. 
Noted. will refactor accordingly.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
srikanthpadakanti: I tried. But, Calcite’s OR operator requires both operands to be composite checkers.
The fallback rule I added (IGNORE, CHARACTER) is still a SqlSingleOperandTypeChecker, so Calcite rejects the merge.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Minimized it to use the same - buildExpandRelNode

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: done.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: yeah this is redundant. I deleted them. Thanks

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: done

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLAbstractTest.java
srikanthpadakanti: I have added it to do my unit testing but later decided not to use it. I am going to cleanup that

docs/user/ppl/index.rst
```diff
@@ -48,6 +48,65 @@ The query start with search command and then flowing a set of command delimited
 
 * **Commands**
 
+  - `Syntax <cmd/syntax.rst>`_
```
dai-chen: I think the doc structure changed. Please add your new command below.

srikanthpadakanti: Done.

srikanthpadakanti:
@dai-chen I believe the test failures in CI are unrelated to my changes.

dai-chen:
> @dai-chen I believe the test failures in CI are unrelated to my changes.

Not sure what's going wrong. Could you see if it goes away after merging from `main` again?

PR #4688, opened 2025-10-28
Introduce RelBuilder wrapper for dynamic fields

ykmr1224:

### Description
- Introduce RelBuilderWrapper and RelFieldBuilder to abstract the field operations with dynamic fields.
- RelBuilderWrapper uses delegation to RelBuilder so that it won't allow direct operation on fields without considering dynamic fields.
  - It uses delegation instead of inheritance since RelBuilder internally call `field` or `fields` methods and inheritance cannot prohibit access to some methods.
  - RelBuilderWrapper hide original field/fields operations, and provide some package private method to allow RelFieldBuilder to access raw operations.
- It leaves fixes for several commands to later PR to avoid this PR becomes too big.

### Related Issues

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/DynamicFieldsResultProcessor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/DynamicFieldsResultProcessor.java
@@ -26,7 +26,8 @@ import org.opensearch.sql.executor.ExecutionEngine.Schema.Column;
 
 /** Utility class for processing dynamic fields in query results. */
 @UtilityClass
-public class DynamicFieldsProcessor {
+public class DynamicFieldsResultProcessor {
+
   /**
    * Expand dynamic fields into individual columns in QueryResponse.
    *
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLEvalMaxMinFunctionIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLEvalMaxMinFunctionIT.java
@@ -32,7 +32,7 @@ public class CalcitePPLEvalMaxMinFunctionIT extends PPLIntegTestCase {
         executeQuery(
             String.format(
                 "source=%s | eval new = max(1, 3, age) | fields age, new", TEST_INDEX_DOG));
-    verifySchema(result, schema("age", "bigint"), schema("new", "int"));
+    verifySchema(result, schema("age", "bigint"), schema("new", "bigint"));
     verifyDataRows(result, rows(2, 3), rows(4, 4));
   }
 
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
@@ -34,7 +34,7 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.sql.ast.statement.Explain.ExplainFormat;
 import org.opensearch.sql.calcite.CalcitePlanContext;
 import org.opensearch.sql.calcite.utils.CalciteToolsHelper.OpenSearchRelRunners;
-import org.opensearch.sql.calcite.utils.DynamicFieldsProcessor;
+import org.opensearch.sql.calcite.utils.DynamicFieldsResultProcessor;
 import org.opensearch.sql.calcite.utils.OpenSearchTypeFactory;
 import org.opensearch.sql.calcite.utils.UserDefinedFunctionUtils;
 import org.opensearch.sql.common.response.ResponseListener;
@@ -255,7 +255,7 @@ public class OpenSearchExecutionEngine implements ExecutionEngine {
       ExprType exprType;
       if (fieldType.getSqlTypeName() == SqlTypeName.ANY) {
         if (!values.isEmpty()) {
-          exprType = DynamicFieldsProcessor.inferColumnType(values, columnName);
+          exprType = DynamicFieldsResultProcessor.inferColumnType(values, columnName);
         } else {
           // Using UNDEFINED instead of UNKNOWN to avoid throwing exception
           exprType = ExprCoreType.UNDEFINED;
@@ -268,7 +268,7 @@ public class OpenSearchExecutionEngine implements ExecutionEngine {
     Schema schema = new Schema(columns);
     QueryResponse response = new QueryResponse(schema, values, null);
 
-    QueryResponse processedResponse = DynamicFieldsProcessor.expandDynamicFields(response);
+    QueryResponse processedResponse = DynamicFieldsResultProcessor.expandDynamicFields(response);
 
     listener.onResponse(processedResponse);
   }


Swiddis (approved):

core/src/main/java/org/opensearch/sql/calcite/rel/RelBuilderWrapper.java
```diff
@@ -0,0 +1,537 @@
...
+import org.apache.calcite.tools.RelBuilder.GroupKey;
+import org.apache.calcite.util.ImmutableBitSet;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * RelBuilderWrapper delegates most public methods to the underlying RelBuilder while prohibiting
+ * direct field access methods to maintain consistency with dynamic fields handling.
+ */
+@RequiredArgsConstructor
+public class RelBuilderWrapper {
```
LantaoJin, line 36: why not just leverage the [OpenSearchRelBuilder](https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java#L179) directly?

LantaoJin: If you want a derived class of RexBuilder, we can create one as well in [OpenSearchPrepareImpl.perform()](https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java#L220)

core/src/main/java/org/opensearch/sql/calcite/plan/PPLAggregateConvertRule.java
```diff
@@ -74,8 +76,10 @@ public void onMatch(RelOptRuleCall call) {
 
   public void apply(RelOptRuleCall call, LogicalAggregate aggregate, LogicalProject project) {
 
-    final RelBuilder relBuilder = call.builder();
+    final RelBuilder rawRelBuilder = call.builder();
```
LantaoJin, line 79: `OpenSearchRelBuilder` should be returned here, so if you reuse `OpenSearchRelBuilder`, no need to wrap again.

core/src/main/java/org/opensearch/sql/calcite/rel/RelBuilderWrapper.java
ykmr1224: As I described in the PR description, inheritance didn't work well since there are some internal access within RelBuilder to the methods we want to prohibit.

core/src/main/java/org/opensearch/sql/calcite/plan/PPLAggregateConvertRule.java
ykmr1224: Please see my above comment.

core/src/main/java/org/opensearch/sql/calcite/rel/RelFieldBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public int getStackPosition(int inputCount, int inputOrdinal) {
+    return inputCount - 1 - inputOrdinal;
+  }
+
+  public List<String> getAllFieldNames(int n) {
+    return relBuilder.peek(n).getRowType().getFieldNames();
+  }
+
+  public RexInputRef staticField(String fieldName) {
```
LantaoJin, line 59: how about simplify all `staticField` to `field`?

core/src/main/java/org/opensearch/sql/calcite/rel/RelBuilderWrapper.java
LantaoJin: Override all `fields` of raw RelBuilder in OpenSearchRelBuilder cannot work?

In `CalciteRelNodeVisitor`, all calling to `context.relBuilder.fields/field` are interrupted in `OpenSearchRelBuilder.fields/field`.
But in `calcite.RelBuilder`, all `fields`/`field` will access its raw `fields`/`field` method. <=== do you want to prohibit this?

Delegation might not work either. For example, `RelBuilder.sort()` calls `RelBuilder.field()`. How to prohibit this calling with delegation class?

ykmr1224: > In CalciteRelNodeVisitor, all calling to context.relBuilder.fields/field are interrupted in OpenSearchRelBuilder.fields/field.
> But in calcite.RelBuilder, all fields/field will access its raw fields/field method. <=== do you want to prohibit this?

No, inheritance won't work like that. If we override fields/field in a subclass, even the call from calcite.RelBuilder will call the method in subclass.

> Delegation might not work either. For example, RelBuilder.sort() calls RelBuilder.field(). How to prohibit this calling with delegation class?

The idea is that, if the method won't work as we expect, customize or prohibit (and provide alternative way) the usage in RelBuilderWrapper.

core/src/main/java/org/opensearch/sql/calcite/rel/RelFieldBuilder.java
ykmr1224: I named it as `staticField` so it would be more explicit to access static field.
We want to avoid someone unintentionally omit the existence of dynamic fields.

core/src/main/java/org/opensearch/sql/calcite/rel/RelBuilderWrapper.java
LantaoJin: > No, inheritance won't work like that. If we override fields/field in a subclass, even the call from calcite.RelBuilder will call the method in subclass.

Ok, can this work for you?

```
public class TestInheritance {

  static class RelBuilder {
    public void fields() {
      System.out.println("fields");
      field();
    }
    public void field() {
      System.out.println("field");
    }
    public void otherUnrelated() {
      System.out.println("others");
    }
  }

  static class DerivedRelBuilder extends RelBuilder {
    @Override
    public void fields() {
      System.out.println("Derived fields");
      super.field(); // change to add super if want to call origin
    }
    @Override
    public void field() {
      System.out.println("Derived field");
    }
  }
  
  public static void main(String[] args) {
    RelBuilder relBuilder = new DerivedRelBuilder();
    System.out.println("Test fields:");
    relBuilder.fields();
    System.out.println("Test field:");
    relBuilder.field();
    System.out.println("Test unrelated:");
    relBuilder.otherUnrelated();
  }
}
```

> Test fields:
> Derived fields
> field
> Test field:
> Derived field
> Test unrelated:
> others

ykmr1224: Thanks for the further deep dive.
The issue is, some methods in original RelBuilder calls methods which we want to prohibit. How can we selectively allow the call to the same method and delegate to super?
We can do it if we override the method which calls the methods we want to prohibit, but that would require copying the original implementation from Calcite.

LantaoJin: > We can do it if we override the method which calls the methods we want to prohibit

Now I know why you use a decorator pattern.
A simple way is adding an override method to throw an exception in `OpenSearchRelBuilder`
```
    @Deprecated
    @Override
    public RexInputRef field(String fieldName) {
      throw new ForbiddenMethodCall("field() is not supported in OpenSearchRelBuilder");
    }
```
But it just throws runtime exception, so our UT and IT will help us to avoid using those forbidden methods. To throw compile error, an alternative is creating an annotation with `com.google.errorprone`.

LantaoJin: Overall, the simple approach is merging current `RelFieldBuilder` and `OpenSearchRelBuilder`. But It's okey that we keep the current `RelBuilderWrapper`.

LantaoJin: Oh, I think we still need to use RelBuilder or its derived class OpenSearchRelBuilder because RelBuilder will be used in Calcite optimization rules. For example, https://github.com/opensearch-project/sql/blob/c9a8f47ea993c95bb6b13023de68ab54ddedba00/core/src/main/java/org/opensearch/sql/calcite/plan/PPLAggGroupMergeRule.java#L84

we do have opportunity to get the OpenSearchRelBuilder from `call.builder()` but no way to get the wrapper.

How about just add dynamic methods in OpenSearchRelBuilder and keep the original `field` and `fields` as static field methods.

ykmr1224: Hi Lantao,
Can you revisit my comment?
I think you haven't understood why inheritance won't work.
If we inherit and override method with throwing exception, internal call from other methods in RelBuilder will also fail. And I don't have good idea to workaround this other than reimplementing the method and call super.field instead, which duplicates the implementation from Calcite and not maintainable.

LantaoJin: Yes, you are right, in inheritance, if the method calling chain is A() -> B() -> C(). The overridden method A'() will still call the overridden C'(): A'() -> B() -> C'() . 

Can we just add `@Deprecated` and java doc on the override method to callout this method should not be used in the project?

LantaoJin: Just a suggestion @ykmr1224 . I am okey on the current POC, we can postpone the problem about how to restrict the usage of RelBuilder in optimize rules (Calcite internal).

PR #4690, opened 2025-10-28
bin command error message enhancement

### LINKED ISSUE #4590: [BUG] `bin` Command Validation enhancement on Non-Numeric Fields
### **Description**

When the bin command is applied to a non-numeric field while using numeric span parameters (span, minspan, etc.), the command currently fails with an unclear or misleading error message.
Users cannot easily identify that the root cause is the use of a non-numeric field for numeric binning.

The expected behavior is that the bin command should gracefully skip numeric binning for non-numeric fields and preserve the original field values instead of throwing an error.
This ensures query continuity and prevents unexpected pipeline failures in mixed-type datasets.

---

### **Steps to Reproduce**

1. Execute a query that applies numeric binning to a non-numeric field:

   ```json
   {
     "query": "source=otellogs | bin severityText minspan=10"
   }
   ```

   (`severityText` is a **non-numeric field**, e.g., categorical text.)
2. Observe the returned error message.

---

### **Expected Behavior**

* The `bin` command should continue without error and return the original non-numeric value
  Example:
With this data
```
{\"_time\":1753722000, \"value\":\"1\"},
{\"_time\":1753722017, \"value\":2},
{\"_time\":1753722023, \"value\":\"invalid\"}
```
If we use
```
bin value span=10 | table _time, value, category | sort _time
```
Result:
_time | value | category
-- | -- | --
2025-07-28 17:00:00 | 0-10 | D
2025-07-28 17:00:17 | 0-10 | D
2025-07-28 17:00:23 | invalid | D

---

### **Actual Behavior**

* The current behavior is to throw error message

---

ahkcs:
### Description
When the bin command is applied to a non-numeric field (e.g., string) with numeric span parameters, the command fails with unclear Calcite type validation errors, making it difficult for users to identify the root cause

This PR Added clear validation in BinFieldValidator that throws SemanticCheckException with message,  Added field type checks in all numeric bin handlers before Calcite validation, providing clearer user-facing error messages

### Related Issues
Partially resolves #4590 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/binning/BinFieldValidator.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/binning/BinFieldValidator.java
@@ -6,13 +6,11 @@
 package org.opensearch.sql.calcite.utils.binning;
 
 import java.util.List;
-import org.apache.calcite.rel.type.RelDataType;
 import org.opensearch.sql.ast.expression.Field;
 import org.opensearch.sql.ast.tree.Bin;
 import org.opensearch.sql.calcite.CalcitePlanContext;
-import org.opensearch.sql.calcite.utils.OpenSearchTypeFactory;
 
-/** Utility class for field validation and type checking in bin operations. */
+/** Utility class for bin-specific field operations. */
 public class BinFieldValidator {
 
   /** Extracts the field name from a Bin node. */
@@ -33,10 +31,5 @@ public class BinFieldValidator {
           String.format(
               "Field '%s' not found in dataset. Available fields: %s", fieldName, availableFields));
     }
-  }
-
-  /** Checks if the field type is time-based. */
-  public static boolean isTimeBasedField(RelDataType fieldType) {
-    return OpenSearchTypeFactory.isTimeBasedType(fieldType);
   }
 }


core/src/main/java/org/opensearch/sql/calcite/utils/binning/BinFieldValidator.java
```diff
@@ -59,4 +60,39 @@ public static boolean isTimeBasedField(RelDataType fieldType) {
...
+      throw new SemanticCheckException(
+          String.format(
+              "Cannot apply binning: field '%s' is non-numeric and not time-related, expected"
+                  + " numeric or time-related type",
+              fieldName));
+    }
+  }
+
+  /** Checks if the field type is numeric. */
+  private static boolean isNumericField(RelDataType fieldType) {
```
Swiddis: thought: This is a pretty useful general utility outside of `bin`

Should we define it closer to where `SqlTypeName` is defined? We have lots of hand-rolled field logic for different commands and it's led to some issues when the type list change (e.g. OpenSearch recently introducing `TIMESTAMP_MICROS`).

core/src/main/java/org/opensearch/sql/calcite/utils/binning/BinFieldValidator.java
```diff
@@ -59,4 +60,39 @@ public static boolean isTimeBasedField(RelDataType fieldType) {
     // Check if type string contains EXPR_TIMESTAMP
     return fieldType.toString().contains("EXPR_TIMESTAMP");
   }
+
+  /** Validates that the field type is numeric for numeric binning operations. */
+  public static void validateNumericField(RelDataType fieldType, String fieldName) {
```
Swiddis: suggestion: Instead, load numeric fields into a data class that directly supports binning operations.

When we validate with a `void` method, we don't track any validation information in the type system, so we rely on downstream logic to "know" validation was run. It's easy to forget it or unnecessarily run it multiple times. If we instead put this validation in a constructor for a relevant class, and Bin uses this class, then the fact the class exists means we've definitely run validation.

This adds safety, makes these field operations easier to test, makes useful "numeric-field-only" operations more reusable, and simplifies the bin command's logic.

core/src/main/java/org/opensearch/sql/calcite/utils/binning/BinFieldValidator.java
ahkcs: Good suggestion! I've moved the type checking utilities to OpenSearchTypeFactory.  Future type additions (like TIMESTAMP_MICROS) will only need updates in one place

core/src/main/java/org/opensearch/sql/calcite/utils/binning/BinFieldValidator.java
ahkcs: Great suggestion! I've refactored to use a type-safe validation approach by creating a `BinnableField` class that validates in its constructor and updated all 5 bin handlers to use it.

core/src/main/java/org/opensearch/sql/calcite/utils/binning/BinnableField.java
```diff
@@ -0,0 +1,64 @@
...
+              fieldName));
+    }
+  }
+
+  /**
+   * Returns true if this field requires numeric binning logic (not time-based binning).
+   *
+   * @return true if the field should use numeric binning, false if it should use time-based binning
+   */
+  public boolean requiresNumericBinning() {
```
ykmr1224, line 61: nit: This method was confusing to me. It might be better changing it to `isNumeric` is it is actually used to check if the field is numeric.

ahkcs:
> QQ: what's the expected behavior? In PR description, semantic exception is expected. In issue #4590, the expectation is "the bin command should gracefully skip numeric binning for non-numeric fields and preserve the original field values instead of throwing an error." ?

Hi @dai-chen , this PR is for short-term fix to enhance error handling, our long-term goal would be the expected behavior in #4590

core/src/main/java/org/opensearch/sql/calcite/utils/binning/BinnableField.java
ahkcs: Hi Tomo, this method exists because some parameters such as `minspan`, `start`, `end` currently don't support time-based binning, so they will use this method to check first

LantaoJin:
@ahkcs backport required, thanks.

PR #4696, opened 2025-10-29
Support 'usenull' option in PPL `top` and `rare` commands

### LINKED ISSUE #4684: [BUG] The `top` `rare` command should not return `null`
**What is the bug?**
The current `top` `rare` command return `null` which cannot be pushed to DSL.

For example:
| ID | Type |
|----|------|
| 1  | "200" |
| 2  | "404" |
| 3  | null |
| 4  | "200" |
| 5  | null |
| 6  | null |
| 7  | "200" |
| 8  | "404" |
| 9  | null |
| 10 | "200" |

```
source = t | top Type
```
Returns
| Type | count |
|----|------|
| "200" | 4 |
| null | 4 |
| "404" | 2 |

```
source = t | top Type
```
Returns
| Type | count |
|----|------|
| "200" | 4 |
| "404" | 2 |

**What is the solution your purpose?**

1. add a new argument `usenull=true` (no breaking)
2. just correct the current behavior (treat as a bug)

#4671

LantaoJin:
### Description
* usenull=\<bool\>: optional (since 3.4.0). whether to output the null value. The default value of ``usenull`` is determined by ``plugins.ppl.syntax.legacy.preferred``:
  * When ``plugins.ppl.syntax.legacy.preferred=true``, ``usenull`` defaults to ``true``
  * When ``plugins.ppl.syntax.legacy.preferred=false``, ``usenull`` defaults to ``false``

See #4684 for examples.

### Related Issues
Resolves #4684

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -1128,7 +1128,7 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     Pair<List<RexNode>, List<AggCall>> aggregationAttributes =
         aggregateWithTrimming(groupExprList, aggExprList, context);
     if (toAddHintsOnAggregate) {
-      addBucketNotNullHintToAggregate(context);
+      addIgnoreNullBucketHintToAggregate(context);
     }
 
     // schema reordering
@@ -1864,12 +1864,12 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     List<UnresolvedExpression> aggExprList =
         List.of(AstDSL.alias(countFieldName, AstDSL.aggregate("count", null)));
 
-    // if usenull=false, add a isNotNull before Aggregate and bucketNotNull hint to this Aggregate
+    // if usenull=false, add a isNotNull before Aggregate and the hint to this Aggregate
     Boolean bucketNullable = (Boolean) argumentMap.get(RareTopN.Option.useNull.name()).getValue();
-    boolean toAddBucketNotNullHintOnAggregate = false;
+    boolean toAddHintsOnAggregate = false;
     if (!bucketNullable && !groupExprList.isEmpty()) {
-      toAddBucketNotNullHintOnAggregate = true;
-      // add isNotNull filter before aggregation for non-nullable buckets
+      toAddHintsOnAggregate = true;
+      // add isNotNull filter before aggregation to filter out null bucket
       List<RexNode> groupByList =
           groupExprList.stream().map(expr -> rexVisitor.analyze(expr, context)).toList();
       context.relBuilder.filter(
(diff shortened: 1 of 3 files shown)


docs/user/ppl/cmd/rare.rst
```diff
@@ -26,10 +26,13 @@ rare [rare-options] <field-list> [by-clause] ``(available from 3.1.0+)``
 * rare-options: optional. options for the rare command. Supported syntax is [countfield=<string>] [showcount=<bool>].
 * showcount=<bool>: optional. whether to create a field in output that represent a count of the tuple of values. Default value is ``true``.
 * countfield=<string>: optional. the name of the field that contains count. Default value is ``'count'``.
+* usenull=<bool>: optional (since 3.4.0). whether to output the null value. The default value of ``usenull`` is determined by ``plugins.ppl.syntax.legacy.preferred``:
 
+ * When ``plugins.ppl.syntax.legacy.preferred=true``, ``usenull`` defaults to ``true``
+ * When ``plugins.ppl.syntax.legacy.preferred=false``, ``usenull`` defaults to ``false``
```
penghuo, line 32: [bullet points in blockquote](https://github.com/LantaoJin/search-plugins-sql/blob/291e3fc183871fe4879c195586b3431f844dbe32/docs/user/ppl/cmd/rare.rst), is it expected?

penghuo (approved):

docs/user/ppl/cmd/rare.rst
LantaoJin: Yes. Keep the format same with the [current stats.rst](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/stats.rst#syntax). I tried to use the new format to align with [Update PPL Command Documentation](https://github.com/opensearch-project/sql/pull/4562). But seems there are still many format problems in that PR. [ref](https://github.com/opensearch-project/sql/pull/4562#discussion_r2476160870)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1879,8 +1863,27 @@ public RelNode visitRareTopN(RareTopN node, CalcitePlanContext context) {
...
+    if (!bucketNullable && !groupExprList.isEmpty()) {
+      toAddHintsOnAggregate = true;
+      // add isNotNull filter before aggregation to filter out null bucket
+      List<RexNode> groupByList =
+          groupExprList.stream().map(expr -> rexVisitor.analyze(expr, context)).toList();
+      context.relBuilder.filter(
+          PlanUtils.getSelectColumns(groupByList).stream()
+              .map(context.relBuilder::field)
+              .map(context.relBuilder::isNotNull)
+              .toList());
```
yuancu, line 1879: Why not apply `isNotNull` directly on `groupByList`, but on their underlying input ref? If some operation converts a `null` field to a non-null one, I think it should not be filtered out.

LantaoJin: @yuancu 
Think about query:
```
| eval a = nullif(status, "200") | stats bucket_nullable = false count() by a
```
The `groupByList` contains `RexCall "AS($9, 'a')"`. So we finally build a 
`context.relBuilder.filter(isNotNull($9))` which $9 is `nullif(status, "200")` instead of `status`

PR #4698, opened 2025-10-29
Support wildcard for replace command

### LINKED ISSUE #3975: [RFC] Add replace Command to PPL (Calcite Engine)

ahkcs:
### Description
Support wildcard for replace command

### Related Issues
* Resolves #3975 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/WildcardUtils.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/WildcardUtils.java
@@ -160,7 +160,7 @@ public class WildcardUtils {
    * @param pattern wildcard pattern with escape sequences
    * @return array of pattern parts
    */
-  public static String[] splitWildcards(String pattern) {
+  private static String[] splitWildcards(String pattern) {
     List<String> parts = new ArrayList<>();
     StringBuilder current = new StringBuilder();
     boolean escaped = false;
@@ -194,7 +194,7 @@ public class WildcardUtils {
    * @param str string to count wildcards in
    * @return number of unescaped wildcards
    */
-  public static int countWildcards(String str) {
+  private static int countWildcards(String str) {
     int count = 0;
     boolean escaped = false;
     for (char c : str.toCharArray()) {
--- a/core/src/test/java/org/opensearch/sql/calcite/utils/WildcardUtilsTest.java
+++ b/core/src/test/java/org/opensearch/sql/calcite/utils/WildcardUtilsTest.java
@@ -5,6 +5,11 @@
 
 package org.opensearch.sql.calcite.utils;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import com.google.common.collect.ImmutableList;
 import java.util.List;
 import org.junit.jupiter.api.BeforeEach;
@@ -75,6 +80,32 @@ class WildcardUtilsTest {
   }
 
   @Test
+  void testMatchesWildcardPatternEdgeCases() {
+    // Test null handling
+    assertFalse(WildcardUtils.matchesWildcardPattern(null, "field"));
+    assertFalse(WildcardUtils.matchesWildcardPattern("pattern", null));
+    assertFalse(WildcardUtils.matchesWildcardPattern(null, null));
+
+    // Test empty strings
+    assertTrue(WildcardUtils.matchesWildcardPattern("", ""));
+    assertFalse(WildcardUtils.matchesWildcardPattern("", "field"));
+    assertFalse(WildcardUtils.matchesWildcardPattern("field", ""));
+
+    // Test single wildcard
+    assertTrue(WildcardUtils.matchesWildcardPattern("*", "anything"));
+    assertTrue(WildcardUtils.matchesWildcardPattern("*", ""));
+
+    // Test multiple consecutive wildcards
+    assertTrue(WildcardUtils.matchesWildcardPattern("**", "field"));
+    assertTrue(WildcardUtils.matchesWildcardPattern("a**b", "ab"));
+    assertTrue(WildcardUtils.matchesWildcardPattern("a**b", "axxxb"));
+
+    // Test wildcards at start and end
+    assertTrue(WildcardUtils.matchesWildcardPattern("*field*", "myfield123"));
+    assertTrue(WildcardUtils.matchesWildcardPattern("*field*", "field"));
+  }
+
+  @Test
   void testExpandWildcardPattern() {
     // Test exact match
     testExpansion("firstname", ImmutableList.of("firstname"));
@@ -98,6 +129,20 @@ class WildcardUtilsTest {
   }
 
   @Test
+  void testExpandWildcardPatternEdgeCases() {
+    // Test null handling
+    assertEquals(List.of(), WildcardUtils.expandWildcardPattern(null, availableFields));
+    assertEquals(List.of(), WildcardUtils.expandWildcardPattern("pattern", null));
+    assertEquals(List.of(), WildcardUtils.expandWildcardPattern(null, null));
+
+    // Test empty list
+    assertEquals(List.of(), WildcardUtils.expandWildcardPattern("*", List.of()));
+
+    // Test single wildcard matches all
+    assertEquals(availableFields, WildcardUtils.expandWildcardPattern("*", availableFields));
+  }
+
+  @Test
   void testContainsWildcard() {
     // Test with wildcard
     testContainsWildcard("field*", true);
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
```diff
@@ -0,0 +1,173 @@
...
+
+package org.opensearch.sql.calcite.utils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utility for wildcard-based string replacement in PPL replace command.
+ *
+ * <p>Supports SPL-style wildcard matching where '*' matches zero or more characters. Captured
```
Swiddis: issue: No way to escape asterisks

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
```diff
@@ -0,0 +1,173 @@
...
+   * @param replacement The replacement (may contain wildcards)
+   * @return The replaced string, or original if no match
+   */
+  public static String replaceWithWildcard(String input, String pattern, String replacement) {
+    if (input == null) {
+      return null;
+    }
+
+    // Fast path: no wildcards = literal replacement
+    if (!pattern.contains("*")) {
```
Swiddis: question: What are the perf implications of reprocessing the pattern from scratch on every record?

Ideally it'd be faster to compile the pattern once and reuse it to reduce branching in the inner loop. But I don't know if it's a measurable difference.

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  /**
+   * Match pattern against input and capture wildcard portions.
+   *
+   * @param input The input string
+   * @param pattern The pattern with wildcards
+   * @return List of captured strings (one per wildcard), or null if no match
+   */
+  public static List<String> matchAndCapture(String input, String pattern) {
```
Swiddis: suggestion (if-minor): Compiling an equivalent Regex matcher might be faster and simpler.

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
ahkcs: That's a great catch! For now I am marking it as a limitation in the docs since the use case for literal asterisks is rare. Do you think we should implement escape instead or we can keep it as a limitation?

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
ahkcs: That's a great suggestion! Refactored to use regex matching

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
ahkcs: That's a good catch! Added LRU pattern cache to compile once per unique pattern string instead of compiling on every row

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
```diff
@@ -0,0 +1,137 @@
...
+      return input;
+    }
+
+    return substituteWildcards(replacement, captures);
+  }
+
+  /** Match pattern against input and capture wildcard portions. */
+  public static List<String> matchAndCapture(String input, String pattern) {
+    Pattern compiledPattern =
+        PATTERN_CACHE.computeIfAbsent(pattern, WildcardReplaceUtils::compileWildcardPattern);
```
ykmr1224: What happens if pattern contains regex? I think we need to escape regex.

core/src/main/java/org/opensearch/sql/expression/function/udf/WildcardReplaceFunctionImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.calcite.sql.type.OperandTypes;
+import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.type.SqlTypeFamily;
+import org.opensearch.sql.calcite.utils.WildcardReplaceUtils;
+import org.opensearch.sql.expression.function.ImplementorUDF;
+import org.opensearch.sql.expression.function.UDFOperandMetadata;
+
+/** UDF for wildcard-based string replacement in PPL replace command. */
+public class WildcardReplaceFunctionImpl extends ImplementorUDF {
```
ykmr1224: If we use regex to execute the replace, should we rather utilize existing function for regex replacement? In that case, we just need to compile pattern during planning, and don't need caching during execution.

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
Swiddis: We can keep it as a limitation for now, but I'm a bit concerned about breaking a potential existing usage with no workaround. I don't know if who's using asterisks today, the number is probably low, but it'd be surprise me if it was 0.

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
Swiddis: He's using `Pattern.quote` to escape the literal parts on line 80.

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardReplaceUtils.java
Swiddis: Especially if we're already building the final pattern with regex split (L76), we might be able to just make the split have negative lookahead with `Pattern.split("(?!\\\\)\\*")`

Feel free to skip this if it's more complicated than that

ahkcs: Updated to support escaping asterisks

core/src/main/java/org/opensearch/sql/expression/function/udf/WildcardReplaceFunctionImpl.java
ahkcs: I think that is a great improvement! I have updated, now wildcard patterns convert to regex at planning time. Also updated to use Clacite's REGEXP_REPLACE_3 operator instead of custom UDF,  removed WildcardReplaceFunctionImpl, WildcardReplaceUtils, and runtime caching

LantaoJin:
The bwc-tests failed due to no space left on device which can be fixed by #4716 , no related to this PR but better to merge upstream (after 4716 merged) then rerun.

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardUtils.java
```diff
@@ -92,4 +93,141 @@ private static boolean matchesCompiledPattern(String[] parts, String fieldName)
...
+
+  /**
+   * Converts a wildcard pattern to a regex pattern.
+   *
+   * <p>Example: "*ada" → "^(.*?)ada$"
+   *
+   * @param wildcardPattern wildcard pattern with '*' and escape sequences (\*, \\)
+   * @return regex pattern with capture groups
+   */
+  public static String convertWildcardPatternToRegex(String wildcardPattern) {
```
ykmr1224, line 105: Let's add comprehensive unit tests for the methods in WildcardUtils.

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardUtils.java
```diff
@@ -92,4 +93,141 @@ private static boolean matchesCompiledPattern(String[] parts, String fieldName)
...
+
+  /**
+   * Splits a wildcard pattern into parts separated by unescaped wildcards.
+   *
+   * <p>Example: "a*b*c" → ["a", "b", "c"]
+   *
+   * @param pattern wildcard pattern with escape sequences
+   * @return array of pattern parts
+   */
+  public static String[] splitWildcards(String pattern) {
```
ykmr1224: Should it be private?

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardUtils.java
```diff
@@ -92,4 +93,141 @@ private static boolean matchesCompiledPattern(String[] parts, String fieldName)
...
+    return parts.toArray(new String[0]);
+  }
+
+  /**
+   * Counts the number of unescaped wildcards in a string.
+   *
+   * @param str string to count wildcards in
+   * @return number of unescaped wildcards
+   */
+  public static int countWildcards(String str) {
```
ykmr1224: ditto

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardUtils.java
ahkcs: Added unit tests in WildcardUtilsTest

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardUtils.java
ahkcs: Changed to private

core/src/main/java/org/opensearch/sql/calcite/utils/WildcardUtils.java
ahkcs: Changed to private

PR #4719, opened 2025-11-03
Support access to nested field of struct after fields command

### LINKED ISSUE #3459: [FEATURE][Calcite Engine] Support access to nested fields after command `FIELD`

qianheng-aws:
### Description
Support access to nested field of struct after fields command

### Related Issues
Resolves #3459

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
@@ -177,11 +177,10 @@ public class OpenSearchTypeFactory extends JavaTypeFactoryImpl {
           return TYPE_FACTORY.createArrayType(
               TYPE_FACTORY.createSqlType(SqlTypeName.ANY, nullable), -1);
         case STRUCT:
-          // TODO: should use RelRecordType instead of MapSqlType here
-          // https://github.com/opensearch-project/sql/issues/3459
           final RelDataType relKey = TYPE_FACTORY.createSqlType(SqlTypeName.VARCHAR);
+          // TODO: should we provide more precise type here?
           return TYPE_FACTORY.createMapType(
-              relKey, TYPE_FACTORY.createSqlType(SqlTypeName.BINARY), nullable);
+              relKey, TYPE_FACTORY.createSqlType(SqlTypeName.ANY), nullable);
         case UNKNOWN:
         default:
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -177,11 +177,10 @@ public static RelDataType convertExprTypeToRelDataType(ExprType fieldType, boole
           return TYPE_FACTORY.createArrayType(
               TYPE_FACTORY.createSqlType(SqlTypeName.ANY, nullable), -1);
         case STRUCT:
-          // TODO: should use RelRecordType instead of MapSqlType here
-          // https://github.com/opensearch-project/sql/issues/3459
           final RelDataType relKey = TYPE_FACTORY.createSqlType(SqlTypeName.VARCHAR);
+          // TODO: should we provide more precise type here?
```
LantaoJin, line 181: Non-blocking: Do you have any idea how to specify precise type here? Calcite 1.41.0 has more strict type checking in RexToLixTranslator. If the Map value is another Array type, it cannot cast a ANY to Array.

qianheng-aws: Maybe we can do type-conversion recursively for the nested fields and store them in a customized type? It will has benefit to resolve proper functions for these fields. 

What's more I'm thinking we may be able to deprecate the flatten fields if these nested fields' type are precise enough. It's not worthy to do that now since we have flatten fields and it only takes effect after `FIELD` command. It should be done as another issue for enhancement.

qianheng-aws: In the beginning, I'm thinking using `RelRecordType `  instead of `MapSqlType` to address this issue, which will get better performance and naturally support variant types. 

But looks like using map to support dynamic fields and type tolerance will be our trend. https://github.com/opensearch-project/sql/issues/4349 and https://github.com/opensearch-project/sql/issues/4433

PR #4742, opened 2025-11-04
Fix binning udf resolution / Add type coercion support for binning UDFs

### LINKED ISSUE #4740: [BUG] Binning UDFs aren’t resolved via `PPLFuncImpTable`
**What is the bug?**

Binning UDFs used by the bin command (e.g., `span_bucket`) are not resolved through `PPLFuncImpTable.resolve(...)`. Instead, they’re instantiated via `makeCall`, which bypasses normal resolution. Because these functions are also missing from `BuiltinFunctionName`, resolution fails and throws an exception below.

```
Unsupported function: span_bucket
java.lang.IllegalArgumentException: Unsupported function: span_bucket
	at org.opensearch.sql.expression.function.PPLFuncImpTable.resolve(PPLFuncImpTable.java:514)
```

1. Call the `PPLFuncImpTable.resolve` API with a binning UDF.
2. Observe `IllegalArgumentException: Unsupported function` exception.

1. Binning UDFs are declared in BuiltinFunctionName.
2. They are always constructed via PPLFuncImpTable.resolve(...)

 - Version: OS `2.19-dev` and latest `main`
 - Plugins: OpenSearch PPL

#### Implications

- **Functionality**
    - Type checking is skipped for these functions.
    - Schemaless operations and implicit casting are not applied.
- **Integrations**
    - Bin command in Spark breaks because the UDF is not resolvable through the standard function table.

ahkcs:
### Description
 Fixes #4740 - Binning UDFs are now resolved through `PPLFuncImpTable.resolve()` instead of being instantiated directly via `makeCall()`.

Implemented automatic type coercion to allow string fields with numeric values to work seamlessly with all binning operations (bins, span, minspan, range).

### Related Issues
* Resolves #4740

Type coercion support for binning UDFs #4356

### DIFF
--- a/integ-test/src/test/resources/expectedOutput/calcite/explain_bin_aligntime.yaml
+++ b/integ-test/src/test/resources/expectedOutput/calcite/explain_bin_aligntime.yaml
@@ -3,8 +3,8 @@ calcite:
     LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
       LogicalProject(category=[$0], value=[$1], timestamp=[$2], @timestamp=[$9])
         LogicalSort(fetch=[5])
-          LogicalProject(category=[$1], value=[$2], timestamp=[$3], _id=[$4], _index=[$5], _score=[$6], _maxscore=[$7], _sort=[$8], _routing=[$9], @timestamp=[FROM_UNIXTIME(*(*(FLOOR(/(/(UNIX_TIMESTAMP($0), 3600), 2)), 2), 3600))])
+          LogicalProject(category=[$1], value=[$2], timestamp=[$3], _id=[$4], _index=[$5], _score=[$6], _maxscore=[$7], _sort=[$8], _routing=[$9], @timestamp=[FROM_UNIXTIME(*(*(FLOOR(DIVIDE(DIVIDE(UNIX_TIMESTAMP($0), 3600), 2)), 2), 3600))])
             CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_time_data]])
   physical: |
-    EnumerableCalc(expr#0..3=[{inputs}], expr#4=[UNIX_TIMESTAMP($t3)], expr#5=[3600], expr#6=[/($t4, $t5)], expr#7=[2], expr#8=[/($t6, $t7)], expr#9=[FLOOR($t8)], expr#10=[*($t9, $t7)], expr#11=[*($t10, $t5)], expr#12=[FROM_UNIXTIME($t11)], proj#0..2=[{exprs}], $f3=[$t12])
+    EnumerableCalc(expr#0..3=[{inputs}], expr#4=[UNIX_TIMESTAMP($t3)], expr#5=[3600], expr#6=[DIVIDE($t4, $t5)], expr#7=[2], expr#8=[DIVIDE($t6, $t7)], expr#9=[FLOOR($t8)], expr#10=[*($t9, $t7)], expr#11=[*($t10, $t5)], expr#12=[FROM_UNIXTIME($t11)], proj#0..2=[{exprs}], $f3=[$t12])
       CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_time_data]], PushDownContext=[[PROJECT->[category, value, timestamp, @timestamp], LIMIT->5, LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":5,"timeout":"1m","_source":{"includes":["category","value","timestamp","@timestamp"],"excludes":[]}}, requestedTotalSize=5, pageSize=null, startFrom=0)])
--- a/integ-test/src/test/resources/expectedOutput/calcite_no_pushdown/explain_bin_aligntime.yaml
+++ b/integ-test/src/test/resources/expectedOutput/calcite_no_pushdown/explain_bin_aligntime.yaml
@@ -3,10 +3,10 @@ calcite:
     LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
       LogicalProject(category=[$0], value=[$1], timestamp=[$2], @timestamp=[$9])
         LogicalSort(fetch=[5])
-          LogicalProject(category=[$1], value=[$2], timestamp=[$3], _id=[$4], _index=[$5], _score=[$6], _maxscore=[$7], _sort=[$8], _routing=[$9], @timestamp=[FROM_UNIXTIME(*(*(FLOOR(/(/(UNIX_TIMESTAMP($0), 3600), 2)), 2), 3600))])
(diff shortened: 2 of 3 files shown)


integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4740.yml
```diff
@@ -0,0 +1,144 @@
...
+          - '{"@timestamp":"2024-01-01T00:20:00.000Z","age":45,"balance":5000.0,"name":"Eve"}'
+
+---
+"bin with numeric field using WIDTH_BUCKET - issue 4740":
+  - skip:
+      features:
+        - headers
+        - allowed_warnings
+  - do:
+      allowed_warnings:
```
dai-chen, line 39: is this required for each test?

dai-chen:
Adding `bug` label since enforce-label workflow doesn't recognize `bugFix` label.

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4740.yml
ahkcs: I think `yamlRestTest` is used to verify issue fixed

core/src/main/java/org/opensearch/sql/calcite/utils/PPLOperandTypes.java
```diff
@@ -128,6 +128,32 @@ private PPLOperandTypes() {}
               SqlTypeFamily.NUMERIC,
               SqlTypeFamily.NUMERIC,
               SqlTypeFamily.NUMERIC));
+  public static final UDFOperandMetadata WIDTH_BUCKET_OPERAND =
```
dai-chen, line 138: np: in future we may need more readable function signature/checker spec or docs for internal function usage too.

ahkcs:
Unit tests in CI seems flaky

core/src/main/java/org/opensearch/sql/calcite/utils/PPLOperandTypes.java
ahkcs: Added comments

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4740.yml
```diff
@@ -0,0 +1,144 @@
...
+        - 'Loading the fielddata on the _id field is deprecated and will be removed in future versions. If you require sorting or aggregating on this field you should also include the id in the body of your documents, and map this field as a keyword field that has [doc_values] enabled'
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: source=test_binning_4740 | bin age bins=2 | stats count() by age | sort age
+
+  - is_true: datarows
+  - gte: { total: 1 }
+  - gte: { size: 1 }
```
penghuo: verify actually results.

ahkcs: Updated expected results

dai-chen:
@ahkcs CI is failing. I retried but still failed. Could you double check?

ahkcs:
> @ahkcs CI is failing. I retried but still failed. Could you double check?

Unit tests passed after rebase

penghuo (approved):

docs/user/ppl/cmd/bin.rst
```diff
@@ -530,3 +530,22 @@ PPL query::
     | 28.0-29.0 | 13             |
     +-----------+----------------+
 
+
+Example 20: Type coercion with string fields
+==============================================
+
+The ``bin`` command automatically converts string fields containing valid numeric values to numeric types, enabling binning operations on keyword/text fields. Only strings that represent valid numbers (e.g., "25", "30.5", "100") are supported. Non-numeric strings will cause an error.
```
penghuo: ditto

docs/user/ppl/cmd/bin.rst
```diff
@@ -16,13 +16,13 @@ bin
...
 ============
-| The ``bin`` command groups numeric values into buckets of equal intervals, making it useful for creating histograms and analyzing data distribution. It takes a numeric field and generates a new field with values that represent the lower bound of each bucket.
+| The ``bin`` command groups numeric values into buckets of equal intervals, making it useful for creating histograms and analyzing data distribution. It takes a numeric or time-based field and generates a new field with values that represent the lower bound of each bucket. String fields containing valid numeric values (e.g., "25", "30.5") are automatically converted to numeric types through type coercion.
 
 Syntax
 ============
 bin <field> [span=<interval>] [minspan=<interval>] [bins=<count>] [aligntime=(earliest | latest | <time-specifier>)] [start=<value>] [end=<value>]
 
-* field: mandatory. The numeric field to bin.
+* field: mandatory. The field to bin. Accepts numeric fields, time-based fields, or string fields containing valid numeric values (e.g., "25", "30.5"). String fields with numeric values are automatically coerced to numeric type. Non-numeric strings will result in an error.
```
penghuo: I suggest remove this section.

Type conversion is a generic feature applied across all commands and functions, so it does not need to be explicitly mentioned in each one.

Additionally, enabling binning operations should target string-type fields, not keyword or text fields specifically.

core/src/main/java/org/opensearch/sql/expression/function/udf/binning/MinspanBucketFunction.java
```diff
@@ -73,7 +73,14 @@ public Expression implement(
     /** Minspan bucket calculation. */
     public static String calculateMinspanBucket(
         Number fieldValue, Number minSpanParam, Number dataRange, Number maxValue) {
-      if (fieldValue == null || minSpanParam == null || dataRange == null || maxValue == null) {
+      // Detect NULL from failed type coercion (e.g., CAST("abc" AS DOUBLE) returns NULL)
+      if (dataRange == null || maxValue == null) {
+        throw new IllegalArgumentException(
+            "Cannot apply binning: field contains non-numeric string values. "
+                + "Only numeric types or string types with valid numeric values are supported.");
```
penghuo: It should return a null value. For example, if only one document contains an invalid field, the entire query should not fail.

penghuo (changes requested):

docs/user/ppl/cmd/bin.rst
ahkcs: Updated to remove comments

docs/user/ppl/cmd/bin.rst
ahkcs: Updated to remove this section

core/src/main/java/org/opensearch/sql/expression/function/udf/binning/MinspanBucketFunction.java
ahkcs: I think it makes sense, I have updated to remove the validation logic. I kept the validation logic in `BinnableField` to only verify if it's supported type(Reject truly unsupported types (e.g., BOOLEAN, ARRAY, MAP))

PR #4743, opened 2025-11-05
Support aggregation/window commands with dynamic fields

ykmr1224:

### Description
- Support aggregation/window commands with dynamic fields
  - stats, eventstats, timechart, trendline
- DebugUtils/JsonUtils are just utility class mainly for tests and debugging.

### Related Issues

### Check List


integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsAggregationIT.java
```diff
@@ -0,0 +1,328 @@
...
+            + " type=[QUERY_SIZE_LIMIT])\n"
+            + "      LogicalSort(sort0=[$0], dir0=[ASC-nulls-first])\n"
+            + "        LogicalAggregate(group=[{0}], c=[COUNT()])\n"
+            + "          LogicalProject(name=[$0])\n"
+            + "            CalciteLogicalIndexScan(table=[[OpenSearch, test_dynamic_fields]])\n"
+            + "  physical: |\n"
+            + "    EnumerableLimit(fetch=[200])\n"
+            + "      EnumerableSort(sort0=[$0], dir0=[ASC-nulls-first])\n"
+            + "        EnumerableAggregate(group=[{0}], c=[COUNT()])\n"
+            + "          CalciteEnumerableIndexScan(table=[[OpenSearch, test_dynamic_fields]])\n"
```
penghuo, line 51: no pushdown?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsAggregationIT.java
```diff
@@ -0,0 +1,328 @@
...
+    verifySchema(result, schema("department", "string"), schema("avg", "double"));
+    verifyDataRows(result, rows("Engineering", 77500), rows("Marketing", 65000), rows(null, 60000));
+  }
+
+  @Test
+  public void testEventstatsByDynamicField() throws IOException {
+    String query =
+        source(
+            TEST_INDEX_DYNAMIC,
+            "eventstats count() as c by department" + "| fields id, city, department, c");
```
penghuo: nit, remove " + "

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsAggregationIT.java
```diff
@@ -0,0 +1,328 @@
...
+    verifySchema(result, schema("department", "string"), schema("c", "bigint"));
+    verifyDataRows(result, rows("Engineering", 3), rows("Marketing", 1), rows(null, 1));
+  }
+
+  @Test
+  public void testStatsWithDynamicField() throws IOException {
+    String query =
+        source(
+            TEST_INDEX_DYNAMIC,
+            "stats avg(cast(salary as int)) as avg by department | fields department, avg");
```
penghuo: what is type of salary? Type coercion does not works for avg?

common/src/main/java/org/opensearch/sql/common/utils/DebugUtils.java
```diff
@@ -0,0 +1,73 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.common.utils;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/** Utility class for debugging operations. Should not be committed. */
```
penghuo: remove it?

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsTimechartIT.java
```diff
@@ -0,0 +1,191 @@
...
+
+  @Test
+  public void testTrendlineWithDynamicFieldWithoutCast() throws IOException {
+    String query =
+        source(
+            TEST_INDEX_DYNAMIC,
+            "trendline sma(2, latency) as latency_trend | fields id, latency, latency_trend");
+
+    Throwable e = assertThrows(IllegalArgumentException.class, () -> executeQuery(query));
+    assertEquals("`latency` needs to be specific type. Please cast explicitly.", e.getMessage());
```
penghuo: I am afraid this error message does not help user too much. 
type coercion does not works for sma?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2144,6 +2154,14 @@ public RelNode visitTimechart(
     }
   }
 
+  private UnresolvedExpression addAliasToDynamicFieldAccess(UnresolvedExpression exp) {
```
penghuo: why releated to dynamicFieldAccess? it just alias, right?

common/src/main/java/org/opensearch/sql/common/utils/DebugUtils.java
ykmr1224: I will fix the comment and leave this class. It is not used in the code, but useful while testing and debugging.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: You are right. I will rename it.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsAggregationIT.java
ykmr1224: Currently testing without pushdown. Will test pushdown separately.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsAggregationIT.java
ykmr1224: As the salary is dynamic field, it is typed as ANY. I think current type coercion doesn't work for ANY.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsTimechartIT.java
ykmr1224: I am thinking to work on the automatic type coercion from ANY to specific types separately. This is a workaround until then.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsAggregationIT.java
penghuo: What if we use STRING? The current type coercion rules already support STRING → NUMERIC.
I believe dynamic field support will depend on type coercion regardless.

ykmr1224: Do you mean typing dynamic fields as STRING? (Convert everything into String?)
That should work, but we need to convert back to STRING every time storing to dynamic fields. (since Map can take single value type)

ykmr1224: PR for type coercion: https://github.com/opensearch-project/sql/pull/4753

ykmr1224:
Updated to utilize type coercion.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2089,7 +2094,13 @@ public RelNode visitTimechart(
...
     // Extract parameters for byField case
     UnresolvedExpression byField = node.getByField();
-    String byFieldName = ((Field) byField).getField().toString();
+
+    String byFieldName = ((Field) node.getByField()).getField().toString();
+    if (!context.fieldBuilder.isFieldSpecificType(byFieldName)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "By field `%s` needs to be specific type. Please cast explicitly.", byFieldName));
+    }
```
penghuo: Can we cast to string for groupBy field?

ykmr1224: I realized timechart requires bigger change due to type assigned to `span` function, which prevents automatic type coercion work properly.
Let me address this in a separate PR.

ykmr1224: Found simpler way to solve the problem, and included the change in this PR.

core/src/main/java/org/opensearch/sql/expression/function/udf/SpanFunction.java
```diff
@@ -98,6 +98,18 @@ public Expression implement(
...
+        ScalarFunctionImpl function =
+            (ScalarFunctionImpl)
+                ScalarFunctionImpl.create(
+                    Types.lookupMethod(
+                        SpanFunction.class,
+                        "evalTimestamp",
+                        String.class,
+                        int.class,
+                        String.class));
+        return function.getImplementor().implement(translator, call, RexImpTable.NullAs.NULL);
```
penghuo: Could u explain it more on this, "// if first argument is string, consider it as timestamp"?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2048,6 +2049,9 @@ public RelNode visitTimechart(
       org.opensearch.sql.ast.tree.Timechart node, CalcitePlanContext context) {
     visitChildren(node, context);
 
+    projectDynamicFieldAsString(node.getBinExpression(), context);
+    projectDynamicFieldAsString(node.getByField(), context);
+
```
penghuo, line 2054: is it required for all visitor?

penghuo: could u add a test in CalciteDynamicFieldsTimechartIT to help understand what is correspond logical plan / sql

core/src/main/java/org/opensearch/sql/expression/function/udf/SpanFunction.java
ykmr1224: Updated.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: Added CalcitePPLDynamicFieldsTest.java‎ for spark SQL. Added explains in IT.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLDynamicFieldsTest.java
```diff
@@ -0,0 +1,336 @@
...
+    RelNode root = getRelNode(ppl);
+    String expectedLogical =
+        "LogicalProject(id=[$0], name=[$1], _MAP=[$2])\n"
+            + "  LogicalSort(sort0=[$3], dir0=[ASC-nulls-first])\n"
+            + "    LogicalProject(id=[$0], name=[$1], _MAP=[$2], $f3=[ITEM($2, 'status')])\n"
+            + "      LogicalTableScan(table=[[scott, test_dynamic]])\n";
+    verifyLogical(root, expectedLogical);
+
+    String expectedSparkSql =
+        "SELECT `id`, `name`, `_MAP`\n"
```
penghuo, line 118: The output always include _MAP columns? 
@dai-chen does it works with unified ppl in spark?

dai-chen: Sure, let me check.

ykmr1224: It contains `_MAP` when the query does not explicitly select fields, since it should output all the dynamic fields along with static fields. (You can refer test case: `testProjectStaticFields`)

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsTimechartIT.java
```diff
@@ -0,0 +1,364 @@
...
+    enableCalcite();
+  }
+
+  @Test
+  public void testTimechartByDynamicField() throws IOException {
+    String query = source(TEST_INDEX_DYNAMIC, "timechart span=1d count() by event");
+
+    JSONObject result = executeQuery(query);
+
+    assertExplainYaml(
```
dai-chen, line 38: Can we only assert the part we're interested in?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLDynamicFieldsTest.java
dai-chen: As I understand if we submit such SQL query on S3 table to Spark directly, the changes include at least:

1. Add `_MAP` to Spark table schema
2. Add result expanding logic similarly as `DynamicFieldsResultProcessor.expandDynamicFields()`

Do you have example for writing `_MAP`? I want to check if more changes required.

ykmr1224: @dai-chen 
`_MAP` should be automatically added to the table schema when permissive mode is enabled, or a command generate dynamic fields (like `spath` command without `output` param)

`_MAP` is collected [here](https://github.com/opensearch-project/sql/blob/feature/permissive/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java#L120)
Refer [this PR](https://github.com/opensearch-project/sql/pull/4611) for further context.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsTimechartIT.java
ykmr1224: I've added this per request from @penghuo to add explain verification, and I think it is better keeping whole part to detect when plan is changed.
I would migrate it to separate file once I merge the change and enabled permissive mode in `main` branch. (it is currently enabled only in integration test and cannot use same test base class)

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLDynamicFieldsTest.java
dai-chen: @ykmr1224 I just want to make sure I’m understanding this correctly.

- **Case 1:** For `_MAP` generated from a table, do we need to update the Spark catalog to add it when permissive mode is enabled? When you say "automatically added to the table", it means current OpenSearch schema right?
- **Case 2:** For `_MAP` generated dynamically by a command like `spath`, could you share a concrete example, including:
  - the PPL query, and  
  - the Spark SQL query generated?

Since our approach is to transpile PPL into Spark SQL, I’d like to ensure that all required semantics are encoded in the SQL we generate. Otherwise, we’ll need to estimate the effort for any changes required in the Spark SQL engine.

ykmr1224: @dai-chen 
Case 1: Yes, it is added to OpenSearch schema (specifically to metadata fields). I am not sure how Spark catalog works, but I suppose we need to add `_MAP` to the catalog schema.

Case 2: Here is the sample SQL for ppl `source=EMP | fields ENAME | spath input=ENAME`
```
SELECT `mvappend`(`ENAME`, `JSON_EXTRACT_ALL`(`ENAME`)['ENAME']) `ENAME`, `MAP_REMOVE`(`JSON_EXTRACT_ALL`(`ENAME`), ARRAY ('ENAME')) `_MAP`
FROM `scott`.`EMP`
```

``` `MAP_REMOVE`(`JSON_EXTRACT_ALL`(`ENAME`), ARRAY ('ENAME')) `_MAP` ``` is where _MAP is assigned. (MAP_REMOVE is to dedupe the fields in static field)

Comment (dai-chen) [reply ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLDynamicFieldsTest.java:118]: Yes, case 1 may need some changes and we can focus on case 2. Posted the Spark SQL query generated in my understanding.

```
# Test data
search source=test_events;
25/11/19 11:10:06 WARN UnifiedQueryParser: PPL translated to Spark SQL:
 SELECT *
FROM `spark_catalog`.`default`.`test_events`

@timestamp	host	packets	message
2025-09-08 10:00:00	server1	60	{"category":1, "resource":"A"}
2025-09-08 10:01:00	server1	120	{"category":2, "resource":"B"}
2025-09-08 10:02:00	server1	60	{"category":3, "resource":"C"}
2025-09-08 10:02:30	server2	180	{"category":4, "resource":"D"}

# PPL query
# source=test_events | spath input=message | eval cat = abs(category) * 10

# Spark SQL query expected
spark-sql (default)>
                   > SELECT
                   >   ABS(TRY_CAST(`_MAP`['category'] AS INT) * 10) AS `cat`
                   > FROM (
                   >   SELECT `JSON_EXTRACT_ALL`(`message`) AS `_MAP`
                   >   FROM `test_events`
                   > );
line 2:14 missing ')' at '('
cat
10
20
30
40
```

If this is correct, the only question is expand logic in `DynamicFieldsResultProcessor.expandDynamicFields()`

PR #4746, opened 2025-11-05
Support join with dynamic fields

ykmr1224:

### Description
- Support join with dynamic fields (join / lookup)
- Mainly two modifications were needed:
  - Resolve dynamic fields while resolving qualified name for join conditions
  - Add cast to join conditions to avoid duplicate `compare(Object, Object)` issue in Calcite: https://issues.apache.org/jira/browse/CALCITE-7206

### Related Issues

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/ExtendedRexBuilder.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/ExtendedRexBuilder.java
@@ -197,7 +197,10 @@ public class ExtendedRexBuilder extends RexBuilder {
     return makeCast(target.getType(), node, true, true);
   }
 
-  /** Utility to cast ANY to specific types to avoid compare issue */
+  /**
+   * Utility to cast ANY to specific types to avoid compare issue in Calcite:
+   * https://issues.apache.org/jira/browse/CALCITE-7206
+   */
   RexNode castAnyToAlignTypes(RexNode rexNode, CalcitePlanContext context) {
     return rexNode.accept(
         new RexConverter() {
--- a/core/src/main/java/org/opensearch/sql/calcite/rel/QualifiedNameResolver.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/rel/QualifiedNameResolver.java
@@ -109,43 +109,24 @@ public class QualifiedNameResolver {
         .orElseThrow(() -> getNotFoundException(nameNode));
   }
 
-  private static String joinParts(List<String> parts, int start, int length) {
-    StringBuilder sb = new StringBuilder();
-    for (int i = 0; i < length; i++) {
-      if (start < i) {
-        sb.append(".");
-      }
-      sb.append(parts.get(start + i));
-    }
-    return sb.toString();
-  }
-
-  private static String joinParts(List<String> parts, int start) {
-    return joinParts(parts, start, parts.size() - start);
-  }
-
   private static Optional<RexNode> resolveFieldWithAlias(
       QualifiedName nameNode, CalcitePlanContext context, int inputCount) {
-    List<String> parts = nameNode.getParts();
     log.debug(
-        "resolveFieldWithAlias() called with nameNode={}, parts={}, inputCount={}",
-        nameNode,
-        parts,
-        inputCount);
+        "resolveFieldWithAlias() called with nameNode={}, inputCount={}", nameNode, inputCount);
 
-    if (parts.size() >= 2) {
+    if (nameNode.getPartsCount() >= 2) {
       // Consider first part as table alias
-      String alias = parts.get(0);
+      String alias = nameNode.getPart(0);
       log.debug("resolveFieldWithAlias() trying alias={}", alias);
 
       // Try to resolve the longest match first
-      for (int length = parts.size() - 1; 1 <= length; length--) {
-        String field = joinParts(parts, 1, length);
+      for (int length = nameNode.getPartsCount() - 1; 1 <= length; length--) {
+        String field = nameNode.sub(1, 1 + length);
         log.debug("resolveFieldWithAlias() trying field={} with length={}", field, length);
 
         Optional<RexNode> fieldNode = tryToResolveField(alias, field, context, inputCount);
         if (fieldNode.isPresent()) {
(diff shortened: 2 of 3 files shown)


integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalciteDynamicFieldsJoinIT.java
```diff
@@ -0,0 +1,484 @@
...
+
+    JSONObject result = executeQuery(query);
+
+    System.out.println(result.toString());
+
+    verifyJoinResult(result);
+  }
+
+  @Test
+  public void testJoinDynamicWithStaticWithoutCast1() throws IOException {
```
ykmr1224: Type coercion cannot solve this issue since EQUAL operator is defined as parameters=[Comparable, Comparable]. (Underlying problem is, when Equal(Object, *) is used, Calcite generate invalid code.)
I am thinking to define another definition of EQUAL operator specific for ANY type to workaround this issue.

core/src/main/java/org/opensearch/sql/calcite/rel/RelFieldBuilder.java
```diff
@@ -129,4 +140,117 @@ private RexNode createItemAccess(RexNode field, String itemName) {
...
+  private static boolean isStaticField(String fieldName) {
+    // use startsWith as `_MAP` can be renamed to `_MAP0`, etc. by RelBuilder
+    return !fieldName.startsWith(DYNAMIC_FIELDS_MAP);
+  }
+
+  @Value
+  @AllArgsConstructor
+  private static class JoinedField {
+    boolean isInLeft;
+    boolean isInRight;
```
Swiddis, line 153: Is keeping this as two booleans a better choice than having a left/right enum?

Keeping two separate booleans introduces the trivial illegal state of having both of them be true or false, if a code path accidentally triggers it. Enum might be safer.

core/src/main/java/org/opensearch/sql/calcite/rel/QualifiedNameResolver.java
```diff
@@ -139,6 +152,28 @@ private static Optional<RexNode> resolveFieldWithAlias(
...
+        "resolveDynamicFieldsWithAlias() called with nameNode={}, parts={}, inputCount={}",
+        nameNode,
+        parts,
+        inputCount);
+
+    if (parts.size() >= 2) {
+      // Consider first part as table alias
+      String alias = parts.get(0);
+
+      String fieldName = String.join(".", parts.subList(1, parts.size()));
```
Swiddis: I feel like this logic would be better-placed as an accessor method in `QualifiedName`.

[`DELIMITER = "."` is a constant there](https://github.com/opensearch-project/sql/blob/1ecf42d5028bc3bdefffff5ecec8cbd81a471aa4/core/src/main/java/org/opensearch/sql/ast/expression/QualifiedName.java#L25)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1225,6 +1225,8 @@ public RelNode visitJoin(Join node, CalcitePlanContext context) {
               context.relBuilder.peek(),
               context.relBuilder.literal(context.sysLimit.joinSubsearchLimit())));
     }
+    List<String> leftAllFields = context.fieldBuilder.getAllFieldNames(1);
+    List<String> rightAllFields = context.fieldBuilder.getAllFieldNames(0);
```
Swiddis, line 1229: Not entirely the fault of this PR, but CalciteRelNodeVisitor is already a very large file, and I dislike that we're adding more complexity with this logic.

I'd like if we could move the bulk of this visit logic to a dedicated class and leverage that to simplify the method. I did similar refactoring for SPath via `rewriteAsEval`.

I also think the class would be a better place to encapsulate the left/right distinction, to make it easier to vet that we've applied any relevant logic to both halves.

core/src/main/java/org/opensearch/sql/calcite/ExtendedRexBuilder.java
```diff
@@ -164,10 +181,36 @@ else if ((SqlTypeUtil.isApproximateNumeric(sourceType) || SqlTypeUtil.isDecimal(
...
+          @Override
+          public RexNode visitCall(RexCall call) {
+            if (call.getKind() == SqlKind.EQUALS) {
+              RexNode n0 = call.operands.get(0);
+              RexNode n1 = call.operands.get(1);
+              return super.visitCall((RexCall) equalsWithCastAsNeeded(n0, n1));
+            } else {
+              return super.visitCall(call);
+            }
+          }
```
Swiddis, line 216: question (non-blocking): Can you explain this more? What's special about `EQUALS` here?

core/src/main/java/org/opensearch/sql/calcite/rel/RelFieldBuilder.java
ykmr1224: Field with the same name could exist in both side at the same time, and that's why I ended up with this approach.

core/src/main/java/org/opensearch/sql/calcite/ExtendedRexBuilder.java
ykmr1224: Ah, forgot to add comment. I'll add comment.
It is a workaround for Calcite issue: https://issues.apache.org/jira/browse/CALCITE-7206

core/src/main/java/org/opensearch/sql/calcite/rel/QualifiedNameResolver.java
ykmr1224: Fixed.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: I totally agree with that.
Although, I don't want to do that in this series of PRs as it would make it difficult to merge back to main branch.

PR #4750, opened 2025-11-06
Pushdown sort by complex expressions to scan

### LINKED ISSUE #3912: [FEATURE] Calcite Engine Framework: pushdown sort with non-reference expression
Support push down sort on a non-reference expression, for example:
```
...| sort (a + 1)

...| sort length(a)
```

songkant-aws:
### Description
Add sort performance enhancement of pushing down sort by complex expressions to scan.

### Related Issues
Resolves #3912 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -209,7 +218,40 @@ public void pushDownSort(List<SortBuilder<?>> sortBuilders) {
...
+   * @param rowType The row type for script generation context
+   * @param cluster The RelOptCluster for script generation context
+   */
+  public void pushDownSortExpression(
+      SortExpressionInfo sortExprInfo,
+      Map<String, ExprType> fieldTypes,
+      RelDataType rowType,
+      RelOptCluster cluster) {
+    if (sortExprInfo.isSimpleFieldReference()) {
+      sourceBuilder.sort(SortBuilders.fieldSort(sortExprInfo.getFieldName()));
```
songkant-aws: Need to add sort order and null direction here.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -209,7 +218,40 @@ public void pushDownSort(List<SortBuilder<?>> sortBuilders) {
...
+    PredicateAnalyzer.ScriptQueryExpression scriptExpr =
+        new PredicateAnalyzer.ScriptQueryExpression(
+            sortExprInfo.getExpression(), rowType, fieldTypes, cluster);
+
+    Script script = scriptExpr.getScript();
+    if (script != null) {
+      // Determine the correct ScriptSortType based on the expression's return type
+      ScriptSortType sortType = getScriptSortType(sortExpr.getType());
+
+      sourceBuilder.sort(SortBuilders.scriptSort(script, sortType));
```
songkant-aws: Need to add sort order here and check if script sort supports null direction

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -531,6 +534,16 @@ static boolean sortByFieldsOnly(Sort sort) {
     return !sort.getCollation().getFieldCollations().isEmpty() && sort.fetch == null;
   }
 
+  static boolean sortReferencesExpr(Sort sort, Project project) {
```
LantaoJin, line 545: javadoc

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -867,6 +867,35 @@ public void testSimpleSortExpressionPushDownWithOnlyExprProjected() throws Excep
     assertJsonEqualsIgnoreId(expected, result);
   }
 
+  @Test
+  public void testComplexSortExpressionPushDownExplain() throws Exception {
```
LantaoJin, line 874: Add more ITs:
```
eval age2 = age + balance | fields age, age2 | sort age2
eval age2 = age + balance, age3 = age2 - age, | sort age3
```

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteSortCommandIT.java
```diff
@@ -102,4 +103,80 @@ public void testPushdownSortCastToDoubleExpression() throws IOException {
     verifySchema(result, schema("age", "int"), schema("age2", "double"));
     verifyOrder(result, rows(28, 28d), rows(32, 32d));
   }
+
+  @Test
+  public void testPushdownSortStringExpression() throws IOException {
```
LantaoJin: why not add all ITs in SortCommandIT.java

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteSortCommandIT.java
```diff
@@ -102,4 +103,80 @@ public void testPushdownSortCastToDoubleExpression() throws IOException {
...
+    String ppl =
+        String.format(
+            "source=%s | eval firstname2 = substring(firstname, 0, 3) | sort firstname2 | fields"
+                + " firstname2, firstname",
+            TEST_INDEX_BANK_WITH_NULL_VALUES);
+    if (!isPushdownDisabled()) {
+      String explained = explainQueryToString(ppl);
+      assertTrue(explained.contains("SORT_EXPR->[SUBSTRING($0, 0, 3) ASCENDING NULLS_FIRST]"));
+    }
```
LantaoJin: no necessary

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/ExpandCollationOnProjectExprRule.java
```diff
@@ -47,8 +54,57 @@ public void onMatch(RelOptRuleCall call) {
     final Project project = call.rel(1);
     final RelTraitSet toTraits = converter.getTraitSet();
     final RelCollation toCollation = toTraits.getTrait(RelCollationTraitDef.INSTANCE);
-    final RelTrait fromTrait =
-        project.getInput().getTraitSet().getTrait(RelCollationTraitDef.INSTANCE);
+
+    // Branch 1: Check if complex expressions are already sorted by scan
+    if (handleComplexExpressionsSortedByScan(call, project, toTraits, toCollation)) {
```
LantaoJin: can you add some IT and ExplainIT for this case?

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
```diff
@@ -30,6 +33,9 @@
 import org.apache.calcite.util.mapping.Mappings;
 import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.sql.calcite.utils.OpenSearchTypeFactory;
+import org.opensearch.sql.opensearch.storage.scan.AbstractCalciteIndexScan;
+import org.opensearch.sql.opensearch.storage.scan.context.PushDownType;
+import org.opensearch.sql.opensearch.storage.scan.context.SortExpressionInfo;
 
 @UtilityClass
 public class OpenSearchRelOptUtil {
```
LantaoJin: What is `OpenSearchRelOptUtil` used for?  I don't see any reason to keep it in opensearch module. Can this be merged to `PlanUtlis`?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -425,4 +427,62 @@ public AbstractRelNode pushDownLimit(LogicalSort sort, Integer limit, Integer of
     }
     return null;
   }
+
+  public CalciteLogicalIndexScan pushDownLimitToScan(RexNode limit, RexNode offset) {
```
LantaoJin: javadoc: what's the difference with `CalciteLogicalIndexScan.pushDownLimit`? It confused we have both methods here.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/SortExpressionInfo.java
```diff
@@ -0,0 +1,115 @@
...
+import org.opensearch.sql.opensearch.storage.scan.AbstractCalciteIndexScan;
+
+/**
+ * Information about a sort expression that has been pushed down to OpenSearch. Contains both the
+ * expression and its collation information. For simple field references, stores the field name for
+ * stability across schema changes.
+ */
+@Getter
+@AllArgsConstructor
+public class SortExpressionInfo {
```
LantaoJin: can you rename this class to `SortExprDigest` and move it to package `org.opensearch.sql.opensearch.storage.scan.context`

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
```diff
@@ -274,4 +280,78 @@ private static String generateUniqueName(String baseName, Set<String> usedNames)
...
+  /**
+   * Check if the scan can provide the required sort collation by matching toCollation's mapped
+   * project RexNodes with sort expressions from PushDownContext using a two-pointer approach.
+   *
+   * @param scan The scan RelNode to check
+   * @param project The project node to match expressions against
+   * @param toCollation The required collation to match
+   * @return true if scan can provide the required collation, false otherwise
+   */
+  public static boolean canScanProvideSortCollation(
```
LantaoJin, line 324: can you add unit tests in `OpenSearchRelOptUtilTest`?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/ExpandCollationOnProjectExprRule.java
songkant-aws: All of sort expression pushdown ITs are for this case.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
songkant-aws: It's redundant. I will remove it.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteSortCommandIT.java
songkant-aws: Sometimes I want the assertion on pushed down context as well, which only Calcite supports it. I'm not strong opinion of adding pushdown context on IT. But feel it's probably safer to add it in case some commit may break the pushdown optimization.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/ExpandCollationOnProjectExprRule.java
LantaoJin: nit: which one for example?
and can this predicate move to definition of ExpandCollationOnProjectExprRule.Config (can reduce node copy)

LantaoJin (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/SortExprIndexScanRule.java
```diff
@@ -0,0 +1,238 @@
...
+                        .oneInput(
+                            b1 ->
+                                b1.operand(LogicalProject.class)
+                                    .predicate(Predicate.not(Project::containsOver))
+                                    .oneInput(
+                                        b2 ->
+                                            b2.operand(CalciteLogicalIndexScan.class)
+                                                .predicate(
+                                                    Predicate.not(
+                                                            CalciteLogicalIndexScan
```
qianheng-aws: Seems duplicated with the below predication of `noAggregatePushed`.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/SortExprIndexScanRule.java
```diff
@@ -0,0 +1,238 @@
...
+  public interface Config extends RelRule.Config {
+    SortExprIndexScanRule.Config DEFAULT =
+        ImmutableSortExprIndexScanRule.Config.builder()
+            .build()
+            .withOperandSupplier(
+                b0 ->
+                    b0.operand(LogicalSort.class)
+                        // Pure limit pushdown should be covered by SortProjectTransposeRule and
+                        // OpenSearchLimitIndexScanRule
+                        .predicate(sort -> !sort.collation.getFieldCollations().isEmpty())
```
qianheng-aws: Why not use PlanUtils::sortByFieldsOnly? Is there any reason we support limit here?

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
```diff
@@ -274,4 +311,70 @@ private static String generateUniqueName(String baseName, Set<String> usedNames)
...
+      AbstractCalciteIndexScan scan, Project project, RelCollation toCollation) {
+
+    // Check if the scan has sort expressions pushed down
+    if (scan.getPushDownContext().stream()
+        .noneMatch(operation -> operation.type() == PushDownType.SORT_EXPR)) {
```
qianheng-aws, line 332: Is it possible that the pushed SORT(i.e. SORT_EXPR with all digests are simple expression) can satisfy sort collation here?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/SortExprIndexScanRule.java
```diff
@@ -0,0 +1,238 @@
...
+    CalciteLogicalIndexScan newScan;
+    // If the scan's sort info already satisfies new sort, just pushdown limit if there is
+    if (scan.isTopKPushed()
+        && scanProvidesRequiredCollation
+        && (sort.fetch != null || sort.offset != null)) {
+      Integer limitValue = LimitIndexScanRule.extractLimitValue(sort.fetch);
+      Integer offsetValue = LimitIndexScanRule.extractOffsetValue(sort.offset);
+      newScan = (CalciteLogicalIndexScan) scan.pushDownLimit(sort, limitValue, offsetValue);
+    } else {
+      // Attempt to push down sort expressions
```
qianheng-aws, line 97: We still push down sort although `scanProvidesRequiredCollation` is true? I think we could just remove the sort directly.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1480,12 +1484,14 @@ public Script getScript() {
         throw new UnsupportedScriptException(
             "ScriptQueryExpression requires a valid current time from hook, but it is not set");
       }
+      Map<String, Object> mergedParams = new HashMap<>(params);
```
qianheng-aws: Better use LinkedHashMap here or the final plan may be unstable.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -425,4 +426,46 @@ public AbstractRelNode pushDownLimit(LogicalSort sort, Integer limit, Integer of
...
+              table,
+              osIndex,
+              getRowType(),
+              pushDownContext.cloneWithoutSort());
+
+      // Create action to push down sort expressions to OpenSearch
+      OSRequestBuilderAction action =
+          requestBuilder -> {
+            for (SortExprDigest info : sortExprDigests) {
+              requestBuilder.pushDownSortExpression(
```
qianheng-aws: [question] Will this method throw exception by any chance? We plan to make the request builder transformation process lazy. We should put any logic which may throw exception outside of the lambda function, so it can avoid failure on the final plan.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -225,7 +238,58 @@ public void pushDownSort(List<SortBuilder<?>> sortBuilders) {
...
+            fieldTypes,
+            cluster,
+            Map.of(
+                PlanUtils.NULL_DIRECTION,
+                sortExprDigest.getNullDirection().name(),
+                PlanUtils.DIRECTION,
+                sortExprDigest.getDirection().name()));
+
+    Script script = scriptExpr.getScript();
+    if (script != null) {
```
qianheng-aws: What if script is null? Seems nothing happens. We better add assertion or throw exception if that's never possible.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -166,6 +167,11 @@ public double estimateRowCount(RelMetadataQuery mq) {
         case SORT_AGG_METRICS -> {
           dRows = dRows * .9 / 10; // *.9 because always bucket IS_NOT_NULL
           dCpu += dRows;
+        }
+        case SORT_EXPR -> {
+          @SuppressWarnings("unchecked")
+          List<SortExprDigest> sortKeys = (List<SortExprDigest>) operation.digest();
+          dCpu += NumberUtil.multiply(dRows, 1.1 * sortKeys.size());
```
qianheng-aws: I think we should filter the simple expr.

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
songkant-aws: It's possible. The cost computation has coarse results by multiplying field count with 1.1. So VolcanoPlanner will prefer regular field sort pushdown. 

Another option is to calculate the accurate complex expression count and multiply this count with 1.1. Put a check in the rule to not pushdown if all of expressions are simple expressions

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/SortExprIndexScanRule.java
songkant-aws: I was thinking if first sort by [a, b] expressions is already in scan. The second sort by [a] expression could narrow down the sort effort. We could still allow the pushdown. 

For topK pushdown, it could be also applied. If sort by [a, b], limit 10 is pushed, say the second sort by [a], limit 2 comes in, we should allow it pushdown as well. This may need replace existing SORT_EXPR and LIMIT pushdown context.

But right now, it has the problem that multiple different sort complex expressions doesn't work well yet.

qianheng-aws: > The second sort by [a] expression could narrow down the sort effort. We could still allow the pushdown.

Shall we do the same thing for the above branch then? It only push down limit when scanProvidesRequiredCollation is true.

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
qianheng-aws: I mean, in this query  `... | sort a | eval b = a + 1 | sort b`, when pushing the `sort b`, this method will always return false. But I expect it return true.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/SortExprIndexScanRule.java
songkant-aws: Use `noAggregatePushed` now

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
songkant-aws: Yes, it returns false in this case but doesn't matter. Now, if all of expressions are simple, it will go through regular pushDownSort. Field sort exists in regular traitset and is propagated well. It will be handled by `handleSimpleExpressionFieldSorts` method in ExpandCollationOnProjectExprRule.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/SortExprIndexScanRule.java
songkant-aws: Revisit the logic, for already pushed down topK and if the scan collation satisfies the output collation, we can still leave the old sort in the scan(remove the new sort in the meantime). It's kind of keeping semantics. For regular sort pushdown, we can override by the new sort.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
songkant-aws: Use LinkedHashMap now

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
songkant-aws: As discussed offline, move most of logic that could throw exception outside. And use a lazy pushdownSortSuppliers method instead.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
songkant-aws: It should never happen. I neglect this check.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
songkant-aws: Now, it will only count complex expr. The calculation is fine grained.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -425,4 +438,95 @@ public AbstractRelNode pushDownLimit(LogicalSort sort, Integer limit, Integer of
...
+                default -> null;
+              };
+          sortBuilderSuppliers.add(
+              () -> SortBuilders.fieldSort(digest.getFieldName()).order(order).missing(missing));
+          continue;
+        }
+        RexNode sortExpr = digest.getExpression();
+        assert sortExpr instanceof RexCall : "sort expression should be RexCall";
+        Map<String, Object> directionParams = new LinkedHashMap<>();
+        directionParams.put(PlanUtils.NULL_DIRECTION, digest.getNullDirection().name());
```
qianheng-aws, line 484: Could these 2 params be combined into 1 like MISSING_MIN or MISSING_MAX?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/sort/CalciteNumberSortScript.java
```diff
@@ -0,0 +1,64 @@
...
+  @Override
+  public double execute() {
+    Object value = calciteScript.execute(this.getDoc(), this.sourceLookup)[0];
+    // There is a limitation here when the Double value is exactly theoretical min/max value.
+    // It can't distinguish the ordering between null and exact Double.NEGATIVE_INFINITY or
+    // Double.NaN.
+    if (value == null) {
+      boolean isAscending = direction == Direction.ASCENDING;
+      boolean isNullFirst = nullDirection == NullDirection.FIRST;
+      return isAscending == isNullFirst ? Double.NEGATIVE_INFINITY : Double.NaN;
```
qianheng-aws, line 60: Shouldn't be `Double.POSITIVE_INFINITY` for `Double.NaN`?

songkant-aws: According to Double's constant definition, 
Double.POSITIVE_INFINITY is equal to the value of Double.longBitsToDouble(0x7ff0000000000000L).
Double.NaN is equal to the value of Double.longBitsToDouble(0x7ff8000000000000L).

compareTo() treats NaN is larger. However operator of comparisons like '>', '=', '<' always return false if comparing with NaN because NaN is not a number.

I searched the sorting order online. It indicates most of SQL engine like PostgreSQL will sort NaN as the largest value for Double in ascending order.

qianheng-aws (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
qianheng-aws: This is not a blocking comment. We have chance to do this when resolving conflict with https://github.com/opensearch-project/sql/pull/4795. These 2 PR both pass parameter and should have conflict. Let's see which will be merged first.

songkant-aws:
@LantaoJin It doesn't support sort expr pushdown for SortMergeJoin yet.

SortMergeJoin is optimized by EnumerableMergeJoinRule, which is a physical plan level rule. It will inserts collation to left and right children of join and then sort is transformed to EnumerableSort.

Our rules mostly optimize logical plans. Considering it's a valid case, I think one more enhancement is to add physical plan optimization support(Enumerable convention optimization) as well. I will create an issue to track this enhancement.

PR #4765, opened 2025-11-08
Add `regexp_replace()` function as alias of `replace()`

### LINKED ISSUE #4764: [FEATURE] Support `REGEXP_REPLACE` function
`REGEXP_REPLACE` function is not supported yet. It is required in ClickBench IT.

Support `REGEXP_REPLACE` function.

Spark: https://spark.apache.org/docs/latest/sql-ref-functions-builtin.html#string-functions
MySQL: https://dev.mysql.com/doc/refman/8.4/en/regexp.html
PostgreSQL: https://www.postgresql.org/docs/18/functions-string.html

LantaoJin:
### Description
- Add `regexp_replace()` function as alias of `replace()`
- Rename `regex_match()` to `regexp_match()`, and keep the `regex_match()` as synonyms of `regexp_match()`.
- Change the implementation of `PATTERNS` method to `replace()` from `INTERNAL_REGEXP_REPLACE_3`.

### Related Issues
Resolves #4764 

### Check List

### DIFF
--- a/docs/user/ppl/functions/condition.rst
+++ b/docs/user/ppl/functions/condition.rst
@@ -587,8 +587,6 @@ Argument type: STRING, STRING
 
 Return type: BOOLEAN
 
-Synonyms: `REGEX_MATCH`_
-
 Example::
 
     #os> source=logs | where regexp_match(message, 'ERROR|WARN|FATAL') | fields timestamp, message


docs/user/ppl/functions/condition.rst
```diff
@@ -571,25 +571,27 @@ Example::
...
-Usage: regex_match(string, pattern) returns true if the regular expression pattern finds a match against any substring of the string value, otherwise returns false.
+Usage: regexp_match(string, pattern) returns true if the regular expression pattern finds a match against any substring of the string value, otherwise returns false.
 
 The function uses Java regular expression syntax for the pattern.
 
 Argument type: STRING, STRING
 
 Return type: BOOLEAN
 
+Synonyms: `REGEX_MATCH`_
```
RyanL1997: I checked the rendered file and it seems like this link does not work.

core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
```diff
@@ -330,7 +330,6 @@ public enum BuiltinFunctionName {
   INTERNAL_UNCOLLECT_PATTERNS(FunctionName.of("uncollect_patterns")),
   INTERNAL_GROK(FunctionName.of("grok"), true),
   INTERNAL_PARSE(FunctionName.of("parse"), true),
-  INTERNAL_REGEXP_REPLACE_3(FunctionName.of("regexp_replace_3"), true),
```
RyanL1997, line 333: just for my knowledge: This functionality is being consolidated rather than removed right? - The 3-parameter `regexp` replace functionality is being handled by the existing `REPLACE` function now.

integ-test/src/test/resources/clickbench/queries/q29.ppl
```diff
@@ -3,9 +3,12 @@ SELECT REGEXP_REPLACE(Referer, '^https?://(?:www\.)?([^/]+)/.*$', '\1') AS k,
...
 FROM hits WHERE Referer <> '' GROUP BY k HAVING COUNT(*) > 100000 ORDER BY l DESC LIMIT 25;
 */
+/*
+OpenSearch accepts Json as restful request payload, convert \. to \\. and \1 to \\1
+*/
 source=hits
-| Referer != ''
-| eval k = regexp_replace(Referer, '^https?://(?:www\.)?([^/]+)/.*$', '\1')
+| where Referer != ''
+| eval k = regexp_replace(Referer, '^https?://(?:www\\.)?([^/]+)/.*$', '\\1')
```
RyanL1997, line 11: Any reason we need to double escaping now?

docs/user/ppl/functions/condition.rst
LantaoJin: Okey, there is no link for `REGEX_MATCH` anymore, the `REGEX_MATCH` should be considered as a typo name. I will remove it.

integ-test/src/test/resources/clickbench/queries/q29.ppl
LantaoJin: That is because the query is added to json entity of restful request. Char `\` is invalid without escaping in json body.
https://github.com/opensearch-project/sql/blob/20f2234002b1a3c29cd6aa7ecba30a4dc047870a/integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java#L133

core/src/main/java/org/opensearch/sql/expression/function/BuiltinFunctionName.java
LantaoJin: `INTERNAL_REGEXP_REPLACE_3` is not a public UDF. It is used in `patterns` and `rex` command. So I removed it from `BuiltinFunctionName.java` because `patterns` and `rex` command should be translated to `replace`, an enhanced `REGEXP_REPLACE_3`

songkant-aws (approved):

qianheng-aws (approved):

yuancu (approved):

Issue #3925, opened 2025-07-25
[BUG] Circuit Breaker Triggered by Reverse Operation on Large Datasets


selsong:
**What is the bug?**
The `reverse` operation is triggering a circuit breaker when applied to large datasets. Testing shows that the operation fails consistently with datasets of 12,000 rows or more, but succeeds with 10,000 rows.

**How can one reproduce the bug?**
1. Use a source with a large number of rows (e.g., `source=big5`)
2. Apply a `head` operation to limit the number of rows greater than 10,000
3. Apply the `reverse` operation
4. Execute the query

```
Query: source=big5 | head 30000 | reverse
  Run 1: FAILED (14189ms)
  Run 2: FAILED (3876ms)
  Run 3: FAILED (3912ms)
```
```
Query: source=big5 | head 12000 | reverse
  Run 1: FAILED (3982ms)
  Run 2: FAILED (3933ms)
  Run 3: FAILED (3944ms)
```
```
Query: source=big5 | head 10000 | reverse
Iterations: 3, Timeout: 180s
  Run 1: 993ms
  Run 2: 766ms
  Run 3: 766ms
  Percentiles: P90=993ms | P95=993ms
```

**What is the expected behavior?**
The operation fails with a CircuitBreakingException when the fielddata for `_id` exceeds the configured breaker limit:

```
CircuitBreakingException[[fielddata] Data too large, data for [_id] would be [12027099143/11.2gb], which is larger than the limit of [12025908428/11.1gb]]
```

Full error stack trace:
```
[2025-07-25T22:33:28,014][WARN ][o.o.i.b.fielddata        ] [ip-172-31-35-163] [fielddata] New used memory 12027099143 [11.2gb] for data of [_id] would be larger than configured breaker: 12025908428 [11.1gb], breaking
[2025-07-25T22:33:28,015][ERROR][o.o.s.p.r.RestPPLQueryAction] [ip-172-31-35-163] Error happened during query handling
java.lang.RuntimeException: java.sql.SQLException: exception while executing query: all shards failed
        at org.opensearch.sql.opensearch.executor.OpenSearchExecutionEngine.lambda$execute$6(OpenSearchExecutionEngine.java:203) ~[?:?]
        at java.base/java.security.AccessController.doPrivileged(AccessController.java:319) ~[?:?]
        ...
[...]

PR #4775, opened 2025-11-10
Implement `reverse` performance optimization

### LINKED ISSUE #3924: [FEATURE] Support `reverse` pushdown with Calcite

### LINKED ISSUE #5125: Reverse Optimization fails when creating consecutive sorts due to Calcite physical optimizer merging
## Description

The `reverse` command's sort-collation-reversal optimization does not work correctly when the reversed sort is placed directly on top of the original sort with no intermediate operator between them. Calcite's physical optimizer merges (or eliminates) consecutive `LogicalSort` nodes, causing the reversed direction to be lost.

Related PR: #4775 

## Root Cause

The `reverse` optimization in `CalciteRelNodeVisitor.visitReverse()` detects an existing sort collation and creates a new `LogicalSort` with the reversed direction on top of the original sort. This produces a plan like:

```
LogicalSort(sort0=[$1], dir0=[DESC])      <-- reverse adds this
  LogicalSort(sort0=[$1], dir0=[ASC])     <-- original sort
    ...
```

While this logical plan is correct, Calcite's physical optimizer (e.g., `SortRemoveRule`) recognizes that two consecutive sorts on the same field are redundant and merges them — but it keeps the **original** (lower) sort direction instead of the **top-level** (reversed) one, effectively making the `reverse` a no-op.

## When It Works vs. When It Fails

**Works** — when there is an intermediate operator (filter, project, eval) between the original sort and the reversed sort. The intermediate operator prevents Calcite from merging the two sorts:

```
-- WORKS: sort -> fields (project) -> reverse
source=accounts | sort account_number | fields account_number | reverse

Logical plan:
  LogicalSort(sort0=[$0], dir0=[DESC])     <-- reversed sort
    LogicalProject(account_number=[$0])    <-- intermediate operator prevents merging
      LogicalSort(sort0=[$0], dir0=[ASC])
        LogicalTableScan
```

```
-- WORKS: sort -> where (filter) -> reverse
source=accounts | sort account_number | where balance > 30000 | reverse

Logical plan:
  LogicalSort(sort0=[$0], dir0=[DESC])
    LogicalFilter(condition=[>($4, 30000)])  <-- intermediate operator prevents merging
      LogicalSort(sort0=[$0], dir0=[ASC])
        LogicalTableScan
```

**Fails** — when the reversed sort sits directly on top of the original sort with no intermediate operator:

```
-- FAILS: stats -> sort -> reverse (two consecutive sorts, no operator between them)
source=accounts | stats count() as c by gender | sort gender | reverse

Logical plan:
  LogicalSort(sort0=[$1], dir0=[DESC])     <-- reversed sort
    LogicalSort(sort0=[$1], dir0=[ASC])    <-- original sort (nothing between them)
      LogicalProject(c=[$1], gender=[$0])
        LogicalAggregate(...)

Expected: M, F (descending)
Actual:   F, M (ascending — reverse was ignored)
```

```
-- FAILS: timechart -> reverse (timechart adds a sort, reverse adds another directly on top)
source=events | timechart span=1m count() | reverse

Logical plan:
  LogicalSort(sort0=[$0], dir0=[DESC])     <-- reversed sort
    LogicalSort(sort0=[$0], dir0=[ASC])    <-- timechart's sort (nothing between them)
      LogicalAggregate(...)

Expected: 00:04, 00:03, 00:02, 00:01, 00:00 (descending)
Actual:   00:00, 00:01, 00:02, 00:03, 00:04 (ascending — reverse was ignored)
```

## Affected Scenarios

| Query Pattern | Why It Fails |
|---|---|
| `stats ... \| sort field \| reverse` | Sort after aggregation + reverse = consecutive sorts |
| `timechart ... \| reverse` | Timechart's built-in sort + reverse = consecutive sorts |
| `timechart timefield=X ... \| reverse` | Same as above with custom time field |
| `timechart ... by field \| reverse` | Same as above with group-by |

## Expected Behavior

`reverse` should always flip the sort direction, regardless of whether the reversed sort is directly adjacent to the original sort.

## Suggested Fix

Instead of creating a **new** sort on top of the existing sort, modify the **existing** sort node in-place to flip its direction. This avoids creating consecutive sorts entirely. For example:

```java
// Instead of:
context.relBuilder.sort(reversedCollation);  // adds a NEW sort on top

// Consider replacing the existing sort:
RelNode currentNode = context.relBuilder.build();  // pop current
// Replace the top-level Sort with one that has reversed collation
RelNode modifiedSort = LogicalSort.create(
    ((Sort) currentNode).getInput(),  // keep the sort's child
    reversedCollation,                // use reversed direction
    ((Sort) currentNode).offset,
    ((Sort) currentNode).fetch);
context.relBuilder.push(modifiedSort);
```

This would produce a single sort node with the correct (reversed) direction, eliminating the problem entirely.

## Tests

The following integration tests in `CalciteReverseCommandIT` are affected and currently use `verifyDataRows` (unordered) as a workaround:

- `testReverseAfterAggregationWithSort`
- `testTimechartWithReverse`
- `testTimechartWithCustomTimefieldAndReverse`
- `testTimechartWithGroupByAndReverse`

The corresponding unit tests in `CalcitePPLReverseTest` verify the logical plan is correct — the issue is only in physical execution.

ahkcs:
### Description

Originally from #4056 by @selsong

This PR implements a significant performance optimization for the `reverse` command by eliminating the expensive ROW_NUMBER() window function and implementing a three-tier logic based on query context.

Resolves #3924
Resolves #5125

#### Motivation

The previous implementation used ROW_NUMBER() window function which:
- Required materializing the entire dataset
- Caused excessive memory usage
- Failed on large datasets (100M+ records) with "insufficient resources" errors

#### Solution: Three-Tier Reverse Logic

The `reverse` command now follows context-aware behavior:

1. **With existing sort/collation**: Reverses all sort directions (ASC ↔ DESC)
2. **With @timestamp field (no explicit sort)**: Sorts by @timestamp in descending order
3. **Without sort or @timestamp**: The command is ignored (no-op)

#### Key Implementation Details

- **In-place sort replacement**: When `reverse` is applied directly on top of a sort, the sort node is replaced in-place using `PlanUtils.replaceTop()` rather than stacking a new sort. This avoids a Calcite optimizer issue where consecutive `LogicalSort` nodes get merged, discarding the reversed direction (#5125).
- **Backtracking through non-blocking operators**: When `reverse` is separated from the sort by non-blocking operators (`where`, `eval`, `fields`), the implementation walks the RelNode tree to find the upstream sort and inserts a reversed sort via `RelHomogeneousShuttle`.
- **Blocking operator detection**: Aggregation, Join, SetOp, Uncollect, and window-function Projects are recognized as blocking operators that destroy collation — reverse will not backtrack past these.

---

#### 1. Reverse with Explicit Sort (Primary Use Case)

**Query:**
```
source=accounts | sort +balance, -firstname | reverse
```

**Behavior:** Flips all sort directions: `+balance, -firstname` → `-balance, +firstname`

**Logical Plan:**
```
LogicalSystemLimit(sort0=[$3], sort1=[$1], dir0=[DESC-nulls-last], dir1=[ASC-nulls-first], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(account_number=[$0], firstname=[$1], ...)
    LogicalSort(sort0=[$3], sort1=[$1], dir0=[DESC-nulls-last], dir1=[ASC-nulls-first])
      CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])
```

**Physical Plan:** (efficiently pushes reversed sort to OpenSearch)
```
CalciteEnumerableIndexScan(table=[[OpenSearch, accounts]],
  PushDownContext=[[..., SORT->[
    {"balance": {"order": "desc", "missing": "_last"}},
    {"firstname.keyword": {"order": "asc", "missing": "_first"}}
  ], LIMIT->10000]])
```

---

#### 2. Reverse with @timestamp (Time-Series Optimization)

**Query:**
```
source=time_series_logs | reverse | head 100
```

**Behavior:** When no explicit sort exists but the index has an @timestamp field, reverse automatically sorts by @timestamp DESC to show most recent events first.

**Use Case:** Common pattern in log analysis - users want recent logs first

**Logical Plan:**
```
LogicalSystemLimit(sort0=[$0], dir0=[DESC], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(@timestamp=[$0], category=[$1], value=[$2])
    LogicalSort(sort0=[$0], dir0=[DESC])
      CalciteLogicalIndexScan(table=[[OpenSearch, time_data]])
```

---

#### 3. Reverse Ignored (No-Op Case)

**Query:**
```
source=accounts | reverse | head 100
```

**Behavior:** When there's no explicit sort AND no @timestamp field, reverse is ignored. Results appear in natural index order.

**Rationale:** Avoid expensive operations when reverse has no meaningful semantic interpretation.

**Logical Plan:**
```
  LogicalProject(account_number=[$0], firstname=[$1], ...)
    CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])
```

Note: No sort node is added - reverse is completely ignored.

---

#### 4. Double Reverse (Cancellation)

**Query:**
```
source=accounts | sort +balance, -firstname | reverse | reverse
```

**Behavior:** Two reverses cancel each other out, returning to original sort order.

**Logical Plan:**
```
LogicalSystemLimit(sort0=[$3], sort1=[$1], dir0=[ASC-nulls-first], dir1=[DESC-nulls-last], fetch=[10000])
  LogicalProject(account_number=[$0], firstname=[$1], ...)
    LogicalSort(sort0=[$3], sort1=[$1], dir0=[ASC-nulls-first], dir1=[DESC-nulls-last])
      CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])
```

Final sort order matches original query: `+balance, -firstname`

---

#### 5. Backtracking Through Non-Blocking Operators

**Query:**
```
source=accounts | sort age | where age > 20 | reverse
```

**Behavior:** Reverse walks past the `where` (filter) node to find and reverse the upstream sort.

**Logical Plan:**
```
LogicalSystemLimit(sort0=[$8], dir0=[DESC-nulls-last], fetch=[10000])
  LogicalProject(account_number=[$0], firstname=[$1], ...)
    LogicalSort(sort0=[$8], dir0=[DESC-nulls-last])
      LogicalSort(sort0=[$8], dir0=[ASC-nulls-first])
        LogicalFilter(condition=[>($8, 20)])
          CalciteLogicalIndexScan(table=[[OpenSearch, accounts]])
```

---

#### 6. Reverse After Aggregation (Blocking Operator)

**Query:**
```
source=accounts | stats count() as cnt by gender | sort cnt | reverse
```

**Behavior:** Aggregation is a blocking operator. The sort after aggregation is visible to reverse and gets flipped.

---

#### 7. Timechart with Reverse

**Query:**
```
source=web_logs | timechart count() as cnt | reverse
```

**Behavior:** Timechart produces a time-bucketed aggregation with an implicit sort. Reverse flips the time order from ASC to DESC.

### Related Issues
Resolves #3924
Resolves #5125

### Check List


ahkcs:
> QQ: I recall the major comment on original PR is early optimization in analyzer layer. Is this new PR trying to address the concern? Ref: [#4056 (comment)](https://github.com/opensearch-project/sql/pull/4056#discussion_r2345168887)

Hi Chen, I think that's a valid concern. However, after trying it out, I think it has significant complexity comparing to the current approach. I think `CalciteRelNodeVisitor` is used as a logical plan builder that constructs the logical representation of the query, so I think optimization can also happen here. In our approach, our `visitReverse` is choosing LogicalSort(reversed) vs LogicalSort(ROW_NUMBER), and I think this is appropriate for logical plan builder.  If we moved the optimization to Calcite rule, we'd be doing something more complex - starting with a naive representation (always ROW_NUMBER) and rewriting it. That adds significant complexity.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -652,19 +653,30 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
+        context.relBuilder.getCluster().getMetadataQuery().collations(context.relBuilder.peek());
+    RelCollation collation = collations != null && !collations.isEmpty() ? collations.get(0) : null;
+
+    if (collation != null && !collation.getFieldCollations().isEmpty()) {
+      // If there's an existing sort, reverse its direction
+      RelCollation reversedCollation = PlanUtils.reverseCollation(collation);
+      context.relBuilder.sort(reversedCollation);
+    } else {
+      // Fallback: use ROW_NUMBER approach when no existing sort
+      RexNode rowNumber =
```
LantaoJin: Just an idea: How about ignore the reverse command if there is not existing collations or no IMPLICIT_FIELD_TIMESTAMP in rowType.
If there is IMPLICIT_FIELD_TIMESTAMP in rowType, no need to add `ROW_NUMBER`

ahkcs: Thanks for the suggestion! I think it makes sense since `reverse` command may not have meaningful semantics if there's no ordering.

ahkcs: Updated the implementation: Now `reverse` command is ignored if no collations/@timestamp found
Updated the test cases and documentation as well

ahkcs:
> Can you add benchmark results on before VS after?

Test is done against big5_v2_first100m dataset(The first 100m docs of big5)

Before: Query failed to execute even with head 10
```
Query: source=big5_v2_first100m | sort +`@timestamp` | reverse | head <size>

  Running single_sort_reverse_10 (5 iterations)... ERROR: {
  "error": {
    "details": "java.sql.SQLException: exception while executing query: insufficient resources to run the query, quit.",
    "type": "RuntimeException"
  },
  "status": 500
}

```

After:

```
=== Test 1: Single Field Sort + Reverse ===
Query: source=big5_v2_first100m | sort +`@timestamp` | reverse | head <size>

  Running single_sort_reverse_10 (5 iterations)... Done
  Running single_sort_reverse_100 (5 iterations)... Done
  Running single_sort_reverse_1000 (5 iterations)... Done

Size         Avg (ms)   P50 (ms)   P90 (ms)   Min (ms)   Max (ms)
---------- ---------- ---------- ---------- ---------- ----------
10                496        495        502        490        502
100               501        498        516        495        516
1K                577        576        586        571        586

=== Test 2: Single Field Sort DESC + Reverse ===
Query: source=big5_v2_first100m | sort -`@timestamp` | reverse | head <size>

  Running single_sort_desc_reverse_10 (5 iterations)... Done
  Running single_sort_desc_reverse_100 (5 iterations)... Done
  Running single_sort_desc_reverse_1000 (5 iterations)... Done

Size         Avg (ms)   P50 (ms)   P90 (ms)   Min (ms)   Max (ms)
---------- ---------- ---------- ---------- ---------- ----------
10                422        423        424        419        424
100               436        438        439        434        439
1K                515        507        548        503        548

=== Test 3: Multi-field Sort + Reverse ===
Query: source=big5_v2_first100m | sort +`host.name`, -`@timestamp` | reverse | head <size>

  Running multi_sort_reverse_10 (5 iterations)... Done
  Running multi_sort_reverse_100 (5 iterations)... Done
  Running multi_sort_reverse_1000 (5 iterations)... Done

Size         Avg (ms)   P50 (ms)   P90 (ms)   Min (ms)   Max (ms)
---------- ---------- ---------- ---------- ---------- ----------
10                610        587        699        582        699
100               601        599        612        590        612
1K                665        666        667        661        667

=== Test 4: Double Reverse (should cancel out) ===
Query: source=big5_v2_first100m | sort +`@timestamp` | reverse | reverse | head <size>

  Running double_reverse_10 (5 iterations)... Done
  Running double_reverse_100 (5 iterations)... Done
  Running double_reverse_1000 (5 iterations)... Done

Size         Avg (ms)   P50 (ms)   P90 (ms)   Min (ms)   Max (ms)
---------- ---------- ---------- ---------- ---------- ----------
10                441        423        514        420        514
100               436        437        445        427        445
1K                506        502        521        502        521

=== Test 5: Sort + Reverse with Filter ===
Query: source=big5_v2_first100m | where `host.name` != 'unknown' | sort +`@timestamp` | reverse | head <size>

  Running filter_sort_reverse_10 (5 iterations)... Done
  Running filter_sort_reverse_100 (5 iterations)... Done
  Running filter_sort_reverse_1000 (5 iterations)... Done

Size         Avg (ms)   P50 (ms)   P90 (ms)   Min (ms)   Max (ms)
---------- ---------- ---------- ---------- ---------- ----------
10                532        511        612        507        612
100               518        520        532        507        532
1K                589        587        598        585        598

=== REVERSE COMMAND PERFORMANCE TEST COMPLETE ===

```

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteReverseCommandIT.java
```diff
@@ -90,21 +99,129 @@ public void testReverseWithComplexPipeline() throws IOException {
...
+        executeQuery(
+            String.format("source=%s | fields account_number | reverse | head 3", TEST_INDEX_BANK));
+    verifySchema(result, schema("account_number", "bigint"));
+    // Without sort or @timestamp, reverse is ignored, so data comes in natural order
+    // The first 3 documents in natural order (ascending by account_number)
+    verifyDataRowsInOrder(result, rows(1), rows(6), rows(13));
+  }
+
+  @Test
+  public void testReverseWithTimestampField() throws IOException {
```
LantaoJin, line 195: Can you add some ITs for `streamstats`? the `streamstats` always sort on `__stream_seq__` to keep the output ordering as same as input's. We'd better add some ITs to verify the pipe `streamstats | reverse`

ahkcs: Added ITs for `streamstats`, since `reverse` command only works when there's a detectable collation now, using `streamstats` only will make `reverse` command ignored. Do we want to keep it this way?

LantaoJin: No, the streamstats contains collation `__stream_seq__` ASC

ahkcs: @LantaoJin After checking, it seems like `streamstats` command has this logic here:
```
 if (hasGroup) {
      context.relBuilder.sort(context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_STREAMSTATS));
      context.relBuilder.projectExcept(context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_STREAMSTATS));
    }
```
where `projectExcept` removes __stream_seq__ from the schema, which makes the `reverse` command unable to detect the collation.

songkant-aws: I'm curious what's the behavior of other pipe language. It seems a common use case where the final project may not choose to output the intermediate sort field. 

Like a query `source table | fields a, b | sort a | fields b`

ahkcs: It does seem common. I think for the mentioned query, it still makes sense to make `reverse` command an no-op since the sorted field is excluded from the final project. Do you think we should keep our current behavior for `reverse`?

songkant-aws: I think it depends on what feature we're going to align with. Like looking for user's query patterns or directly seeking feedbacks. Run some similar queries on other engines and figure out what's the behavior. It's probably not an emergent issue to be resolved in this PR. 

I think the question is if there is no explicit sort collation of the child but the output rows are indeed sorted in some order, do we want to reverse the implicit order?

LantaoJin: > @LantaoJin After checking, it seems like `streamstats` command has this logic here:
> 
> ```
>  if (hasGroup) {
>       context.relBuilder.sort(context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_STREAMSTATS));
>       context.relBuilder.projectExcept(context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_STREAMSTATS));
>     }
> ```
> 
> where `projectExcept` removes **stream_seq** from the schema, which makes the `reverse` command unable to detect the collation.

Okey, this is the same problem with current implementation for
```
source table | sort a | fields b | reverse
```

Any idea on this? @ahkcs @songkant-aws . We'd better fix it, at least for `streamstats`.

LantaoJin: Another possible solution would be pushdown the reverse to sort `_doc` desc when `@timestamp` is absent. Image the case of PIT search, we can't reverse such a huge data in memory, only pushdown sort tiebreaker (`_doc` + `_sharddoc`) can work with search_after API.

songkant-aws: To reverse the implicit collation generated by previous operators, I think a simple solution is to backtrack the first RelNode's non-empty collation and insert a reverse-order sort there. It could be done at building RelNode stage.

For example,
`source table | sort a | fields b | reverse` could be rebuilt with `source table | sort a | sort -a | fields b`
And we can stop backtrack when we encounter blocking operators like aggregation, join, etc.

songkant-aws: I tried the `_doc` pushdown. It could work. But I doubt the user realize it is sorted by that. If there is no sorting before `reverse`. I feel OK to abandon it as we don't guarantee the reverse order if the output order seems to be non-deterministic.

ahkcs: I have updated the implementation to add backtracking logic as @songkant-aws suggested, it works on both `streamstats` and `source table | sort a | fields b | reverse` scenario

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -669,25 +671,115 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
+   *
+   * @param node the starting RelNode to backtrack from
+   * @return the collation found, or null if no sort or blocking operator encountered
+   */
+  private RelCollation backtrackForCollation(RelNode node) {
+    while (node != null) {
+      // Check for blocking operators that destroy collation
+      if (node instanceof Aggregate
+          || node instanceof org.apache.calcite.rel.core.Join
+          || node instanceof Correlate) {
```
songkant-aws: Could be simplified to `Aggregate || BiRel || SetOp`. It will cover more cases including set operations like union.

songkant-aws: Thanks for the update! I go through all possible RelNodes built by RelBuilder and figure out more corner cases used in current PPL. We need to stop backtracking as well by then.
1. Uncollect RelNode (it's usually used as a child of LogicalCorrelate, so we should be fine in most cases)
2. When project contains window functions, aka RexOver, the ordering will be determined by window function's order by and partitions. So we should not destroy its output order as well. 

For the other part, it seems fine for now. Please also add some complex query combined with different operators in tests.

ahkcs: Updated, also added UTs and ITs for different operators

songkant-aws (approved):

songkant-aws:
LGTM. Please get other signoffs.

yuancu:
Hi @ahkcs , #4784 allows user to specify a timestamp field in `timechart` command. For this specific case, maybe we need to use the specified timefield instead of the hard-coded `@timestamp`.

Although I doubt that there isn't much impact because all timechart commands have a sort at the end of their plans, making them to fall into your first tier. Can you please double check?

songkant-aws:
@ahkcs any progress on this issue? Related to this original issue: https://github.com/opensearch-project/sql/issues/3925

ahkcs:
There's some Calcite physical optimizer merging issue
Created github issue: https://github.com/opensearch-project/sql/issues/5125 for the details

ahkcs:
@songkant-aws Updated PR with a new approach to resolve the issue mentioned, details in the PR description

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -681,25 +686,162 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
+
+      // Project with window functions has ordering determined by the window's ORDER BY clause
+      // We should not destroy its output order by inserting a reversed sort
+      if (node instanceof LogicalProject && ((LogicalProject) node).containsOver()) {
+        return null;
+      }
+
+      // Check for Sort node with collation
+      if (node instanceof org.apache.calcite.rel.core.Sort) {
+        org.apache.calcite.rel.core.Sort sort = (org.apache.calcite.rel.core.Sort) node;
```
songkant-aws: Nit: Use imported class. -> node instanceof Sort

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -681,25 +686,162 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
+   *
+   * @param root the root of the tree to rebuild
+   * @param reversedCollation the reversed collation to insert
+   * @param context the Calcite plan context
+   * @return the rebuilt tree with reversed sort inserted
+   */
+  private RelNode insertReversedSortInTree(
+      RelNode root, RelCollation reversedCollation, CalcitePlanContext context) {
+    return root.accept(
+        new org.apache.calcite.rel.RelHomogeneousShuttle() {
```
songkant-aws: Nit: Same as above comment. Use imported class in this method

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -681,25 +686,162 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
+        context.relBuilder.getCluster().getMetadataQuery().collations(context.relBuilder.peek());
+    RelCollation collation = collations != null && !collations.isEmpty() ? collations.get(0) : null;
+
+    if (collation != null && !collation.getFieldCollations().isEmpty()) {
+      // If there's an existing sort, reverse its direction
+      RelCollation reversedCollation = PlanUtils.reverseCollation(collation);
+      RelNode currentNode = context.relBuilder.peek();
+      if (currentNode instanceof org.apache.calcite.rel.core.Sort) {
+        org.apache.calcite.rel.core.Sort existingSort =
+            (org.apache.calcite.rel.core.Sort) currentNode;
```
songkant-aws: Nit: Same here. Use imported class in this method

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -681,25 +686,162 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
+            && existingSort.fetch == null
+            && existingSort.offset == null) {
+          // Pure collation sort (no fetch/offset) - replace in-place to avoid consecutive
+          // sorts. Calcite's physical optimizer merges consecutive LogicalSort nodes and may
+          // discard the reversed direction. Replacing in-place avoids this issue.
+          RelCollation reversedFromSort = PlanUtils.reverseCollation(existingSort.getCollation());
+          RelNode replacedSort =
+              org.apache.calcite.rel.logical.LogicalSort.create(
+                  existingSort.getInput(), reversedFromSort, null, null);
+          PlanUtils.replaceTop(context.relBuilder, replacedSort);
```
songkant-aws, line 858: Is this observation correct? The logic is weird because that means an explicit query like `source =t | sort -a | sort a` could be equivalent to the query of `source=t | sort -a`. I think it's not merge. The worst case is it needs double sort in case of no pushdown. But in pushdown case, I think the sort will be pushed twice. And finally the top sort will override the scan's traitset. 

I checked the PR out in my local env. When I remove this piece of code, the explainIT still gives me the expected collation in physical plan.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -681,25 +686,162 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
+                return org.apache.calcite.rel.logical.LogicalSort.create(
+                    visitedBarrier, reversedCollation, null, null);
+              }
+              // Found a collation Sort - insert reversed sort on top of it
+              if (sort.getCollation() != null
+                  && !sort.getCollation().getFieldCollations().isEmpty()) {
+                sortFound = true;
+                RelNode visitedSort = super.visit(other);
+                return org.apache.calcite.rel.logical.LogicalSort.create(
+                    visitedSort, reversedCollation, null, null);
```
songkant-aws: If the merge sort theory below is correct(I think it needs careful verification), this part should replace the visitedSort with reversedSort. The current behavior is appending visitedSort to the child of reversedSort, which creates consecutive sorts.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: The in-place replacement is intentional and necessary. The issue is that explainIT only shows the logical plan before Calcite's physical optimizer runs. During actual execution, SortRemoveRule merges consecutive LogicalSort nodes and keeps the lower sort's direction, effectively discarding the reversed one. So for a query like sort a | reverse, the plan becomes Sort(ASC) → Sort(DESC), and the optimizer collapses it back to Sort(ASC).
                                                                                                                                                                                                                                                                                                                   
I confirmed this by running the integration tests (CalciteReverseCommandIT) — they fail without the in-place replacement. The workaround replaces the existing Sort node directly instead of stacking a new one on top, so there are no consecutive sorts for the optimizer to merge.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Good catch!  I've updated this to replace the sort in-place: instead of super.visit(other) (which returns the Sort node and stacks on top), it now uses sort.getInput().accept(this) to visit only the children, then creates a new Sort with the reversed collation using those children as input. This is the same in-place replacement pattern used in visitReverse above.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: detailed description regarding this issue: https://github.com/opensearch-project/sql/issues/5125

songkant-aws: I think the conclusion is not true. Explained plans in explainIT include both original logical plan and optimized physical plan. Calcite's VolcanoPlanner optimizes both logical plans and physical plans, then the final optimized physical plan will generate a compiled Java code to execute the query. Rules never execute the query.

The SortRemoveRule replaces an explicit sort operator with a trait requirement on the child RelSubset. It doesn't do the simple merge. It assigns ordering trait to different RelSubsets. VolcanoPlanner will compute the cheapest plan from different RelSubsets. A simplified process is like 1st round SortRemoveRule changes Sort(1 Desc) - Sort(1 Asc) - RelSubset#0(NONE) => Sort(1 Desc) - RelSubset#1(1 Asc), 2nd round SortRemoveRule changes Sort(1 Desc) - RelSubset#2(1 Asc) => RelSubset#3(1 Desc). The final plan must satisfy the (1 Desc) ordering requirements.

The actual root cause is there is a shallow copy bug in AggPushDownAction. In case of `testReverseAfterAggregationWithSort` test case failure you mentioned in the issue. The actual optimized physical plan is:
```
// query: source=%s | stats count() as c by gender | sort gender | reverse
calcite:
  logical: |
    LogicalSystemLimit(sort0=[$1], dir0=[DESC-nulls-last], fetch=[10000], type=[QUERY_SIZE_LIMIT])
      LogicalSort(sort0=[$1], dir0=[DESC-nulls-last])
        LogicalSort(sort0=[$1], dir0=[ASC-nulls-first])
          LogicalProject(c=[$1], gender=[$0])
            LogicalAggregate(group=[{0}], c=[COUNT()])
              LogicalProject(gender=[$4])
                CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_bank]])
  physical: |
    CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_bank]], PushDownContext=[[AGGREGATION->rel#50:LogicalAggregate.NONE.[](input=RelSubset#49,group={0},c=COUNT()), PROJECT->[c, gender], SORT->[1 DESC LAST], LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"composite_buckets":{"composite":{"size":10000,"sources":[{"gender":{"terms":{"field":"gender.keyword","missing_bucket":true,"missing_order":"first","order":"asc"}}}]}}}}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
```
You can see PushDownContext has correct SORT action, aka `SORT->[1 DESC LAST]`, which means SortRemoveRule didn't pick the ASC-nulls-first as winner. The problem is the agg request builder doesn't have correct ordering on buckets. You can easily verify SortRemoveRule doesn't do anything wrong by manually removing our own rules like `SortIndexScanRule`. The `testReverseAfterAggregationWithSort` will pass without our `SortIndexScanRule`.

Regardless of the actual bug. I think your current approach still has some value. The collapse of consecutive sorts will reduce VolcanoPlanner's optimization space size. You can keep the current design.

songkant-aws (approved):
LGTM so far. Please let others review and approve.

songkant-aws:
@LantaoJin Please also review this PR when you're free, thanks!

PR #4777, opened 2025-11-11
Fix bug that `Streamstats` command incorrectly treats null as a valid group

### LINKED ISSUE #4751: [BUG] Streamstats command incorrectly treats null as a valid group
**What is the bug?**
When using the `streamstats by <field>` clause, the current implementation treats `null` values as a valid group key and computes separate statistics for them. Which is not consistent with the expected behavior.

1. Set data:

```
PUT stream_test_null
{
  "mappings": {
    "properties": {
      "name": {
        "type": "keyword"
      },
      "age": {
        "type": "integer"
      },
      "state": {
        "type": "text",
        "fields": {
          "keyword": {
            "type": "keyword",
            "ignore_above": 256
          }
        }
      },
      "country": {
        "type": "text"
      },
      "year": {
        "type": "integer"
      },
      "month": {
        "type": "integer"
      }
    }
  }
}
```
```
POST stream_test_null/_bulk?refresh=true
{"index":{"_id":"1"}}
{"name":"Jake","age":70,"state":"California","country":"USA","year":2023,"month":4}
{"index":{"_id":"2"}}
{"name":"Hello","age":30,"state":"New York","country":"USA","year":2023,"month":4}
{"index":{"_id":"3"}}
{"name":"John","age":25,"state":"Ontario","country":"Canada","year":2023,"month":4}
{"index":{"_id":"4"}}
{"name":"Jane","age":20,"state":"Quebec","country":"Canada","year":2023,"month":4}
{"index":{"_id":"5"}}
{"name":null,"age":10,"state":null,"country":"Canada","year":2023,"month":4}
{"index":{"_id":"6"}}
{"name":"Kevin","year":2023,"month":4}
```

2. Run this query

```
POST _plugins/_ppl
{
  "query": """
  source=stream_test_null | streamstats avg(age) as avg_age by state
  """
}
```

3. See error

```
{
  "schema": [
    {
      "name": "name",
      "type": "string"
    },
    {
      "name": "country",
      "type": "string"
    },
    {
      "name": "state",
      "type": "string"
    },
    {
      "name": "month",
      "type": "int"
    },
    {
      "name": "year",
      "type": "int"
    },
    {
      "name": "age",
      "type": "int"
    },
    {
      "name": "avg_age",
      "type": "double"
    }
  ],
  "datarows": [
    [
      "Jake",
      "USA",
      "California",
      4,
      2023,
      70,
      70
    ],
    [
      "Hello",
      "USA",
      "New York",
      4,
      2023,
      30,
      30
    ],
    [
      "John",
      "Canada",
      "Ontario",
      4,
      2023,
      25,
      25
    ],
    [
      "Jane",
      "Canada",
      "Quebec",
      4,
      2023,
      20,
      20
    ],
    [
      null,
      "Canada",
      null,
      4,
      2023,
      10,
      10
    ],
    [
      "Kevin",
      null,
      null,
      4,
      2023,
      null,
      10
    ]
  ],
  "total": 6,
  "size": 6
}
```
The two `null` value for state are incorrectly considered in one valid group.

```
"datarows": [
    [
      "Jake",
      "USA",
      "California",
      4,
      2023,
      70,
      70
    ],
    [
      "Hello",
      "USA",
      "New York",
      4,
      2023,
      30,
      30
    ],
    [
      "John",
      "Canada",
      "Ontario",
      4,
      2023,
      25,
      25
    ],
    [
      "Jane",
      "Canada",
      "Quebec",
      4,
      2023,
      20,
      20
    ],
    [
      null,
      "Canada",
      null,
      4,
      2023,
      10,
      null
    ],
    [
      "Kevin",
      null,
      null,
      4,
      2023,
      null,
      null
    ]
  ]
```

 - Plugins

ishaoxy:
### Description

1. Added a `groupNotNull` predicate  

2. Wrapped each window expression in a conditional form:

`CASE WHEN groupNotNull THEN raw_expr ELSE CAST(NULL AS expr_type) END`

### Related Issues
* Resolve #4751 

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
@@ -413,39 +413,40 @@ public class CalciteStreamstatsCommandIT extends PPLIntegTestCase {
             + " \"Quebec\",\"country\": \"USA\",\"year\": 2023,\"month\":"
             + " 4}\n");
     client().performRequest(insertRequest);
+    try {
+      JSONObject actual =
+          executeQuery(
+              String.format(
+                  "source=%s | streamstats window=2 global=false avg(age) as avg by country",
+                  TEST_INDEX_STATE_COUNTRY));
 
-    JSONObject actual =
-        executeQuery(
-            String.format(
-                "source=%s | streamstats window=2 global=false avg(age) as avg by country",
-                TEST_INDEX_STATE_COUNTRY));
+      verifyDataRows(
+          actual,
+          rows("Jake", "USA", "California", 4, 2023, 70, 70),
+          rows("Hello", "USA", "New York", 4, 2023, 30, 50),
+          rows("John", "Canada", "Ontario", 4, 2023, 25, 25),
+          rows("Jane", "Canada", "Quebec", 4, 2023, 20, 22.5),
+          rows("Jay", "USA", "Quebec", 4, 2023, 40, 35));
 
-    verifyDataRows(
-        actual,
-        rows("Jake", "USA", "California", 4, 2023, 70, 70),
-        rows("Hello", "USA", "New York", 4, 2023, 30, 50),
-        rows("John", "Canada", "Ontario", 4, 2023, 25, 25),
-        rows("Jane", "Canada", "Quebec", 4, 2023, 20, 22.5),
-        rows("Jay", "USA", "Quebec", 4, 2023, 40, 35));
+      JSONObject actual2 =
+          executeQuery(
+              String.format(
+                  "source=%s | streamstats window=2 global=true avg(age) as avg by country",
+                  TEST_INDEX_STATE_COUNTRY));
 
-    JSONObject actual2 =
-        executeQuery(
-            String.format(
-                "source=%s | streamstats window=2 global=true avg(age) as avg by country",
-                TEST_INDEX_STATE_COUNTRY));
-
-    verifyDataRows(
-        actual2,
-        rows("Jake", "USA", "California", 4, 2023, 70, 70),
-        rows("Hello", "USA", "New York", 4, 2023, 30, 50),
-        rows("John", "Canada", "Ontario", 4, 2023, 25, 25),
-        rows("Jane", "Canada", "Quebec", 4, 2023, 20, 22.5),
-        rows("Jay", "USA", "Quebec", 4, 2023, 40, 40));
-
-    Request deleteRequest =
-        new Request(
-            "DELETE", String.format("/%s/_doc/%d?refresh=true", TEST_INDEX_STATE_COUNTRY, docId));
-    client().performRequest(deleteRequest);
+      verifyDataRows(
+          actual2,
+          rows("Jake", "USA", "California", 4, 2023, 70, 70),
+          rows("Hello", "USA", "New York", 4, 2023, 30, 50),
+          rows("John", "Canada", "Ontario", 4, 2023, 25, 25),
+          rows("Jane", "Canada", "Quebec", 4, 2023, 20, 22.5),
+          rows("Jay", "USA", "Quebec", 4, 2023, 40, 40));
+    } finally {
+      Request deleteRequest =
+          new Request(
+              "DELETE", String.format("/%s/_doc/%d?refresh=true", TEST_INDEX_STATE_COUNTRY, docId));
+      client().performRequest(deleteRequest);
+    }
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
RyanL1997: nit: unnecessary change.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1692,21 +1652,61 @@ public RelNode visitStreamWindow(StreamWindow node, CalcitePlanContext context)
...
+  private List<RexNode> wrapWindowFunctionsWithGroupNotNull(
+      List<RexNode> overExpressions, RexNode groupNotNull, CalcitePlanContext context) {
+    List<RexNode> wrappedOverExprs = new ArrayList<>(overExpressions.size());
+    for (RexNode overExpr : overExpressions) {
+      RexNode rawExpr = overExpr;
+      String aliasName = null;
+      if (overExpr instanceof RexCall rc && rc.getOperator() == SqlStdOperatorTable.AS) {
+        rawExpr = rc.getOperands().get(0);
+        if (rc.getOperands().size() >= 2 && rc.getOperands().get(1) instanceof RexLiteral lit) {
+          aliasName = lit.getValueAs(String.class);
```
RyanL1997, line 1692: Potential NPE: Add null check before calling `lit.getValueAs(String.class)`. The literal value could be null.

```java
  if (rc.getOperands().size() >= 2 && rc.getOperands().get(1) instanceof RexLiteral lit
      && lit.getValue() != null) {
      aliasName = lit.getValueAs(String.class);
  }
```

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -745,6 +742,7 @@ public void testExplainOnFirstLast() throws IOException {
   }
 
   // Only for Calcite
+  @Test
```
RyanL1997, line 745: nice :)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1692,21 +1652,61 @@ public RelNode visitStreamWindow(StreamWindow node, CalcitePlanContext context)
...
+      // wrap each expr: CASE WHEN groupNotNull THEN rawExpr ELSE CAST(NULL AS rawType) END
+      List<RexNode> wrappedOverExprs =
+          wrapWindowFunctionsWithGroupNotNull(overExpressions, groupNotNull, context);
+      context.relBuilder.projectPlus(wrappedOverExprs);
+      // resort when there is by condition
       context.relBuilder.sort(context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_STREAMSTATS));
       context.relBuilder.projectExcept(context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_STREAMSTATS));
+    } else {
+      List<RexNode> overExpressions =
+          node.getWindowFunctionList().stream().map(w -> rexVisitor.analyze(w, context)).toList();
```
RyanL1997: The same `overExpressions` calculation appears twice. Consider extracting this to a variable before the if-else block:

```java
List<RexNode> overExpressions = node.getWindowFunctionList().stream()
    .map(w -> rexVisitor.analyze(w, context)).toList();
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1692,21 +1652,61 @@ public RelNode visitStreamWindow(StreamWindow node, CalcitePlanContext context)
...
+      List<RexNode> groupByList =
+          groupList.stream().map(expr -> rexVisitor.analyze(expr, context)).toList();
+      List<RexNode> notNullList =
+          PlanUtils.getSelectColumns(groupByList).stream()
+              .map(context.relBuilder::field)
+              .map(context.relBuilder::isNotNull)
+              .toList();
+      RexNode groupNotNull =
+          notNullList.isEmpty()
+              ? context.relBuilder.literal(true)
```
RyanL1997: Just for my understanding: when `notNullList.isEmpty()`, returning `literal(true)` means all rows pass the null check. Is this the intended behavior when there are no grouping fields?

RyanL1997 (commented):
Hi @ishaoxy , thanks for taking this on. I just left some comments. In addition, for the testing, I saw we have covered most of the 'happy' cases, consider adding more negative cases (e.g. “What happens when `GROUP BY col1, col2` and only one column is null?”).

RyanL1997:
nit: I noticed that there were multiple empty space line deletion in the diff - consider revert these changes.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: `null instanceOf Class` always return false. So no need to check again.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
ishaoxy: :)

ishaoxy:
> Hi @ishaoxy , thanks for taking this on. I just left some comments. In addition, for the testing, I saw we have covered most of the 'happy' cases, consider adding more negative cases (e.g. “What happens when `GROUP BY col1, col2` and only one column is null?”).

Thank you for the suggestion! I have added such negative test.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ishaoxy: fixed.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1641,7 +1641,7 @@ public RelNode visitStreamWindow(StreamWindow node, CalcitePlanContext context)
           new String[] {ROW_NUMBER_COLUMN_FOR_STREAMSTATS});
     }
 
-    // Default
+    // default
     if (hasGroup) {
```
yuancu, line 1648: I'm a little curious why is this patch only applied to the default case. How about case 2: `node.isGlobal() && hasWindow && hasGroup`, why won't the current implementation regard two rows with `null` grouping fields as the same group? The results seem correct, but I haven't figured out the trick.

yuancu (dismissed):

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ishaoxy: This code now has been modified; sorry for this redundant logic. Because in this branch always hasGroup==true, notNullList will never be empty, and it will get result like `[state IS NOT NULL, country IS NOT NULL]` based on the group field.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ishaoxy: Because `global` path grouping uses `return context.relBuilder.equals(rightGroup, outerGroup);` to achieve grouping, `null == null` will not return true, so `null` value will not be treated as a group. However, the `default` path uses SQL's native partition semantics. In the future, we may add a button to support both these behaviors.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
```diff
@@ -644,10 +645,43 @@ public void testMultipleStreamstatsWithNull() throws IOException {
...
+        rows("Jake", "USA", "California", 4, 2023, 70, 70, 70),
+        rows("Hello", "USA", "New York", 4, 2023, 30, 30, 50),
+        rows("John", "Canada", "Ontario", 4, 2023, 25, 25, 25),
+        rows("Jane", "Canada", "Quebec", 4, 2023, 20, 20, 22.5),
+        rows("Jay", "USA", null, 4, 2023, 28, null, 50));
+
+    Request deleteRequest =
+        new Request(
+            "DELETE", String.format("/%s/_doc/%d?refresh=true", TEST_INDEX_STATE_COUNTRY, docId));
+    client().performRequest(deleteRequest);
```
LantaoJin: please change the code (and all others) to
```
...
client().performRequest(insertRequest);
try {
...
} finally {
    Request deleteRequest =
        new Request(
            "DELETE", String.format("/%s/_doc/%d?refresh=true", TEST_INDEX_STATE_COUNTRY, docId));
    client().performRequest(deleteRequest);
}
```

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
```diff
@@ -644,10 +645,43 @@ public void testMultipleStreamstatsWithNull() throws IOException {
         rows("Hello", "USA", "New York", 4, 2023, 30, 30, 50),
         rows("John", "Canada", "Ontario", 4, 2023, 25, 25, 25),
         rows("Jane", "Canada", "Quebec", 4, 2023, 20, 20, 22.5),
-        rows(null, "Canada", null, 4, 2023, 10, 10, 18.333333333333332),
+        rows(null, "Canada", null, 4, 2023, 10, null, 22.5),
         rows("Kevin", null, null, 4, 2023, null, null, null));
   }
 
+  @Test
+  public void testMultipleStreamstatsWithNull2() throws IOException {
```
LantaoJin, line 657: can you add a test for complex expression:
```
source=%s
| eval new_state=lower(state), new_country=lower(country) 
| streamstats avg(age) as avg_age by new_state, new_country
```

ishaoxy: added.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteStreamstatsCommandIT.java
ishaoxy: fixed.

LantaoJin (approved):

RyanL1997 (approved):
@ishaoxy LGTM thanks for the change.

PR #4784, opened 2025-11-12
Specify timestamp field with `timefield` in timechart command

### LINKED ISSUE #4576: [FEATURE] timechart with option to specify timestamp column, overriding the implicit @timestamp field
Using PPL version 3.4.0.

The [timechart](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/timechart.rst) command implicitly requires a timestamp field which must be name exactly `@timestamp`. When the index has a timestamp column with a different name, the command fails by a runtime error. In the example below, the `ocsf-1.1.0-4003` index has a timestamp column named `time`

```
source=ocsf-1.1.0-4003
| where TIMESTAMPDIFF(HOUR, time, NOW()) <= 24
| where rcode='NXDomain'
| timechart span=1h COUNT()
```

Failed with error

```json
{
  "error": {
    "reason": "Invalid Query",
    "details": "Field [@timestamp] not found.",
  },
  "status": 400
}
```

Provide the `timechart` command with an option `timefield=<FieldName>` to use a user-defined timestamp column.  The inital query above would then become

```
timechart timefield=time span=1h COUNT() by rcode
```

As of 2025-10-15, using PPL 3.4.0. The workaround is simply to rename the existing timestamp field to `@timestamp` to satisfy the implicit harcoded field name required by `timechart`

```
source=ocsf-1.1.0-4003
| where TIMESTAMPDIFF(HOUR, time, NOW()) <= 24
| where rcode='NXDomain'
| rename time as @timestamp
| timechart span=1h COUNT()
```

yuancu:
### Description

Allow users to specify a timestamp field instead of assuming there exists a `@timestamp` field when using `timechart` command.

For example, the following query is made possible:
```bash
source=events | timechart timefield=start_at span=1hour by category
```

### Related Issues
Resolves #4576 

### Check List


ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -760,41 +758,28 @@ private List<UnresolvedExpression> parseAggTerms(
...
-      }
-    }
+    List<Argument> arguments = ArgumentFactory.getArgumentList(ctx, expressionBuilder);
+    ArgumentMap argMap = ArgumentMap.of(arguments);
+    Literal spanLiteral = argMap.getOrDefault("spanliteral", AstDSL.stringLiteral("1m"));
+    String timeFieldName =
+        Optional.ofNullable(argMap.get("timefield"))
+            .map(l -> (String) l.getValue())
+            .orElse(OpenSearchConstants.IMPLICIT_FIELD_TIMESTAMP);
+    Field spanField = AstDSL.field(timeFieldName);
```
dai-chen, line 778: Could you check if we need to use this specified time field in `transformPerFunction` for per functions as well?

yuancu: Yes, I think we need to use the specified `timefield` in per functions because otherwise they'll refer to the non-existing `@timestamp` field. I added an integration test for this case.

LantaoJin:
non-blocking: @yuancu looks like this PR would impact the reverse enhancement https://github.com/opensearch-project/sql/pull/4775 more or less.

yuancu:
> non-blocking: @yuancu looks like this PR would impact the reverse enhancement #4775 more or less.

Yes it will. If this PR gets merged first, I'll help @ahkcs replace time field reference from hard-coded ones to the timefield variable.

qianheng-aws (approved):

PR #4793, opened 2025-11-12
Fix function identify problem in converting to sql dialect

### LINKED ISSUE #4774: [BUG] Scalar and aggregate `MIN`/`MAX` conflict when translating PPL to SQL
**What is the bug?**

PPL introduced scalar version of `MIN`/`MAX` function in https://github.com/opensearch-project/sql/pull/4333. In PPL this is unambiguous because scalar vs aggregate version are distinguished by the surrounding command (`eval` vs. `stats`).

However, when we translate PPL to SQL (either as SQL text or as a `SqlNode`), the same function names (`MIN`/`MAX`) are reused for both scalar and aggregate semantics. Depending on the engine’s function registry and resolution rules, this name collision can lead to ambiguous resolution or the wrong implementation being chosen.

```
-- Standard aggregate usage
SELECT MIN(age) FROM accounts;

-- Scalar-style usage that comes from PPL translation
SELECT MIN(age, 'test', NULL) FROM accounts;
```

One possible solution is to keep exposing scalar `MIN`/`MAX` to PPL users under the same names, but have the translator map scalar calls to internal, non-conflicting function names.

 - OS: 3.4

N/A

This issue was discovered during the PPL function unification PoC in
https://github.com/opensearch-project/opensearch-spark/issues/1281#issuecomment-3500524870. After registering the PPL scalar min/max functions in the shared function registry, SparkSQL’s native aggregate MIN implementation is effectively overridden. For example, in spark-sql:

```
spark-sql (default)> SELECT MIN(packets) FROM test_events;
min(packets)
60
120
60
180
```

Instead of returning a single aggregated value, multiple rows are returned, which indicates that the scalar variant (or a conflicting resolution) is being used.

LantaoJin:
### Description
Fix function identify problem in converting to sql dialect:
- change min (scalar function) to scalar_min in plan level
- change max (scalar function) to scalar_max in plan level
- remove the UDF backtick in converting to sql dialect 

### Related Issues
Resolves #4774 

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
@@ -1298,19 +1298,19 @@ public class CalciteExplainIT extends ExplainIT {
 
   @Test
   public void testExplainEvalMax() throws IOException {
-    String expected = loadExpectedPlan("explain_eval_max.json");
-    assertJsonEqualsIgnoreId(
+    String expected = loadExpectedPlan("explain_eval_max.yaml");
+    assertYamlEqualsIgnoreId(
         expected,
-        explainQueryToString(
+        explainQueryYaml(
             "source=opensearch-sql_test_index_account | eval new = max(1, 2, 3, age, 'banana')"));
   }
 
   @Test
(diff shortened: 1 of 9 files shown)


LantaoJin:
cc @dai-chen

Issue #4757, opened 2025-11-07
[RFC] RexNode standardization for script push down


qianheng-aws:
**Problem Statement**
The Calcite engine in PPL supports pushing down various operators (e.g., FILTER, AGGREGATION) by utilizing scripts when UDFs cannot be directly translated to DSL features.

However, OpenSearch cluster imposes certain restrictions on script usage when submitting search requests:
- Size limitation: 65,535 bytes (default)
- Rate limitation: 75 requests per 5 minutes (default)

The rate restriction poses a significant challenge, particularly as more and more OpenSearch components frequently use PPL or when usage of PPL from users spikes during specific periods.

While OpenSearch provides a script-caching mechanism using the LRU algorithm with two key configurations:
- script.cache.max_size: Maximum number of cached scripts (default: 100)
- script.cache.expire: Script expiration time (default: 0, indicating no time-based expiration)

Despite this caching mechanism, PPL is not effectively utilizing the cache as expected, leading to potential performance issues and rate limit violations.

**Current State**
Calcite engine: 
In the script push down process, calcite engine will serialize 3 parts of content into the final script:
- RexNode expression in the pushed down operator
- RowType of the current scan operator
- FieldTypes of the index mapping 

The `RexNode` contains the specific information from the PPL query, like which fields, literal values or functions are used. `RowType` and `FieldTypes` both derived from the index mapping. They are all able to prevent script cache sharing among different queries.

V2 engine: 
v2 engine has similar issues as calcite engine, and what's more it has defect of encoding current timestamp into the final scripts, which means scripts are always changing and the script cache can never be hit even for the same query.

**Proposal**
We should perform expression for script push down, which aims to transform the current RexNode expression into a common pattern applicable to different RexNode expression. This way, its script cache can also apply to different expressions even across PPL queries.

In the standardization process, we also expect to reduce the size of a script since it will make the expression simpler by somehow.

**Approach**
1. Literal standardization
In this step, we replace the `RexLiteral` with `RexDynamicParameter` so it can apply to different values of literal in its position, or even fields. e.g.
```
// Before
... | where LENGTH(a) > 10 | eval b = a + "SUFFIX" | stats count() by b

// With standardization
... | where LENGTH(a) > `?0` | eval b  = a + `?1` | stats count() by b
```
In the standardized expression, `FIELD1` and `FIELD2` can be any literal value or even dynamic values from a field of the index.

[...]

PR #4795, opened 2025-11-13
Perform RexNode expression standardization for script push down.

qianheng-aws:
### Description
This PR includes changes:
1. Implement the step 1&2(Replace field and literal with parameters) described in the RFC:https://github.com/opensearch-project/sql/issues/4757. This will enhance our script cache to get higher hitting ratio.
2. Remove `ROW_TYPE` and `EXPR_MAP` in our script. Then the average script size can be reduced by ***2 to 5*** times than before.
3. Remove `OpenSearchRequestBuilder` when computing digest for `OpenSearchIndexScanOperator`, while keep it when generating explain plan.
4. Remove `OpenSearchRequestBuilder` in `PushDownContext` and make the related action lazy perform. Since we have change 3, it's less valuable to hold that object in each `PushDownContext`.
5. Tiny enhancement on the parameter of SORT_EXPR, see https://github.com/opensearch-project/sql/pull/4750#discussion_r2532987617

### Related Issues
Partly resolves https://github.com/opensearch-project/sql/issues/4757

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
```diff
@@ -62,6 +72,78 @@ public static Pair<RexNode, RelDataType> getRemappedRexAndType(
     return Pair.of(newMappedRex, builder.build());
   }
 
+  public static RexNode standardizeRexNodeExpression(
+      final RexNode rexNode,
+      final RelDataType inputRowType,
+      final Map<String, ExprType> fieldTypes,
+      List<Integer> sources,
+      List<Object> digests,
+      List<Object> literals) {
```
yuancu: Javadoc is appreciated

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggregationBuilderAction.java
```diff
@@ -7,13 +7,9 @@
...
-  /**
-   * Apply the action on the target {@link AggPushDownAction} and add the operation to the context
-   *
-   * @param context the context to add the operation to
-   * @param operation the operation to add to the context
-   */
-  default void transform(PushDownContext context, PushDownOperation operation) {
+  default void pushOperation(PushDownContext context, PushDownOperation operation) {
+    // Apply transformation to aggregation builder in the optimization phase as some transformation
+    //  may cause exception. We need to detect that exception in advance.
```
penghuo, line 12: what is meaning of this error message? TODO?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/PushDownContext.java
```diff
@@ -94,12 +96,19 @@ ArrayDeque<PushDownOperation> getOperationsForAgg() {
...
       isAggregatePushed = true;
       this.aggPushDownAction = (AggPushDownAction) operation.action();
     }
     if (operation.type() == PushDownType.LIMIT) {
+      startFrom += ((LimitDigest) operation.digest()).offset();
+      if (startFrom >= osIndex.getMaxResultWindow()) {
+        throw new PushDownUnSupportedException(
+            String.format(
+                "Requested offset %d should be less than the max result window %d",
+                startFrom, osIndex.getMaxResultWindow()));
```
penghuo, line 111: this is not customer facing error, right?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -166,76 +155,31 @@ public UnsupportedScriptException(Throwable cause) {
...
         SourceLookup sourceLookup,
         Map<String, Object> params) {
       this.docProvider = docProvider;
       this.sourceLookup = sourceLookup;
-      this.params = params;
+      this.utcTimestamp = (long) params.get(Variable.UTC_TIMESTAMP.camelName);
+      this.sources =
+          ((List<Integer>) params.get("SOURCES")).stream().map(Source::fromValue).toList();
+      this.digests = (List<Object>) params.get("DIGESTS");
+      this.literals = (List<Object>) params.get("LITERALS");
```
penghuo: Is there any test cases show decoded pushdown script?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -256,9 +200,21 @@ public QueryProvider getQueryProvider() {
...
-        return params.get(Variable.UTC_TIMESTAMP.camelName);
+      if (Variable.UTC_TIMESTAMP.camelName.equals(name)) return this.utcTimestamp;
+
+      try {
+        int index = paramNameToIndex.get(name);
+        return switch (sources.get(index)) {
+          case DOC_VALUE -> getFromDocValue((String) digests.get(index));
+          case SOURCE -> getFromSource((String) digests.get(index));
+          case LITERAL -> getFromLiteral((Integer) digests.get(index));
+        };
```
penghuo, line 214: could u add a developer doc to explain the spec of pushdown script, after encoding, is not easy to read I guess.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -125,17 +113,18 @@ public String getType() {
...
-        (Map<String, ExprType>) objectMap.get(RelJsonSerializer.FIELD_TYPES);
-
-    JavaTypeFactory typeFactory = OpenSearchTypeFactory.TYPE_FACTORY;
-    RexToLixTranslator.InputGetter getter = new ScriptInputGetter(typeFactory, rowType, fieldTypes);
+    RexNode rexNode = relJsonSerializer.deserialize(scriptCode);
+
+    RexToLixTranslator.InputGetter getter =
+        (blockBuilder, i, type) -> {
+          throw new UnsupportedScriptException("There shouldn't be RexInputRef in the RexNode.");
+        };
```
penghuo, line 129: By design, it is never throw exception, right?
if happened, it will customer facing exception?
if yes, Add [BUG]  in error message

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
```diff
@@ -62,6 +72,78 @@ public static Pair<RexNode, RelDataType> getRemappedRexAndType(
     return Pair.of(newMappedRex, builder.build());
   }
 
+  public static RexNode standardizeRexNodeExpression(
```
penghuo: Add UT for this method.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/PushDownContext.java
qianheng-aws: Yeah, it will caught by `pushDownLimit` and then it will prevent the push down the limit operator then.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: See `RelJsonSerializerTest`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggregationBuilderAction.java
qianheng-aws: > Remove OpenSearchRequestBuilder in PushDownContext and make the related action lazy perform. Since we have change 3, it's less valuable to hold that object in each PushDownContext.

I made the transformation of each push down operation lazy except `operationsForAgg`. This comment explains why we have this exception of not making it lazy.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: Yeah, it should be a bug if this exception happens.

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
qianheng-aws: We have test `RelJsonSerializerTest` which will cover this actually. That tests verify the whole process `standardizeRexNodeExpression -> encoding -> decoding` while the `encoding -> decoding` could be ignored since we just invoked a Base64 encoder/decoder.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/PushDownContext.java
qianheng-aws: Will add [INTERNAL] tag in the error message

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: Added in https://github.com/opensearch-project/sql/pull/4795/commits/12780545aecec0b076e7172d49523f42c82e91c0

opensearch/src/main/java/org/opensearch/sql/opensearch/util/OpenSearchRelOptUtil.java
```diff
@@ -62,6 +72,79 @@ public static Pair<RexNode, RelDataType> getRemappedRexAndType(
...
+      final RelDataType inputRowType,
+      final Map<String, ExprType> fieldTypes,
+      List<Integer> sources,
+      List<Object> digests,
+      List<Object> literals) {
+    final List<RelDataTypeField> inputFieldList = inputRowType.getFieldList();
+
+    final int[] currentIndex = {0};
+    final RexShuttle rexShuttle =
+        new RexShuttle() {
```
songkant-aws: Could we make it a singleton instance? Maybe leverage `RexBiVisitorImpl`?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
penghuo: Thanks. It is very clear. 

Nit: Is Literals array necessary? DIGESTS and LITERALS can combined?
```
        "params": {
            "utcTimestamp": 17630261838681530000,
            "SOURCES": [0, 2, 2, 1],
            "DIGESTS": ["age", 0, 1, "email"],
            "LITERALS": [35, "u35"]
        }
```
vs
```
        "params": {
            "utcTimestamp": 17630261838681530000,
            "SOURCES": [0, 2, 2, 1],
            "DIGESTS": ["age", 35, "u35", "email"]
        }
```

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1463,10 +1471,23 @@ public ScriptQueryExpression(
...
       RelJsonSerializer serializer = new RelJsonSerializer(cluster);
+      this.sources = new ArrayList<>();
+      this.digests = new ArrayList<>();
+      this.literals = new ArrayList<>();
       this.codeGenerator =
           () ->
               SerializationWrapper.wrapWithLangType(
-                  ScriptEngineType.CALCITE, serializer.serialize(rexNode, rowType, fieldTypes));
+                  ScriptEngineType.CALCITE,
+                  serializer.serialize(rexNode, rowType, fieldTypes, sources, digests, literals));
```
penghuo: why include `sources, digests, literals` as paramater in serialize() function and client create empty array?

penghuo: I see. It is required when create Script on L1504.
I also found sources, digests, literals exposed been used in multiple place without encapsulation. e.g. ScriptDataContext and standardizeRexNodeExpression. 

can we encapsulate our [script protocol](https://github.com/opensearch-project/sql/blob/12780545aecec0b076e7172d49523f42c82e91c0/docs/dev/intro-scripts.md) in a class?  e.g. 
```
class ParameterBindings {
   void putValue(String name, Object value)
   Object getValue(String name)
}
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
qianheng-aws: Agree, it could work. I split it for 2 reasons:
1. keep alignment concept, in which we view `LITERALS` as a kind of source and retrieve actual value via `DIGEST`, just like looking up `DOC_VALUE` or `SOURCE`.
 
2. in case there is duplicated literal value and `LITERALS` can reduce them, e.g. 
```
        "params": {
            "utcTimestamp": 17630261838681530000,
            "SOURCES": [0, 2, 2, 2],
            "DIGESTS": ["age", 0, 0, 0],
            "LITERALS": ["<LITERAL_VALUE_WIHT_BIG_SIZE>"]
        }
```
Though in most cases, it shouldn't have much benefit. I only found 1 case  in our IT hit this, like `case(age < 30, 30 else 100)` will reuse literal 30.  So I don't have strongly incline on the current way.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -31,6 +31,7 @@
 import static com.google.common.base.Preconditions.checkState;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
+import static javax.swing.UIManager.put;
```
songkant-aws: nit question: Is it a right import?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1463,10 +1471,23 @@ public ScriptQueryExpression(
...
       this.codeGenerator =
           () ->
               SerializationWrapper.wrapWithLangType(
-                  ScriptEngineType.CALCITE, serializer.serialize(rexNode, rowType, fieldTypes));
+                  ScriptEngineType.CALCITE,
+                  serializer.serialize(rexNode, rowType, fieldTypes, sources, digests, literals));
+    }
+
+    // For filter script, this method will be called after planning phase;
+    // For the agg-script, this will be called in planning phase to generate agg builder
```
songkant-aws: Not relevant to this PR. It would be nice to be lazy as well. Does the exception come from script generation or else?

qianheng-aws: Agree. Our push down operations after aggregation have strongly dependance on the exception throwing mechanism, so it can prevent such push down easily if exception happens. 

We'd better extract all cases which is actually not our targets out of the transformation lambda function. But they are too many and too complex to refactor at once in this PR.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
qianheng-aws: Thanks! Will remove

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/ExtendedRelJson.java
```diff
@@ -364,7 +365,11 @@ public RelDataType toType(RelDataTypeFactory typeFactory, Object o) {
       // Reconstruct UDT from its udt tag
       Object udtName = ((Map<?, ?>) o).get("udt");
       OpenSearchTypeFactory.ExprUDT udt = OpenSearchTypeFactory.ExprUDT.valueOf((String) udtName);
-      return ((OpenSearchTypeFactory) typeFactory).createUDT(udt);
+      // View IP as string to avoid using a value of customized java type in the script.
+      if (udt == ExprUDT.EXPR_IP) return super.toType(typeFactory, o);
```
songkant-aws, line 369: One more thing that comes up in my mind, how about the sorting by ExprIpValue? It was one of blocker to prevent last change made by yuanchuan to directly convert IP to string. cc @yuancu 

Another discussion I remember is whether IP sorting is necessary. Forgot the result of discussion.

qianheng-aws: Luckily seems we don't have any function/script return ExprIPValue for now, so I think at least it won't block any script push down.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
penghuo: looks good. 
Can ScriptParameterHelper been used to hide protocol?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -166,76 +158,28 @@ public UnsupportedScriptException(Throwable cause) {
...
+        Map<String, Object> params,
+        Map<String, Integer> parameterToIndex) {
       this.docProvider = docProvider;
       this.sourceLookup = sourceLookup;
-      this.params = params;
+      this.utcTimestamp = (long) params.get(Variable.UTC_TIMESTAMP.camelName);
+      this.sources = ((List<Integer>) params.get(SOURCES)).stream().map(Source::fromValue).toList();
+      this.digests = (List<Object>) params.get(DIGESTS);
+      this.literals = (List<Object>) params.get(LITERALS);
+      this.parameterToIndex = parameterToIndex;
```
penghuo, line 185: Can ScriptParameterHelper been used to hide protocol?

qianheng-aws: Do you mean put `ScriptParameterHelper` object in the script parameter directly? I remember the opensearch request will be formatted to json by using `XContentSerializer` which cannot serialize self defined class. That's why I convert `ScriptParameterHelper` into a map when constructing request.

qianheng-aws: By my local test, it will throw exception if putting ScriptParameterHelper in request directly.
```
Caused by: java.lang.IllegalArgumentException: cannot write xcontent for unknown value of type class org.opensearch.sql.opensearch.storage.serde.ScriptParameterHelper
```

penghuo (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/RexStandardizer.java
```diff
@@ -0,0 +1,199 @@
...
+            ? null
+            : OpenSearchTextType.toKeywordSubField(field.getName(), exprType);
+    int newIndex = helper.currentIndex[0]++;
+    if (docFieldName != null) {
+      helper.digests.add(docFieldName);
+      helper.sources.add(Source.DOC_VALUE.getValue());
+    } else {
+      helper.digests.add(field.getName());
+      helper.sources.add(Source.SOURCE.getValue());
+    }
```
yuancu, line 80: Just curious: is there any benefit of retrieving from doc values over from source?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/ExtendedRelJson.java
yuancu: I think this looks good for now

yuancu (commented):
Conflicts exist

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/RexStandardizer.java
qianheng-aws: Retrieving from source will be slower

qianheng-aws:
The latest change: https://github.com/opensearch-project/sql/pull/4795/commits/f74600ff6c4f002b5e018b0cece2dcc162b6056d includes: 
1. an enhancement on the parameter from SORT_EXPR: https://github.com/opensearch-project/sql/pull/4750#discussion_r2532987617
2. Apply comments: https://github.com/opensearch-project/sql/pull/4795#discussion_r2528384908

LantaoJin (approved):

PR #4814, opened 2025-11-14
Support `split` eval function

ahkcs:
## **Summary**

This PR implements the `split` eval function for PPL, enabling users to split strings into multivalue arrays based on a delimiter. 

### **Basic split with semicolon**

```ppl
source=people | eval result = split('a;b;c', ';')
```

Result: `['a', 'b', 'c']`

---

### **Split into individual characters (empty delimiter)**

```ppl
source=people | eval result = split('abcd', '')
```

Result: `['a', 'b', 'c', 'd']`

---

### **Multi-character delimiter**

```ppl
source=people | eval result = split('name::value', '::')
```

Result: `['name', 'value']`

---

### **Split field value**

```ppl
source=people | eval words = split(employer, ' ')
```

Splits the `employer` field on spaces.

---

### Check List


dai-chen (commented):
Can this be reused? https://issues.apache.org/jira/browse/CALCITE-6951

ahkcs:
> https://issues.apache.org/jira/browse/CALCITE-6951

Hi @dai-chen , currently we are using `SPLIT`, which is also supported by Calcite. And we did some special handling for empty delimiter on top of that.

Here's the documentation for Calcite `SPLIT` function:

```

SPLIT(string [, delimiter ]) 

```

Returns the string array of string split at delimiter (if omitted, default is comma). If the string is empty it returns an empty array, otherwise, if the delimiter is empty, it returns an array containing the original string.

dai-chen:
> > https://issues.apache.org/jira/browse/CALCITE-6951
> 
> Hi @dai-chen , currently we are using `SPLIT`, which is also supported by Calcite. And we did some special handling for empty delimiter on top of that.
> 
> Here's the documentation for Calcite `SPLIT` function:
> 
> ```
> 
> SPLIT(string [, delimiter ]) 
> ```
> 
> Returns the string array of string split at delimiter (if omitted, default is comma). If the string is empty it returns an empty array, otherwise, if the delimiter is empty, it returns an array containing the original string.

I see. So the only reason of `SplitFunctionImp` is special handling for `delimiter=""`, right?
Can the extract function below help?

- Case 1: Delimiter is not empty string, translate split to

```
SELECT SPLIT('a;b;c;d', ';');
+--------------+
|    EXPR$0    |
+--------------+
| [a, b, c, d] |
+--------------+
```

- Case 2: Delimiter is empty string, translate split to:

```
SELECT REGEXP_EXTRACT_ALL('abcd', '.');
+--------------+
|    EXPR$0    |
+--------------+
| [a, b, c, d] |
+--------------+
```

ahkcs:
@dai-chen Thanks for the suggestion! I think it makes sense. I have updated the PR to move the implementation to `PPLFuncImpTable`

docs/user/ppl/functions/collection.rst
```diff
@@ -186,6 +186,60 @@ Example::
...
+    |--------------|
+    | [name,value] |
+    +--------------+
+
+    os> source=people | eval test = 'hello', result = split(test, ',') | fields result | head 1
+    fetched rows / total rows = 1/1
+    +---------+
+    | result  |
+    |---------|
+    | [hello] |
```
dai-chen, line 240: This covers delimiter not found case right? The result should be a original string or an array of original string?

ahkcs: Yes, if delimiter is not found, the result should be an array of original string

dai-chen (approved):
Thanks for the changes!

PR #4822, opened 2025-11-19
Support spath command to extract all fields

ykmr1224:

### Description
- Support spath command to extract all fields when output and path are not specified
  - It will parse input field as JSON and extract all the paths into fields.
  - If same field name already exists, it will append the result to existing value (the value becomes an array).
  - Non-existing fields will be added to `_MAP`(dynamic fields) as we cannot know the contents at planning time.
  - Static fields will become `ANY` type because append could happen depending on the JSON content.
- Introduced internal function `APPEND` to keep existing field scalar, and make single item result scalar
  - Originally, planning to use `MVAPPEND` which returns array even for single value result, but realized it makes other commands require conversion from array to scalar. I considered automatic type coercion from array to scalar, but it does not align with Splunk behavior where multivalue with single item would be automatically converted as scalar value.

### Related Issues

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -691,11 +691,11 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     if (node.getPath() != null) {
       return visitEval(node.rewriteAsEval(), context);
     } else {
-      return spathWithoutPath(node, context);
+      return spathExtractAll(node, context);
     }
   }
 
-  private RelNode spathWithoutPath(SPath node, CalcitePlanContext context) {
+  private RelNode spathExtractAll(SPath node, CalcitePlanContext context) {
     visitChildren(node, context);
 
     // 1. Extract all fields from JSON in `inField` and merge with existing dynamic fields.
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/standalone/MapAppendFunctionIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/standalone/MapAppendFunctionIT.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.sql.calcite.standalone;
 
+import static org.hamcrest.Matchers.*;
+
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.List;
@@ -41,11 +43,14 @@ public class MapAppendFunctionIT extends CalcitePPLRelNodeIntegTestCase {
         relNode,
         resultSet -> {
           Map<String, Object> result = getResultMapField(resultSet);
-          assertEquals(4, result.size());
-          assertMapValue(result, "key1", "value1");
-          assertMapValue(result, "key2", "value2");
-          assertMapValue(result, "key3", "value3");
-          assertMapValue(result, "key4", "value4");
+          assertThat(
+              result,
+              allOf(
+                  hasEntry("key1", "value1"),
+                  hasEntry("key2", "value2"),
+                  hasEntry("key3", "value3"),
+                  hasEntry("key4", "value4"),
+                  aMapWithSize(4)));
         });
   }
 
@@ -68,10 +73,13 @@ public class MapAppendFunctionIT extends CalcitePPLRelNodeIntegTestCase {
         relNode,
         resultSet -> {
           Map<String, Object> result = getResultMapField(resultSet);
-          assertEquals(3, result.size());
-          assertMapValue(result, "key1", "value1");
-          assertMapListValue(result, "key2", "value2", "value3");
-          assertMapValue(result, "key3", "value4");
+          assertThat(
+              result,
+              allOf(
+                  aMapWithSize(3),
+                  hasEntry("key1", (Object) "value1"),
+                  hasEntry("key2", (Object) List.of("value2", "value3")),
+                  hasEntry("key3", (Object) "value4")));
         });
   }
 
@@ -100,8 +108,7 @@ public class MapAppendFunctionIT extends CalcitePPLRelNodeIntegTestCase {
         relNode,
         resultSet -> {
           Map<String, Object> result = getResultMapField(resultSet);
-          assertEquals(1, result.size());
-          assertMapValue(result, "key1", "value1");
+          assertThat(result, allOf(hasEntry("key1", "value1"), aMapWithSize(1)));
         });
   }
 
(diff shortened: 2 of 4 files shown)


integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLSpathCommandIT.java
```diff
@@ -12,37 +12,313 @@
...
+    verifySchema(
+        result,
+        schema("id", "bigint"),
+        schema("nested{}.a", "array"),
+        schema("nested{}.arr{}", "array"));
+    verifyDataRows(result, rows(3L, arr("v1", "v2"), arr(1, 2, 3, 4, 5)));
+  }
+
+  @Test
+  @Ignore("Join does not work now")
```
ykmr1224, line 280: Pending this PR: https://github.com/opensearch-project/sql/pull/4746

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -688,7 +688,50 @@ public RelNode visitParse(Parse node, CalcitePlanContext context) {
...
   public RelNode visitSpath(SPath node, CalcitePlanContext context) {
-    return visitEval(node.rewriteAsEval(), context);
+    if (node.getPath() != null) {
+      return visitEval(node.rewriteAsEval(), context);
+    } else {
+      return spathWithoutPath(node, context);
+    }
+  }
+
+  private RelNode spathWithoutPath(SPath node, CalcitePlanContext context) {
```
Swiddis: Should this be part of the SPath class instead of in Visitor?

nit: `spathWithoutPath` might be better named something like `spathExtractAll`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -688,7 +688,50 @@ public RelNode visitParse(Parse node, CalcitePlanContext context) {
...
+    if (context.fieldBuilder.isDynamicFieldsExist()) {
+      map =
+          context.rexBuilder.makeCall(
+              BuiltinFunctionName.MAP_APPEND, context.fieldBuilder.getDynamicFieldsMap(), map);
+    }
+
+    // 2. Merge dynamic fields with static fields.
+    //   static_field1 = APPEND(static_field1, _MAP[static_field1]), static_attr2 = ...
+    List<String> staticFieldNames = context.fieldBuilder.getStaticFieldNames();
+    List<RexNode> fields = new ArrayList<>();
```
Swiddis, line 714: I mentioned in standup this morning that I think we're making this more complex by working directly on the `List` of fields.

I'd like to consider putting the bulk of this logic in a data class that can start supplying a higher-level interface for most field operations. Can incrementally move more commands to use it over time.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/MapAppendFunctionIT.java
```diff
@@ -42,10 +42,10 @@ public void testMapAppendWithNonOverlappingKeys() throws Exception {
...
           Map<String, Object> result = getResultMapField(resultSet);
           assertEquals(4, result.size());
-          assertMapListValue(result, "key1", "value1");
-          assertMapListValue(result, "key2", "value2");
-          assertMapListValue(result, "key3", "value3");
-          assertMapListValue(result, "key4", "value4");
+          assertMapValue(result, "key1", "value1");
+          assertMapValue(result, "key2", "value2");
+          assertMapValue(result, "key3", "value3");
+          assertMapValue(result, "key4", "value4");
```
Swiddis: Should we have a higher-level assertion helper here so we can do this all in one assert?

if this ever fails with multiple key mismatches, might not be obvious what the problem is by just looking at the first mismatch (or if the length is wrong, we end up with "expected 4 found 6" or something)

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/AppendCore.java
```diff
@@ -8,14 +8,32 @@
...
+ * Core logic for `mvappend` and internal `append` function to collect elements from list of args.
+ */
+public class AppendCore {
 
   /**
    * Collect non-null elements from `args`. If an item is a list, it will collect non-null elements
-   * of the list. See {@ref MVAppendFunctionImplTest} for detailed behavior.
+   * of the list. See {@ref AppendFunctionImplTest} for detailed behavior.
+   */
+  public static Object collectElements(Object... args) {
```
Swiddis, line 20: thought: Can we make this typing any more specific than `Object`?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: I think we shouldn't move Calcite specific implementation to SPath, which is rather AST layer.

Agree with the name change.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ykmr1224: This one is very special logic for `spath` to append values to static fields and remove appended ones from dynamic fields. I don't think it is something reusable for other use case, and should be coded in `spath` implementation rather than abstraction layer.
`spath` command is special since it generate new dynamic fields while others only consume.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/AppendCore.java
ykmr1224: We cannot since input/output could be any type.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
Swiddis: If spath command is special enough for this, I imagine other commands would have their own special cases later. It would still be worth finding an abstraction here in my view. To me it seems like this is one of the main operations of adding dynamic fields to that structure & would eventually be reusable as more dynamic commands are visited.

ykmr1224: Expected commands which would add dynamic fields are: `multikv` and `timechart` (There are some more with lower priorities: chart, extract, kv, xpath, xmlkv, transpose, xyseries, pivot)

`multikv` would simply overwrite existing fields.
`timechart` would also overwrite existing fields (it is a bit different since it would rather make new frame rather than updating existing frame)

There might be a chance we could reuse current logic in the future, but I think we should generalize when we implement that. I don't have very good idea to generalize some part right now. Let me know if you already have clear picture which part should be generalized.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/MapAppendFunctionIT.java
ykmr1224: Fixed

penghuo:
Close this PR. Since it target feature branch.

PR #4825, opened 2025-11-19
Push down filters on nested fields as nested queries

### LINKED ISSUE #4508: [BUG] PPL filter on computed fields from nested paths returns zero results
## Query Information
**PPL Command/Query:**
```ppl
# This works - returns 3 results with computed NameLen field
source=test-nested-eval-filter | eval NameLen=LENGTH(items.name) | fields id, items.name, NameLen

# This fails - returns 0 results (should return 2: banana and orange with length > 5)
source=test-nested-eval-filter | eval NameLen=LENGTH(items.name) | fields id, items.name, NameLen | where NameLen > 5
```

**Expected Result:**
The query should return 2 documents where the computed `NameLen` field (length of `items.name`) is greater than 5:
- order2 with banana (length 6)
- order3 with orange (length 6)

**Actual Result:**
The query returns 0 results:
```json
{
  "schema": [...],
  "datarows": [],
  "total": 0,
  "size": 0
}
```

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "id": {"type": "keyword"},
      "items": {
        "type": "nested",
        "properties": {
          "name": {"type": "keyword"}
        }
      }
    }
  }
}
```

**Sample Data**
```json
{"id":"order1","items":[{"name":"apple"}]}
{"id":"order2","items":[{"name":"banana"}]}
{"id":"order3","items":[{"name":"orange"}]}
```

## Bug Description
**Issue Summary:**
Filtering on computed fields (created with `eval`) that reference nested field paths returns zero results, even though the computed field is correctly calculated when not filtered. The issue is specific to nested fields - the same pattern works correctly with regular (non-nested) fields.

**Steps to Reproduce:**
1. Create an index with a nested field mapping
2. Insert documents with nested objects
3. Run `eval` to create a computed field based on a nested field path (e.g., `LENGTH(items.name)`)
4. Add a `where` clause to filter on the computed field
5. Observe that the query returns 0 results instead of the expected filtered results

**Minimal Reproduction:**
```bash
# Create test index
curl -X PUT "localhost:9200/test-nested-eval-filter" -H 'Content-Type: application/json' -d '{
  "mappings": {
    "properties": {
      "id": {"type": "keyword"},
      "items": {
        "type": "nested",
        "properties": {
          "name": {"type": "keyword"}
        }
      }
    }
  }
}'

# Insert test data
curl -X POST "localhost:9200/test-nested-eval-filter/_bulk" -H 'Content-Type: application/json' -d '
{"index":{"_id":"1"}}
{"id":"order1","items":[{"name":"apple"}]}
{"index":{"_id":"2"}}
{"id":"order2","items":[{"name":"banana"}]}
{"index":{"_id":"3"}}
{"id":"order3","items":[{"name":"orange"}]}
'

curl -X POST "localhost:9200/test-nested-eval-filter/_refresh"

# This works - returns 3 results
curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d '{
  "query": "source=test-nested-eval-filter | eval NameLen=LENGTH(items.name) | fields id, items.name, NameLen"
}'

# This fails - returns 0 results (should return 2)
curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d '{
  "query": "source=test-nested-eval-filter | eval NameLen=LENGTH(items.name) | fields id, items.name, NameLen | where NameLen > 5"
}'
```

**Verification with Regular Fields:**
The same pattern works correctly with regular (non-nested) fields:
```bash
# Create index with regular field
curl -X PUT "localhost:9200/test-regular-eval-filter" -H 'Content-Type: application/json' -d '{
  "mappings": {
    "properties": {
      "id": {"type": "keyword"},
      "name": {"type": "keyword"}
    }
  }
}'

# Insert data
curl -X POST "localhost:9200/test-regular-eval-filter/_bulk" -H 'Content-Type: application/json' -d '
{"index":{"_id":"1"}}
{"id":"order1","name":"apple"}
{"index":{"_id":"2"}}
{"id":"order2","name":"banana"}
{"index":{"_id":"3"}}
{"id":"order3","name":"orange"}
'

curl -X POST "localhost:9200/test-regular-eval-filter/_refresh"

# This works correctly - returns 2 results
curl -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d '{
  "query": "source=test-regular-eval-filter | eval NameLen=LENGTH(name) | fields id, name, NameLen | where NameLen > 5"
}'
```

**Impact:**
This bug prevents users from filtering on computed fields that are based on nested field values, which is a common use case in observability data (OTEL, Jaeger) and other structured data with nested objects.

## Environment Information
**OpenSearch Version:**
OpenSearch 3.3.0-SNAPSHOT

**Additional Details:**
- Affects PPL queries using Calcite engine (V3)
- The issue is specific to nested fields; regular fields work correctly
- The `eval` command successfully computes values from nested fields when no filter is applied

---

### Core Issue
When a `where` clause filters on a computed field that references a nested field, the filter is pushed down to OpenSearch as a script query. The script attempts to access the nested field value using doc values (via `docProvider.get()`), but nested fields in OpenSearch do not have doc values. This causes the script to return null for all documents, making the filter evaluate to false for every document.

### Code References

1. **Script Execution**: `/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java` (lines 226-275)
   - `ScriptDataContext.get(String name)` method (line 263)
   - Attempts to retrieve field values from `docProvider.get(name)`
   - Returns null when doc values are empty or missing (line 267)

```java
@Override
public Object get(String name) {
  // ...
  ScriptDocValues<?> docValue = this.docProvider.get(name);
  if (docValue == null || docValue.isEmpty()) {
    return null; // No way to differentiate null and missing from doc value
  }
  // ...
}
```

2. **Alternative Access Method**: The same class has a `getFromSource(String name)` method that could access nested fields from `_source`, but it's not used by the generated Calcite code.

### Why This Happens

1. **Doc Values Limitation**: Nested fields in OpenSearch are stored differently and don't have doc values accessible via the standard `doc[]` API
2. **Script Generation**: The Calcite-generated code always uses the `DataContext.get()` method, which relies on doc values
3. **No Nested Field Detection**: The code doesn't detect when a field path references a nested field and switch to `_source` access
4. **Silent Failure**: When doc values are unavailable, the method returns null rather than falling back to `_source` access

### Direct OpenSearch Query Verification

Testing the underlying OpenSearch script query directly confirms the issue:
```bash
curl -X POST "localhost:9200/test-nested-eval-filter/_search" -H 'Content-Type: application/json' -d '{
  "query": {
    "script": {
      "script": {
        "source": "doc[\"items.name\"].value.length() > 5",
        "lang": "painless"
      }
    }
  }
}'
```

Returns error:
```
"caused_by": {
  "type": "illegal_state_exception",
  "reason": "A document doesn't have a value for a field! Use doc[<field>].size()==0 to check if a document is missing a field!"
}
```

---

## Tentative Proposed Fix

### Possible Approach (Tentative - Needs Investiagtion)

The fix requires modifying how nested fields are accessed in filter scripts. Several approaches are possible:

**Option 1: Fallback to Source Access**
Modify `ScriptDataContext.get()` to detect nested fields and fall back to `_source` access:

1. **Detection**: Check if the field path corresponds to a nested field in the index mapping
2. **Fallback**: If doc values are unavailable and the field is nested, use `sourceLookup.get(name)` instead
3. **Implementation Location**: `/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java` (ScriptDataContext class)

```java
@Override
public Object get(String name) {
  // ... existing UTC_TIMESTAMP handling ...
  
  ScriptDocValues<?> docValue = this.docProvider.get(name);
  if (docValue == null || docValue.isEmpty()) {
    // Try to get from source for nested fields
    Object sourceValue = this.sourceLookup.get(name);
    if (sourceValue != null) {
      return sourceValue;
    }
    return null;
  }
  // ... existing value processing ...
}
```

**Option 2: Nested Field Detection in Code Generation**
Modify the Calcite code generation to use `getFromSource()` for nested fields:
- Detect nested fields during query planning
- Generate code that calls `getFromSource()` instead of `get()` for nested field paths
- More complex but potentially more efficient

**Option 3: Nested Query Wrapper**
Instead of using script queries for nested fields, generate proper nested queries:
- Detect when filter expressions reference nested fields
- Generate OpenSearch nested query DSL instead of script queries
- Most efficient but requires significant refactoring

### Implementation Considerations

- **Performance**: `_source` access is slower than doc values, but it's the only option for nested fields
- **Mapping Access**: Need access to index mapping to detect nested fields
- **Array Handling**: Nested fields are arrays; need to handle multiple values correctly
- **Backward Compatibility**: Ensure the fix doesn't break existing functionality for regular fields

---

## Related Issues

- #4487 - PPL dedup command returns zero results on nested field paths (different manifestation of nested field handling issues)
- #3452 - Support nested (object) fields with Calcite

---

## Workaround

**No easy workaround available.** The issue is fundamental to how nested fields are accessed in filter scripts. Possible alternatives:

1. **Restructure Data**: Flatten nested structures into regular fields (not always feasible)
2. **Pre-filter Before Eval**: Apply filters before computing fields (doesn't work when filter depends on computed value)
3. **Use Regular Fields**: If possible, use non-nested field mappings

None of these are practical workarounds for most use cases involving nested data structures.

yuancu:
### Description

Because [nested fields](https://docs.opensearch.org/latest/mappings/supported-field-types/nested/) are indexed as hidden documents, we cannot query them directly. Instead, we have to use the [nested query](https://docs.opensearch.org/latest/query-dsl/joining/nested/) to access them.

For example, for the following mapping:
```json
{
  "mappings": {
    "properties": {
      "id": {"type": "keyword"},
      "items": {
        "type": "nested",
        "properties": {
          "name": {"type": "keyword"}
        }
      }
    }
  }
}
```

if we want to use term query to match those items with name `banana`, instead of using:
```json
{
  "query": {
    "term": {
      "items.name": {
        "value": "banana",
        "boost": 1
      }
    }
  }
}
```

We should use
```json
{
  "query": {
    "nested": {
      "path": "items",
      "query": {
        "term": {
          "items.name": {
            "value": "banana",
            "boost": 1
          }
        }
      }
    }
  }
}
```

Here, the `nested` clause *steps down* into the nested `items` field. It no longer has access to fields in the root document, nor fields in any other nested document.

### Work items

- [x] Fix the case in the issue #4508: script's access to a single nested field.
- [x] Support querying deep nested objects: if a field of a nested object is another nested object, we should construct a nested query using the deepest level
- [x] Support nested query in other types of filter.
  - [x] Term query
  - [x] Range query
  - [x] Multi-terms query
- [x] ~~Support filters on both nested and root objects, where they can be separated. E.g. `| where items.name = 'banana' and id = 2` can be separated to two filters: ` items.name = 'banana'` stays under a nested query, while `id = 2` stays on a normal term query, combined with a `bool` query.~~  **Already implemented**.
- [x] Support filters on nested and root objects, where they can not be separated

### Related Issues
Resolves #4508 

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
@@ -61,6 +61,7 @@ import java.util.Set;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import org.apache.calcite.plan.RelOptCluster;
@@ -1613,7 +1614,7 @@ public class PredicateAnalyzer {
 
     private final String name;
     private final ExprType type;
-    @Getter private final String nestedPath;
+    @Getter @Nullable private final String nestedPath;
 
     public NamedFieldExpression(
         int refIndex, List<String> schema, Map<String, ExprType> filedTypes) {
@@ -1623,7 +1624,7 @@ public class PredicateAnalyzer {
     }
 
     private NamedFieldExpression() {
-      this(null, null, "");
+      this(null, null, (String) null);
     }
 
     private NamedFieldExpression(
@@ -1635,7 +1636,7 @@ public class PredicateAnalyzer {
     }
 
     private NamedFieldExpression(RexLiteral literal) {
-      this(literal == null ? null : RexLiteral.stringValue(literal), null, "");
+      this(literal == null ? null : RexLiteral.stringValue(literal), null, null);
     }
 
     public String getRootName() {


opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1485,7 +1504,25 @@ private String getOrCreateGeneratedCode() {
...
+      if (nestedPaths.size() > 1) {
+        throw new UnsupportedScriptException(
+            String.format(
+                Locale.ROOT,
+                "Accessing multiple nested fields under different hierarchies in script is not"
+                    + " supported: %s",
+                nestedPaths));
+      }
+      if (!nestedPaths.isEmpty()) {
+        return nestedQuery(nestedPaths.get(0), scriptQuery, ScoreMode.None);
```
qianheng-aws, line 1534: Since we have supported Agg script and Sort script now,  I think we need to identify whether it's a filter script before wrapping it with nested query. And please add a test on agg script with nested  fields.

yuancu: - Script sort does not make use of `PredicateAnalyzer.QueryExpression`, thus does not interleave with the current change:
https://github.com/opensearch-project/sql/blob/c31227cd98e0030f7a1a89cf1819c6354e480104/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java#L499
- The same is with field sort
https://github.com/opensearch-project/sql/blob/c31227cd98e0030f7a1a89cf1819c6354e480104/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java#L476

However, they still need nested queries to access nested fields. As a result, the following queries does not work (`author` is a nested field)
- script sort: `source=opensearch-sql_test_index_cascaded_nested | eval lower_name = lower(author.name) | sort lower_name`
- field sort: `source=opensearch-sql_test_index_cascaded_nested | sort author.name`

Should I fix them in this PR or another one?

For agg script, I added a test case. Yet I doubt I still miss many cases? Originally I thought they were only handled in `AggregateAnalyzer`, so I could raise another PR for nested fields in aggregations.

LantaoJin (commented):
This PR doesn't count in the case that the filter is not pushed down.
```
source=t | head 10000 | where items.name = "xx"
```
Above query will return incorrect results.

This issue is similar to the #3696 which I am working on refactor to avoid the correctness issue in non-pushdown case.

integ-test/src/test/resources/expectedOutput/calcite/explain_dedup_expr4_alternative_mutated.yaml
```diff
@@ -0,0 +1,12 @@
+calcite:
```
LantaoJin: why the PR add this yaml file?

integ-test/src/test/resources/expectedOutput/calcite/explain_dedup_with_expr4_mutated.yaml
```diff
@@ -0,0 +1,12 @@
+calcite:
```
LantaoJin: ditto

integ-test/src/test/resources/expectedOutput/calcite/explain_dedup_expr4_alternative_mutated.yaml
yuancu: I accidentally added them. Removed.

Thanks for pointing out. They were used only for local tests.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1501,7 +1515,25 @@ private String getOrCreateGeneratedCode() {
...
     public QueryBuilder builder() {
-      return new ScriptQueryBuilder(getScript());
+      ScriptQueryBuilder scriptQuery = QueryBuilders.scriptQuery(getScript());
+      List<String> nestedPaths =
+          referredFields.stream()
+              .map(p -> Utils.resolveNestedPath(p, fieldTypes))
+              .filter(Predicate.not(Strings::isNullOrEmpty))
+              .distinct()
+              .collect(Collectors.toUnmodifiableList());
+      if (nestedPaths.size() > 1) {
```
qianheng-aws, line 1525: Do we have test for this case?

qianheng-aws: Seems like a very corner case.

yuancu: Yes, `testScriptFilterOnDifferentNestedHierarchyShouldThrow` tests this case. It happens when users try to operate on two nested fields of different levels.

PR #4830, opened 2025-11-20
Support sort expression pushdown for SortMergeJoin

### LINKED ISSUE #4823: [FEATURE] Support sort expressions pushdown for SortMergeJoin
See response here: https://github.com/opensearch-project/sql/pull/4750#issuecomment-3546140480

Sort by complex expressions pushdown is better to support Enumerable convention. So SortMergeJoin with complex expressions will be able to push down sort by complex expressions as well

songkant-aws:
### Description
We observed there are some common user queries following the pattern like `source=tableA | rex '...<?rex_field>...' | join left=a right=b on a.rex_field=b.field tableB`. Join keys reference to a expression output instead of original table field. This query will be transformed to SortMergeJoin where it sorts by join keys on each join child.

SortMergeJoin inserts EnumerableSort operators at physical plan optimization. Previous sort expression pushdown optimization #4750 only provides ability of logical plan optimization, aka pushdown LogicalSort into scan. This PR expands the ability to push down EnumerableSort by simple or complex sort expressions in case of SortMergeJoin.

### Related Issues
Resolves #4823 

### Check List


integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -956,6 +956,37 @@ public void testSortComplexExprMixedWithSimpleExpr() throws Exception {
...
+    String expected = loadExpectedPlan("explain_complex_sort_expr_pushdown_for_smj.yaml");
+    assertYamlEqualsIgnoreId(expected, result);
+  }
+
+  @Test
+  public void testSimpleSortExprPushdownForSMJ() throws Exception {
+    String query =
+        "source=opensearch-sql_test_index_bank | join left=a right=b on a.age + 1 = b.balance - 20"
+            + " opensearch-sql_test_index_bank";
+    var result = explainQueryYaml(query);
```
yuancu, line 1030: I found that this query is not executable with the following error:

```
java.lang.RuntimeException: java.sql.SQLException: exception while executing query: class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')
        at org.opensearch.sql.opensearch.executor.OpenSearchExecutionEngine.lambda$execute$2(OpenSearchExecutionEngine.java:217) ~[opensearch-sql-3.4.0.0-SNAPSHOT.jar:3.4.0.0-SNAPSHOT]
```
It may result from the type difference between age and balance. I see quite a lot type cast to `Integer` in the generated code.

This should be another problem that's irrelevant to this optimization though.

LantaoJin: @songkant-aws does this issue be fixed? I don't see any IT for correctness. can you add more tests in `CalcitePPLJoinIT`?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -1012,6 +1012,37 @@ public void testSortComplexExprMixedWithSimpleExpr() throws Exception {
     assertYamlEqualsIgnoreId(expected, result);
   }
 
+  @Test
+  public void testComplexSortExprPushdownForSMJ() throws Exception {
+    String query =
+        "source=opensearch-sql_test_index_bank | rex field=lastname \\\"(?<initial>^[A-Z])\\\" |"
+            + " join left=a right=b on a.initial = b.firstname opensearch-sql_test_index_bank";
```
LantaoJin, line 1019: add a case of
```
source=opensearch-sql_test_index_bank
| rex field=lastname \\\"(?<initial>^[A-Z])\\\"
| join type=left max=1 initial opensearch-sql_test_index_bank
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -409,4 +496,22 @@ public boolean isScriptPushed() {
...
+   *
+   * @param relDataType the return type of the expression
+   * @return the appropriate ScriptSortType
+   */
+  private ScriptSortType getScriptSortType(RelDataType relDataType) {
+    if (SqlTypeName.CHAR_TYPES.contains(relDataType.getSqlTypeName())) {
+      return ScriptSortType.STRING;
+    } else if (SqlTypeName.INT_TYPES.contains(relDataType.getSqlTypeName())
+        || SqlTypeName.APPROX_TYPES.contains(relDataType.getSqlTypeName())) {
+      return ScriptSortType.NUMBER;
```
LantaoJin, line 511: could it check type family?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
songkant-aws: This is caused by type mismatch at runtime due to different index mapping types. The plan explanation can still pass. I think it's not a scope of this PR. Dynamic type conversion at runtime or type validation checker could potentially address it.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
songkant-aws: The SqlTypeFamily check is almost the same with above raw types check. Also, SqlTypeFamily check allows ANY type, which is not expected here.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
songkant-aws: Added this explain test, and some other IT test cases to verify correctness.

PR #4839, opened 2025-11-20
Support `mvfind`  eval function

ahkcs:
## Description

  This PR implements the `mvfind` eval function for PPL, which searches a multivalue array and returns the 0-based index of the first element matching a regular expression pattern. Returns NULL if no match is found.

  ## Implementation Approach

  ### Attempted: Native Calcite SQL Functions
  We initially explored using native Calcite SQL functions to avoid creating a UDF. We tried ARRAY-related Calcite functions like `ARRAY_POSITION`, but we didn't find any functions that support REGEX. For functions that do support REGEX (like `REGEXP_CONTAINS`, `REGEXP_LIKE`), they all take string as a parameter instead of array.

  ### Final Solution: UDF Implementation
  Implemented as a UDF following the pattern of `EXISTS` and `FORALL` functions, using Java's `Pattern` and `Matcher` for regex matching.

  ## Examples

  ```sql
  -- Basic pattern matching
  source=people | eval result = mvfind(array('apple', 'banana', 'apricot'), 'ban.*') | fields result
  > 1

  -- No match returns NULL
  source=people | eval result = mvfind(array('cat', 'dog', 'bird'), 'fish') | fields result
  > null

  -- Regex with character class
  source=people | eval result = mvfind(array('error123', 'info', 'error456'), 'error[0-9]+') | fields result
  > 0

  -- Case-insensitive matching
  source=people | eval result = mvfind(array('Apple', 'Banana', 'Cherry'), '(?i)banana') | fields result
  > 1

```

### Check List

### DIFF
--- a/core/src/test/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImplTest.java
+++ b/core/src/test/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImplTest.java
@@ -8,6 +8,7 @@ package org.opensearch.sql.expression.function.CollectionUDF;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Arrays;
 import java.util.Collections;
@@ -168,8 +169,8 @@ public class MVFindFunctionImplTest {
             IllegalArgumentException.class,
             () -> MVFindFunctionImpl.evalWithString(array, "[invalid"));
     // Verify error message contains pattern details
-    assert exception.getMessage().contains("Invalid regex pattern");
-    assert exception.getMessage().contains("[invalid");
+    assertTrue(exception.getMessage().contains("Invalid regex pattern"));
+    assertTrue(exception.getMessage().contains("[invalid"));
   }
 
   // Type conversion tests

### CODE REVIEW

File: core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java [hunk 1]
```diff
@@ -0,0 +1,99 @@
...
+ * <p>Usage: mvfind(array, regex)
+ *
+ * <p>Returns the 0-based index of the first array element matching the regex pattern, or NULL if no
+ * match is found.
+ *
+ * <p>Example: mvfind(array('apple', 'banana', 'apricot'), 'ban.*') returns 1
+ */
+public class MVFindFunctionImpl extends ImplementorUDF {
+  public MVFindFunctionImpl() {
+    super(new MVFindImplementor(), NullPolicy.ALL);
```
ykmr1224: Should it be NullPolicy.ANY? (looks it returns null when one of the parameter is null)

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,99 @@
...
+
+    List<Object> array = (List<Object>) args[0];
+    String regex = (String) args[1];
+
+    if (array == null || regex == null) {
+      return null;
+    }
+
+    try {
+      Pattern pattern = Pattern.compile(regex);
```
ykmr1224: Is it possible to avoid compiling everytime?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return ReturnTypes.INTEGER_NULLABLE;
+  }
+
+  @Override
+  public UDFOperandMetadata getOperandMetadata() {
+    return UDFOperandMetadata.wrap(
+        OperandTypes.family(SqlTypeFamily.ARRAY, SqlTypeFamily.CHARACTER));
```
ykmr1224, line 51: Does type coercion work and accept number as regex?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    List<Object> array = (List<Object>) args[0];
+    String regex = (String) args[1];
+
+    if (array == null || regex == null) {
+      return null;
+    }
+
+    try {
```
ykmr1224, line 159: nit: I'd recommend having separate method `mvfind(List<Object> array, String regex)` to separate preprocess and core logic.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,99 @@
...
+        Object element = array.get(i);
+        if (element != null) {
+          String strValue = element.toString();
+          if (pattern.matcher(strValue).find()) {
+            return i; // Return 0-based index
+          }
+        }
+      }
+    } catch (Exception e) {
+      throw new RuntimeException("Error in mvfind function: " + e.getMessage(), e);
```
ykmr1224: Does this lead invalid regex to be 5xx error? Can we check regex earlier (during planning) and raise 4xx error instead?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Updated to NullPolicy.ANY

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Moved the compiling to planning time

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Updated to support type coercion, also added corresponding UT

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Updated to separate method

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Updated error handling

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ykmr1224: I thought `CoercionUtils` would automatically add cast.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  /**
+   * Core mvfind logic that searches for a pattern in an array.
+   *
+   * @param array The array to search
+   * @param pattern The pre-compiled regex pattern
+   * @return The 0-based index of the first matching element, or null if no match
+   */
```
ykmr1224: nit: seems redundant comment.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,175 @@
...
+                Types.lookupMethod(
+                    MVFindFunctionImpl.class, "evalWithPattern", List.class, Pattern.class),
+                arrayExpr,
+                Expressions.constant(compiledPattern, Pattern.class));
+          } catch (PatternSyntaxException e) {
+            // Convert to IllegalArgumentException so it's treated as a client error (400)
+            throw new IllegalArgumentException(
+                String.format("Invalid regex pattern '%s': %s", patternString, e.getDescription()),
+                e);
+          }
```
ykmr1224: nit: Let's extract for maintainability/readability.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  /**
+   * Evaluates mvfind with a String pattern. Compiles the regex pattern and executes search.
+   *
+   * @param array The array to search
+   * @param regex The regex pattern string
+   * @return The 0-based index of the first matching element, or null if no match
+   */
+  public static Integer mvfind(List<Object> array, String regex) {
```
ykmr1224: Seems it is used only by `evalWithString`. Let's move it to just below `evalWithString`.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,175 @@
...
+          } catch (PatternSyntaxException e) {
+            // Convert to IllegalArgumentException so it's treated as a client error (400)
+            throw new IllegalArgumentException(
+                String.format("Invalid regex pattern '%s': %s", patternString, e.getDescription()),
+                e);
+          }
+        }
+      }
+
+      // For null or dynamic patterns, use evalWithString
```
ykmr1224: I think calcite automatically return null in case of null parameter.

core/src/test/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImplTest.java
```diff
@@ -0,0 +1,232 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+/** Unit tests for MVFindFunctionImpl. */
```
ykmr1224: nitnit: redundant comment

ahkcs: Removed

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Updated comment

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Updated

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Updated

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Removed

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: After checking, the tests would fail if we go through CoercionUtils, it seems that we don't have coercion rule for Integer -> String?

ykmr1224:
Can you resolve conflict?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,178 @@
...
+      // Check if regex pattern is a literal - compile at planning time
+      if (call.operands.size() >= 2 && call.operands.get(1) instanceof RexLiteral) {
+        RexLiteral patternLiteral = (RexLiteral) call.operands.get(1);
+        Expression literalPatternExpr = tryCompileLiteralPattern(patternLiteral, arrayExpr);
+        if (literalPatternExpr != null) {
+          return literalPatternExpr;
+        }
+      }
+
+      // For dynamic patterns, use evalWithString
```
dai-chen, line 70: There seems no IT to cover non-literal regex. Just wonder is this required and supported in other PPL regex function/command? Not sure what's the performance impact for this.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
```diff
@@ -0,0 +1,178 @@
...
+
+  @Override
+  public SqlReturnTypeInference getReturnTypeInference() {
+    return ReturnTypes.INTEGER_NULLABLE;
+  }
+
+  @Override
+  public UDFOperandMetadata getOperandMetadata() {
+    // Accept ARRAY and any type for the regex pattern (coerced to string via toString())
+    return UDFOperandMetadata.wrap(OperandTypes.family(SqlTypeFamily.ARRAY, SqlTypeFamily.ANY));
```
dai-chen: For the second argument, I think type coercion should be defined and performed in single place?

ahkcs: Hi Chen, after checking, I updated to restrict regex parameter to STRING type only. Removed implicit type coercion
(integer -> string) since the regex pattern should always be a string.
Non-string types will now fail at type checking.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImpl.java
ahkcs: Updated to add an IT `testMvfindWithDynamicRegex`

core/src/test/java/org/opensearch/sql/expression/function/CollectionUDF/MVFindFunctionImplTest.java
```diff
@@ -0,0 +1,215 @@
...
+  public void testMvfindWithInvalidRegex() {
+    List<Object> array = Arrays.asList("test");
+    // Invalid regex should throw IllegalArgumentException (client error - 400)
+    IllegalArgumentException exception =
+        assertThrows(
+            IllegalArgumentException.class,
+            () -> MVFindFunctionImpl.evalWithString(array, "[invalid"));
+    // Verify error message contains pattern details
+    assert exception.getMessage().contains("Invalid regex pattern");
+    assert exception.getMessage().contains("[invalid");
```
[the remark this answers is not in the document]
ahkcs: Updated

ykmr1224 (approved):

Issue #4811, opened 2025-11-14
[BUG] Push down redundant filter for time span


qianheng-aws:
**What is the bug?**
Currently, for time span agg, we always view it as `bucket_nullalbe=false` and won't show `null` bucket after this PR:https://github.com/opensearch-project/sql/pull/4327. We implements this by adding another filter operator on the time field before aggregation.

After push down span by using `date_histogram`, it actually won't have `null` bucket already. But we still push down a redundant filter into the scan. e.g. 

```
// PPL 
source=events  |  stats count() by span(@timestamp, 1d)

// Final plan
CalciteEnumerableIndexScan(table=[[OpenSearch, events]], PushDownContext=[[PROJECT->[@timestamp], FILTER->IS NOT NULL($0), AGGREGATION->rel#630:LogicalAggregate.NONE.[](input=RelSubset#629,group={0},count()=COUNT()), PROJECT->[count(), span(@timestamp,1d)], LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","query":{"exists":{"field":"@timestamp","boost":1.0}},"_source":{"includes":["@timestamp"],"excludes":[]},"aggregations":{"composite_buckets":{"composite":{"size":10000,"sources":[{"span(@timestamp,1d)":{"date_histogram":{"field":"@timestamp","missing_bucket":false,"order":"asc","fixed_interval":"1d"}}}]}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```

It has `FILTER->IS NOT NULL($0)` in the PushDownContext and `"query":{"exists":{"field":"@timestamp","boost":1.0}}` in the DSL.

It will introduce more performance downgrade if the bucket field is a derived field, which will generate script in the DSL query.

**How can one reproduce the bug?**
1. Create a index with time field, e.g.
```
PUT localhost:9200/events
{
  "mappings": {
    "properties": {
      "@timestamp": {
        "type": "date"
      },
      "host": {
        "type": "text"
      },
      "cpu_usage": {
        "type": "double"
      },
      "region": {
        "type": "keyword"
      }
    }
  }
}
```

2. Run explain on a query with time span, e.g.
```
source=events  |  stats count() by span(@timestamp, 1d)
```

**What is the expected behavior?**
The final plan shouldn't contain the filter derived from time span aggregation.

**What is your host/environment?**
 - Version 3.4-SNAPSHOT
 - Plugins

[...]

PR #4843, opened 2025-11-21
Remove redundant push-down-filters derived for bucket-non-null agg

qianheng-aws:
### Description
In our current implementation/final plan, we always push down a redundant filter of ISNOTNULL for group fields if params `BUCKET_NULLABLE=false` or the group field is time span. However, since we've set `MISSING_BUCKET` to false in the aggregation builder, we actually don't need this filter which will introduce more overhead especially the filter is pushed down to be a script.

This PR enhance the above case by adding a new rule, in which we detect the derived filter from bucket-non-null agg and then ignore them in the push down process.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/4811

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/AggregateIndexScanRule.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/AggregateIndexScanRule.java
@@ -55,11 +55,11 @@ public class AggregateIndexScanRule extends RelRule<AggregateIndexScanRule.Confi
       final LogicalFilter filter = call.rel(2);
       final LogicalProject bottomProject = call.rel(3);
       final CalciteLogicalIndexScan scan = call.rel(4);
-      boolean isBucketNullable = Config.isBucketNullableAgg.test(aggregate);
+      boolean ignoreNullBucket = Config.aggIgnoreNullBucket.test(aggregate);
       List<Integer> groupRefList =
           aggregate.getGroupSet().asList().stream()
               .map(topProject.getProjects()::get)
-              .filter(rex -> isBucketNullable || isTimeSpan(rex))
+              .filter(rex -> ignoreNullBucket || isTimeSpan(rex))
               .flatMap(expr -> PlanUtils.getInputRefs(expr).stream())
               .map(RexSlot::getIndex)
               .toList();
@@ -199,7 +199,7 @@ public class AggregateIndexScanRule extends RelRule<AggregateIndexScanRule.Confi
                                         Predicate.not(AbstractCalciteIndexScan::isLimitPushed)
                                             .and(AbstractCalciteIndexScan::noAggregatePushed))
                                     .noInputs()));
-    Predicate<Aggregate> isBucketNullableAgg =
+    Predicate<Aggregate> aggIgnoreNullBucket =
         agg ->
             agg.getHints().stream()
                 .anyMatch(
@@ -221,7 +221,7 @@ public class AggregateIndexScanRule extends RelRule<AggregateIndexScanRule.Confi
             .withOperandSupplier(
                 b0 ->
                     b0.operand(LogicalAggregate.class)
-                        .predicate(isBucketNullableAgg)
+                        .predicate(aggIgnoreNullBucket)
                         .oneInput(
                             b1 ->
                                 b1.operand(LogicalFilter.class)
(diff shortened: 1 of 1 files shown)


integ-test/src/test/resources/expectedOutput/calcite/clickbench/q33.yaml
```diff
@@ -8,4 +8,4 @@ calcite:
...
\ No newline at end of file
+    CalciteEnumerableIndexScan(table=[[OpenSearch, hits]], PushDownContext=[[AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0, 1},c=COUNT(),sum(IsRefresh)=SUM($2),avg(ResolutionWidth)=AVG($3)), PROJECT->[c, sum(IsRefresh), avg(ResolutionWidth), WatchID, ClientIP], SORT_AGG_METRICS->[0 DESC LAST], LIMIT->10, LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"WatchID|ClientIP":{"multi_terms":{"terms":[{"field":"WatchID"},{"field":"ClientIP"}],"size":10,"min_doc_count":1,"shard_min_doc_count":0,"show_term_doc_count_error":false,"order":[{"c":"desc"},{"_key":"asc"}]},"aggregations":{"sum(IsRefresh)":{"sum":{"field":"IsRefresh"}},"avg(ResolutionWidth)":{"avg":{"field":"ResolutionWidth"}},"c":{"value_count":{"field":"_index"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```
LantaoJin, line 11: why the groupset and SORT_AGG_METRICS digest changed.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/AggregateIndexScanRule.java
```diff
@@ -190,24 +251,66 @@ public interface Config extends OpenSearchRuleConfig {
...
-                  .allMatch(AggregateIndexScanRule.Config::isNotNullOnRef));
+              && rexCall.getOperands().stream().allMatch(AggregateIndexScanRule.Config::isNotNull));
     }
 
-    private static boolean isNotNullOnRef(RexNode rex) {
+    private static boolean isNotNull(RexNode rex) {
       return rex instanceof RexCall rexCall
-          && rexCall.getOperator().equals(SqlStdOperatorTable.IS_NOT_NULL)
-          && rexCall.getOperands().get(0) instanceof RexInputRef;
+          && rexCall.getOperator().equals(SqlStdOperatorTable.IS_NOT_NULL);
```
LantaoJin: could be simplified to `rex.isA(SqlKind.IS_NOT_NULL)`

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q33.yaml
qianheng-aws: Before this PR, the plan can only match the rule `agg-filter-project-scan` after being transformed by some calcite's transpose rule and project merge rule. So the index in group set is different from the original plan(i.e. logical plan).

After adding another new rule `agg-project-filter-project-scan`, the original plan can match this new one after trimming(it will add another project before scan), so the new digest has exactly the same ref index in the group set as logical plan.

But overall, though the aggregation digest is different, their generated request builders are exactly the same.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/AggregateIndexScanRule.java
```diff
@@ -40,23 +49,46 @@ protected AggregateIndexScanRule(Config config) {
...
   @Override
   public void onMatch(RelOptRuleCall call) {
-    if (call.rels.length == 4) {
+    if (call.rels.length == 5) {
+      final LogicalAggregate aggregate = call.rel(0);
+      final LogicalProject topProject = call.rel(1);
+      final LogicalFilter filter = call.rel(2);
+      final LogicalProject bottomProject = call.rel(3);
+      final CalciteLogicalIndexScan scan = call.rel(4);
+      boolean isBucketNullable = Config.isBucketNullableAgg.test(aggregate);
```
yuancu: nit: I feel it's a little confusing because it's semantics seem to be *buket not nullable*

qianheng-aws: You're right. That's a typo. Will correct it.

PR #4844, opened 2025-11-21
Convert `dedup` pushdown to composite + top_hits

### LINKED ISSUE #4797: [ENHANCEMENT] Convert dedup pushdown to `composite + top_hits`
#3972 converts `dedup` command to `collapse search` when pushdown enabled. By benchmarking, the performance of  `collapse search` shows worse than composite + top_hits or terms + top_hits. Here is the benchmarking result:
1. terms + top_hits **2000ms**
2. composite + top_hits (missing_bucket false/true:) **3500ms/4500ms**
3. collapse **13000ms**

Additional, (1) and (2) can support dedup on **bool**, **date** etc fields and **script** expr, (3) only works on keyword/numeric field. and (2) can support **search_after** in future

Refactor to approach (2)

```
curl -XPOST "http://localhost:9200/big5/_search" -H 'Content-Type: application/json' -d'
{
  "size": 0,
  "_source": false,
  "aggs": {
    "packets": {
      "terms": {
        "field": "cloud.region",
        "size": 10000
      },
      "aggs": {
        "topN": {
          "top_hits": {
            "size": 1,
            "_source": true
          }
        }
      }
    }
  }
}'
```
2000ms

```
curl -XPOST "http://localhost:9200/big5/_search" -H 'Content-Type: application/json' -d'
{
  "size": 0,
  "timeout": "1m",
  "_source": false,
  "aggs": {
    "packets": {
      "composite": {
        "size": 10000,
        "sources": [
          {
            "name": {
              "terms": {
                "field": "cloud.region",
                "missing_bucket": true
              }
            }
          }
        ]
      },
      "aggs": {
        "topN": {
          "top_hits": {
            "size": 1,
            "_source": true
          }
        }
      }
    }
  }
}'
```
missing_bucket false 3500ms 
missing_bucket true  4500ms
```
curl -XPOST "http://localhost:9200/big5/_search" -H 'Content-Type: application/json' -d'
{
  "from": 0,
  "size": 10000,
  "timeout": "1m",
  "_source": true,
  "collapse": {
    "field": "cloud.region"
  }
}'
```
13000ms

LantaoJin:
### Description
Convert `dedup` pushdown to composite + top_hits
Main changes:
1. Add a rule `DedupPushdownRule` to convert the dedup plan pattern to Aggregate with TopHits metrics
2. Upgrade TopHitsParser to support parsing data from `_source` (fetchField cannot work for OS Object type)
3. Corresponding API change on `MetricParser`: `Map<String, Object> parse(Aggregation aggregation);` ->  `List<Map<String, Object>> parse(Aggregation aggregation);`

Follow-ups: Support dedup on script (expression)

### Related Issues
Resolves #4797 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -341,10 +347,13 @@ private static Pair<Builder, List<MetricParser>> processAggregateCalls(
     return Pair.of(metricBuilder, metricParserList);
   }
 
-  private static List<RexNode> convertAggArgThroughProject(AggregateCall aggCall, Project project) {
+  private static List<RexNode> convertAggArgThroughProject(
+      AggregateCall aggCall, Project project, AggregateAnalyzer.AggregateBuilderHelper helper) {
     return project == null
         ? List.of()
-        : aggCall.getArgList().stream().map(project.getProjects()::get).toList();
+        : PlanUtils.getObjectFromLiteralAgg(aggCall) != null
```
qianheng-aws, line 363: Can you elaborate more on why calling this method here?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -341,10 +347,13 @@ private static Pair<Builder, List<MetricParser>> processAggregateCalls(
     return Pair.of(metricBuilder, metricParserList);
   }
 
-  private static List<RexNode> convertAggArgThroughProject(AggregateCall aggCall, Project project) {
+  private static List<RexNode> convertAggArgThroughProject(
+      AggregateCall aggCall, Project project, AggregateAnalyzer.AggregateBuilderHelper helper) {
```
qianheng-aws: nit: `helper` seems never used?

integ-test/src/test/resources/expectedOutput/calcite/explain_dedup_complex1.yaml
```diff
@@ -0,0 +1,10 @@
...
...eEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]], PushDownContext=[[PROJECT->[account_number, firstname, address, balance, gender, city, employer, state, age, email, lastname], FILTER->IS NOT NULL($4), AGGREGATION->rel#:LogicalAggregate.NONE.[](input=LogicalProject#,group={0},agg#0=LITERAL_AGG(1)), LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","query":{"exists":{"field":"gender","boost":1.0}},"_source":{"includes":["account_number","firstname","address","balance","gender","city","employer","state","age","email","lastname"],"excludes":[]},"aggregations":{"composite_buckets":{"composite":{"size":1000,"sources":[{"gender":{"terms":{"field":"gender.keyword","missing_bucket":false,"order":"asc"}}}]},"aggregations":{"$f1":{"top_hits":{"from":0,"size":1,"version":false,"seq_no_primary_term":false,"explain":false,"_source":false,"fields":[{"field":"gender"},{"field":"account_number"},{"field":"firstname"},{"field":"address"},{"field":"balance"},{"field":"city"},{"field":"employer"},{"field":"state"},{"field":"age"},{"field":"email"},{"field":"lastname"}]}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```
qianheng-aws: We won't have  FILTER->IS NOT NULL($4) push down for common aggregate push down, after this PR: https://github.com/opensearch-project/sql/pull/4843

Could you check if we can remove them as well for dedup push down?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -37,87 +41,149 @@ protected DedupPushdownRule(Config config) {
...
+        .filter(rex -> !rex.isA(SqlKind.ROW_NUMBER))
+        .filter(Predicate.not(dedupColumns::contains))
+        .forEach(reordered::add);
+    relBuilder.project(reordered, ImmutableList.of(), true);
+    // childProject includes all list of finalOutput columns
+    LogicalProject childProject = (LogicalProject) relBuilder.peek();
+
+    // 3. Push an Aggregate
+    final List<RexNode> newDedupColumns = RexUtil.apply(mappingForDedupColumns, dedupColumns);
+    relBuilder.aggregate(relBuilder.groupKey(newDedupColumns), relBuilder.literalAgg(dedupNumer));
```
qianheng-aws, line 165: LITERAL_AGG in Calcite has totally different function as we use here. It seems to be tricky to implement this in this way. Do we have alternative approach? Or at least add some comments to notice that.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -135,6 +136,52 @@ public void extendTypeMapping(Map<String, OpenSearchDataType> typeMapping) {
               (c, dt) -> new OpenSearchExprBinaryValue(c.stringValue()))
           .build();
 
+  private static final Map<ExprType, TriFunction<Content, ExprType, Boolean, ExprValue>>
+      typeActionMap2 =
```
qianheng-aws: Seems used nowhere.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -530,6 +539,33 @@ yield switch (functionName) {
...
       }
+      case LITERAL_AGG -> {
+        RexLiteral literal = PlanUtils.getObjectFromLiteralAgg(aggCall);
+        if (literal == null || !(literal.getValue() instanceof Number)) {
+          throw new AggregateAnalyzer.AggregateAnalyzerException(
+              String.format("Unsupported push-down aggregator %s", aggCall.getAggregation()));
+        }
+        Integer dedupNumber = literal.getValueAs(Integer.class);
+        TopHitsAggregationBuilder topHitsAggregationBuilder =
+            AggregationBuilders.topHits(aggFieldName).from(0).fetchSource(false).size(dedupNumber);
```
qianheng-aws: Can you elaborate more on why we set `fetchSource(false)` here? The default value is true?

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/TopHitsParser.java
```diff
@@ -21,45 +26,96 @@ public class TopHitsParser implements MetricParser {
...
+      //       }
+      //     }
+      //   ]
+      // }
+      // will converts to:
+      // List[
+      //   LinkedHashMap["name" -> "A", "category" -> "X"],
+      //   LinkedHashMap["name" -> "A", "category" -> "X"]
+      // ]
+      List<Map<String, Object>> res =
```
qianheng-aws: [question] Is this the only case that agg metric parser should return a List of value for each bucket? I'm thinking Is it still appropriate to  translate `dedup x` to `aggregate`? Normally speaking, `aggregate` should return only 1 row value for the metric in each bucket. While `dedup x` can return more than 1 rows and it now affects the API of agg metric parser.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: Is it for identifying LITERAL_AGG?

LantaoJin: Yes. `LITERAL_AGG` passes the `numberOfDedup` here

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
LantaoJin: will remove it

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
LantaoJin: I don't have other alternative approach. I think it's safe since `LITERAL_AGG` usually used in Project removing for Aggregate. And PPL doesn't have explicit syntax to call `LITERAL_AGG`. I will add comment to elaborate.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
LantaoJin: Update: change to use `fetchSource` instead of `fetchField` due to `fetchField` cannot work on OS object type.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/TopHitsParser.java
LantaoJin: > [question] Is this the only case that agg metric parser should return a List of value for each bucket?

Yeah, just `top_hits` metric agg does. `top_hits` metric agg in OpenSearch can be used in any bucket agg in DSL as a sub-aggregation, it's quite different with SQL' aggregate.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
LantaoJin: will remove

integ-test/src/test/resources/expectedOutput/calcite/explain_dedup_complex1.yaml
LantaoJin: done

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1601,15 +1600,15 @@ private static void buildDedupNotNull(
             .partitionBy(dedupeFields)
             .orderBy(dedupeFields)
```
yuancu, line 1601: Will `dedupe` work without ordering by deduped fields?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
```diff
@@ -73,6 +73,8 @@ public class OpenSearchRequestBuilder {
 
   @ToString.Exclude private final Settings settings;
 
+  @ToString.Exclude private boolean topHitsAgg = false;
```
yuancu, line 76: What is this field used for? It seems it's not referred anywhere

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -1788,4 +1792,115 @@ public void testInternalItemAccessOnStructs() throws IOException {
...
+
+  @Ignore("https://github.com/opensearch-project/sql/issues/4789")
+  public void testDedupExpr() throws IOException {
+    enabledOnlyWhenPushdownIsEnabled();
+    String expected = loadExpectedPlan("explain_dedup_expr1.yaml");
+    assertYamlEqualsIgnoreId(
+        expected,
+        explainQueryYaml(
+            "source=opensearch-sql_test_index_account | eval new_gender = lower(gender) | dedup 1"
+                + " new_gender"));
```
yuancu, line 1828: Isn't such case covered by `CalcitePPLDedupIT.testDedupExpr`

qianheng-aws (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -530,6 +548,38 @@ yield switch (functionName) {
                   String.format("Unsupported push-down aggregator %s", aggCall.getAggregation()));
         };
       }
+      case LITERAL_AGG -> {
```
qianheng-aws, line 551: Can you check if we will never produce a LITERAL_AGG in RelBuilder or by some rules in planner?

Otherwise, we may push down a real LITERAL_AGG to be tophits while it shouldn't be.

qianheng-aws: One alternative in my mind is using a self-defined class extends `calcite.Aggregate`. It should also be able to leverage our AggregateAnalyzer.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: em, it should work in non-pushdown case. maybe we could remove this orderBy in window.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java
LantaoJin: yes. can be deleted in current impl

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
yuancu: It looks a little strange to me because the sort keys in a window should be the same (the partition key)

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
LantaoJin: > Can you check if we will never produce a LITERAL_AGG in RelBuilder or by some rules in planner?
> 
> Otherwise, we may push down a real LITERAL_AGG to be tophits while it shouldn't be.

Checked, so far no LITERAL_AGG can be produced because PPL doesn't support literal in aggregators.
In SQL
```
SELECT 
    dept_id,
    COUNT(*) as emp_count,
    1 as constant_val
FROM employees
GROUP BY dept_id;
```
Could be rewritten to 
```
SELECT 
    dept_id,
    COUNT(*) as emp_count,
    LITERAL_AGG(1) as constant_val
FROM employees
GROUP BY dept_id;
```
to reduce a Project upon Aggregate.

From
```
Project(dept_id, emp_count, constant_val=1)
  Aggregate(GROUP BY dept_id, compute COUNT(*) as emp_count)
    Scan(employees)
```
To
```
Aggregate(GROUP BY dept_id, compute COUNT(*) as emp_count, compute LITERAL_AGG(1) as constant_val)
  Scan(employees)
```
But `stats` only support pre-defined aggregators. cc @qianheng-aws

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
LantaoJin: Yes. it was not introduced by this pr. let's fix them in followup PR since it will change the plan a lot.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: I see `SubQueryRemoveRule` will produce LITERAL_AGG for `some`(we don't have) and `in` subquery. But I'm not sure whether it will be triggered.

And RelBuilder's literalAgg method is public, we should avoid call that method by developers then.

qianheng-aws: https://issues.apache.org/jira/browse/CALCITE-4334

PR #4856, opened 2025-11-24
Support `mvmap` eval function

ahkcs:
## **Description**

This PR adds support for the `mvmap` eval function, which iterates over each element of a multivalue array, applies a given expression, and returns a new array containing the transformed results.

**Syntax:**  
`mvmap(array, expression)`

During iteration, the field name in the expression is implicitly bound to each element of the input array (e.g., `mvmap(arr, arr * 10)`).

---

### **Core Approach**

`mvmap` is implemented as an alias for PPL's existing `TRANSFORM` operator, which applies a lambda function to each array element.

To support implicit binding syntax—`mvmap(arr, arr * 10)` instead of requiring explicit lambda syntax (`mvmap(arr, x -> x * 10)`)—AST transformation logic was added in `AstExpressionBuilder.java`.

### **AST Transformation Steps**

1. Intercept `mvmap` function calls in `visitEvalFunctionCall()`
2. Extract the field name from the first argument (e.g., `arr`)
3. Wrap the second argument expression into a `LambdaFunction`, using the extracted field name as the parameter
4. Transform:

### Check List

### DIFF
--- a/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
+++ b/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
@@ -504,9 +504,12 @@ public class AstExpressionBuilder extends OpenSearchPPLParserBaseVisitor<Unresol
 
         @Override
         public QualifiedName visitFunction(Function node, Void context) {
-          List<UnresolvedExpression> funcArgs = node.getFuncArgs();
-          if (!funcArgs.isEmpty()) {
-            return funcArgs.get(0).accept(this, context);
+          // Visit each funcArg and return on first non-null (qualified name found)
+          for (UnresolvedExpression arg : node.getFuncArgs()) {
+            QualifiedName result = arg.accept(this, context);
+            if (result != null) {
+              return result;
+            }
           }
           return null;
         }


ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -431,6 +431,12 @@ public UnresolvedExpression visitEvalFunctionCall(EvalFunctionCallContext ctx) {
     final String mappedName =
         FUNCTION_NAME_MAPPING.getOrDefault(functionName.toLowerCase(Locale.ROOT), functionName);
 
+    // Handle mvmap with implicit binding: mvmap(field, field * 10) -> mvmap(field, field -> field *
+    // 10)
+    if ("mvmap".equalsIgnoreCase(mappedName)) {
+      return buildMvmapFunction(ctx.functionArgs().functionArg());
+    }
```
dai-chen: it's cleaner to handle this in its own visit method.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -446,6 +452,58 @@ private Function buildFunction(
...
+    // Extract field name from first argument for implicit binding
+    QualifiedName fieldName = extractFieldName(firstArg);
+    if (fieldName == null) {
+      // Can't determine field name, fall back to regular function call
+      return new Function("mvmap", Arrays.asList(firstArg, secondArg));
+    }
+
+    // Wrap expression in lambda: expr -> LambdaFunction(expr, [fieldName])
+    LambdaFunction lambda = new LambdaFunction(secondArg, Collections.singletonList(fieldName));
+    return new Function("mvmap", Arrays.asList(firstArg, lambda));
```
dai-chen: I'm not sure if I understand this part. Why `secondArg` could be a lambda already? Why we do nothing if `firstArg` doesn't have field name? Both are illegal according to the grammar right?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -446,6 +452,58 @@ private Function buildFunction(
...
+  }
+
+  /** Extracts the field name from an expression for implicit lambda binding. */
+  private QualifiedName extractFieldName(UnresolvedExpression expr) {
```
dai-chen: What's the expected behavior if first arg is an complex expression? From the code, it seems we try to find the field deep nested and bind it to lambda?

dai-chen (changes requested):

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ahkcs: Hi Chen, I think it makes sense to make both cases illegal, I have updated the code to throw `SyntaxCheckException`

We allow this initially because we use `TRANSFORM` for implementation and we want to keep the lambda input as an option, but I think it's better to throw exception.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ahkcs: Yes. For complex expressions like mvmap(mvindex(arr, 1, 3), arr * 10):

  - At parse time: we recursively extract the field name (arr) from the nested function to use as the lambda variable name → mvmap(mvindex(arr, 1, 3), arr -> arr * 10)
  - At runtime: iteration happens over the evaluated result of mvindex(arr, 1, 3)

Also added IT and UT to confirm

ahkcs: Also added documentation in collection.rst

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ahkcs: Updated, Added dedicated grammar rule `mvmapFunctionCall` and `visitMvmapFunctionCall()` method

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -440,12 +440,61 @@ public UnresolvedExpression visitEvalFunctionCall(EvalFunctionCallContext ctx) {
...
     return new Function(
         functionName, args.stream().map(this::visitFunctionArg).collect(Collectors.toList()));
   }
 
+  /**
+   * Extracts the field name from the first argument for implicit lambda binding. The second
+   * argument must reference this same field. E.g., {@code mvmap(mvindex(arr, 1, 2), arr * 10)}
+   * extracts 'arr' and creates {@code arr -> arr * 10}.
+   */
+  private QualifiedName extractFieldName(UnresolvedExpression expr) {
```
dai-chen, line 486: np: use visitor to simplify and cover all cases

ykmr1224:
@coderabbitai review

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ahkcs: Refactored to use `AbstractNodeVisitor`

ahkcs:
> Is it possible to refer other fields in the second parameter?
> 
> e.g. `... | fields array, a | eval b = mvmap(array, array * a)`
> 
> If yes, we might want to add test case for that.

Updated to support singe-value field

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -448,12 +450,74 @@ public UnresolvedExpression visitEvalFunctionCall(EvalFunctionCallContext ctx) {
...
+        @Override
+        public QualifiedName visitQualifiedName(QualifiedName node, Void context) {
+          return node;
+        }
+
+        @Override
+        public QualifiedName visitFunction(Function node, Void context) {
+          List<UnresolvedExpression> funcArgs = node.getFuncArgs();
+          if (!funcArgs.isEmpty()) {
+            return funcArgs.get(0).accept(this, context);
```
dai-chen: Shall we visit each funcArg and return on first non-null (qualified name found)?

ahkcs: Updated the `visitFunction` method to iterate through all function arguments and return on the first non-null qualified name found

dai-chen (approved):

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
dai-chen: Thanks! This will also help answer Tomo's comment https://github.com/opensearch-project/sql/pull/4856#pullrequestreview-3553890924 which also ask clarify for the behavior.

PR #4892, opened 2025-12-02
Use Calcite's validation system for type checking & coercion

### LINKED ISSUE #3865: [RFC] Implicit Type Coercion with Calcite
### Problem Statement

Implicit type coercion is missing from the current PPL validation. It is useful in at least two scenarios:

1. when the actual argument types of a function don't match with its expected types (should coerce to expected param type)
2. when the two operand types of a comparator are different (should coerce to a common type)

**Relevant issue**: https://github.com/opensearch-project/sql/issues/3761

**Status update**:

- Approach 1: https://github.com/yuancu/sql-plugin/pull/6 Discarded due to the re-introduction of `SqlNode`
- Approach 2 (alternative proposal): https://github.com/opensearch-project/sql/pull/3914 Merged

### Current State

As a result, the current function implementation must have an override for each parameter type it supports, even when they could be implicitly cast. For example, a function `foo` that accepts a date as an argument should also accept `"2020-12-10"` as an argument. However, without implicit type coercion, we have to define both `foo(date)` and `foo(string)`.

### How Calcite Implements Implicit Type Coercion

Calcite checks operand types and attempts to coerce them if necessary during the **validation phase**.

For example, in FamilyOperandTypeChecker, when an argument type doesn't match expectations, Calcite tries to determine if the argument can be coerced to the expected type. The simplified [code snippet](https://github.com/apache/calcite/blob/dafb8d7a1fde884daf76d2a24aa6aea9c39ee5ca/core/src/main/java/org/apache/calcite/sql/type/FamilyOperandTypeChecker.java#L144) below illustrates this process:

```java
 public boolean checkOperandTypes(
      SqlCallBinding callBinding,
      boolean throwOnFailure){
      ...
if (!checkSingleOperandType(callBinding, op.e, op.i, families.get(op.i), false)) {
	    // try to coerce type if it is allowed.
	  boolean coerced = false;
	  if (callBinding.isTypeCoercionEnabled()) {
	    TypeCoercion typeCoercion = callBinding.getValidator().getTypeCoercion();
	    ImmutableList.Builder<RelDataType> builder = ImmutableList.builder();
	    for (int i = 0; i < callBinding.getOperandCount(); i++) {
	      builder.add(callBinding.getOperandType(i));
	    }
	    ImmutableList<RelDataType> dataTypes = builder.build();
	    coerced = typeCoercion.builtinFunctionCoercion(callBinding, dataTypes, families);
	  }
}
```

In this process, `builtinFunctionCoercion` calls [`coerceOperandType`](https://github.com/apache/calcite/blob/dafb8d7a1fde884daf76d2a24aa6aea9c39ee5ca/core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java#L119C1-L128C17) to rewrite the operands when needed. The [coerceOperandType](https://github.com/apache/calcite/blob/dafb8d7a1fde884daf76d2a24aa6aea9c39ee5ca/core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java#L119C1-L128C17) method rewrites the query by wrapping the argument with a `CAST` call:

```java
 protected boolean coerceOperandType(SqlValidatorScope scope, SqlCall call, int index, RelDataType targetType) {
		 ...
    RelDataType targetType1 = syncAttributes(operandType, targetType);
    SqlNode desired = castTo(operand, targetType1);
    call.setOperand(index, desired);
    updateInferredType(desired, targetType1);
    return true;
  }
```

To determine whether an argument can be cast to the target type, Calcite defines rules in 
`SqlTypeCoercionRule` and `SqlTypeAssignmentRule`. Additionally, the `TypeCoercion` interface defines how a type should be cast to another.

### Challenges in Applying Calcite's Type Coercion

The validation phase of Calcite SQL validates a parsed `SqlNode` AST. It resolves identifiers, derives return types, validates argument types, etc. However, since we created a new language interface outside of SQL, this phase isn't directly applicable and is **skipped** in our current implementation.

As a result, we can't directly utilize Calcite's type coercion mechanism.

### Proposal

We faced a similar issue with function argument type checking. We [implemented](https://github.com/opensearch-project/sql/pull/3626) our own argument validation initially. ~~However, later we [discovered](https://github.com/opensearch-project/sql/pull/3626#discussion_r2104225644) that other language interfaces—like [calcite-piglet](https://calcite.apache.org/javadocAggregate/org/apache/calcite/piglet/package-summary.html)—successfully leverage Calcite's type validation.~~ **Update**: with a deeper inspection, I found that piglet actually uses its own type check, without re-using Calcite's type checking mechanism at all.

Calcite-piglet translates its logical plans into a SQL AST (represented with `SqlNode`) to reuse Calcite's type checking. It first constructs its own AST via a custom parser, then converts each node into a corresponding `SqlNode`. ~~This approach allows it to access Calcite's standard validation flow and leverage its built-in type checking. The key implementation is in [PigRelToSqlConverter](https://calcite.apache.org/javadocAggregate/org/apache/calcite/piglet/PigRelToSqlConverter.html).~~

~~I propose we adopt a similar approach by converting PPL to `SqlNode`.~~ However, we could still try doing so. This should eliminate many future issues by reusing more of Calcite's functionality.

**My concern over this proposal**

With this approach, we **re-introduce** the `SqlNode` layer. This contradicts our previous design, which skips the layer.

- Calcite's workflow
  ```mermaid
  flowchart LR
  SQL -->|parse| SqlNode1[SqlNode] -->|validation| SqlNode2[SqlNode] -->|convert| RelNode1[RelNode] -->|optimize| RelNode2[RelNode]
  ```
- Our existing design
  ```mermaid
  flowchart LR
  PPL -->|parse| AST -->|convert| RelNode1[RelNode] -->|optimize| RelNode2[RelNode]
  ```
- With this proposal (the blocks in blue are added phases)
  ```mermaid
  flowchart LR
  PPL -->|parse| AST -->|convert| RelNode1[RelNode] -->|convert| SqlNode1[SqlNode]:::blue -->|validate| SqlNode2[SqlNode]:::blue -->|convert| RelNode2[RelNode]:::blue -->|optimize| RelNode3[RelNode]
  
  classDef blue fill:#2196F3,stroke:#0D47A1,color:white
  ```

It's like converting PPL to SQL for validation purpose, making people doubting why is the SqlNode phase skipped at the first place.

### Approach

This solution requires the following steps:

- [ ]  Convert PPL's logical plan to SqlNode by creating a similar `PplRelToSqlConverter`
    - [ ]  The handling of UDT needs more investigation
- [ ]  Refactor the current argument type validation with the converted SQL AST
- [ ]  Implement function type coercion based on this

**Long-Term Goals**
* Automatic coercion makes defining functions simpler and easier to maintain.
* Make use of more Calcite's validation mechanism to reduce maintenance effort and make the implementation more robust.

**Alternative**

```mermaid
flowchart LR
PPL -->|parse| AST -->|convert & cast| RelNode1[RelNode]:::blue -->|optimize| RelNode3[RelNode]

classDef blue fill:#2196F3,stroke:#0D47A1,color:white
```

- Make coercion directly at the `RelNode` level
  - Merits: Aligns with our existing design
  - Potential drawbacks: We make less use of what's already there in Calcite. It may leads to more investment in validation in the future. For example, we could have eliminated the effort to implement type checkers if we introduced the `SqlNode` layer.

**Implementation Discussion**

### LINKED ISSUE #4636: [FEATURE] Type Conversion - Leverage RelDataType from Calcite in V3 engine.
Currently. [V3 engine Type Coercion depend on ExprCoreType used in V2](https://github.com/opensearch-project/sql/pull/3914#discussion_r2244919778), and [the Type Coerciion logic does not support all TypeChecker.](https://github.com/opensearch-project/sql/pull/4599/files#diff-d9124925480da5626d6d9ff1243913234ffa4fd640411f439008fadb55b7935f) 

Leverage Calcite RelDataType and TypeCoercionImpl logic.

Customize TypeCoercion logic for RexNode.

n/a

### LINKED ISSUE #5175: [BUG] COALESCE(null, 42) returns wrong type string instead of int due to leastRestrictive type inference
## Query Information
**PPL Command/Query:**
```ppl
```
**Expected Result:**

**Actual Result:**

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
    }
  }
}
```
**Sample Data**
```json
{ "dummy": "row" }
```

## Bug Description
**Issue Summary:**

**Steps to Reproduce:**
2. Run both queries above.

**Impact:**

yuancu:
### Description

**Please review the sub-PRs instead**: https://github.com/opensearch-project/sql/pull/4892#issuecomment-4020754059

This PR migrates from our custom type checking mechanism to **Apache Calcite's native validation system** by introducing a SqlNode validation layer. This addresses the lack of a proper SQL validation phase and resolves compatibility issues with user-defined types (UDTs).

This implements **Approach 1** described in #3865.

### How It Works

The PR introduces a validation layer in the query execution pipeline:

```
RelNode → SqlNode → [Validation & Type Coercion] → SqlNode → RelNode
```

This approach leverages Calcite's robust type system while preserving OpenSearch's custom type semantics through careful type mapping and restoration.

### Benefits

- **Comprehensive type checking**: Access to Calcite's full type checker suite (variable-length operands, same-operand constraints, OR-relationship type checkers) vs. our limited family-based checking
- **Function optimization**: Ability to rewrite functions at the SqlNode level (e.g., `sqrt(x)` → `pow(x, 0.5)`) before generating physical plans
- **Automatic type coercion**: Implicit casting where type-safe and appropriate, improving query compatibility
- **Reduced maintenance**: Leverage Calcite's battle-tested validation logic instead of maintaining custom implementations

#### Core Infrastructure
- [x] Introduced SqlNode validation layer with round-trip conversion (RelNode ↔ SqlNode)
- [x] Extended Calcite validator to support OpenSearch UDTs via type substitution
- [x] Implemented custom `deriveType`, `coerceOperandType`, and type inference methods
- [x] Made validator thread-safe with proper synchronization

#### Type System Enhancements
- [x] **UDT Support**: To support OpenSearch-specific UDTs (which Calcite doesn't natively support), we bridge Calcite's `SqlTypeName` enum with OpenSearch UDTs through dynamic type mapping:
  1. Convert UDTs to standard SQL types (e.g., `EXPR_TIME` → `SqlTypeName.TIME`) before validation
  2. Apply Calcite's validation and type coercion
  3. Restore UDTs after validation to maintain plan correctness
- [x] **Cross-type Comparisons**: Override `commonTypeForBinaryComparison` to enable datetime cross-type comparisons (DATE vs TIME, etc.)
- [x] **IP Type Handling**: Emulated IP type using standard SQL types for validation, with special handling in operators
- [x] **Composite Types**: Extended SQL-UDT conversion to handle nested/composite types
- [x] **Type Coercion**: Use `SAFE_CAST` for string-to-number conversions to tolerate malformatted data; use `CAST` for literal numbers

#### Function & Operator Handling
- [x] Defined operand type checkers for all built-in functions:
  - Array functions (`array_slice`, `reduce`, `mvappend`, etc.) with proper type inference
  - JSON functions (`json_extract`, `json_set`, etc.)
  - Mathematical functions (corrected `atan` overloading, percentile approximations)
  - Aggregation functions (`DISTINCT_COUNT_APPROX`, `COUNT(*)` rewriting, etc.)
- [x] Function overloading support:
  - `ADD` operator: String concatenation vs numeric addition
  - `ATAN`: Single-operand vs two-operand versions
  - `GEOIP`: String overrides due to UDT erasure
- [x] Arithmetic operations between strings and numerics (implicit coercion)
- [x] Define `SqlKind` for `DIVIDE` and `MOD` UDFs

#### Query Construct Support
- [x] Preserve null ordering and collation through SqlNode round-trips
- [x] Support SEMI and ANTI joins in SQL conversion
- [x] Preserve sort orders in subqueries
- [x] Pass `RelHint` through conversions (added `SqlHint` for `LogicalAggregate`)
- [x] Handle windowed aggregates and `bucket_nullable` flags
- [x] Support `IN`/`NOT IN` with tuple inputs via row constructor rewriting

#### Dialect & Compatibility
- [x] Extended OpenSearch Spark SQL dialect for custom syntax
- [x] Fixed interval semantics mismatch between SQL and PPL
- [x] Fixed quarter interval bug in Calcite
- [x] Handle identifier expansion in aggregate functions
- [x] Properly handle `LogicalValues` for empty row generation

#### Edge Cases & Fixes
- [x] Skip validation for unsupported patterns:
  - Bin-on-timestamp operations (not yet implemented)
  - Group-by window functions (return original plan gracefully)
  - Specific aggregation patterns with LogicalValues
- [x] Fixed nullability attribute preservation through `SAFE_CAST`
- [x] Trim unused fields after SQL→Rel conversion
- [x] Ensure `RelToSqlConverter` is instantiated per-use (stateful component)
- [x] Fixed float literal handling with explicit casts
- [x] Remove JSON_TYPE operator insertion where inappropriate

#### Test Fixes
- [x] Fixed integration tests across multiple suites:
  - `CalcitePPLDateTimeBuiltinFunctionIT`: Interval semantics
  - `CalcitePPLBuiltinFunctionIT`: LOG function, sarg deserialization
  - `CalciteArrayFunctionIT`: Type checkers, reduce function inference
  - `CalciteMathematicalFunctionIT`, `CalcitePPLAggregationIT`
  - `CalciteBinCommandIT`: Timestamp operations, windowed aggregates in GROUP BY
  - `CalciteStreamstatsCommandIT`: Sort columns, bucket_nullable
  - `CalcitePPLJsonBuiltinFunctionIT`: String conversion
- [x] Updated all explain integration tests for plan changes
- [x] Fixed YAML tests, doctests, and unit tests
- [x] Updated ClickBench query plans

#### Code Cleanup
- [x] Removed legacy type checking from `PPLFuncImpTable`
- [x] Deprecated `UDFOperandMetadata.wrapUDT` interface
- [x] Removed unused classes: `CalciteFuncSignature`, `PPLTypeChecker`
- [x] Consolidated `EnhancedCoalesce` into built-in `Coalesce`
- [x] Removed type checkers from operator registration (now handled by Calcite)

#### Optimizations
- [x] Eliminated `SAFE_CAST` on non-string literal numbers (use `CAST` for better performance)
- [x] Investigated and addressed `dedup` optimization issues

### Performance Impact

  | Dataset    | Scenario           | Analyze (ms)   | Optimize (ms)  | Execute (ms)   | Total Avg (ms) |
  |------------|--------------------|----------------|----------------|----------------|----------------|
  | clickbench | Without Validation | 7.93           | 8.29           | 0.95           | 17.53          |
  |            | With Validation    | 9.59           | 8.72           | 1.00           | 19.69          |
  |            | Difference         | +1.66 (+20.9%) | +0.43 (+5.2%)  | +0.05 (+5.3%)  | +2.16 (+12.3%) |
  | big5       | Without Validation | 3.52           | 4.27           | 0.80           | 8.99           |
  |            | With Validation    | 3.37           | 3.80           | 0.72           | 8.22           |
  |            | Difference         | -0.15 (-4.3%)  | -0.47 (-11.0%) | -0.08 (-10.0%) | -0.77 (-8.6%)  |
  | tpch       | Without Validation | 8.65           | 692.00         | 183.15         | 1028.80        |
  |            | With Validation    | 7.91           | 689.16         | 175.63         | 1008.77        |
  |            | Difference         | -0.74 (-8.6%)  | -2.84 (-0.4%)  | -7.52 (-4.1%)  | -20.03 (-1.9%) |

**Profiled on personal laptop, each test runs twice, then averaged.**

**Conclusion**: No significant performance degradation. ClickBench shows slight overhead (+12.3%) during analyze phase due to validation, but big5 and TPCH show improvements, likely from better query optimization enabled by proper type information.

### Related Issues
Resolves #4636, resolves #3865, resolves #5175

### Check List


core/src/main/java/org/opensearch/sql/expression/function/udf/math/DivideFunction.java
```diff
@@ -41,6 +42,11 @@ public DivideFunction() {
     super(new DivideImplementor(), NullPolicy.ANY);
   }
 
+  @Override
+  public SqlKind getKind() {
+    return SqlKind.DIVIDE;
+  }
```
yuancu, line 49: This is to enable operator-dependent type derivation.

core/src/main/java/org/opensearch/sql/expression/function/CalciteFuncSignature.java
```diff
@@ -7,16 +7,13 @@
...
 import java.util.List;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.sql.type.SqlOperandTypeChecker;
 
 /** Function signature is composed by function name and arguments list. */
-public record CalciteFuncSignature(FunctionName functionName, PPLTypeChecker typeChecker) {
+public record CalciteFuncSignature(FunctionName functionName, SqlOperandTypeChecker typeChecker) {
 
+  // TODO: Refactor this match method
   public boolean match(FunctionName functionName, List<RelDataType> argTypes) {
```
penghuo: Does match() method is used?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -363,15 +357,15 @@ default RexNode resolve(RexBuilder builder, RexNode... args) {
    * implementations are independent of any specific data storage, should be registered here
    * internally.
    */
-  private final ImmutableMap<BuiltinFunctionName, List<Pair<CalciteFuncSignature, FunctionImp>>>
+  private final ImmutableMap<BuiltinFunctionName, Pair<CalciteFuncSignature, FunctionImp>>
```
penghuo: Why is a FunctionRegistry required? It is needed because CalciteRelNodeVisitor must resolve functions during analysis and translation?

core/src/main/java/org/opensearch/sql/calcite/validate/ValidationUtils.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.sql.SqlCollation;
+import org.apache.calcite.sql.type.NonNullableAccessors;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.type.SqlTypeUtil;
+import org.opensearch.sql.calcite.utils.OpenSearchTypeFactory;
+
+@UtilityClass
+public class ValidationUtils {
```
penghuo, line 20: Could add comments/javadoc if some function is copy from Calcite?

core/src/main/java/org/opensearch/sql/calcite/validate/converters/OpenSearchRelToSqlConverter.java
```diff
@@ -0,0 +1,172 @@
...
+   *
+   * <p>The override implementation uses ANTI / SEMI join syntax:
+   *
+   * <pre>{@code
+   * SELECT ... FROM table1 AS t0
+   * LEFT ANTI JOIN table2 AS t2 ON t0.name = t2.name
+   * }</pre>
+   */
+  @Override
+  protected Result visitAntiOrSemiJoin(Join e) {
```
penghuo, line 99: Add qualifier can also solve the issue? e.g. t0.name. Then we can avoid the ANTI JOIN visit in PplSqlToRelConverter?

penghuo, line 99: Add qualifier can also solve the issue? e.g. t0.name. Then we can avoid the ANTI JOIN visit in PplSqlToRelConverter?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
yuancu: it's still used when building logical plans from PPL's AST. yes, it's used in CalviteRelNodeVisitor. Now it's only name-based resolve

core/src/main/java/org/opensearch/sql/calcite/validate/converters/OpenSearchRelToSqlConverter.java
yuancu: Yes, adding a qualifier does solve the problem.

#### RCA

I investigated this query: `source = opensearch-sql_test_index_state_country | where country = 'Canada' OR country = 'England' | left semi join left=a, right=b ON a.name = b.name opensearch-sql_test_index_occupation | sort a.age`. It's original logical plan from CalciteRelNodeVisitor is:

```
 LogicalSort(sort0=[$5], dir0=[ASC-nulls-first])
  LogicalJoin(condition=[=($0, $6)], joinType=[semi])
    LogicalProject(name=[$0], country=[$1], state=[$2], month=[$3], year=[$4], age=[$5])
      LogicalFilter(condition=[SEARCH($1, Sarg['Canada':VARCHAR, 'England':VARCHAR]:VARCHAR)])
        CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_state_country]])
    LogicalProject(name=[$0], country=[$1], occupation=[$2], month=[$3], salary=[$4], year=[$5])
      CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_occupation]])
```

Its SQL is however

```sql
SELECT `name`, `country`, `state`, `month`, `year`, `age`
FROM `OpenSearch`.`opensearch-sql_test_index_state_country` AS `t0`
WHERE `country` IN ('Canada', 'England') AND EXISTS (SELECT 1
FROM (SELECT `name`, `country`, `occupation`, `month`, `salary`, `year`
FROM `OpenSearch`.`opensearch-sql_test_index_occupation`) AS `t1`
WHERE `name` = `t1`.`name`)  -- left `name` is unqualified
ORDER BY `age`
```

*Why is the first `name` unqualified*

In [RelToSqlConverter](https://github.com/apache/calcite/blob/5c6456d1f783d057f9027f32aa6c37d664a2a598/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java), the condition is [initially](https://github.com/apache/calcite/blob/5c6456d1f783d057f9027f32aa6c37d664a2a598/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java#L290-L291) created with qualifier `WHERE t0.name = t1.name`:
```java
SqlNode sqlCondition =
    convertConditionToSqlNode(e.getCondition(), leftContext, rightContext);
```
The following lines removes the qualifier because it's attempting to create a correlated subquery, and in standard SQL, outer query references in correlated subqueries are typically unqualified.

```java
if (leftResult.neededAlias != null) {
  SqlShuttle visitor =
      new AliasReplacementShuttle(leftResult.neededAlias,
          e.getLeft().getRowType(), sqlSelect.getSelectList());
  sqlCondition = sqlCondition.accept(visitor);
}
```

The visitor replaces the identifiers whose first qualifier equals to `leftResult.neededAlias`:
```java
@Override public SqlNode visit(SqlIdentifier id) {
...
  if (tableAlias.equals(id.names.get(0))) {
    int index =
        requireNonNull(tableType.getField(id.names.get(1), false, false),
            () -> "field " + id.names.get(1) + " is not found in "
                + tableType)
        .getIndex();
    SqlNode selectItem = source.get(index);
    if (selectItem.getKind() == SqlKind.AS) {
      selectItem = ((SqlCall) selectItem).operand(0);
    }
    return selectItem.clone(id.getParserPosition());
...
}
```

This replacement shuttle is there because the alias of the outer query `t0` does not exist under some conditions. However,  SQL's name resolution rules resolve the unqualified name to the innermost scope first, and there happen to exist such name in the innermost scope, so the name resolution is falsely resolved to `t1.name == t1.name` in the end.

#### Resolution

It can be resolved by either removing the replacement shuttle or enabling ANTI / SEMI join syntax. But the former solution risks encountering an non-existing `t0` alias if [the following condition](https://github.com/apache/calcite/blob/5c6456d1f783d057f9027f32aa6c37d664a2a598/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java#L329-L333) is not met:

```java
    if (leftResult.neededAlias != null && sqlSelect.getFrom() != null) {
      sqlSelect.setFrom(as(sqlSelect.getFrom(), leftResult.neededAlias));
    }
```
(**Note**: the above condition is updated by [CALCITE-3810: Render ANTI and SEMI join to NOT EXISTS and EXISTS in the JDBC adapter](https://issues.apache.org/jira/browse/CALCITE-3810). The latest update may have resolved this problem, but the update is not included in Calcite 1.41)

I think adopting the second solution, enabling ANTI/SEMI join, is better:
- It's safer because we will not run into trouble when the above condition is not met.
- Spark SQL [supports](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-join.html) ANTI and SEMI JOIN
- As we are creating ANTI / SEMI JOIN in CalciteRelNodeVisitor already, it will be more consistent if the logical plan translated back also contains them.

  https://github.com/opensearch-project/sql/blob/d0195cb5109f787e29b32f5505a1ab0b0f01dc45/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java#L1288-L1293

core/src/main/java/org/opensearch/sql/expression/function/CalciteFuncSignature.java
yuancu: No, it's no longer used. I removed it.

yuancu:
@coderabbitai generate docstrings

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
```diff
@@ -101,6 +121,47 @@ private CalcitePlanContext(CalcitePlanContext parent) {
...
+   * @return cached SqlValidator instance
+   */
+  public SqlValidator getValidator() {
+    if (validator == null) {
+      final CalciteServerStatement statement;
+      try {
+        statement = connection.createStatement().unwrap(CalciteServerStatement.class);
+      } catch (SQLException e) {
+        throw new RuntimeException(e);
+      }
```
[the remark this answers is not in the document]
yuancu: `CalciteServerStatement` does not implement `AutoCloseable` (although the statement created with `connection.createStatement()` does). Does this mean that I do not have to close it?

core/src/main/java/org/opensearch/sql/calcite/validate/ValidationUtils.java
yuancu: Yes, I annotated every method I copied from Calcite's implementation

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEnhancedCoalesceTest.java
```diff
@@ -42,13 +42,12 @@ public void testCoalesceWithMixedTypes() {
...
         "LogicalSort(fetch=[3])\n"
-            + "  LogicalProject(EMPNO=[$0], COMM=[$6], result=[COALESCE($6, $0,"
-            + " 'fallback':VARCHAR)])\n"
+            + "  LogicalProject(EMPNO=[$0], COMM=[$6], result=[COALESCE($6, $0)])\n"
             + "    LogicalTableScan(table=[[scott, EMP]])\n";
     verifyLogical(root, expectedLogical);
 
     String expectedSparkSql =
-        "SELECT `EMPNO`, `COMM`, COALESCE(`COMM`, `EMPNO`, 'fallback') `result`\n"
+        "SELECT `EMPNO`, `COMM`, COALESCE(`COMM`, `EMPNO`) `result`\n"
```
yuancu, line 50: to check: is this optimized because `EMPNO` is nonnull?

the rest optimization seems reasonable.

penghuo (approved):
Thanks!

core/src/main/java/org/opensearch/sql/calcite/validate/ValidationUtils.java
```diff
@@ -0,0 +1,165 @@
...
+   *   ORDER BY 2
+   * </pre>
+   *
+   * <p><b>TODO:</b> Remove this workaround when upgrading to a Calcite version that fixes the bugs.
+   *
+   * @param e the exception to check
+   * @return {@code true} if the exception should be tolerated as a known Calcite bug, {@code false}
+   *     otherwise
+   */
+  public static boolean tolerantValidationException(Exception e) {
```
qianheng-aws, line 157: How many cases will hit this? It will use the original plan if it's tolerant, right?

core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -271,6 +291,66 @@ public LogicalPlan analyze(UnresolvedPlan plan, QueryType queryType) {
...
+   * PPL queries.
+   *
+   * @param relNode the relation node to validate
+   * @param context the Calcite plan context containing the validator
+   * @return the validated (and potentially modified) relation node
+   */
+  private RelNode validate(RelNode relNode, CalcitePlanContext context) {
+    SkipRelValidationShuttle skipShuttle = new SkipRelValidationShuttle();
+    relNode.accept(skipShuttle);
+    if (skipShuttle.shouldSkipValidation()) {
```
qianheng-aws, line 318: Is this a limitation? How many cases in our IT will hit this?

core/src/main/java/org/opensearch/sql/calcite/validate/converters/PplSqlToRelConverter.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.sql.JoinType;
+import org.apache.calcite.sql.SqlJoin;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.sql2rel.SqlRexConvertletTable;
+import org.apache.calcite.sql2rel.SqlToRelConverter;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class PplSqlToRelConverter extends SqlToRelConverter {
```
qianheng-aws: Since we've extended Calcite's SqlToRelConverter, we have chance to customized our plan by using hints.

e.g. When converting RelNode to SqlNode, we convert LogicalDedup to SqlSelect with `dedup` hints. And later when converting SqlNode to RelNode, we can detect that hint and add LogicalDedup on the original LogicalProject.

qianheng-aws:
This PR validate the whole SQL including type compatible, SQL grammer, field access and others. However what we need is only type compatibility since the relbuilder will not build an incorrect plan.

I'm thinking if we can only convert recall to sqlnode and invoke validateCall to do that part of validation(should include type coercion), and then convert it back to recall.

core/src/main/java/org/opensearch/sql/calcite/validate/ValidationUtils.java
yuancu: Roughly 5 cases. Yes, it will use the original plan if hit

core/src/main/java/org/opensearch/sql/executor/QueryService.java
yuancu: Yes, they are limitations, either from our implementation or calite's. Around a dozen ITs will hit this.

core/src/main/java/org/opensearch/sql/calcite/validate/converters/PplSqlToRelConverter.java
yuancu: Yes, I was also considering this approach. I hesitated because I was not sure which sql operation conforms with dedup's attributes. SqlSelect should to be a reasonable one.

SQL hints are already used to the empty bucket hints in the current implementation.

yuancu:
> I'm thinking if we can only convert recall to sqlnode and invoke validateCall to do that part of validation(should include type coercion), and then convert it back to recall.

It's kind of tricky, the validation of these are deeply entangled. E.g. if a function call is at a very bottom level of the AST, the validator has to go down the AST and validate all the nodes it comes across to build a proper validation scope for the function call. If we want to try to build them for these function calls, we may need to replicate those steps, which does not differ much from the original validation.

yuancu:
> Curious how much overhead will this validation process introduce?

I have updated a profiling result in the PR description. It's profiled with [this script](https://gist.github.com/yuancu/e722d937059c2aec5c1a04fa34ec8d59). Based on the result, the validation layer brings no significant performance degration.

yuancu:
Hi @ykmr1224 , do you have any concern over this PR? I foresee that it will bring some impact to the schema-on-read solution, be it in a helpful or negative way. I'm worrying that this PR will make your solution difficult to proceed.

qianheng-aws (dismissed):

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
```diff
@@ -107,6 +127,52 @@ private CalcitePlanContext(CalcitePlanContext parent) {
...
+  /**
+   * Gets the SqlValidator instance (singleton per CalcitePlanContext).
+   *
+   * @return cached SqlValidator instance
+   */
+  public SqlValidator getValidator() {
+    if (validator == null) {
+      synchronized (this) {
+        //  Double-Checked Locking for thread-safety
+        if (validator == null) {
```
LantaoJin: why thread problem could be happened here? the `CalcitePlanContext` is created per query. Do you want create a `SqlValidator` per query either? Or share a `SqlValidator` in diff queries.

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
```diff
@@ -78,6 +87,17 @@ public class CalcitePlanContext {
...
+  /**
+   * -- SETTER -- Sets the SQL operator table provider. This must be called during initialization by
+   * the opensearch module.
+   *
+   * @param provider the provider to use for obtaining operator tables
+   */
+  @Setter private static SqlOperatorTableProvider operatorTableProvider;
+
+  /** Cached SqlValidator instance (lazy initialized). */
+  private volatile SqlValidator validator;
```
LantaoJin: cache for what? do you mean some queries could be validated-free? Since `CalcitePlanContext` is create per query, if a query has to be validated no matter what it is, lazy initialization is not needed.

core/src/main/java/org/opensearch/sql/calcite/validate/PplValidator.java
```diff
@@ -0,0 +1,205 @@
...
+      RelDataTypeFactory typeFactory,
+      SqlValidator.Config validatorConfig) {
+    SchemaPlus defaultSchema = frameworkConfig.getDefaultSchema();
+
+    final CalcitePrepare.Context prepareContext = statement.createPrepareContext();
+    final CalciteSchema schema =
+        defaultSchema != null ? CalciteSchema.from(defaultSchema) : prepareContext.getRootSchema();
+    CalciteCatalogReader catalogReader =
+        new CalciteCatalogReader(
+            schema.root(), schema.path(null), typeFactory, prepareContext.config());
```
LantaoJin: if we use `CalciteConnectionConfig.DEFAULT` as the last parameter of `CalciteCatalogReader()`, is it possible to remove the `CalciteServerStatement` in `create`'s parameter.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1769,6 +1777,84 @@ public RelNode visitStreamWindow(StreamWindow node, CalcitePlanContext context)
...
+   *
+   * <pre>
+   * LogicalProject(SUM($5) OVER (ORDER BY $5 ROWS 1 PRECEDING))  ← ORDER BY embedded
+   * </pre>
+   *
+   * @param overExpressions Window function expressions (may contain nested {@link RexOver})
+   * @param context Plan context for building RexNodes
+   * @return Expressions with ORDER BY embedded in all window specifications
+   */
+  private List<RexNode> embedExistingCollationsIntoOver(
```
LantaoJin, line 2148: Can this changes fix the issue https://github.com/opensearch-project/sql/issues/5004?

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -330,82 +329,66 @@ public Type getJavaClass(RelDataType type) {
...
+        } else if (OpenSearchTypeUtil.isTime(t)) {
+          timeCount++;
+        } else if (OpenSearchTypeUtil.isIp(t)) {
+          ipCount++;
+        } else if (OpenSearchTypeUtil.isBinary(t)) {
+          binaryCount++;
+        }
+      }
+      if (nullCount == 0 && anyCount == 0) {
+        RelDataType udt;
```
LantaoJin: why return udt only when `nullCount == 0 && anyCount == 0`?

core/src/main/java/org/opensearch/sql/expression/function/UserDefinedFunctionBuilder.java
```diff
@@ -12,8 +12,10 @@
 import org.apache.calcite.sql.SqlOperandCountRange;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.InferTypes;
+import org.apache.calcite.sql.type.OperandTypes;
 import org.apache.calcite.sql.type.SqlReturnTypeInference;
 import org.apache.calcite.sql.validate.SqlUserDefinedFunction;
+import org.jspecify.annotations.NonNull;
```
yuancu, line 18: Why using `org.jspecify.annotations.NonNull`: `lombok.NonNull` is not for annotating return value; Calcite adopts `org.checkerframework.checker.nullness.qual.NonNull`, which is heavier than `org.jspecify.annotations.NonNull`.

core/src/main/java/org/opensearch/sql/expression/function/udf/math/EulerFunction.java
```diff
@@ -39,7 +40,7 @@ public SqlReturnTypeInference getReturnTypeInference() {
   }
 
   @Override
-  public UDFOperandMetadata getOperandMetadata() {
+  public @NonNull UDFOperandMetadata getOperandMetadata() {
     return PPLOperandTypes.NONE;
   }
```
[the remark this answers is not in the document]
yuancu: It can inherit the interface's javadoc

core/src/main/java/org/opensearch/sql/calcite/validate/OpenSearchSparkSqlDialect.java
```diff
@@ -68,4 +90,24 @@ private void unparseFunction(
...
     writer.endList(frame);
   }
+
+  @Override
+  public SqlConformance getConformance() {
+    return new ConformanceDelegate(super.getConformance());
+  }
+
+  /** SqlConformance delegator that enables liberal mode for LEFT SEMI/ANTI JOIN support. */
+  private static class ConformanceDelegate implements SqlConformance {
```
LantaoJin: why not extends `SqlDelegatingConformance`? If a caller calls one method of `ConformanceDelegate` which is not override, will it throw error?

core/src/main/java/org/opensearch/sql/calcite/validate/OpenSearchSparkSqlDialect.java
```diff
@@ -68,4 +90,24 @@ private void unparseFunction(
...
+    @Delegate private final SqlConformance delegate;
+
+    ConformanceDelegate(SqlConformance delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override
+    public boolean isLiberal() {
+      // This allows SQL feature LEFT ANTI JOIN & LEFT SEMI JOIN
```
LantaoJin: can you explain more here? I didn't get the point why it must return `true`

core/src/main/java/org/opensearch/sql/calcite/validate/PplTypeCoercion.java
```diff
@@ -0,0 +1,211 @@
...
+public class PplTypeCoercion extends TypeCoercionImpl {
+
+  /**
+   * Creates a custom TypeCoercion instance for PPL. This can be used as a TypeCoercionFactory.
+   *
+   * @param typeFactory the type factory
+   * @param validator the SQL validator
+   * @return custom PplTypeCoercion instance
+   */
+  public static TypeCoercion create(RelDataTypeFactory typeFactory, SqlValidator validator) {
```
LantaoJin, line 56: you added lots of static method in folder `validate`. IMO, it required sufficient unit tests for those static methods, at least for all new added files. Can you add more unit test to increase the coverage with the help of AI tool?

integ-test/src/test/resources/expectedOutput/calcite/big5/range_numeric.yaml
```diff
@@ -1,9 +1,10 @@
+
 calcite:
   logical: |
     LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
-      LogicalProject(agent=[$0], process=[$6], log=[$8], message=[$11], tags=[$12], cloud=[$13], input=[$15], @timestamp=[$17], ecs=[$18], data_stream=[$20], meta=[$24], host=[$26], metrics=[$27], aws=[$30], event=[$35])
-        LogicalSort(fetch=[10])
-          LogicalFilter(condition=[SEARCH($28, Sarg[[20..200]])])
+      LogicalSort(fetch=[10])
+        LogicalProject(agent=[$0], process=[$6], log=[$8], message=[$11], tags=[$12], cloud=[$13], input=[$15], @timestamp=[$17], ecs=[$18], data_stream=[$20], meta=[$24], host=[$26], metrics=[$27], aws=[$30], event=[$35])
+          LogicalFilter(condition=[AND(>=($28, 20), <=($28, 200))])
```
LantaoJin, line 7: I just curious that how to generate `SEARCH` expression from SqlNode

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q2.yaml
```diff
@@ -2,7 +2,7 @@ calcite:
...
...,count()=COUNT()), LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","query":{"script":{"script":{"source":"{\"langType\":\"calcite\",\"script\":\"rO0ABXQCSnsKICAib3AiOiB7CiAgICAibmFtZSI6ICI8PiIsCiAgICAia2luZCI6ICJOT1RfRVFVQUxTIiwKICAgICJzeW50YXgiOiAiQklOQVJZIgogIH0sCiAgIm9wZXJhbmRzIjogWwogICAgewogICAgICAib3AiOiB7CiAgICAgICAgIm5hbWUiOiAiU0FGRV9DQVNUIiwKICAgICAgICAia2luZCI6ICJTQUZFX0NBU1QiLAogICAgICAgICJzeW50YXgiOiAiU1BFQ0lBTCIKICAgICAgfSwKICAgICAgIm9wZXJhbmRzIjogWwogICAgICAgIHsKICAgICAgICAgICJkeW5hbWljUGFyYW0iOiAwLAogICAgICAgICAgInR5cGUiOiB7CiAgICAgICAgICAgICJ0eXBlIjogIlNNQUxMSU5UIiwKICAgICAgICAgICAgIm51bGxhYmxlIjogdHJ1ZQogICAgICAgICAgfQogICAgICAgIH0KICAgICAgXSwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiSU5URUdFUiIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZQogICAgICB9CiAgICB9LAogICAgewogICAgICAiZHluYW1pY1BhcmFtIjogMSwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiQklHSU5UIiwKICAgICAgICAibnVsbGFibGUiOiB0cnVlCiAgICAgIH0KICAgIH0KICBdCn0=\"}","lang":"opensearch_compounded_script","params":{"utcTimestamp": 0,"SOURCES":[0,2],"DIGESTS":["AdvEngineID",0]}},"boost":1.0}},"track_total_hits":2147483647}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```
LantaoJin: Seems an unexpected script was introduced here

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q37.yaml
```diff
@@ -6,7 +6,8 @@ calcite:
...
...AgIm9wIjogewogICAgICAgICJuYW1lIjogIlNBRkVfQ0FTVCIsCiAgICAgICAgImtpbmQiOiAiU0FGRV9DQVNUIiwKICAgICAgICAic3ludGF4IjogIlNQRUNJQUwiCiAgICAgIH0sCiAgICAgICJvcGVyYW5kcyI6IFsKICAgICAgICB7CiAgICAgICAgICAiZHluYW1pY1BhcmFtIjogMCwKICAgICAgICAgICJ0eXBlIjogewogICAgICAgICAgICAidHlwZSI6ICJTTUFMTElOVCIsCiAgICAgICAgICAgICJudWxsYWJsZSI6IHRydWUKICAgICAgICAgIH0KICAgICAgICB9CiAgICAgIF0sCiAgICAgICJ0eXBlIjogewogICAgICAgICJ0eXBlIjogIklOVEVHRVIiLAogICAgICAgICJudWxsYWJsZSI6IHRydWUKICAgICAgfQogICAgfSwKICAgIHsKICAgICAgImR5bmFtaWNQYXJhbSI6IDEsCiAgICAgICJ0eXBlIjogewogICAgICAgICJ0eXBlIjogIkJJR0lOVCIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZQogICAgICB9CiAgICB9CiAgXQp9\"}","lang":"opensearch_compounded_script","params":{"utcTimestamp": 0,"SOURCES":[0,2],"DIGESTS":["IsRefresh",0]}},"boost":1.0}},{"bool":{"must":[{"exists":{"field":"URL","boost":1.0}}],"must_not":[{"term":{"URL":{"value":"","boost":1.0}}}],"adjust_pure_negative":true,"boost":1.0}}],"adjust_pure_negative":true,"boost":1.0}},"aggregations":{"URL":{"terms":{"field":"URL","size":10,"min_doc_count":1,"shard_min_doc_count":0,"show_term_doc_count_error":false,"order":[{"_count":"desc"},{"_key":"asc"}]}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```
LantaoJin: ditto

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q38.yaml
```diff
@@ -6,7 +6,8 @@ calcite:
...
...ogewogICAgICAgICJuYW1lIjogIlNBRkVfQ0FTVCIsCiAgICAgICAgImtpbmQiOiAiU0FGRV9DQVNUIiwKICAgICAgICAic3ludGF4IjogIlNQRUNJQUwiCiAgICAgIH0sCiAgICAgICJvcGVyYW5kcyI6IFsKICAgICAgICB7CiAgICAgICAgICAiZHluYW1pY1BhcmFtIjogMCwKICAgICAgICAgICJ0eXBlIjogewogICAgICAgICAgICAidHlwZSI6ICJTTUFMTElOVCIsCiAgICAgICAgICAgICJudWxsYWJsZSI6IHRydWUKICAgICAgICAgIH0KICAgICAgICB9CiAgICAgIF0sCiAgICAgICJ0eXBlIjogewogICAgICAgICJ0eXBlIjogIklOVEVHRVIiLAogICAgICAgICJudWxsYWJsZSI6IHRydWUKICAgICAgfQogICAgfSwKICAgIHsKICAgICAgImR5bmFtaWNQYXJhbSI6IDEsCiAgICAgICJ0eXBlIjogewogICAgICAgICJ0eXBlIjogIkJJR0lOVCIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZQogICAgICB9CiAgICB9CiAgXQp9\"}","lang":"opensearch_compounded_script","params":{"utcTimestamp": 0,"SOURCES":[0,2],"DIGESTS":["IsRefresh",0]}},"boost":1.0}},{"bool":{"must":[{"exists":{"field":"Title","boost":1.0}}],"must_not":[{"term":{"Title":{"value":"","boost":1.0}}}],"adjust_pure_negative":true,"boost":1.0}}],"adjust_pure_negative":true,"boost":1.0}},"aggregations":{"Title":{"terms":{"field":"Title","size":10,"min_doc_count":1,"shard_min_doc_count":0,"show_term_doc_count_error":false,"order":[{"_count":"desc"},{"_key":"asc"}]}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```
LantaoJin: ditto

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q39.yaml
```diff
@@ -6,9 +6,10 @@ calcite:
...
+                  LogicalFilter(condition=[AND(=($103, 62), >=($0, TIMESTAMP('2013-07-01 00:00:00')), <=($0, TIMESTAMP('2013-07-31 00:00:00')), =(SAFE_CAST($72), 0), <>(SAFE_CAST($49), 0), =(SAFE_CAST($35), 0))])
```
LantaoJin: ditto

core/src/main/java/org/opensearch/sql/calcite/validate/OpenSearchSparkSqlDialect.java
yuancu: To allow converted SQL adopt LEFT SEMI JOIN, I first extended RelToSqlConverter’s `visitAntiOrSemiJoin` method to convert rel join of type LEFT_SEMI_JOIN to sql join of LEFT_SEMI_JOIN.

> JoinRelType.LEFT_SEMI_JOIN -> JoinType.LEFT_SEMI_JOIN:

https://github.com/opensearch-project/sql/blob/7418cda6926b2dc73081cf9a34c5bcb4affc5659/core/src/main/java/org/opensearch/sql/calcite/validate/converters/OpenSearchRelToSqlConverter.java#L98-L119

Later when the sql join gets converted back to rel join, it will examine whether the join type is acceptable in [SqlValidatorImpl.java](https://github.com/apache/calcite/blob/077d355ec337aba4272b7adfbcc982629839883e/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java#L3904C1-L3910C8). That is when this conformance will be checked against.

> validate JoinType.LEFT_SEMI_JOIN:

```java
protected void validateJoin(SqlJoin join, SqlValidatorScope scope) {
    ...
    switch (joinType) {
    case LEFT_ANTI_JOIN:
    case LEFT_SEMI_JOIN:
      if (!this.config.conformance().isLiberal()) {
        throw newValidationError(join.getJoinTypeNode(),
            RESOURCE.dialectDoesNotSupportFeature(joinType.name()));
      }
```

Full modifications to support LEFT SEMI JOIN can be viewed in this [diff](https://github.com/opensearch-project/sql/pull/4892/changes/1aea9e377fb520381bb856888780e1f6be5e9b41#diff-21b0d43bd35c924c5f7278c1ebda0fcabb80b5013b7602859bca21713c2d126c).

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q2.yaml
yuancu: The script is introduced by `SAFE_CAST(AdvEngineID AS INTEGER)`.

I am thinking maybe this cast here does make sense. The PPL query is `source=hits | where AdvEngineID!=0 | stats count()`, where `AdvEngineID` is a `smallint`, and `0` is parsed as an `integer`. Therefore, `AdvEngineID` is promoted to their common type `integer`.

The cast seems to be unnecessary in 2 ways:
1. why use `SAFE_CAST` instead of `CAST`, which won't introduce scripts.
2. why cast `AdvEngineID` to integer instead of casting `0` to `smallint`

After some investigation, there are the reasons:
1. There is a PR that introduces SAFE_CAST so that malformatted data won't cause runtime errors but return null. It enables the following use case (especially the last column)

https://github.com/opensearch-project/sql/blob/42b89f01c5b578588ffde6c950190dace536d463/docs/user/ppl/functions/conversion.md#L131-L148

2. I can of course always cast the literal side, i.e. cast 0 to smallint in this case, but what if it is not `0` but Integer.MAX? Casting `AdvEngineID` to integer would be safer.

What's your opinion on how to fix it?

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -330,82 +329,70 @@ public Type getJavaClass(RelDataType type) {
...
+        } else if (binaryCount + nullCount == types.size()) {
+          udt = createUDT(ExprUDT.EXPR_BINARY, nullable);
+        }
+        // There exists a mix of time, date, and timestamp (and optionally null)
+        else if (binaryCount == 0 && ipCount == 0) {
+          udt = createUDT(ExprUDT.EXPR_TIMESTAMP, nullable);
+        } else {
+          udt = createSqlType(SqlTypeName.VARCHAR, nullable);
+        }
+        return udt;
```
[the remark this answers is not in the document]
yuancu: @coderabbitai nope, all null casts won't pass the initial condition `if (types.stream().anyMatch(OpenSearchTypeUtil::isUserDefinedType))`, thus won't fall into the date UDT case.

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q2.yaml
LantaoJin: > What's your opinion on how to fix it?

No problem to adding `SAFE_CAST(AdvEngineID AS INTEGER)` in the original logical plan. But we need to remove the cast in pushdown IMO.

Refer to SparkSQL
```
create table t1(col1 ShortType) using parquet;
explain extended select * from t1 where col1 = 0;
```

In its logical plan, we see the `cast(col1#5 as int)`
```
== Analyzed Logical Plan ==
col1: smallint
Project [col1#5]
+- Filter (cast(col1#5 as int) = 0)
   +- SubqueryAlias spark_catalog.default.t1
      +- Relation spark_catalog.default.t1[col1#5] parquet
```

But the cast removed in optimized logical plan
```
== Optimized Logical Plan ==
Filter (isnotnull(col1#5) AND (col1#5 = 0))
+- Relation spark_catalog.default.t1[col1#5] parquet
```
and its physical `PushedFilters`
```
== Physical Plan ==
*(1) Filter (isnotnull(col1#5) AND (col1#5 = 0))
+- *(1) ColumnarToRow
   +- FileScan parquet spark_catalog.default.t1[col1#5] Batched: true, DataFilters: [isnotnull(col1#5), (col1#5 = 0)], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/Users/ltjin/Package/spark-3.5.4-bin-hadoop3/spark-warehouse/t1], PartitionFilters: [], PushedFilters: [IsNotNull(col1), EqualTo(col1,0)], ReadSchema: struct<col1:smallint>
```

LantaoJin: TODO: we need a [UnwrapCastInBinaryComparison](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/UnwrapCastInBinaryComparison.scala) like rule in optimizer

LantaoJin: @coderabbitai please open an issue about implementing a [UnwrapCastInBinaryComparison](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/UnwrapCastInBinaryComparison.scala) like rule in optimizer

LantaoJin: > After some investigation, there are the reasons:

> 1. There is a PR that introduces SAFE_CAST so that malformatted data won't cause runtime errors but return null. It enables the following use case (especially the last column)

@yuancu, back to your investigation, how a runtime error could be happened? `where a != 0` implicit that the `a` is not null. Since if the `a` is null, it should return `false` in database. So, we can optimize the `a!=0` to `isnotnull(a) AND NOT(a = 0)`, `a=0` to `isnotnull(a) AND a = 0`.

LantaoJin: Example on PostgreSQL:
```
-- create
CREATE TABLE EMPLOYEE (
  empId INTEGER PRIMARY KEY,
  name TEXT NOT NULL,
  dept TEXT NOT NULL,
  col1 SMALLINT
);

-- insert
INSERT INTO EMPLOYEE VALUES (0001, 'Clark', 'Sales', 1);
INSERT INTO EMPLOYEE VALUES (0002, 'Dave', 'Accounting', 0);
INSERT INTO EMPLOYEE VALUES (0003, 'Ava', 'Sales');

-- fetch 
SELECT * FROM EMPLOYEE WHERE col1 != 0;
```
returns 
```
 empid | name  | dept  | col1 
-------+-------+-------+------
     1 | Clark | Sales |    1
```
and
```
SELECT * FROM EMPLOYEE WHERE col1 = 0;
```
returns
```
 empid | name |    dept    | col1 
-------+------+------------+------
     2 | Dave | Accounting |    0
```

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
yuancu: I used to think a validator could be re-used by multiple queries. But if a new context is created for every query, there is no need to do so. Also, I later found out that `SqlValidator` is stateful and could not be shared. I have updated the code to remove it.

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
yuancu: Removed

core/src/main/java/org/opensearch/sql/calcite/validate/PplValidator.java
yuancu: Yes, it's possible. Updated the implementation to eliminate the dependency.

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
yuancu: I have updated the condition to `anyCount == 0` alone.

If there is `ANY` involved, the common type will be `ANY` anyway, I delegate this to calcite's implementation of `leastRestrictive`. Since `SqlTypeName.NULL` can be casted to any possible type (not `ANY` type) including UDTs, I include it now.

core/src/main/java/org/opensearch/sql/calcite/validate/OpenSearchSparkSqlDialect.java
yuancu: I did not discover the existence of `SqlDelegatingConformance`. Updated to use it.

Calling to non-override method would not throw error because it uses the `lomok.Delegate` decorator -- these calls will be delegated to the underlying delegate.

core/src/test/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactoryTest.java
```diff
@@ -0,0 +1,279 @@
...
+    // When IP UDT is mixed with OTHER type (which is used as intermediate for IP)
+    RelDataType ipUdt = TYPE_FACTORY.createUDT(ExprUDT.EXPR_IP);
+    RelDataType nullType = TYPE_FACTORY.createSqlType(SqlTypeName.NULL);
+
+    RelDataType result = TYPE_FACTORY.leastRestrictive(List.of(ipUdt, nullType));
+
+    assertNotNull(result);
+    assertTrue(OpenSearchTypeUtil.isIp(result));
+    assertTrue(result.isNullable());
+  }
```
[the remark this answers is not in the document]
yuancu: @coderabbitai I think it's fixed, can you check again?

core/src/test/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactoryTest.java
```diff
@@ -0,0 +1,279 @@
...
+    RelDataType dateUdt = TYPE_FACTORY.createUDT(ExprUDT.EXPR_DATE);
+    RelDataType nullType1 = TYPE_FACTORY.createSqlType(SqlTypeName.NULL);
+    RelDataType nullType2 = TYPE_FACTORY.createSqlType(SqlTypeName.NULL);
+
+    RelDataType result = TYPE_FACTORY.leastRestrictive(List.of(dateUdt, nullType1, nullType2));
+
+    assertNotNull(result);
+    assertTrue(OpenSearchTypeUtil.isDate(result));
+    assertTrue(result.isNullable());
+  }
```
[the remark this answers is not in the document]
yuancu: I added some, can you check again @coderabbitai

LantaoJin:
This PR is too large to review, is it possible to separate into multiple smaller ones?
In future, if we want to merge such a large change, I think we'd better to create a feature branch and review PRs targeted to feature branch, then we can merge the feature branch back to main without review.

yuancu:
## PR Split Plan

As suggested by @LantaoJin, this PR has been split into 4 sub-PRs targeting the `feature/validation` branch:

| Sub-PR | Title | Files | Dependency |
|--------|-------|-------|------------|
| #5213 | Add validation infrastructure and type system | 32 | Independent |
| #5214 | Define operand type checkers for all PPL built-in functions | 86 | Independent |
| #5215 | Enable validation pipeline and update all test expected outputs | 583 | After #5213 + #5214 |
| #5216 | Remove legacy type checking code and update documentation | 13 | After #5215 |

**Merge order**: #5213 and #5214 can be reviewed/merged in parallel → #5215 → #5216 → merge `feature/validation` → `main`

**Reviewer guide**:
- **#5213** (32 files): Core validation pipeline design — focus on round-trip correctness, graceful fallback, type coercion rules
- **#5214** (86 files): Function operand types — highly repetitive pattern, batch-reviewable
- **#5215** (583 files, ~546 YAML): If CI passes, the YAML updates are correct — focus on the ~37 source files
- **#5216** (13 files): Pure deletion + docs — quick review

yuancu:
> **Note**: This PR is superseded by 4 smaller sub-PRs targeting the `feature/validation` branch, as suggested in https://github.com/opensearch-project/sql/pull/4892#issuecomment-3857871514.

Please review the sub-PRs instead:

| Order | PR | Title | Files | Status |
|-------|-----|-------|-------|--------|
| 1 | #5213 | Add validation infrastructure and type system | 32 | Ready for review |
| 2 | #5214 | Define operand type checkers for all PPL built-in functions | 86 | Ready for review |
| 3 | #5215 | Enable validation pipeline and update all test expected outputs | 583 | Merge after #5213 + #5214 |
| 4 | #5216 | Remove legacy type checking code and update documentation | 13 | Merge after #5215 |

This PR will be closed once all sub-PRs are merged and `feature/validation` is merged back to `main`.

PR #4897, opened 2025-12-03
Fix parsing of dot-containing field names

### LINKED ISSUE #4896: [BUG] `ArrayIndexOutOfBoundsException` when querying index with disabled objects containing dot-only field names
## Bug Description

When querying an OpenSearch index that contains documents with field names consisting only of dots (e.g., `"."`, `".."`) **inside disabled objects** (`"enabled": false`), the SQL plugin throws:

```
ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
```

The root cause is in the constructor of `OpenSearchExprValueFactory.JsonPath`:

```java
public JsonPath(String rawPath) {
  this.paths = List.of(rawPath.split("\\."));
}
```

When `rawPath` is `"."`, the call `rawPath.split("\\.")` returns an empty array (`[]`), causing `paths.get(0)` to fail.

---

## Steps to Reproduce

1. **Create index with disabled object mapping**

```bash
curl -X PUT "localhost:9200/logs-disabled-test" -H 'Content-Type: application/json' -d '{
  "mappings": {
    "properties": {
      "log": {"type": "object", "enabled": false},
      "@timestamp": {"type": "date"}
    }
  }
}'
```

2. **Index a document with a dot-only field name**

```bash
curl -X POST "localhost:9200/logs-disabled-test/_doc" -H 'Content-Type: application/json' -d '{
  "@timestamp": "2025-11-26T17:10:00.000Z",
  "log": {".": "problematic value"}
}'
```

3. **Run a PPL query**

```bash
curl -X POST "localhost:9200/_plugins/_ppl" -H "Content-Type: application/json" -d '{
  "query": "source=logs-disabled-test"
}'
```

4. **Observed result**

OpenSearch returns HTTP 500 with an internal exception:

```json
{
  "error": {
    "details": "java.sql.SQLException: exception while executing query: Index 0 out of bounds for length 0",
    "type": "RuntimeException"
  },
  "status": 500
}
```

---

## Expected Behavior

The query should:

1. **Either** return results successfully, treating `"."` as a literal field name
2. **Or** return a clean, user-facing validation error

The plugin should not crash with an ArrayIndexOutOfBoundsException.

---

## Environment

* OS: Any
* OpenSearch Version: 2.19-dev
* SQL Plugin Version: 2.19-dev (Calcite engine enabled)

---

## Additional Context

This occurs because OpenSearch setting **allows arbitrary inner field names** for `"enabled": false` object fields. These documents are indexed without validation. When log pipelines generate malformed keys like `"."`, they pass through fine, but the SQL plugin fails when trying to parse the field path.

### Stack Trace (excerpt)

```
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
    at java.base/java.util.ImmutableCollections$ListN.get(ImmutableCollections.java:687)
    at org.opensearch.sql.opensearch.data.value.OpenSearchExprValueFactory$JsonPath.getRootPath(...)
    at org.opensearch.sql.opensearch.data.value.OpenSearchExprValueFactory.populateValueRecursive(...)
    ...
```

ahkcs:
<h2>Description</h2>
<p>Resolves #4896 — <code inline="">ArrayIndexOutOfBoundsException</code> when querying an index containing dot-only field names (e.g., <code inline="">"."</code>, <code inline="">".."</code>) inside disabled object fields.</p>
<h2>Root Cause</h2>
<p>In <code inline="">OpenSearchExprValueFactory.JsonPath</code>, the constructor uses:</p>
<pre><code class="language-java">rawPath.split("\\.")
</code></pre>
<p>For field names that consist only of dots (e.g., <code inline="">"."</code>, <code inline="">".."</code>), <code inline="">split("\\.")</code> returns an empty array <code inline="">[]</code>.<br>
Subsequent access to <code inline="">paths.get(0)</code> results in <code inline="">ArrayIndexOutOfBoundsException</code>.</p>
<p>This occurs frequently when an index contains disabled objects (<code inline="">"enabled": false</code>), because disabled objects bypass field name validation and allow invalid field names such as <code inline="">"."</code>.</p>
<h2>Fix</h2>
<p>If <code inline="">split("\\.")</code> returns an empty array, treat the entire raw path as a literal field name instead of a parsed path:</p>
<pre><code class="language-java">if (parts.length == 0) {
    this.paths = List.of(rawPath);   // Use "." as a literal field name
} else {
    this.paths = List.of(parts);
}
</code></pre>
<h2>Result</h2>

Input JSON | Before | After
-- | -- | --
{"log": {".": "value"}} | Crash (ArrayIndexOutOfBounds) | Successfully returns the field value

## Summary by CodeRabbit

* **Bug Fixes**
  * Field names containing dots (leading, trailing, middle, multiple, or malformed) are preserved and remain queryable instead of being split.

* **Tests**
  * Added extensive unit tests validating dot-containing field-name parsing and retrieval.
  * Added an integration test exercising end-to-end queries and behavior for various dot-containing field-name scenarios.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
+++ b/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
@@ -31,85 +31,33 @@ setup:
               message:
                 type: text
 
-  # Document 1: Single dot field name "."
+  # Bulk index all test documents
   - do:
-      index:
+      bulk:
         index: test_disabled_object_4896
-        id: 1
-        body:
-          "@timestamp": "2025-11-26T17:10:00.000Z"
-          message: "single dot"
-          log:
-            ".": "single dot value"
-
-  # Document 2: Multiple dots field name ".."
-  - do:
-      index:
-        index: test_disabled_object_4896
-        id: 2
-        body:
-          "@timestamp": "2025-11-26T17:11:00.000Z"
-          message: "multiple dots"
-          log:
-            "..": "double dot value"
-
-  # Document 3: Trailing dots field name "a..."
-  - do:
-      index:
-        index: test_disabled_object_4896
-        id: 3
-        body:
-          "@timestamp": "2025-11-26T17:12:00.000Z"
-          message: "trailing dots"
-          log:
-            "a...": "trailing dots value"
-
-  # Document 4: Leading dot field name ".a"
-  - do:
-      index:
-        index: test_disabled_object_4896
-        id: 4
-        body:
-          "@timestamp": "2025-11-26T17:13:00.000Z"
-          message: "leading dot"
-          log:
-            ".a": "leading dot value"
-
-  # Document 5: Middle dots field name "a..b"
-  - do:
-      index:
-        index: test_disabled_object_4896
-        id: 5
-        body:
-          "@timestamp": "2025-11-26T17:14:00.000Z"
-          message: "middle dots"
-          log:
-            "a..b": "middle dots value"
-
-  # Document 6: Multiple unusual fields in same object
-  - do:
-      index:
-        index: test_disabled_object_4896
-        id: 6
-        body:
-          "@timestamp": "2025-11-26T17:15:00.000Z"
-          message: "mixed fields"
-          log:
-            ".": "dot1"
-            "..": "dot2"
-            "normal": "normal value"
-
-  # Document 7: Malformed top-level field name "log." (trailing dot)
-  - do:
-      index:
-        index: test_disabled_object_4896
-        id: 7
         refresh: true
         body:
-          "@timestamp": "2025-11-26T17:16:00.000Z"
-          message: "malformed top-level"
-          "log.":
-            "nested": "value in log."
+          # Document 1: Single dot field name "."
+          - '{"index": {"_id": "1"}}'
+          - '{"@timestamp": "2025-11-26T17:10:00.000Z", "message": "single dot", "log": {".": "single dot value"}}'
+          # Document 2: Multiple dots field name ".."
+          - '{"index": {"_id": "2"}}'
+          - '{"@timestamp": "2025-11-26T17:11:00.000Z", "message": "multiple dots", "log": {"..": "double dot value"}}'
+          # Document 3: Trailing dots field name "a..."
+          - '{"index": {"_id": "3"}}'
+          - '{"@timestamp": "2025-11-26T17:12:00.000Z", "message": "trailing dots", "log": {"a...": "trailing dots value"}}'
+          # Document 4: Leading dot field name ".a"
+          - '{"index": {"_id": "4"}}'
+          - '{"@timestamp": "2025-11-26T17:13:00.000Z", "message": "leading dot", "log": {".a": "leading dot value"}}'
+          # Document 5: Middle dots field name "a..b"
+          - '{"index": {"_id": "5"}}'
+          - '{"@timestamp": "2025-11-26T17:14:00.000Z", "message": "middle dots", "log": {"a..b": "middle dots value"}}'
+          # Document 6: Multiple unusual fields in same object
+          - '{"index": {"_id": "6"}}'
+          - '{"@timestamp": "2025-11-26T17:15:00.000Z", "message": "mixed fields", "log": {".": "dot1", "..": "dot2", "normal": "normal value"}}'
+          # Document 7: Malformed top-level field name "log." (trailing dot)
+          - '{"index": {"_id": "7"}}'
+          - '{"@timestamp": "2025-11-26T17:16:00.000Z", "message": "malformed top-level", "log.": {"nested": "value in log."}}'
 
 ---
 teardown:
@@ -183,6 +131,27 @@ teardown:
           query: source=test_disabled_object_4896 | where message = 'leading dot' | fields log
   - match: { "total": 1 }
   - match: { "datarows.0.0": { ".a": "leading dot value" } }
+
+  # Querying "log.a" should NOT return the ".a" field value
+  # Because "log.a" means nested field "a" inside "log", not literal ".a"
+  - do:
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: source=test_disabled_object_4896 | where message = 'leading dot' | fields log.a
+  - match: { "total": 1 }
+  - match: { "datarows.0.0": null }
+
+  # To access literal field ".a" inside log, use backticks: log.`.a`
+  - do:
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: source=test_disabled_object_4896 | where message = 'leading dot' | fields log.`.a`
+  - match: { "total": 1 }
+  - match: { "datarows.0.0": "leading dot value" }
 
 ---
 "Middle dots field name preserves original name":


opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -373,15 +377,25 @@ private ExprValue parseStruct(Content content, String prefix, boolean supportArr
...
-                        type(makeField(prefix, entry.getKey())),
-                        supportArrays)));
+                    new JsonPath(fieldKey),
+                    parse(entry.getValue(), fullFieldPath, type(fullFieldPath), supportArrays));
+              } catch (IllegalArgumentException e) {
+                // Return null for invalid field names (e.g., fields consisting only of dots)
+                // This can happen with disabled object fields that bypass field name validation
+                // Log warning to hint users about corrupt data
+                LOG.warn(
+                    "Field '{}' has invalid name and will return null: {}",
```
Swiddis: This is going to be a lot of logging if the user ingests a lot of data like this. Maybe not too bad after compression though?

Swiddis: Ideally we should only log one message like this per query (e.g. listing every field in one shot), instead of doing it for every call to parseStruct. Not sure how much effort that is.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -373,15 +377,25 @@ private ExprValue parseStruct(Content content, String prefix, boolean supportArr
...
                     result,
-                    new JsonPath(entry.getKey()),
-                    parse(
-                        entry.getValue(),
-                        makeField(prefix, entry.getKey()),
-                        type(makeField(prefix, entry.getKey())),
-                        supportArrays)));
+                    new JsonPath(fieldKey),
+                    parse(entry.getValue(), fullFieldPath, type(fullFieldPath), supportArrays));
+              } catch (IllegalArgumentException e) {
```
RyanL1997: This catches all `IllegalArgumentException`s, not just the one from `JsonPath` constructor. This could mask other unexpected issues.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -373,15 +377,25 @@ private ExprValue parseStruct(Content content, String prefix, boolean supportArr
...
+                // Return null for invalid field names (e.g., fields consisting only of dots)
+                // This can happen with disabled object fields that bypass field name validation
+                // Log warning to hint users about corrupt data
+                LOG.warn(
+                    "Field '{}' has invalid name and will return null: {}",
+                    fullFieldPath,
+                    e.getMessage());
+                result.tupleValue().put(fieldKey, ExprNullValue.of());
```
RyanL1997: Is returning null the right behavior? At this point users still not notice their data is corrupt.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
RyanL1997: +1 to
> This is going to be a lot of logging if the user ingests a lot of data like this. Maybe not too bad after compression though?

ahkcs: Updated. Now using ThreadLocal to collect invalid fields and log once per query

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
ahkcs: Good catch! Updated to validate before creating JsonPath instead of catching the broad exception

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
ahkcs: Our expectation is that query engine can tolerate corrupt dataset, return null if can not be processed, and give user some hints that data is corrupt.

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
```diff
@@ -0,0 +1,90 @@
...
+  # Before the fix: ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
+  # After the fix: Query succeeds, invalid field returns null with a warning log,
+  # and valid fields (@timestamp, message) are returned normally
+  # Query valid fields - should succeed
+  - do:
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: source=test_disabled_object_4896 | fields @timestamp, message
```
dai-chen: The issue happened even if corrupted `log` field used?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
dai-chen: I think we should document this somewhere. I see Spark CSV reader for example, actually has `PERMISSIVE`, `DROPMALFORMED` and `FAILFAST` mode to let user decide. Ref: https://spark.apache.org/docs/latest/sql-data-sources-csv.html

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
ahkcs: The expected behavior is that valid fields (@timestamp, message) are returned normally and invalid field(log) returns null with a warning log

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -373,18 +403,38 @@ private ExprValue parseStruct(Content content, String prefix, boolean supportArr
...
         .map()
         .forEachRemaining(
-            entry ->
+            entry -> {
+              String fieldKey = entry.getKey();
+              String fullFieldPath = makeField(prefix, fieldKey);
+              // Check for invalid field names (e.g., fields consisting only of dots like "." or
+              // "..")
+              // before creating JsonPath to avoid masking other IllegalArgumentExceptions
+              if (isInvalidFieldName(fieldKey)) {
```
dai-chen: Can we move this and logging logic to higher level? I thought this should be general handling for any malformed field we cannot process? Ref: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/FailureSafeParser.scala#L62

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
dai-chen: Nvm, I was confused by the comments above. Probably you can update it.

RyanL1997 (dismissed):
Generally LGTM. Please also address Chen's comments.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
penghuo: +1

3 questions
1. Why is defintion of corrupted dataset? The example in PR description is valid json.
2. How should the query engine handle corrupted data? Should it skip the malformed record, replace it with null, or use a user-defined placeholder such as "CORRUPTED"?
3. How can error details be surfaced to users who do not have permission to access server-side logs?

opensearch/src/test/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactoryTest.java
```diff
@@ -1041,6 +1041,41 @@ public void testPopulateValueRecursive() {
...
+    // Field name "." (single dot) should return the actual value, not crash or return null
+    // This can happen with disabled object fields in OpenSearch
+    Map<String, ExprValue> result = tupleValue("{\"structV\":{\".\":\"value\"}}");
+    ExprValue structValue = result.get("structV");
+    // The "." field should contain the actual value
+    assertEquals(stringValue("value"), structValue.tupleValue().get("."));
+  }
+
+  @Test
+  public void constructWithMultipleDotsFieldNameReturnsValue() {
```
dai-chen, line 1056: np: Could you verify what if the malformed field is not on top level? E.g., `a..`.

ahkcs:
Updated PR to treat dot-containing field names as literal field names (returning actual data) instead of treating them as invalid.

opensearch/src/test/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactoryTest.java
ahkcs: Updated to have more comprehensive test cases for both UT and yaml test

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
```diff
@@ -0,0 +1,246 @@
...
+              log:
+                type: object
+                enabled: false
+              "@timestamp":
+                type: date
+              message:
+                type: text
+
+  # Document 1: Single dot field name "."
+  - do:
```
penghuo, line 35: Change L35-L112 single doc index to bulk index 

```
  - do:
      bulk:
        index: test
        refresh: true
        body:
          - '{"index": {}}'
          - '{"id": 1}'
          - '{"index": {}}'
          - '{"id": 2}'
          - '{"index": {}}'
          - '{"id": 3}'
```

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
```diff
@@ -0,0 +1,246 @@
...
+      features:
+        - headers
+
+  # Field ".a" should NOT create nested structure
+  - do:
+      headers:
+        Content-Type: 'application/json'
+      ppl:
+        body:
+          query: source=test_disabled_object_4896 | where message = 'leading dot' | fields log
```
penghuo, line 131: what is results if using  `fields log.a`?  should be log.a not exist, right?, user should use `fields log.'.a'`

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -411,11 +414,53 @@ static void populateValueRecursive(ExprTupleValue result, JsonPath path, ExprVal
...
+     *       strings, use literal to avoid creating nested structures with empty keys
+     * </ul>
+     *
+     * @param path The dot-separated path
+     * @return A JsonPath with components split by dots, or literal if splitting would produce empty
+     *     keys
+     */
+    public static JsonPath fromPath(String path) {
+      // Use -1 limit to preserve trailing empty strings (e.g., "a..." -> ["a", "", "", ""])
+      String[] parts = path.split("\\.", -1);
```
Swiddis, line 447: thought: Is this the right place for this logic?

As a wider design point (not just this PR): We've had bugs in different features that boil down to people rolling their own logic for field parsing. Spath was a victim too, and I've called out similar cases in past PRs.

Ideally we should be trying to move this logic into a central module so every part of the code is using the same logic with appropriate edge-case handling. I know  `QualifiedName` defines `DELIMITER = "."` in a full constant and contains some logic like this, while other places still do these kinds of array operations directly.

Especially since there's more-or-less an implicit assumption in many code paths that `.` defines nesting, I'm expecting this isn't the last we'll see of this issue. Having the code be reusable for later seems like a good investment.

Swiddis (dismissed):
lgtm functionally, consider following up with docs

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
ahkcs: Updated

integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/4896.yml
ahkcs: Yes, I have added test cases to verify this behavior
fields `log.a` → returns null
fields `log.'.a'`→ returns"leading dot value"

ahkcs:
Moved to #4907

PR #4900, opened 2025-12-03
Remove doPrivileged call in Calcite script execution

Swiddis:
### Description
Seems to delete a major perf bottleneck: we create an access controller context for every individual document we process in a script, which involves a lot of stack trace traversal and locking for accessing that context. Curious what the breakage is if I just delete this controller step, since it passes tests locally. At least for the query I was interested in, it reduced the runtime from 70 seconds (2mil documents) to 4.

Better approach if we need it: move the privilege step to outside the core script loop.

Measured from slow query (big5 dataset):
```
source = big5
| where `agent.name` = 'filebeat'
| where `@timestamp` >= '2023-01-01 00:00:00' and `@timestamp` <= '2023-01-02 00:00:00'
| where `metrics.size` > 1000 and `metrics.size` != 5000
| stats count(`agent.name`)
```

Before (70s/query):
image

After (4s/query):
image

### Related Issues
N/A

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/core/CalciteScript.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/core/CalciteScript.java
@@ -5,8 +5,6 @@
 
 package org.opensearch.sql.opensearch.storage.script.core;
 
-import java.security.AccessController;
-import java.security.PrivilegedAction;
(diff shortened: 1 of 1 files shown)


songkant-aws:
Grreat! I remember somehow 2.19 still needs it. But it was removed in OpenSearch since 3.x version.

PR #4914, opened 2025-12-08
RexCall and RelDataType standardization for script push down

### LINKED ISSUE #4757: [RFC] RexNode standardization for script push down

qianheng-aws:
### Description
This PR continues https://github.com/opensearch-project/sql/pull/4795 work, it includes change:
1. Implement RexCall standardization by using `RexNormalize.normalize`
2. Implement RelDataType standardization by widening the type, see details in `RexStandardizer::widenType`
3. Support Sarg literal value by expanding `SEARCH` to conjunction expressions.
4. Support decimal literal value by downgrading to double value.

### Related Issues
Resolves #4757 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/RexStandardizer.java
```diff
@@ -78,7 +106,7 @@ public RexNode visitInputRef(RexInputRef inputRef, ScriptParameterHelper helper)
       helper.digests.add(field.getName());
       helper.sources.add(Source.SOURCE.getValue());
     }
-    return new RexDynamicParam(field.getType(), newIndex);
+    return new RexDynamicParam(widenType(field.getType(), helper.stack.peek()), newIndex);
```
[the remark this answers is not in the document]
qianheng-aws: `helper.stack` here is an `ArrayDeque`, while used as a stack

integ-test/src/test/resources/expectedOutput/calcite/agg_case_cannot_push.yaml
```diff
@@ -6,4 +6,4 @@ calcite:
...
...ICB7CiAgICAgICAgICAgICAgImR5bmFtaWNQYXJhbSI6IDUsCiAgICAgICAgICAgICAgInR5cGUiOiB7CiAgICAgICAgICAgICAgICAidHlwZSI6ICJCSUdJTlQiLAogICAgICAgICAgICAgICAgIm51bGxhYmxlIjogdHJ1ZQogICAgICAgICAgICAgIH0KICAgICAgICAgICAgfSwKICAgICAgICAgICAgewogICAgICAgICAgICAgICJkeW5hbWljUGFyYW0iOiA2LAogICAgICAgICAgICAgICJ0eXBlIjogewogICAgICAgICAgICAgICAgInR5cGUiOiAiQklHSU5UIiwKICAgICAgICAgICAgICAgICJudWxsYWJsZSI6IHRydWUKICAgICAgICAgICAgICB9CiAgICAgICAgICAgIH0KICAgICAgICAgIF0KICAgICAgICB9CiAgICAgIF0KICAgIH0sCiAgICB7CiAgICAgICJkeW5hbWljUGFyYW0iOiA3LAogICAgICAidHlwZSI6IHsKICAgICAgICAidHlwZSI6ICJWQVJDSEFSIiwKICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICJwcmVjaXNpb24iOiAtMQogICAgICB9CiAgICB9LAogICAgewogICAgICAiZHluYW1pY1BhcmFtIjogOCwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgfQogICAgfQogIF0KfQ==\"}","lang":"opensearch_compounded_script","params":{"utcTimestamp": 0,"SOURCES":[0,2,2,2,0,0,2,2,2],"DIGESTS":["age",30,"u30",30,"age","age",40,"u40","u100"]}},"missing_bucket":true,"missing_order":"first","order":"asc"}}}]},"aggregations":{"avg_age":{"avg":{"field":"age"}}}}}}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
```
LantaoJin, line 9: can you add a IT to verify the generated script with skip encoding via format=extended.

qianheng-aws: We have test `testSkipScriptEncodingOnExtendedFormat`. Do you mean set `format=extended` for  `agg_case_cannot_push`?

qianheng-aws: Added in test case `testRexStandardizationForScript`

The script is formatted below, with several changes:
- `>=`   -> `<=`
- SEARCH -> AND(...)
- INT nullable=false -> BIGINT nullable=true
- CHAR[xxx] nullable=false -> VARCHAR nullable=true

```
{
  "op": {
    "name": "CASE",
    "kind": "CASE",
    "syntax": "SPECIAL"
  },
  "operands": [
    {
      "op": {
        "name": "<",
        "kind": "LESS_THAN",
        "syntax": "BINARY"
      },
      "operands": [
        {
          "dynamicParam": 0,
          "type": {
            "type": "BIGINT",
            "nullable": true
          }
        },
        {
          "dynamicParam": 1,
          "type": {
            "type": "BIGINT",
            "nullable": true
          }
        }
      ]
    },
    {
      "dynamicParam": 2,
      "type": {
        "type": "VARCHAR",
        "nullable": true,
        "precision": -1
      }
    },
    {
      "op": {
        "name": "AND",
        "kind": "AND",
        "syntax": "BINARY"
      },
      "operands": [
        {
          "op": {
            "name": "<=",
            "kind": "LESS_THAN_OR_EQUAL",
            "syntax": "BINARY"
          },
          "operands": [
            {
              "dynamicParam": 3,
              "type": {
                "type": "BIGINT",
                "nullable": true
              }
            },
            {
              "dynamicParam": 4,
              "type": {
                "type": "BIGINT",
                "nullable": true
              }
            }
          ]
        },
        {
          "op": {
            "name": "<=",
            "kind": "LESS_THAN_OR_EQUAL",
            "syntax": "BINARY"
          },
          "operands": [
            {
              "dynamicParam": 5,
              "type": {
                "type": "BIGINT",
                "nullable": true
              }
            },
            {
              "dynamicParam": 6,
              "type": {
                "type": "BIGINT",
                "nullable": true
              }
            }
          ]
        }
      ]
    },
    {
      "dynamicParam": 7,
      "type": {
        "type": "VARCHAR",
        "nullable": true,
        "precision": -1
      }
    },
    {
      "dynamicParam": 8,
      "type": {
        "type": "VARCHAR",
        "nullable": true,
        "precision": -1
      }
    }
  ]
}
```

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -2079,4 +2079,17 @@ public void testAliasTypeField() throws IOException {
...
+    enabledOnlyWhenPushdownIsEnabled();
+    assertYamlEqualsIgnoreId(
+        loadExpectedPlan("explain_extended_for_standardization.json"),
+        explainQueryToString(
+            String.format(
+                "source=%s | eval age_range = case(age < 30, 'u30', age >= 30 and age <= 40, 'u40'"
+                    + " else 'u100') | stats avg(age) as avg_age by age_range",
+                TEST_INDEX_BANK),
+            true));
+  }
```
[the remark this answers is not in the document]
qianheng-aws: We should avoid reusing `FORMAT` parameter for  `EXPLAIN OPTION` and add a new parameter for it. `EXPLAIN OPTION` affects the content of this API while `FORMAT` only affects the format like JSON, CSV, YAML.

@coderabbitai  Could you create an issue to track this?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/RexStandardizer.java
```diff
@@ -47,17 +64,28 @@ protected RexStandardizer(boolean deep) {
...
-    if (update[0]) {
-      // REVIEW jvs 8-Mar-2005:  This doesn't take into account
-      // the fact that a rewrite may have changed the result type.
-      // To do that, we would need to take a RexBuilder and
-      // watch out for special operators like CAST and NEW where
-      // the type is embedded in the original call.
-      return call.clone(call.getType(), clonedOperands);
-    } else {
-      return call;
+    // Try to expand `SEARCH` before standardization since we cannot translate literal `Sarg`.
```
yuancu, line 67: I encountered a similar problem -- `Sarg` can be serialized but can not be properly deserialized. I addressed it by replacing `Sarg sarg = sargFromJson((Map) literal)` with `Sarg sarg = sargFromJson((Map) literal, type)` in `ExtendedRelJson.java`. Maybe it helps.

yuancu (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/RexStandardizer.java
qianheng-aws: The problem I met here is I cannot translate a Sarg literal into a java object by using `RexToLixTranslator.translateLiteral`.  Thus, we cannot generate a parameter for it.

qianheng-aws: From the implementation of `RexToLixTranslator `, it seems Sarg can only be used as literal in RexNode expression but not DynamicParams.

PR #4921, opened 2025-12-09
Add match support in search cmd

### LINKED ISSUE #4666: [ENHANCEMENT] pushdown simple search query to `match`
```
source = big5 process.name=kernel
```
Now above query is pushdown to `query_string`, it can simplify to `match`

For simple search query (no logical operators and no wildcards), push down to `match` to get better performance.

keep as is

https://docs.opensearch.org/latest/query-dsl/full-text/match/

xinyual:
### Description
This PR push down the simple case of search command to use match instead of using query string.
Pushdown the following message 
```
source = table field=expression
```
To match(field, expression) but omit wildcard cases.

Some behaviors changed 
1. For non-existing field, previous when using query_string it will return empty result but now it will raise error.
2. For type mismatch query, previous when using query_string it will return empty result but now it will raise error.
3. change a Wildcard escaping: Previously I think the escaping it too much: 
Our target is `C:\\Users\\admin` and in PPL query we should only escape once like `\"C:\\\\Users\\\\admin\"`. And in java string, due to the java string escape, we should escape again so it would be `\\\"C:\\\\\\\\Users\\\\\\\admin\\\"`. Not quite sure about lucene escape in the [comment](https://github.com/opensearch-project/sql/blob/4bf5c9c776e7f8cb11714d68fbc2c9163475ef23/integ-test/src/test/java/org/opensearch/sql/ppl/SearchCommandIT.java#L519). I could find result by using match query like
```
{
    "query": {
        "match": {
            "attributes.error.type": "C:\\Users\\admin"
        }
    }
}
``` 

### Related Issues
Resolves #4666 

### Check List


integ-test/src/test/resources/expectedOutput/calcite/big5/keyword_in_range.yaml
```diff
@@ -3,7 +3,7 @@ calcite:
...
...agent, process, process.name, log, message, tags, cloud, input, @timestamp, ecs, data_stream, meta, host, metrics, aws, event], FILTER->AND(match(MAP('field_name', $2), MAP('value', 'kernel':VARCHAR)), SEARCH($8, Sarg[['2023-01-01 00:00:00':VARCHAR..'2023-01-03 00:00:00':VARCHAR)]:VARCHAR)), LIMIT->10, PROJECT->[agent, process, log, message, tags, cloud, input, @timestamp, ecs, data_stream, meta, host, metrics, aws, event], LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":10,"timeout":"1m","query":{"bool":{"must":[{"match":{"process.name":{"query":"kernel","operator":"OR","prefix_length":0,"max_expansions":50,"fuzzy_transpositions":true,"lenient":false,"zero_terms_query":"NONE","auto_generate_synonyms_phrase_query":true,"boost":1.0}}},{"range":{"@timestamp":{"from":"2023-01-01T00:00:00.000Z","to":"2023-01-03T00:00:00.000Z","include_lower":true,"include_upper":false,"format":"date_time","boost":1.0}}}],"adjust_pure_negative":true,"boost":1.0}},"_source":{"includes":["agent","process","log","message","tags","cloud","input","@timestamp","ecs","data_stream","meta","host","metrics","aws","event"],"excludes":[]}}, requestedTotalSize=10, pageSize=null, startFrom=0)])
```
penghuo, line 9: What is improvement of change to match query? In following example, query_string and match generate same Lucene term query.

```
## query_string
{
  "query": {
    "query_string": {
      "query": "hello",
      "default_field": "name"
    }
  }
}

## match
{
  "query": {
    "match": {
      "name": "hello"
    }
  }
}

TermQuery(name:hello)
```

@vamsimanohar

xinyual:
Did a test on big5 dataset with the following query:
```
{"query": {"query_string": {"query": "cloud.region:\"us-east-1\""}}}
{ "query": { "match": { "cloud.region":"us-east-1" } } }
```
And get the following result:
Query_string:
|                                        50th percentile latency |                    query_string |    2.37813 |     ms |
|                                        90th percentile latency |                    query_string |    2.51361 |     ms |
|                                        99th percentile latency |                    query_string |    2.66104 |     ms |
|                                       100th percentile latency |                    query_string |    2.66104 |     ms |

match:
|                                        50th percentile latency |                    match |    2.24579 |     ms |
|                                        90th percentile latency |                    match |    2.34332 |     ms |
|                                        99th percentile latency |                    match |    2.46091 |     ms |
|                                       100th percentile latency |                    match |    2.46091 |     ms |

Looks the performance is almost same.
So I think the optimization is meaningless? @LantaoJin @penghuo

xinyual:
Close since the performance is almost same and the optimization is useless.

PR #4926, opened 2025-12-10
Replace duplicated aggregation logic with `aggregateWithTrimming()`

### LINKED ISSUE #4925: [BUG] Replace duplicated aggregation logic with `aggregateWithTrimming()`
**What is the bug?**
In the current implementation of `buildStreamWindowJoinPlan` in the `Streamstats` command, the aggregation for window functions was partially re-implemented inside this method. But we should just use `aggregateWithTrimming()` to avoid duplicating existing functionality.

 - Plugins

``

ishaoxy:
### Description
 Just use `aggregateWithTrimming()` to avoid duplicating existing functionality.

### Related Issues
Resolves #4925 

### Check List


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1829,8 +1828,7 @@ private RelNode buildStreamWindowJoinPlan(
     context.relBuilder.filter(filter);
 
     // aggregate all window functions on right side
-    List<AggCall> aggCalls = buildAggCallsForWindowFunctions(node.getWindowFunctionList(), context);
-    context.relBuilder.aggregate(context.relBuilder.groupKey(), aggCalls);
```
LantaoJin, line 1833: @yuancu @songkant-aws we'd better not to call `context.relBuilder.aggregate` directly, instead, use `aggregateWithTrimming` to build aggregation in stack. I see some codes in `visitChart`, `rankByColumnSplit` and `visitPatterns`. Can you create a issue to replace?

LantaoJin: Or co-auther with @ishaoxy to address them in this PR.

yuancu: I did so because I needed to build multiple aggregations for the chart command. `aggregateWithTrimming` works on PPL's AST (composed of `UnresolvedPlan`) and creates RexNode AST, but I have already gone through this in the first aggregation. Therefore, I had to call `relBuilder.aggregate` the second time I created an aggregation to aggregate on RexNode.

Issue #4789, opened 2025-11-12
[FEATURE] Support pushdown `dedup` complex expression


LantaoJin:
**Is your feature request related to a problem?**
Current we only support pushdown the pattern of `dedup N field1, field2...` to aggregate. But cannot support pushdown the pattern of  `dedup N expr1, expr2...` or `| eval expr1 = ..., expr2 =... | dedup N expr1, expr2`

**What solution would you like?**
Support pushdown dedup complex expression

PR #4929, opened 2025-12-10
Pushdown join with `max=n` option to TopHits aggregation

### LINKED ISSUE #4927: [FEATURE] Speedup Join with option `max=n` by converting to TopHits Aggregation
Currently, the join without option `max=n` will generate a `SortMergeJoin` operator and the `Sort` can be pushed down to each sides.

But for join with option `max=n`, the logical plan contains a window function in right side, we cannot push down the `Sort` to DSL.

This issue will pushdown the right side (with option `max=n`) to a TopHits aggregation which could highly speedup the join processing, even convert the whole join to a `HashJoin`.

#4844

LantaoJin:
### Description
Pushdown join with `max=n` option to TopHits aggregation:
- The right side subsearch with `max=n` will be converted to TopHits aggregation.
- For inner join, the `SortMergeJoin` may be converted to `HashJoin` by reordering the sides of join
- For non-inner join, the right side will be fully pushed down to DSL, rather than executing `WindowFunction` in memory.

### Related Issues
Resolves #4927 

### Check List


LantaoJin:
> <summary>core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java (1)</summary><blockquote>
> 
> `1323-1374`: **Address the gap between documentation and SEMI/ANTI join implementation.**
> 
> The documentation allows `max` option with `semi` and `anti` join types, but the code returns early for SEMI/ANTI joins (lines 1328-1332) before processing the `max` option and applying dedup/limit optimizations. This creates an inconsistency: users can specify `max` with SEMI/ANTI joins per the documented syntax, but it will be silently ignored.
> 
> Either remove the early return to enable `max` support for SEMI/ANTI joins, or add a validation error when `max` is specified with SEMI/ANTI, and update documentation to clarify the limitation.

@coderabbitai the `max` option takes no effect for `semi` and `anti` join types because `semi` and `anti` joins just use left side to filter the records in left side. So the join results in whatever max=1, max=2 or max=∞ are totally same.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -30,6 +30,7 @@
 public class CalciteExplainIT extends ExplainIT {
   @Override
   public void init() throws Exception {
+    GlobalPushdownConfig.enabled = false;
```
qianheng-aws: Remove it?

LantaoJin: sure.

integ-test/src/test/resources/expectedOutput/calcite_no_pushdown/explain_complex_sort_expr_pushdown_for_smj_w_max_option.yaml
```diff
@@ -20,8 +20,8 @@ calcite:
             EnumerableCalc(expr#0..18=[{inputs}], expr#19=['(?<lastname>^[A-Z])'], expr#20=['lastname'], expr#21=[REX_EXTRACT($t6, $t19, $t20)], lastname=[$t21])
               CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_bank]])
           EnumerableSort(sort0=[$6], dir0=[ASC])
-            EnumerableCalc(expr#0..19=[{inputs}], expr#20=[1], expr#21=[<=($t19, $t20)], proj#0..12=[{exprs}], $condition=[$t21])
-              EnumerableWindow(window#0=[window(partition {6} order by [6] rows between UNBOUNDED PRECEDING and CURRENT ROW aggs [ROW_NUMBER()])])
-                EnumerableCalc(expr#0..18=[{inputs}], expr#19=[IS NOT NULL($t6)], proj#0..18=[{exprs}], $condition=[$t19])
-                  EnumerableLimit(fetch=[50000])
-                    CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_bank]])
+            EnumerableLimit(fetch=[50000])
```
qianheng-aws, line 23: The behavior of system limit has changed from `limitation of source` to `limitation of the results after top hits`.

So if we cannot push down the window, it will scan all rows from the source. @LantaoJin

LantaoJin: I corrected the behavior since the define of `plugins.ppl.join.subsearch_maxout` is

> The size configures the maximum of rows from subsearch to join against.

qianheng-aws: Can we ensure the window will always be pushed down? Otherwise it will get regression than before?

LantaoJin: No if (1) the join keys contain text (not keyword); (2) the join keys contain expression (I am working on support #4789 which could resolve it)

PR #4944, opened 2025-12-11
Mvexpand feature

### LINKED ISSUE #4439: Add native support for mvexpand command in PPL

srikanthpadakanti:
### Description
This pull request adds native support for the mvexpand command in PPL to OpenSearch SQL, enabling users to expand multivalue fields (arrays) into separate rows directly within queries. This functionality is analogous to Splunk's mvexpand command and streamlines analytics, dashboarding, and data preparation involving arrays or multivalue fields.

Key features introduced:

Native mvexpand command for PPL queries to expand array fields into separate rows/events.
Optional limit parameter to restrict the number of expanded values per event/document.
Robust handling of empty/null arrays, large arrays (with memory/resource limits), and non-array fields.
Streaming/distributable execution for performance and scalability.
Comprehensive documentation and edge case coverage.
This feature makes OpenSearch SQL more powerful and user-friendly for log analytics, data exploration, and migration from platforms like Splunk.

### Related Issues
Resolves #4439
https://github.com/opensearch-project/sql/issues/4439

### Check List
- [ X] New functionality includes testing.
- [ X] New functionality has been documented.
 - [ X] New functionality has javadoc added.
 - [ X] New functionality has a user manual doc added.
- [ X] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ X] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ X] Commits are signed per the DCO using `--signoff` or `-s`.

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
@@ -26,6 +26,7 @@ import static org.opensearch.sql.util.MatcherUtils.verifyErrorMessageContains;
 
 import java.io.IOException;
 import java.util.Locale;
+import org.apache.commons.text.StringEscapeUtils;
 import org.junit.Assert;
 import org.junit.Ignore;
 import org.junit.Test;
@@ -2523,6 +2524,21 @@ public class CalciteExplainIT extends ExplainIT {
             StringUtils.format(
                 "source=%s | stats count(eval(author.name < 'K')) as george_and_jk",
                 TEST_INDEX_CASCADED_NESTED)));
+  }
+
+  @Test
+  public void testFieldFormatExplain() throws Exception {
+
+    enabledOnlyWhenPushdownIsEnabled();
+    String expected = loadExpectedPlan("explain_field_format.yaml");
+    assertYamlEqualsIgnoreId(
+        expected,
+        explainQueryYaml(
+            StringEscapeUtils.escapeJson(
+                StringUtils.format(
+                    "source=%s | head 5| fieldformat formatted_balance ="
+                        + " \"$\".tostring(balance,\"commas\") ",
+                    TEST_INDEX_ACCOUNT))));
   }
 
   @Test
--- a/core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
+++ b/core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
@@ -527,6 +527,12 @@ public class Analyzer extends AbstractNodeVisitor<LogicalPlan, AnalysisContext> 
     return new LogicalEval(child, expressionsBuilder.build());
   }
 
+  /** Build {@link LogicalEval}. */
+  @Override
+  public LogicalPlan visitFieldFormat(Eval node, AnalysisContext context) {
+    throw getOnlyForCalciteException("fieldformat");
+  }
+
   @Override
   public LogicalPlan visitAddTotals(AddTotals node, AnalysisContext context) {
     throw getOnlyForCalciteException("addtotals");
--- a/core/src/main/java/org/opensearch/sql/ast/AbstractNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/ast/AbstractNodeVisitor.java
@@ -265,6 +265,10 @@ public abstract class AbstractNodeVisitor<T, C> {
     return visitChildren(node, context);
   }
 
+  public T visitFieldFormat(Eval node, C context) {
+    return visitChildren(node, context);
+  }
+
   public T visitParse(Parse node, C context) {
     return visitChildren(node, context);
   }
--- a/core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
@@ -628,6 +628,12 @@ public class FieldResolutionVisitor extends AbstractNodeVisitor<Node, FieldResol
   }
 
   @Override
+  public Node visitFieldFormat(Eval node, FieldResolutionContext context) {
+    visitChildren(node, context);
+    return node;
+  }
+
+  @Override
   public Node visitExpand(Expand node, FieldResolutionContext context) {
     Set<String> expandFields = extractFieldsFromExpression(node.getField());
     context.pushRequirements(context.getCurrentRequirements().or(expandFields));
--- a/core/src/main/java/org/opensearch/sql/ast/expression/Let.java
+++ b/core/src/main/java/org/opensearch/sql/ast/expression/Let.java
@@ -20,6 +20,8 @@ import org.opensearch.sql.calcite.plan.OpenSearchConstants;
 public class Let extends UnresolvedExpression {
   private final Field var;
   private final UnresolvedExpression expression;
+  private final Literal concatPrefix;
+  private final Literal concatSuffix;
 
   public Let(Field var, UnresolvedExpression expression) {
     String varName = var.getField().toString();
@@ -29,6 +31,21 @@ public class Let extends UnresolvedExpression {
     }
     this.var = var;
     this.expression = expression;
+    this.concatPrefix = null;
+    this.concatSuffix = null;
+  }
+
+  public Let(
+      Field var, UnresolvedExpression expression, Literal concatPrefix, Literal concatSuffix) {
+    String varName = var.getField().toString();
+    if (OpenSearchConstants.METADATAFIELD_TYPE_MAP.containsKey(varName)) {
+      throw new IllegalArgumentException(
+          String.format("Cannot use metadata field [%s] as the eval field.", varName));
+    }
+    this.var = var;
+    this.expression = expression;
+    this.concatPrefix = concatPrefix;
+    this.concatSuffix = concatSuffix;
   }
 
   @Override
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -398,6 +398,28 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
   @Override
   public RexNode visitLet(Let node, CalcitePlanContext context) {
     RexNode expr = analyze(node.getExpression(), context);
+    if (node.getConcatPrefix() != null) {
+
+      expr =
+          context.rexBuilder.makeCall(
+              SqlStdOperatorTable.CONCAT,
+              context.rexBuilder.makeLiteral(
+                  node.getConcatPrefix().getValue(),
+                  context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.VARCHAR),
+                  true),
+              expr);
+    }
+    if (node.getConcatSuffix() != null) {
+
+      expr =
+          context.rexBuilder.makeCall(
+              SqlStdOperatorTable.CONCAT,
+              expr,
+              context.rexBuilder.makeLiteral(
+                  node.getConcatSuffix().getValue(),
+                  context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.VARCHAR),
+                  true));
+    }
     return context.relBuilder.alias(expr, node.getVar().getField().toString());
   }
 
--- a/docs/category.json
+++ b/docs/category.json
@@ -17,6 +17,7 @@
     "user/ppl/cmd/describe.md",
     "user/ppl/cmd/eventstats.md",
     "user/ppl/cmd/eval.md",
+    "user/ppl/cmd/fieldformat.md",
     "user/ppl/cmd/fields.md",
     "user/ppl/cmd/fillnull.md",
     "user/ppl/cmd/grok.md",
--- a/docs/user/ppl/cmd/fieldformat.md
+++ b/docs/user/ppl/cmd/fieldformat.md
@@ -1,0 +1,128 @@
+
+# fieldformat
+
+The `fieldformat` command sets the value to a field with the specified expression and appends the field with evaluated result to the search results. The command is an alias of eval command.
+Additionally, it also provides string concatenation dot operator followed by and/or follows a string that will be concatenated to the expression.
+
+
+## Syntax
+
+The `fieldformat` command has the following syntax:
+
+```syntax
+ fieldformat <field>=[(prefix).]<expression>[.(suffix)] ["," <field>=[(prefix).]<expression>[.(suffix)] ]...
+
+```
+
+## Parameters
+
+The `fieldformat` command supports the following parameters.
+
+| Parameter| Required/Optional | Description                                                                                                                                   |
+|----------------|-------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|
+| `<field>`      | Required          | The name of the field to create or update. If the field does not exist, a new field is added. If it already exists, its value is overwritten. |
+| `<expression>` | Required          | The expression to evaluate.  The expression can have a prefix and/or suffix string part that will be concatenated to the expression.          |
+| `prefix`       | Optional          | A string before the expression followed by dot operator which will be concatenated as prefix to the evaluated expression value.               |
+| `suffix`       | Optional          | A string that follows  the expression and dot operator which will be concatenated as suffix to the evaluated expression value.                |
+  
+
+## Example 1: Create a new field  
+
+The following query creates a new `doubleAge` field for each document:
+  
+```ppl
+source=accounts
+| fieldformat doubleAge = age * 2
+| fields age, doubleAge
+```
+  
+The query returns the following results:
+  
+```text
+fetched rows / total rows = 4/4
++-----+-----------+
+| age | doubleAge |
+|-----+-----------|
+| 32  | 64        |
+| 36  | 72        |
+| 28  | 56        |
+| 33  | 66        |
++-----+-----------+
+```
+  
+
+## Example 2: Override an existing field  
+
+The following query overrides the `age` field by adding `1` to its value:
+  
+```ppl
+source=accounts
+| fieldformat age = age + 1
+| fields age
+```
+  
+The query returns the following results:
+  
+```text
+fetched rows / total rows = 4/4
++-----+
+| age |
+|-----|
+| 33  |
+| 37  |
+| 29  |
+| 34  |
++-----+
+```
+  
+
+  
+
+## Example 3: String concatenation with prefix 
+
+The following query uses the `.` (dot) operator for string concatenation. You can concatenate string literals and field values as follows:
+  
+```ppl
+source=accounts 
+| fieldformat greeting = 'Hello '.tostring( firstname) 
+| fields firstname, greeting
+```
+  
+The query returns the following results:
+  
+```text
+fetched rows / total rows = 4/4
++-----------+---------------+
+| firstname | greeting      |
+|-----------+---------------|
+| Amber     | Hello Amber   |
+| Hattie    | Hello Hattie  |
+| Nanette   | Hello Nanette |
+| Dale      | Hello Dale    |
++-----------+---------------+
+```
+  
+
+## Example 4: String concatenation with dot operator, prefix and suffix
+
+The following query performs prefix and suffix string concatenation operations using dot operator:
+
+```ppl
+source=accounts | fieldformat age_info =  'Age: '.CAST(age AS STRING).' years.' | fields firstname, age, age_info
+```
+
+The query returns the following results:
+
+```text
+fetched rows / total rows = 4/4
++-----------+-----+----------------+
+| firstname | age | age_info       |
+|-----------+-----+----------------|
+| Amber     | 32  | Age: 32 years. |
+| Hattie    | 36  | Age: 36 years. |
+| Nanette   | 28  | Age: 28 years. |
+| Dale      | 33  | Age: 33 years. |
++-----------+-----+----------------+
+```
+
+
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/CalciteNoPushdownIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/CalciteNoPushdownIT.java
@@ -34,6 +34,7 @@ import org.opensearch.sql.ppl.PPLIntegTestCase;
   CalciteDedupCommandIT.class,
   CalciteDescribeCommandIT.class,
   CalciteExpandCommandIT.class,
+  CalciteFieldFormatCommandIT.class,
   CalciteMvExpandCommandIT.class,
   CalciteFieldsCommandIT.class,
   CalciteFillNullCommandIT.class,
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFieldFormatCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteFieldFormatCommandIT.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.remote;
+
+import static org.opensearch.sql.util.MatcherUtils.*;
+
+import java.io.IOException;
+import org.apache.commons.text.StringEscapeUtils;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+import org.opensearch.client.Request;
+import org.opensearch.sql.ppl.PPLIntegTestCase;
+
+public class CalciteFieldFormatCommandIT extends PPLIntegTestCase {
+
+  @Override
+  public void init() throws Exception {
+    super.init();
+    enableCalcite();
+
+    loadIndex(Index.BANK);
+
+    // Create test data for string concatenation
+    Request request1 = new Request("PUT", "/test_eval/_doc/1?refresh=true");
+    request1.setJsonEntity("{\"name\": \"Alice\", \"age\": 25, \"title\": \"Engineer\"}");
+    client().performRequest(request1);
+
+    Request request2 = new Request("PUT", "/test_eval/_doc/2?refresh=true");
+    request2.setJsonEntity("{\"name\": \"Bob\", \"age\": 30, \"title\": \"Manager\"}");
+    client().performRequest(request2);
+
+    Request request3 = new Request("PUT", "/test_eval/_doc/3?refresh=true");
+    request3.setJsonEntity("{\"name\": \"Charlie\", \"age\": null, \"title\": \"Analyst\"}");
+    client().performRequest(request3);
+  }
+
+  @Test
+  public void testFieldFormatStringConcatenation() throws IOException {
+    JSONObject result =
+        executeQuery(
+            StringEscapeUtils.escapeJson(
+                "source=test_eval | fieldformat greeting = 'Hello ' + name"));
+    verifySchema(
+        result,
+        schema("name", "string"),
+        schema("title", "string"),
+        schema("age", "bigint"),
+        schema("greeting", "string"));
+    verifyDataRows(
+        result,
+        rows("Alice", "Engineer", 25, "Hello Alice"),
+        rows("Bob", "Manager", 30, "Hello Bob"),
+        rows("Charlie", "Analyst", null, "Hello Charlie"));
+  }
+
+  @Test
+  public void testFieldFormatStringConcatenationWithNullFieldToString() throws IOException {
+    JSONObject result =
+        executeQuery(
+            StringEscapeUtils.escapeJson(
+                "source=test_eval | fieldformat age_desc = \"Age: \".tostring(age,\"commas\") |"
+                    + " fields name, age, age_desc"));
+    verifySchema(
+        result, schema("name", "string"), schema("age", "bigint"), schema("age_desc", "string"));
+    verifyDataRows(
+        result,
+        rows("Alice", 25, "Age: 25"),
+        rows("Bob", 30, "Age: 30"),
+        rows("Charlie", null, null));
+  }
+
+  @Test
+  public void testFieldFormatStringConcatenationWithNullField() throws IOException {
+    JSONObject result =
+        executeQuery(
+            StringEscapeUtils.escapeJson(
+                "source=test_eval | fieldformat age_desc = \"Age: \".CAST(age AS STRING) | fields"
+                    + " name, age, age_desc"));
+    verifySchema(
+        result, schema("name", "string"), schema("age", "bigint"), schema("age_desc", "string"));
+    verifyDataRows(
+        result,
+        rows("Alice", 25, "Age: 25"),
+        rows("Bob", 30, "Age: 30"),
+        rows("Charlie", null, null));
+  }
+
+  @Test
+  public void testFieldFormatStringConcatWithSuffix() throws IOException {
+    JSONObject result =
+        executeQuery(
+            StringEscapeUtils.escapeJson(
+                "source=test_eval | fieldformat age_desc = CAST(age AS STRING).\" years\" | fields"
+                    + " name, age, age_desc"));
+    verifySchema(
+        result, schema("name", "string"), schema("age", "bigint"), schema("age_desc", "string"));
+    verifyDataRows(
+        result,
+        rows("Alice", 25, "25 years"),
+        rows("Bob", 30, "30 years"),
+        rows("Charlie", null, null));
+  }
+
+  @Test
+  public void testFieldFormatStringConcatWithPrefixSuffix() throws IOException {
+    JSONObject result =
+        executeQuery(
+            StringEscapeUtils.escapeJson(
+                "source=test_eval | fieldformat age_desc = \"Age: \".CAST(age AS STRING).\" years\""
+                    + " | fields name, age, age_desc"));
+    verifySchema(
+        result, schema("name", "string"), schema("age", "bigint"), schema("age_desc", "string"));
+    verifyDataRows(
+        result,
+        rows("Alice", 25, "Age: 25 years"),
+        rows("Bob", 30, "Age: 30 years"),
+        rows("Charlie", null, null));
+  }
+}
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/NewAddedCommandsIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/NewAddedCommandsIT.java
@@ -228,6 +228,19 @@ public class NewAddedCommandsIT extends PPLIntegTestCase {
     verifyQuery(result);
   }
 
+  @Test
+  public void testFieldFormatCommand() throws IOException {
+    JSONObject result;
+    try {
+      executeQuery(
+          String.format(
+              "search source=%s  | fieldformat double_balance = balance * 2 ", TEST_INDEX_BANK));
+    } catch (ResponseException e) {
+      result = new JSONObject(TestUtils.getResponseBody(e.getResponse()));
+      verifyQuery(result);
+    }
+  }
+
   private void verifyQuery(JSONObject result) throws IOException {
     if (isCalciteEnabled()) {
       assertFalse(result.getJSONArray("datarows").isEmpty());
--- a/integ-test/src/test/java/org/opensearch/sql/security/CalciteCrossClusterSearchIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/security/CalciteCrossClusterSearchIT.java
@@ -13,6 +13,7 @@ import static org.opensearch.sql.util.MatcherUtils.verifyDataRows;
 import static org.opensearch.sql.util.MatcherUtils.verifySchema;
 
 import java.io.IOException;
+import org.apache.commons.text.StringEscapeUtils;
 import org.json.JSONObject;
 import org.junit.Test;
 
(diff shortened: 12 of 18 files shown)


core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1098,6 +1098,14 @@ void populate() {
                   OperandTypes.family(SqlTypeFamily.ARRAY, SqlTypeFamily.INTEGER)
                       .or(OperandTypes.family(SqlTypeFamily.MAP, SqlTypeFamily.ANY)),
               false));
+      // Allow using INTERNAL_ITEM when the element type is unknown/undefined at planning time.
+      // Some datasets (or Calcite's type inference) may give the element an UNDEFINED type.
+      // Accept a "ignore" first-argument family so INTERNAL_ITEM(elem, 'key') can still be planned
+      // and resolved at runtime (fallback semantics handled at execution side). - Used in MVEXPAND
+      registerOperator(
```
dai-chen, line 1119: Not sure if missed some context. What is this used for? Please point me to comment related if any.

docs/user/ppl/index.rst
```diff
@@ -0,0 +1,138 @@
+
+===============================
```
dai-chen: Is this PR diff issue? Why is all old content gone?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
srikanthpadakanti: This code defines a fallback type-checker for INTERNAL_ITEM to handle cases where the element type is undefined at planning time. It ensures the operator accepts composite types (IGNORE, CHARACTER) for resilient type validation during query execution.

docs/user/ppl/index.rst
srikanthpadakanti: Removed the file. It shouldn't be there as per upstream.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
dai-chen: > Some datasets (or Calcite's type inference) may give the element an UNDEFINED type

Could you clarify a little more which query/test dataset you have this problem? If any previous comment related, please let me know. Thanks!

srikanthpadakanti: Without that code, tests fail because the INTERNAL_ITEM operator requires specific argument types ({ARRAY, INTEGER} or {STRUCT, ANY}) for type validation during query execution. When Calcite assigns an UNDEFINED type at planning time, the absence of fallback handling in registerOperator causes validation to fail, triggering ExpressionEvaluationException.

With that code, the fallback type-checker accepts UNDEFINED types as valid arguments (IGNORE, CHARACTER), allowing queries to pass type validation, execute correctly, and produce expected results in the tests.

Example - I commented that code and ran my IT below is the failure 

testMvexpandDuplicate

org.opensearch.client.ResponseException: method [POST], host [http://127.0.0.1:65246], URI [/_plugins/_ppl], status line [HTTP/1.1 400 Bad Request]
{
  "error": {
    "reason": "Invalid Query",
    "details": "INTERNAL_ITEM function expects {[ARRAY,INTEGER]|[STRUCT,ANY]}, but got [UNDEFINED,STRING]",
  },
  "status": 400
}

dai-chen (approved):

srikanthpadakanti:
Hello @penghuo @ykmr1224 Can you please look into this PR. Thanks

core/src/main/java/org/opensearch/sql/ast/tree/MvExpand.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import javax.annotation.Nullable;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import org.opensearch.sql.ast.AbstractNodeVisitor;
+import org.opensearch.sql.ast.expression.Field;
+
+/** AST node representing an {@code mvexpand <field> [limit N]} operation. */
```
ykmr1224: `[limit N]` -> `[limit=N]`
It might be better referring as `mvexpand PPL command` to avoid inconsistency.

docs/user/ppl/cmd/mvexpand.md
```diff
@@ -0,0 +1,126 @@
...
+- The test index name used in these examples is `mvexpand_logs`.
+
+## Examples
+
+### Example 1: Basic Expansion (single document)
+Input document (case "basic") contains three tag values.
+
+PPL query:
+```ppl
+source=mvexpand_logs | where case='basic' | mvexpand tags | fields tags.value
```
ykmr1224: Why do we output value to `field.value`? Are we intentionally making it different from Splunk command spec or due to some reason?

docs/user/ppl/cmd/mvexpand.md
```diff
@@ -0,0 +1,126 @@
...
+```
+mvexpand <field> [limit=<int>]
+```
+
+- `<field>`: The multivalue (array) field to expand. (Required)
+- `limit`: Maximum number of values per document to expand. (Optional)
+
+## Notes about these doctests
+- The tests below target a single, deterministic document by using `where case='<name>'` so the doctests are stable.
+- The test index name used in these examples is `mvexpand_logs`.
```
ykmr1224: Looking at the examples, it would be easier to understand if we use `array` function to make and multivalue field, and expand it by `mvexpand` command. And then, we don't need to add new index `mvexpand_logs` for this doc.

Refer here for examples: https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/collection.md

integ-test/src/test/resources/expectedOutput/calcite/explain_mvexpand.yaml
```diff
@@ -0,0 +1,7 @@
+calcite:
+  logical: |
+    LogicalSystemLimit(sort0=[$2], dir0=[ASC], fetch=[10000], type=[QUERY_SIZE_LIMIT])
+      LogicalProject(skills=[$0], username=[$3], VALUE=[$10])
+        LogicalValues(tuples=[[]])
+  physical: |
+    EnumerableValues(tuples=[[]])
```
ykmr1224: Looks it does not do anything related to `mvexpand`.

integ-test/src/test/resources/expectedOutput/calcite_no_pushdown/explain_mvexpand.yaml
```diff
@@ -0,0 +1,7 @@
+calcite:
+  logical: |
+    LogicalSystemLimit(sort0=[$2], dir0=[ASC], fetch=[10000], type=[QUERY_SIZE_LIMIT])
+      LogicalProject(skills=[$0], username=[$3], VALUE=[$10])
+        LogicalValues(tuples=[[]])
+  physical: |
+    EnumerableValues(tuples=[[]])
```
ykmr1224: Same as above

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3093,11 +3099,102 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+
+    Field field = mvExpand.getField();
+
+    String fieldName = extractFieldName(field);
+
+    // 2. Lookup field
+    RelDataTypeField matched = rowType.getField(fieldName, false, false);
+
+    // 2A. Missing field → true EMPTY relation (no schema, no rows)
+    if (matched == null) {
```
ykmr1224: Why don't we raise error when the field is not found?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3093,11 +3099,102 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+
+    RelBuilder relBuilder = context.relBuilder;
+    RelDataType rowType = relBuilder.peek().getRowType();
+
+    Field field = mvExpand.getField();
+
+    String fieldName = extractFieldName(field);
+
+    // 2. Lookup field
+    RelDataTypeField matched = rowType.getField(fieldName, false, false);
```
ykmr1224: We should rely on `rexVisitor.analyze` to resolve field node.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3093,11 +3099,102 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+   * it unnests the target multivalue field and joins back to the original relation. The additional
+   * mvexpand semantics (such as an optional per-document limit) are surfaced via the MVExpand AST
+   * node but reuse the same underlying RelBuilder pipeline as expand at this layer.
+   *
+   * @param mvExpand MVExpand command to be visited
+   * @param context CalcitePlanContext containing the RelBuilder and other context
+   * @return RelNode representing records with the expanded multi-value field
+   */
+  @Override
+  public RelNode visitMvExpand(MvExpand mvExpand, CalcitePlanContext context) {
```
ykmr1224, line 3372: Why does it have lots of difference with visitExpand? Should it be mostly same other than `limit` parameter?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,366 @@
...
+            "source=%s | mvexpand skills | where username='u_int' | fields username, skills", idx);
+
+    ResponseException ex = assertThrows(ResponseException.class, () -> executeQuery(query));
+    String msg = ex.getMessage();
+    Assertions.assertTrue(
+        msg.contains("Cannot expand field 'skills': expected ARRAY type but found INTEGER"),
+        "Expected SemanticCheckException about non-array integer field, got: " + msg);
+  }
+
+  private static String createTempIndexWithMapping(String baseName, String mappingJson)
```
ykmr1224: I think we already have these helper methods in SQLIntegTestCase. Please use existing ones whenever possible.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,366 @@
...
+      deleteIndexIfExists(INDEX + "_int_field");
+    } catch (Exception ignored) {
+    }
+  }
+
+  @Test
+  public void testMvexpandSingleElement() throws Exception {
+    String query =
+        String.format(
+            "source=%s | mvexpand skills | where username='single' | fields username, skills.name",
```
ykmr1224, line 42: Why does it bring `skills.name` although it expanded `skills`?

core/src/main/java/org/opensearch/sql/ast/tree/MvExpand.java
srikanthpadakanti: Switched to [limit=N] and explicitly referenced this as the mvexpand PPL command

docs/user/ppl/cmd/mvexpand.md
srikanthpadakanti: `field.value` was not intended as a deviation from Splunk. After `mvexpand`, the expanded element remains under the same field name (e.g., `tags`, `ids`). I updated the doctests and added a short note clarifying output naming.

docs/user/ppl/cmd/mvexpand.md
srikanthpadakanti: Agreed. Updated the mvexpand examples to generate multivalue fields inline using the array() function

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: This is intentional. 
For PPL compatibility and pipeline safety, mvexpand treats a missing field as an empty expansion (0 rows) instead of throwing, matching Splunk behavior and avoiding hard failures in chained queries.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Agreed. 
Updated to rely exclusively on rexVisitor.analyze for field resolution.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Refactored so both expand and mvexpand share buildExpandRelNode.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: swapped the local helpers with the SQLIntegTestCase

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: This is expected behavior. 
mvexpand skills expands the array elements into rows but preserves the nested object structure, so subfields like skills.name remain accessible.

This matches Splunk’s mvexpand semantics and is required for downstream field projections to work correctly.

integ-test/src/test/resources/expectedOutput/calcite/explain_mvexpand.yaml
srikanthpadakanti: Previous plan shortcircuited and didn’t hit mvexpand. 
Fixed to validate the real Unnest path now.

integ-test/src/test/resources/expectedOutput/calcite_no_pushdown/explain_mvexpand.yaml
srikanthpadakanti: Previous plan shortcircuited and didn’t hit mvexpand. 
Fixed to validate the real Unnest path now.

srikanthpadakanti:
@ykmr1224 
can you please trigger the workflows and review my changes. thanks.

ppl/src/test/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizerTest.java
```diff
@@ -863,13 +863,15 @@ public void testMvindex() {
   }
 
   @Test
-  public void testMvzip() {
```
dai-chen, line 873: Any reason why we override this existing test?

integ-test/src/test/resources/mvexpand_int_mapping.json
```diff
@@ -0,0 +1,8 @@
+{
```
dai-chen: np: I feel it's unnecessary to create several such index mapping. We can use single one and test different case by different document data?

dai-chen (approved):

docs/user/ppl/cmd/mvexpand.md
```diff
@@ -0,0 +1,166 @@
+# mvexpand
+
+## Description
+The `mvexpand` command expands each value in a multivalue (array) field into a separate row, similar to Splunk's `mvexpand` command. For each document, every element in the specified array field is returned as a new row.
```
penghuo: Please remove the phrase `similar to Splunk's mvexpand command`.
The term similar is vague and may confuse readers by implying partial equivalence or undefined behavioral differences.

docs/user/ppl/cmd/mvexpand.md
```diff
@@ -0,0 +1,166 @@
...
+```
+mvexpand <field> [limit=<int>]
+```
+
+- `<field>`: The multivalue (array) field to expand. (Required)
+- `limit`: Maximum number of values per document to expand. (Optional)
+
+## Notes about these doctests
+- The examples below generate deterministic multivalue fields using `eval` + `array()` so doctests are stable.
+- All examples run against a single source index (`people`) and use `head 1` to keep output predictable.
```
penghuo: Remove it or put it at the end. It is not releated to mvexpand command itself.

docs/user/ppl/cmd/mvexpand.md
```diff
@@ -0,0 +1,166 @@
...
+Null array (case "null"):
+
+PPL query:
+```ppl
+source=people
+| eval tags = array('dummy')
+| fields tags
+| head 1
+| mvexpand tags
+| where false
```
penghuo: I did not understand it. Why add `where false` after mvexpand?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,164 @@
...
+    String ppl = "source=DEPT | mvexpand EMPNOS limit=2";
+    // Smoke test: planning should succeed with mvexpand limit parameter.
+    getRelNode(ppl);
+  }
+
+  @Test
+  public void testMvExpandProjectNested() {
+    String ppl = "source=DEPT | mvexpand EMPNOS | fields DEPTNO, EMPNOS";
+    // Smoke test: projection after mvexpand should plan cleanly.
+    getRelNode(ppl);
```
penghuo: what is this test for? no assertion.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMvExpandTest.java
```diff
@@ -0,0 +1,164 @@
...
+    String ppl = "source=DEPT | where DEPTNO = 999 | mvexpand EMPNOS";
+    getRelNode(ppl);
+  }
+
+  @Test
+  public void testMvExpandPrimitiveArray() {
+    // EMPNOS is already an array of primitives (INTEGER), so this is the primitive-array case.
+    String ppl = "source=DEPT | mvexpand EMPNOS";
+    getRelNode(ppl);
+  }
```
penghuo: ditto

integ-test/src/test/resources/mvexpand_int_mapping.json
penghuo: +1

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3093,11 +3099,76 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+    buildExpandRelNode(arrayFieldRex, fieldName, fieldName, mvExpand.getLimit(), context);
+
+    return relBuilder.peek();
+  }
+
+  private static boolean isMissingFieldException(RuntimeException e) {
+    Throwable t = e;
+    while (t != null) {
+      final String msg = t.getMessage();
+      if (msg != null && msg.matches("Field \\[.+\\] not found\\.?")) {
```
ykmr1224: Please don't rely on exception message. It is very easy to break.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
penghuo: I expected the results is map.
```
{"name", "go"}
```

Did u verify this in splunk?
```
| makeresults format=json data="[
{\"_time\":1755622404, \"jsonarray\": [{\"name\":\"John\"}, {\"name\":\"Bob\"}]}
]"
| eval skills=json_array_to_mv(jsonarray)
| mvexpand skills
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3093,11 +3099,76 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+    final RelBuilder relBuilder = context.relBuilder;
+    final Field field = mvExpand.getField();
+
+    final String fieldName = field.getField().toString();
+
+    final RexInputRef arrayFieldRex;
+    try {
+      arrayFieldRex = (RexInputRef) rexVisitor.analyze(field, context);
+    } catch (IllegalArgumentException e) {
+      if (isMissingFieldException(e)) {
```
ykmr1224: Not sure why we need this exception handling. If this is needed, we should prefer not relying on Exception. (If we cannot avoid, we should throw special Exception class instead of checking message)

docs/user/ppl/cmd/mvexpand.md
ykmr1224: Can you explain the reason why we make this deviation?

docs/user/ppl/index.md
```diff
@@ -78,8 +78,10 @@ source=accounts
 | [describe command](cmd/describe.md) | 2.1 | stable (since 2.1) | Query the metadata of an index. |
 | [explain command](cmd/explain.md) | 3.1 | stable (since 3.1) | Explain the plan of query. |
 | [show datasources command](cmd/showdatasources.md) | 2.4 | stable (since 2.4) | Query datasources configured in the PPL engine. |
- | [addtotals command](cmd/addtotals.md) | 3.4 | stable (since 3.4) | Adds row and column values and appends a totals column and row. | 
- | [addcoltotals command](cmd/addcoltotals.md) | 3.4 | stable (since 3.4) | Adds column values and appends a totals row. | 
+| [addtotals command](cmd/addtotals.md) | 3.4 | stable (since 3.4) | Adds row and column values and appends a totals column and row. | 
+| [addcoltotals command](cmd/addcoltotals.md) | 3.4 | stable (since 3.4) | Adds column values and appends a totals row. |
+| [mvexpand command](cmd/mvexpand.md) | 3.4 | experimental (since 3.4) | Expand a multi-valued field into separate documents (one per value). |
```
ykmr1224: nit: Let's move next to `expand` command.

integ-test/src/test/resources/expectedOutput/calcite/explain_mvexpand.yaml
ykmr1224: I still don't see any plan related to mvexpand.

integ-test/src/test/resources/expectedOutput/calcite_no_pushdown/explain_mvexpand.yaml
ykmr1224: ditto

ppl/src/main/java/org/opensearch/sql/ppl/utils/PPLQueryDataAnonymizer.java
```diff
@@ -654,6 +655,17 @@ public String visitAppend(Append node, String context) {
     return StringUtils.format("%s | append [%s ]", child, subsearch);
   }
 
+  @Override
+  public String visitMvExpand(MvExpand node, String context) {
+    String child = node.getChild().get(0).accept(this, context);
+    String field = MASK_COLUMN; // Always anonymize field names
+    // Optionally handle limit if needed (e.g., | mvexpand identifier limit=***)
```
ykmr1224: nit: redundant comments

srikanthpadakanti:
@penghuo @anasalkouz @dai-chen Can you please review this. Thanks

docs/user/ppl/cmd/mvexpand.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+Expected output:
+```text
+fetched rows / total rows = 0/0
++------+
+| tags |
+|------|
++------+
+```
```
LantaoJin: shouldn't throw semantic exception for a non-existed field.

docs/user/ppl/cmd/mvexpand.md
```diff
@@ -0,0 +1,143 @@
...
```
LantaoJin: I don't see much business meaning for this example. why not add some examples on `projects` fields? for example `source=people | mvexpand projects`

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
```diff
@@ -0,0 +1,347 @@
...
+        "{\"username\":\"duplicate\",\"skills\":[{\"name\":\"dup\"},{\"name\":\"dup\"}]}",
+        "{\"username\":\"complex\",\"skills\":[{\"name\":\"ml\",\"level\":\"expert\"},{\"name\":\"ai\"},{\"level\":\"novice\"}]}",
+        "{\"username\":\"large\",\"skills\":["
+            + "{\"name\":\"s1\"},{\"name\":\"s2\"},{\"name\":\"s3\"},{\"name\":\"s4\"},{\"name\":\"s5\"},"
+            + "{\"name\":\"s6\"},{\"name\":\"s7\"},{\"name\":\"s8\"},{\"name\":\"s9\"},{\"name\":\"s10\"}"
+            + "]}",
+        "{\"username\":\"hetero_types\",\"skills\":[{\"level\":\"senior\"},{\"level\":3}]}",
+        "{\"username\":\"u1\",\"skills_not_array\":\"scala\"}",
+        "{\"username\":\"u_int\",\"skills_int\":5}",
+        "{\"username\":\"limituser\",\"skills\":[{\"name\":\"a\"},{\"name\":\"b\"},{\"name\":\"c\"},{\"name\":\"d\"},{\"name\":\"e\"}]}");
```
LantaoJin: I see you've already add files `mvexpand_edge_cases.json` and `mvexpand_edge_cases_mapping.json`. Why duplicate them in code?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -314,6 +315,14 @@ public void testExplainMultisearchTimestampInterleaving() throws IOException {
     assertYamlEqualsIgnoreId(expected, result);
   }
 
+  @Test
+  public void testMvexpandExplain() throws IOException {
+    // mvexpand explain plan validation
+    String expected = loadExpectedPlan("explain_mvexpand.yaml");
+    explainQueryYaml(
+        "source=mvexpand_edge_cases | eval skills_arr = array(1, 2, 3) | mvexpand skills_arr");
```
LantaoJin, line 357: if just `mvexpand skills_arr` here, no need to load the new index `MVEXPAND_EDGE_CASES`.

how about change to `source=mvexpand_edge_cases | mvexpand skills`?

docs/user/ppl/cmd/mvexpand.md
```diff
@@ -0,0 +1,143 @@
...
+
+
+## Examples
+
+### Example 1: Basic Expansion (single document)
+Input document (case "basic") contains three tag values.
+
+PPL query:
+```ppl
+source=people
```
LantaoJin, line 28: I found you add a new index `mvexpand_logs` for doctest, but never use it?

doctest/test_data/mvexpand_logs.json
```diff
@@ -0,0 +1,6 @@
+{"case":"basic","tags":[{"value":"error"},{"value":"warning"},{"value":"info"}]}
```
LantaoJin: what is this new index used for? I didn't see any test on it.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3111,11 +3117,82 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+   *
+   * @param mvExpand MVExpand command to be visited
+   * @param context CalcitePlanContext containing the RelBuilder and other context
+   * @return RelNode representing records with the expanded multi-value field
+   */
+  /**
+   * MVExpand command visitor.
+   *
+   * <p>For Calcite remote planning, mvexpand reuses the same expansion mechanics as {@link Expand}:
+   * it unnests the target multivalue field and joins back to the original relation.
```
LantaoJin: duplicated javadoc

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3111,11 +3117,82 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+    visitChildren(mvExpand, context);
+
+    final RelBuilder relBuilder = context.relBuilder;
+    final Field field = mvExpand.getField();
+    final String fieldName = field.getField().toString();
+
+    // Missing-field: produce no rows (but keep schema stable).
+    final RelDataType inputType = relBuilder.peek().getRowType();
+    final RelDataTypeField inputField =
+        inputType.getField(fieldName, /*caseSensitive*/ false, /*elideRecord*/ false);
```
LantaoJin: ppl is case sensitive for field name

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3111,11 +3117,82 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+    if (inputField == null) {
+      return buildEmptyResultWithStableSchema(relBuilder, fieldName);
+    }
```
LantaoJin, line 3386: I think we can simplify to throw `SemanticCheckException`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3111,11 +3117,82 @@ public RelNode visitExpand(Expand expand, CalcitePlanContext context) {
...
+    // Resolve field ref using rexVisitor for consistent semantics (same as expand).
+    final RexInputRef arrayFieldRex = (RexInputRef) rexVisitor.analyze(field, context);
+
+    // Enforce ARRAY type before UNNEST so we return SemanticCheckException.
+    final SqlTypeName actual = arrayFieldRex.getType().getSqlTypeName();
+    if (actual != SqlTypeName.ARRAY) {
+      throw new SemanticCheckException(
+          String.format(
+              "Cannot expand field '%s': expected ARRAY type but found %s", fieldName, actual));
+    }
```
LantaoJin, line 3395: This is over restricted. `mvexpand` could apply to a non-array type field. We can treat them an array with only one item.

docs/user/ppl/cmd/mvexpand.md
srikanthpadakanti: Updated the documentation wording.

docs/user/ppl/cmd/mvexpand.md
srikanthpadakanti: Updated Example 3 [Empty Expansion] to use `projects` field.

docs/user/ppl/cmd/mvexpand.md
srikanthpadakanti: This index was originally added for mvexpand examples; following the feedback to reuse existing doctest indices, I’ve cleaned it up and updated the related metadata doctest.

doctest/test_data/mvexpand_logs.json
srikanthpadakanti: This index was originally added for mvexpand examples; following the feedback to reuse existing doctest indices, I’ve cleaned it up and updated the related metadata doctest.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: cleaned.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: I’d like to keep the current behavior here: when the target field isn’t present in the input schema, mvexpand returns an empty result (0 rows) with a stable schema rather than throwing SemanticCheckException. This aligns with the documented mvexpand semantics and avoids failing pipelines where the field is projected out.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Agreed — removed the ARRAY-only restriction.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: Right.. updated.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMvExpandCommandIT.java
srikanthpadakanti: Removed the manual creation and insertion.

LantaoJin:
One question in my mind is what's the different between `mvexpand` and the existed `expand` and should we combine them?

@penghuo do you know why `mvexpand` is in the roadmap and is it a task to refactor the existed `expand` command?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
srikanthpadakanti: done

srikanthpadakanti:
> One question in my mind is what's the different between `mvexpand` and the existed `expand` and should we combine them?
> 
> @penghuo do you know why `mvexpand` is in the roadmap and is it a task to refactor the existed `expand` command?

Refactoring the existing expand command was not part of the original task. However, @ykmr1224 suggested reusing a common implementation for both expand and mvexpand. Based on that guidance, I refactored expand to align with and reuse the shared logic. 

@LantaoJin Please look into this and see if this can be merged. Thanks

srikanthpadakanti:
@LantaoJin @anasalkouz @penghuo Can you please review this. Thanks

srikanthpadakanti:
Looks like these CI failures are not related to my changes. I do not see any mvexpand related issues in the failure logs.

srikanthpadakanti:
Diverged long back. Instead of going through merge issues, closing this and opening a new one.

PR #4957, opened 2025-12-15
Support pushdown dedup with expression

### LINKED ISSUE #4789: [FEATURE] Support pushdown `dedup` complex expression

LantaoJin:
### Description
Support pushdown `dedup` with expression:
- the dedup columns contain expressions
  - `| eval new_gender = lower(gender), new_name = lower(name) | dedup 2 new_gender, new_name`
- the other columns contain expressions
  - `| eval new_gender = lower(gender), new_name = lower(name) | dedup 2 gender, name`
 
This PR also implicitly support pushdown `join with max option` with expressions
- the join keys contain expressions
  - `source = t1 | eval new_gender = lower(gender) | join new_gender [source = t2 | eval new_gender = lower(gender) ]`
- the other columns contain expresssions
  - `source = t1 | eval new_gender = lower(gender) | join gender [source = t2 | eval new_gender = lower(gender) ]`

### Related Issues
Resolves #4789 

### Check List


opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/TopHitsParser.java
```diff
@@ -90,7 +90,12 @@ public List<Map<String, Object>> parse(Aggregation agg) {
       //   LinkedHashMap["name" -> "A", "category" -> "Y"]
       // ]
       return Arrays.stream(hits)
-          .<Map<String, Object>>map(hit -> new LinkedHashMap<>(hit.getSourceAsMap()))
+          .<Map<String, Object>>map(
+              hit -> {
+                Map map = new LinkedHashMap<>(hit.getSourceAsMap());
```
yuancu: nit: use `var` or `Map<String, Object>`

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/TopHitsParser.java
```diff
@@ -90,7 +90,12 @@ public List<Map<String, Object>> parse(Aggregation agg) {
       //   LinkedHashMap["name" -> "A", "category" -> "Y"]
       // ]
```
yuancu, line 91: Seems the comment also needs updated, as it no longer returns source but fields

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -111,79 +105,97 @@ && dedupColumnsContainRexCall(rexCallsExceptWindow, dedupColumns)) {
...
+    // 1 Initial a RelBuilder by pushing Scan and Project
+    if (projectWithExpr == null) {
+      // 1.1 if projectWithExpr not existed, push a scan then create a new project
+      relBuilder.push(scan);
+      List<RexNode> columnsFromScan = relBuilder.fields();
+      List<String> colNamesFromScan = relBuilder.peek().getRowType().getFieldNames();
+      List<Pair<RexNode, String>> namedPairFromScan =
+          Streams.zip(columnsFromScan.stream(), colNamesFromScan.stream(), Pair::new).toList();
+      List<Pair<RexNode, String>> reordered =
+          advanceDedupColumns(namedPairFromScan, dedupColumnIndices, dedupColumnNames);
```
yuancu, line 118: Can you explain the intuition / reason for reordering the columns?

LantaoJin: That's why I suggested you to use aggregateWithTrimming() instead of calling the relBuilder.aggregate() directly. In SQL, the Project added for Aggregate always keeps the group columns in the front of others, it means the group keys set always contain {0}, or there could be trigger a Calcite bug.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/TopHitsParser.java
LantaoJin: It still returns source, the field will be added when scripts are existed.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -73,35 +81,21 @@ protected void apply(
...
+    List<String> dedupColumnNames =
+        dedupColumnIndices.stream()
+            .map(
+                i ->
+                    projectWithWindow.getNamedProjects().stream()
+                        .filter(pair -> pair.getKey().isA(SqlKind.INPUT_REF))
+                        .filter(pair -> ((RexInputRef) pair.getKey()).getIndex() == i)
+                        .map(Pair::getValue)
+                        .findFirst()
+                        .get())
```
qianheng-aws, line 94: Seems this be simplified to 
```
  dedupColumnIndices.stream()
     .map(projectWithWindow.getInput().getRowType().getFieldNames()::get)
     .toList()
```

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -73,35 +81,21 @@ protected void apply(
...
             .toList();
-    if (!rexCallsExceptWindow.isEmpty()
-        && dedupColumnsContainRexCall(rexCallsExceptWindow, dedupColumns)) {
-      // TODO https://github.com/opensearch-project/sql/issues/4789
-      // | eval new_gender = lower(gender) | fields new_gender, age | dedup 1 new_gender
-      if (LOG.isDebugEnabled()) {
-        LOG.debug("Cannot pushdown the dedup since the dedup columns contain RexCall");
-      }
+    if (dedupColumnIndices.size() != dedupColumnNames.size()) {
```
qianheng-aws, line 96: Will this ever happened since dedupColumnNames is derived from dedupColumnIndices?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
qianheng-aws: I don't get it as well. Since the `aggregate` is going to be pushed into scan and the final generated RelNode is a single `Scan` operator, will it really trigger the bug in Calcite? 

Do we do similar thing in the previous PR for supporting dedup push down?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -356,19 +357,22 @@ private static Pair<Builder, List<MetricParser>> processAggregateCalls(
    *
    * @param aggCall the aggregate call
    * @param project the project
-   * @return the converted RexNode list
+   * @return the converted Pair<RexNode, String> list
    */
-  private static List<RexNode> convertAggArgThroughProject(AggregateCall aggCall, Project project) {
+  private static List<org.apache.calcite.util.Pair<RexNode, String>> convertAggArgThroughProject(
```
qianheng-aws: There is already `import org.apache.commons.lang3.tuple.Pair` imported. Could it be switched to that or at least unified?

LantaoJin: sure. will convert to common.lang3's Pair

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/TopHitsParser.java
LantaoJin: The definition is in L93 `.<Map<String, Object>>map(`. But yes, I can add `Map<String, Object>` here and remove L93. The reason use L93 instead of `Map<String, Object>` is we used unnamed variable, but now it has to define a name.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
LantaoJin: No. above code doesn't cover the rename case:
```
source=test | rename status as http_status | dedup http_status | fields http_status | sort http_status
```

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
LantaoJin: I am 100% sure they have same size. Better to keep this check.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
LantaoJin: > I don't get it as well. Since the `aggregate` is going to be pushed into scan and the final generated RelNode is a single `Scan` operator, will it really trigger the bug in Calcite?
> 
> Do we do similar thing in the previous PR for supporting dedup push down?

Yes. I see bugs when the group key columns are not in the front of child Project in Calcite in developing support Aggregate with Calcite. Not sure is it fixed or not. Check the comment here https://github.com/opensearch-project/sql/blob/7dfabcea94952ead0a27463d810097d74446acc4/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java#L1049

qianheng-aws: That happens in CalciteRelNodeVisitor, but is similar change needed in the push down process?

LantaoJin: This logic has no diff with CalciteRelNodeVisitor. It is building an aggregate via relbuilder. I didn't test if I remove the reorder logic. It may not easy to trigger the bug in dedup, but for safe, I added the reorder logic

yuancu (approved):

Issue #4931, opened 2025-12-08
[FEATURE] Prune old RelNode in push down rules


qianheng-aws:
**Is your feature request related to a problem?**
Calcite's planner supports RelNode pruning, which reduces unnecessary exploration of candidate plans within a RelSet. This optimization should be applied when a new plan is definitively cheaper than existing alternatives and offers no additional value for discovering the optimal execution plan.

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**
Calcite has interface `SubstitutionRule`, where there is API `autoPruneOld`  determines whether the planner should automatically prune old node when there is at least 1 equivalent rel generated by the rule. The default behavior is returning false and substitution rules will be executed first in the rule on-match queue.

RyanL1997:
@opensearch-project/triage , Can we move this issue to sql repo? https://github.com/opensearch-project/sql. Thanks in advance.

qianheng-aws:
After some tests on pruning old in our rules, I've found some bad cases:

1. Prune old in project/filter push down rules will prevent aggregate push down
**cases**:  testExplainMaxOnStringField, testAliasTypeField, testCountAggPushDownExplain, ....
**root-cause**: Since we've pruned the old RelNodes, the pattern of agg push down rules cannot match a valid RelSubSets.
**fix way:** 
    - Make agg push down extends SubstitutionRule so it will get higher priority in the rule match queue
    - Extends agg push down rules to support pattern `AGG-SCAN` so we can still push down agg after project and filter push down.

2. Prune old may cause exception in the plan preparing process.
**cases**: org.opensearch.sql.calcite.remote.CalciteExplainIT#testExplainScalarCorrelatedSubqueryInSelect
org.opensearch.sql.calcite.remote.CalciteExplainIT#testJoinWithPushdownSortIntoAgg
org.opensearch.sql.calcite.remote.CalciteExplainIT#testExplainAppendCommand
org.opensearch.sql.calcite.remote.CalciteExplainIT#testSortComplexExprMixedWithSimpleExpr
org.opensearch.sql.calcite.remote.CalciteExplainIT#testComplexSortExpressionProjectThenSort
**root-cause**: some old RelNodes has multiple parents, if we prune it directly in a rule, it cannot be converted to a valid enumerable node in the other branches. In the end, the planner cannot generate a valid final plan.
**fix way**: Detect its parents before pruning a RelNode.

[...]

PR #4992, opened 2025-12-23
Prune old in operator push down rules

qianheng-aws:
### Description
This PR primally supports pruning the old operator if we get a better plan than before. Therefore, it will improve the efficiency of planning process by avoid exploring meaningless equivalent plans.

The performance gain is shown below:
1. Average query cost on big5 and clickbench:

  | previous | pruneOld
-- | -- | --
big5 | 17ms | 14ms
clickbench | 26 ms | 19ms

2. Average optimization cost on big5 and clickbench (By enable calcite's debug mode and adding timingTracer, it will induce more time cost than before)

  | previous | pruneOld
-- | -- | --
big5 | 36.3 ms | 25.9 ms
clickbench | 265 ms | 110.9 ms

3. Average number of applied rules on big5 and clickbench:

  | previous | pruneOld
-- | -- | --
big5 | 114 | 50
clickbench | 474 | 215

Some positive cases on plan with this PR:
- testDedupRename
- testCasePushdownAsRangeQueryExplain
- testExplainOnAggregationWithFunction
- testDedupExpr

### Implementation Details

As described in https://github.com/opensearch-project/sql/issues/4931#issuecomment-3645518565, there is also many issues and bug spotted after pruning old. So there is additional change to fix them and make it compatible:

- As there is Subset reuse in Calcite, pruning a Subset which is the only child of other Subset will cause preparing failure. So we should prune the old from top to down and stop if the current node cannot be pruned. One node cannot be pruned if it's physical node(see the point5 in the above comment) or it has multiple parents(except the root of the call, as we are generating a new root to replace it).
- Make `PPLAggregateConvertRule`, `PPLAggGroupMergeRule`, 'RareTopPushdownRule', 'DedupPushDownRule'  implements `SubstitutionRule` so they will get higher priority on rule match and then we can get optimized aggregates in RelSubset before pruning.
- Support removing `project`, `sort` and agg derived `filter` when doing aggregate push down.
- Slightly refactor `AggregateIndexScanRule` so it can support pushing down on more cases
- Refactor and simplify `DedupPushDownRule` so it can get compatible with the current pruning mechanism
- Continue pushing down limit if it can reduce the estimated row count.
- Fix several bugs in `AggregateAnalyzer` when the project is null. See UT in `AggregateAnalyzerTest`, its expected results is wrong before. 

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/4931

### Check List


core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -646,4 +654,97 @@ private static boolean isNotNullOnRef(RexNode rex) {
...
+   */
+  static void tryPruneRelNodes(RelOptRuleCall call) {
+    if (call.getPlanner() instanceof VolcanoPlanner volcanoPlanner) {
+      Arrays.stream(call.rels)
+          .takeWhile(
+              rel ->
+                  // Don't prune the physical RelNode as it may prevent sort expr push down
+                  rel.getConvention() == Convention.NONE
+                      && (rel == call.rels[0]
+                          || volcanoPlanner.getSubsetNonNull(rel).getParentRels().size() == 1))
```
yuancu, line 746: How does the condition `volcanoPlanner.getSubsetNonNull(rel).getParentRels().size() == 1` guarantee that the rel node's parent is pruned?

Is it because that if the index is greater than 0, then index 0 must have been traversed?

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -33,7 +32,8 @@
 import org.opensearch.sql.opensearch.storage.scan.CalciteLogicalIndexScan;
 
 @Value.Enclosing
-public class DedupPushdownRule extends InterruptibleRelRule<DedupPushdownRule.Config> {
+public class DedupPushdownRule extends InterruptibleRelRule<DedupPushdownRule.Config>
+    implements SubstitutionRule {
```
yuancu, line 36: Why do some rules implement `SubstitutionRule`, while some don't?

core/src/main/java/org/opensearch/sql/calcite/plan/PPLAggGroupMergeRule.java
```diff
@@ -41,7 +43,8 @@
  * these UDFs' output must have equivalent cardinality as `a`.
  */
 @Value.Enclosing
-public class PPLAggGroupMergeRule extends RelRule<PPLAggGroupMergeRule.Config> {
+public class PPLAggGroupMergeRule extends RelRule<PPLAggGroupMergeRule.Config>
+    implements SubstitutionRule {
```
LantaoJin, line 47: should be wrapped in `InterruptibleRelRule`?
```
public class PPLAggGroupMergeRule extends InterruptibleRelRule<PPLAggGroupMergeRule.Config> implements SubstitutionRule
```

core/src/main/java/org/opensearch/sql/calcite/plan/PPLAggGroupMergeRule.java
```diff
@@ -101,6 +104,7 @@ public void apply(RelOptRuleCall call, LogicalAggregate aggregate, LogicalProjec
     parentProjections.addAll(aggCallRefs);
     relBuilder.project(parentProjections);
     call.transformTo(relBuilder.build());
+    PlanUtils.tryPruneRelNodes(call);
```
LantaoJin, line 107: how about move `tryPruneRelNodes(call)` into `InterruptibleRelRule`?
```
    onMatchImpl(call);
+   if (this instanceof SubstitutionRule) {
+     tryPruneRelNodes(call);
+   }
```

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q17.yaml
```diff
@@ -8,4 +8,4 @@ calcite:
...
\ No newline at end of file
+    CalciteEnumerableIndexScan(table=[[OpenSearch, hits]], PushDownContext=[[AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0, 1},count()=COUNT()), SORT_AGG_METRICS->[2 DESC LAST], PROJECT->[count(), UserID, SearchPhrase], LIMIT->10, LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"SearchPhrase|UserID":{"multi_terms":{"terms":[{"field":"SearchPhrase"},{"field":"UserID"}],"size":10,"min_doc_count":1,"shard_min_doc_count":0,"show_term_doc_count_error":false,"order":[{"_count":"desc"},{"_key":"asc"}]}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```
LantaoJin, line 11: Critical: the original query is `count() by UserID, SearchPhrase`, the order matter for results

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -81,169 +71,81 @@ protected void apply(
...
-  }
-
   @Value.Immutable
   public interface Config extends OpenSearchRuleConfig {
-    // Can only push the case with KEEPEMPTY=false:
-    // +- LogicalProject(no _row_number_dedup_)
-    //    +- LogicalFilter(condition contains _row_number_dedup_)
-    //       +- LogicalProject(contains _row_number_dedup_)
-    //          +- LogicalFilter(condition=IS NOT NULL(dedupColumn))"
-    //             +- CalciteLogicalIndexScan
```
yuancu, line 208: I'm curious why can this rule be safely eliminated

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q18.yaml
```diff
@@ -8,4 +8,4 @@ calcite:
...
\ No newline at end of file
+    CalciteEnumerableIndexScan(table=[[OpenSearch, hits]], PushDownContext=[[AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0, 1},count()=COUNT()), PROJECT->[count(), UserID, SearchPhrase], LIMIT->10, LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"composite_buckets":{"composite":{"size":10,"sources":[{"SearchPhrase":{"terms":{"field":"SearchPhrase","missing_bucket":false,"order":"asc"}}},{"UserID":{"terms":{"field":"UserID","missing_bucket":false,"order":"asc"}}}]}}}}, requestedTotalSize=10, pageSize=null, startFrom=0)])
```
LantaoJin, line 11: ditto

integ-test/src/test/resources/expectedOutput/calcite/explain_agg_with_script.yaml
```diff
@@ -6,5 +6,5 @@ calcite:
...
+    EnumerableCalc(expr#0..2=[{inputs}], expr#3=[100], expr#4=[*($t2, $t3)], expr#5=[+($t1, $t4)], expr#6=[CHAR_LENGTH($t0)], sum=[$t5], len=[$t6], gender=[$t0])
+      CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_bank]], PushDownContext=[[AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0},sum_SUM=SUM($1),sum_COUNT=COUNT($1)), LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","aggregations":{"composite_buckets":{"composite":{"size":1000,"sources":[{"gender":{"terms":{"field":"gender.keyword","missing_bucket":true,"missing_order":"first","order":"asc"}}}]},"aggregations":{"sum_SUM":{"sum":{"field":"balance"}},"sum_COUNT":{"value_count":{"field":"balance"}}}}}}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
```
LantaoJin, line 10: Critical: Seems a bad case? the group key is an expression which should be pushed to script agg pushdown

LantaoJin (commented):
I'd like to verify the checksum of clickbench result before being merged

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
LantaoJin: Good question! We need to update dev-doc to explain how to choose proper base rule.

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
qianheng-aws: Yes, if the current RelNode has only 1 parent, it must be the RelNode in `call.rels[current_offset - 1]` and it has been pruned in the previous step of this stream

qianheng-aws: The best way is looking up its all parents in `VolcanoPlanner.prunedNodes` while its unaccessible

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
qianheng-aws: I refactor the dedup push down rule by unifying 2 rules into 1.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
qianheng-aws: Basically, we can only add `SubstitutionRule` if we can ensure it's better than the previous ones. As said in Calcite's comment for `SubstitutionRule`

> A rule that implements this interface indicates that the new RelNode is typically better than the old one.

core/src/main/java/org/opensearch/sql/calcite/plan/PPLAggGroupMergeRule.java
qianheng-aws: As discussed offline, `prune old` shouldn't forcefully be bound to `SubstitutionRule`.  Keep flexibility currently.

integ-test/src/test/resources/expectedOutput/calcite/explain_agg_with_script.yaml
qianheng-aws: This is a good case of applying `PPLAggGroupMergeRule`.

integ-test/src/test/resources/expectedOutput/calcite/clickbench/q17.yaml
qianheng-aws: In SQL, `group by a, b` is equivalent to `group by b, a`, many logic follows this principle in Calcite. And this is case is affected by `AggregateProjectMergeRule` in Calcite.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -180,18 +182,28 @@ <T> T inferValue(RexNode node, Class<T> clazz) {
       }
       throw new IllegalStateException(String.format("Cannot infer value from RexNode %s", node));
     }
+
+    RexNode inferRexNodeFromIndex(int index, Project project) {
```
LantaoJin, line 186: add `private`

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
```diff
@@ -180,18 +182,28 @@ <T> T inferValue(RexNode node, Class<T> clazz) {
...
+      return project == null ? RexInputRef.of(index, rowType) : project.getProjects().get(index);
+    }
+
+    String inferFieldNameFromIndex(int index, Project project) {
```
LantaoJin, line 190: ditto

core/src/main/java/org/opensearch/sql/calcite/plan/PPLAggGroupMergeRule.java
qianheng-aws: InterruptibleRelRule is in package `opensearch` and has dependency on `OpenSearchTimeoutException` while this rule is package `core`.

Therefore, we cannot make this extends `InterruptibleRelRule` unless move that from  package `opensearch` to `core` and add library `opensearch` in core gradle.

On the other hand, if there is interrupt triggered in planning process, it should be detected in our push down rules in package `opensearch`.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/AggregateAnalyzer.java
qianheng-aws: We should give the methods in `AggregateBuilderHelper` package level accessibility. I see all methods in this class are using default symbol

yuancu (approved):

integ-test/src/test/resources/expectedOutput/calcite/explain_dedup_with_expr4.yaml
```diff
@@ -9,5 +9,4 @@ calcite:
...
..."script_fields":{"new_state":{"script":{"source":"{\"langType\":\"calcite\",\"script\":\"rO0ABXQA/HsKICAib3AiOiB7CiAgICAibmFtZSI6ICJMT1dFUiIsCiAgICAia2luZCI6ICJPVEhFUl9GVU5DVElPTiIsCiAgICAic3ludGF4IjogIkZVTkNUSU9OIgogIH0sCiAgIm9wZXJhbmRzIjogWwogICAgewogICAgICAiZHluYW1pY1BhcmFtIjogMCwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgfQogICAgfQogIF0KfQ==\"}","lang":"opensearch_compounded_script","params":{"utcTimestamp": 0,"SOURCES":[0],"DIGESTS":["state.keyword"]}},"ignore_failure":false},"new_gender":{"script":{"source":"{\"langType\":\"calcite\",\"script\":\"rO0ABXQA/HsKICAib3AiOiB7CiAgICAibmFtZSI6ICJMT1dFUiIsCiAgICAia2luZCI6ICJPVEhFUl9GVU5DVElPTiIsCiAgICAic3ludGF4IjogIkZVTkNUSU9OIgogIH0sCiAgIm9wZXJhbmRzIjogWwogICAgewogICAgICAiZHluYW1pY1BhcmFtIjogMCwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgfQogICAgfQogIF0KfQ==\"}","lang":"opensearch_compounded_script","params":{"utcTimestamp": 0,"SOURCES":[0],"DIGESTS":["gender.keyword"]}},"ignore_failure":false}}}}}}}}, requestedTotalSize=10000, pageSize=null, startFrom=0)])
```
LantaoJin, line 12: It's odd why this IT passed with `{"field":"state.keyword","missing_bucket":false,"order":"asc"}}}]}`

It should be `desc`

qianheng-aws: ASC is the default order. SORT is removed as it’s before DEDUP

PR #5009, opened 2025-12-31
Fix PIT context leak in Legacy SQL for non-paginated queries

### LINKED ISSUE #5002: [BUG] SQL legacy engine creates PIT unnecessarily and leak context
**What is the bug?**

When no cursor requested (`fetch_size` is not specified), the SQL Legacy engine creates Point-in-Time (PIT) contexts unnecessarily. Meanwhile, the PIT is also not closed properly, causing PIT contexts to leak and accumulate until they expire after the keep-alive timeout (default: 1 minute). Under repeated or concurrent query workloads, this can exhaust PIT capacity and hit the default limit of 300 open PIT contexts, resulting in query failures.

1. Create test index:
```bash
PUT /test-logs-2025-01-01
{
  "mappings": {
    "properties": {
      "action": {"type": "text", "fields": {"keyword": {"type": "keyword"}}},
      "timestamp": {"type": "date"}
    }
  }
}
```

2. Insert test data:
```bash
POST /test-logs-2025-01-01/_bulk
{"index":{}}
{"action":"login_success","timestamp":"2025-01-01T10:00:00Z"}
{"index":{}}
{"action":"login_success","timestamp":"2025-01-01T10:01:00Z"}
{"index":{}}
{"action":"login_failed","timestamp":"2025-01-01T10:02:00Z"}
```

3. Run SQL query (with backtick for V2 test, e.g., `` `test-logs-2025-01-01` `` ):
```bash
POST /_plugins/_sql
{
  "query": "SELECT * FROM test-logs-2025-01-01 WHERE action LIKE 'login%' ORDER BY timestamp ASC"
}
```

4. Check PIT contexts:
```bash
GET /_nodes/stats/indices/search?filter_path=nodes.*.indices.search.point_in_time_current
```

For queries that complete in a single page (no cursor requested / no follow-up page needed):

- Do not create PIT, or;
- If PIT is created internally, close/delete it before returning the response.

Note: The SQL/PPL V2 engine may also create PIT implicitly, but in that case it can be intentional: V2 may fetch until the last page and then close the PIT to ensure correctness for any required in-memory post-processing.

- OpenSearch Version: 2.19-dev, main branch
- Plugin: SQL plugin

N/A

- Legacy engine (SQL): always creates PIT aggressively when handling queries. The conditional check from 2.5 was removed.
    - Code: https://github.com/opensearch-project/sql/blob/main/legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java#L103
- V2 Engine (SQL/PPL): PIT is created when `from + size > maxResultWindow`. Request size is `Integer.MAX_VALUE` by default and size limit setting is not applied. The V2 engine closes PIT contexts by fetching all pages until an empty response is received, then deleting the PIT to ensure no PIT leak.
    - 2.5 code: https://github.com/opensearch-project/sql/blob/2.5/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java#L99
    - main code: https://github.com/opensearch-project/sql/blob/main/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchRequestBuilder.java#L63
- V3 Engine (Calcite - PPL only): doesn't have the issue because Calcite optimizer automatically applies a system size limit to every query. The system limit is read from a setting which is 10k by default.
    - Doc: https://github.com/opensearch-project/sql/blob/main/docs/user/admin/settings.rst#plugins-query-size-limit
    - Code: https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/executor/QueryService.java#L284

aalva500-prog:
### Description
This PR fixes Point-in-Time (PIT) context leak in the Legacy SQL engine when executing queries without `fetch_size` parameter.

**Problem:**
The Legacy SQL engine was creating PIT contexts for ALL queries but only cleaning them up when cursors were created (paginated queries with `fetch_size > 0`). Non-paginated queries leaked PITs, causing accumulation until the 300 PIT limit was exhausted and queries to fail.

**Solution:**
- Only create PIT when `fetch_size > 0` and not `null`  (pagination requested)

**Impact:**
- Non-paginated queries no longer leak PIT contexts
- Paginated queries continue to work correctly with cursor-based PIT management

### Related Issues
Resolves #5002

### Check List


integ-test/src/test/java/org/opensearch/sql/legacy/PointInTimeLeakIT.java
```diff
@@ -0,0 +1,225 @@
...
+        StringUtils.format(
+            "SELECT * FROM %s WHERE action LIKE 'login%%' ORDER BY timestamp ASC", TEST_INDEX);
+
+    JSONObject response = executeQueryWithFetchSize(query, 2);
+
+    assertTrue("Should have cursor with fetch_size", response.has("cursor"));
+    String cursor = response.getString("cursor");
+
+    JSONObject closeResponse = executeCursorCloseQuery(cursor);
+    assertTrue("Cursor close should succeed", closeResponse.getBoolean("succeeded"));
```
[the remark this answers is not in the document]
aalva500-prog: Wrong suggestion, there is no compilation error as the executeCursorCloseQuery method is not missing - it's already implemented in the parent class SQLIntegTestCase and is being correctly inherited by PointInTimeLeakIT.

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -91,41 +92,75 @@ public String execute(Client client, Map<String, String> params, QueryAction que
...
+    queryAction.explain();
 
-      protocol = new Protocol(client, queryAction, response.getHits(), format, defaultCursor);
-    } else {
-      protocol = new Protocol(client, queryAction, response.getHits(), format, Cursor.NULL_CURSOR);
+    int fetchSize = queryAction.getSqlRequest().fetchSize();
+    if (fetchSize > 0) {
+      pit = new PointInTimeHandlerImpl(client, queryAction.getSelect().getIndexArr());
+      pit.create();
+    }
```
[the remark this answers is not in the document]
aalva500-prog: Will take care of it

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -91,41 +92,75 @@ public String execute(Client client, Map<String, String> params, QueryAction que
...
 
     return protocol;
   }
 
   protected boolean isDefaultCursor(SearchResponse searchResponse, DefaultQueryAction queryAction) {
-    return queryAction.getSqlRequest().fetchSize() != 0
+    return queryAction.getSqlRequest().fetchSize() > 0
         && Objects.requireNonNull(searchResponse.getHits().getTotalHits()).value()
             >= queryAction.getSqlRequest().fetchSize();
   }
```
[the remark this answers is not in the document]
aalva500-prog: Will take care of it

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -91,42 +92,78 @@ public String execute(Client client, Map<String, String> params, QueryAction que
...
+      if (pit != null) {
+        try {
+          pit.delete();
+          pit = null; // Prevent double deletion in finally
+        } catch (RuntimeException cleanupException) {
+          LOG.error("Failed to delete PIT during exception handling", cleanupException);
+          Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
+        }
+      }
+      throw e;
```
Swiddis: issue (non-blocking): Why do we need this catch block? The finally block does the same thing?

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -91,42 +92,78 @@ public String execute(Client client, Map<String, String> params, QueryAction que
...
+    } finally {
+      // Cursor owns PIT lifecycle when created; otherwise clean up here
+      if (pit != null && !cursorCreated) {
+        try {
+          pit.delete();
+        } catch (RuntimeException e) {
+          LOG.error("Failed to delete PIT in finally block", e);
+          Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
+        }
```
Swiddis: issue (readability): It's not obvious how this logic avoids deleting the cursor for paginated queries.

`cursorCreated` could stand to be more clearly named, and all the cursor config could be split out to a smaller method to make it easier to follow this branching. `isDefaultCursor` isn't a very descriptive name either

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
aalva500-prog: I'll remove the catch block and leave the finally block, thanks!

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
aalva500-prog: Sure, will simplify the implementation and make it more readable, thanks!

integ-test/src/test/java/org/opensearch/sql/legacy/PointInTimeLeakIT.java
```diff
@@ -0,0 +1,231 @@
...
+              "SELECT * FROM %s WHERE action LIKE 'login%%' ORDER BY timestamp ASC", TEST_INDEX);
+
+      JSONObject response = executeQueryWithoutFetchSize(query);
+
+      assertTrue("Query should succeed", response.has("datarows"));
+      JSONArray dataRows = response.getJSONArray("datarows");
+      assertThat("Should return results", dataRows.length(), greaterThan(0));
+      assertFalse("Should not have cursor for non-paginated query", response.has("cursor"));
+
+      System.out.println(
```
dai-chen: Please remove all sysout

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -91,42 +92,78 @@ public String execute(Client client, Map<String, String> params, QueryAction que
...
-    SearchRequestBuilder searchRequest = queryAction.getRequestBuilder();
-    searchRequest.setPointInTime(new PointInTimeBuilder(pit.getPitId()));
-    response = searchRequest.get();
-
     Protocol protocol;
-    if (isDefaultCursor(response, queryAction)) {
-      DefaultCursor defaultCursor = new DefaultCursor();
-      defaultCursor.setLimit(queryAction.getSelect().getRowCount());
-      defaultCursor.setFetchSize(queryAction.getSqlRequest().fetchSize());
+    boolean cursorCreated = false;
```
dai-chen: Could you separate the logic in 2 methods instead of flow control by this flag?

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -91,42 +92,78 @@ public String execute(Client client, Map<String, String> params, QueryAction que
...
 
-      defaultCursor.setPitId(pit.getPitId());
-      defaultCursor.setSearchSourceBuilder(queryAction.getRequestBuilder().request().source());
-      defaultCursor.setSortFields(
-          response.getHits().getAt(response.getHits().getHits().length - 1).getSortValues());
+    queryAction.explain();
```
dai-chen, line 99: What is this used for?

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -91,42 +92,78 @@ public String execute(Client client, Map<String, String> params, QueryAction que
...
+      }
+    } catch (Exception e) {
+      if (pit != null) {
+        try {
+          pit.delete();
+          pit = null; // Prevent double deletion in finally
+        } catch (RuntimeException cleanupException) {
+          LOG.error("Failed to delete PIT during exception handling", cleanupException);
+          Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
+        }
```
dai-chen: Not necessary because finally block will executed anyway?

aalva500-prog: Yeah, I'm fixing it, thanks!

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
```diff
@@ -91,42 +92,78 @@ public String execute(Client client, Map<String, String> params, QueryAction que
...
-      defaultCursor.setSearchSourceBuilder(queryAction.getRequestBuilder().request().source());
-      defaultCursor.setSortFields(
-          response.getHits().getAt(response.getHits().getHits().length - 1).getSortValues());
+    queryAction.explain();
 
-      protocol = new Protocol(client, queryAction, response.getHits(), format, defaultCursor);
-    } else {
-      protocol = new Protocol(client, queryAction, response.getHits(), format, Cursor.NULL_CURSOR);
+    Integer fetchSize = queryAction.getSqlRequest().fetchSize();
+    if (fetchSize != null && fetchSize > 0) {
```
vamsimanohar, line 102: can you explain the logic here in words? What does fetchsize mean here?

vamsimanohar:
Have you also checked in other scenarios if the PIT contexts are closed properly

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
aalva500-prog: Sure, I'm working on it, thanks!

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
aalva500-prog: fetchsize is the size of the fetch_size parameter of the query, I'm only creating a PIT if the user specifies the fetch_size parameter and it is greater than zero. Before this, the PIT was always created regardless the user providing fetchsize in the query or not. My logic is to avoid creating PIT if `fetchSize` is 0 or not specified, meaning no pagination is needed and no PIT is created. Hope that makes sense.

aalva500-prog:
> Have you also checked in other scenarios if the PIT contexts are closed properly

@vamsimanohar The PIT context leak only happens in Legacy engine when the index name has special characters like `test-logs-2025.01.01`, it is not wrapped in backticks, and the fecth_size is not provided. It doesn't happen when the index name is like `test-logs-2025-01-01`, as it never falls to Legacy engine and it is handled by V2/V3. The IT class `PointInTimeLeakIT` should cover such scenarios, specifically this test case `testCompareV1AndV2EnginePitBehavior` compares V1 vs V2 by using backticks around the index so that the query is handle by V2 only. Note: I did manual testing in my own local cluster and AOS domain to confirm the same.

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
aalva500-prog: It was originally in the code, will take a further look and update here, thanks!

__Update:__ Ran the integration tests without this line and got failures in `CursorIT`, `PrettyFormatResponseIT`, `SQLCorrectnessIT`, `PaginationFallbackIT`, and `PaginationIT`.

According to my investigation, the call to `queryAction.explain()` is necessary because it builds the OpenSearch `SearchRequestBuilder` before executing the query. Without it, the `request` field in `DefaultQueryAction` remains null, which causes failures when `queryAction.getRequestBuilder()` is called later.

The original code was:

```java
SqlOpenSearchRequestBuilder sqlOpenSearchRequestBuilder = queryAction.explain();
```

Since the returned variable is never used, I simplified it to:

```java
queryAction.explain();
```

integ-test/src/test/java/org/opensearch/sql/legacy/PointInTimeLeakIT.java
```diff
@@ -0,0 +1,176 @@
...
+    return new JSONObject(TestUtils.getResponseBody(response));
+  }
+
+  private int getCurrentPitCount() throws IOException {
+    Request statsRequest = new Request("GET", PIT_STATS_ENDPOINT);
+    Response response = client().performRequest(statsRequest);
+    JSONObject stats = new JSONObject(TestUtils.getResponseBody(response));
+
+    int totalPits = 0;
+    if (stats.has("nodes")) {
```
dai-chen: np: I recall org.json supports JSON path. Could you check `query()` or `optQuery()` API?

legacy/src/main/java/org/opensearch/sql/legacy/executor/format/PrettyFormatRestExecutor.java
dai-chen: Thanks for the refactor. It looks much cleaner!

integ-test/src/test/java/org/opensearch/sql/legacy/PointInTimeLeakIT.java
aalva500-prog: Sure, I'll modify this method accordingly, thanks!

Issue #4984, opened 2025-12-22
[RFC] Support Schema-on-Read in PPL


ykmr1224:
## 1. Summary
This RFC proposes a solution for OpenSearch PPL to support schema-on-read data sources where fields and types are unknown until data reading begins. The proposal focuses on a long-term solution to support both schema-on-read and commands that generate fields dynamically, such as `spath`. The core approach introduces a field resolution phase to determine the input schema based on the AST before logical planning.

Related issues
- https://github.com/opensearch-project/sql/issues/4112 (motivation of this RFC)
- https://github.com/opensearch-project/sql/issues/4433 (This RFC will be deprecate)
- https://github.com/opensearch-project/sql/issues/4349

## 2. Motivation
The current PPL engine utilizes Apache Calcite for planning and execution, but Calcite requires a fixed schema during the planning stage. To handle the case where input schema is not known (or partially known), and also dynamic field generation commands, we need mechanism to handle query without schema.

Existing mechanisms in Calcite, such as `DynamicRecordType`, are insufficient for PPL due to several gaps:
* **Validation Stage Missing:** PPL lacks a query validation stage where `DynamicRecordType` typically collects fields.
* **Field Generation:** `DynamicRecordType` does not support commands that dynamically produce fields (e.g., `spath`, `multikv`).
* **Merge Conflicts:** PPL merges fields from multiple inputs by overwriting non-null values, whereas Calcite retains dynamic stars as separate fields.
* **Complexity:** Previous attempts to mix static and dynamic fields resulted in complicated planning logic.

## 3. Proposed Solution: Field Resolution Phase
To resolve the schema issues, we propose introducing a **Field Resolution** phase. This phase analyzes the Abstract Syntax Tree (AST) to decide the input schema before converting it to a Calcite `RelNode` tree.

### 3.1 The Process
The new execution flow is as follows:
1.  **Parse PPL:** Build AST from input PPL query.
2.  **Field Resolution (New):** Decide input schema based on AST.
    * Traverse AST from root to leaf.
    * Calculate required input fields for each data source based on output requirements.
3.  **Logical Plan:** Convert AST to Calcite `RelNode` tree using the resolved static schema.
4.  **Physical Plan & Execution:** Apply rules and execute.

### 3.2 Key Logic
* **Static Mapping:** By calculating requirements through the AST, fields used by commands are mapped as static fields in the data schema.
[...]

Issue #4960, opened 2025-12-15
[FEATURE] Reduce boilerplate for adding new PPL commands and functions


dai-chen:
**Is your feature request related to a problem?**

Adding new PPL commands or functions requires extensive boilerplate code across multiple layers of the codebase, making it error-prone and time-consuming. Developers must manually:

1. Update ANTLR grammar files (lexer and parser)
2. Create AST node classes and implement AST/RelNode visitor pattern methods
3. Register functions in multiple locations (PPLBuiltinOperators, PPLFuncImpTable, BuiltinFunctionName)
4. Create multiple test classes (IT, Yaml IT, ExplainIT, AnonymousIT, CrossClusterIT)
5. Remember engine-specific setup (e.g., `enableCalcite()` for V3-only functions)
6. Update documentation

This repetitive process leads to:
- **Human errors**: Easy to miss required steps or files - some are subtle even if reference similar PRs.
- **Inconsistency**: Different developers may structure code differently.
- **Slow onboarding**: New contributors face a steep learning curve.
- **Maintenance burden**: Changes to architecture require updating many files.

**What solution would you like?**

Below are several ideas that address the problem from different levels (tooling, specs, and architecture):

- **Interactive scaffolding tool**: Add a Gradle task that generates the required files and code insertions from a few prompts, with a strict “expected files changed” checklist to prevent missed steps (human or AI) and copy-paste drift.

- **Spec-driven code generation**: Define a YAML/Markdown spec as the single source of truth for commands/functions (metadata, signatures, engine support, tests/docs) and generate the repetitive glue at build time—AI-friendly by design.

- **Better abstractions**: Introduce a small DSL for AST → RelNode translation and shared test base classes/mixins to standardize setup. This reduces long-term maintenance and makes both scaffolding and codegen simpler and safer.

**What alternatives have you considered?**

1. Better documentation only: Document the process better, but doesn't prevent human errors
2. Copy-paste from examples: Current approach, leads to inconsistencies and forgotten steps
3. Code review checklists: Helpful but reactive - catches errors after they're made

**Do you have any additional context?**

1. This issue was motivated by #4947, where a backport PR failed because the integration test was missing `enableCalcite()`.
2. Success metrics may include:
    - Time to add new PPL command/function reduced by X%+
    - New contributors can onboard and ship first PR within X days
    - Consistent code structure across all commands/functions

Swiddis:
[...]

PR #5028, opened 2026-01-07
Implement spath command with field resolution

ykmr1224:
### Description
- Implement spath command with field resolution
- This change is for step1 (spath limited version) in https://github.com/opensearch-project/sql/issues/4984
- Identify referred fields from query and extract only the referred fields from JSON.
  - Raise error when referred fields cannot be identified (like when no `fields` command) or wildcard is used.
- All the fields are extracted as STRING regardless of the actual value type in JSON
  - Later it would be extended to ANY, but that requires functions to handle ANY well.
- Originally planned to use `mvappend` function, but added new internal `append` function to avoid returning single element array. (`mvappend` will always make array regardless of array size, this is consistent with other array functions in PPL. `append` will return scalar value when the result become single element, this is similar to Splunk behavior where single element array is automatically converted to scalar value.)
  - This prevents fields not merged by `spath` command becoming an array.

### Related Issues
https://github.com/opensearch-project/sql/issues/4984

### Check List

### DIFF
--- a/docs/user/ppl/cmd/spath.md
+++ b/docs/user/ppl/cmd/spath.md
@@ -256,3 +256,37 @@ source=structured
 source=structured
 | spath input=doc_multi
 | fields a, b*  # ERROR: Spath command cannot extract arbitrary fields
+```
+
+## Example 10: Performance Considerations
+
+**Important**: The `spath` command processes data on the coordinator node after retrieval from data nodes. Commands placed after `spath` cannot utilize OpenSearch index capabilities, which significantly impacts performance on large datasets.
+
+### Best Practice: Filter Before spath
+
+Always place filter conditions (`where` clauses) **before** the `spath` command to leverage index pushdown:
+
+```ppl
+# Slow - filters after spath
+source=structured
+| spath input=doc_multi
+| where id=1
+| fields id, a
+
+# Fast - filters before spath
+source=structured
+| where id=1
+| spath input=doc_multi
+| fields id, a
+```
+
+### Performance Impact
+
+- **After spath**: All data is retrieved from OpenSearch, then filtered in memory on the coordinator
+- **Before spath**: Only matching documents are retrieved from OpenSearch utilizing index
+
+### Recommendations
+
+1. **Index nested fields directly** when possible instead of using `spath` for filtering
+2. **Place all filters before `spath`** to maximize index utilization
+3. **Limit result sets** with filters before applying `spath` on large datasets


docs/user/ppl/cmd/spath.md
```diff
@@ -2,14 +2,28 @@
...
 * output: optional. The destination field that the data will be loaded to. **Default:** value of `path`.  
-* path: mandatory. The path of the data to load for the object. For more information on path syntax, see [json_extract](../functions/json.md#json_extract).  
-  
+* path: mandatory. The path of the data to load for the object. For more information on path syntax, see [json_extract](../functions/json.md#json_extract).
+
+### Field Resolution-based Extraction (Experimental)
+spath input=\<field\>
+* input: mandatory. The field containing JSON data to extract from.
+* Extracts fields based on downstream requirements
+* **Limitation**: It raises error if extracted fields cannot be identified by following commands (i.e. `fields`, or `stats` command is needed)
```
penghuo, line 39: Are these limitations will be addressed in following PR?

docs/user/ppl/cmd/spath.md
```diff
@@ -2,14 +2,28 @@
...
 
+### Path-based Extraction
 spath input=\<field\> [output=\<field\>] [path=]\<path\>
 * input: mandatory. The field to scan for JSON data.  
 * output: optional. The destination field that the data will be loaded to. **Default:** value of `path`.  
-* path: mandatory. The path of the data to load for the object. For more information on path syntax, see [json_extract](../functions/json.md#json_extract).  
-  
+* path: mandatory. The path of the data to load for the object. For more information on path syntax, see [json_extract](../functions/json.md#json_extract).
+
+### Field Resolution-based Extraction (Experimental)
```
penghuo, line 18: `Field Resolution-based` is a solution, not user facing feature? Or we want to tell user spath will auto extract fields reference in the query.

docs/user/ppl/cmd/spath.md
```diff
@@ -107,4 +121,97 @@ fetched rows / total rows = 3/3
 | false | 2 |
 +-------+---+
 ```
-  
\ No newline at end of file
+
+## Example 5: Field Resolution-based Extraction
```
penghuo, line 142: Add a test case to demo field merge capability. 
`{"a.b", 1, "a":{"b":2}}`, results will be `"a.b": [1, 2]`

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
```diff
@@ -206,4 +212,24 @@ public RexLambdaRef captureVariable(RexNode fieldRef, String fieldName) {
...
+  /**
+   * Resolves required fields for a target node in the PPL query plan by analyzing the AST from
+   * root. Used for schema-on-read features like `spath` command.
+   *
+   * @param target the plan node to resolve field requirements for
+   * @return field resolution result with regular fields and wildcard patterns
+   * @throws IllegalStateException if root node not set via {@link #setRootNode}
+   */
+  public FieldResolutionResult resolveFields(UnresolvedPlan target) {
+    if (rootNode == null) {
```
penghuo, line 225: error message will return to user? it is not actionable.

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
```diff
@@ -206,4 +212,24 @@ public RexLambdaRef captureVariable(RexNode fieldRef, String fieldName) {
...
+
+  /**
+   * Resolves required fields for a target node in the PPL query plan by analyzing the AST from
+   * root. Used for schema-on-read features like `spath` command.
+   *
+   * @param target the plan node to resolve field requirements for
+   * @return field resolution result with regular fields and wildcard patterns
+   * @throws IllegalStateException if root node not set via {@link #setRootNode}
+   */
+  public FieldResolutionResult resolveFields(UnresolvedPlan target) {
```
penghuo, line 224: This function not use CalcitePlanContext info. Could fit into analyzer module.

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
```diff
@@ -0,0 +1,624 @@
...
+  private void acceptAndVerifyNodeVisited(Node node, FieldResolutionContext context) {
+    Node result = node.accept(this, context);
+    if (result != node) {
+      throw new IllegalArgumentException(
+          "Unsupported command for field resolution: " + node.getClass().getSimpleName());
+    }
+  }
+
+  @Override
+  public Node visitProject(Project node, FieldResolutionContext context) {
```
penghuo, line 112: Consider extracting per‑command “required field” logic to reducing duplication when new commands are added.

docs/user/ppl/cmd/spath.md
ykmr1224: Yes, do we want to note those will be addressed in near future? (I suppose those would be in later version)

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
ykmr1224: It would be 5xx error as it throws IllegalStateException.

docs/user/ppl/cmd/spath.md
ykmr1224: Yes, it is actually more like a internal solution to realize `spath` command.
As it is interim solution until the dynamic fields are added in the next step, I'd like to keep it as it is.
Once dynamic fields are added, we don't need to mention about `field resolution-based extraction`.

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
ykmr1224: Per‑command “required field” logic depends on the command logic, which specific part do you expect to be extracted?

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
ykmr1224: This function refers `rootNode` in CalcitePlanContext, and that's why I put it here.

docs/user/ppl/cmd/spath.md
ykmr1224: Added example.

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
penghuo: add [BUG] as prefix?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLSpathCommandIT.java
```diff
@@ -17,32 +18,155 @@
...
 
-    loadIndex(Index.BANK);
+    putItem(1, "simple", sj("{'a': 1, 'b': 2, 'c': 3}"));
+    putItem(2, "simple", sj("{'a': 1, 'b': 2, 'c': 3}"));
+    putItem(3, "nested", sj("{'nested': {'d': [1, 2, 3], 'e': 'str'}}"));
+    putItem(4, "join1", sj("{'key': 'k1', 'left': 'l'}"));
+    putItem(5, "join2", sj("{'key': 'k1', 'right': 'r1'}"));
+    putItem(6, "join2", sj("{'key': 'k2', 'right': 'r2'}"));
+    putItem(7, "overwrap", sj("{'a.b': 1, 'a': {'b': 2, 'c': 3}}"));
+  }
```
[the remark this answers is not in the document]
ykmr1224: @coderabbitai It is cleaned by parent class. Don't raise similar comment.

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
```diff
@@ -0,0 +1,624 @@
...
+import org.opensearch.sql.common.patterns.PatternUtils;
+import org.opensearch.sql.expression.parse.PatternsExpression;
+import org.opensearch.sql.expression.parse.RegexCommonUtils;
+
+/**
+ * Visitor to analyze and collect required fields from PPL AST using stack-based traversal. The
+ * result is used to support `spath` command (Field Resolution-based Extraction) and schema-on-read
+ * support in the future.
+ */
+public class FieldResolutionVisitor extends AbstractNodeVisitor<Node, FieldResolutionContext> {
```
dai-chen, line 76: Not sure if any better way to extract the input/output fields for each operator. But we can enforce by adding input/output to AST interface? I'm just concerned about the complexity of adding new PPL feature: https://github.com/opensearch-project/sql/issues/4960

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
penghuo: When new AST node introduced, it need to add field resolve logic FieldResolutionVisitor. 
One workaround is and define a interface on UnresolvedPlan to resolve field, then each UnresolvedPlan need to implement it.

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
ykmr1224: I think you mean adding a method to Node or UnresolvedPlan. That is an option, but has following downside:
- AST nodes will have implementation specific methods, though AST is abstraction of PPL/SQL query structure.
- We have several nodes which is not used by PPL, we will need to implement them. I suppose we would put dummy implementation as they won't be used for now, and we could forget to update them when we start using that node.

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
ykmr1224: I wanted to avoid that as answered at https://github.com/opensearch-project/sql/pull/5028#discussion_r2687492605

core/src/main/java/org/opensearch/sql/calcite/CalcitePlanContext.java
ykmr1224: 5xx error indicates service issue, and we shouldn't need to tag it as `[BUG]`. We already have bunch of codes to throw IllegalStateException with `Failed to xxxx` message.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -2321,6 +2321,13 @@ public void testNotBetweenPushDownExplain() throws Exception {
             "source=opensearch-sql_test_index_bank | where age not between 30 and 39"));
   }
 
+  @Test
+  public void testSpathWithoutPathExplain() throws IOException {
+    String expected = loadExpectedPlan("explain_spath_without_path.yaml");
+    assertYamlEqualsIgnoreId(
+        expected, explainQueryYaml(source(TEST_INDEX_LOGS, "spath input=message | fields test")));
```
penghuo, line 2328: With spath command, the query performance will be difference, let's call-out in doc.
e.g., bytes is a index field. msg is text field include json.
```
## slow
source=log00001 | spath input=msg | where bytes=100 | fields bytes

## fast bytes=100 pushdown as term query
source=log00001 | where bytes=100 | spath input=msg | fields bytes
```

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
penghuo: what if traverse RelNode 2 times? at least RelNode not change quite a lot.

dai-chen: @ykmr1224 No, I'm thinking add contract on AST node to enforce each subclass to implement it. Could you check `SqlNode.getOperandList()` or `SqlValidatorScope.findAllColumnNames()`? Is it similar as what we want?

ykmr1224: > @penghuo

Traverse RelNode twice does not work well, because we need to decide the field list before building RelNode tree during the first path. RelNode requires fixed schema, and this is why I proposed field resolution phase to decide the schema before building RelNodes.

> @dai-chen 

AST node is represented by subclass of [UnresolvedPlan](https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/ast/tree/UnresolvedPlan.java#L16) / [Node](https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/ast/Node.java#L15).
So I thought the idea is adding unimplemented method there. Then each node such as [Project](https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/ast/tree/Project.java) needs to implement it. Though the downside is as described above.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
ykmr1224: Fixed.

dai-chen (approved):

docs/dev/ppl-commands.md
```diff
@@ -27,6 +27,7 @@ If you are working on contributing a new PPL command, please read this guide and
 - [ ] **Visitor Pattern:**
   - Add `visit*` in `AbstractNodeVisitor`
   - Overriding `visit*` in `Analyzer`, `CalciteRelNodeVisitor` and `PPLQueryDataAnonymizer`
+  - Override `visit*` in `FieldResolutionVisitor` for `spath` command support.
```
LantaoJin, line 30: @ykmr1224 can you add an IT (or guide how to add an necessary IT) to ensure new commands can work well with `spath`.
Not sure how does a new command impact `spath`, of else developer has to add following code in FieldResolutionVisitor to pass the UT CI.
```
  @Override
  public Node visit*(Transpose node, FieldResolutionContext context) {
    visitChildren(node, context);
    return node;
  }
```

ykmr1224: FieldResolutionVisitorCoverageTest will detect if visitXXX is implemented in CalciteRelNodeVisitor but not implemented in FieldResolutionVisitor (this works as safe guard to forget implement it).
Let me add a implementation guide doc based on https://github.com/opensearch-project/sql/issues/4984 in a follow-up PR, but it might be better delayed until fully implement schema-on-read.

PR #5029, opened 2026-01-08
Make V2 also reference plugins.query.buckets

Swiddis:
### Description
Straightforward update to use the setting in V2 as we do in Calcite. Lots of test changes though :/

### Related Issues
N/A

### Check List


integ-test/src/test/java/org/opensearch/sql/util/MatcherUtils.java
Swiddis, line 1: bonus: yaml tests no longer whitespace sensitive, since i had to rewrap a bunch to get these integ tests to pass

Swiddis: nvm that just causes infinitely more failures :/

integ-test/src/test/resources/expectedOutput/ppl/clickbench/q15.yaml
```diff
@@ -18,7 +18,7 @@ root:
...
               \"}\",\"lang\":\"opensearch_compounded_script\"},\"boost\":1.0}},\"\
-              aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":1000,\"\
+              aggregations\":{\"composite_buckets\":{\"composite\":{\"size\":10000,\"\
```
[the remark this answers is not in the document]
Swiddis: @coderabbitai The 10k value comes from the default bucket setting and various places in the integration test suite where the value gets reset, your `rg` search found cases where the value was being updated to 1000 and those are the files that are the same. In the test logs you can see lines like `updating [plugins.query.buckets] from [1000] to [10000]`.

When the value isn't explicitly set, the chain is:

1. index.max_result_window = 10000 (OpenSearch default)
2. plugins.query.size_limit defaults to index.max_result_window = 10000
3. plugins.query.buckets defaults to plugins.query.size_limit = 10000

PR #5042, opened 2026-01-13
Separate explain mode from format params

qianheng-aws:
### Description
Introduce a new query parameter `mode`(allows `SIMPLE`, `STANDARD`, `EXTENDED`, `COST`) specifically for controlling the explain verbosity, while keeping format strictly for output serialization format(allows `JSON`, `YAML` for explain API or command).

e.g.
```
POST .../_plugins/_ppl/_explain?mode=extended&format=yaml
{
"query": "source=..."
}
```
Before this PR, when setting `format=yaml`, users cannot specify the explain mode anymore with the explain mode bounded to the default `STANDARD`.

For backward compatibility, it also supports `format=[simple, standard, extended, cost]` and get the same behavior as before

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/4940

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/request/PPLQueryRequestFactory.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/request/PPLQueryRequestFactory.java
@@ -7,6 +7,8 @@ package org.opensearch.sql.plugin.request;
 
 import java.util.Map;
 import java.util.Optional;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.opensearch.rest.RestRequest;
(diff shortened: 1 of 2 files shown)


docs/user/ppl/interfaces/endpoint.md
```diff
@@ -61,7 +61,8 @@ You can send HTTP explain request to endpoint **/_plugins/_ppl/_explain** with y
 ### Description  
 
 To translate your query, send it to explain endpoint. The explain output is OpenSearch domain specific language (DSL) in JSON format. You can just copy and paste it to your console to run it against OpenSearch directly.
-Explain output could be set different formats: `standard` (the default format), `simple`, `extended`, `dsl`.
+Explain API supports various mode: `standard` (the default format), `simple`, `extended`, `cost`.
```
penghuo, line 64: +1, mode is more clear then format. 
Q, PPL should support format=[simple, extended, cost] until 4.x, then we can deprecated it, right? If yes, we should add deprecated warnning info in resposne.

qianheng-aws: Do you mean we should keep backward compatible? The current code won't support `format=[simple, extended, cost]` anymore.

LantaoJin: +1 to add `format=[simple, extended, cost]` as alias of `mode=[simple, extended, cost]` before 4.x and clarify them deprecated, even they are not wildly used, but as a policy of open source community, we don't want to introduce API level breaking changes.

qianheng-aws: Addressed in https://github.com/opensearch-project/sql/pull/5042/commits/17e9334ed199c74ac248aab27869ed47a4147ea6, `format=[simple, standard, extended, cost]` is supported as well.

yuancu (approved):
LGTM

PR #5058, opened 2026-01-20
Support spath with dynamic fields

ykmr1224:
### Description
- Extend spath command to support dynamic fields, where it no longer requires field selection.
  - It has limitations as it does not fully support field ordering of dynamic fields.
- This change is for step2-1 in https://github.com/opensearch-project/sql/issues/4984
- The fields not explicitly referred in the query would be stored in `_MAP`(dynamic fields), and expanded when result is organized.
- It does not support partial wildcard like `prefix*`, `*suffix`, and raise error in case those are used.
- When result contains dynamic fields, OpenSearchExecutionEngine puts dynamic fields at the end of the fields when organizing result. (This is temporal behavior until field ordering is implemented)
  - To align this behavior, it raises error when `fields` command specify `*` in the middle of field list.
- Following commands will be adopted in later PR (command with multiple input requires change) : `appendcol`, `multisearch`, `lookup`

### Related Issues
https://github.com/opensearch-project/sql/issues/4984

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/DynamicFieldsHelper.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/DynamicFieldsHelper.java
@@ -112,13 +112,9 @@ class DynamicFieldsHelper {
   /** Adjust fields to align the static/dynamic fields in `target` to `theOtherInput` */
   static RelNode adjustFieldsForDynamicFields(
       RelNode target, RelNode theOtherInput, CalcitePlanContext context) {
-    if (hasDynamicFields(theOtherInput)) {
+    if (hasDynamicFields(theOtherInput) && !hasDynamicFields(target)) {
       List<String> requiredStaticFields = getStaticFields(theOtherInput);
-      if (!hasDynamicFields(target)) {
-        return adjustFieldsForDynamicFields(target, requiredStaticFields, context);
-      } else {
-        return addRequiredStaticFields(target, requiredStaticFields, context);
-      }
+      return adjustFieldsForDynamicFields(target, requiredStaticFields, context);
     }
     return target;
   }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/opensearch/sql/calcite/utils/JoinAndLookupUtils.java
```diff
@@ -96,12 +103,57 @@ static void renameToExpectedFields(
...
+   */
+  static void mergeDynamicFieldsAsNeeded(CalcitePlanContext context, OverwriteMode overwriteMode) {
+    List<String> fieldNames = context.relBuilder.peek().getRowType().getFieldNames();
+    List<RexNode> fields = context.relBuilder.fields();
+    List<RexNode> dynamicFieldMaps = new ArrayList<>();
+    List<RexNode> result = new ArrayList<>();
+    for (int i = 0; i < fieldNames.size(); i++) {
+      String fieldName = fieldNames.get(i);
+      RexNode field = fields.get(i);
+      if (fieldName.startsWith(DYNAMIC_FIELDS_MAP)) {
```
penghuo, line 133: 1. change startsWith to equals?
2. document _MAP is reserved keyword of PPL.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -744,20 +756,119 @@ private RelNode spathExtractAll(SPath node, CalcitePlanContext context) {
                 BuiltinFunctionName.INTERNAL_APPEND,
                 context.relBuilder.field(fieldName),
                 item);
+        item = castToString(item, context);
       }
       fields.add(context.relBuilder.alias(item, fieldName));
     }
 
+    // 3. Add _MAP field for dynamic fields when wildcards present
+    if (resolutionResult.hasWildcards()) {
```
penghuo, line 751: Can we delay the wildcard handling in DynamicFieldsResultProcessor?

ykmr1224: It is needed to put everything else (fields not explicitly referred in following commands) into dynamic fields. By aligning the fields in static fields and dynamic fields, we can avoid complicated process to handle both static & dynamic fields in most of the commands. And join can simply append two maps since required fields are always in static fields, and others are in dynamic fields (static/dynamic fields handle different set of fields, and don't require cross check or dedupe between static/dynamic fields)

The logic here:
0. (before here) required static fields are decided by FieldResolutionVisitor
1. Make `_MAP` with non-required fields (exclude required fields from the map made from JSON
2. Add existing non-required fields to the `_MAP`
3. Merge `_MAP` with existing `_MAP`

core/src/main/java/org/opensearch/sql/calcite/utils/JoinAndLookupUtils.java
ykmr1224: 1 -> It needs to be `startsWith` since Calcite automatically rename field to `_MAP0` when both input contains `_MAP`.
2 -> Added doc description in identifiers.md

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionResult.java
```diff
@@ -100,8 +110,10 @@ private Set<String> and(Set<String> fields) {
         .collect(Collectors.toSet());
   }
 
+  /** Creates new result intersecting this result with another (intersection). */
   public FieldResolutionResult and(FieldResolutionResult other) {
```
dai-chen, line 114: np: this API design is different from others which accept set of fields? Is it where the complexity coming from?

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
```diff
@@ -490,43 +493,56 @@ public Node visitStreamWindow(StreamWindow node, FieldResolutionContext context)
...
+    context.popRequirements();
     return node;
   }
 
   @Override
   public Node visitAppendCol(AppendCol node, FieldResolutionContext context) {
-    visitChildren(node, context);
-    return node;
+    throw new IllegalArgumentException(
+        "AppendCol command cannot be used together with spath command");
```
dai-chen, line 511: This visitor is not generic and only for spath?

core/src/main/java/org/opensearch/sql/ast/tree/FillNull.java
```diff
@@ -63,6 +63,10 @@ public List<Field> getFields() {
     return getReplacementPairs().stream().map(Pair::getLeft).toList();
   }
 
+  public boolean isAgainstAllFields() {
```
dai-chen, line 66: I'm not clear when we should call this? Any PPL command doesn't allow `*`?

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionVisitor.java
ykmr1224: The error message is specific for `spath` command since we currently use it only for `spath` command (I made it specific for user to understand easier). Once we generalize it for schema-on-read, we need to change the message to be generic.

core/src/main/java/org/opensearch/sql/ast/tree/FillNull.java
ykmr1224: When `fillnull` command is used without field list, it applies to all the fields. This method is used for dynamic fields to raise error since we cannot support it.

core/src/main/java/org/opensearch/sql/ast/analysis/FieldResolutionResult.java
ykmr1224: This method is for when wildcard is specified, and it can utilize the constructors we already have.
(Simply for when taking `and` with fields + wildcards)

PR #5109, opened 2026-02-04
Support fetch_size API for PPL

### LINKED ISSUE #3314: [BUG] fetch_size does not work when sending PPL request
**What is the bug?**
According to documentation, [https://opensearch.org/docs/latest/search-plugins/sql/sql-ppl-api/#request-body-fields](https://opensearch.org/docs/latest/search-plugins/sql/sql-ppl-api/#request-body-fields), `fetch_size` should control the size of the responses we got back for both PPL and SQL requests. However, currently when i send requests with `fetch_size` in the dev tool, only SQL works; PPL responses do not seem to respect the fetch_size param and still return the full size responses. 

1. Go to dev tool
2. type 
```
POST _plugins/_ppl/
{
  "fetch_size" : 1,
  "query" : "source = opensearch_dashboards_sample_data_flights",
}
```

3. Send the request
4. See responses return with full size, but not size 1

 - Version 2.18

https://github.com/user-attachments/assets/1f9b329e-a229-474f-9813-61a3bfa28d19

ahkcs:
<html><head></head><body><h2>Summary</h2>
<p>Add <code>fetch_size</code> parameter to the PPL REST API (<code>POST /_plugins/_ppl</code>) that limits the number of rows returned in a single response.</p>
<ul>
<li>When <code>fetch_size</code> is specified, a <code>Head</code> AST node is injected on top of the query plan, which pushes the limit down to OpenSearch — so only the requested number of documents are fetched, not the full result set</li>
<li>Unlike SQL's <code>fetch_size</code> which enables cursor-based pagination, PPL's <code>fetch_size</code> returns a single complete response with no cursor</li>
</ul>
<p>Resolves #3314</p>
<h2>Design</h2>
<p>When <code>fetch_size &gt; 0</code>, we inject <code>Head(fetchSize, 0)</code> before <code>addSelectAll</code> in <code>AstStatementBuilder.visitPplStatement()</code>. This is equivalent to appending <code>| head N</code> to the query at the AST level.</p>
<p>The existing optimizer handles limit pushdown to OpenSearch for both the Calcite and Legacy execution paths — no changes to the execution engine are needed.</p>
<h2>API Example</h2>
<pre><code class="language-json">POST /_plugins/_ppl
{
  "fetch_size": 500,
  "query": "source = my_index"
}
</code></pre>
<h2>Constraints</h2>
<ul>
<li><code>fetch_size</code> must be between <strong>1</strong> and <strong>10,000</strong></li>
<li>A value of <code>0</code> or omitting the parameter uses the system default (current behavior)</li>
</ul>
<h2>Interaction with <code>head</code> command</h2>
<p>The smaller limit always wins:</p>

fetch_size | Query | Rows returned
-- | -- | --
5 | ... \| head 100 | 5
10 | ... \| head 3 | 3

</body></html>

### Related Issues
Resolves #3314 

### Check List

### DIFF
--- a/docs/user/interfaces/endpoint.rst
+++ b/docs/user/interfaces/endpoint.rst
@@ -279,19 +279,20 @@ The ``fetch_size`` parameter limits the number of rows returned in a PPL query r
 
 If ``fetch_size`` is larger than ``plugins.query.size_limit``, the result is capped at ``plugins.query.size_limit``. The effective number of rows returned is always ``min(fetch_size, plugins.query.size_limit)``.
 
-.. note::
+Note
+----
 
-   Unlike SQL's ``fetch_size`` which enables cursor-based pagination, PPL's ``fetch_size`` does not return a cursor and does not support fetching additional pages. The response is always complete and final.
+Unlike SQL's ``fetch_size`` which enables cursor-based pagination, PPL's ``fetch_size`` does not return a cursor and does not support fetching additional pages. The response is always complete and final.
 
-   +--------------------+-------------------------------------+------------------------------------+
-   | Aspect             | SQL ``fetch_size``                  | PPL ``fetch_size``                 |
-   +====================+=====================================+====================================+
-   | Purpose            | Cursor-based pagination             | Response size limiting             |
-   +--------------------+-------------------------------------+------------------------------------+
-   | Returns cursor?    | Yes                                 | No                                 |
-   +--------------------+-------------------------------------+------------------------------------+
-   | Can fetch more?    | Yes (with cursor)                   | No (single response)               |
-   +--------------------+-------------------------------------+------------------------------------+
++--------------------+-------------------------------------+------------------------------------+
+| Aspect             | SQL ``fetch_size``                  | PPL ``fetch_size``                 |
++====================+=====================================+====================================+
+| Purpose            | Cursor-based pagination             | Response size limiting             |
++--------------------+-------------------------------------+------------------------------------+
+| Returns cursor?    | Yes                                 | No                                 |
++--------------------+-------------------------------------+------------------------------------+
+| Can fetch more?    | Yes (with cursor)                   | No (single response)               |
++--------------------+-------------------------------------+------------------------------------+
 
 Example 1: JSON body
 -------


ppl/src/main/java/org/opensearch/sql/ppl/domain/PPLQueryRequest.java
```diff
@@ -18,6 +18,8 @@
 public class PPLQueryRequest {
 
   private static final String DEFAULT_PPL_PATH = "/_plugins/_ppl";
+  private static final String FETCH_SIZE_FIELD = "fetch_size";
+  private static final int MAX_FETCH_SIZE = 10000;
```
Swiddis: suggestion: This limit could probably be based on the index.max_result_window setting. https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/index-settings/#dynamic-index-level-index-settings

If we have this 10k limit artificially defined for our plugin elsewhere, we should reuse that constant instead of making a new one (it might just be the max size setting).

core/src/main/java/org/opensearch/sql/ast/statement/Query.java
```diff
@@ -19,13 +18,18 @@
 @Setter
 @ToString
 @EqualsAndHashCode(callSuper = false)
-@RequiredArgsConstructor
```
Swiddis, line 22: polish: Why switch to an explicit constructor?

ppl/src/main/java/org/opensearch/sql/ppl/domain/PPLQueryRequest.java
```diff
@@ -93,4 +95,24 @@ public Format format() {
...
   }
+
+  /**
+   * Get the maximum number of results to return. Unlike SQL's fetch_size which enables cursor-based
+   * pagination, PPL's fetch_size simply limits the response to N rows without cursor support.
+   *
+   * @return fetch_size value from request, or 0 if not specified (meaning use system default)
+   * @throws IllegalArgumentException if fetch_size exceeds MAX_FETCH_SIZE (10000)
+   */
+  public int getFetchSize() {
```
Swiddis, line 106: suggestion: is might also be worth supporting a `size` url parameter?

Similar to how core lets you supply both `?size=n` or `"size": n` in requests. Also similar to how we support specifying response formats in both the URL and the body.

core/src/main/java/org/opensearch/sql/ast/statement/Query.java
ahkcs: Reverted

ppl/src/main/java/org/opensearch/sql/ppl/domain/PPLQueryRequest.java
ahkcs: Good catch. Removed the hardcoded MAX_FETCH_SIZE = 10000 — the effective cap is already enforced by LogicalSystemLimit, which uses the plugins.query.size_limit cluster setting. This way it also respects custom configurations.

ppl/src/main/java/org/opensearch/sql/ppl/domain/PPLQueryRequest.java
ahkcs: Added. fetch_size is now supported as a URL parameter (e.g., POST /_plugins/_ppl?fetch_size=5)

docs/user/interfaces/endpoint.rst
```diff
@@ -266,3 +266,58 @@ Result set::
       "size": 5,
       "status": 200
     }
+
+Fetch Size (PPL)
```
penghuo: Mark as experimental. We can target prod at 3.7 release.

docs/user/interfaces/endpoint.rst
```diff
@@ -266,3 +266,58 @@ Result set::
...
+================
+
+Description
+-----------
+
+PPL also supports the ``fetch_size`` parameter, but with different semantics from SQL. In PPL, ``fetch_size`` limits the number of rows returned in a single, complete response. **PPL does not support cursor-based pagination** — no cursor is returned and there is no way to fetch additional pages. The value of ``fetch_size`` should be greater than ``0``. In absence of ``fetch_size`` or a value of ``0``, it will use the system default behavior (no limit). The effective upper bound is governed by the ``plugins.query.size_limit`` cluster setting (defaults to ``index.max_result_window``, which is 10000).
```
penghuo: Repharse. 
1. Introdroduce what is PPL fetch_sisze in description section.
2. Move PPL and SQL difference as a notes


> system default behavior (no limit)

not correct. plugins.query.size_limit is the limit.

> index.max_result_window

this setting is not visible to PPL user.

docs/user/ppl/limitations/limitations.md
```diff
@@ -62,7 +62,7 @@ For the following functionalities, the query will be forwarded to the V2 query e
     * ML  
     * Kmeans  
 * `show datasources` and command  
-* Commands with `fetch_size` parameter  
+* SQL commands with `fetch_size` parameter (cursor-based pagination). Note: PPL's `fetch_size` (response size limiting, no cursor) is supported in Calcite Engine.
```
penghuo: What is SQL commands? Rephrase it.

integ-test/src/test/java/org/opensearch/sql/ppl/FetchSizeIT.java
```diff
@@ -0,0 +1,234 @@
...
+
+  @Test
+  public void testFetchSizeJsonBodyTakesPrecedenceOverUrlParam() throws IOException {
+    // JSON body fetch_size=3 should take precedence over URL param fetch_size=10
+    Request request = new Request("POST", QUERY_API_ENDPOINT + "?fetch_size=10");
+    String jsonBody =
+        String.format(
+            Locale.ROOT,
+            "{\n  \"query\": \"source=%s | fields firstname\",\n  \"fetch_size\": 3\n}",
+            TEST_INDEX_ACCOUNT);
```
penghuo, line 221: 1. Any reason we support query paramaters and query body?
2. It is not documented in API doc.

docs/user/interfaces/endpoint.rst
```diff
@@ -266,3 +266,58 @@ Result set::
...
+| Can fetch more?    | Yes (with cursor)                   | No (single response)               |
++--------------------+-------------------------------------+------------------------------------+
+
+Example
+-------
+
+PPL query::
+
+	>> curl -H 'Content-Type: application/json' -X POST localhost:9200/_plugins/_ppl -d '{
+	  "fetch_size" : 5,
```
penghuo, line 303: what if fetch_size larger than  plugins.query.size_limit . which one should follow?

docs/user/interfaces/endpoint.rst
ahkcs: Updated

docs/user/ppl/limitations/limitations.md
ahkcs: Updated to `SQL queries`

integ-test/src/test/java/org/opensearch/sql/ppl/FetchSizeIT.java
ahkcs: URL parameter support was added per @Swiddis's suggestion — https://github.com/opensearch-project/sql/pull/5109#discussion_r2765270495. 
For documentation, `endpoint.rst` covers this: "fetch_size can be specified either as a URL parameter or in the JSON request body. If both are provided, the JSON body value takes precedence." The Example 2 section also demonstrates the URL parameter syntax.

docs/user/interfaces/endpoint.rst
ahkcs: Since our implementation for PPL `fetch_size` API is essentially appending a `HEAD` command at the end of the query, so if `fetch_size` is larger than `plugins.query.size_limit`, it would behave the same as using a `HEAD` command larger than `plugins.query.size_limit`, which would follow the cap set by `plugins.query.size_limit`

docs/user/interfaces/endpoint.rst
ahkcs: Rephrased. Description now introduces PPL fetch_size directly, SQL/PPL differences moved to a note section. Corrected the default behavior — now states result size is governed by `plugins.query.size_limit`. Removed `index.max_result_window` reference.

PR #5136, opened 2026-02-11
Proposal: Use MongoDB naming convention

LantaoJin:
### Description
This PR refactors the `graphLookup` command field names to align with MongoDB's $graphLookup aggregation stage naming convention, improving consistency with industry standards and making the API more intuitive for users familiar with MongoDB.

The original field names (`startField`, `fromField`, `toField`) were ambiguous and didn't clearly convey the relationship between fields. MongoDB's `$graphLookup` naming convention is
  well-established and more descriptive:
  - `startWith` - clearly indicates the initial value to start traversal with
  - `connectFromField` - explicitly shows this is the field to recurse through
  - `connectToField` - clearly indicates the field to match against

This alignment provides better developer experience and reduces confusion for users familiar with graph operations in MongoDB.

Field Name Mapping

  | Old Name | New Name | Description |
  |----------|----------|-------------|
  | `startField` | `startWith` | Field in source documents whose value starts the traversal |
  | `fromField` | `connectFromField` | Field in lookup documents containing values to recurse through |
  | `toField` | `connectToField` | Field in lookup documents to match against during traversal |

Example Query

  Before:
```
  source = employees
    | graphLookup employees
      startField=reportsTo
      fromField=reportsTo
      toField=name
      as reportingHierarchy
```
  After:
```
  source = employees
    | graphLookup employees
      startWith=reportsTo
      connectFromField=reportsTo
      connectToField=name
      as reportingHierarchy
```

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5138, opened 2026-02-11
[Feature] Support bi-directional graph traversal command `graphlookup`

### LINKED ISSUE #5088: [RFC] Implement a `graphLookup` command to demonstrate multi-hop queries
**Problem Statement**
We can functionally implement a PPL `graphLookup` command that is semantically equivalent to MongoDB’s `$graphLookup` [pipeline aggregation query](https://www.mongodb.com/docs/manual/reference/operator/aggregation/graphLookup/), which is to demonstrate multi-hop queries based on a graph model.

**Current State**
Lack of multi-hop ability in query

**Long-Term Goals**
* Describe the ideal outcome or future state you aim to achieve.
  * A new PPL command to support multi-hop query with high performance.
* What are the primary objectives of this solution should fulfill in the long run?
  * Support GraphRag query with PPL

**Proposal**
PPL syntax:
```
source = <vertex_index> 
| graphlookup
  <edge_index>
  startWith=<expression>
  fromField=<field>
  toField=<field>
  [maxDepth=<number>]
  [match=<expression>]
  as <new_field>
```

**Approach**
Approach 1
Implement with Calcite Recursive CTE: [POC here](https://github.com/LantaoJin/search-plugins-sql/tree/poc/poc_graphlookup)
```

                    ┌──────────────┐                                     
                    │ Graph G=(V,E)│                                     
                    └──────┬───────┘                                     
                           │ represented as                              
                           ▼                                             
                    ┌──────────────┐                                     
                    │edges(src,dst)│                                     
                    └──────┬───────┘                                     
                           │                                             
           ┌───────────────┼───────────────┐                             
           │               │               │                             
           ▼               ▼               ▼                             
      ┌─────────┐    ┌───────────┐    ┌───────────┐                      
      │  JOIN   │    │ UNION ALL │    │ RECURSIVE │                      
      │=traverse│    │=accumulate│    │= iterate  │                      
      │one edge │    │           │    │           │                      
      └─────────┘    └───────────┘    └───────────┘                      
           │               │               │                             
           └───────────────┼───────────────┘                             
                           │                                             
                           ▼                                             
                    ┌───────────────┐                                    
                    │ Graph Search  │                                    
                    │   Result      │                                    
                    │(reachable set)│                                    
                    └───────────────┘   
```
Approach 2 (perferred)
Provide a native Enumerable operator to implement high performance multi-hop traverse.

LantaoJin:
### Description
The `graphlookup` command performs recursive graph traversal on OpenSearch indices using a breadth-first search (BFS) algorithm.

```
┌───────────────────────────────────────────────────────────────┐
│              CalciteEnumerableGraphLookup                     │
│  ┌─────────────┐     ┌──────────────────────────────────────┐ │
│  │   Source    │────▶│         BFS Enumerator               │ │
│  │  Scannable  │     │  ┌────────────────────────────────┐  │ │
│  └─────────────┘     │  │  For each source row:          │  │ │
│                      │  │  1. Extract startField value   │  │ │
│  ┌─────────────┐     │  │  2. Query: toField = value     │  │ │
│  │   Lookup    │◀────│  │  3. Collect fromField values   │  │ │
│  │ IndexScan   │     │  │  4. Repeat until maxDepth      │  │ │
│  └─────────────┘     │  └────────────────────────────────┘  │ │
│                      └──────────────────────────────────────┘ │
└───────────────────────────────────────────────────────────────┘
```

### Syntax

Option 1

```
graphLookup <lookupIndex> startField=<field> fromField=<field> toField=<field> [maxDepth=<maxDepth>] [depthField=<field>] [direction=(uni | bi)] [supportArray=(true | false)] [batchMode=(true | false)] [usePIT=(true | false)] [filter=(<condition>)] as <outputField>
```
Example Query
```
  source = employees
    | graphLookup employees
      startField=reportsTo
      fromField=reportsTo
      toField=name
      as reportingHierarchy
```

Option 2 (follow MongoDB naming convention)

```
graphLookup <lookupIndex> startWith=<field> connectFromField=<field> connectToField=<field> [maxDepth=<maxDepth>] [depthField=<field>] [direction=(uni | bi)] [supportArray=(true | false)] [batchMode=(true | false)] [usePIT=(true | false)] [filter=(<condition>)] as <outputField>
```
Example Query
```
  source = employees
    | graphLookup employees
      startWith=reportsTo
      connectFromField=reportsTo
      connectToField=name
      as reportingHierarchy
```

### Performance

OpenSearch `graphlookup` command has implemented several features, which make it more applicable on RAG scenarios as well as suitable for different requirements.

* Using params of `direction=uni` and `usePIT=true`, OpenSearch `graphlookup` achieves the same result as MongoDB's `$graphlookup` aggregation command with **5-6x** faster on SNB SF-1 dataset, **29x** faster on SNB SF-10. And OpenSearch can finish all benchmark on SNB SF-10 in 734ms, SNB SF-30 in 8000ms while **MongoDB fails to complete**.
* In multiple start value scenarios, OpenSearch supports `batchMode`, which is more compatible with the requirements of RAG. Apart from already deduped on the final results, it also gets **58x** faster than MongoDB on SNB SF-1 dataset. And also OpenSearch can finish multiple start value benchmark on SNB SF-10 in 3508ms, on SNB SF-30 in 10992ms **while MongoDB fails to complete**.
* OpenSearch supports both uni-direction and bi-direction traversal. The latter will cost **4x** on query time than the former but also can retrieve **4x** more vertices.
* OpenSearch also support `PIT` and `non-PIT` mode, the latter will get better performance on large dataset by sacrificing the accuracy. It gets **5x** faster than the using PIT mode on SNB SF-30 but getting 27% nodes missed.

image (13)
image (14)

### Related Issues
Resolves #5088

### Check List


LantaoJin:
@penghuo can you review and merge this?

docs/user/ppl/index.md
```diff
@@ -82,6 +82,8 @@ source=accounts
 | [addcoltotals command](cmd/addcoltotals.md) | 3.5 | stable (since 3.5) | Adds column values and appends a totals row. |
 | [transpose command](cmd/transpose.md) | 3.5 | stable (since 3.5) | Transpose rows to columns. |
 | [mvcombine command](cmd/mvcombine.md) | 3.5 | stable (since 3.4) | Combines values of a specified field across rows identical on all other fields. |
+| [graphlookup command](cmd/graphlookup.md) | 3.5 | experimental (since 3.5) | Performs recursive graph traversal on a collection using a BFS algorithm.|
```
penghuo: since 3.6

docs/user/ppl/cmd/graphlookup.md
```diff
@@ -0,0 +1,357 @@
...
+# graphLookup (Experimental)
+
+The `graphLookup` command performs recursive graph traversal on a collection using a breadth-first search (BFS) algorithm. It searches for documents matching a start value and recursively traverses connections between documents based on specified fields. This is useful for hierarchical data like organizational charts, social networks, or routing graphs.
+
+## Syntax
+
+The `graphLookup` command has the following syntax:
+
+```syntax
+graphLookup <lookupIndex> startField=<startField> fromField=<fromField> toField=<toField> [maxDepth=<maxDepth>] [depthField=<depthField>] [direction=(uni | bi)] [supportArray=(true | false)] [batchMode=(true | false)] [usePIT=(true | false)] [filter=(<condition>)] as <outputField>
```
penghuo, line 11: Did we align on syntax? https://github.com/opensearch-project/sql/pull/5136#issuecomment-3886445182

cc: @qianheng-aws @model-collapse

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLGraphLookupIT.java
```diff
@@ -0,0 +1,738 @@
...
+        result,
+        schema("airport", "string"),
+        schema("connects", "string"),
+        schema("reachableAirports", "array"));
+    verifyDataRows(
+        result,
+        rows(
+            "JFK",
+            List.of("BOS", "ORD"),
+            List.of(List.of("JFK", List.of("BOS", "ORD")), List.of("BOS", List.of("JFK", "PWM")))));
```
penghuo, line 236: Corrected expectation would be:
```
  rows("JFK", List.of("BOS", "ORD"),
      List.of(
          List.of("JFK", List.of("BOS", "ORD")),
          List.of("BOS", List.of("JFK", "PWM")),
          List.of("ORD", List.of("JFK"))))
```
The point is: ORD is discovered by the BFS at Level 1, but then line 408's if (!nextValues.isEmpty()) drops it because ORD's only connection (JFK) was already visited. It that dropping is intentional?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteEnumerableGraphLookup.java
```diff
@@ -0,0 +1,545 @@
...
+            new LimitDigest(maxResultWindow, 0),
+            (OSRequestBuilderAction)
+                requestBuilder -> requestBuilder.pushDownLimit(maxResultWindow, 0));
+      }
+      // When usePIT is true, no limit is set, allowing PIT-based pagination for complete results
+
+      // Get the source enumerator
+      if (graphLookup.getSource() instanceof Scannable scannable) {
+        Enumerable<?> sourceEnum = scannable.scan();
+        this.sourceEnumerator = (Enumerator<@Nullable Object>) sourceEnum.enumerator();
```
penghuo, line 209: If an exception is thrown after line 209 but before the constructor finishes, the object is never fully constructed, so close() is never called by the caller. The sourceEnumerator leaks.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLGraphLookupIT.java
qianheng-aws: Yes, it's intentional not to include loop edges in the result in the current implementation. And The edge `ORD -> JFK` is a loop edge, which won't contribute to discover new vertices. Actually this edge should be dropped in the search query, before line 408.

There are some reasons dong that:
1. To exclude loop edges, we can add another filter `target no in visited nodes` directly in our BFS search query(line 462 & 474). It will be more efficient and that somehow explains why we are faster than mongoDB.
2. If we really want to include loop edge in the results, we also have to distinguish that from duplicated edges based on unique id like `doc_id`.

PR #5140, opened 2026-02-12
Add auto-extract mode for `spath` command

dai-chen:
### Description

As a follow-up after https://github.com/opensearch-project/sql/pull/5139, this PR implements the auto-extract mode for the `spath` command per the proposal in #4307. When `path` parameter is omitted, `spath` rewrites to `json_extract_all(input)` function and returns a `map<string, string>` with flattened keys instead of expanding dynamic columns.

Minor differences from proposal:
- Malformed JSON returns partial results

### Related Issues

Resolves https://github.com/opensearch-project/sql/issues/4307 (partially, ergonomic map access syntax are follow-ups)

### Check List

### DIFF
--- a/docs/user/ppl/cmd/spath.md
+++ b/docs/user/ppl/cmd/spath.md
@@ -4,7 +4,7 @@
 The `spath` command extracts fields from structured JSON data. It operates in two modes:
 
 - **Path-based mode**: When `path` is specified, extracts a single value at the given JSON path.
-- **Auto-extract mode**: When `path` is omitted, extracts all fields from the JSON into a map.
+- **Auto-extract mode** (experimental): When `path` is omitted, extracts all fields from the JSON into a map.
 
 > **Note**: The `spath` command is not executed on OpenSearch data nodes. It extracts fields from data after it has been returned to the coordinator node, which is slow on large datasets. We recommend indexing fields needed for filtering directly instead of using `spath` to filter nested fields.
 
@@ -28,7 +28,7 @@ The `spath` command supports the following parameters.
 
 For more information about path syntax, see [json_extract](../functions/json.md#json_extract).
 
-## Auto-extract mode
+## Auto-extract mode (experimental)
 
 When `path` is omitted, the `spath` command runs in auto-extract mode. Instead of extracting a single value, it flattens the entire JSON into a `map<string, string>` column using the following rules:
 


integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLSpathCommandIT.java
```diff
@@ -36,6 +36,16 @@ public void init() throws Exception {
...
+
+    // Auto-extract mode: one doc field per flatten rule
+    Request autoExtractDoc = new Request("PUT", "/test_spath_auto/_doc/1?refresh=true");
+    autoExtractDoc.setJsonEntity(
+        "{\"nested_doc\": \"{\\\"user\\\":{\\\"name\\\":\\\"John\\\"}}\","
+            + " \"array_doc\": \"{\\\"tags\\\":[\\\"java\\\",\\\"sql\\\"]}\","
+            + " \"merge_doc\": \"{\\\"a\\\":{\\\"b\\\":1},\\\"a.b\\\":2}\","
+            + " \"stringify_doc\":"
+            + " \"{\\\"n\\\":30,\\\"b\\\":true,\\\"x\\\":null}\"}");
+    client().performRequest(autoExtractDoc);
```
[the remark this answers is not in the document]
penghuo: Any test cover malformed JSON input in auto-extract mode?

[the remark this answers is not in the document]
dai-chen: Addressed in https://github.com/opensearch-project/sql/pull/5140/commits/4890a4edf1631fd95613d3d827f9a433212b8e4d. Thanks!

docs/user/ppl/cmd/spath.md
```diff
@@ -20,11 +23,25 @@ The `spath` command supports the following parameters.
...
+
+- Nested objects use dotted keys: `user.name`, `user.age`
+- Arrays use `{}` suffix: `tags{}`, `users{}.name`
+- Duplicate logical keys merge into arrays: `c{}.b = [2, 3]`
+- Null values are preserved: a JSON `null` becomes the string `"null"` in the map
+- All values are stringified: numbers and booleans are converted to their string representation (for example, `30` becomes `"30"`, `true` becomes `"true"`, and arrays become `"[a, b, c]"`)
+
+> **Note**: Auto-extract mode processes the entire input field with no character limit. For large JSON payloads, consider using path-based extraction to target specific fields.
+
+> **Note**: Invalid or malformed JSON returns partial results containing any fields successfully parsed before the error. Empty JSON object (`{}`) returns an empty map.
```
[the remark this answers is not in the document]
dai-chen: Addressed in https://github.com/opensearch-project/sql/pull/5140/commits/4890a4edf1631fd95613d3d827f9a433212b8e4d.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLSpathCommandIT.java
```diff
@@ -45,4 +65,102 @@ public void testSimpleSpath() throws IOException {
...
+  @Test
+  public void testSpathAutoExtractArraySuffix() throws IOException {
+    JSONObject result =
+        executeQuery(
+            "source=test_spath_auto | spath input=array_doc output=result | fields result");
+
+    // Arrays use {} suffix: tags{}
+    verifySchema(result, schema("result", "struct"));
+    verifyDataRows(result, rows(new JSONObject("{\"tags{}\":\"[java, sql]\"}")));
+  }
```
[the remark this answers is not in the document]
dai-chen: Ignore.

docs/user/ppl/cmd/spath.md
```diff
@@ -123,3 +140,35 @@ fetched rows / total rows = 3/3
...
+The query returns the following results:
+  
+```text
+fetched rows / total rows = 3/3
++---------------------------------------------------------------------------------+------------------------------------------------------------------------------------+
+| doc_auto                                                                        | result                                                                             |
+|---------------------------------------------------------------------------------+------------------------------------------------------------------------------------|
+| {"user":{"name":"John","age":30},"tags":["java","sql"],"active":true}           | {'user.age': '30', 'tags{}': '[java, sql]', 'user.name': 'John', 'active': 'true'} |
+| {"user":{"name":"Jane","age":25},"tags":["python"],"active":null}               | {'user.age': '25', 'tags{}': 'python', 'user.name': 'Jane', 'active': 'null'}      |
+| {"user":{"name":"Bob","age":35},"tags":["go","rust","sql"],"user.name":"Bobby"} | {'user.age': '35', 'tags{}': '[go, rust, sql]', 'user.name': '[Bob, Bobby]'}       |
```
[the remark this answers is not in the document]
dai-chen: This is by design. The value could be scalar or array.

Swiddis, line 163: issue (non-blocking): Is using single quotes for the keys (python-style) by design?

On instinct I would prefer to use standard JSON syntax for safer consistency with other json operators down the road. Not massively opposed if this is the direction we want.

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
```diff
@@ -82,7 +88,18 @@ public static Object eval(Object... args) {
       return null;
     }
 
-    return parseJson(jsonStr);
+    Map<String, Object> parsed = parseJson(jsonStr);
+    return parsed == null ? null : stringifyMap(parsed);
+  }
+
+  // TODO: JSON parsing dominates cost; consider stringify scalars in place during parsing
+  //  to avoid this extra pass.
```
Swiddis, line 96: question (non-blocking): Can you elaborate more on stringifying scalars in place?

docs/user/ppl/cmd/spath.md
Swiddis: Actually if I understand right this is just CLI output, over HTTP it's getting returned as a plain JSON object

core/src/test/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImplTest.java
```diff
@@ -66,281 +40,371 @@ public void testOperandMetadata() {
...
+                "other": 5
+              },
+              "another": [6, [7, 8], 9]
+            }\
+            """),
+        is(
+            Map.of(
+                "data.items{}{}", "[1, 2]",
+                "data.items{}{}.hello", "3",
+                "data.items{}", "4",
```
Swiddis, line 253: thought: Is this bracket syntax easily supported by downstream commands that want to extract these fields, or will there need to be odd escaping?

That might only be a concern for the mapping functions in a future PR, but I would still carefully review if this is the extraction syntax we want for arrays. At a glance this isn't intuitive to me but after squinting at the docs for a bit I kinda got it?

e.g. It's not obvious to me what will happen if I try `"items": [[1, 2], [3, 4]]`, do I get two `items{}{}` keys? Does one overwrite the other? Are they merged into one array? (From code review I know which one it is, but not as a user given only the docs & this example)

docs/user/ppl/cmd/spath.md
dai-chen: Yes, it's CLI output. The value is just plain string in JSON.

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImpl.java
dai-chen: Sure, currently we have this second pass from Map<String,Object> to Map<String,String>. I was thinking maintain and update Map<String,String> in place. But the string manipulation will be tricky when merging a key's value with the new one into a string in array format.

core/src/test/java/org/opensearch/sql/expression/function/jsonUDF/JsonExtractAllFunctionImplTest.java
dai-chen: Yes, I think we don't have good workaround until we move to Map<String,Any> or schemaless. Currently it just preserves all the values from "conflicting" key. Let me verify your example.

dai-chen: The result is `items{}{}` as expected. I can add more cases in doctest or this UT if that's your suggestion. Thanks!

PR #5141, opened 2026-02-12
[Feature] Calcite PPL search result highlighting

### LINKED ISSUE #5059: [FEATURE] Supporting Query Highlight Feature into PPL API
When you use DQL in discover, the results set highlight the search query into the results, see the following screen shot:

Image

The following are an example of query with highlight:

`GET opensearch_dashboards_sample_data_logs/_search
{
      "sort": [
        {
          "timestamp": {
            "order": "desc",
            "unmapped_type": "boolean"
          }
        }
      ],
      "size": 500,
      "version": true,
      "aggs": {
        "2": {
          "date_histogram": {
            "field": "timestamp",
            "calendar_interval": "1d",
            "time_zone": "America/Los_Angeles",
            "min_doc_count": 1
          }
        }
      },
      "stored_fields": [
        "*"
      ],
      "script_fields": {
        "hour_of_day": {
          "script": {
            "source": "doc['timestamp'].value.getHour()",
            "lang": "painless"
          }
        }
      },
      "docvalue_fields": [
        {
          "field": "@timestamp",
          "format": "date_time"
        },
        {
          "field": "timestamp",
          "format": "date_time"
        },
        {
          "field": "utc_time",
          "format": "date_time"
        }
      ],
      "_source": {
        "excludes": []
      },
      "query": {
        "bool": {
          "must": [],
          "filter": [
            {
              "bool": {
                "should": [
                  {
                    "match_phrase": {
                      "agent": "Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322)"
                    }
                  }
                ],
                "minimum_should_match": 1
              }
            },
            {
              "range": {
                "timestamp": {
                  "gte": "2025-10-03T19:04:39.469Z",
                  "lte": "2026-01-16T20:04:39.469Z",
                }
              }
            }
          ],
          "should": [],
          "must_not": []
        }
      },
      "highlight": {
        "pre_tags": [
          "@opensearch-dashboards-highlighted-field@"
        ],
        "post_tags": [
          "@/opensearch-dashboards-highlighted-field@"
        ],
        "fields": {
          "*": {}
        },
        "fragment_size": 2147483647
      }
    }`

We need to support something similar using PPL API.

### LINKED ISSUE #5156: [FEATURE RFC] PPL Search Result Highlighting
# PPL Search Result Highlighting — Design

**Author**: Jialiang Liang
**Date**: 2026-02-17

---

## 1. Scope

This design covers **search result highlighting for PPL queries** executed through the
Calcite engine. There are two distinct user stories:

### User Story 1: OSD Explore users

When a user switches to PPL mode in Explore and runs `search source=logs "error"`,
matching terms should be highlighted in the results table — the same experience they get
with DSL today. The user does not configure highlighting; OSD handles it automatically
using its existing `getHighlightRequest()` function.

### User Story 2: API / CLI users

API consumers sending PPL queries via `POST /_plugins/_ppl` may want to highlight
specific fields with custom tags (e.g., `<em>` for HTML). They need full control over
which fields are highlighted, what tags are used, and the fragment size.

### Both stories, one mechanism

Both are served by the same API: an optional `highlight` object in the PPL request body.
OSD and API users construct the config differently, but the backend treats them
identically — it is a pure pass-through. When no `highlight` is provided, behavior is
unchanged (backward compatible).

### Out of scope

- Porting the V2 `highlight()` PPL function to the Calcite engine
- Customizable highlight settings in OSD UI (e.g., per-field configuration from Explore)

---

## 2. Highlighting Behavior

PPL highlighting is fully aligned with the existing DSL highlight feature — the same
OpenSearch highlighting engine, the same rules for what gets highlighted, and the same
rendering in OSD Explore.

### When do highlights appear?

Only when the PPL query contains a **full-text search** — e.g.,
`search source=logs "error"` or `search source=logs "connection timeout"`. The search
term is translated to a `query_string` query, and OpenSearch's highlighter identifies
where that term appears in the document.

If the query contains only structured filters (e.g., `search source=logs | where status = 200`),
no highlights are produced — even if the highlight config is present in the request.

### What gets highlighted?

The matching search terms inside **`text` and `keyword` field values**. For example,
searching `"Holmes"` highlights the term `Holmes` wherever it appears — in `firstname`,
`lastname`, `address`, or any other string field that contains the match.

### What does NOT get highlighted?

- **Non-string fields**: Numeric, date, boolean, and other non-text field types never
  produce highlight fragments, even if the document matched because of those fields.
- **Structured filters**: PPL commands like `where`, `stats`, or conditions using
  comparison operators (`>`, `<`, `=`) do not produce highlights. These are translated
  to range/term/bool filters in the DSL, which OpenSearch's highlighter does not act on.

### What about piped commands?

Piped commands narrow the result set but do not affect which terms are highlighted.
For example:

```
search source=logs "error" | where status > 400
```

The `where status > 400` filters rows, but only the `"error"` full-text search produces
highlights. This is the same behavior as a DSL `bool` query with a `query_string` in
`must` and a `range` in `filter` — the filter narrows results without contributing to
highlights.

### What does it look like?

Matching terms in field values are wrapped in configurable tags:

- **OSD Explore**: Tags are OSD internal markers (`@opensearch-dashboards-highlighted-field@`)
  that OSD renders as bold/colored text in the results table — identical to DSL behavior.
- **API/CLI**: Users choose their own tags (e.g., `<em>`, `<mark>`, or custom markers)
  and see them in the JSON response.

---

### Principle: caller-driven, backend pass-through

- The **caller** (OSD, API client, CLI) controls highlighting by providing a `highlight`
  object in the PPL request body
- The **backend** forwards the config as-is to OpenSearch and returns highlight data in
  the response
- When **no `highlight`** is provided, no highlighting is applied

This is consistent with how DSL works: OSD injects the highlight clause, not OpenSearch.

### Request API

```json
POST /_plugins/_ppl
{
  "query": "search source=logs \"error\"",
  "highlight": {
    "fields": { "*": {} },
    "pre_tags": ["@opensearch-dashboards-highlighted-field@"],
    "post_tags": ["@/opensearch-dashboards-highlighted-field@"],
    "fragment_size": 2147483647
  }
}
```

The `highlight` object supports the same structure as OpenSearch's [highlighting API](https://opensearch.org/docs/latest/search-plugins/searching-data/highlight/):

| Field | Description |
|-------|-------------|
| `fields` | Map of field names to per-field config. `"*"` for wildcard. |
| `pre_tags` | Array of tags inserted before highlighted terms |
| `post_tags` | Array of tags inserted after highlighted terms |
| `fragment_size` | Max character length of each fragment. OSD sets `2^31 - 1` so the entire field value is returned rather than OpenSearch's default 100-char truncation. |

**API/CLI example** — specific field with custom tags:

```json
POST /_plugins/_ppl
{
  "query": "search source=logs \"error\"",
  "highlight": {
    "fields": { "message": {} },
    "pre_tags": ["<em>"],
    "post_tags": ["</em>"],
    "fragment_size": 200
  }
}
```

### Response format

The response includes a `highlights` array parallel to `datarows`:

```json
{
  "schema": [{ "name": "firstname", "type": "string" }, ...],
  "datarows": [["Holmes", ...], ["Blanche", ...], ["Amber", ...]],
  "highlights": [
    { "firstname": ["<tag>Holmes</tag>"], "firstname.keyword": ["<tag>Holmes</tag>"] },
    { "lastname": ["<tag>Holmes</tag>"], "lastname.keyword": ["<tag>Holmes</tag>"] },
    { "address": ["880 <tag>Holmes</tag> Lane"] }
  ],
  "total": 3,
  "size": 3
}
```

- Each entry in `highlights` corresponds to the row at the same index in `datarows`
- Entries are `null` when a row has no highlight data for the requested fields
- The `highlights` array is **omitted entirely** when no highlight config is provided

---

### 4.1 No command-level scoping — follow DSL behavior

DSL does not scope highlighting to specific query clauses — `getHighlightRequest()`
applies the same wildcard config regardless of query structure. To match this behavior,
when a caller provides a `highlight` config, the backend attaches it to the OpenSearch
request regardless of which PPL commands are in the pipeline.

### 4.2 Relationship with V2 `highlight()` function

PPL has an existing per-field `highlight()` function in the V2 engine (e.g.,
`highlight(msg, pre_tags='<em>')`). This is **not supported in the Calcite engine**.
The request-level API covers the same use cases:

| Capability | V2 `highlight()` function | Request-level API |
|-----------|--------------------------|-------------------|
| Per-field control | `highlight(msg)` in PPL query | `"fields": { "msg": {} }` in request body |
| Custom tags | `highlight(msg, pre_tags='<em>')` | `"pre_tags": ["<em>"]` in request body |
| Wildcard fields | `highlight(*)` in PPL | `"fields": { "*": {} }` in request body |

Porting `highlight()` to Calcite is out of scope. The backend plumbing supports both
approaches if an in-query function is added later.

### 4.3 Backend is a pure pass-through — no hardcoded defaults

An alternative considered was a `?highlight=true` query parameter that triggers the
backend to inject a default config. This was rejected because it would hardcode
OSD-specific knowledge (tags, fragment size) into the backend.

Instead, the backend simply forwards whatever config the caller provides. One mechanism
for all callers, no OSD-specific knowledge in the backend.

---

### Methodology

A/B benchmark comparing PPL query execution **with** and **without** highlighting on a
worst-case dataset where every document matches and every text field produces highlights:

- **Dataset**: 10,000 documents, search term `"error"` in **4 text fields** per document
- **Query**: `search source=highlight_perf_test "error"` — returns all 10,000 rows
- **Iterations**: 20 measured runs after 3 warmup runs
- **Environment**: Single-node OpenSearch (1 shard, 0 replicas), local dev machine

### Results

| Scenario | Avg Latency | P50 | P95 | Response Size |
|----------|-------------|-----|-----|---------------|
| **No highlight** | 163 ms | 161 ms | 197 ms | 2.84 MB |
| **Highlight (wildcard `*`)** | 617 ms | 613 ms | 669 ms | 9.29 MB |
| **Highlight (single field)** | 299 ms | 298 ms | 325 ms | 4.06 MB |

| Comparison | Latency Overhead | Size Overhead |
|------------|-----------------|---------------|
| Wildcard vs no highlight | +278% (~3.8x) | +227% (~3.3x) |
| Single field vs no highlight | +83% (~1.8x) | +43% (~1.4x) |

### Analysis

1. **Worst-case benchmark.** Every row matches, every text field highlights. Real-world
   result sets are smaller and sparser.

2. **The overhead is in OpenSearch, not the SQL plugin.** The backend is a pure
   pass-through. The latency and size cost is OpenSearch's native highlighting — the
   same cost DSL users already pay today.

3. **Single-field highlighting is significantly cheaper.** API users who specify only the
   fields they need pay roughly half the overhead of wildcard.

4. **Response size is proportional to highlight data.** Wildcard adds ~6.4 MB of
   highlight fragments (tags + full field values for 4 fields x 10k rows).

---

## 6. Limitations

1. **Same limitations as DSL highlighting.** Only full-text queries produce highlights;
   only `text`/`keyword` fields; no query-aware scoping. These are inherent OpenSearch
   limitations (see [Background: DSL Highlighting](#background-dsl-highlighting) in the Appendix).

2. **`shouldHighlight` is hardcoded to `true`.** The OSD frontend currently always sends
   the highlight config. It should read from the `doc_table:highlight` UI setting.

3. **No in-query `highlight()` function in Calcite.** The request-level API covers the
   same capabilities, but interactive PPL users cannot use `highlight()` in queries.

---

## Appendix

<details>
<summary><strong>Background: DSL Highlighting</strong></summary>

### How OSD enables DSL highlighting

OSD's `SearchSource.flatten()` injects a `highlight` clause into every DSL request when
the `doc_table:highlight` UI setting is enabled (default: `true`). The
`getHighlightRequest()` function returns a hardcoded config: wildcard fields, OSD custom
tags, and `fragment_size: 2^31 - 1`.

| Aspect | Behavior |
|--------|----------|
| Fields | Always `"*"` (wildcard) — hardcoded |
| Tags | Always OSD custom tags — hardcoded |
| Fragment size | Always max int — hardcoded |
| On/off | `doc_table:highlight` UI setting (default: `true`) |
| Query-aware scoping | **None** — receives the query but never inspects it |

### DSL highlighting limitations

These are inherent OpenSearch limitations, not OSD-specific:

1. **Only full-text queries produce highlights.** `query_string`, `match`,
   `match_phrase`, etc. Structured filters (`range`, `term`, `bool` filter) do not.

2. **Only `text` and `keyword` fields are highlighted.** Numeric, boolean, date, and
   other non-string types produce no fragments.

3. **No query-aware scoping.** Same wildcard config regardless of query structure.

4. **Keyword subfields are included.** `"*"` matches both `firstname` and
   `firstname.keyword`.

These limitations apply equally to PPL highlighting since we use the same mechanism.

</details>

<details>
<summary><strong>Backend Implementation Details</strong></summary>

### Data flow

```
PPLQueryRequest         ->  PPLService            ->  AbstractPlan (carries highlightConfig)
   .getHighlight()          .setHighlightOnPlan()      across thread boundary

                        ->  QueryPlan.execute()    ->  CalcitePlanContext ThreadLocal
                            (on worker thread)         .setHighlightConfig()

                        ->  CalciteEnumerableIndexScan.scan()
                            reads ThreadLocal, builds HighlightBuilder,
                            attaches to OpenSearchRequestBuilder

OpenSearch response     ->  OpenSearchIndexEnumerator
   _highlight field         COLLECTED_HIGHLIGHTS ThreadLocal side-channel

                        ->  OpenSearchExecutionEngine.buildResultSet()
                            merges highlights into QueryResult

                        ->  SimpleJsonResponseFormatter
                            writes "highlights" array in JSON response
```

### Key files

| File | Role |
|------|------|
| `PPLQueryRequest.getHighlight()` | Extracts `highlight` JSONObject from request body |
| `PPLService.setHighlightOnPlan()` | Attaches config to `AbstractPlan` for cross-thread transport |
| `AbstractPlan.highlightConfig` | Carries config from REST handler thread to worker thread |
| `QueryPlan.execute()` / `ExplainPlan.execute()` | Sets `CalcitePlanContext` ThreadLocal on worker thread |
| `AbstractCalciteIndexScan.applyHighlightConfig()` | Converts config map to `HighlightBuilder` |
| `OpenSearchIndexEnumerator` | Collects `_highlight` from OpenSearch response via side-channel |
| `OpenSearchExecutionEngine.buildResultSet()` | Merges highlights into `QueryResult` |
| `SimpleJsonResponseFormatter` | Writes `highlights` array in JSON response |

### Threading model

The PPL endpoint runs on a REST handler thread, but query execution runs on a separate
`sql-worker` thread pool. ThreadLocals do not cross thread boundaries.

**Solution**: The highlight config is carried via the `AbstractPlan` object (a normal
Java reference). `execute()` runs on the worker thread and sets the ThreadLocal there.

```
REST handler thread:  PPLService -> setHighlightOnPlan(plan) -> queryManager.submit(plan)
                                                                        |
Worker thread:        plan.execute() -> setHighlightThreadLocal() -> CalcitePlanContext.run()
                                                                        |
                      analyze() -> optimize() -> scan() -> applyHighlightConfig()
```

</details>

<details>
<summary><strong>Frontend Implementation Details (OSD)</strong></summary>

### Changes

| File | Change |
|------|--------|
| `ppl_search_strategy.ts` | Calls `getHighlightRequest()` and attaches config to `request.body.highlight` |
| `facet.ts` | Forwards `request.body.highlight` into `params.body` sent to `/_plugins/_ppl` |

### Response handling

| File | Role |
|------|------|
| `ppl_search_strategy.ts` | Stores `rawResponse.data.highlights` on `dataFrame.meta` |
| `data_frames/utils.ts` | Attaches `meta.highlights[i]` as `hit.highlight` per row in `convertResult()` |

`getHighlightRequest()` is the same function OSD uses for DSL — no new OSD-specific code
needed.

</details>

<details>
<summary><strong>Sample Queries & Responses</strong></summary>

Test data: `accounts.json` (1000 documents with fields like `firstname`, `lastname`,
`address`, `age`, etc.)

### With highlight (OSD-style — wildcard fields, OSD tags)

```bash
curl -s -X POST "localhost:9200/_plugins/_ppl" \
  -H 'Content-Type: application/json' -d '{
    "query": "search source=accounts \"Holmes\"",
    "highlight": {
      "fields": { "*": {} },
      "pre_tags": ["@opensearch-dashboards-highlighted-field@"],
      "post_tags": ["@/opensearch-dashboards-highlighted-field@"],
      "fragment_size": 2147483647
    }
  }' | jq
```

**Response** (trimmed to highlights):

```json
{
  "highlights": [
    {
      "firstname": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"],
      "firstname.keyword": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"]
    },
    {
      "lastname": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"],
      "lastname.keyword": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"]
    },
    {
      "address": ["880 @opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@ Lane"]
    }
  ],
  "total": 3,
  "size": 3
}
```

### Without highlight (backward compatible)

```bash
curl -s -X POST "localhost:9200/_plugins/_ppl" \
  -H 'Content-Type: application/json' -d '{
    "query": "search source=accounts \"Holmes\""
  }' | jq
```

**Response**: No `highlights` field — backward compatible.

### Explain (highlight appears in generated DSL)

```bash
curl -s -X POST "localhost:9200/_plugins/_ppl/_explain" \
  -H 'Content-Type: application/json' -d '{
    "query": "search source=accounts \"Holmes\"",
    "highlight": {
      "fields": { "*": {} },
      "pre_tags": ["@opensearch-dashboards-highlighted-field@"],
      "post_tags": ["@/opensearch-dashboards-highlighted-field@"],
      "fragment_size": 2147483647
    }
  }' | jq
```

**Response** (relevant portion):

```json
{
  "query": { "query_string": { "query": "Holmes" } },
  "highlight": {
    "pre_tags": ["@opensearch-dashboards-highlighted-field@"],
    "post_tags": ["@/opensearch-dashboards-highlighted-field@"],
    "fields": { "*": { "fragment_size": 2147483647 } }
  }
}
```

### Custom tags + specific field (API user scenario)

```bash
curl -s -X POST "localhost:9200/_plugins/_ppl" \
  -H 'Content-Type: application/json' -d '{
    "query": "search source=accounts \"Holmes\"",
    "highlight": {
      "fields": { "address": {} },
      "pre_tags": ["<em>"],
      "post_tags": ["</em>"],
      "fragment_size": 200
    }
  }' | jq
```

**Response**:

```json
{
  "highlights": [
    null,
    null,
    { "address": ["880 <em>Holmes</em> Lane"] }
  ],
  "total": 3,
  "size": 3
}
```

Only `address` is highlighted. Rows 1 and 2 are `null` because "Holmes" appears in
`firstname`/`lastname`, not `address`.

### Text search + piped filter

```bash
curl -s -X POST "localhost:9200/_plugins/_ppl" \
  -H 'Content-Type: application/json' -d '{
    "query": "search source=accounts \"Holmes\" | where balance > 40000",
    "highlight": {
      "fields": { "*": {} },
      "pre_tags": ["@opensearch-dashboards-highlighted-field@"],
      "post_tags": ["@/opensearch-dashboards-highlighted-field@"],
      "fragment_size": 2147483647
    }
  }' | jq
```

**Response**:

```json
{
  "highlights": [
    {
      "lastname": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"],
      "lastname.keyword": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"]
    }
  ],
  "total": 1,
  "size": 1
}
```

The `where balance > 40000` narrows results but does not produce highlights (it is a
range filter, not a full-text query).

</details>

RyanL1997:
### Description. 
> **For ease of review, please reference the design doc in issue #5156.**                                                                                                                                                                                                               
- Add request-level highlight API for PPL queries. Callers (OSD, API, CLI) can include an optional highlight JSON object in the PPL request body, which the backend forwards as-is to OpenSearch.                                              
- Highlight config is carried across the thread boundary (REST handler → sql-worker) via AbstractPlan, then set as a ThreadLocal for Calcite planning and execution.
- Highlight metadata from OpenSearch hits is collected via a side-channel ThreadLocal in OpenSearchIndexEnumerator and merged back into the JDBC/SimpleJson response as a highlights array parallel to datarows.

### Related Issues
* Resolve #5156 
* Resolve #5059

### Check List

### DIFF
--- a/docs/user/ppl/interfaces/endpoint.md
+++ b/docs/user/ppl/interfaces/endpoint.md
@@ -59,14 +59,7 @@ Expected output:
 
 You can include an optional `highlight` object in the request body to request search result highlighting. When a PPL query contains a full-text search (e.g., `search source=logs "error"`), OpenSearch identifies where the search terms appear in document fields and returns the matching fragments wrapped in the specified tags.
 
-The `highlight` object supports the following parameters:
-
-| Parameter | Type | Required | Description |
-|-----------|------|----------|-------------|
-| `fields` | Object | Yes | Map of field names to per-field configuration. Use `"*"` to highlight all `text` and `keyword` fields that match. |
-| `pre_tags` | Array of strings | No | Tags inserted before each highlighted term. Default: `["<em>"]`. |
-| `post_tags` | Array of strings | No | Tags inserted after each highlighted term. Default: `["</em>"]`. |
-| `fragment_size` | Integer | No | Maximum character length of each highlight fragment. Default is OpenSearch's default (100). Set to `2147483647` to return the entire field value without truncation. |
+The `highlight` object is passed as-is to OpenSearch. See the [OpenSearch highlighting documentation](https://docs.opensearch.org/latest/search-plugins/searching-data/highlight/) for all supported parameters.
 
 ### What gets highlighted
 
@@ -202,7 +195,7 @@ Only the `address` field is highlighted. The `<mark>` custom tags are used inste
 ### Limitations
 
 - Highlighting is supported only in the Calcite engine.
-- The backend forwards the highlight config as-is to OpenSearch. The same highlighting behavior and limitations as [OpenSearch's highlighting API](https://opensearch.org/docs/latest/search-plugins/searching-data/highlight/) apply.
+- The backend forwards the highlight config as-is to OpenSearch. The same highlighting behavior and limitations as [OpenSearch's highlighting API](https://docs.opensearch.org/latest/search-plugins/searching-data/highlight/) apply.
 - Highlighting works with **single-source queries only**, consistent with DSL where highlighting is inherently single-index per request. Behavior with joins (`| join`), subqueries, and multi-source queries is not validated.
 
 ## Explain


ppl/src/main/java/org/opensearch/sql/ppl/domain/PPLQueryRequest.java
```diff
@@ -109,4 +109,18 @@ public int getFetchSize() {
...
   }
+
+  /**
+   * Get the highlight configuration from the request body. The caller (OSD, API, CLI) controls
+   * highlighting by providing a highlight object in the PPL request. The backend forwards this
+   * config as-is to OpenSearch.
+   *
+   * @return highlight JSONObject from request, or null if not specified
+   */
+  public JSONObject getHighlight() {
```
penghuo: 1. Add doc to describe `highligh` parameters.  include, what field can be highlight.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
```diff
@@ -278,6 +281,21 @@ private QueryResponse buildResultSet(
       values.add(ExprTupleValue.fromExprValueMap(row));
     }
 
+    // Merge highlight data collected by the enumerator back into ExprTupleValues.
+    // The Calcite row pipeline only carries schema column values, so highlight metadata
+    // is collected as a side channel in OpenSearchIndexEnumerator and merged here.
+    List<ExprValue> collectedHighlights =
+        OpenSearchIndexEnumerator.getAndClearCollectedHighlights();
+    for (int i = 0; i < Math.min(values.size(), collectedHighlights.size()); i++) {
```
penghuo: Highlight-to-row positional misalignment when Calcite operators sit above the index scan. The COLLECTED_HIGHLIGHTS ThreadLocal collects one entry per row the enumerator produces (raw scan order). After Calcite processing, buildResultSet iterates the final ResultSet, which may contain fewer rows in different order if non-pushable operators (filter, sort, dedup, head/tail) were applied above the scan.

integ-test/src/yamlRestTest/resources/rest-api-spec/test/ppl_highlight.yml
```diff
@@ -0,0 +1,100 @@
...
+          highlight:
+            fields:
+              "*": {}
+            pre_tags:
+              - "<em>"
+            post_tags:
+              - "</em>"
+
+  - match: {"size": 2}
+  - is_true: highlights
```
penghuo: assertions only check is_true/is_false, no content validation

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
```diff
@@ -27,6 +31,27 @@
  */
 public class OpenSearchIndexEnumerator implements Enumerator<Object> {
 
+  /**
+   * Thread-local collector for highlight data. Since the Calcite row pipeline only carries schema
+   * column values, highlight metadata from OpenSearch hits is collected here as a side channel.
+   * After execution, {@link #getAndClearCollectedHighlights()} retrieves the collected data so it
+   * can be merged back into the ExprTupleValues for the JDBC response.
+   */
+  private static final ThreadLocal<List<ExprValue>> COLLECTED_HIGHLIGHTS =
```
penghuo: Is highlighting intended only for single-source search commands, or should it work with joins/subqueries?

COLLECTED_HIGHLIGHTS is a single static ThreadLocal shared across all enumerator instances. If a query involves a join or union of two OpenSearch indices, multiple enumerator instances will append to the same list.

ppl/src/main/java/org/opensearch/sql/ppl/domain/PPLQueryRequest.java
RyanL1997: added at: docs/user/ppl/interfaces/endpoint.md ([link](https://github.com/opensearch-project/sql/pull/5141/changes#diff-a5740f2d94985b7ece9712540bdd0c84cba20cc8179720b12e020b2e71593c85R56))

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
RyanL1997: Good catch. I've addressed this by replacing the `COLLECTED_HIGHLIGHTS` ThreadLocal side-channel entirely. Instead, `_highlight` is now a hidden column in the Calcite `RowType` (`SqlTypeName.ANY`).                                                 
                                                                                                                                                                                                                                                 
How it works: `CalciteLogicalIndexScan.buildInitialSchema()` appends `_highlight` to the scan's schema when highlight config is present. `OpenSearchResponse.addHighlightsToBuilder()` builds the highlight data as an `ExprTupleValue` and puts it directly into the row tuple. `OpenSearchIndexEnumerator.current()` carries it as an opaque `ExprValue` in the Calcite row array.

Because `_highlight` is now a regular column, Calcite operators (filter, sort, dedup, head/tail) naturally preserve it alongside the row data — no positional misalignment is possible. When a filter removes a row, its highlight goes with it. When sort reorders rows, highlights move with them.

On the output side, `OpenSearchExecutionEngine.buildResultSet()` extracts `_highlight` from each row and excludes it from the response schema (hidden from users). `QueryResult.highlights()` produces a parallel array for the formatters. `| fields` projections also preserve `_highlight` automatically via `CalciteRelNodeVisitor.visitProject()`, matching DSL behavior where `_source` filtering never affects highlights.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java
RyanL1997: Yes, Highlighting is single-source only, consistent with DSL where highlighting is inherently single-index per request — there's no DSL equivalent of highlighting across a join.

Regarding the `COLLECTED_HIGHLIGHTS` ThreadLocal concern — this is no longer applicable. We replaced the ThreadLocal side-channel entirely with a hidden `_highlight` column approach (see the response on the positional misalignment [comment](https://github.com/opensearch-project/sql/pull/5141#discussion_r2831326458)). There is no shared static list anymore.

docs/user/ppl/interfaces/endpoint.md
```diff
@@ -53,7 +53,119 @@ Expected output:
...
 ```
   
-## Explain  
+## Highlight
+
+### Description
+
+You can include an optional `highlight` object in the request body to request search result highlighting. When a PPL query contains a full-text search (e.g., `search source=logs "error"`), OpenSearch identifies where the search terms appear in document fields and returns the matching fragments wrapped in the specified tags.
+
+The `highlight` object supports the following parameters:
```
dai-chen: Should we simply reference https://docs.opensearch.org/latest/search-plugins/searching-data/highlight ?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLHighlightIT.java
```diff
@@ -0,0 +1,166 @@
...
+    JSONArray highlights = result.getJSONArray("highlights");
+
+    for (int i = 0; i < highlights.length(); i++) {
+      if (!highlights.isNull(i)) {
+        JSONObject hl = highlights.getJSONObject(i);
+        // Only address field should be highlighted, not other text fields
+        assertFalse("Should not highlight firstname field", hl.has("firstname"));
+      }
+    }
+  }
```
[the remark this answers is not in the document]
RyanL1997: fixed.

protocol/src/test/java/org/opensearch/sql/protocol/response/format/SimpleJsonResponseFormatterTest.java
```diff
@@ -178,4 +183,38 @@ void formatErrorPretty() {
...
+  @Test
+  void formatResponseWithoutHighlightsOmitsField() {
+    QueryResult response =
+        new QueryResult(
+            schema,
+            Collections.singletonList(tupleValue(ImmutableMap.of("firstname", "John", "age", 20))));
+    SimpleJsonResponseFormatter formatter = new SimpleJsonResponseFormatter(COMPACT);
+    String result = formatter.format(response);
+    assertFalse(result.contains("\"highlights\""));
+  }
```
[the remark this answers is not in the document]
RyanL1997: fixed

protocol/src/test/java/org/opensearch/sql/protocol/response/QueryResultTest.java
```diff
@@ -106,4 +111,89 @@ void iterate() {
...
+                        "age",
+                        ExprValueUtils.integerValue(20),
+                        "_highlight",
+                        ExprValueUtils.LITERAL_MISSING))),
+            Cursor.None);
+
+    List<Map<String, Object>> highlights = response.highlights();
+    assertEquals(1, highlights.size());
+    assertNull(highlights.get(0));
+  }
```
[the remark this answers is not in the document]
RyanL1997: fixed

dai-chen (commented):
Although we only expose highlight via the PPL API today, internally can we make it generic — accept arbitrary DSL and merge them with the SearchSourceBuilder already constructed in the index scan operator? That way future extensions don't each need their own end-to-end pipeline.

protocol/src/test/java/org/opensearch/sql/protocol/response/format/JdbcResponseFormatterTest.java
```diff
@@ -207,6 +208,55 @@ void format_server_error_response_due_to_opensearch() {
...
+        new QueryResult(
+            new Schema(
+                ImmutableList.of(
+                    new Column("name", null, STRING), new Column("age", null, INTEGER))),
+            Collections.singletonList(tupleValue(ImmutableMap.of("name", "John", "age", 20))));
+
+    // When no highlights, the highlights field should not appear in the JSON
+    String formatted = formatter.format(response);
+    assertEquals(false, formatted.contains("\"highlights\""));
+  }
```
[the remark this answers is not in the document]
RyanL1997: fixed

integ-test/src/yamlRestTest/resources/rest-api-spec/test/ppl_highlight.yml
RyanL1997: fixed - added more assertions to check the content

RyanL1997:
> Although we only expose highlight via the PPL API today, internally can we make it generic:
Make the highlight part of PPL query as early as possible behind the API, as discussed inhttps://github.com/https://github.com/opensearch-project/sql/issues/5156#issuecomment-3924993361 (highlight function/command or search command parameter?)
If not possible, can we accept arbitrary DSL and merge them with the SearchSourceBuilder already constructed in the index scan operator? That way future extensions don't each need their own end-to-end pipeline.

Hi @dai-chen. Thanks for the suggestion. After checking, I went with option 2 — accept arbitrary DSL and merge it with the `SearchSourceBuilder` at the index scan level.
                                                                                                                                                                          
### Why not option 1 (make highlight part of PPL syntax)?     

The main issue is the response format. The V2 highlight() function returns highlights inline as columns in datarows:
```json
  {
    "schema": [
      { "name": "Tags", "type": "text" },
      { "name": "highlight('Tags')", "type": "nested" }
    ],
    "datarows": [
      ["yeast home-brew", ["<em>yeast</em> home-brew"]]
    ]
  }
```

OSD Explore already knows how to consume highlights from DSL, where they come back as a separate metadata array parallel to the hits. Our current approach matches that shape:
```json
  {
    "schema": [{ "name": "Tags", "type": "text" }],
    "datarows": [["yeast home-brew"]],
    "highlights": [{ "Tags": ["<em>yeast</em> home-brew"] }]
  }
```

If we went with the `highlight()` function, OSD would need to handle a fundamentally different response shape — highlights mixed into datarows as columns instead of a separate highlights array it already consumes today. The request-body approach keeps the schema/datarows untouched and gives the caller (OSD, API users, CLI) full control over highlight config without changing the PPL query text.

### Current design:

The plumbing is now generic. Instead of a highlight-specific pipeline (`Map<String, Object>` → manual HighlightBuilder construction), we:

1. `PPLQueryRequest.getExtraSearchSource()` wraps caller-provided fields into a search-source-compatible JSON string (`{"highlight": {...}`})
2. This String flows through `AbstractPlan` → `ThreadLocal` → index scan as an opaque blob
3. At the index scan level, `applyExtraSearchSource()` parses it via `SearchSourceBuilder.fromXContent()` and selectively merges recognized clauses (currently: `if(extra.highlighter() != null) target.highlighter(extra.highlighter())`)

To add a future extension (e.g. suggest, rescore, post_filter), you would:
- Add the new key to the wrapper in `getExtraSearchSource()`
- Add one merge line in `applyExtraSearchSource()`

No new end-to-end plumbing needed. So please take a look again thanks!

docs/user/ppl/interfaces/endpoint.md
RyanL1997: good point, fixed.

RyanL1997:
Closing this and moving to https://github.com/opensearch-project/sql/pull/5234 for new AST level approach. Thanks. cc @penghuo @dai-chen

Issue #5153, opened 2026-02-17
Investigation: Multisearch union boundary loses UDT types for @timestamp (#5145, #5146, #5147)


ahkcs:
## Summary

Issues #5145, #5146, and #5147 all stem from the same architectural problem: **the multisearch union creates a boundary that breaks the connection between the `@timestamp` alias column and the underlying OpenSearch index data**, and complicates Calcite's plan preparation when nested subfields (e.g., `geo.src`, `machine.os`) are present.

## Affected Queries

| Issue | Query Pattern | Symptom |
|-------|--------------|---------|
| #5145 | `\| multisearch [sub1] [sub2]` (no further processing) | `Error while preparing plan` (500) |
| #5146 | `\| multisearch [sub1] [sub2] \| stats avg(bytes) by span(5m)` | `Error while preparing plan` (500) |
| #5147 | `\| multisearch [sub1] [sub2] \| bin @timestamp span=5m \| stats avg(bytes) by @timestamp` | `@timestamp` becomes `null` for all rows |

### How the plan is constructed

1. **Subsearch parsing** (`AstBuilder.visitSubSearch`, line 185): Each subsearch is wrapped in `projectExceptMeta()`, which adds a `Project(AllFieldsExcludeMeta)` that removes metadata fields (`_id`, `_index`, `_score`, etc.).

2. **@timestamp alias** (`AliasFieldsWrappable.wrapProjectForAliasFields`): The `CalciteLogicalIndexScan` adds `@timestamp` as a projected alias of the `timestamp` field via `relBuilder.projectPlus(relBuilder.alias(relBuilder.field("timestamp"), "@timestamp"))`.

3. **visitMultisearch** (`CalciteRelNodeVisitor.java:2255`): Processes each subsearch, uses `SchemaUnifier` to align schemas, creates a `LogicalUnion`, and adds a `LogicalSort` on `@timestamp` for timestamp interleaving.

4. **Default Project(\*)**: When no explicit command follows multisearch, the top-level `Project(AllFields)` triggers `handleAllFieldsProject` → `tryToRemoveNestedFields()`, which removes nested subfields (fields with dots where the parent is also present, e.g., `geo.coordinates` when `geo` exists).

### Issue #5145: Plan preparation failure

The resulting plan for an index with nested fields (like `opensearch_dashboards_sample_data_logs`) is:

```
LogicalSystemLimit(sort0=[$20], fetch=10000)
[...]

PR #5154, opened 2026-02-17
[BugFix]Fix multisearch UDT type loss through UNION (#5145, #5146, #5147)

### LINKED ISSUE #5145: [BUG] PPL multisearch without further processing fails
**What is the bug?**
When doing a multisearch with two (or more) subsearches, the multisearch fails if no fields are selected either in the subsearch or the multisearch or if no further processing is done (e.g. by using the stats command).

All example queries are runnable in the OpenSearch Playground, using the sample data (specifically the `opensearch_dashboards_sample_data_logs` index/datasource).
The following query:
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US'] 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT']
```
results in the following error (formatted for better readability):
```
{
  "error": {
    "details": "java.sql.SQLException: Error while preparing plan [LogicalSystemLimit(sort0=[$20], dir0=[DESC], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], utc_time=[$15], bytes=[$16], machine=[$17], response=[$20], clientip=[$21], host=[$22], event=[$23], phpmemory=[$25], timestamp=[$26], @timestamp=[$27])
    LogicalSort(sort0=[$27], dir0=[DESC])
      LogicalUnion(all=[true])
        LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], @timestamp=[$33])
          LogicalFilter(condition=[=($13, 'US')])
            LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], _id=[$27], _index=[$28], _score=[$29], _maxscore=[$30], _sort=[$31], _routing=[$32], @timestamp=[$26])
              CalciteLogicalIndexScan(table=[[OpenSearch, opensearch_dashboards_sample_data_logs]])
        LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], @timestamp=[$33])
          LogicalFilter(condition=[=($13, 'IT')])
            LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], _id=[$27], _index=[$28], _score=[$29], _maxscore=[$30], _sort=[$31], _routing=[$32], @timestamp=[$26])
              CalciteLogicalIndexScan(table=[[OpenSearch, opensearch_dashboards_sample_data_logs]])]",
    "type": "RuntimeException"
  },
  "status": 500
}
```

The following two queries work:
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US'] 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT'] 
| fields @timestamp, geo, bytes
```
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US' | fields @timestamp, geo, bytes] 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT' | fields @timestamp, geo, bytes] 
```

Trying to select all fields using `*` does not work. The following query:
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US'] 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT'] 
| fields `*`
```
results in the following error (formatted for better readability):
```
{
  "error": {
    "details": "java.sql.SQLException: Error while preparing plan [LogicalSystemLimit(sort0=[$27], dir0=[DESC], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalSort(sort0=[$27], dir0=[DESC])
    LogicalUnion(all=[true])
      LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], @timestamp=[$33])
        LogicalFilter(condition=[=($13, 'US')])
          LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], _id=[$27], _index=[$28], _score=[$29], _maxscore=[$30], _sort=[$31], _routing=[$32], @timestamp=[$26])
            CalciteLogicalIndexScan(table=[[OpenSearch, opensearch_dashboards_sample_data_logs]])
      LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], @timestamp=[$33])
        LogicalFilter(condition=[=($13, 'IT')])
          LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], _id=[$27], _index=[$28], _score=[$29], _maxscore=[$30], _sort=[$31], _routing=[$32], @timestamp=[$26])
            CalciteLogicalIndexScan(table=[[OpenSearch, opensearch_dashboards_sample_data_logs]])]",
    "type": "RuntimeException"
  },
  "status": 500
}
```

1. Go to [the OpenSearch playground](https://playground.opensearch.org/).
2. Use the queries provided above either in the Dev Console or the Query Workbench.

In the absence of any field selection, select all the fields in the result.

 - OS: N/A
 - Version 3.4.0
 - Plugins: N/A

N/A

The following query also works:
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US'] 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT'] 
| stats avg(bytes)
```

### LINKED ISSUE #5146: [BUG] PPL span-expression used after multisearch fails
**What is the bug?**
Using a span-expression in the `stats` command after a `multisearch` command fails.

All example queries are runnable in the OpenSearch Playground, using the sample data (specifically the `opensearch_dashboards_sample_data_logs` index/datasource).

The following query:
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US'] 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT'] 
| stats avg(bytes) by span(5m)
```
fails with the following error (formatted for better readability):
```
{
  "error": {
    "details": "java.sql.SQLException: Error while preparing plan [LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(avg(bytes)=[$1], span(5m)=[$0])
    LogicalAggregate(group=[{1}], avg(bytes)=[AVG($0)])
      LogicalProject(bytes=[$16], span(5m)=[SPAN($27, 5, 'm')])
        LogicalFilter(condition=[IS NOT NULL($27)])
          LogicalSort(sort0=[$27], dir0=[DESC])
            LogicalUnion(all=[true])
              LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], @timestamp=[$33])
                LogicalFilter(condition=[=($13, 'US')])
                  LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], _id=[$27], _index=[$28], _score=[$29], _maxscore=[$30], _sort=[$31], _routing=[$32], @timestamp=[$26])
                    CalciteLogicalIndexScan(table=[[OpenSearch, opensearch_dashboards_sample_data_logs]])
              LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], @timestamp=[$33])
                LogicalFilter(condition=[=($13, 'IT')])
                  LogicalProject(referer=[$0], request=[$1], agent=[$2], extension=[$3], memory=[$4], ip=[$5], index=[$6], message=[$7], url=[$8], tags=[$9], geo=[$10], geo.coordinates=[$11], geo.dest=[$12], geo.src=[$13], geo.srcdest=[$14], utc_time=[$15], bytes=[$16], machine=[$17], machine.os=[$18], machine.ram=[$19], response=[$20], clientip=[$21], host=[$22], event=[$23], event.dataset=[$24], phpmemory=[$25], timestamp=[$26], _id=[$27], _index=[$28], _score=[$29], _maxscore=[$30], _sort=[$31], _routing=[$32], @timestamp=[$26])
                    CalciteLogicalIndexScan(table=[[OpenSearch, opensearch_dashboards_sample_data_logs]])
]",
    "type": "RuntimeException"
  },
  "status": 500
}
```

1. Go to [the OpenSearch playground](https://playground.opensearch.org/).
2. Use the queries provided above either in the Dev Console or the Query Workbench.

The result of the above query should be the same as if the following query was run:
```
source=opensearch_dashboards_sample_data_logs 
| where geo.src='US' or geo.src='IT' 
| stats avg(bytes) by span(5m)
```

 - OS: N/A
 - Version 3.4.0
 - Plugins: N/A

N/A

N/A

### LINKED ISSUE #5147: [BUG] PPL `bin` used after `multisearch` has no effect
**What is the bug?**
The `bin` command has no effect when used after a `multisearch` command.

All example queries are runnable in the OpenSearch Playground, using the sample data (specifically the opensearch_dashboards_sample_data_logs index/datasource).

The following query:
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US'] 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT'] 
| bin @timestamp span=5m 
| stats avg(bytes) by @timestamp
```
gives the following response:
```json
{
  "schema": [
    {
      "name": "avg(bytes)",
      "type": "double"
    },
    {
      "name": "@timestamp",
      "type": "string"
    }
  ],
  "datarows": [
    [
      5713.0848985725015,
      null
    ]
  ],
  "total": 1,
  "size": 1
}
```
which is almost identical to the following query:
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US']
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT'] 
| stats avg(bytes)
```
and its response:
```json
{
  "schema": [
    {
      "name": "avg(bytes)",
      "type": "double"
    }
  ],
  "datarows": [
    [
      5713.0848985725015
    ]
  ],
  "total": 1,
  "size": 1
}
```

The only difference I can see is that the first query also returns the `@timestamp` field, which is wierdly `null`. I think this might have something to do with the issue. The following query and it response also illustrate that the `bin` command followed by nothing else seems to set the value of the `@timestamp` field to `null`.
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US']
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT'] 
| fields @timestamp, geo, bytes 
|  bin @timestamp span=5m
```
```json
{
  "schema": [
    {
      "name": "geo",
      "type": "struct"
    },
    {
      "name": "bytes",
      "type": "bigint"
    },
    {
      "name": "@timestamp",
      "type": "string"
    }
  ],
  "datarows": [
    [
      {
        "srcdest": "US:TR",
        "src": "US",
        "coordinates": {
          "lat": 34.80434056,
          "lon": -96.6712775
        },
        "dest": "TR"
      },
      2950,
      null
    ],
    [
      {
        "srcdest": "US:CN",
        "src": "US",
        "coordinates": {
          "lat": 35.82149222,
          "lon": -81.61073639
        },
        "dest": "CN"
      },
      209,
      null
    ],
//[...]
  "total": 1331,
  "size": 1331
}
```

1. Go to [the OpenSearch playground](https://playground.opensearch.org/).
2. Use the queries provided above either in the Dev Console or the Query Workbench.

The query:
```
| multisearch 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='US'] 
    [search source=opensearch_dashboards_sample_data_logs | where geo.src='IT'] 
| bin @timestamp span=5m 
| stats avg(bytes) by @timestamp
```
should have the same behaviour as the following query:
```
source=opensearch_dashboards_sample_data_logs 
| where geo.src='US' or geo.src='IT' 
| bin @timestamp span=5m 
| stats avg(bytes) by @timestamp
```

 - OS: N/A
 - Version 3.4.0
 - Plugins: N/A

N/A

N/A

ahkcs:
## Summary

- Override `leastRestrictive()` in `OpenSearchTypeFactory` to preserve UDT types (e.g., `ExprTimestampType`) through set operations (UNION ALL) used by multisearch
- Without this fix, Calcite's default `leastRestrictive()` downgrades UDT-wrapped types to their underlying SQL type (e.g., `VARCHAR`), breaking downstream operators that rely on type checks
- Add integration tests verifying span, bin, and bin+stats work correctly after multisearch

## Issues Resolved

Resolves #5145, resolves #5146, resolves #5147

See investigation details: #5153

## Root Cause

`OpenSearchTypeFactory` (extends `JavaTypeFactoryImpl`) did not override `leastRestrictive()`. When Calcite builds UNION ALL for multisearch, it calls `leastRestrictive()` on column types from both branches. Both branches have `ExprTimestampType` (a UDT backed by `SqlTypeName.VARCHAR`), but the default implementation creates a plain `BasicSqlType(VARCHAR)`, **losing the UDT wrapper**.

This caused:
- **#5146**: `SpanFunction.SpanImplementor` checks `fieldType instanceof ExprSqlType` → fails for plain VARCHAR → `"Unsupported expr type: STRING"`
- **#5147**: `BinnableField.isTimeBasedType()` can't recognize plain VARCHAR as time-based → bin produces null timestamps
- **#5145**: Same type loss causes plan preparation failures depending on downstream operators

## Fix

When all input types share the same UDT (e.g., all are `EXPR_TIMESTAMP`), the result retains the UDT wrapper. Nullability is handled correctly — if any input is nullable, the result is nullable. Otherwise, delegates to the default Calcite implementation.

## Test Plan

- [x] `testMultisearchWithoutFurtherProcessing` — basic multisearch returns all 51 rows
- [x] `testMultisearchWithSpanExpression` — `stats avg(value) by span(@timestamp, 5m)` after multisearch (36 rows verified)
- [x] `testMultisearchBinTimestamp` — `bin @timestamp span=5m` produces non-null timestamps (36 rows verified)
- [x] `testMultisearchBinAndStats` — `bin + stats` produces correct time buckets (36 rows verified)
- [x] Existing `testMultisearchWithTimestampInterleaving` schema updated from `"string"` to `"timestamp"` (correct behavior)
- [x] All existing multisearch integration tests pass

### DIFF
--- a/core/src/test/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactoryTest.java
+++ b/core/src/test/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactoryTest.java
@@ -89,10 +89,9 @@ public class OpenSearchTypeFactoryTest {
 
     RelDataType result = TYPE_FACTORY.leastRestrictive(List.of(udt, plain));
 
-    // Falls back to super.leastRestrictive which may return a plain type or null
-    if (result != null) {
-      assertEquals(SqlTypeName.VARCHAR, result.getSqlTypeName());
-    }
+    // Falls back to super.leastRestrictive — both backed by VARCHAR, so result is non-null
+    assertNotNull(result);
+    assertEquals(SqlTypeName.VARCHAR, result.getSqlTypeName());
   }
 
   @Test
@@ -102,10 +101,9 @@ public class OpenSearchTypeFactoryTest {
 
     RelDataType result = TYPE_FACTORY.leastRestrictive(List.of(timestamp, date));
 
-    // Different UDTs — falls back to super.leastRestrictive
-    if (result != null) {
-      assertEquals(SqlTypeName.VARCHAR, result.getSqlTypeName());
-    }
+    // Different UDTs — falls back to super.leastRestrictive, both backed by VARCHAR
+    assertNotNull(result);
+    assertEquals(SqlTypeName.VARCHAR, result.getSqlTypeName());
   }
 
   @Test


core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -377,6 +378,44 @@ public static boolean isNumericType(RelDataType fieldType) {
...
+        boolean allSameUdt =
+            types.stream()
+                .allMatch(
+                    t ->
+                        t instanceof AbstractExprRelDataType<?> udt
+                            && udt.getUdt() == firstUdt.getUdt());
+        if (allSameUdt) {
+          boolean anyNullable = types.stream().anyMatch(RelDataType::isNullable);
+          if (anyNullable && !first.isNullable()) {
+            return firstUdt.createWithNullability(this, true);
```
Swiddis: nit: Could do this in a single pass with a plain for loop instead of splitting the match checks and null checks.

```java
@Override
public @Nullable RelDataType leastRestrictive(List<RelDataType> types) {
  if (types.size() > 1) {
    RelDataType first = types.get(0);
    if (first instanceof AbstractExprRelDataType<?> firstUdt) {
      boolean anyNullable = false;
      for (RelDataType t : types) {
        if (t instanceof AbstractExprRelDataType<?> udt
            && udt.getUdt() == firstUdt.getUdt()) {
          anyNullable |= t.isNullable();
        } else {
          return super.leastRestrictive(types); // If we hit a non-match, fallback
        }
      }
      if (anyNullable && !first.isNullable()) {
        return firstUdt.createWithNullability(this, true);
      }
      return first;
    }
  }
  return super.leastRestrictive(types);
}
```

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
```diff
@@ -344,6 +344,203 @@ public void testMultisearchCrossIndexFieldSelection() throws IOException {
...
+        rows("E", 2001, "2025-08-01 04:00:00"),
+        rows("E", 2003, "2025-08-01 01:00:00"),
+        rows("E", 2005, "2025-07-31 20:45:00"),
+        rows("E", 2007, "2025-07-31 16:00:00"),
+        rows("E", 2009, "2025-07-31 12:30:00"),
+        rows("E", 2011, "2025-07-31 08:00:00"),
+        rows("E", 2013, "2025-07-31 04:30:00"),
+        rows("E", 2015, "2025-07-31 01:00:00"),
+        rows("E", 2017, "2025-07-30 21:30:00"),
+        rows("E", 2019, "2025-07-30 18:00:00"));
```
Swiddis: these test assertions are pretty long, is that the minimal reproducing example? would 1-day intervals be clearer?

ahkcs:
> How are we confident that this is always correct outside of the multisearch special case?
> 
> My understanding is leastRestrictive is assumed to convert types to the plain SQL types, so keeping around UDTs in some circumstances might quietly break other code's assumptions. But maybe we only hit this branch with multiple types if we're in an expression that needs it, or maybe we don't actually have such an assumption from any caller code.

The leastRestrictive override only activates when all input types are the same `AbstractExprRelDataType` with the same UDT — in that case, preserving the UDT is the semantically correct result (it's the least restrictive type that encompasses all inputs). When any type differs, it falls back to Calcite's default implementation. This is consistent with how other type systems handle UDTs in union operations.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultisearchCommandIT.java
ahkcs: Updated

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
ahkcs: Updated

core/src/test/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactoryTest.java
```diff
@@ -0,0 +1,131 @@
...
+  @Test
+  public void testLeastRestrictiveFallsBackForMixedUdtAndNonUdt() {
+    RelDataType udt = TYPE_FACTORY.createUDT(ExprUDT.EXPR_TIMESTAMP);
+    RelDataType plain = TYPE_FACTORY.createSqlType(SqlTypeName.VARCHAR);
+
+    RelDataType result = TYPE_FACTORY.leastRestrictive(List.of(udt, plain));
+
+    // Falls back to super.leastRestrictive which may return a plain type or null
+    if (result != null) {
+      assertEquals(SqlTypeName.VARCHAR, result.getSqlTypeName());
```
RyanL1997: nit: We should know which outcome is expected and assert it unconditionally. If it should be `null`, assert `assertNull`. If it should be `VARCHAR`, drop the if.

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -377,6 +378,43 @@ public static boolean isNumericType(RelDataType fieldType) {
...
+   *     super#leastRestrictive})
+   */
+  @Override
+  public @Nullable RelDataType leastRestrictive(List<RelDataType> types) {
+    if (types.size() > 1) {
+      RelDataType first = types.get(0);
+      if (first instanceof AbstractExprRelDataType<?> firstUdt) {
+        boolean anyNullable = false;
+        for (RelDataType t : types) {
+          if (t instanceof AbstractExprRelDataType<?> udt && udt.getUdt() == firstUdt.getUdt()) {
```
RyanL1997, line 403: This uses reference equality (`==`). This is fine if `getUdt()` returns an enum value, but could you confirm that's always the case? If it ever returns a non-enum object, this should be `.equals()` instead.

core/src/test/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactoryTest.java
ahkcs: Good catch — you're right, both types are backed by `SqlTypeName.VARCHAR` so `super.leastRestrictive` will always return a non-null result here. I'll remove the conditional and assert unconditionally with assertNotNull. Will fix the same pattern in `testLeastRestrictiveFallsBackForDifferentUdts` too.

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
ahkcs: Yes, `getUdt()` always returns an `ExprUDT` enum value (declared at OpenSearchTypeFactory.java:73), and the field is typed as `ExprUDT` in `AbstractExprRelDataType`. So == is safe here — and actually preferred for enum comparison in Java since it's null-safe and provides compile-time type checking.

PR #5034, opened 2026-01-10
Adding PPL `convert` command with numeric and memory conversion functions

aalva500-prog:
### Description
Implement a `convert` command in PPL for numeric and memory data type conversions. This command provides a streamlined interface for converting field values to numeric types, reducing the need for multiple eval statements and improving query readability.

**Key Features:**
- **Numeric Conversions**: `auto()`, `num()`, `rmcomma()`, `rmunit()` , `memk()` functions
- **Field Preservation**: AS clause support for creating new fields while preserving originals
- **Multiple Conversions**: Support for multiple conversion functions in a single command

**Syntax:**
```
... | convert <convert-function>(<field>) [AS <field>] [<convert-function>(<field>) [AS <field>]]...
```

**Example Usage:**
```sql
source=accounts | convert auto(balance)
source=data | convert rmcomma(price) AS clean_price
source=metrics | convert rmunit(duration) num(count)
source=processor | convert memk(virtual_memory)
```

**Future Enhancements:**
This initial implementation focuses on numeric and memory conversions. Time conversions (`ctime()`, `dur2sec()`, etc.) and wildcard support will be implemented in subsequent PRs.

This PR partially addresses the current limitation where users must chain multiple evaluation functions or write complex expressions for common conversion operations.

### Related Issues
Resolves partially (only numeric and memory conversions in this PR) #5001 

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
- [x] New functionality has javadoc added.
- [ ] New functionality has a user manual doc added.
- [x] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #5001, opened 2025-12-26
[RFC] PPL `convert` Command


aalva500-prog:
## Problem Statement

PPL currently lacks a unified command for converting field values between different data types and formats. Users frequently need to convert data for analysis, calculations, and reporting purposes, but must rely on chaining multiple evaluation functions or writing complex expressions. Common conversion needs include:

- Converting string representations to numeric values
- Transforming timestamp formats for readability or calculations
- Converting duration strings to seconds for mathematical operations
- Normalizing memory values to consistent units
- Extracting numeric values from mixed alphanumeric strings

This creates verbose queries and increases the learning curve for users who need to remember multiple function names and their specific syntax.

## Current State

Users must currently use evaluation functions within the `eval` command to perform conversions:

```sql
source=logs | eval balance_num = tonumber(balance) | eval time_readable = strftime(_time, "%H:%M:%S")
```

Limitations of the current approach:
- Requires knowledge of specific function names scattered across documentation
- No unified interface for common conversion operations
- Difficult to apply conversions to multiple fields efficiently
- No support for specialized conversions like duration-to-seconds or memory unit normalization
- Verbose syntax for simple operations

## Long-Term Goals

- Provide a unified `convert` command that consolidates common data type conversion operations
- Simplify PPL queries by reducing the need for multiple eval statements
- Support bulk field conversions with wildcard patterns
- Enable in-place field conversion or creation of new fields via AS clause
- Maintain high performance with large datasets through optimized conversion implementations
- Ensure the solution is extensible for future conversion function additions
- Align with SQL plugin architecture and PPL command patterns
- Support 90% confidence that this addresses user conversion needs based on common log analysis patterns

## Proposal

Implement a `convert` command in PPL with the following conversion functions:

1. **Numeric Conversions**
   - `auto()` - Automatically converts fields to numbers using best-fit heuristics
   - `num()` - Converts to numbers, removing non-convertible values
   - `rmcomma()` - Removes commas from numeric strings
   - `rmunit()` - Extracts leading numbers and removes trailing text

2. **Time Conversions**
   - `ctime()` - Converts UNIX timestamps to human-readable format
   - `mktime()` - Converts human-readable time strings to epoch time
   - `mstime()` - Converts MM:SS.SSS format to seconds
   - `dur2sec()` - Converts duration format [D+]HH:MM:SS to seconds

3. **Memory Conversions**
   - `memk()` - Converts memory values (with k/m/g suffixes) to kilobytes

[...]

PR #5157, opened 2026-02-19
feat: Implement PPL convert command with 5 conversion functions

ritvibhatt:
### Description
Copy of #5034, resolves merge conflicts

### Related Issues
Resolves partially (only numeric and memory conversions in this PR) https://github.com/opensearch-project/sql/issues/5001

### Check List


core/src/main/java/org/opensearch/sql/expression/function/udf/AutoConvertFunction.java
```diff
@@ -0,0 +1,28 @@
...
+
+/** PPL auto() conversion function. */
+public class AutoConvertFunction extends BaseConversionUDF {
+
+  public AutoConvertFunction() {
+    super(AutoConvertFunction.class);
+  }
+
+  public static Object convert(Object value) {
+    return new AutoConvertFunction().convertValue(value);
```
penghuo: Each static convert() method creates a new UDF instance per row (new AutoConvertFunction().convertValue(value)). Additionally, AutoConvertFunction.applyConversion creates new NumConvertFunction() per row at line 26. For large datasets this produces 2 object allocations per row per convert call. Since convertValue and applyConversion are stateless, use a cached singleton instance instead (e.g. private static final AutoConvertFunction INSTANCE = new AutoConvertFunction()). Same pattern applies to all 5 UDF classes.

docs/user/ppl/cmd/convert.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Parameters
+
+The `convert` command supports the following parameters.
+
+| Parameter | Required/Optional | Description |
+| --- | --- | --- |
+| `<convert-function>` | Required | One of the conversion functions: `auto()`, `num()`, `rmcomma()`, `rmunit()`, `memk()`, or `none()`. |
+| `<field>` | Required | Single field name to convert, or `*` to convert all fields. |
+| `AS <field>` | Optional | Create new field with converted value, preserving original field. |
```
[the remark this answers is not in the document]
penghuo: +1. 
suggestion, "remove the wildcard claim from docs"

core/src/main/java/org/opensearch/sql/expression/function/udf/BaseConversionUDF.java
```diff
@@ -0,0 +1,188 @@
...
+/** Base class for PPL conversion functions. */
+@Log4j2
+public abstract class BaseConversionUDF extends ImplementorUDF {
+
+  private static final Pattern COMMA_PATTERN = Pattern.compile(",");
+  private static final Pattern LEADING_NUMBER_WITH_UNIT_PATTERN =
+      Pattern.compile("^([+-]?(?:\\d+\\.?\\d*|\\.\\d+)(?:[eE][+-]?\\d+)?)(.*)$");
+  private static final Pattern CONTAINS_LETTER_PATTERN = Pattern.compile(".*[a-zA-Z].*");
+  private static final Pattern STARTS_WITH_SIGN_OR_DIGIT = Pattern.compile("^[+-]?[\\d.].*");
+  private static final Pattern MEMK_PATTERN = Pattern.compile("^([+-]?\\d+\\.?\\d*)([kmgKMG])?$");
```
penghuo, line 34: memk("512 m") (with space before unit) returns null. Is this the intended Splunk compatible behavior?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -1202,6 +1203,42 @@ public UnresolvedPlan visitFillNullValueAllFields(
...
+            .collect(Collectors.toList());
+    return new Convert(conversions);
+  }
+
+  private Let buildConversion(OpenSearchPPLParser.ConvertFunctionContext funcCtx) {
+    if (funcCtx.fieldExpression().isEmpty()) {
+      throw new IllegalArgumentException("Convert function requires a field argument");
+    }
+
+    String functionName = funcCtx.functionName.getText();
```
penghuo, line 1227: convert currently accepts any identifier as a function name and forwards it without semantic validation, so unsupported/non-convert functions (for example convert lower(field)) are accepted even though the command contract documents a fixed set (auto/num/rmcomma/rmunit/memk/none). This makes convert behave like a generic single-arg eval and can silently produce non-numeric output types, so function names should be rejected early unless they are in the supported conversion list.

core/src/main/java/org/opensearch/sql/expression/function/udf/AutoConvertFunction.java
ritvibhatt: Updated to use cached singleton instance instead of creating new instance every row for the 5 UDF classes

core/src/main/java/org/opensearch/sql/expression/function/udf/BaseConversionUDF.java
ritvibhatt: Yes, Splunk also returns null when there is a space between the value and unit

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
ritvibhatt: Updated to only accept supported convert functions

penghuo (dismissed):
Thanks!

Please fix merge conflict.

PR #5162, opened 2026-02-20
[Feature] Add grammar bundle generation API for PPL language features

mengweieric:
### Description
Implements the backend grammar metadata API for PPL autocomplete support.

This endpoint serves a versioned grammar bundle containing ANTLR metadata required for downstream consumers (for example OpenSearch Dashboards) to reconstruct a functional PPL lexer/parser at runtime using antlr4ng interpreter APIs. This enables full client-side parsing/autocomplete with zero per-keystroke server calls, while keeping backend grammar as the source of truth.

#### What the bundle contains:
- Serialized lexer and parser ATNs via `ATNSerializer.serialize().toArray()`, compatible with antlr4ng `ATNDeserializer.deserialize()`
- Token vocabulary (literal/symbolic names), lexer/parser rule names, channel names, and mode names for interpreter reconstruction
- `tokenDictionary`, `ignoredTokens`, and `rulesToVisit` for autocomplete behavior
- `grammarHash` (ATNs + lexer/parser rule names + literal/symbolic vocabulary + ANTLR version) for client-side change detection
- `bundleVersion` and `antlrVersion` for compatibility validation

#### Backend behavior:
- Synchronized lazy initialization (bundle built on first request and cached for node lifecycle)
- Deterministic output (same plugin JAR -> same bundle shape/content)
- Endpoint marked `@ExperimentalApi`
- Grammar endpoint now performs authorization via PPL transport action before serving bundle data (aligns with existing PPL permission model)

#### Also included:
- REST API spec registration for `ppl.grammar` + YAML REST response-shape test
- Unit tests for grammar REST handler (including authorization-failure path)
- Security integration tests for grammar endpoint access with/without PPL permission
- `THIRD-PARTY` updated to reflect ANTLR 4.13.2

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/5218

### Check List

### DIFF
--- a/ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
+++ b/ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
@@ -162,7 +162,7 @@ public final class PPLGrammarBundleBuilder {
             "integerLiteral",
             "decimalLiteral",
             "keywordsCanBeId",
-            "renameClasue",
+            "renameClause",
             "qualifiedName",
             "tableQualifiedName",
             "wcQualifiedName",


ahkcs:
QQ:

What would be the estimated payload size when using this API? Should we consider adding compression or `Accept-Encoding` handling to compress the payload?

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  private static void updateDigest(MessageDigest digest, int[] data) {
+    for (int v : data) {
+      digest.update((byte) (v >>> 24));
+      digest.update((byte) (v >>> 16));
+      digest.update((byte) (v >>> 8));
+      digest.update((byte) (v));
+    }
+  }
```
Swiddis, line 235: suggestion (perf): Directly load the buffer?

`bb = ByteBuffer.allocate(4 * data.length)`, `bb.asIntBuffer().put(data)`, `digest.update(bb)`

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,120 @@
...
+
+    // Parser ATN & metadata
+    builder.field("parserSerializedATN", bundle.getParserSerializedATN());
+    builder.field("parserRuleNames", bundle.getParserRuleNames());
+
+    // Vocabulary
+    builder.field("literalNames", bundle.getLiteralNames());
+    builder.field("symbolicNames", bundle.getSymbolicNames());
+
+    builder.endObject();
```
Swiddis, line 135: thought: I dislike that we're going to be directly exposing ANTLR implementation details like this.

It means any frontends wanting to build their own autocomplete (CLIs, CWL Console, anything else other than OSD) will need to introduce ANTLR dependencies. There's also just generally a lot of validation and conversion happening between AST/RelNode that lives outside of the antlr directly, so frontends still need to be pretty highly aware of how the backend works and per-command semantics. Did we already research & eliminate the option of making a custom format that encodes more useful information?

Guessing there isn't a better way or format [without some heavy lifting](https://github.com/antlr/antlr4/issues/233) :/

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,120 @@
...
+
+    return channel -> {
+      try {
+        GrammarBundle bundle = getOrBuildBundle();
+        XContentBuilder builder = channel.newBuilder();
+        serializeBundle(builder, bundle);
+        channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
+      } catch (Exception e) {
+        log.error("Error building or serializing PPL grammar", e);
+        channel.sendResponse(new BytesRestResponse(channel, RestStatus.INTERNAL_SERVER_ERROR, e));
```
Swiddis: issue: Very wide catch returning 500s

Should be using the existing infra for error codes instead of hardcoding `INTERNAL_SERVER_ERROR`, or catch a more specific exception for serialization failures. The "error building or serializing" context also seems relevant compared to returning a raw exception message.

Swiddis (commented):
Technically fine but I'm not very happy with the solution overall.

One issue for error handling, can approve after that's fixed.

Swiddis:
> What would be the estimated payload size when using this API? Should we consider add compression or Accept-Encoding handling to compress the payload?

Tested locally (`xh localhost:9200/_plugins/_ppl/_grammar -h`), not super light but it could be worse, based on the actual ATN contents the cache's memory footprint is going to be around 280 KB.

```http
HTTP/1.1 200 OK
Content-Encoding: gzip
Content-Length: 68514
Content-Type: application/json; charset=UTF-8
X-Opensearch-Version: OpenSearch/3.6.0-SNAPSHOT (opensearch)
```

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,120 @@
...
+      }
+    };
+  }
+
+  // Thread-safe lazy initialization with double-checked locking.
```
dai-chen: Is Atomic reference or simply synchronized the method easier?

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/GrammarBundle.java
```diff
@@ -0,0 +1,58 @@
...
+package org.opensearch.sql.ppl.autocomplete;
+
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+/** Serialized ANTLR grammar data served by {@code GET /_plugins/_ppl/_grammar}. */
+@Value
+@Builder
+public class GrammarBundle {
```
dai-chen, line 19: Just curious can this and the endpoint reused by SQL in future? Or we need duplicate code and endpoint?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,120 @@
...
+/*
+ * REST handler for {@code GET /_plugins/_ppl/_grammar}.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+@Log4j2
+public class RestPPLGrammarAction extends BaseRestHandler {
+
+  private static final String ENDPOINT_PATH = "/_plugins/_ppl/_grammar";
```
dai-chen, line 39: We need to update endpoint documentation?

plugin/src/test/java/org/opensearch/sql/plugin/rest/RestPPLGrammarActionTest.java
```diff
@@ -0,0 +1,247 @@
...
+    assertEquals("/_plugins/_ppl/_grammar", action.routes().get(0).getPath());
+  }
+
+  @Test
+  public void testGetGrammar_ReturnsBundle() throws Exception {
+    FakeRestRequest request =
+        new FakeRestRequest.Builder(NamedXContentRegistry.EMPTY)
+            .withMethod(RestRequest.Method.GET)
+            .withPath("/_plugins/_ppl/_grammar")
+            .build();
```
dai-chen: Each test method is a little long. Maybe extract such duplicate to private method while maintaining the readability?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,120 @@
...
+  public List<Route> routes() {
+    return ImmutableList.of(new Route(GET, ENDPOINT_PATH));
+  }
+
+  @Override
+  protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client)
+      throws IOException {
+
+    return channel -> {
+      try {
```
penghuo, line 56: This endpoint bypasses the transport action layer that other PPL handlers use (e.g., `RestPPLQueryAction` → `PPLQueryAction`). While the `/_plugins/_ppl/` route prefix is likely covered by the security plugin's route-level rules, the inconsistency means this endpoint's access control depends entirely on that assumption rather than explicit action-level permissions. Worth confirming that the security plugin's default config protects this path, or add a transport action wrapper for consistency.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,120 @@
+/*
```
penghuo, line 1: No integration test covers this new endpoint. Without an integ-test or yamlRestTest, there's no verification that the endpoint works when the plugin is loaded in a real OpenSearch node (route registration, serialization over the wire, security plugin interaction). Add at least a basic integ-test that hits the endpoint and validates the response shape.

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/GrammarBundle.java
```diff
@@ -0,0 +1,58 @@
...
+  @NonNull private String bundleVersion;
+
+  /** ANTLR runtime version used to generate the grammar. */
+  @NonNull private String antlrVersion;
+
+  /** SHA-256 hash of the serialized ATN data. Clients may use this to detect grammar changes. */
+  @NonNull private String grammarHash;
+
+  /** Serialized lexer ATN as int array (ATNSerializer output). */
+  @NonNull private int[] lexerSerializedATN;
```
penghuo, line 34: `@Value` with `int[]` and `String[]` fields doesn't provide true immutability — callers can mutate the backing arrays. Since `RestPPLGrammarAction` caches a single `GrammarBundle` instance and serves it to all requests, any future code path that obtains the arrays and mutates them would corrupt the global cache. Consider defensive copies in the builder's `build()` method, or switch to `List<Integer>` / `List<String>`.

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+        .literalNames(literalNames)
+        .symbolicNames(symbolicNames)
+        .build();
+  }
+
+  private static int resolveStartRuleIndex(String[] ruleNames) {
+    int idx = Arrays.asList(ruleNames).indexOf("root");
+    return Math.max(idx, 0);
+  }
+
```
penghuo, line 198: The grammar hash includes lexer ATN + parser ATN + ANTLR version, but not the vocabulary (literal/symbolic names) or rule names. If a grammar change only alters a token's literal name without changing ATN structure, the hash won't change and clients won't know to refresh. Is this intentional, or should vocabulary data be included in the hash?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+@Log4j2
+public class RestPPLGrammarAction extends BaseRestHandler {
+
+  private static final String ENDPOINT_PATH = "/_plugins/_ppl/_grammar";
+
+  // Lazy-initialized singleton bundle (built once per JVM lifecycle)
```
penghuo: Is this endpoint path registered in the OpenSearch API specification? The PR checklist shows "API changes companion pull request" is unchecked. Since this endpoint will be consumed by OpenSearch Dashboards for autocomplete, it should be documented in the API spec for cross-version compatibility validation.

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
mengweieric: Fixed. Grammar hash now includes ATNs + lexer/parser rule names + literal/symbolic vocabulary + ANTLR version.

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/GrammarBundle.java
mengweieric: Fixed. Added defensive copies in builder/getters and unmodifiable map for token dictionary.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
mengweieric: Fixed. Added YAML REST API spec + integration test for /_plugins/_ppl/_grammar response shape.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
mengweieric: Fixed. The handler now logs context and returns via OpenSearch exception serialization (new BytesRestResponse(channel, e)) instead of hardcoded INTERNAL_SERVER_ERROR

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
mengweieric: Fixed. Switched to synchronized lazy initialization for simplicity and thread safety.

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/GrammarBundle.java
mengweieric: Good question. The payload is grammar-agnostic and reusable for SQL. This PR is intentionally PPL-only since there’s no SQL grammar consumer/endpoint yet. If SQL autocomplete needs it, we can extract shared bundle/handler/builder abstractions and add /_plugins/_sql/_grammar instead of duplicating logic.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
mengweieric: Fixed. Added endpoint docs and stabilized doctest example output.

plugin/src/test/java/org/opensearch/sql/plugin/rest/RestPPLGrammarActionTest.java
mengweieric: Fixed. Extracted duplicated request setup into a helper to keep tests shorter and readable.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
mengweieric: Fixed. Endpoint is now registered in REST API spec as ppl.grammar.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
mengweieric: Good point. Fixed in this PR: /_plugins/_ppl/_grammar now performs an explicit transport-action authorization check via PPLQueryAction before serving the bundle, so access control is no longer only route-prefix-based. I also added security tests for grammar endpoint access with and without PPL permission.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,154 @@
...
+  @Override
+  protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client)
+      throws IOException {
+
+    return channel -> {
+      try {
+        authorizeRequest(
+            client,
+            new ActionListener<>() {
+              @Override
```
penghuo: The response contains two serialized ATN int arrays plus vocabulary, rule names, and dictionaries — likely 200KB–1MB for a grammar this size. Since `grammarHash` is already included for client-side change detection, have you considered supporting `ETag`/`If-None-Match` based on `grammarHash` for HTTP caching? Dashboards could skip re-downloading the bundle if the hash has not changed. Not blocking for the initial experimental release, but worth considering for a follow-up.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
```diff
@@ -0,0 +1,154 @@
...
+  @VisibleForTesting
+  protected GrammarBundle buildBundle() {
+    return new PPLGrammarBundleBuilder().build();
+  }
+
+  /** Invalidate the cached bundle, forcing a rebuild on the next request. */
+  @VisibleForTesting
+  protected synchronized void invalidateCache() {
+    cachedBundle = null;
+  }
```
penghuo: PPLGrammarBundleBuilder has no constructor and no instance state — its build() method creates a lexer/parser locally, does all the work, and returns a GrammarBundle. So it's  effectively a stateless utility class.

  Option : Static holder pattern (lazy, thread-safe, no synchronization on hot path)

  public class PPLGrammarBundleBuilder {

      private static class BundleHolder {
          static final GrammarBundle INSTANCE = new PPLGrammarBundleBuilder().build();
      }

      public static GrammarBundle getBundle() {
          return BundleHolder.INSTANCE;
      }

      // existing build logic stays private
      private GrammarBundle build() { ... }
  }

  Caller: PPLGrammarBundleBuilder.getBundle()

mengweieric: Implemented, thanks for the suggestion.

I refactored PPLGrammarBundleBuilder to use a static holder singleton (getBundle()), made it stateless with a private constructor, and updated RestPPLGrammarAction to consume it directly (removed handler-level cache/synchronization). Tests were updated and all relevant checks are passing.

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryRequest.java
```diff
@@ -119,7 +119,16 @@ public String getRequest() {
...
+    return path != null && path.endsWith("/_explain");
+  }
+
+  /**
+   * Check if request is for grammar metadata endpoint.
+   *
+   * @return true if it is a grammar metadata request
+   */
+  public boolean isGrammarRequest() {
+    return path != null && path.endsWith("/_grammar");
```
ahkcs, line 131: path.endsWith("/_grammar") would match a hypothetical future endpoint like `/_plugins/_sql/_grammar`
Should we consider matching against the full path constant like `/_ppl/_grammar`

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+   */
+  private static int[] buildRulesToVisit(String[] ruleNames) {
+    List<String> ruleNamesToVisit =
+        Arrays.asList(
+            "statsFunctionName",
+            "takeAggFunction",
+            "integerLiteral",
+            "decimalLiteral",
+            "keywordsCanBeId",
+            "renameClasue",
```
ahkcs: nit: "renameClasue" seems to be a typo, in the .g4 file as well, should be updated in both places

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryRequest.java
Swiddis: We've done the same thing historically: https://github.com/opensearch-project/sql/blob/94649fbe781741d5f9b15c30b4e0e5f5811abc86/sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java#L118

It's a bit sloppy, but it works for the most part lol. Should maybe revisit using proper path routing some time, I'd like that as it'd make it easier to navigate how our code responds to requests.

ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
RyanL1997: I saw that in the OSD PR `src/plugins/data/public/antlr/opensearch_ppl/runtime_ppl_grammar/opensearch_ppl_autocomplete.ts`(https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11428/changes#diff-b4928cad849ab101a00e0ed40faadb5abe68a3b91f515c5c9323b7aa27b1aa82R406) we have actually corrected the name. And this will cause a mismatch. 
However, from the frontend perspective do we need to check if there is anything auto generated are still reference to the typo version?

ahkcs (approved):

PR #5189, opened 2026-02-26
Fix #5178: SUM of all-null column returns NULL per SQL standard

### LINKED ISSUE #5178: [BUG] SUM of all-null column returns 0.0 instead of NULL per SQL standard
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-types | stats sum(num_field) as total_sum
```
**Expected Result:**
When all values are null, `sum` should return `null`.

**Actual Result:**
Returns `0` with HTTP 200.

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "num_field": { "type": "integer" }
    }
  }
}
```
**Sample Data**
```json
[
  { "num_field": null },
  { "num_field": null },
  { "num_field": null }
]
```

## Bug Description
**Issue Summary:**
`SUM` on all-null column returns `0` instead of `null`.

**Steps to Reproduce:**
1. Create index with nullable numeric field.
2. Insert only null values for that field.
3. Run the query above.
4. Observe `total_sum = 0`.

**Impact:**
Aggregation semantics are incorrect and may hide missing-data conditions.

opensearchpplteam:
### Description

**Root Cause**: OpenSearch's `sum` aggregation returns `0.0` when all input values are null, which violates the SQL standard that requires `SUM` of an all-null column to return `NULL`.

**Fix**: Replace the `sum` aggregation with `stats` aggregation for SUM pushdown. The `stats` aggregation provides both `sum` and `count` fields. When `count == 0`, all input values were null, so the parser returns `null` instead of `0.0`. This is implemented in a new `SumStatsParser` class that follows the existing `MetricParser` pattern.

Both the Calcite code path (`AggregateAnalyzer`) and the legacy code path (`MetricAggregationBuilder`) are updated consistently.

### Related Issues

#5178

### Check List

- [x] Commits are signed per the DCO using --signoff or -s.


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
```diff
@@ -336,6 +338,23 @@ private String getAggregationPath(
...
+   * Resolve the order path for a bucket order. Stats aggregation is a multi-value metrics
+   * aggregation, so when ordering on it, the sort path must specify the sub-metric name (e.g.,
+   * "sum(balance).sum").
+   */
+  private String resolveOrderPath(String aggName, CompositeAggregationBuilder composite) {
+    AggregationBuilder sortMetric =
+        composite.getSubAggregations().stream()
+            .filter(sub -> sub.getName().equals(aggName))
+            .findFirst()
+            .orElse(null);
```
opensearchpplteam, line 351: The `resolveOrderPath` method uses `.orElse(null)` when looking up the sub-aggregation. If the aggregation name is not found (which should not happen in normal flow), the method silently returns `aggName` without the `.sum` suffix rather than failing fast. Consider using `.orElseThrow()` to surface unexpected state earlier, or add a brief comment explaining why the fallback is intentional.

penghuo:
This is opensearch issue. https://github.com/opensearch-project/OpenSearch/issues/17581.

opensearchpplteam:
Thanks for flagging this, @penghuo. You're right that the root cause is in OpenSearch's `sum` aggregation, which returns `0.0` instead of `null` when all input values are null. OpenSearch/OpenSearch#17581 tracks that core fix.

This PR provides a SQL-plugin-level workaround so that `SUM` returns the correct SQL-standard result (`NULL`) now, by switching to the `stats` aggregation and using `count == 0` to detect all-null input. Once the core issue is resolved, this can be simplified back to using the `sum` aggregation directly.

PR #5190, opened 2026-02-26
Fix MAP field overwrite in eval with dotted-path assignments

### LINKED ISSUE #5185: [BUG] `eval` overwrites MAP root field when assigning multiple dotted-path names
**What is the bug?**

When using `eval` to materialize multiple dotted paths from the same MAP column in a single command (e.g., `eval doc.user.name=doc.user.name, doc.user.age=doc.user.age`), the second assignment fails with `Field [doc.user.age] not found`. The first assignment succeeds, but it removes the MAP root column `doc` from the schema, making subsequent paths unresolvable.

```
  -d '{"doc": "{\"user\":{\"name\":\"John\",\"age\":30}}"}'
  -d '{"doc": "{\"user\":{\"name\":\"Alice\",\"age\":25}}"}'
```
2. Enable Calcite and run:
```
source=t | spath input=doc | eval doc.user.name=doc.user.name, doc.user.age=doc.user.age`
Error: Field [doc.user.age] not found
```

3. Note: splitting into two separate eval commands also fails for the same reason:

```
source=t | spath input=doc | eval doc.user.name=doc.user.name | eval doc.user.age=doc.user.age
```

Both dotted-path assignments should succeed. The MAP root column doc should not be removed when a dotted-path column like `doc.user.name` is added — the dotted name is a new flat column, not a nested sub-field of doc.

 - Version: PPL Calcite V3
 - Plugins: 3.6

N/A

- Single dotted-path eval works: `eval doc.user.name=doc.user.name succeeds`.
- The workaround is to use non-dotted aliases: `eval username=doc.user.name, age=doc.user.age`.
- The `shouldOverrideField` heuristic was designed for OpenSearch nested/struct fields where resource.attributes.key is a real sub-field of resource. For MAP columns produced by spath, the dotted name is just a flat column name that happens to contain dots — it's not a nested sub-field.
- The root cause is in `projectPlusOverriding` → `shouldOverrideField`, which checks `newName.startsWith(originalName + ".")`. When eval creates a column named `doc.user.name`, the check `"doc.user.name".startsWith("doc.")` returns true, so `projectPlusOverriding` removes the original `doc` MAP column. The second eval expression `doc.user.age=doc.user.age` then fails because `doc` no longer exists in the schema.

```java
// shouldOverrideField (CalciteRelNodeVisitor.java)
private boolean shouldOverrideField(String originalName, List<String> newNames) {
    return newNames.stream()
        .anyMatch(newName ->
            newName.equals(originalName)
            || newName.startsWith(originalName + "."));  // ← incorrectly matches MAP root
}
```

utafrali:
eval was overwriting MAP fields when multiple dotted-path names were assigned. The issue was assuming any dotted path starting with a field name meant it was a nested access, but MAP columns have flat names with dots in them. Now we check the field type before deciding to override.

Fixes #5185

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -981,10 +981,11 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
 
   private void projectPlusOverriding(
       List<RexNode> newFields, List<String> newNames, CalcitePlanContext context) {
-    List<String> originalFieldNames = context.relBuilder.peek().getRowType().getFieldNames();
+    RelDataType rowType = context.relBuilder.peek().getRowType();
+    List<String> originalFieldNames = rowType.getFieldNames();
     List<RexNode> toOverrideList =
         originalFieldNames.stream()
-            .filter(originalName -> shouldOverrideField(originalName, newNames))
+            .filter(originalName -> shouldOverrideField(originalName, newNames, rowType))
             .map(a -> (RexNode) context.relBuilder.field(a))
             .toList();
     // 1. add the new fields, For example "age0, country0"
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1007,15 +1005,19 @@ private void projectPlusOverriding(
...
+  private boolean shouldOverrideField(
+      String originalName, List<String> newNames, RelDataType rowType) {
+    RelDataTypeField field = rowType.getField(originalName, true, false);
+    boolean isMapType = field != null && field.getType().getSqlTypeName() == SqlTypeName.MAP;
     return newNames.stream()
         .anyMatch(
             newName ->
                 // Match exact field names (e.g., "age" == "age") for flat fields
                 newName.equals(originalName)
                     // OR match nested paths (e.g., "resource.attributes..." starts with
```
dai-chen, line 1017: @ahkcs Could you help check why previously we want to override root field (PR https://github.com/opensearch-project/sql/pull/4606)? Just want to confirm  if current MAP path issue is by design or we should fix it like this.

utafrali:
Thanks for the review! I'll address the DCO failure by amending the commits with `--signoff` and will add an integration test in CalcitePPLSpathCommandIT.

Regarding the MAP path override question: Looking at PR #4606, the root field override was used to handle schema derivation for the root context. However, for MAP types, we need to preserve the original schema structure since MAP inherently allows arbitrary nested keys. The fix ensures we don't incorrectly override the field descriptor for MAP types, which was causing the path extraction issues. If there's a specific reason the override should apply to MAP types as well, I'm happy to adjust the approach.

utafrali:
Closing — can't sign the DCO. Thanks\!

PR #5193, opened 2026-02-27
Add isnotempty PPL function

### LINKED ISSUE #5182: [FEATURE] `isnotempty` function
We have an [isempty](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/condition.md#isempty) PPL function. Arguably more useful is its counterpart, non-emptiness. This currently is done with `not isempty`.

An `isnotempty` function. This would follow the same pattern as e.g. [isnotnull](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/condition.md#isempty).

N/A

You could probably make the same case for ispresent/isnotpresent and isblank/isnotblank.

lucacavenaghi97:
## Description

Adds `isnotempty(field)` to PPL. Returns `true` when a field is not null and not empty string — logical negation of `isempty(field)`.

Resolves #5182

## Changes

- Added `ISNOTEMPTY` token and parser rules in ppl, language-grammar, and async-query-core grammars
- Added `IS_NOT_EMPTY` enum constant to `BuiltinFunctionName`
- Registered Calcite implementation as `NOT(IS_NULL(arg) OR IS_EMPTY(arg))` in `PPLFuncImpTable`
- Added `testIsNotEmpty` integration test
- Added docs

### DIFF
--- a/docs/user/ppl/functions/condition.md
+++ b/docs/user/ppl/functions/condition.md
@@ -642,8 +642,40 @@ fetched rows / total rows = 4/4
 | False         |         | True              | null     |
 +---------------+---------+-------------------+----------+
 ```
-  
-## EARLIEST  
+
+## ISNOTEMPTY
+
+### Description
+
+Usage: `isnotempty(field)` returns true if the field is not null and is not an empty string.
+
+**Argument type:** All supported data types.
+**Return type:** `BOOLEAN`
+
+### Example
+
+```ppl
+source=accounts
+| eval temp = ifnull(employer, '   ')
+| eval `isnotempty(employer)` = isnotempty(employer), `isnotempty(temp)` = isnotempty(temp)
+| fields `isnotempty(temp)`, temp, `isnotempty(employer)`, employer
+```
+
+Expected output:
+
+```text
+fetched rows / total rows = 4/4
++------------------+---------+----------------------+----------+
+| isnotempty(temp) | temp    | isnotempty(employer) | employer |
+|------------------+---------+----------------------+----------|
+| True             | Pyrami  | True                 | Pyrami   |
+| True             | Netagy  | True                 | Netagy   |
+| True             | Quility | True                 | Quility  |
+| True             |         | False                | null     |
++------------------+---------+----------------------+----------+
+```
+
+## EARLIEST
 
 ### Description  
 
--- a/async-query-core/src/main/antlr/OpenSearchPPLLexer.g4
+++ b/async-query-core/src/main/antlr/OpenSearchPPLLexer.g4
@@ -373,6 +373,7 @@ REPLACE:                            'REPLACE';
 REVERSE:                            'REVERSE';
 CAST:                               'CAST';
 ISEMPTY:                            'ISEMPTY';
+ISNOTEMPTY:                         'ISNOTEMPTY';
 ISBLANK:                            'ISBLANK';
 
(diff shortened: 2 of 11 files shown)


docs/user/ppl/functions/condition.md
```diff
@@ -642,8 +642,40 @@ fetched rows / total rows = 4/4
...
+Usage: `isnotempty(field)` returns true if the field is not null and is not an empty string.
+
+**Argument type:** All supported data types.
+**Return type:** `BOOLEAN`
+
+### Example
+
+```ppl
+source=accounts
+| eval temp = ifnull(employer, '   ')
```
Swiddis, line 659: note: I didn't realize that isempty counted blank strings, I thought it only matched empty (zero-length) strings.

Weird imo, not this PR's fault.

docs/user/ppl/functions/condition.md
```diff
@@ -642,8 +642,40 @@ fetched rows / total rows = 4/4
...
+```text
+fetched rows / total rows = 4/4
++------------------+---------+----------------------+----------+
+| isnotempty(temp) | temp    | isnotempty(employer) | employer |
+|------------------+---------+----------------------+----------|
+| True             | Pyrami  | True                 | Pyrami   |
+| True             | Netagy  | True                 | Netagy   |
+| True             | Quility | True                 | Quility  |
+| True             |         | False                | null     |
++------------------+---------+----------------------+----------|
```
Swiddis: chore: Doctests are failing because of bad table formatting at the corner.

Will have to fix that before this can be formally approved & accepted. (Can run doctest locally with `./gradlew doctest`).

image

Swiddis (changes requested):
Overall solid, nice work. Good call on just registering this as an inversion of the existing function. One chore to fix before this can be merged (in general we're gonna need green CI).

Ideally this could use a unit test, but doctest & existing isempty tests mostly covers everything we care about.

docs/user/ppl/functions/condition.md
lucacavenaghi97: Fixed the table corner formatting, doctest passes locally now.

docs/user/ppl/functions/condition.md
lucacavenaghi97: I actually thought isempty only matched null and zero-length strings, not blank/whitespace. Isn't that what isblank is for? Or am I misunderstanding something?

penghuo:
@lucacavenaghi97 
`NOT isempty(field)` already provides the expected "not empty" semantics. Based on this behavior, a separate `isnotempty` function is not required.

Sample dataset

```json
PUT notisempty_demo
{
  "mappings": {
    "properties": {
      "id": {"type": "integer"},
      "name": {"type": "keyword"}
    }
  }
}

POST notisempty_demo/_bulk?refresh=true
{"index":{}}
{"id":1,"name":"alice"}
{"index":{}}
{"id":2,"name":""}
{"index":{}}
{"id":3,"name":null}
{"index":{}}
{"id":4,"name":"bob"}
{"index":{}}
{"id":5}
```

Query

```ppl
source=notisempty_demo
| where NOT isempty(name)
| fields id, name
| sort id
```

Result:
- Returns only `id=1 (alice)` and `id=4 (bob)`
- Rows with `name = ''`, `name = null`, or missing `name` are excluded

Issue #5191, opened 2026-02-26
fetch_size: head command does not correctly reflect total hits or override fetch_size


ahkcs:
## Description

When using the `fetch_size` API (default `fetch_size = 500`), the result status bar does not correctly display the total document count in certain `head` command scenarios. Both cases below are bugs — the status should always reflect the total number of hits, not just the number of documents retrieved by the query.

Related PR: https://github.com/opensearch-project/sql/pull/5109

## Bug 1: `head N` where N < fetch_size

**Example:** `head 200` with default `fetch_size = 500`

- **Current behavior:** Shows `200 of 200 results`
- **Expected behavior:** Should show `200 results` 

The status should reflect how many total documents match the query, not just how many were returned.

## Bug 2: `head N` where N > fetch_size

**Example:** `head 600` with default `fetch_size = 500`

- **Current behavior:** Shows `500 of 600 results`
- **Expected behavior:** Should show `600 of {total_hits} results`

When the user-specified `head` value exceeds the default `fetch_size`, the `head` value should override `fetch_size` so that all requested rows are returned.

## Summary of Expected Behavior

- The result status should always show the **total document count** (total hits), not just the number of documents retrieved from the query.
- If `head` value > `fetch_size`, the `head` value should **override** `fetch_size`.
- If `head` value < `fetch_size`, the status should show `head_value of total_hits`.

anasalkouz:
For Bug 2: head N where N > fetch_size

Expected behavior: Should show 600 of {total_hits} results

ahkcs:
> For Bug 2: head N where N > fetch_size
> 
> Expected behavior: Should show 600 of {total_hits} results

Updated issue

PR #5194, opened 2026-02-27
Explicit head command should override fetch_size in PPL

ahkcs:
## Summary

Fixes bug 2 in #5191

When a user's PPL query contains an explicit `head` command, the `fetch_size` API parameter should not inject an additional `Head` node on top. Previously, `AstStatementBuilder.visitPplStatement()` always wrapped the plan with `Head(fetchSize)`, creating a double-Head that capped results unexpectedly.

**Example:** `source=index | head 100` with `fetch_size=5` would only return **5 rows** instead of the expected **100 rows**, because the outer `Head(5)` silently overrode the user's explicit `head 100`.

## Test plan
- [x] Manual verification via curl on local cluster:
  - `head 5` + `fetch_size=10` → 5 rows ✓
  - `head 8` + `fetch_size=3` → 8 rows ✓ (head overrides fetch_size)
  - `fetch_size=3` (no head) → 3 rows ✓ (fetch_size still works when no head)

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLJoinIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLJoinIT.java
@@ -5,9 +5,11 @@
 
 package org.opensearch.sql.calcite.remote;
 
+import static org.opensearch.sql.legacy.TestUtils.getResponseBody;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_HOBBIES;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_OCCUPATION;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_STATE_COUNTRY;
+import static org.opensearch.sql.plugin.rest.RestPPLQueryAction.QUERY_API_ENDPOINT;
 import static org.opensearch.sql.util.MatcherUtils.assertJsonEquals;
 import static org.opensearch.sql.util.MatcherUtils.rows;
 import static org.opensearch.sql.util.MatcherUtils.schema;
@@ -17,9 +19,13 @@ import static org.opensearch.sql.util.MatcherUtils.verifyNumOfRows;
 import static org.opensearch.sql.util.MatcherUtils.verifySchema;
 
 import java.io.IOException;
+import java.util.Locale;
 import org.json.JSONObject;
+import org.junit.Assert;
 import org.junit.Test;
 import org.opensearch.client.Request;
+import org.opensearch.client.RequestOptions;
+import org.opensearch.client.Response;
 import org.opensearch.sql.common.setting.Settings;
 import org.opensearch.sql.legacy.TestsConstants;
 import org.opensearch.sql.ppl.PPLIntegTestCase;
@@ -1074,5 +1080,33 @@ public class CalcitePPLJoinIT extends PPLIntegTestCase {
         rows("David", "USA", "Washington", 4, 2023, 40, "a"),
         rows("Jake", "USA", "California", 4, 2023, 70, "a"),
         rows("Hello", "USA", "New York", 4, 2023, 30, "e"));
+  }
+
+  @Test
+  public void testFetchSizeAppliesWhenHeadOnlyInJoinSubquery() throws IOException {
+    // head inside a join subquery should NOT suppress fetch_size on the outer query.
+    // The subquery's head operates at a different scope.
+    // Self-join STATE_COUNTRY on name, right side limited to head 3.
+    // fetch_size=2 should still cap the outer result to 2 rows.
+    String query =
+        String.format(
+            Locale.ROOT,
+            "source=%s | inner join left=a, right=b ON a.name = b.name"
+                + " [source=%s | sort name | head 3] | sort a.name | fields a.name, a.age",
+            TEST_INDEX_STATE_COUNTRY,
+            TEST_INDEX_STATE_COUNTRY);
+    Request request = new Request("POST", QUERY_API_ENDPOINT);
+    request.setJsonEntity(
+        String.format(Locale.ROOT, "{\"query\": \"%s\", \"fetch_size\": 2}", query));
+    RequestOptions.Builder restOptionsBuilder = RequestOptions.DEFAULT.toBuilder();
+    restOptionsBuilder.addHeader("Content-Type", "application/json");
+    request.setOptions(restOptionsBuilder);
+    Response response = client().performRequest(request);
+    Assert.assertEquals(200, response.getStatusLine().getStatusCode());
+    JSONObject result = new JSONObject(getResponseBody(response, true));
+    verifySchema(result, schema("name", "string"), schema("age", "int"));
+    // Right side after sort+head 3: David, Hello, Jake. Self-join matches all 3.
+    // Outer sort by name + fetch_size=2 caps to first 2: David, Hello
+    verifyDataRows(result, rows("David", 40), rows("Hello", 30));
   }
 }


ppl/src/main/java/org/opensearch/sql/ppl/parser/AstStatementBuilder.java
```diff
@@ -69,6 +69,23 @@ public static class StatementBuilderContext {
     private final String explainMode;
   }
 
+  /**
+   * Recursively checks if the AST contains a {@link Head} node. When the user's query already
+   * includes an explicit {@code head} command, we should not inject an additional Head for
+   * fetch_size so that the user's explicit limit takes precedence.
```
dai-chen: So if a query ends with `... | head 3 | head 500`, the limit in final physical plan is Limit 500?

ahkcs: It should be 3, added a corresponding test case

ahkcs: > So if a query ends with `... | head 3 | head 500`, the limit in final physical plan is Limit 500?

The new logic only decides if fetch_size will inject a new head node or not, so it won't affect the original head command behavior(which is that the smaller head command will take effect)

dai-chen: Sorry I meant without current change, it will be LIMIT 500? If so, is there bug elsewhere or this is Calcite planner expected behavior?

ahkcs: Here's an example physical plan:
https://github.com/opensearch-project/sql/pull/5194/changes#diff-0c3c28474cebca090844af02cd01618a9ec578f69a272774934c7b3fb1c020aeR8

If without the current change, it will be LIMIT 3 in the final `OpenSearchRequestBuilder(sourceBuilder`
I think it's the expected behavior

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstStatementBuilder.java
```diff
@@ -69,6 +69,23 @@ public static class StatementBuilderContext {
...
+   */
+  private boolean containsHead(UnresolvedPlan plan) {
+    if (plan instanceof Head) {
+      return true;
+    }
+    for (var child : plan.getChild()) {
+      if (child instanceof UnresolvedPlan && containsHead((UnresolvedPlan) child)) {
```
dai-chen: No matter where is the head command in the query, fetch_size will be ignored? Why not only check if the last command is head?

ahkcs: Yes.
For example this query:
```
source=t | head 100 | fields age with fetch_size=5
```
In the AST, fields (Project) is the outermost node, wrapping Head:
```
  Project([age], Head(100, Relation("t")))                                 
```                                                                                                     
   
If we only checked whether the last/outermost command is Head, this query would not match — rawPlan instanceof Head would be false because rawPlan is Project. So fetch_size=5 would still be injected, capping the user's head 100 to 5 rows.

dai-chen: I feel this makes the meaning of `fetch_size` parameter more confusing. With pagination, it's aligned with V2 and represent page size. Without pagination, it represents response size now. Either way it controls how many returned. But with this logic, it is impacted by any command in the query, like join, subquery or stats if they have head in/before it?

ahkcs: Good point. I've updated containsHead to only walk the main pipeline chain (first child) instead of recursively traversing all children. This means a head inside a join subquery or nested structure won't affect fetch_size injection for the outer query — only head commands in the main pipeline are considered.

ahkcs: Also updated documentation to explicitly explain the interaction between `head` command and `fetch_size` API to address the confusion.
`fetch_size` API should be a default limit instead of a fixed response size cap and it can be overridden by `head` command

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstStatementBuilder.java
```diff
@@ -32,7 +34,7 @@ public class AstStatementBuilder extends OpenSearchPPLParserBaseVisitor<Statemen
   @Override
   public Statement visitPplStatement(OpenSearchPPLParser.PplStatementContext ctx) {
     UnresolvedPlan rawPlan = astBuilder.visit(ctx);
-    if (context.getFetchSize() > 0) {
+    if (context.getFetchSize() > 0 && !containsHead(rawPlan)) {
```
dai-chen, line 37: In raw plan, isn't head command the first? Not sure why the containsHead still needs loop and check per child.

ahkcs: Added a test for `source=t | head 100 | fields age` with fetch_size=5. The fields command creates a Project that wraps Head in the raw plan, 
rawPlan = `Project([age], Head(100, Relation))` so rawPlan instanceof Head is false. 
Without the walk, fetch_size=5 gets injected and incorrectly caps the user's head 100 to 5 rows.

ahkcs: @dai-chen
Regarding subqueries — I checked and the walk should be safe. In the codebase, `getChild()` consistently returns only the main pipeline input at index 0. Subqueries are stored in separate fields (e.g., Join stores its right side in a `right` field, not in `getChild()`; same for Lookup's `lookupRelation`, Append's `subSearch`, etc.). So `children.get(0)` always follows the main pipeline and won't enter subqueries. But please let me know if I'm missing something\!

docs/user/interfaces/endpoint.rst
```diff
@@ -294,6 +294,37 @@ Unlike SQL's ``fetch_size`` which enables cursor-based pagination, PPL's ``fetch
...
+
+When a PPL query contains an explicit ``head`` command in the main pipeline, the ``head`` command takes precedence over ``fetch_size``. Because PPL's ``fetch_size`` does not support pagination, capping the result below the user's explicit ``head`` limit would silently discard rows with no way to retrieve them. To avoid this, ``fetch_size`` is ignored when the main pipeline contains a ``head`` command, and the query returns the number of rows specified by ``head``.
+
+If the query does **not** contain a ``head`` command in the main pipeline, ``fetch_size`` limits the result as usual. Note that ``head`` commands inside subqueries (e.g., within a ``join``) do not affect ``fetch_size`` behavior for the outer query, since they operate at a different scope.
+
++-----------------------------------------------+---------------------------------------------------+
+| Query                                         | Behavior with ``fetch_size=5``                    |
++===============================================+===================================================+
+| ``source=t \| fields age``                    | ``fetch_size`` applies — returns 5 rows           |
```
dai-chen: np: is the \ required?

dai-chen (approved):

docs/user/interfaces/endpoint.rst
ahkcs: Nice catch! Removed

docs/user/interfaces/endpoint.rst
```diff
@@ -294,6 +294,37 @@ Unlike SQL's ``fetch_size`` which enables cursor-based pagination, PPL's ``fetch
...
+
+If the query does **not** contain a ``head`` command in the main pipeline, ``fetch_size`` limits the result as usual. Note that ``head`` commands inside subqueries (e.g., within a ``join``) do not affect ``fetch_size`` behavior for the outer query, since they operate at a different scope.
+
++-----------------------------------------------+---------------------------------------------------+
+| Query                                         | Behavior with ``fetch_size=5``                    |
++===============================================+===================================================+
+| ``source=t | fields age``                     | ``fetch_size`` applies — returns 5 rows           |
++-----------------------------------------------+---------------------------------------------------+
+| ``source=t | head 100 | fields age``          | ``head`` takes precedence — returns 100 rows      |
```
penghuo, line 309: fetch_size is optional — if you don't want it, don't send it. The server should honor what the client explicitly requests. The fix belongs in the client that's injecting fetch_size by default.

From a REST API perspective, fetch_size is an explicit, optional parameter the caller chooses to send. If the caller sends fetch_size=5, they're requesting at most 5 rows — that's the API contract. Adding server-side logic to silently ignore that parameter based on query content breaks the principle of least surprise at the API layer.

ahkcs: I think this is a very valid point. If we fix it server-side, we're working around a client behavior by making the API semantics inconsistent. If we fix it client-side, the API contract stays clean.

ahkcs: Opened PR on OSD side: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11430

ahkcs:
closing as PR merged on OSD

Issue #5152, opened 2026-02-17
[BUG] Inconsistent MAP path resolution across PPL commands


dai-chen:
**What is the bug?**

When a PPL command produces a MAP-typed column (e.g., spath auto-extract), downstream commands that reference nested values via dotted paths like `doc.user.name` fail in many commands. The Calcite engine resolves `doc.user.name` as ITEM(doc, 'user.name') through `QualifiedNameResolver`, and this works in commands that use `rexVisitor.analyze()` (eval, where, stats, sort, dedup, eventstats, streamstats). However, commands that match field names by iterating `getRowType().getFieldNames()` fail because the row type only contains `doc` (the MAP column), not `doc.user.name`.

**How can one reproduce the bug?**

1. Create an index with a JSON string field and enable Calcite:

```
PUT /test/_doc/1 {"doc": "{\"user\":{\"name\":\"John\",\"age\":30,\"city\":\"NYC\"}}"}
PUT /test/_doc/2 {"doc": "{\"user\":{\"name\":\"Alice\",\"age\":25,\"city\":\"LA\"}}"}
```

2. Run any of these queries (each demonstrates a different failure):

```
# Field name matching failures (Category A):
source=test | spath input=doc | rename doc.user.name as username | fields username
source=test | spath input=doc | fillnull using doc.user.name = 'N/A'
source=test | spath input=doc | replace 'John' WITH 'Jonathan' IN doc.user.name
source=test | spath input=doc | top 1 doc.user.name by doc.user.city
source=test | spath input=doc | rare doc.user.name by doc.user.city
source=test | spath input=doc | fields - doc.user.name
source=test | spath input=doc | rename doc.user.* as extracted.*

# Projection naming failures (Category B):
source=test | spath input=doc | top 2 doc.user.name
source=test | spath input=doc | rare doc.user.name
source=test | spath input=doc
  | inner join left=a, right=b ON a.doc.user.name = b.name other_index
```

**What is the expected behavior?**

Full PartiQL-like path navigation support: doc.user.name should work uniformly across all PPL commands when doc is a MAP/STRUCT column.

**What is your host/environment?**
 - PPL Calcite V3 engine

**Do you have any screenshots?**

N/A

**Do you have any additional context?**

- **User-side workarounds**
  - **Generic workaround:** materialize nested MAP values via `eval`, then reference the materialized columns downstream.  
    Example: `eval username = doc.user.name, age = doc.user.age`
[...]

PR #5198, opened 2026-03-03
Fix MAP path resolution for symbol-based PPL commands

dai-chen:
### Description

This PR resolves the failure Category A in [Impact Analysis](https://github.com/opensearch-project/sql/issues/5152#issuecomment-3917179411) by pre-materializing referenced MAP dotted paths for symbol-based PPL commands. Referenced paths such as `doc.user.name` are resolved and projected as flat named columns before command-specific planning so existing schema-based matching logic can work uniformly on those references.

#### Implementation notes

The current implementation makes the following interim trade-offs:

1. **Operand extraction is implemented as a per-command switch** rather than a generic `getOperands()` contract on AST nodes, to avoid a broader AST interface change in this PR. This is similar in spirit to Calcite’s `SqlCall.getOperandList()`, but PPL AST nodes do not yet expose a uniform operand API.
2. **Name resolution is triggered from visitChildren as a post-processing hook**, because the current PPL engine lacks a dedicated analysis phase where this would more naturally live. Longer term, this is closer in spirit to Calcite’s `CustomColumnResolvingTable`, but that broader resolver abstraction is out of scope for this PR.

#### Examples

<pre>
# Before changes
opensearchsql> source=t 
.............. | spath input=doc
.............. | replace 'John' WITH 'Jonathan' <b>IN doc.user.name</b>
.............. | fields doc.user.name;

{'reason': 'Invalid Query', 'details': '<b>field [doc.user.name] not found</b>;
  input fields are: [_id, _index, _score, _maxscore, _sort, _routing, doc]', 'type': 'IllegalArgumentException'}

# After changes
opensearchsql> source=t 
.............. | spath input=doc
.............. | replace 'John' WITH 'Jonathan' <b>IN doc.user.name</b>
.............. | fields doc.user.name;
fetched rows / total rows = 5/5
+-------------------------------------------------------------+-----------------+
| doc                                                         | doc.user.name   |
|-------------------------------------------------------------+-----------------|
| {'user.age': '30', 'user.name': 'John', 'user.city': 'NYC'} | Jonathan        |
| {'user.age': '25', 'user.name': 'Alice', 'user.city': 'LA'} | Alice           |
| {'user.age': '35', 'user.name': 'John', 'user.city': 'SF'}  | Jonathan        |
| {'user.age': '40', 'user.name': 'Bob', 'user.city': 'NYC'}  | Bob             |
| {'user.age': '40', 'user.city': 'NYC'}                      | null            |
+-------------------------------------------------------------+-----------------+

opensearchsql> explain source=t 
.............. | spath input=doc
.............. | replace 'John' WITH 'Jonathan' IN doc.user.name
.............. | fields doc.user.name;
OpenSearchException({'calcite': {
'logical': "
  LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])\n
    LogicalProject(doc=[JSON_EXTRACT_ALL($0)],
      <b>doc.user.name=[REPLACE(ITEM(JSON_EXTRACT_ALL($0), 'user.name'), 'John':VARCHAR, 'Jonathan':VARCHAR)])</b>\n
      CalciteLogicalIndexScan(table=[[OpenSearch, t]])\n", ...
</pre>

### Related Issues

Resolves partially https://github.com/opensearch-project/sql/issues/5152

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/MapPathPreMaterializer.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/MapPathPreMaterializer.java
@@ -73,15 +73,17 @@ public class MapPathPreMaterializer {
     Set<String> existingFields =
         new HashSet<>(context.relBuilder.peek().getRowType().getFieldNames());
     List<RexNode> newColumns = new ArrayList<>();
-    for (Field f : fields) {
+    for (Field field : fields) {
       try {
-        RexNode resolved = rexVisitor.analyze(f, context);
-        String name = f.getField().toString();
+        RexNode resolved = rexVisitor.analyze(field, context);
+        String name = field.getField().toString();
         if (resolved.getKind() == SqlKind.ITEM && !existingFields.contains(name)) {
           newColumns.add(context.relBuilder.alias(resolved, name));
+          existingFields.add(name);
         }
       } catch (RuntimeException e) {
-        log.debug("Skipping field resolution for '{}': {}", f.getField(), e.getMessage(), e);
+        // Skip unresolvable fields (e.g. wildcards); let the command itself handle them
+        log.debug("Skipping field resolution for '{}': {}", field.getField(), e.getMessage(), e);
       }
     }
 
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -193,15 +193,12 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     return unresolved.accept(this, context);
   }
 
-  /**
-   * Override visitChildren to auto-materialize MAP dotted paths after children are visited. When a
-   * command references dotted paths into MAP columns, they are materialized as flat columns before
-   * the command's own logic runs.
-   */
   @Override
   public RelNode visitChildren(Node node, CalcitePlanContext context) {
     RelNode result = super.visitChildren(node, context);
     if (node instanceof UnresolvedPlan plan) {
+      // Materialize MAP dotted paths as flat columns after children are analyzed
+      // (so MAP/struct types are known) but before the command's own visit logic runs.
       mapPathMaterializer.materializePaths(plan, context);
     }
     return result;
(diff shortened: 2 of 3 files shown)


penghuo (approved):

core/src/main/java/org/opensearch/sql/calcite/MapPathPreMaterializer.java
```diff
@@ -0,0 +1,106 @@
...
+      case MvCombine mvCombine -> List.of(mvCombine.getField());
+      default -> List.of();
+    };
+  }
+
+  private void doMaterializeMapPaths(List<Field> fields, CalcitePlanContext context) {
+    Set<String> existingFields =
+        new HashSet<>(context.relBuilder.peek().getRowType().getFieldNames());
+    List<RexNode> newColumns = new ArrayList<>();
+    for (Field field : fields) {
```
Swiddis, line 76: note: This is going to be expensive for indices with lots of fields, since we're calling the materialization on every node (except the leaf) during planning.

core/src/main/java/org/opensearch/sql/calcite/MapPathPreMaterializer.java
```diff
@@ -0,0 +1,106 @@
...
+  private List<Field> extractFieldOperands(UnresolvedPlan node) {
+    return switch (node) {
+      case Rename rename -> toFields(rename.getRenameList(), m -> m.getOrigin());
+      case FillNull fillNull -> toFields(fillNull.getReplacementPairs(), Pair::getLeft);
+      case Replace replace -> toFields(replace.getFieldList());
+      // Only fields-exclusion needs pre-materialization; inclusion resolves paths via visitProject
+      case Project project -> project.isExcluded() ? toFields(project.getProjectList()) : List.of();
+      case AddTotals addTotals -> toFields(addTotals.getFieldList());
+      case MvCombine mvCombine -> List.of(mvCombine.getField());
+      default -> List.of();
```
Swiddis, line 68: thought: I'm a bit concerned that people implementing commands will need to implicitly know if adding to this case statement is necessary.

Is there a way to define a contract so this is less likely to drift over time? This specific method could probably also use related documentation.

dai-chen: Yes, as I mentioned in Implementation Notes, we can add contract on AST node to enforce this.

core/src/main/java/org/opensearch/sql/calcite/MapPathPreMaterializer.java
dai-chen: Yeah, I'm trying to minimize the impact. Currently this materialize logic is only activated on the 6 commands which gives input operands above.

PR #5199, opened 2026-03-04
[BugFix] Fix PushDownContext shallow copy bug

### LINKED ISSUE #5125: Reverse Optimization fails when creating consecutive sorts due to Calcite physical optimizer merging

songkant-aws:
### Description
PushdownContext.clone() / cloneWithoutSort() / cloneForAggregate() replays add(operation) to construct new context. Aggregation operations reuses the same `AggPushDownAction` instance in both old and new context due to shallow copy side effect happened during cloning PushdownContext. The following pushdown actions from other equivalent RelSubsets will modify this shared mutable action object, which could contaminate previous context's agg states.

### Related Issues
Resolves #5125 

### Check List


LantaoJin:
cc @qianheng-aws as proper reviewer

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
```diff
@@ -64,6 +66,156 @@ public AggPushDownAction(
...
+    if (builder instanceof HistogramAggregationBuilder histogram) {
+      return copyHistogramAggregationBuilder(histogram);
+    }
+    if (builder instanceof TopHitsAggregationBuilder topHits) {
+      return copyTopHitsAggregationBuilder(topHits);
+    }
+    if (builder instanceof NestedAggregationBuilder nested) {
+      return copyNestedAggregationBuilder(nested);
+    }
+    return builder;
```
qianheng-aws: What if there is other missed or upcoming agg builder?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggPushDownAction.java
```diff
@@ -64,6 +66,156 @@ public AggPushDownAction(
...
+    return metadata == null || metadata.isEmpty() ? null : metadata;
+  }
+
+  private static final class CompositeAggregationBuilderCopy extends CompositeAggregationBuilder {
+    private CompositeAggregationBuilderCopy(CompositeAggregationBuilder source) {
+      super(source, copySubAggregations(source), copyMetadataOrNull(source));
+    }
+  }
+
+  private static final class TermsAggregationBuilderCopy extends TermsAggregationBuilder {
```
qianheng-aws: Can this be replaced with 
```
source.shallowCopy(source, copySubAggregations(source), copyMetadataOrNull(source))
```
instead of defining these extended class?

qianheng-aws (commented):
Have you ever tried lazy push-down for `AggregationBuilderAction`? If that approach can work, we don't need to create AggregationBuilder in the planning phase, neither need to do deep copy for AggregationBuilder then.

songkant-aws:
@qianheng-aws Now the fix is changed to lazily push down agg request at the createRequest phase. But it introduces some intermediate immutable spec object to record any logical pushdown operations. It touches more logic now but I think it could mitigate the current problem of AggPushDownAction's mutation during planning.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
```diff
@@ -0,0 +1,520 @@
...
+  private static LimitPushdownMode inferLimitPushdownMode(AggKind kind) {
+    return switch (kind) {
+      case COMPOSITE, TERMS, MULTI_TERMS, TOP_HITS, RARE_TOP -> LimitPushdownMode.BUCKET_SIZE;
+      case OTHER, DATE_HISTOGRAM, HISTOGRAM -> LimitPushdownMode.UNSUPPORTED;
+    };
+  }
+
+  private static LimitPushdownMode inferLimitPushdownMode(
+      List<AggregationBuilder> builders, @Nullable AggregationBuilder rootBuilder) {
+    if (builders.isEmpty()) {
```
qianheng-aws: Nit: seems no need to pass in param `builders` here. This can be replaced with rootBuilder==null.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
```diff
@@ -0,0 +1,520 @@
...
+      // count() optimization uses hits.total and leaves the builder list empty. Keeps
+      // LIMIT in PushDownContext for these cases even though no request-side limit is applied.
+      return LimitPushdownMode.ESTIMATE_ONLY;
+    }
+    AggregationBuilder builder = unwrapNestedBuilder(rootBuilder);
+    if (builder instanceof ValuesSourceAggregationBuilder.LeafOnly<?, ?>) {
+      // Treats leaf metric aggregations as limit-pushable because they produce a single row.
+      return LimitPushdownMode.LEAF_METRIC;
+    }
+    return inferLimitPushdownMode(inferKind(rootBuilder));
```
qianheng-aws: Nit: pass in `kind` to avoid calling inferKind method here, it's already evaluated before calling `inferLimitPushdownMode`.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
```diff
@@ -0,0 +1,520 @@
...
+        outputFields,
+        rowType,
+        fieldTypes,
+        cluster,
+        bucketNullable,
+        queryBucketSize,
+        extendedTypeMapping,
+        bucketNames,
+        bucketNames,
+        new AggPushDownAction(builderAndParser, extendedTypeMapping, bucketNames).getScriptCount(),
```
qianheng-aws: It's weird to new a AggPushDownAction here only to get the script count. Replace it with builderAndParser.getLeft().stream().mapToInt(AggPushDownAction::getScriptCount).sum()

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
```diff
@@ -0,0 +1,520 @@
...
+        fieldTypes,
+        cluster,
+        bucketNullable,
+        queryBucketSize,
+        extendedTypeMapping,
+        initialBucketNames,
+        bucketNames,
+        scriptCount,
+        measureSortTarget,
+        inferLimitPushdownMode(measureSortTarget),
```
qianheng-aws: Why invoking infer again while measureSortTarget is unchanged here?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
```diff
@@ -0,0 +1,520 @@
...
+        rareTopDigest,
+        size);
+  }
+
+  public Pair<List<AggregationBuilder>, OpenSearchAggregationResponseParser> build() {
+    try {
+      AggregateAnalyzer.AggregateBuilderHelper helper =
+          new AggregateAnalyzer.AggregateBuilderHelper(
+              rowType, fieldTypes, cluster, bucketNullable, queryBucketSize);
+      Pair<List<AggregationBuilder>, OpenSearchAggregationResponseParser> builderAndParser =
```
qianheng-aws: Why should we analyze the `aggregate` again? Can we reuse the `builderAndParser` generated when pushing down the `aggregate`? Even more, we may be able to reuse the `AggPushDownAction`.

Since we pushdown everything lazily until this build() method, which will only be invoked once. It should be safe to reuse.

qianheng-aws (commented):
As commented here:https://github.com/opensearch-project/sql/pull/5199/changes#r2980342004, if the `AggPushDownAction` can be reusable, plenty of params can be removed, which are maintained all the way but only for creating `AggPushDownAction` in the end in `build()` method.

qianheng-aws:
> As commented here:https://github.com/opensearch-project/sql/pull/5199/changes#r2980342004, if the `AggPushDownAction` can be reusable, plenty of params can be removed, which are maintained all the way but only for creating `AggPushDownAction` in the end in `build()` method.

As far as I can see, only these params are getting changed in the push down process for agg.
```
bucketNames, kind, limitPushdownMode, bucketSortCollations, bucketSortFieldNames, digest, bucketSize
```

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
songkant-aws: Yes, replace it with `rootBuilder == null` check now.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
songkant-aws: Done

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
songkant-aws: Done. It's now calling getScriptCount instead

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
songkant-aws: This was incorrect. Now it will infer rewriteTarget in local method and based on rewritten AggKind to inferLimitPushdownMode again.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/context/AggSpec.java
songkant-aws: As discussed offline. Ideally if we don't care about afterKey rewrite in agg pagination, then the `builderAndParser` can be reusable. I change the behavior to reuse passed builderAndParser to AggSpec.

However, I reviewed the whole process. I still think we still need AggSpec like object to hold planning stage agg state immutable copies(can't fully AggPushDownAction). AggPushDownAction is used at the final materialization of replayed PushDownOperations for agg. It will be better to separate them into two different objects. During optimizer planning stage, we still need to record some agg state changes in different RelSubsets. One for planning side agg state immutable objects, the other one is for replaying final operations to materialize final AggregationBuilder(lazily generate request builder).

songkant-aws:
@qianheng-aws Now I have streamlined the AggSpec. And choose not to reuse AggPushDownAction because we need several immutable objects copies in different RelSubsets to represent different logical agg states per RelSubset. However, replaying pushdown operations to change internal agg parser is a one-time mutable state change, which is better lazily built at last.

qianheng-aws (approved):

LantaoJin (approved):

Issue #5031, opened 2026-01-08
[FEATURE] Implement convert command in ppl


asifabashar:
The convert command transforms field values—typically into numeric or time formats.
Without AS, the original field is replaced.
Some of the functionality can be achieved by functions like strftime(),  or tonumber() for conversions.
convert [timeformat=<string>] (<convert-function> [AS <field>])...
Optional

timeformat=time format used in the field.
Default: %m/%d/%Y %H:%M:%S.

AS <field>
Writes results to a new field instead of overwriting the original.

convert functions:
Convert function parameter can be field names space separated or to specify all fields  use wildcard ( * ) character.
Function | Description
-- | --
auto() | Auto-convert fields to numbers; fails if any value cannot convert.
ctime() | UNIX time → human-readable time.
dur2sec() | [D+]HH:MM:SS → seconds.
memk() | Converts values with units (k/m/g) → kilobytes.
mktime() | Human-readable → epoch time.
mstime() | [MM:]SS.SSS → seconds.
none() | Exclude fields from wildcard conversion.
num() | Like auto(), but drops non-convertible values.
rmcomma() | Removes commas (e.g., 1,000 → 1000).
rmunit() | Strips units from numeric values ("212 sec" → 212).

Example:
Following example converts all fields to numeric, keeps values unchanged that cannot be converted.

... | convert auto(*)

Following example  converts all values except for values in specified fields in none function

... | convert auto(*) none(hostname)

Following example changes the duration values in format  [D+]HH:MM:SS to seconds for the specified fields. If field name is duration in following format  [D+]HH:MM:SS. For instance, duration="00:05:08" will be converted to duration="308" seconds
... | convert dur2sec(duration) dur2sec(duration)

This example uses the rmunit convert function to remove unit from the value. For example, if duration="101 sec", the resulting value is duration="101". 
... | convert rmunit(duration)

**Is your feature request related to a problem?**
To add feature part of roadmap #4287 

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

asifabashar:
@LantaoJin @penghuo  please review and  assign this to me if it makes sense and no one else is working on it.

asifabashar:
Duplicate of [RFC] PPL convert Command](https://github.com/opensearch-project/sql/issues/5001)
[untriaged](https://github.com/opensearch-project/sql/issues?q=is%3Aissue%20state%3Aopen%20label%3Auntriaged)
Status: Open.

PR #5210, opened 2026-03-06
Add time conversion functions for convert command

ritvibhatt:
### Description

Implement ctime, mktime, mstime, and dur2sec conversion functions for the PPL convert command, along with timeformat parameter support.

### Changes

- ctime(field) — converts UNIX epoch timestamps to human-readable time strings
- mktime(field) — converts time strings to UNIX epoch timestamps
- mstime(field) — converts [MM:]SS.SSS time strings to total seconds
- dur2sec(field) — converts HH:MM:SS duration strings to total seconds
- timeformat parameter — allows ctime and mktime to use custom strftime format specifiers (default: %m/%d/%Y %H:%M:%S)
- Register all four functions in PPLFuncImpTable
- Preserve timeformat parameter in PPLQueryDataAnonymizer output
- Add ANY_OPTIONAL_STRING operand type for functions accepting 1 or 2 arguments
- Add toJavaPattern() utility to convert strftime specifiers to Java DateTimeFormatter patterns

### Testing

- Unit tests for all four conversion functions
- PPL parser tests for convert with timeformat
- Anonymizer tests for convert command variants
- Integration tests for all functions with default and custom timeformat

### Related Issues
Related to #5031 

### Check List


docs/user/ppl/cmd/convert.md
```diff
@@ -241,6 +246,108 @@ fetched rows / total rows = 3/3
...
+```
+
+The query returns the following results:
+
+```text
+fetched rows / total rows = 1/1
++----------+
+| duration |
+|----------|
+| 5025.0   |
```
dai-chen, line 347: The result is double?

docs/user/ppl/cmd/convert.md
```diff
@@ -241,6 +246,108 @@ fetched rows / total rows = 3/3
...
++--------------+
+| date_str     |
+|--------------|
+| 1.066507633E9|
++--------------+
+```
+
+## Example 11: Using timeformat with ctime() and mktime()
+
+The `timeformat` parameter specifies a strftime format string for `ctime()` and `mktime()`:
```
dai-chen, line 291: np: `mktime` is not demonstrated below?

docs/user/ppl/cmd/convert.md
```diff
@@ -241,6 +246,108 @@ fetched rows / total rows = 3/3
...
+```
+
+The query returns the following results:
+
+```text
+fetched rows / total rows = 1/1
++----------+
+| time_str |
+|----------|
+| 225.5    |
```
dai-chen, line 367: same question as above

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -995,6 +999,10 @@ void populate() {
       registerOperator(RMCOMMA, PPLBuiltinOperators.RMCOMMA);
       registerOperator(RMUNIT, PPLBuiltinOperators.RMUNIT);
       registerOperator(MEMK, PPLBuiltinOperators.MEMK);
+      registerOperator(CTIME, PPLBuiltinOperators.CTIME);
+      registerOperator(MKTIME, PPLBuiltinOperators.MKTIME);
+      registerOperator(MSTIME, PPLBuiltinOperators.MSTIME);
+      registerOperator(DUR2SEC, PPLBuiltinOperators.DUR2SEC);
```
dai-chen, line 1005: All these have to a UDF?

integ-test/src/test/resources/expectedOutput/calcite/explain_convert_ctime.yaml
```diff
@@ -0,0 +1,8 @@
+calcite:
+  logical: |
+    LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
+      LogicalProject(ts=[CTIME(1066507633)])
```
dai-chen, line 4: Please double check if it's necessary to add 2 yaml files for new PPL function. Also this ymal is almost the same as no pushdown version, are these new function not pushed down anyway?

ritvibhatt: The functions aren't pushed down, pushdown implementation will be in a follow up PR but I think the explain tests look for the files in both folders so both files are necessary.

docs/user/ppl/cmd/convert.md
ritvibhatt: Thanks for catching that! Updated to include an example with ```mktime```

docs/user/ppl/cmd/convert.md
ritvibhatt: Yes fractional seconds are allowed so should return a double

docs/user/ppl/cmd/convert.md
ritvibhatt: Yes, if there is a numeric value passed to the function like "5025.5", it should return 5025.5

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ritvibhatt: Yeah, I couldn't find any Calcite operators that handled parsing time the formats that was needed for these functions

dai-chen (approved):

ahkcs (approved):

PR #5219, opened 2026-03-09
added cloudwatch style contains operator

### LINKED ISSUE #5181: [FEATURE] Cloudwatch-style `like` operator
I quite like how Cloudwatch's log insights QL lets us easily search string fields with an inline `like`:
```
| filter @message like 'keyword'
```

I'd like if we supported cloudwatch-style inline `like` statements in `where` clauses.

PPL already has a `LIKE` function:
```
where LIKE(message, '%keyword%')
```
I just find the syntax more cumbersome than the alternative.

N/A

ThyTran1402:
### Description
- Adds a contains operator to PPL where clauses as a more readable compare to LIKE(field, '%keyword%') or field like '%keyword%'.
- | where message contains 'error is equivalent to | where message like '%error%' .

Behavior:
- Case-insensitive substring match (backed by ilike)
- RHS must be a string literal; a field reference on the RHS raises a SemanticCheckException                                                                          - % and _ in the keyword are treated as LIKE wildcards (consistent with the underlying operator)
- contains can still be used as a field/identifier name in other positions 

### Related Issues
Resolves #5181 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/operator/predicate/BinaryPredicateOperators.java
+++ b/core/src/main/java/org/opensearch/sql/expression/operator/predicate/BinaryPredicateOperators.java
@@ -53,6 +53,7 @@ public class BinaryPredicateOperators {
     repository.register(greater());
     repository.register(gte());
     repository.register(like());
+    repository.register(ilike());
     repository.register(notLike());
     repository.register(regexp());
   }
@@ -389,6 +390,12 @@ public class BinaryPredicateOperators {
         BuiltinFunctionName.LIKE.getName(),
         impl(nullMissingHandling(OperatorUtils::matches2), BOOLEAN, STRING, STRING),
         impl(nullMissingHandling(OperatorUtils::matches3), BOOLEAN, STRING, STRING, BOOLEAN));
+  }
+
+  private static DefaultFunctionResolver ilike() {
+    return define(
+        BuiltinFunctionName.ILIKE.getName(),
+        impl(nullMissingHandling(OperatorUtils::matches2), BOOLEAN, STRING, STRING));
   }
 
   private static DefaultFunctionResolver regexp() {
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/FilterQueryBuilder.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/FilterQueryBuilder.java
@@ -68,6 +68,7 @@ public class FilterQueryBuilder extends ExpressionNodeVisitor<QueryBuilder, Obje
           .put(BuiltinFunctionName.LTE.getName(), new RangeQuery(Comparison.LTE))
           .put(BuiltinFunctionName.GTE.getName(), new RangeQuery(Comparison.GTE))
           .put(BuiltinFunctionName.LIKE.getName(), new LikeQuery())
+          .put(BuiltinFunctionName.ILIKE.getName(), new LikeQuery())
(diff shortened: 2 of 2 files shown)


integ-test/src/test/java/org/opensearch/sql/ppl/WhereCommandIT.java
```diff
@@ -144,6 +144,34 @@ public void testLikeOperatorCaseInsensitive() throws IOException {
     verifyDataRows(result3, rows("Amber"));
   }
 
+  @Test
+  public void testContainsOperator() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | where firstname contains 'mbe' | fields firstname",
```
LantaoJin, line 152: is the `contains` case-sensitive or insensitive? Need more tests for this.
Can you link the cloudwatch doc link in description?
And this PR should include user document updates. maybe in `condition.md`

ThyTran1402: I think contains is case-insensitive since it's implemented with ILIKE with %value% wrapping.
If a user writes firstname contains '%', the wrapped pattern becomes %%% which would match everything, which is   likely unintended. I'll fix this and add more test cases.
Here is the doc provided: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_QuerySyntax-Filter.html

ThyTran1402:
Hi @LantaoJin 

I think the PR is ready for review.

Thank you!

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -213,6 +213,17 @@ public UnresolvedExpression visitCompareExpr(CompareExprContext ctx) {
     String operator = ctx.comparisonOperator().getText();
     if ("==".equals(operator)) {
       operator = EQUAL.getName().getFunctionName();
+    } else if ("contains".equalsIgnoreCase(operator)) {
```
Swiddis, line 216: nit: In principle this should be a const like `LIKE` but it's probably fine

We already broke the const in the previous if with `"=="`, if someone doesn't like it then they can review this file for consts in another PR.

Swiddis:
Failing unit test CI due to spotless violations, can fix by running `./gradlew :spotlessApply`.

Doctests (`./gradlew doctest`) and integ tests (`./gradlew integTest --tests WhereCommandIT`) are also failing with assertion errors.

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
ThyTran1402: That makes sense to me. Thank you for reviewing it!

ThyTran1402:
> Failing unit test CI due to spotless violations, can fix by running `./gradlew :spotlessApply`.
> 
> Doctests (`./gradlew doctest`) and integ tests (`./gradlew integTest --tests WhereCommandIT`) are also failing with assertion errors.

Hi @Swiddis 

I just clean it up. Can you please review it again?

Thank you!

ThyTran1402:
> upd: ^

Hi @Swiddis 

Are there any changes that you recommend? since I think the PR is ready for review.

Swiddis:
Doctests and integ tests are still failing.

ThyTran1402:
> Doctests and integ tests are still failing.

hmm that's right 😢 . I fixed doctests and inges tests already. Can you run the testcase again?

Thank you!

Swiddis (approved):

PR #5221, opened 2026-03-10
Fix PIT (Point in Time) resource leaks in v2 query engine

### LINKED ISSUE #5220: [BUG] PPL query failed with Trying to create too many Point In Time contexts. Must be less than or equal to: [300]
**What is the bug?**
Running PPL query like below, will get errors

**PPL**
`
source=cwl*  |where @timestamp >'2011-03-07 00:12:38.000' and @timestamp <'2026-03-07 00:12:36.000' | where match(@message, 'error') | head 20000
`
**Error**

`
"{\n  \"error\": {\n    \"reason\": \"Error occurred in OpenSearch engine: Partial shards failure\",\n    \"details\": \"Shard[1]: org.opensearch.core.concurrency.OpenSearchRejectedExecutionException: Trying to create too many Point In Time contexts. Must be less than or equal to: [300]. This limit can be set by changing the [search.max_open_pit_context] setting.\\nShard[1]
`

 - Plugins

Image

1. cwl* will match 55 indices
2. calcite is not enabled

LantaoJin:
### Description

This PR fixes multiple PIT (Point in Time) resource leak scenarios in the **v2** query engine. PITs are server-side resources created for paginated search; if not properly deleted, they accumulate and eventually hit the `search.max_open_pit_context limit`, causing all subsequent queries to fail.

Three distinct PIT leak paths were identified:

1. `explain()` never closes the physical plan — `OpenSearchExecutionEngine.explain()` builds a physical plan (which creates a PIT) to generate explain output, then discards it without calling `close()`. Every explain call leaks a PIT per index matched. For example, `source=test*` against 50 indices leaks 50 PITs per call, exhausting the default 300 limit in just 6 calls.

2. `OpenSearchIndexScan.close()` skips PIT deletion on non-final pages — `close()` delegates to `request.clean()` which only deletes the PIT when `needClean=true` (last page consumed). If the scan is closed early due to LIMIT, exception, or timeout, the PIT is orphaned.

3. Explicit cursor close doesn't delete PIT — `CursorCloseOperator` propagates `close()` to children, but `OpenSearchIndexScan.close()` → `cleanup()` → `clean()` skips PIT deletion because `needClean=false` mid-pagination. The PIT is never reclaimed.

### Related Issues
Resolves #5220

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/data/utils/OpenSearchJsonContent.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/data/utils/OpenSearchJsonContent.java
@@ -13,6 +13,8 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchParseException;
 import org.opensearch.common.Numbers;
 import org.opensearch.common.geo.GeoPoint;
@@ -25,7 +27,7 @@ import org.opensearch.core.xcontent.XContentParser;
 /** The Implementation of Content to represent {@link JsonNode}. */
 @RequiredArgsConstructor
 public class OpenSearchJsonContent implements Content {
-
+  private static final Logger LOG = LogManager.getLogger();
   private final JsonNode value;
 
   @Override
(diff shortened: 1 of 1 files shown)


qianheng-aws:
> OpenSearchExecutionEngine.explain() builds a physical plan (which creates a PIT) to generate explain output

PIT isn't really needed for plan explain, I think creating PIT here is a kind of resource wasting.

LantaoJin:
> > OpenSearchExecutionEngine.explain() builds a physical plan (which creates a PIT) to generate explain output
> 
> PIT isn't really needed for plan explain, I think creating PIT here is a kind of resource wasting.

~~Exactly, I just release the resource, will check how to avoid to create.~~

I checked the code and find it maybe not worth to fix it in `explain` due to change so many interfaces in v2.

PR #4024, opened 2025-08-12
Implement type checking for aggregation functions with Calcite

yuancu:
### Description
Previously, using aggregation functions on improper fields will throw errors at runtime (in both v2 and v3). This PR  implements type checking for aggregation functions in OpenSearch PPL with Calcite enabled.

Additionally, this PR refactored type checker to make its association with function implementation more straightforward.

### Related Issues
Resolves #4000

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
 - [x] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5222, opened 2026-03-10
Fix memory leak: ExecutionEngine recreated per query appending to global function registry

ahkcs:
## Summary
- **Root cause:** `OpenSearchPluginModule.executionEngine()` was missing `@Singleton`, so Guice created a new `OpenSearchExecutionEngine` on every SQL/PPL request. Each construction called `registerOpenSearchFunctions()`, which appended to `PPLFuncImpTable.INSTANCE.externalFunctionRegistry` via `compute()` — growing the list by 1 entry per query, forever.
- After ~730K queries, the GEOIP function list contained ~730K entries, each holding 14 objects (GeoIpFunction, SqlIdentifier, CalciteFuncSignature, PPLTypeChecker, etc.), retaining ~2.12GB of heap.
- The bug existed pre-3.3 but was invisible because each leaked entry was only ~3 lightweight objects. PR #4024 changed registration from lazy (lambda) to eager (full object tree), making each entry ~14 objects and pushing the leak into OOM territory.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
@@ -425,7 +425,8 @@ public class PPLFuncImpTable {
     CalciteFuncSignature signature = new CalciteFuncSignature(functionName.getName(), typeChecker);
     externalFunctionRegistry.put(
         functionName,
-        List.of(Pair.of(signature, (FunctionImp) (builder, args) -> builder.makeCall(operator, args))));
+        List.of(
+            Pair.of(signature, (FunctionImp) (builder, args) -> builder.makeCall(operator, args))));
   }
 
   /**


ahkcs:
> Can you confirm that `registerExternalOperator` is never called more than once for the same `BuiltinFunctionName`? The old `compute()` logic accumulated overloads, while the new `put()` replaces them. If single-registration is guaranteed, this simplification LGTM. Also, please verify that no downstream code mutates the list retrieved from `externalFunctionRegistry`, since `List.of()` is immutable.

`registerExternalOperator` is only called once per function name (GEOIP and DISTINCT_COUNT_APPROX), and the new `@Singleton` on the engine ensures it only runs once anyway. So put() replacing compute() is safe — there's nothing to accumulate.                                
                                                                                                                                                                                                                                                                                                                   Verified that there's no downstream mutation. All code that reads from the registry just iterates the list; nothing tries to .add() or modify it. List.of() being immutable is fine.

PR #5234, opened 2026-03-14
[Feature] PPL Highlight Support

### LINKED ISSUE #5059: [FEATURE] Supporting Query Highlight Feature into PPL API

### LINKED ISSUE #5156: [FEATURE RFC] PPL Search Result Highlighting

RyanL1997:
### Description

> Preview of the updated documentation at my fork:  
https://github.com/RyanL1997/sql/blob/ppl-highlight-cmd/docs/user/ppl/interfaces/endpoint.md#highlight

This PR replaces the request-level `ThreadLocal` approach from #5141. Instead of passing highlight config through thread-local state, we thread a `HighlightConfig` record through the AST and Calcite plan — following the same pattern as `fetch_size` (#5109). The highlight config flows from the API request into the AST via `StatementBuilderContext`, becomes a `LogicalHighlight` Calcite node, and gets pushed down to the OpenSearch scan via an optimizer rule. This avoids the thread-safety concerns and implicit coupling of `ThreadLocal` and keeps the data flow explicit and traceable through the plan.

Updated design according to [review comment](https://github.com/opensearch-project/sql/pull/5234#issuecomment-4079004691)
> Done. Dropped LogicalHighlight, HighlightIndexScanRule, and the Highlight AST node (3 classes deleted). Highlight config now threads through Query → QueryPlan → QueryService → CalcitePlanContext, and pushDownHighlight() is called eagerly in visitRelation() before alias wrapping — same consume-once pattern as other scan hints. Introduced a HighlightPushDown interface in core to bridge the module boundary with opensearch (since CalciteRelNodeVisitor can't reference CalciteLogicalIndexScan directly). Config is cleared immediately after pushdown so it's applied exactly once.

- Add PPL search result highlighting via the `highlight` API parameter (not as a PPL command)
- Support both the simple array format (`["*"]`) and the rich OpenSearch Dashboards object format with `pre_tags`, `post_tags`, `fields`, and `fragment_size`
- Introduce `HighlightConfig` record to carry highlight options through the full pipeline: `PPLQueryRequest` → `AstStatementBuilder` → `Highlight` AST node → `CalciteRelNodeVisitor` → `LogicalHighlight` → `HighlightIndexScanRule` pushdown → OpenSearch `HighlightBuilder`
- Response includes a top-level `highlights` array parallel to `datarows`, containing per-field highlighted fragments
- Add endpoint documentation with examples and expected output

### Related Issues
* Resolve #5156 
* Resolve #5059
* Relate https://github.com/opensearch-project/sql/pull/5141

### Sample Request / Response
<details>  

```bash
curl -s -X POST "localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d '{"query":"search source=accounts \"Holmes\"","highlight":{"fields":{"*":{}},"pre_tags":["@opensearch-dashboards-highlighted-field@"],"post_tags":["@/opensearch-dashboards-highlighted-field@"],"fragment_size":2147483647}}' | jq
{
  "schema": [
    {
      "name": "account_number",
      "type": "bigint"
    },
    {
      "name": "firstname",
      "type": "string"
    },
    {
      "name": "address",
      "type": "string"
    },
    {
      "name": "balance",
      "type": "bigint"
    },
    {
      "name": "gender",
      "type": "string"
    },
    {
      "name": "city",
      "type": "string"
    },
    {
      "name": "employer",
      "type": "string"
    },
    {
      "name": "state",
      "type": "string"
    },
    {
      "name": "age",
      "type": "bigint"
    },
    {
      "name": "email",
      "type": "string"
    },
    {
      "name": "lastname",
      "type": "string"
    }
  ],
  "datarows": [
    [
      578,
      "Holmes",
      "969 Metropolitan Avenue",
      34259,
      "M",
      "Aguila",
      "Cubicide",
      "PA",
      37,
      "holmesmcknight@cubicide.com",
      "Mcknight"
    ],
    [
      828,
      "Blanche",
      "605 Stryker Court",
      44890,
      "F",
      "Loomis",
      "Motovate",
      "KS",
      33,
      "blancheholmes@motovate.com",
      "Holmes"
    ],
    [
      1,
      "Amber",
      "880 Holmes Lane",
      39225,
      "M",
      "Brogan",
      "Pyrami",
      "IL",
      32,
      "amberduke@pyrami.com",
      "Duke"
    ]
  ],
  "highlights": [
    {
      "firstname": [
        "@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"
      ],
      "firstname.keyword": [
        "@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"
      ]
    },
    {
      "lastname.keyword": [
        "@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"
      ],
      "lastname": [
        "@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"
      ]
    },
    {
      "address": [
        "880 @opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@ Lane"
      ]
    }
  ],
  "total": 3,
  "size": 3
}
```

</details>  

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/security/CalciteCrossClusterSearchIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/security/CalciteCrossClusterSearchIT.java
@@ -16,6 +16,7 @@ import static org.opensearch.sql.util.MatcherUtils.verifySchema;
 import java.io.IOException;
 import java.util.Locale;
 import org.apache.commons.text.StringEscapeUtils;
+import org.json.JSONArray;
 import org.json.JSONObject;
 import org.junit.Assert;
 import org.junit.Test;
@@ -485,10 +486,18 @@ public class CalciteCrossClusterSearchIT extends CrossClusterTestBase {
             String.format(
                 "search source=%s \\\"Hattie\\\" | fields firstname", TEST_INDEX_BANK_REMOTE),
             "[\"*\"]");
-    verifySchema(result, schema("firstname", "string"));
-    verifyDataRows(result, rows("Hattie"));
-    var highlights = result.getJSONArray("highlights");
-    var highlight = highlights.getJSONObject(0);
+    JSONArray schemaArray = result.getJSONArray("schema");
+    int hlIndex = -1;
+    for (int i = 0; i < schemaArray.length(); i++) {
+      if ("_highlight".equals(schemaArray.getJSONObject(i).getString("name"))) {
+        hlIndex = i;
+        break;
+      }
+    }
+    Assert.assertTrue("Schema should contain _highlight column", hlIndex >= 0);
+    JSONArray dataRows = result.getJSONArray("datarows");
+    Assert.assertTrue("Should have at least one row", dataRows.length() > 0);
+    var highlight = dataRows.getJSONArray(0).getJSONObject(hlIndex);
     Assert.assertTrue(
         "Highlight should contain <em>Hattie</em>",
         highlight.getJSONArray("firstname").getString(0).contains("<em>Hattie</em>"));


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -495,4 +495,49 @@ public boolean isScriptPushed() {
...
+    org.opensearch.search.fetch.subphase.highlight.HighlightBuilder hb =
+        new org.opensearch.search.fetch.subphase.highlight.HighlightBuilder();
+    if (fields.size() == 1 && "*".equals(fields.get(0))) {
+      // Wildcard: highlight search query matches in all fields
+      hb.field("*");
+    } else {
+      // Highlight specific terms across all fields
+      String queryStr =
+          fields.stream()
+              .map(term -> "\"" + term + "\"")
```
penghuo: User-supplied highlight terms are interpolated into a Lucene `query_string` with only double-quote wrapping (`"\"" + term + "\""`) and no escaping. A term like `foo\"` breaks out of the quoting and allows arbitrary Lucene query syntax injection, potentially surfacing field content in highlight fragments that the user did not search for. Sanitize by escaping backslashes and quotes before wrapping (`term.replace("\\", "\\\\").replace("\"", "\\\"")`) or switch to `matchPhraseQuery` per-term instead of `queryStringQuery`, which avoids query parsing entirely.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -495,4 +495,49 @@ public boolean isScriptPushed() {
...
+              .collect(java.util.stream.Collectors.joining(" OR "));
```
penghuo: Non-wildcard highlight fields (e.g., `["error", "login"]`) are treated as search terms for a new `query_string` highlight query on `"*"`, not as field names for scoping the original search query's highlights. This is semantically different from standard OpenSearch `highlight.fields` behavior where keys are field names to highlight. If a user sends `{"fields": {"message": {}}}` via the object format, the extracted key `"message"` gets treated as a search term. Consider using `hb.field(fieldName)` per field for standard field-based highlighting, or document very prominently that the simple array format takes search terms, not field names.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
```diff
@@ -495,4 +495,49 @@ public boolean isScriptPushed() {
...
+    // Apply pre_tags / post_tags if provided
+    if (highlightConfig.preTags() != null && !highlightConfig.preTags().isEmpty()) {
+      hb.preTags(highlightConfig.preTags().toArray(new String[0]));
+    }
+    if (highlightConfig.postTags() != null && !highlightConfig.postTags().isEmpty()) {
+      hb.postTags(highlightConfig.postTags().toArray(new String[0]));
+    }
+
+    // Apply fragment_size (default to Integer.MAX_VALUE when not specified)
+    hb.fragmentSize(
```
penghuo: When no `fragment_size` is specified, this defaults to `Integer.MAX_VALUE`, meaning each highlight fragment can be the entire field value. For large text fields this causes significant response size inflation. OpenSearch's own default is 100. Consider omitting the `fragmentSize()` call when not specified (letting OpenSearch use its default) or using a bounded default.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -82,6 +86,23 @@ public CalciteLogicalIndexScan(
         new PushDownContext(osIndex));
   }
 
+  public RelNode pushDownHighlight(HighlightConfig highlightConfig) {
```
penghuo, line 90: `pushDownHighlight` unconditionally appends `_highlight` to the schema without checking for duplicates. `LogicalHighlight.create` has a guard (`!getFieldNames().contains(HIGHLIGHT_FIELD)`), but the scan-level pushdown does not. If the Calcite planner fires `HighlightIndexScanRule` twice during exploration, this could produce a duplicate column. Add the same containment check here.

ppl/src/main/java/org/opensearch/sql/ppl/domain/PPLQueryRequest.java
```diff
@@ -109,4 +114,65 @@ public int getFetchSize() {
...
+  public HighlightConfig getHighlightConfig() {
+    if (jsonContent == null || !jsonContent.has(HIGHLIGHT_FIELD)) {
+      return null;
+    }
+
+    // Simple array format: ["*"] or ["error", "login"]
+    JSONArray arr = jsonContent.optJSONArray(HIGHLIGHT_FIELD);
+    if (arr != null) {
+      List<String> fields = new ArrayList<>();
+      for (int i = 0; i < arr.length(); i++) {
```
penghuo: No input validation on highlight config: no limit on field count (unbounded loop), no field name validation, no cap on `pre_tags`/`post_tags` array size or content, and `fragment_size` has no upper bound check. Malicious or malformed input flows deep into the system before being caught. Consider adding basic validation: cap array sizes (e.g., max 50 fields), validate field name patterns, bound `fragment_size` to a reasonable max.

docs/user/ppl/interfaces/endpoint.md
```diff
@@ -201,3 +201,88 @@ Expected output (trimmed):
...
+
+```bash ppl ignore
+curl -sS -H 'Content-Type: application/json' \
+  -X POST localhost:9200/_plugins/_ppl \
+  -d '{
+        "query": "source=accounts \"Holmes\"",
+        "highlight": {
+          "pre_tags": ["@opensearch-dashboards-highlighted-field@"],
+          "post_tags": ["@/opensearch-dashboards-highlighted-field@"],
+          "fields": {"*": {}},
```
penghuo, line 236: What is the meaning of fields? Why `{}` is required?
```
"fields": {"*": {}},
```

docs/user/ppl/interfaces/endpoint.md
```diff
@@ -201,3 +201,88 @@ Expected output (trimmed):
...
+    {
+      "firstname": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"],
+      "firstname.keyword": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"]
+    },
+    {
+      "lastname": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"],
+      "lastname.keyword": ["@opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@"]
+    },
+    {
+      "address": ["880 @opensearch-dashboards-highlighted-field@Holmes@/opensearch-dashboards-highlighted-field@ Lane"]
```
penghuo, line 262: Datarows does not include `address`, should address been highlighted?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
RyanL1997: You're right — the non-wildcard entries are field names to scope highlighting to, not search terms. Fixed by replacing the `queryStringQuery`/`matchPhraseQuery` branch with a simple loop:
```java
  for (String fieldName : fields) {
    hb.field(fieldName);
  }
```
This matches standard OpenSearch highlight.fields behavior where keys are field names. The wildcard special-case ("*") is now just handled uniformly by the same loop.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
RyanL1997: Good catch. On closer look for the [next comment](https://github.com/opensearch-project/sql/pull/5234/changes#r2948284540), the entire query-building branch was semantically wrong — non-wildcard entries are field names (which fields to highlight), not search terms. Replaced with a simple `hb.field(fieldName)` loop, which matches standard OpenSearch `highlight.fields` behavior. This also eliminates the injection surface entirely since there's no query construction at all.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteHighlightIT.java
```diff
@@ -0,0 +1,208 @@
...
+import org.json.JSONObject;
+import org.junit.Assert;
+import org.junit.jupiter.api.Test;
+import org.opensearch.client.Request;
+import org.opensearch.client.RequestOptions;
+import org.opensearch.client.Response;
+import org.opensearch.sql.ppl.PPLIntegTestCase;
+
+/** Integration tests for PPL highlight parameter. */
+public class CalciteHighlightIT extends PPLIntegTestCase {
```
penghuo, line 24: These are all valid search expressions that produce query_string queries in the DSL. The highlighting should work for all of them. Could u add IT to verify the behaviour. 
- Boolean full-text: search "error" OR "warning" source=t with highlight
- Multiple unquoted terms: search user email source=t with highlight
- Mixed full-text + structured filter: search "error" AND status>400 source=t with highlight
- Wildcards in search text: search log_* source=t with highlight
- NOT with full-text: search NOT "error" source=t with highlight

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/AbstractCalciteIndexScan.java
RyanL1997: Agreed and fixed — when fragment_size isn't specified, we should defer to OpenSearch's default (100) rather than forcing `Integer.MAX_VALUE`. Changed to only call `hb.fragmentSize()` when the user explicitly provides it. + I have also updated the doc to reflect the change, including the above field name fix.

penghuo:
Highlight is a scan hint, not a relational operator. Consider dropping `LogicalHighlight`, `HighlightIndexScanRule`, and the `Highlight` AST node. Instead, pass `HighlightConfig` through `CalcitePlanContext` and call `pushDownHighlight()` eagerly in `visitRelation()` — same pattern as existing `PushDownContext` operations. Simpler, fewer classes, same result.

cc: @dai-chen

RyanL1997:
> Highlight is a scan hint, not a relational operator. Consider dropping LogicalHighlight, HighlightIndexScanRule, and the Highlight AST node. Instead, pass HighlightConfig through CalcitePlanContext and call pushDownHighlight() eagerly in visitRelation() — same pattern as existing PushDownContext operations. Simpler, fewer classes, same result.

Done. Dropped `LogicalHighlight`, `HighlightIndexScanRule`, and the `Highlight` AST node (3 classes deleted). Highlight config now threads through `Query` → `QueryPlan` → `QueryService` → `CalcitePlanContext`, and `pushDownHighlight()` is called eagerly in `visitRelation()` before alias wrapping — same consume-once pattern as other scan hints. Introduced a `HighlightPushDown` interface in `core` to bridge the module boundary with `opensearch` (since `CalciteRelNodeVisitor` can't reference `CalciteLogicalIndexScan` directly). Config is cleared immediately after pushdown so it's applied exactly once.

cc @penghuo @dai-chen

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
RyanL1997: Reference to [comment](https://github.com/opensearch-project/sql/pull/5234#issuecomment-4079004691)


This one is auto-resolved by the refactor. `pushDownHighlight()` is now called exactly once in `visitRelation()` and the config is cleared immediately after (`context.setHighlightConfig(null)`), so there's no possibility of duplicate application. The optimizer rule path (which could fire multiple times during exploration) no longer exists.

ppl/src/main/java/org/opensearch/sql/ppl/domain/PPLQueryRequest.java
RyanL1997: Good catch. Added input validation in `PPLQueryRequest.getHighlightConfig()`: 
- field count capped at 100
- field names validated as non-empty strings, 
- `fragment_size` must be positive, 
- `pre_tags`/`post_tags` arrays capped at 10 entries each. 
- Invalid input now throws `IllegalArgumentException` early before reaching the engine.

docs/user/ppl/interfaces/endpoint.md
RyanL1997: The `"fields": {"*": {}}` format aligns with the [OpenSearch highlight API](https://docs.opensearch.org/latest/search-plugins/searching-data/highlight/). In OpenSearch's `_search` endpoint, the highlight `fields` parameter is an object where each key is a field name (or wildcard) and each value is a per-field options object. For example, the following PPL request with per-field highlight options:

```bash
  -d '{
    "query": "source=accounts \"Holmes\"",
    "highlight": {
      "fields": {
        "address": {"fragment_size": 200},
        "firstname": {"number_of_fragments": 3},
        "lastname": {}
      },
      "pre_tags": ["<b>"],
      "post_tags": ["</b>"]
    }
  }'
```

produces the same highlight fragments as the equivalent DSL query:

```bash
curl -X POST "localhost:9200/accounts/_search" \
  -d '{
    "query": {"query_string": {"query": "Holmes"}},
    "highlight": {
      "fields": {
        "address": {"fragment_size": 200},
        "firstname": {"number_of_fragments": 3},
        "lastname": {}
      },
      "pre_tags": ["<b>"],
      "post_tags": ["</b>"]
    }
  }'
```

docs/user/ppl/interfaces/endpoint.md
RyanL1997: Yes, this is intentional and aligns with OpenSearch DSL behavior. In OpenSearch's `_search` endpoint, the `highlight` response is independent of `_source` filtering — even if you use `"_source": ["firstname", "lastname"]`, the highlight section will still return fragments for any field that matched the query, including `address`. Our PPL highlight follows the same semantics: `"fields": {"*": {}}` highlights all fields that matched the search term, regardless of which fields are projected in the `schema`/`datarows`. Added a note in the docs to clarify this.

<details>
<summary>Verified against OpenSearch DSL — _source filtered but highlight still includes address</summary>

```bash
curl -X POST "localhost:9200/accounts/_search" \
  -d '{
    "query": {"query_string": {"query": "Holmes"}},
    "_source": ["firstname", "lastname"],
    "highlight": {
      "fields": {"*": {}},
      "pre_tags": ["<b>"],
      "post_tags": ["</b>"]
    }
  }'
```

```json
{
  "hits": {
    "hits": [
      {
        "_id": "578",
        "_source": { "firstname": "Holmes", "lastname": "Mcknight" },
        "highlight": {
          "firstname": ["<b>Holmes</b>"],
          "firstname.keyword": ["<b>Holmes</b>"]
        }
      },
      {
        "_id": "828",
        "_source": { "firstname": "Blanche", "lastname": "Holmes" },
        "highlight": {
          "lastname": ["<b>Holmes</b>"],
          "lastname.keyword": ["<b>Holmes</b>"]
        }
      },
      {
        "_id": "1",
        "_source": { "firstname": "Amber", "lastname": "Duke" },
        "highlight": {
          "address": ["880 <b>Holmes</b> Lane"]
        }
      }
    ]
  }
}
```

`_source` is filtered to only `firstname`/`lastname`, but highlight still returns `address` for account 1.
</details>

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteHighlightIT.java
RyanL1997: Good call out. Added at `integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteHighlightIT.java`. The following search expression patterns are now covered by IT tests:

| Test | PPL | DSL `query_string` |
|---|---|---|
| `testHighlightBooleanOrSearch` | `source=accounts "Holmes" OR "Bond"` | `(Holmes OR Bond)` |
| `testHighlightBooleanAndSearch` | `source=accounts "Holmes" AND "Lane"` | `(Holmes AND Lane)` |
| `testHighlightNotSearch` | `source=accounts NOT "Holmes"` | `NOT(Holmes)` |
| `testHighlightBooleanOrWithFilter` | `source=accounts "Holmes" OR "Bond" \| where age > 30` | OR + structured filter |
| `testHighlightWildcardInSearchText` | `source=accounts "Holm*"` | `Holm*` |

All produce `query_string` queries in the DSL. The `OR` and `AND` tests verify highlight fragments contain both search terms. The `NOT` test verifies the query succeeds — highlights are naturally empty since negated terms have no positive matches to highlight.

docs/user/ppl/interfaces/endpoint.md
```diff
@@ -201,3 +201,89 @@ Expected output (trimmed):
 - Plan node names use Calcite physical operator names (for example, `EnumerableCalc` or `CalciteEnumerableIndexScan`).
 - Plan `time_ms` is inclusive of child operators and represents wall-clock time; overlapping work can make summed plan times exceed `summary.total_time_ms`.
 - Scan nodes reflect operator wall-clock time; background prefetch can make scan time smaller than total request latency.
+
+## Highlight
+
+You can add a `highlight` parameter to the PPL request body to enable search-result highlighting. This parameter follows the same semantics as the [OpenSearch highlight API](https://docs.opensearch.org/latest/search-plugins/searching-data/highlight/). When enabled, the response includes a top-level `highlights` array containing matching fragments with the specified tags. Each entry in the `highlights` array corresponds to the row at the same index in `datarows`.
```
dai-chen: I see there is max tag/fields in query request. Did we mention it somewhere in the user doc?

core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -87,8 +88,17 @@ public void execute(
...
       QueryType queryType,
       ResponseListener<ExecutionEngine.QueryResponse> listener) {
+    execute(plan, queryType, null, listener);
+  }
+
+  /** Execute with optional highlight config. */
+  public void execute(
+      UnresolvedPlan plan,
+      QueryType queryType,
+      HighlightConfig highlightConfig,
```
penghuo, line 98: nit: If a third request-level parameter is added (timeout, routing, preference), that's the time to refactor to QueryOptions — the pattern exists (ExecutionContext, PlanContext), and it's a straightforward migration. No need to speculate now.

docs/user/ppl/interfaces/endpoint.md
RyanL1997: Good catch, just updated the doc

RyanL1997:
Transforming some offline conversation with @penghuo and @dai-chen :
For the current response in commit https://github.com/opensearch-project/sql/commit/e2ef1b2040d97979ca3205cd7da3042bade6ba5d:

<details>

```bash
curl -sS -X POST "localhost:9200/_plugins/_ppl" \
  -d '{
    "query": "source=accounts \"Holmes\"",
    "highlight": {
      "fields": {
        "address": {"fragment_size": 200},
        "firstname": {"number_of_fragments": 3},
        "lastname": {}
      },
      "pre_tags": ["<b>"],
      "post_tags": ["</b>"]
    }
  }' | python3 -m json.tool
{
    "schema": [
        {
            "name": "account_number",
            "type": "bigint"
        },
        {
            "name": "firstname",
            "type": "string"
        },
        {
            "name": "address",
            "type": "string"
        },
        {
            "name": "balance",
            "type": "bigint"
        },
        {
            "name": "gender",
            "type": "string"
        },
        {
            "name": "city",
            "type": "string"
        },
        {
            "name": "employer",
            "type": "string"
        },
        {
            "name": "state",
            "type": "string"
        },
        {
            "name": "age",
            "type": "bigint"
        },
        {
            "name": "email",
            "type": "string"
        },
        {
            "name": "lastname",
            "type": "string"
        }
    ],
    "datarows": [
        [
            578,
            "Holmes",
            "969 Metropolitan Avenue",
            34259,
            "M",
            "Aguila",
            "Cubicide",
            "PA",
            37,
            "holmesmcknight@cubicide.com",
            "Mcknight"
        ],
        [
            828,
            "Blanche",
            "605 Stryker Court",
            44890,
            "F",
            "Loomis",
            "Motovate",
            "KS",
            33,
            "blancheholmes@motovate.com",
            "Holmes"
        ],
        [
            1,
            "Amber",
            "880 Holmes Lane",
            39225,
            "M",
            "Brogan",
            "Pyrami",
            "IL",
            32,
            "amberduke@pyrami.com",
            "Duke"
        ]
    ],
    "highlights": [
        {
            "firstname": [
                "<b>Holmes</b>"
            ]
        },
        {
            "lastname": [
                "<b>Holmes</b>"
            ]
        },
        {
            "address": [
                "880 <b>Holmes</b> Lane"
            ]
        }
    ],
    "total": 3,
    "size": 3
}
```

</details>

We are proposing - instead of doing highlight in a separate response formatter to isolate everything out in json body, we should keep it in the original data rows / schemas, and let the frontend handle the transformation for render:

<details>

```bash
 curl -sS -X POST "localhost:9200/_plugins/_ppl" \
  -d '{
    "query": "source=accounts \"Holmes\"",
    "highlight": {
      "fields": {
        "address": {"fragment_size": 200},
        "firstname": {"number_of_fragments": 3},
        "lastname": {}
      },
      "pre_tags": ["<b>"],
      "post_tags": ["</b>"]
    }
  }' | python3 -m json.tool
{
    "schema": [
        {
            "name": "account_number",
            "type": "bigint"
        },
        {
            "name": "firstname",
            "type": "string"
        },
        {
            "name": "address",
            "type": "string"
        },
        {
            "name": "balance",
            "type": "bigint"
        },
        {
            "name": "gender",
            "type": "string"
        },
        {
            "name": "city",
            "type": "string"
        },
        {
            "name": "employer",
            "type": "string"
        },
        {
            "name": "state",
            "type": "string"
        },
        {
            "name": "age",
            "type": "bigint"
        },
        {
            "name": "email",
            "type": "string"
        },
        {
            "name": "lastname",
            "type": "string"
        },
        {
            "name": "_highlight",
            "type": "struct"
        }
    ],
    "datarows": [
        [
            578,
            "Holmes",
            "969 Metropolitan Avenue",
            34259,
            "M",
            "Aguila",
            "Cubicide",
            "PA",
            37,
            "holmesmcknight@cubicide.com",
            "Mcknight",
            {
                "firstname": [
                    "<b>Holmes</b>"
                ]
            }
        ],
        [
            828,
            "Blanche",
            "605 Stryker Court",
            44890,
            "F",
            "Loomis",
            "Motovate",
            "KS",
            33,
            "blancheholmes@motovate.com",
            "Holmes",
            {
                "lastname": [
                    "<b>Holmes</b>"
                ]
            }
        ],
        [
            1,
            "Amber",
            "880 Holmes Lane",
            39225,
            "M",
            "Brogan",
            "Pyrami",
            "IL",
            32,
            "amberduke@pyrami.com",
            "Duke",
            {
                "address": [
                    "880 <b>Holmes</b> Lane"
                ]
            }
        ]
    ],
    "total": 3,
    "size": 3
}
```

</details>

PR #5241, opened 2026-03-18
Support PPL timewrap command 

### LINKED ISSUE #5237: [RFC] PPL timewrap command for time-period comparison
## Summary

Add a `timewrap` command to PPL that reshapes `timechart` output by splitting time series into recurring periods for side-by-side comparison (e.g., day-over-day, week-over-week).

## Syntax

```
... | timechart <agg> ... | timewrap <span> [align=now|end]
```

- **`<span>`**: Wrapping interval — `[int]<timescale>` (e.g., `1day`, `2week`). Integer defaults to 1 if omitted.
- **`align`**: `end` (default) aligns periods to search end time; `now` aligns to current time.
- Reuses the existing `spanLiteral` grammar rule — all PPL timescale aliases work (`s`, `sec`, `min`, `hr`, `day`, `week`, `mon`, `month`, `qtr`, `year`, etc.).

## What it does

Given timechart output with hourly counts over 3 weeks:

```
| @timestamp           | count() |
|----------------------|---------|
| 2024-06-17 00:00:00  | 10      |
| 2024-06-17 01:00:00  | 15      |
| ...                  | ...     |
| 2024-07-07 23:00:00  | 13      |
```

`timewrap 1week` produces:

```
| @timestamp | 1week_before | 2weeks_before | 3weeks_before |
|------------|--------------|---------------|---------------|
| 00:00:00   | 9            | 11            | 10            |
| 01:00:00   | 13           | 8             | 15            |
| ...        | ...          | ...           | ...           |
| 23:00:00   | 13           | 14            | 12            |
```

Columns named `<N><timescale>_before` — `1week_before` is the most recent period, counting backward. Singular for N=1, plural for N>1.

With a `BY` clause: `count() host-a: 1week_before`, `count() host-a: 2weeks_before`, etc.

### 1. Post-processing command, not standalone

`timewrap` only operates on `timechart` output — it is a display-time reshaping command (like `transpose` or `trendline`). Using it without a preceding `timechart` throws a semantic error. Validation is done in the Calcite visitor rather than at parse time, allowing potential intermediate commands (e.g., `where`) between timechart and timewrap.

### 2. Calcite PIVOT-based transformation

The core transformation:
1. Assign each row a **period number**: `FLOOR((end - timestamp) / span) + 1`
2. Normalize timestamps to **offset within period**: `timestamp MOD span`
3. **PIVOT** on period number, grouped by offset

Reuses Calcite's existing PIVOT infrastructure (same approach as `transpose`).

### 3. `m` means minutes, not months

Some query languages use `m` for months in timewrap. PPL keeps `m` = **minutes**, consistent with the existing `spanLiteral` grammar used by `timechart`, `bin`, and all other span-aware commands. Use `mon`/`month` for months.

### 4. Incomplete periods filled with null

If the search time range doesn't divide evenly by the wrap span, the first/last period may have fewer rows. Missing values are `null`.

### 5. Scope

- All timescales that `timechart` span supports
- `timechart` with and without `BY` clause
- `align=end` (default) and `align=now`

## Test plan

- Custom test data spanning multiple days (existing test data is too narrow for timewrap)
- Day-over-day, week-over-week wrapping
- BY clause producing per-group period columns
- Incomplete periods with null fill
- Semantic error when no timechart predecessor
- Backward compatibility with existing timechart tests

Full design doc: `docs/dev/ppl-timewrap-command.md`

ahkcs:
## Description

Adds the `timewrap` command to PPL, which reshapes `timechart` output by wrapping each time period into separate columns for side-by-side comparison (day-over-day, week-over-week, etc.). Behavior verified against Splunk.

**Syntax:** `... | timechart <agg> | timewrap <span> [align=end|now]`

### Key features
- Column naming matches Splunk: `<agg>_<N><unit>_before`, `<agg>_latest_<unit>`, `<agg>_<N><unit>_after`
- Column order: oldest first (leftmost)
- Dynamic column count: only populated period columns appear (no trailing nulls)
- `align=end` (default): uses WHERE upper bound as reference for deterministic column names
- `align=now`: uses current time as reference
- Supports all timescales: sec, min, hr, day, week, month, quarter, year

### Implementation
- Grammar: `timewrapCommand` rule reusing existing `spanLiteral`
- AST: `Timewrap` node with unit, value, align
- Calcite: PIVOT-based transformation with `UNIX_TIMESTAMP`/`FROM_UNIXTIME` for epoch conversion
- Execution engine: post-processing to strip all-null columns, rename to absolute offsets using `__base_offset__`, and handle `latest_<unit>` naming
- WHERE upper bound extraction: walks AST to find `@timestamp <=` filter for `align=end` reference

### Known limitations
- `BY` clause not yet supported
- Month/quarter/year use approximate fixed-length conversions (30d, 91d, 365d)
- Max 20 period columns
- Timestamps show data dates, not reference date (Splunk rewrites to reference date)

## Testing
- 26 integration tests covering all timescales, aggregation types, WHERE filters, align modes, null fill, and edge cases
- 5 doctest examples in user documentation
- All existing timechart and transpose tests pass (backward compatible)

## Issues Resolved

Resolves #5237

## Check List

- [x] New functionality has been documented
- [x] Commits are signed per the DCO using `--signoff`


core/src/main/java/org/opensearch/sql/calcite/utils/TimewrapPivot.java
```diff
@@ -0,0 +1,208 @@
...
+    String singular = parts[1];
+    String plural = parts[2];
+
+    long absolutePeriod = (baseOffset + relativePeriod - 1) * spanValue;
+
+    String mode = seriesMode == null ? "relative" : seriesMode;
+
+    return switch (mode) {
+      // series=exact (+ time_format) is not yet implemented; it intentionally falls back to the
+      // short "s<N>" naming. TODO: format the period start date with time_format.
```
noCharger, line 192: Is this tracked somewhere?

core/src/main/java/org/opensearch/sql/calcite/utils/TimewrapUtils.java
```diff
@@ -0,0 +1,348 @@
...
+  public static long spanToSeconds(SpanUnit unit, int value) {
+    return switch (unit.getName()) {
+      case "s" -> value;
+      case "m" -> value * 60L;
+      case "h" -> value * 3_600L;
+      case "d" -> value * 86_400L;
+      case "w" -> value * 7L * 86_400L;
+      case "M" -> value * 30L * 86_400L;
+      case "q" -> value * 91L * 86_400L;
+      case "y" -> value * 365L * 86_400L;
```
noCharger: Should this be approximate or exact? What's expected for leap-year February boundary?

core/src/main/java/org/opensearch/sql/calcite/utils/TimewrapUtils.java
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  private static Long parseTimestampLiteral(UnresolvedExpression expr) {
+    if (expr instanceof Literal lit && lit.getValue() instanceof String s) {
+      try {
+        java.time.LocalDateTime ldt =
+            java.time.LocalDateTime.parse(
+                s, java.time.format.DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"));
+        return ldt.toEpochSecond(java.time.ZoneOffset.UTC);
+      } catch (Exception e) {
```
noCharger: Narrow to catch (DateTimeParseException)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3823,6 +3827,148 @@ public RelNode visitChart(Chart node, CalcitePlanContext context) {
     return relBuilder.peek();
   }
 
+  @Override
+  public RelNode visitTimewrap(Timewrap node, CalcitePlanContext context) {
+    visitChildren(node, context);
+
+    // Signal the execution engine to strip all-null columns and rename with absolute offsets
+    CalcitePlanContext.stripNullColumns.set(true);
```
noCharger, line 3835: Can we add a leak-guard test: a timewrap query followed by a non-timewrap query on the same pooled thread, asserting the second result has no __base_offset__/__period__ artifacts?

core/src/main/java/org/opensearch/sql/calcite/utils/TimewrapPivot.java
ahkcs: Not previously due to low usage — opened a follow-up issue https://github.com/opensearch-project/sql/issues/5606 to track `series=exact` with `time_format`

core/src/main/java/org/opensearch/sql/calcite/utils/TimewrapUtils.java
ahkcs: `spanToSeconds` is only used on the fixed-length path (s/m/h/d/w), where the second count is exact — no calendar variability. The variable-length units (M/q/y) don't use this method; they go through `calendarUnitNumber`/`calendarUnitStartEpoch`, which handle the Feb/leap-year boundary precisely via `cumDaysBeforeMonth`. I've updated `spanToSeconds` to throw on M/q/y so the approximate values can never be silently used, and clarified the javadoc.

core/src/main/java/org/opensearch/sql/calcite/utils/TimewrapUtils.java
ahkcs: Done — narrowed both to `DateTimeParseException`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Added `TimewrapSignalsLeakTest` in 41f00eee9

noCharger (approved):
Thanks for the change!

PR #5254, opened 2026-03-23
Add query cancellation support via _tasks/_cancel API for PPL queries

### LINKED ISSUE #4887: [FEATURE] Support task cancellation for long-running PPL queries.
Followup to #4857, #4886 -- if a query has a very long timeout and is unexpectedly taking longer than expected, we should be able to cancel queries via the `_tasks/_cancel` API.

For the most part cancellation should already be supported fundamentally by #4857, we just need to integrate with the actual task endpoint and let that interrupt the query thread. We may also need to support canceling whatever background IO is running during that process (but I think this will self-interrupt anyway, since the socket will get closed & OpenSearch detects that).

N/A

N/A

sunil9977:
Fixes #4887

1.   PPL queries now register as CancellableTask, making them visible in GET /_tasks and cancellable via POST /_tasks/{task_id}/_cancel
2. Supports optional queryId in request body for task identification


ahkcs:
Hi @sunil9977, thanks for the changes! Please take a look at the CI failures

plugin/src/main/java/org/opensearch/sql/plugin/transport/SQLQueryTask.java
```diff
@@ -0,0 +1,44 @@
...
+    @Override
+    public boolean shouldCancelChildrenOnCancellation() {
+        return false;
+    }
+
+    public void setExecutionThread(Thread thread) {
+        executionThread.set(thread);
+    }
+
+    public void clearExecutionThread() {
```
ahkcs: Minor: there's a small window between task registration and `setExecutionThread()` where a cancel would see a null thread and be a no-op. In practice this is nearly impossible to hit (cancellation requires separate HTTP roundtrips), but it could matter if the sql-worker pool is saturated.

Since the fix is small, might be worth hardening:

```java
private final AtomicBoolean cancelled = new AtomicBoolean(false);

public void setExecutionThread(Thread thread) {
    executionThread.set(thread);
    if (cancelled.get()) {
        thread.interrupt();
    }
}

@Override
public void onCancelled() {
    cancelled.set(true);
    Thread thread = executionThread.get();
    if (thread != null) {
        thread.interrupt();
    }
}
```

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchQueryManager.java
```diff
@@ -33,22 +34,44 @@ public class OpenSearchQueryManager implements QueryManager {
   public static final String SQL_WORKER_THREAD_POOL_NAME = "sql-worker";
   public static final String SQL_BACKGROUND_THREAD_POOL_NAME = "sql_background_io";
```
ahkcs, line 36: Nit: could be `private static` since the accessor methods already exist.

ahkcs (commented):
Overall the approach looks good — aligns well with the plan from the issue discussion. Left a couple of minor suggestions inline.

plugin/src/main/java/org/opensearch/sql/plugin/transport/SQLQueryTask.java
```diff
@@ -0,0 +1,44 @@
+/*
```
ahkcs: We can rename this file to `PPLQueryTask`

ahkcs:
CI failure seems to be caused by spotless check, please run `./gradlew spotlessApply` to fix this

ahkcs:
During e2e testing with the OSD cancel button, we noticed that `Thread.interrupt()` only takes effect when the execution thread hits a blocking call (I/O, `Future.get()`, etc.). For CPU-bound queries like large joins, this can mean a 50+ second delay between clicking cancel and the query actually stopping.

Adding an interrupt check in `OpenSearchIndexEnumerator.moveNext()` would make cancellation near-instant, since `moveNext()` is called for every row:

```java
// opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/OpenSearchIndexEnumerator.java

@Override
public boolean moveNext() {
    if (Thread.interrupted()) {
        throw new NonFallbackCalciteException(
            String.format("Query was cancelled after processing %d rows.", queryCount));
    }

    if (queryCount >= maxResponseSize) {
        return false;
    }
    // ... rest of method
}
```

This is outside the scope of this PR but would be a nice follow-up to make the cancel experience snappy. In our testing it brought cancel response time from ~50s down to <5s.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchQueryManager.java
Swiddis: These consts are directly referenced in a few places iirc, I originally refactored to this from a lot of places duplicating a `"sql-worker"` string.

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryRequest.java
```diff
@@ -147,12 +158,30 @@ public ActionRequestValidationException validate() {
...
+      return new SQLQueryTask(id, type, action, getDescription() , parentTaskId, headers);
+  }
+
+  @Override
+  public String getDescription()
+  {
+      String prefix = (queryId != null) ? "PPL [queryId=" + queryId + "]: " : "PPL: ";
+
+      if (pplQuery != null && pplQuery.length() > 512) {
+          return prefix + pplQuery.substring(0,512) + "...";
```
Swiddis: suggestion: I don't think this is worth truncating.

The only way to access this description is to directly check the task by ID. Seeing the full query might be useful debug context. Similar debug logic in OSD core exists and doesn't truncate.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchQueryManager.java
ahkcs: Sorry, I wasn't referring to the thread pool name constants — those should stay public static. I meant the `cancellationCallBackThreadLocal` field added in this PR (line 45). It's declared public static but has setCancellationCallback/clearCancellationCallback accessor methods, so the field itself could be `private static` to prevent direct access.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchQueryManager.java
```diff
@@ -70,13 +93,22 @@ private void schedule(NodeClient client, Runnable task, TimeValue timeout) {
 
                 // Special-case handling of timeout-related interruptions
                 if (Thread.interrupted() || e.getCause() instanceof InterruptedException) {
+                    if (callBack != null && callBack.isCancelled()) {
+                        LOG.info("Query was cancelled");
+                        throw new OpenSearchException("Query was cancelled.");
```
Swiddis: issue: This exception will propagate as a 500, which is bad for people monitoring errors.

OpenSearchException is the generic top-level exception class, it defaults as internal server error when the status isn't set. You want to use one of its subclasses.

[TaskCancelledException](https://github.com/opensearch-project/OpenSearch/blob/568189b2fc4bdb9581bc22cf5ad576baf8010d01/libs/core/src/main/java/org/opensearch/core/tasks/TaskCancelledException.java#L44) seems like a better choice.

plugin/src/main/java/org/opensearch/sql/plugin/transport/SQLQueryTask.java
Swiddis: I don't see any precedent for this in OS core, I wonder if we're missing something? Only 2 hits:

```
% rg -w 'onCancelled' -B2 -A10
server/src/main/java/org/opensearch/index/reindex/BulkByScrollTask.java
203-
204-    @Override
205:    public void onCancelled() {
206-        /*
207-         * If this task is a leader, we don't need to do anything extra because the cancel action cancels child tasks for us
208-         * If it's is a worker, we know how to cancel it here
209-         * If we don't know whether it's a leader or worker yet, we do nothing here. If the task is later set to be a worker, we cancel the
210-         * worker at that time.
211-         */
212-        if (isWorker()) {
213-            workerState.handleCancel();
214-        }
215-    }

server/src/main/java/org/opensearch/tasks/CancellableTask.java
93-        assert reason != null;
94-        if (cancelledInfo.trySet(new CancelledInfo(reason))) {
95:            onCancelled();
96-        }
97-    }
98-
99-    public boolean isCancelled() {
100-        return cancelledInfo.get() != null;
101-    }
102-
103-    /**
104-     * Returns true if this task can potentially have children that need to be cancelled when it parent is cancelled.
105-     */
--
113-     * Called after the task is cancelled so that it can take any actions that it has to take.
114-     */
115:    protected void onCancelled() {}
116-
117-    /**
118-     * Returns true if this task should be automatically cancelled if the coordinating node that
119-     * requested this task left the cluster.
120-     */
121-    public boolean cancelOnParentLeaving() {
122-        return true;
123-    }
124-
125-    @Nullable
```

Swiddis:
> During e2e testing with the OSD cancel button, we noticed that Thread.interrupt() only takes effect when the execution thread hits a blocking call (I/O, Future.get(), etc.). For CPU-bound queries like large joins, this can mean a 50+ second delay between clicking cancel and the query actually stopping.

That's on me lol, I thought when I implemented it that the CPU part of joins would be fast (the bulk of the time on most joins is waiting for batches), so it'd be overkill to check every iteration and we could just throw when we request new batches. Happy to take it up if there's a followup task. Do you have more precise repro steps? Since I haven't noticed this issue before.

I'm also interested in if doing this on every row would have a negative effect on benchmarks.

plugin/src/main/java/org/opensearch/sql/plugin/transport/SQLQueryTask.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final AtomicReference<Thread> executionThread = new AtomicReference<>();
+
+    public SQLQueryTask(long id, String type, String action, String description, TaskId parentTaskId, Map<String, String> headers) {
+        super(id, type, action, description, parentTaskId, headers);
+    }
+
+    @Override
+    public boolean shouldCancelChildrenOnCancellation() {
+        return false;
```
Swiddis: Should this be false? Query tasks tend to spawn background IO tasks, but I don't know if those count as children.

ahkcs:
> Do you have more precise repro steps? 

I was doing a `join` query from OSD using POC OSD side change for the cancel button. And after query cancelled, the stacktrace showed up 50 seconds later

ahkcs:
Based on @Swiddis's [observation](https://github.com/opensearch-project/sql/pull/5254#discussion_r2977856846) that almost no core code overrides `onCancelled()` — it looks like core's cancellation model is cooperative: `cancel()` flips `isCancelled()` to true, and the task is expected to poll that flag and stop itself (e.g., search tasks check `task.isCancelled()` per Lucene segment).

I think a simpler approach might work here: pass the `CancellableTask` reference to the worker thread, and check `task.isCancelled()` in `OpenSearchIndexEnumerator.moveNext()`. This would remove the need for the `CancellationCallBack` interface, the ThreadLocal callback, `onCancelled()` override, thread interruption, and the `AtomicReference<Thread>` — all replaced by one `isCancelled()` check per row.

Swiddis:
Oh, interesting. If I'm understanding right they implemented their own thread interruption mechanism? I wonder why not use regular `interrupt()`. Given we already implemented timeouts via interrupt, I wonder if it's worthwhile to refactor everything or we should just have cancellation point to interrupt. Open to either but given the current impl it'd be less work to stick with interrupts in the core places.

Swiddis:
Ok, talked with the core folks, task cancellation is the better approach. Since opensearch uses a thread pool system, interrupting the thread has potential to poison the whole pool. It works for us because we're careful to uninterrupt the thread when handling interruption errors, but there might theoretically be a path that leaves a perpetually-interrupted thread (either currently or in the future). Cancellation also has the benefit of propagating across clusters better (which is more relevant to Core which is fanning out requests to many data nodes regularly).

So the correct approach is change our thread interrupts to use the native task cancellation feature. If all that understanding is correct, it means we can also clean up the logic that un-interrupts the thread, but I think it's only like 2 lines anyway so not a massive win.

plugin/src/main/java/org/opensearch/sql/plugin/transport/SQLQueryTask.java
sunil9977: Yes, updated it to true.

sunil9977:
Thank you all for reviews.
I've refactored it to use existing methods rather than using thread.interrupt().
Please let me know if my understanding is correct.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchQueryManager.java
Swiddis, line 76: future PR: this interrupt needs to be changed to cancellation as well (and all code paths that depend on it)

Swiddis:
@sunil9977 LGTM, one more chore: please add DCO signoffs to the commits as described in https://github.com/opensearch-project/sql/pull/5254/checks?check_run_id=68346380999 & https://github.com/opensearch-project/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin.

Issue #5247, opened 2026-03-18
[FEATURE] PPL query support for Analytics engine integration


dai-chen:
**Is your feature request related to a problem?**

PPL queries currently only execute against Lucene-backed indices. With the Analytics engine providing Parquet-backed storage, PPL queries targeting non-Lucene indices need to be routed through the unified query pipeline to the Analytics engine for execution. The Unified Query API already supports PPL V3 Calcite-based RelNode generation, but the end-to-end integration — query routing, schema building, execution handoff, and response formatting — is not yet ready.

Technical requirements:

- PPL queries against non-Lucene indices return correct results through _plugins/_ppl endpoint
- PPL explain API (_plugins/_ppl/_explain) returns the logical plan handed off to the Analytics engine
- Default response format (JDBC JSON with schema, datarows, total, size, status) matches existing PPL response format
- Clear error messages distinguishing client errors (query parsing/planning in SQL plugin) from server errors (query optimization/distributed planning/execution from Analytics engine)
- Observability: metrics and latency tracking across both SQL plugin (routing, parsing, planning) and Analytics engine (optimization, execution)
- PPL queries on Lucene indices are unaffected (no regression)

**What solution would you like?**

1. **Plugin wiring and dependency integration**: Add analytics-engine as extendedPlugins dependency, resolve Calcite jar conflicts (classloader sharing and bundlePlugin excludes), wire `SchemaBuilder` and `QueryPlanExecutor` from the Analytics engine via Guice.
2. **Query routing and execution handoff**: Add `RestUnifiedQueryAction` and `AnalyticsExecutionEngine` that detect non-Lucene indices, route PPL queries through `UnifiedQueryPlanner.plan()` → `QueryPlanExecutor.execute()`, and schedule execution on sql-worker thread pool with security context propagation.
3. **Response formatting and explain support**: Format `Iterable<Object[]>` results via existing `JdbcResponseFormatter`; return logical `RelNode` plan via `_plugins/_ppl/_explain`, include physical plan if Analytics engine provides `QueryPlanExecutor.explain()` API.
4. **Error handling and observability**: Client vs server / SQL plugin vs Analytics engine error classification, query size limit enforcement, request/failure metrics, and planning/execution latency logging.
5. **Integration and regression tests**: End-to-end ITs with analytics-engine plugin verifying PPL query, explain, response format, error handling, non-Lucene routing, and Lucene regression.

**What alternatives have you considered?**

[...]

PR #5264, opened 2026-03-24
Add query routing and execution handoff for Parquet-backed indices

ahkcs:
## Summary
Implements query routing and execution handoff for Project Mustang's unified query pipeline ([#5247](https://github.com/opensearch-project/sql/issues/5247)).

This PR covers **Step 2** and **Step 3** (partial) from the issue:

- **Index detection**: `isAnalyticsIndex()` extracts index name from PPL query via regex, checks for `parquet_` prefix. Production will use index settings (e.g., `index.storage_type`).
- **Query routing**: `RestPPLQueryAction.prepareRequest()` branches -- parquet indices go to `RestUnifiedQueryAction`, everything else goes to existing Lucene path unchanged.
- **Planning**: `UnifiedQueryPlanner.plan()` parses PPL and generates Calcite `RelNode` against a stub schema (will be replaced by `EngineContext.getSchema()` from the analytics engine).
- **Execution**: `AnalyticsExecutionEngine` calls `QueryPlanExecutor.execute(relNode)`, converts `Iterable<Object[]>` to `QueryResponse` with type mapping via `OpenSearchTypeFactory` and query size limit enforcement.
- **Thread scheduling**: Runs on `sql-worker` thread pool via `client.threadPool().schedule()`.
- **Stub executor**: `StubQueryPlanExecutor` returns canned data for `parquet_logs` and `parquet_metrics` tables for development and testing.

- **Response formatting**: `QueryResponse` -> `QueryResult` with `PPL_SPEC` -> `JdbcResponseFormatter` produces standard JDBC JSON (`schema`, `datarows`, `total`, `size`, `status`). Same format as existing PPL responses.
- **Explain support**: Deferred to next PR.

- **Error classification**: `SyntaxCheckException`/`SemanticCheckException` -> HTTP 400 + `PPL_FAILED_REQ_COUNT_CUS`; engine failures -> HTTP 500 + `PPL_FAILED_REQ_COUNT_SYS`.
- **Latency logging**: `[unified] Planning completed in Xms`, `[unified] Execution completed in Xms, N rows returned`.

### Not in this PR (next PR / blocked)
- **Step 1** (plugin wiring): Blocked -- analytics-engine JAR not available yet
- **Step 3** (explain): Will be in next PR
- **Step 5** (full integration tests with real analytics engine): Blocked -- will use stub for now

## New Files
| File | Purpose |
|---|---|
| `core/.../analytics/QueryPlanExecutor.java` | `@FunctionalInterface` -- contract for analytics engine execution |
| `core/.../analytics/AnalyticsExecutionEngine.java` | `ExecutionEngine` impl -- converts raw rows to `QueryResponse` |
| `plugin/.../rest/RestUnifiedQueryAction.java` | Orchestrates analytics query path end-to-end |
| `plugin/.../rest/StubQueryPlanExecutor.java` | Mock data for development/testing |
| `integ-test/.../ppl/AnalyticsPPLIT.java` | Integration tests -- full pipeline verification |

## Modified Files
| File | Change |
|---|---|
| `plugin/.../rest/RestPPLQueryAction.java` | Added routing branch for `parquet_` indices |
| `plugin/.../SQLPlugin.java` | Passes `ClusterService` + `NodeClient` to `RestPPLQueryAction` |
| `plugin/build.gradle` | Added `:api` dependency for `UnifiedQueryPlanner` |

## Test plan
- 10 unit tests -- `AnalyticsExecutionEngine` (type mapping, row conversion, size limit, nulls, error propagation, PhysicalPlan rejection)
- 10 unit tests -- `RestUnifiedQueryAction` (index extraction regex, routing detection for various PPL patterns)
- 9 integration tests -- `AnalyticsPPLIT` (full pipeline: schema+data verification with `verifySchema`/`verifyDataRows`, response format, projection, syntax error -> 400, Lucene regression)

### DIFF
--- a/core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
+++ b/core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.executor.analytics;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.opensearch.sql.calcite.CalcitePlanContext;
+import org.opensearch.sql.calcite.utils.OpenSearchTypeFactory;
+import org.opensearch.sql.common.response.ResponseListener;
+import org.opensearch.sql.data.model.ExprTupleValue;
+import org.opensearch.sql.data.model.ExprValue;
+import org.opensearch.sql.data.model.ExprValueUtils;
+import org.opensearch.sql.data.type.ExprType;
+import org.opensearch.sql.executor.ExecutionContext;
+import org.opensearch.sql.executor.ExecutionEngine;
+import org.opensearch.sql.executor.pagination.Cursor;
+import org.opensearch.sql.planner.physical.PhysicalPlan;
+
+/**
+ * Execution engine adapter for the analytics engine (Project Mustang).
+ *
+ * <p>Bridges the analytics engine's {@link QueryPlanExecutor} with the SQL plugin's {@link
+ * ExecutionEngine} response pipeline. Takes a Calcite {@link RelNode}, delegates execution to the
+ * analytics engine, and converts the raw results into {@link QueryResponse}.
+ */
+public class AnalyticsExecutionEngine implements ExecutionEngine {
+
+  private final QueryPlanExecutor planExecutor;
+
+  public AnalyticsExecutionEngine(QueryPlanExecutor planExecutor) {
+    this.planExecutor = planExecutor;
+  }
+
+  /** Not supported. Analytics queries use the RelNode path exclusively. */
+  @Override
+  public void execute(PhysicalPlan plan, ResponseListener<QueryResponse> listener) {
+    listener.onFailure(
+        new UnsupportedOperationException("Analytics engine only supports RelNode execution"));
+  }
+
+  /** Not supported. Analytics queries use the RelNode path exclusively. */
+  @Override
+  public void execute(
+      PhysicalPlan plan, ExecutionContext context, ResponseListener<QueryResponse> listener) {
+    listener.onFailure(
+        new UnsupportedOperationException("Analytics engine only supports RelNode execution"));
+  }
+
+  /** Not supported. Analytics queries use the RelNode path exclusively. */
+  @Override
+  public void explain(PhysicalPlan plan, ResponseListener<ExplainResponse> listener) {
+    listener.onFailure(
+        new UnsupportedOperationException("Analytics engine only supports RelNode execution"));
+  }
+
+  @Override
+  public void execute(
+      RelNode plan, CalcitePlanContext context, ResponseListener<QueryResponse> listener) {
+    try {
+      Integer querySizeLimit = context.sysLimit.querySizeLimit();
+      Iterable<Object[]> rows = planExecutor.execute(plan, null);
+
+      List<RelDataTypeField> fields = plan.getRowType().getFieldList();
+      List<ExprValue> results = convertRows(rows, fields, querySizeLimit);
+      Schema schema = buildSchema(fields);
+
+      listener.onResponse(new QueryResponse(schema, results, Cursor.None));
+    } catch (Exception e) {
+      listener.onFailure(e);
+    }
+  }
+
+  private List<ExprValue> convertRows(
+      Iterable<Object[]> rows, List<RelDataTypeField> fields, Integer querySizeLimit) {
+    List<ExprValue> results = new ArrayList<>();
+    for (Object[] row : rows) {
+      if (querySizeLimit != null && results.size() >= querySizeLimit) {
+        break;
+      }
+      Map<String, ExprValue> valueMap = new LinkedHashMap<>();
+      for (int i = 0; i < fields.size(); i++) {
+        String columnName = fields.get(i).getName();
+        Object value = (i < row.length) ? row[i] : null;
+        valueMap.put(columnName, ExprValueUtils.fromObjectValue(value));
+      }
+      results.add(ExprTupleValue.fromExprValueMap(valueMap));
+    }
+    return results;
+  }
+
+  private Schema buildSchema(List<RelDataTypeField> fields) {
+    List<Schema.Column> columns = new ArrayList<>();
+    for (RelDataTypeField field : fields) {
+      ExprType exprType = convertType(field.getType());
+      columns.add(new Schema.Column(field.getName(), null, exprType));
+    }
+    return new Schema(columns);
+  }
+
+  private ExprType convertType(RelDataType type) {
+    try {
+      return OpenSearchTypeFactory.convertRelDataTypeToExprType(type);
+    } catch (IllegalArgumentException e) {
+      return org.opensearch.sql.data.type.ExprCoreType.UNKNOWN;
+    }
+  }
+}
--- a/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
@@ -17,6 +17,7 @@ import java.util.Set;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchException;
+import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.core.action.ActionListener;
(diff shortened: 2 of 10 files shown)


plugin/src/test/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryActionTest.java
```diff
@@ -0,0 +1,96 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class RestUnifiedQueryActionTest {
+
+  // --- isAnalyticsIndex ---
+
+  @Test
```
dai-chen, line 19: FYI, I will add a generic `UnifiedQueryParser` for this extraction. Probably you simplify the test below and also think about the test scope between this and `AnalyticsExecutionEngineTest`.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -0,0 +1,263 @@
...
+            .add("cpu", SqlTypeName.DOUBLE)
+            .add("memory", SqlTypeName.DOUBLE)
+            .add("host", SqlTypeName.VARCHAR)
+            .build();
+      }
+    };
+  }
+
+  /** Classify whether the exception is a client error (bad query) or server error (engine bug). */
+  private static boolean isClientError(Exception e) {
```
dai-chen, line 227: I recall we have a long list in current SQL action. Could you double check?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -0,0 +1,263 @@
...
+      }
+    };
+  }
+
+  /**
+   * Stub schema for development and testing. Returns a hardcoded table definition for any
+   * "parquet_*" table. Will be replaced by EngineContext.getSchema() when the analytics engine is
+   * ready.
+   */
+  private static AbstractSchema buildStubSchema() {
```
dai-chen, line 185: Could you make this a stub class and move it with StubQueryPlanExecutor (and maybe your temporary extraction logic together?) into a sub package? I'm just thinking this makes clear what we're stubbing.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -0,0 +1,263 @@
...
+        isClientError(e) ? RestStatus.BAD_REQUEST : RestStatus.INTERNAL_SERVER_ERROR;
+    String reason = e.getMessage() != null ? e.getMessage() : "Unknown error";
+    // Escape characters that would break JSON
+    reason =
+        reason.replace("\\", "\\\\").replace("\"", "\\\"").replace("\n", "\\n").replace("\r", "");
+    channel.sendResponse(
+        new BytesRestResponse(
+            status,
+            "application/json; charset=UTF-8",
+            "{\"error\":{\"type\":\""
```
dai-chen, line 255: I forgot is this the standard way we do in SQL/PPL rest action today?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -0,0 +1,263 @@
...
+    return e instanceof SyntaxCheckException
+        || e instanceof SemanticCheckException
+        || e instanceof IllegalArgumentException
+        || e instanceof NullPointerException;
+  }
+
+  private static void recordFailureMetric(Exception e) {
+    if (isClientError(e)) {
+      LOG.warn("[unified] Client error in query execution", e);
+      Metrics.getInstance().getNumericalMetric(MetricName.PPL_FAILED_REQ_COUNT_CUS).increment();
```
dai-chen, line 237: Please see how we can make this flexible because we're going to enable unified SQL in this same class soon. Same for other code if PPL-specific.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
```diff
@@ -86,6 +92,12 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient nod
     TransportPPLQueryRequest transportPPLQueryRequest =
         new TransportPPLQueryRequest(PPLQueryRequestFactory.getPPLRequest(request));
 
+    // Route to analytics engine for non-Lucene (e.g., Parquet-backed) indices
```
dai-chen, line 95: Could you double check whether we need to do this here or after the transport action below? Maybe see why we added PPL transport action previously.

core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
```diff
@@ -0,0 +1,115 @@
...
+  public void explain(PhysicalPlan plan, ResponseListener<ExplainResponse> listener) {
+    listener.onFailure(
+        new UnsupportedOperationException("Analytics engine only supports RelNode execution"));
+  }
+
+  @Override
+  public void execute(
+      RelNode plan, CalcitePlanContext context, ResponseListener<QueryResponse> listener) {
+    try {
+      Integer querySizeLimit = context.sysLimit.querySizeLimit();
```
dai-chen, line 68: Could you check how PPL V3 handle this today? I'm thinking if we want to be consistent, this needs to be part of the RelNode plan we pass to `QueryPlanExecutor` instead of post-processing? Otherwise we may pull lots of data back?

plugin/src/test/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryActionTest.java
ahkcs: FYI, PR updates have now been moved to https://github.com/opensearch-project/sql/pull/5267 due to accidental merge
Thanks for the heads up on UnifiedQueryParser. I've simplified the tests to just 2 routing behavior tests (parquet index routes to analytics, non-parquet routes to Lucene). Removed all the extractIndexName tests since that logic will be replaced.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Good catch. Updated isClientError() to match the full list from RestPPLQueryAction

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Moved all stub code into plugin/.../rest/analytics/stub/ sub-package: 
 ```
 StubIndexDetector (temporary index extraction + parquet_ prefix check),
 StubSchemaProvider (hardcoded Calcite schema)
 StubQueryPlanExecutor (canned data)
 ```
 RestUnifiedQueryAction now delegates to these.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Good call. Replaced the hand-crafted JSON with `ErrorMessageFactory.createErrorMessage(e, status.getStatus()).toString()` -- same pattern as `RestPPLQueryAction.reportError()`.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Made metrics and LangSpec query-type-aware. Now uses QueryType to select:
  - Metrics: PPL_REQ_TOTAL / PPL_FAILED_REQ_COUNT_* for PPL, REQ_TOTAL / FAILED_REQ_COUNT_* for SQL
  - Response formatting: PPL_SPEC for PPL, LangSpec.SQL_SPEC for SQL

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
ahkcs: Moved routing from RestPPLQueryAction to TransportPPLQueryAction.doExecute().
`RestPPLQueryAction` and `SQLPlugin` are reverted to their original state

core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
ahkcs: Good catch. Moved the query size limit into the RelNode plan using `LogicalSystemLimit.create()`, same pattern as `QueryService.convertToCalcitePlan()`. The limit is now part of the plan the analytics engine receives, so it can enforce it during execution instead of us pulling all rows back first.

Removed the post-processing truncation from `AnalyticsExecutionEngine`

PR #5265, opened 2026-03-25
Add cluster command for PPL

### LINKED ISSUE #5255: [RFC] PPL `cluster` Command
## Problem Statement

Users need a `cluster` command that groups log events by text similarity. The command should support parameters (`t`, `match`, `field`, etc.) while being optimized for OpenSearch's architecture and capabilities.

Key requirements:
- **Parameter compatibility**: Support expected syntax and parameters
- **Scalability**: Handle large log datasets without memory issues
- **Performance**: Reasonable response times for interactive queries
- **Correctness**: Stable, thread-safe implementation
 
## Syntax

```
cluster [field] [t=threshold] [match=mode] [labelfield=name] [countfield=name]
        [labelonly=bool] [showcount=bool] [delims=chars]
```

**Parameters:**

- **field**: Field to cluster on (required, no default field to fall back to)
- **t**: Similarity threshold between 0.0-1.0 (default: 0.8)
- **match**: Vectorization mode - `termlist` (positional), `termset` (bag-of-words), or `ngramset` (character trigrams) (default: `termlist`)
- **labelfield**: Name for cluster label output field (default: `cluster_label`)
- **countfield**: Name for cluster count output field (default: `cluster_count`)
- **labelonly**: If true, only output cluster labels, not counts (default: false)
- **showcount**: If true, include cluster counts in output (default: true)
- **delims**: Delimiter characters for tokenization (default: space)

## Algorithm Options Considered

**Greedy Single-Pass Clustering**
- Compare each document against existing cluster representatives
- Assign to first cluster above similarity threshold, or create new cluster
- Time: O(n×k), Space: O(k×d) where k=clusters, d=dimensions
- Order-dependent results (SPL results also order dependent)

**Union-Find Connected Components**
- Compare all document pairs, union those above threshold
- Find globally optimal connected components
- Time: O(n²×α(n)), Space: O(n)
- Order-independent, but much slower for large datasets

**Hierarchical Agglomerative Clustering**
- Build distance matrix, iteratively merge closest clusters
- High quality results but O(n²) space and O(n³) time
- Impractical for large log datasets

**Token-Based Vectors (termlist/termset/ngramset)**
- Simple, fast, preserves operational distinctions
- Handles exact patterns and variations well

**Pre-trained Embeddings**
- Semantic understanding, handles paraphrases
- May incorrectly cluster operationally distinct events
- Computationally expensive

### Algorithm: Greedy Single-Pass with Dynamic Representatives

**Decision**: Use Option A (Greedy Single-Pass) with token-based similarity.

**Rationale**:
- **Performance**: O(n×k) scales well for typical log datasets where k << n
- **Simplicity**: Straightforward to implement and debug
- **Memory Efficiency**: Only stores cluster representatives, not full similarity matrix
- **Lexical Focus**: Token-based similarity preserves important operational distinctions (service names, error types, etc.)

### Core Algorithm

**High-Level Flow**:
1. **Buffer Management**: Collect events in batches (like LogPatternAggFunction)
2. **Incremental Processing**: Process each batch against existing global cluster state
3. **Representative Comparison**: Compare new events against cluster representatives only
4. **Dynamic Assignment**: Join existing cluster or create new one based on threshold

**Key Design Elements**:
- **Cluster Representatives**: First event in each cluster becomes the representative
- **Global State**: Maintain cluster list across buffer flushes
- **Memory Bounds**: Limit buffer size (50K events) and max clusters (10K)
- **Vector Caching**: Cache vectorized representations to avoid recomputation

### Key Components

**1. AST Node (`core/src/main/java/org/opensearch/sql/ast/tree/Cluster.java`)**
- Store command parameters: field, threshold, match mode, output options
- Integrate with visitor pattern

**2. Clustering Engine (`common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java`)**
- Three vectorization modes: termlist (positional), termset (bag-of-words), ngramset (character trigrams)
- Cosine similarity computation using `commons-text`
- Vector caching for performance

**3. Aggregate Function (`core/src/main/java/org/opensearch/sql/calcite/udf/udaf/ClusterLabelAggFunction.java`)**
- Buffered processing following LogPatternAggFunction pattern
- Thread-safe accumulator with synchronized collections
- Incremental cluster state management across batches

**4. Grammar Integration (`ppl/src/main/antlr/OpenSearchPPLParser.g4`, `OpenSearchPPLLexer.g4`)**
- Add CLUSTER_CMD token and clusterCommand rule
- Support parameters: t, match, field, labelfield, countfield, labelonly, showcount, delims

### Integration Points

**Parser/AST**:
- Add `visitClusterCommand` to `AstBuilder.java`
- Add `visitCluster` to `AbstractNodeVisitor.java` and implementations

**Analysis**:
- Map to window function in `CalciteRelNodeVisitor.java`

**Function Registration**:
- Register ClusterLabelAggFunction in `PPLBuiltinOperators.java`
- Add to function implementation table

## Risks and Considerations

**Performance**: O(n×k) algorithm may struggle with datasets having many small clusters. Monitor cluster count growth.

**Memory**: Buffering strategy prevents unbounded growth, but large buffers still require significant memory.

**Thread Safety**: Synchronization overhead may impact performance in highly concurrent scenarios.

**Order Dependency**: Results depend on event processing order, which may surprise users expecting deterministic clustering.

### Required Field
The `field` parameter is required. There is no default _raw text field in OpenSearch to fall back to like SPL, so users must always specify which field to cluster on.

### High Cardinality Performance
The clustering algorithm is O(n×k) where k is the number of distinct clusters. When k is small relative to n, performance scales linearly — 1M events with ~15 clusters completes in ~8s. When k approaches n (most events are unique), performance degrades: 10K events with 10K clusters takes ~23s, 20K takes ~4 minutes. A max cluster cap of 10,000 is enforced to bound memory usage.

poc: https://github.com/ritvibhatt/sql/commit/8f2804ed825c72b02d7d16a934af3114a0c10088

ritvibhatt:
## Description

Adds the cluster command to PPL, which groups documents into clusters based on text similarity. More info on decisions made in #5255 

### Syntax
```
cluster <field> [t=<threshold>] [match=<algorithm>] [labelfield=<field>] [countfield=<field>] [showcount=<bool>] [labelonly=<bool>] [delims=<string>]
```

### Supported algorithms
- termlist (default) — positional term frequency, word order matters
- termset — bag-of-words, word order ignored
- ngramset — character trigram frequency for fuzzy matching

All algorithms use cosine similarity against the configured threshold to determine cluster membership.

### Behavior
- By default (labelonly=false), deduplicates results to one
representative row per cluster
- labelonly=true keeps all rows and adds the cluster label
- showcount=true adds a count of how many documents belong to each cluster
- Rows where the source field is null are filtered out before clustering

### Changes
- CalciteRelNodeVisitor: implements visitCluster with window-based clustering, null filtering, and optional dedup/showcount
- ClusterLabelAggFunction: window aggregate that buffers rows, runs greedy clustering, and returns an array of cluster labels
- TextSimilarityClustering: cosine similarity computation with LRU-cached vectorization for termlist, termset, and ngramset modes
- Cluster AST node, parser rules, and PPL grammar additions
- Documentation with 7 doctest-verified examples
- Unit tests for logical plan verification
- Integration tests covering all parameters, dedup behavior, labelonly mode, and multi-row clustering

### Related Issues
Resolves #5255 

### Check List


docs/user/ppl/cmd/cluster.md
```diff
@@ -0,0 +1,211 @@
...
+| Parameter | Required/Optional | Description |
+| --- | --- | --- |
+| `<field>` | Required | The text field to use for clustering analysis. |
+| `t` | Optional | Similarity threshold between 0.0 and 1.0. Documents with similarity above this threshold are grouped together. Default is `0.8`. |
+| `match` | Optional | Clustering algorithm to use. Valid values are `termlist`, `termset`, `ngramset`. Default is `termlist`. |
+| `labelfield` | Optional | Name of the field to store the cluster label. Default is `cluster_label`. |
+| `countfield` | Optional | Name of the field to store the cluster size. Default is `cluster_count`. |
+| `showcount` | Optional | Whether to include the cluster count field in the output. Default is `false`. |
+| `labelonly` | Optional | When `true`, keeps all rows and only adds the cluster label. When `false` (default), deduplicates by keeping only the first representative row per cluster. Default is `false`. |
+| `delims` | Optional | Delimiter characters used for tokenization. Default is `non-alphanumeric` (splits on any non-alphanumeric character). |
```
penghuo: what is allowed value of `delims`?

common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java
```diff
@@ -0,0 +1,169 @@
...
+
+  private static double validateThreshold(double threshold) {
+    if (threshold <= 0.0 || threshold >= 1.0) {
+      throw new IllegalArgumentException(
+          "The threshold must be > 0.0 and < 1.0, got: " + threshold);
+    }
+    return threshold;
+  }
+
+  private static String validateMatchMode(String matchMode) {
```
Swiddis: ideally this should be enum instead of str, which makes it hard to misuse the interface

docs/user/ppl/cmd/cluster.md
Swiddis: if minor: this probably also could stand to be tagged somehow instead of just working directly on strings

Swiddis (commented):
merge conflicts

common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * <p>Optimized for incremental processing with vector caching and memory-efficient operations.
+ */
+public class TextSimilarityClustering {
+
+  private static final CosineSimilarity COSINE = new CosineSimilarity();
+
+  // Cache vectorized representations to avoid recomputation
+  private final Map<String, Map<CharSequence, Integer>> vectorCache =
+      new LinkedHashMap<>(MAX_CACHE_SIZE, 0.75f, true) {
```
penghuo: initialCapacity equal to MAX_CACHE_SIZE?

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/ClusterLabelAggFunction.java
```diff
@@ -0,0 +1,221 @@
...
+ * batches to avoid memory issues.
+ *
+ * <p>When used as a window function over an unbounded frame, the result is a List where each
+ * element corresponds to the cluster label for that row position.
+ */
+public class ClusterLabelAggFunction
+    implements UserDefinedAggFunction<ClusterLabelAggFunction.Acc> {
+
+  private int bufferLimit = 50000; // Configurable buffer size
+  private int maxClusters = 10000; // Limit cluster count to prevent memory explosion
```
penghuo: Does this setting configurable by user?

penghuo (commented):
Did u benchmark cluster command on http_logs dataset?

noCharger:
Benchmarked the `cluster` command on a 3-node cluster (m5.xlarge, 8GB heap) built from
this PR at `a51e8f95f`. I used a controlled-K synthetic log corpus (it matches the command's
`normalizeToken` tokenization, so each template is exactly one `termlist` cluster). More
importantly, I compared it against three baselines on the **identical data in the same run**
so the numbers are interpretable: a plain scan (`| head N | stats count()`), the O(n)
`patterns` sibling, and pushed-down `stats count() by <field>`. Single-query, serial,
p50/p90 of 5 iterations.

cluster_bench_charts


**Panel A: the command is read-bound, not compute-bound, at low and moderate k.** The four
lines overlap. At k=15, `cluster` costs about the same as a plain scan across 1K to 1M rows
(within ~10%, and at 1M it is even marginally faster than scan and `stats_by`). When k is
small, the greedy O(n·k) loop adds essentially nothing on top of pulling rows to the
coordinator. This is expected, since execution is a coordinator-side scan plus an in-JVM
window aggregate with no pushdown.

| n | scan | stats_by | patterns | **cluster** |
|---|------|----------|----------|-------------|
| 1K  | 0.06s | 0.08s | 0.05s | 0.07s |
| 10K | 0.21s | 0.28s | 0.20s | 0.18s |
| 100K| 0.96s | 1.04s | 1.01s | 1.05s |
| 1M  | 9.63s | 9.38s | 9.74s | **8.75s** |

**Panel B: cluster count `k` is the only real cost regime.** At n=100K, latency is flat
(~0.6s) through k=1000, but **k=10,000 did not complete within 30 minutes** (~10⁹ cosine
comparisons). This is the expected high-cardinality degradation, and it is exactly what the
`maxClusters=10,000` cap guards against. Worth documenting that clustering a
near-unique-value field is unsupported.

**Panel C: match modes** (n=50K, k=100): `termlist`, `termset`, and `ngramset` are all
~0.38s. `ngramset` merged the 100 seeded templates into ~18 clusters at t=0.8 because
character-trigram similarity is fuzzier, which is expected and not a correctness issue.

Net: `cluster` is scan-cheap and practical for typical log volumes and cardinalities, and the
guardrail on `k` is the important documented limit.

docs/user/ppl/cmd/cluster.md
noCharger: `delims` accepts one of two forms: the keyword `non-alphanumeric` (the default, which splits on any character that is not a letter, digit, or underscore), or a string whose individual characters are each treated as a delimiter. For example `delims=" ,;"` splits on space, comma, and semicolon. I have updated the parameter table in cluster.md to spell this out.

common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java
noCharger: Done. Added a `MatchMode` enum (`TERMLIST`/`TERMSET`/`NGRAMSET`) with `MatchMode.fromString`. `TextSimilarityClustering` now takes the enum in its constructor and dispatches vectorization on it, so the mode can no longer be misused as a free-form string, and validation lives in one place.

docs/user/ppl/cmd/cluster.md
noCharger: I documented the two accepted forms of `delims` (the `non-alphanumeric` keyword or an arbitrary set of delimiter characters). Since the second form is genuinely an open-ended character set rather than a closed vocabulary, I kept it as a string but clarified the contract in the doc. If you would prefer a stricter typed form (for example an explicit enum of common delimiter presets), I am happy to add it.

common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java
noCharger: Good catch. Setting the initial capacity equal to `MAX_CACHE_SIZE` under the default 0.75 load factor would rehash before reaching the cap. I changed it to `(int)(MAX_CACHE_SIZE / 0.75f) + 1` so the LRU cache can hold `MAX_CACHE_SIZE` entries without rehashing, and moved the `MAX_CACHE_SIZE` constant above the field so the dependency reads top to bottom.

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/ClusterLabelAggFunction.java
noCharger: Both are now operator-configurable cluster settings, mirroring `plugins.ppl.pattern.buffer.limit`:

- `plugins.ppl.cluster.buffer.limit` (default 50000, NodeScope, dynamic) bounds how many rows are buffered on the coordinator before an incremental clustering pass. It affects transient memory only and does not change the clustering result.
- `plugins.ppl.cluster.max.clusters` (default 10000, NodeScope, dynamic) caps the number of distinct clusters, which bounds the per-row comparison cost that keeps the greedy O(rows*clusters) pass from degrading on near-unique-value fields. Note this one does change the output, since it caps how many distinct labels are possible, so raising it can make high-cardinality queries slow; I documented that in the Performance section of cluster.md.

I kept them as NodeScope operator settings rather than per-query options because they are administrative guardrails, not query semantics. I also moved these (and threshold/match/delims) off the shared function instance and onto the per-query accumulator, so a reused function object cannot race across concurrent aggregations. Long-running queries remain bounded by the existing `plugins.ppl.query.timeout` (default 300s), so no command-specific timeout was added.

docs/user/ppl/index.md
```diff
@@ -55,6 +55,7 @@ source=accounts
 | [bin command](cmd/bin.md) | 3.3 | experimental (since 3.3) | Group numeric values into buckets of equal intervals. |
 | [timechart command](cmd/timechart.md) | 3.3 | experimental (since 3.3) | Create time-based charts and visualizations. |
 | [chart command](cmd/chart.md) | 3.4 | experimental (since 3.4) | Apply statistical aggregations to search results and group the data for visualizations. |
+| [cluster command](cmd/cluster.md) | 3.7 | experimental (since 3.7) | Group documents into clusters based on text similarity using various clustering algorithms. |
```
penghuo, line 59: change 3.7 -> 3.8.

docs/user/ppl/cmd/cluster.md
```diff
@@ -0,0 +1,229 @@
...
+| `t` | Optional | Similarity threshold between 0.0 and 1.0. Documents with similarity above this threshold are grouped together. Default is `0.8`. |
+| `match` | Optional | Clustering algorithm to use. Valid values are `termlist`, `termset`, `ngramset`. Default is `termlist`. |
+| `labelfield` | Optional | Name of the field to store the cluster label. Default is `cluster_label`. |
+| `countfield` | Optional | Name of the field to store the cluster size. Default is `cluster_count`. |
+| `showcount` | Optional | Whether to include the cluster count field in the output. Default is `false`. |
+| `labelonly` | Optional | When `true`, keeps all rows and only adds the cluster label. When `false` (default), deduplicates by keeping only the first representative row per cluster. Default is `false`. |
+| `delims` | Optional | Characters used to split the field into tokens. Accepts either the keyword `non-alphanumeric` (the default, which splits on any character that is not a letter, digit, or underscore) or a string whose individual characters are each treated as a delimiter. For example, `delims=" ,;"` splits on space, comma, and semicolon. |
```
penghuo, line 26: It is better to follow same pattern as other paramaters. At the end, add "Default is non-alphanumeric"

docs/user/ppl/cmd/cluster.md
```diff
@@ -0,0 +1,229 @@
...
+
+
+## Cluster settings
+
+The `cluster` command exposes two cluster-level guardrails. Both are `NodeScope` and dynamic, so a cluster administrator can update them at runtime with the `_cluster/settings` API. They are administrative controls rather than per-query options, and they are shared by every `cluster` query on the node.
+
+| Setting | Default | Minimum | Description |
+| --- | --- | --- | --- |
+| `plugins.ppl.cluster.buffer.limit` | `50000` | `1` | Number of rows buffered on the coordinating node before an incremental clustering pass runs. This bounds transient memory only and does not change the clustering result. Lower it on memory constrained nodes, raise it to reduce the number of incremental passes. |
+| `plugins.ppl.cluster.max.clusters` | `10000` | `1` | Maximum number of distinct clusters the command will create. Once this many clusters exist, every remaining row is folded into its closest existing cluster instead of forming a new one. This value changes the output because it caps how many distinct `cluster_label` values are possible. |
```
penghuo, line 221: Consider default value as 100/1000. As [benchmark results](https://github.com/opensearch-project/sql/pull/5265#issuecomment-5006878043) indicated, when clusters=10000, the query will took 30mins+.

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/ClusterLabelAggFunction.java
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * <p>When used as a window function over an unbounded frame, the result is a List where each
+ * element corresponds to the cluster label for that row position.
+ */
+public class ClusterLabelAggFunction
+    implements UserDefinedAggFunction<ClusterLabelAggFunction.Acc> {
+
+  private static final int DEFAULT_BUFFER_LIMIT = 50000; // rows buffered before a partial merge
+  // caps distinct clusters, which bounds the per-row comparison cost
+  private static final int DEFAULT_MAX_CLUSTERS = 10000;
```
penghuo, line 28: OpenSearchSettings.java should share the same default value.

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/ClusterLabelAggFunction.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    public void clearBuffer() {
+      buffer.clear();
+    }
+
+    @Override
+    public Object value(Object... argList) {
+      partialMerge();
+      clearBuffer();
```
penghuo: clearBuffer only been called when get value? What if exception throw before caller execute value method?

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/ClusterLabelAggFunction.java
```diff
@@ -0,0 +1,245 @@
...
+          ClusterRepresentative newCluster = new ClusterRepresentative(1, value, 1);
+          globalClusters.add(newCluster);
+          nextClusterId = 2;
+          return new ClusterAssignment(1, 1);
+        }
+      }
+    }
+
+    public void clearBuffer() {
+      buffer.clear();
```
penghuo, line 240: clear globalClusters and allLabels also

common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java
```diff
@@ -0,0 +1,156 @@
...
+      java.util.regex.Pattern.compile("^\\d+$");
+
+  private static String normalizeToken(String token) {
+    return NUMERIC_PATTERN.matcher(token).matches() ? "*" : token;
+  }
+
+  /** Positional term frequency — token order matters. */
+  private Map<CharSequence, Integer> vectorizeTermList(String value) {
+    String[] tokens = tokenize(value);
+    Map<CharSequence, Integer> vector = new HashMap<>((int) (tokens.length * 1.4));
```
penghuo: 1.4 is magic number?

common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java
```diff
@@ -0,0 +1,156 @@
...
+
+  private static final java.util.regex.Pattern NUMERIC_PATTERN =
+      java.util.regex.Pattern.compile("^\\d+$");
+
+  private static String normalizeToken(String token) {
+    return NUMERIC_PATTERN.matcher(token).matches() ? "*" : token;
+  }
+
+  /** Positional term frequency — token order matters. */
+  private Map<CharSequence, Integer> vectorizeTermList(String value) {
```
penghuo, line 104: Clustering fetches raw _source and re-tokenizes/re-vectorizes on the coordinating node, discarding OpenSearch's index-time analysis.

docs/user/ppl/cmd/cluster.md
noCharger: > Consider default value as 100/1000. As [benchmark results](https://github.com/opensearch-project/sql/pull/5265#issuecomment-5006878043) indicated, when clusters=10000, the query will took 30mins+.

lowered max.clusters default to 1000

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/ClusterLabelAggFunction.java
noCharger: > clearBuffer only been called when get value? What if exception throw before caller execute value method?

Done. value() now drains inside a try/finally.

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/ClusterLabelAggFunction.java
noCharger: > clear globalClusters and allLabels also

Done. reset() now clears globalClusters and allLabels in addition to buffer and nextClusterId, and it runs in the finally block of value().

common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java
noCharger: > 1.4 is magic number?

Replaced the manual ×1.4 capacity math with HashMap.newHashMap(keys.size()) (JDK 19+), which sizes the map to hold that many entries without rehashing.

common/src/main/java/org/opensearch/sql/common/cluster/TextSimilarityClustering.java
noCharger: > Clustering fetches raw _source and re-tokenizes/re-vectorizes on the coordinating node, discarding OpenSearch's index-time analysis.

The coordinator-side re-tokenization is a deliberate choice already weighed in the RFC ([#5255](https://github.com/opensearch-project/sql/issues/5255), see "Other Approaches Considered"): the index-time-analysis alternatives (fingerprint analyzer + terms agg, min_hash + terms agg) were both rejected because they require index-time setup and cannot handle eval-derived fields like ... | eval combined = concat(a, b) | cluster combined.

Issue #5261, opened 2026-03-24
[META] Error Handling Enhancements


Swiddis:
This is a meta issue collecting a few different ideas from different sources.

The core problem is that misbehaving queries are typically hard to debug & require a lot of knowledge of PPL/SQL engine behavior, and there's room for enhancement.

Discussing the concept with @anasalkouz, there are three main problem classes we want to enhance:

1. Something went wrong (execution error)
2. 0 results are returned, why?
3. The query is slow, why?

So, compiling existing issues for the backend:
- https://github.com/opensearch-project/sql/issues/4919 is the backend for the first one, it introduces a general reporting interface that can integrate with our current exceptions and add context. We can incrementally drill into error classes over time.
- https://github.com/opensearch-project/sql/issues/4343 can give the foundation for the latter 2, by providing analyze metrics for results returned & timings at all stages.

Once those have reasonably complete implementations, errors will be in structured reports like this:

```json
{
  "status": 400,
  "error": {
    "type": "SemanticCheckException",
    "code": "FIELD_NOT_FOUND",
    "reason": "Invalid Query",
    "details": "Failed to resolve field 'foo'",
    "location": [
      "while planning the query",
      "while resolving fields in the index mapping"
    ],
    "context": {
      "index_pattern": "logs-*",
      "position": {"line": 1, "column": 25},
      "query": "source=logs-* | fields foo",
      "query_id": "b6627794-3939-4ac4-8c5b-821ccc400f4f"
    },
    "suggestion": "Did you mean: 'foobar'"
  }
}
```

Frontends can choose to do whatever they want with this: render specific fixed pieces of context (e.g. `position` is pointing to a spot in the query, highlight it?), render details/locations/suggestions, throw it in an LLM, etc. It also helps with oncall debugging when given these responses (either directly or from har files). https://github.com/opensearch-project/sql/issues/4919#issuecomment-3863005661 shows me doing this quickly for the SQL CLI based on a proof-of-concept implementation.

From the frontend, there's a separate meta issue:
- https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11577 is the catch-all for integrating with the reporting interface in the frontend.
- Explain-analyze --> debug will be a call to action when a query runs either over some threshold (1000ms?) or with 0 results.

Once the core flows here are done, what's left is to start vetting specific error cases:
- #4771 
- #4872 
- #4869 
- #4896 
- #5065 
- Field not found (issue pending)
[...]

PR #5266, opened 2026-03-25
Initial implementation of report-builder interface

### LINKED ISSUE #4919: [RFC] Error Report-building Mechanism for PPL Errors
## Problem Statement

End-users currently receive error messages with minimal context about how errors occurred, making it difficult to diagnose and fix issues. When exceptions propagate through multiple layers of the system (parsing → semantic analysis → execution → REST response), contextual information is lost, leaving users with generic error messages that don't provide actionable guidance.

**Examples of current pain points:**

- #4872: Error shows `Failed to read mapping for index pattern [[Ljava.lang.String;@9acdf8]` instead of explaining which indices have incompatible mappings
- #4869: Error shows "Insufficient resources to run the query" without indicating which resource (memory/CPU), current values, or how to fix it
- #4896: Error is `ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0` with no context on what type of array might have been involved. In this case, we removed the error condition entirely, but debugging took several hours from multiple engineers

## Current State

The OpenSearch SQL/PPL plugin uses a simple exception hierarchy where exceptions carry only:

1. Exception type (e.g., `SemanticCheckException`)
2. A single error message string
3. Optional cause chain (for wrapped exceptions)

**Architecture limitations:**

- **No context accumulation**: As errors bubble up through layers (Parser → Analyzer → Executor → REST), intermediate layers cannot add context without modifying the original exception message
- **Single-point error creation**: All context must be known at the point where the exception is first thrown
- **Loss of structured information**: Important details like query text, table names, field positions, datasource IDs, and execution context are unavailable to error handlers
- **Generic error responses**: The `ErrorMessage` class can only extract `type`, `reason`, and `details` from the exception's localized message

**Example error flow:**

```
[Semantic Analyzer] throw new SemanticCheckException("Field 'x' not found")
  ↓
[Query Service] catch + rethrow (no context added)
  ↓
[REST Handler] ErrorMessage.toString() → {"type": "SemanticCheckException", "details": "Field 'x' not found"}
  ↓
[User] Receives minimal context, must guess which table, datasource, or query caused the issue
```

## **Long-Term Goals**

- **Rich error context chains**: Users should see the full story of how an error occurred, with context added at each layer:

  ```
  Error: Field 'timestamp' not found
  -> in table 'logs' (index pattern: 'logs-*')
  -> while analyzing query: "source=logs-* | fields timestamp, message | where status > 500"
  -> at position: line 1, column 25
  -> datasource: 'default-opensearch'
  ```

- **Actionable error messages**: Each error should guide users toward resolution with specific details about what went wrong and how to fix it

- **Structured error information**: Enable programmatic error handling with machine-readable error codes and structured context fields

- **Consistent error experience**: Unified error handling across SQL, PPL, Calcite, and legacy engines

**How confident are we that this solution solves the problem?**

High confidence. Similar mechanisms (Rust's anyhow/eyre, Java's Spring NestedExceptionUtils, Python's exception chaining) have proven effective in providing rich error context in existing projects. The OpenSearch SQL plugin's multi-layered architecture works well for context accumulation at each layer.

## Proposal

Introduce a **Report-building error mechanism** inspired by Rust's anyhow/eyre libraries that:

1. **Wraps exceptions** with contextual information as they propagate through system layers
2. **Accumulates context** without modifying original exception messages
3. **Formats rich error responses** with context chains for end-users
4. **Maintains backwards compatibility** with existing exception types and error handlers

**Core components:**

```java
// Report wrapper that accumulates context
public class ErrorReport {
    private final Throwable rootCause;
    private final List<ErrorContext> contextChain;
    
    // Idempotent wrapper: determines whether `e` contains a Report and either returns that report or starts a new one.
    public static ErrorReport wrap(Throwable e);

    public ErrorReport context(String key, Object value);
    public ErrorReport context(ErrorContext ctx);
    public String toDetailedMessage();
    public JSONObject toJSON();
}

// Context information at each layer
public class ErrorContext {
    private final String layer;  // "Parser", "Analyzer", "Executor". Can come from a constants file for consistency.
    private final Map<String, Object> attributes;  // query, table, position, datasource, etc.
}

// Enhanced error response
{
  "status": 400,
  "error": {
    "type": "SemanticCheckException",
    "code": "FIELD_NOT_FOUND",
    "reason": "Invalid Query",
    "details": "Field 'timestamp' not found in table 'logs'",
    "context": [
      {"layer": "semantic_analysis", "table": "logs", "index_pattern": "logs-*"},
      {"layer": "query_parser", "query": "source=logs-* | fields timestamp", "position": {"line": 1, "column": 25}},
      {"layer": "datasource", "datasource_name": "default-opensearch"}
    ],
    "suggestion": "Check that field 'timestamp' exists in indices matching 'logs-*' pattern"
  }
}
```

This structured output approach makes it straightforward for the front-end to enrich the error presentation, e.g. highlighting the position of the error.

### Phase 1: Core Infrastructure (Non-breaking)

1. **Create Report Wrapper Classes**
   - `ErrorReport`: Wrapper that accumulates context around any Throwable
   - `ErrorContext`: Structured context at each layer (query, position, datasource, table, etc.)
   - `ErrorCode`: Enum of machine-readable error codes (FIELD_NOT_FOUND, SYNTAX_ERROR, INDEX_NOT_FOUND, etc.)

2. **Enhance Exception Base Classes**
   - Add optional `ErrorReport` field to `QueryEngineException` base class
   - Add builder methods: `.withContext(key, value)`, `.withQuery(query)`, `.withPosition(line, col)`
   - Maintain backwards compatibility: existing constructors work unchanged

3. **Update ErrorMessage Formatting**
   - Enhance `ErrorMessage.java` to extract and format `ErrorReport` if present
   - Fall back to current behavior for exceptions without reports

### Phase 2: Incremental Context Adoption

4. **Add Context at Key Layers** (can be done incrementally per component):
   - **Parsing layer** (`PPLSyntaxParser`, `SQLSyntaxParser`): Add query text, position info
   - **Semantic analysis** (`Analyzer`): Add table names, field names, datasource context
   - **Query execution** (`QueryService`): Add execution engine (Calcite/V2/Legacy), memory usage
   - **REST handlers** (`RestPPLQueryAction`, `RestSQLQueryAction`): Add request ID, user context

5. **Enhance Specific Error Types** (prioritize by user impact):
   - Start with errors from GH issues #4872, #4869
   - `SemanticCheckException`: Add table, field, datasource context
   - `MemoryUsageException`: Add current/max memory, settings to adjust
   - Index pattern errors: Add which indices matched, mapping differences

## Implementation details

```java
// Example: Adding context in semantic analyzer
public class Analyzer {
    public LogicalPlan analyze(UnresolvedPlan plan, AnalysisContext context) {
        try {
            return doAnalyze(plan, context);
        } catch (SemanticCheckException e) {
            throw ErrorReport.wrap(e)
                    .withContext("table", context.getTableName())
                    .withContext("datasource", context.getDatasourceName())
                    .withContext("index_pattern", context.getIndexPattern());
        }
    }
}

// Example: Adding context at REST layer
public class RestPPLQueryAction {
    private void executeQuery(String query, ResponseListener listener) {
        try {
            pplService.execute(query, listener);
        } catch (Exception e) {
            ErrorReport report = ErrorReport.wrap(e)
                .withContext("query", query)
                .withContext("request_id", QueryContext.current().getRequestId())
                .withContext("format", format);

            listener.onFailure(report.toException());
        }
    }
}

// Example: Enhanced error response
public class ErrorMessage {
    protected String fetchDetails() {
        if (exception instanceof QueryEngineException qee && qee.hasReport()) {
            return qee.getReport().toDetailedMessage();
        }
        return exception.getLocalizedMessage();
    }

    private JSONObject getErrorAsJson() {
        JSONObject errorJson = new JSONObject();
        errorJson.put("type", type);
        errorJson.put("reason", reason);
        errorJson.put("details", details);

        // Add context chain if available
        if (exception instanceof QueryEngineException qee && qee.hasReport()) {
            errorJson.put("code", qee.getReport().getErrorCode());
            errorJson.put("context", qee.getReport().getContextChain());
            errorJson.put("suggestion", qee.getReport().getSuggestion());
        }

        return errorJson;
    }
}
```

### 1. Structured Exception Messages (Simpler approach)

Instead of wrapping exceptions, enhance exception constructors to accept structured context:

```java
public class SemanticCheckException extends QueryEngineException {
    public SemanticCheckException(String message,
                                  String table,
                                  String field,
                                  String datasource) {
        super(formatMessage(message, table, field, datasource));
    }
}
```

**Pros:**

- Simpler implementation, less infrastructure
- Backwards compatible if old constructors remain

**Cons:**

- Context must be known at exception creation time
- Each exception class needs custom constructor for its context
- Intermediate layers can't add context without catching/rethrowing with new exception
- Harder to extract context programmatically (must parse message string)

### 2. ThreadLocal Context (MDC-style)

Use ThreadLocal storage to accumulate context throughout request lifecycle:

```java
public class ErrorContext {
    private static final ThreadLocal<Map<String, Object>> context = new ThreadLocal<>();

    public static void addContext(String key, Object value) {
        context.get().put(key, value);
    }
}
```

**Pros:**

- No changes to exception hierarchy
- Context automatically available to all error handlers
- Commonly used pattern (SLF4J MDC)

**Cons:**

- Risk of context leaking between requests if not properly cleared
- Harder to track which context is added where
- Context not attached to specific exception, harder to serialize/deserialize
- Complex in async execution contexts

### 3. Status Quo with Better Messages (Minimal change)

Keep current architecture but improve individual error messages:

- Add more details to existing exception messages
- Better formatting in `ErrorMessage.fetchDetails()`
- Fix specific issues like #4872, #4869 individually

**Pros:**

- No architectural changes
- Can be done incrementally

**Cons:**

- Doesn't solve systematic context loss problem
- Still limited to single error message string
- No structured error information for programmatic handling
- Difficult to add context at intermediate layers

Swiddis:
### Description

Well, this is turning out quite long.

The fundamental idea of this PR is to add a lot more context to error messages. For example, take this bad query:
```kql
source = big5
| eval range_bucket = case(
   `metrics.size` < -10, 'range_1',
   `metrics.size` >= -10 and `metrics.size` < 10, 'range_2',
   `metrics.size` >= 10 and `metrics.size` < 100, 'range_3',
   `metrics.size` >= 100 and `metrics.size` < 1000, 'range_4',
   `metrics.size` >= 1000 and `metrics.size` < 2000, 'range_5',
   `metrics.size` >= 2000, 'range_6')
| stats count() by range_bucket, span(`@timestamp`, 1h) as auto_span
| sort + range_bucket, + auto_spa
```

Now there's more context, provided at various layers:
```json
{
  "error": {
    "context": {
      "stage": "analyzing",
      "stage_description": "Parsing and validating the query"
    },
    "details": "Field [nonex] not found.",
    "location": [
      "while preparing and validating the query plan"
    ],
    "code": "FIELD_NOT_FOUND",
  },
  "status": 400
}
```

Or, the humble 'index not found':
```json
{
  "error": {
    "context": {
      "stage": "analyzing",
      "index_name": "big5_nonex",
      "stage_description": "Parsing and validating the query"
    },
    "details": "no such index [big5_nonex]",
    "location": [
      "while preparing and validating the query plan",
      "while fetching index mappings"
    ],
    "code": "INDEX_NOT_FOUND",
    "type": "IndexNotFoundException"
  },
  "status": 404
}
```

The doctest deltas have a few concrete cases where I went through and did light touch-ups. Improving any particular error takes O(2 minutes) to wrap in a report with context, and it'll integrate with any higher-up wrapping to give the completed report at the end.

A lot of this PR in terms of line count is just patching specific places where tests are asserting errors behave in specific ways. The core of the PR is:
- `ErrorReport` and `StageErrorHandler`, which adds utilities for incrementally building detailed errors
- `QueryService` where we hook up the outer shell of the stage handlers, and
- `RestPPLQueryAction` where we handle and present the errors.

Simple tests are in `error_reports.yml`, as well as doctest and various other scattered tests. (There's also lots of tests that have no diff, but did catch bugs in earlier versions of this PR.)

The enhancements aren't super complete in this PR as the emphasis is scaffolding (and the integration work + fixing tests is already quite significant), for the most part the changes are backwards compatible in normal cases. The `type`, `details`, and `status` fields are in the same spots and by default pull from the underlying exception. I'll be adding more thorough probes for specific error cases in followups.

### Related Issues
Meta: #5261
Closes #4919 

### Check List


Swiddis:
> The error enrichment logic extracts the raw SQL query plan string from internal exception messages and...

In context, this is pulling details out of a potentially very fuzzy Calcite error. The error logic without any of this ultimately returns the plan with a message resembling "Failed to prepare plan: [blob]." The plan is provided as context as the innermost error might be weird and only apply to one node, e.g. `CalciteEnumerableNestedAggregate` having a problem means nothing without access to the plan containing said aggregate. In OSD this plan probably won't be presented to users directly.

Any details about our internal implementation can be easily scrutinized by navigating to https://github.com/opensearch-project/sql in a recent version of Netscape Navigator.

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -398,6 +401,54 @@ protected RelFieldTrimmer newFieldTrimmer() {
...
+          && message.contains("CalciteEnumerableNestedAggregate");
+    }
+
+    // Detect if error is due to window functions in unsupported context (bins on time fields)
+    private static boolean isWindowBinOnTimeField(SQLException e) {
+      String errorMsg = e.getMessage();
+      return errorMsg != null
+          && errorMsg.contains("Error while preparing plan")
+          && errorMsg.contains("WIDTH_BUCKET");
+    }
```
Swiddis, line 414: self-suggestion: These conditions can be simplified by using error codes at a lower level

The checks here just become `report.getCode() == ErrorCode.ILLEGAL_AGGREGATE_PUSHDOWN` and `report.getCode() == ErrorCode.INVALID_WINDOW_BIN` or something, and we can remove the helpers. Can add this if it comes up or is in the way of other changes.

I mostly bring it up to highlight why I'm introducing codes in the first place: They can let us specify much more specific scenarios than generic exceptions do. I just didn't fully integrate it here in the interest of time & keeping the PR as small as I could get it.

anasalkouz:
Not sure if understand the improvements: what is the error experience before and after this PR?

Swiddis:
You can see some concrete examples by looking at the diffs for doctests that check for errors https://github.com/opensearch-project/sql/pull/5266/changes#diff-601c336135ceed5ca58ff75bf3783afd2722610e3b5583160476fc9378b34cd6

dai-chen:
In meta issue, we mention the "The core problem is that misbehaving queries are typically hard to debug". So the improvements like `location` are mostly for developer? Probably need some clarification for the goal.

Swiddis:
> So the improvements like location are mostly for developer?

It effectively plays the role of a stack trace, but we control the content and allow it to be closer to non-dev-readable. So not as much noise is shared, it helps developers to localize where in the query process an issue happened, and it helps users/debugging agents to figure out what stage things are breaking at as well. Not all errors make it obvious if it happened at planning or execution, for instance.

> Probably need some clarification for the goal.

In general the goal is to make it easy to add arbitrary context to errors at any layer of the code.

Generic exceptions don't have great support for the notion of "at step A, we tried to do B, which tried to do C, which failed. We need to provide context on all of A, B, and C." Exception wrapping is the main way to get there but unwrapping at the serialization layer gets messy, and the messages aren't really structured. And as hinted with location, it's hard to read stack traces if you don't have dev training. Intuitively this can be thought of as moving from unstructured narrow-event logging to structured wide-event logging, but for exceptions.

This is a framework, so we can add any information to any error case, and frontends can flexibly handle what they need to. The errors here are somewhat minimal since the focus is the framework itself. For example, one easy context bit to look at adding in the future is a cursor position for syntax errors, which the frontend can then render as an error which says "the problem is _here_." (Shoutout to [Chumsky's errors](https://docs.rs/chumsky/latest/chumsky/index.html) which spoiled me in this regard.) Context that's useful for users we can add to the frontend, context that's only useful for devs we can leave in the scope of HAR files or debugging agents.

opensearch/src/main/java/org/opensearch/sql/opensearch/client/OpenSearchNodeClient.java
```diff
@@ -99,7 +101,11 @@ public Map<String, IndexMapping> getIndexMappings(String... indexExpression) {
                   Map.Entry::getKey, cursor -> new IndexMapping(cursor.getValue())));
     } catch (IndexNotFoundException | OpenSearchSecurityException e) {
       // Re-throw directly to be treated as client error finally
-      throw e;
+      throw ErrorReport.wrap(e)
```
ahkcs, line 104: This catch block handles both `IndexNotFoundException` and `OpenSearchSecurityException`, but both get `ErrorCode.INDEX_NOT_FOUND`. A user who lacks index permissions will see code: `INDEX_NOT_FOUND` instead of something like `PERMISSION_DENIED`

ahkcs (dismissed):

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
```diff
@@ -49,17 +46,47 @@ public RestPPLQueryAction() {
...
+
+  private static int getRawErrorCode(Exception ex) {
+    if (ex instanceof ErrorReport) {
+      return getRawErrorCode(((ErrorReport) ex).getCause());
+    }
+    if (ex instanceof OpenSearchException) {
+      return ((OpenSearchException) ex).status().getStatus();
+    }
+    // Possible future work: We currently do this on exception types, when we have more robust
+    // ErrorCodes in more locations it may be worth switching this to be based on those instead.
```
RyanL1997, line 69: `SemanticCheckException` and `ExpressionEvaluationException` were removed from `isClientError()`. During this transitional period, if these exceptions aren't always wrapped in `ErrorReport`, they'll fall through to the `return 500` default and return a 500 instead of the previous 400. Was this removal intentional, or should they stay in the list until all call sites are guaranteed to wrap them?

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -429,18 +480,15 @@ public RelNode visit(TableScan scan) {
         optimizeTime.set(System.nanoTime() - startTime);
         return preparedStatement;
       } catch (SQLException e) {
+        Throwables.throwIfUnchecked(e);
```
RyanL1997: `Throwables.throwIfUnchecked(e)` is a no-op here — `SQLException` is a checked exception, so this will never throw. Was this meant to do something else (e.g. unwrap the cause), or can it be removed?

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -398,6 +401,54 @@ protected RelFieldTrimmer newFieldTrimmer() {
...
+      if (e.getCause() != null) {
+        rc = rootCauseMessage(e.getCause());
+      }
+      for (int i = 0; rc == null && i < e.getSuppressed().length; i++) {
+        rc = rootCauseMessage(e.getSuppressed()[i]);
+      }
+      return rc != null ? rc : e.getMessage();
+    }
+
+    private static void enrichErrorsForSpecialCases(ErrorReport.Builder report, SQLException e) {
```
RyanL1997, line 429: This traverses suppressed exceptions as well as the cause chain. Is it intentional that the "root cause" message could come from a suppressed exception rather than the main cause chain?

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -398,6 +401,54 @@ protected RelFieldTrimmer newFieldTrimmer() {
...
+                    + " (2) the timestamp field to be used as an aggregation bucket (e.g., 'stats"
+                    + " count() by @timestamp').")
+            .code(ErrorCode.UNSUPPORTED_OPERATION)
+            .context("is_window_bin_on_time_field", true)
+            .suggestion("check pushdown is enabled and review the aggregation");
+      }
+    }
+
     /**
      * Runs a relational expression by existing connection. This class copied from {@link
```
RyanL1997, line 452: Nit: the two `if` blocks aren't mutually exclusive. The first one sets `details` to `rootCauseMessage(e)`, and the second one (`isWindowBinOnTimeField`) can overwrite `details` again. It works today because WIDTH_BUCKET errors always match the first condition too, but an `else if` (or early return) would make the intent clearer and avoid surprises if conditions change later.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/error/ErrorMessage.java
```diff
@@ -62,12 +63,14 @@ public String toString() {
   }
 
   private JSONObject getErrorAsJson() {
-    JSONObject errorJson = new JSONObject();
+    if (exception instanceof ErrorReport) {
+      return new JSONObject(((ErrorReport) exception).toJsonMap());
+    }
```
RyanL1997: Question:

When the exception is an `ErrorReport`, `toJsonMap()` returns a map with no `reason` field — only `type`, `code`, `details`, `location`, `context`, and `suggestion`. Existing clients that parse `error.reason` will get `null`. This is a user-visible breaking change. Should we keep populating `reason` (e.g. map it to `details`) for backward compatibility, or at minimum document this as a known breaking change?

common/src/main/java/org/opensearch/sql/common/error/QueryProcessingStage.java
```diff
@@ -0,0 +1,46 @@
...
+   * ANALYZING stage: Semantic validation and type checking. Errors: Field not found, type
+   * mismatches, semantic violations.
+   */
+  ANALYZING("Parsing and validating the query"),
+
+  /**
+   * PLAN_CONVERSION stage: Conversion to Calcite execution plan with system limits. Errors:
+   * Unsupported operations, plan conversion failures.
+   */
+  PLAN_CONVERSION("Preparing the query for physical execution"),
```
RyanL1997, line 27: [Caught by claude] Nit: ANTLR parsing happens before this stage — the `ANALYZING` stage covers semantic analysis and type-checking, not parsing. The description "Parsing and validating the query" could be misleading. Something like "Validating and type-checking the query" might be more accurate?

legacy/src/main/java/org/opensearch/sql/legacy/plugin/RestSqlAction.java
```diff
@@ -168,14 +167,28 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+    int code = getRawErrorCode(ex);
+    return RestStatus.fromCode(code);
+  }
+
+  private static int getRawErrorCode(Exception ex) {
+    // Recursively unwrap ErrorReport to get to the underlying cause
+    if (ex instanceof ErrorReport) {
+      return getRawErrorCode(((ErrorReport) ex).getCause());
     }
+    if (ex instanceof OpenSearchException) {
```
RyanL1997, line 185: This is more like a question from my end:

The SQL path's `isClientError()` wasn't updated in the same way as the PPL path's `isClientError()` — e.g. `NullPointerException` is still treated as a client error here. This asymmetry could lead to the same exception type returning 400 via SQL but 500 via PPL. Should both paths be aligned?

RyanL1997 (commented):
Hi @Swiddis , thanks for this enhancement. I think the change itself looks good, however, I do have some questions regarding to the user experience-wise design pattern/minor design philosophy.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
Swiddis: Those two patterns were redundant, they're subclasses of other exception types that are already client errors. Just cleaned them since I was already refactoring there.

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
Swiddis: It's from the inner logic of `throwAsRuntime` since I was breaking its internals out to make it a report, probably redundant. Can clean

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
Swiddis: I'm not sure what the semantic difference is between suppressed exceptions and the main cause chain, but when I was patching this, the root cause was in the `suppressed` exception, so it seems like it's worth doing as a special-case here if Calcite relies on the suppressed exception chain for something. I think this behavior is correct.

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
Swiddis: By design -- this is updating a generic error with more specific context & message when the special case matches.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLQueryAction.java
RyanL1997: Thanks, confirmed — both extend `QueryEngineException` which is still in the list. Makes sense.

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
RyanL1997: Makes sense for the Calcite context — Calcite does stash meaningful errors as suppressed exceptions.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/error/ErrorMessage.java
Swiddis: wow, not sure how i missed that. will add a reason field and switch some things around to use it. `details` is probably more optional in the current setup in that case since the context system covers a lot of the same ground, most clients should support optional `details` since it's not always present today anyway

legacy/src/main/java/org/opensearch/sql/legacy/plugin/RestSqlAction.java
Swiddis: Just didn't do the same careful treatment for SQL since it's still V2 and might be more sensitive to sudden changes, planning on doing this mostly for Calcite. This is just a small set of patches to make it minimally handle reports in shared code (e.g. NodeClient is used by both engines)

RyanL1997: Makes sense to me.

RyanL1997 (approved):

ahkcs (approved):

PR #5273, opened 2026-03-26
Classify TaskCancelledException as client error instead of system error

ahkcs:
## Summary
When a user cancels a PPL query via the `_tasks/_cancel` API, the `TaskCancelledException` is wrapped by Calcite (`RuntimeException → SQLException → TaskCancelledException`) and reaches the REST error handler as a generic exception. This results in:
- HTTP 500 instead of 400
- Incrementing `PPL_FAILED_REQ_COUNT_SYS` (system error metric) instead of `PPL_FAILED_REQ_COUNT_CUS` (client error metric)

Since query cancellation is user-initiated, it should not be classified as a system error.

### Changes
- Add `hasCauseOf()` helper to walk the exception cause chain
- Add `TaskCancelledException` to `isClientError()` via cause chain check
- Add unit tests for wrapped and direct `TaskCancelledException`

### Related
- Follow-up to [#5254](https://github.com/opensearch-project/sql/pull/5254) (query cancellation support) and https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11593#issuecomment-4137212444
- Related issue: [#4887](https://github.com/opensearch-project/sql/issues/4887)

## Test plan
- [x] Unit tests: wrapped `TaskCancelledException` classified as client error
- [x] Unit tests: direct `TaskCancelledException` classified as client error
- [x] Unit tests: generic `RuntimeException` still classified as system error

OSD side log:
```
server    log   [19:02:02.468] [error][plugins][queryEnhancements] pplSearchStrategy: {
  "error": {
    "reason": "Query cancelled",
    "details": "The task is cancelled.",
    "type": "TaskCancelledException"
  },
  "status": 400
```

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/response/error/ErrorMessage.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/response/error/ErrorMessage.java
@@ -8,6 +8,7 @@ package org.opensearch.sql.opensearch.response.error;
 import lombok.Getter;
 import org.json.JSONObject;
 import org.opensearch.core.rest.RestStatus;
+import org.opensearch.core.tasks.TaskCancelledException;
 
 /** Error Message. */
 public class ErrorMessage {
@@ -37,6 +38,9 @@ public class ErrorMessage {
   }
 
   protected String fetchReason() {
+    if (exception instanceof TaskCancelledException) {
+      return "Query cancelled";
+    }
     return status == RestStatus.BAD_REQUEST.getStatus()
         ? "Invalid Query"
         : "There was internal problem at backend";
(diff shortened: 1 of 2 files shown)


ahkcs:
Note: `TaskCancelledException` inherits `INTERNAL_SERVER_ERROR` (500) from `OpenSearchException` by default. Ideally, core would classify this with a more appropriate status (e.g., 400 ) so all plugins benefit. This PR is a customized fix on the plugin side to avoid inflating system error metrics for user-initiated cancellations. If core updates the classification in the future, this special-casing can be removed.

Swiddis:
If we still need this after error enhancements, we can do it by expanding on the OpenSearchException check in the new interface to add some overridden exception types: https://github.com/opensearch-project/sql/pull/5266/changes#diff-3ad41d4c6d562cf927e38e1186467082b21d979bbeb732b87f73651bed1c3fa2R65-R67

PR #5267, opened 2026-03-25
Add query routing and execution handoff for Parquet-backed indices

ahkcs:
## Summary
Implements query routing and execution handoff for Project Mustang's unified query pipeline ([#5247](https://github.com/opensearch-project/sql/issues/5247)).

This PR covers **Step 2** and **Step 3** (partial) from the issue:

- **Index detection**: `StubIndexDetector.isAnalyticsIndex()` extracts index name from PPL query via regex, checks for `parquet_` prefix. Will be replaced by `UnifiedQueryParser` and index settings.
- **Query routing**: `TransportPPLQueryAction.doExecute()` routes parquet indices to `RestUnifiedQueryAction`, everything else to existing Lucene path unchanged. Routing happens after PPL enabled check, metrics, request ID, and profiling setup.
- **Planning**: `UnifiedQueryPlanner.plan()` parses PPL and generates Calcite `RelNode` against a stub schema. Query size limit added to plan as `LogicalSystemLimit`, consistent with PPL V3 (`QueryService.convertToCalcitePlan`).
- **Execution**: `AnalyticsExecutionEngine` calls `QueryPlanExecutor.execute(relNode)`, converts `Iterable<Object[]>` to `QueryResponse` with type mapping via `OpenSearchTypeFactory`.
- **Thread scheduling**: Runs on `sql-worker` thread pool with security context propagation via `withCurrentContext()`.
- **Profiling**: Uses `clearingListener` from `wrapWithProfilingClear()` to ensure profiling thread-local state is cleaned up.
- **Stub executor**: `StubQueryPlanExecutor` returns canned data for `parquet_logs` and `parquet_metrics` tables.

- **Response formatting**: `QueryResponse` -> `QueryResult` with query-type-aware `LangSpec` (PPL_SPEC or SQL_SPEC) -> `JdbcResponseFormatter` produces standard JDBC JSON (`schema`, `datarows`, `total`, `size`, `status`).
- **Metrics**: Query-type-aware success/failure metrics (PPL_REQ_TOTAL vs REQ_TOTAL, PPL_FAILED_REQ_COUNT_* vs FAILED_REQ_COUNT_*).

- **Error classification**: Matches `RestPPLQueryAction.isClientError()` list minus NPE (NPE is a server bug in the analytics path). Will sync with #5266.
- **Error formatting**: Uses `ErrorMessageFactory.createErrorMessage()` for standard error response format.
[...]

PR #5275, opened 2026-03-26
Add explain support for analytics engine path

ahkcs:
## Summary
Adds explain support for the analytics engine path ([#5247](https://github.com/opensearch-project/sql/issues/5247)). This is the second PR building on #5267 (query routing and execution handoff).

This PR covers **Step 3** (explain)

- **AnalyticsExecutionEngine.explain()**: Returns logical plan via `RelOptUtil.toString()` with mode-aware `SqlExplainLevel` (SIMPLE/STANDARD/COST). Physical and extended plans are null since the analytics engine is not yet available.
- **RestUnifiedQueryAction.explain()**: New entry point for explain requests on the analytics path. Formats response via `JsonResponseFormatter` with `normalizeLf()`.
- **TransportPPLQueryAction routing**: Checks `isExplainRequest()` for analytics indices and routes to `unifiedQueryHandler.explain()` instead of `execute()`.

- **AnalyticsExplainIT**: Dedicated explain test class following CalciteExplainIT pattern. Compares actual explain JSON against expected output files in `resources/expectedOutput/analytics/`. Covers: simple scan, project, filter+project, aggregation, sort (with collation propagation), eval.

### Explain response format
```json
{
  "calcite": {
    "logical": "LogicalSystemLimit(...)\n  LogicalProject(ts=[$0], message=[$2])\n    LogicalFilter(condition=[=($1, 200)])\n      LogicalTableScan(table=[[opensearch, parquet_logs]])\n"
  }
}
```
Physical plan will be populated when the analytics engine provides a `QueryPlanExecutor.explain()` API.

## Changed Files
| File | Change |
|---|---|
| `core/.../analytics/AnalyticsExecutionEngine.java` | Added `explain(RelNode, ExplainMode, ...)` method |
| `plugin/.../rest/RestUnifiedQueryAction.java` | Added `explain()` entry point and `createExplainListener()` |
| `plugin/.../transport/TransportPPLQueryAction.java` | Route explain requests for analytics indices |
| `integ-test/.../ppl/AnalyticsExplainIT.java` | New: 6 explain tests with expected output file comparison |
| `integ-test/.../resources/expectedOutput/analytics/*.json` | New: 6 expected explain output files |

## Test plan
- 2 unit tests -- explain logical plan, explain error propagation
- 5 integration tests in AnalyticsPPLIT -- explain response structure, project+scan, filter, aggregation, sort
- 6 integration tests in AnalyticsExplainIT -- JSON comparison against expected output files (scan, project, filter+project, aggregation, sort, eval)

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
@@ -6,13 +6,11 @@
 package org.opensearch.sql.plugin.rest;
 
 import static org.opensearch.sql.executor.ExecutionEngine.ExplainResponse;
-import static org.opensearch.sql.executor.ExecutionEngine.ExplainResponse.normalizeLf;
 import static org.opensearch.sql.lang.PPLLangSpec.PPL_SPEC;
 import static org.opensearch.sql.opensearch.executor.OpenSearchQueryManager.SQL_WORKER_THREAD_POOL_NAME;
 import static org.opensearch.sql.protocol.response.format.JsonResponseFormatter.Style.PRETTY;
 
 import java.util.Map;
-import java.util.Optional;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.schema.impl.AbstractSchema;
 import org.apache.logging.log4j.LogManager;
@@ -35,11 +33,8 @@ import org.opensearch.sql.plugin.rest.analytics.stub.StubSchemaProvider;
 import org.opensearch.sql.plugin.transport.TransportPPLQueryResponse;
 import org.opensearch.sql.ppl.domain.PPLQueryRequest;
 import org.opensearch.sql.protocol.response.QueryResult;
-import org.opensearch.sql.protocol.response.format.Format;
 import org.opensearch.sql.protocol.response.format.JdbcResponseFormatter;
-import org.opensearch.sql.protocol.response.format.JsonResponseFormatter;
 import org.opensearch.sql.protocol.response.format.ResponseFormatter;
-import org.opensearch.sql.protocol.response.format.YamlResponseFormatter;
 import org.opensearch.transport.client.node.NodeClient;
 
 /**
@@ -87,21 +82,25 @@ public class RestUnifiedQueryAction {
     client
         .threadPool()
         .schedule(
-            withCurrentContext(() -> doExecute(query, queryType, pplRequest, false, listener)),
+            withCurrentContext(() -> doExecute(query, queryType, pplRequest, listener)),
             new TimeValue(0),
             SQL_WORKER_THREAD_POOL_NAME);
   }
 
-  /** Explain a query through the unified query pipeline on the sql-worker thread pool. */
+  /**
+   * Explain a query through the unified query pipeline on the sql-worker thread pool. Returns
+   * ExplainResponse via ResponseListener so the caller (TransportPPLQueryAction) can format it
+   * using its own createExplainResponseListener, reusing the existing format-aware logic.
+   */
   public void explain(
       String query,
       QueryType queryType,
       PPLQueryRequest pplRequest,
-      ActionListener<TransportPPLQueryResponse> listener) {
+      ResponseListener<ExplainResponse> listener) {
     client
         .threadPool()
         .schedule(
-            withCurrentContext(() -> doExecute(query, queryType, pplRequest, true, listener)),
+            withCurrentContext(() -> doExplain(query, queryType, pplRequest, listener)),
             new TimeValue(0),
             SQL_WORKER_THREAD_POOL_NAME);
   }
@@ -110,7 +109,6 @@ public class RestUnifiedQueryAction {
       String query,
       QueryType queryType,
       PPLQueryRequest pplRequest,
-      boolean isExplain,
       ActionListener<TransportPPLQueryResponse> listener) {
     try {
       long startTime = System.nanoTime();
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -115,8 +133,13 @@ private void doExecute(
             (planTime - startTime) / 1_000_000,
             queryType);
 
-        analyticsEngine.execute(
-            plan, planContext, createQueryListener(queryType, planTime, listener));
+        if (isExplain) {
+          analyticsEngine.explain(
+              plan, ExplainMode.STANDARD, planContext, createExplainListener(listener));
```
penghuo: ExplainMode.STANDARD is hardcoded here, so the user's `?mode=` parameter is ignored. `pplRequest.mode()` (PPLQueryRequest.java:103) already parses the request mode — pass it through instead of the constant so that `simple`, `cost`, and `extended` requests produce the correct SqlExplainLevel.

ahkcs: Good catch. Now using pplRequest.mode() instead of hardcoded ExplainMode.STANDARD

core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
```diff
@@ -77,6 +80,26 @@ public void execute(
...
+      ExplainMode mode,
+      CalcitePlanContext context,
+      ResponseListener<ExplainResponse> listener) {
+    try {
+      SqlExplainLevel level =
+          mode == ExplainMode.SIMPLE
+              ? SqlExplainLevel.NO_ATTRIBUTES
+              : mode == ExplainMode.COST
+                  ? SqlExplainLevel.ALL_ATTRIBUTES
+                  : SqlExplainLevel.EXPPLAN_ATTRIBUTES;
```
dai-chen: Add a method in ExplainMode enum to encapsulate this, e.g., `toExplainLevel`?

integ-test/src/test/java/org/opensearch/sql/ppl/AnalyticsExplainIT.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private String loadAnalyticsExpectedPlan(String fileName) {
+    return loadFromFile("expectedOutput/analytics/" + fileName);
+  }
+
+  @Test
+  public void testExplainSimpleScan() throws IOException {
+    String query = "source = opensearch.parquet_logs";
+    String result = explainQueryYaml(query);
+    LOG.info("[testExplainSimpleScan] query: {}\nresult:\n{}", query, result);
```
dai-chen: remove all logging if no special reason?

integ-test/src/test/java/org/opensearch/sql/ppl/AnalyticsPPLIT.java
```diff
@@ -150,6 +150,8 @@ public void testSingleFieldProjection() throws IOException {
     verifyNumOfRows(result, 3);
   }
 
+  // Explain is covered by AnalyticsExplainIT with expected output file comparison.
```
dai-chen: remove such comment?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -162,6 +187,42 @@ public void onFailure(Exception e) {
     };
   }
 
+  private ResponseListener<ExplainResponse> createExplainListener(
+      PPLQueryRequest pplRequest, ActionListener<TransportPPLQueryResponse> transportListener) {
+    return new ResponseListener<ExplainResponse>() {
```
dai-chen: Is this exactly the same as `TransportPPLQueryAction.createExplainResponseListener`?

core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
ahkcs: Added `toExplainLevel()` to ExplainMode enum. AnalyticsExecutionEngine uses it directly. `OpenSearchExecutionEngine` has the same ternary logic and can be updated in a separate follow-up PR

integ-test/src/test/java/org/opensearch/sql/ppl/AnalyticsExplainIT.java
ahkcs: Removed logging

integ-test/src/test/java/org/opensearch/sql/ppl/AnalyticsPPLIT.java
ahkcs: Removed

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Yes, it was duplicated. Removed `createExplainListener` from `RestUnifiedQueryAction`. Now explain() returns `ExplainResponse` via `ResponseListener`, and `TransportPPLQueryAction` wraps it with its existing `createExplainResponseListener` for format-aware formatting.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
dai-chen: Please note down this as part of contract between analytics engine. Either they can provide explain API (e.g., `QueryPlanExecutor.explain()`) or we only show logical plan for this execution path and document it somewhere. Thanks!

PR #5274, opened 2026-03-26
Add unified query parser API

dai-chen:
### Description

Extract parsing logic from UnifiedQueryPlanner into a `UnifiedQueryParser<R>` interface with language-specific implementations: `PPLQueryParser` (returns `UnresolvedPlan`) and `CalciteSqlQueryParser` (returns `SqlNode`).

#### Implementation Notes

- **Context-owned parser**: Following the Spark/Flink pattern, `UnifiedQueryContext` creates and owns the parser, ensuring consistent configuration across all components.
- **No parse with visitor API yet**: A language-neutral `parse(query, visitor)` was explored but deferred. PPL's custom AST and Calcite's SqlNode tree have fundamentally different node structures. Until this tension resolves, callers use native visitors on the typed `parse(query)` result.

### Related Issues
Part of https://github.com/opensearch-project/sql/issues/5248

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
 - [x] New functionality has javadoc added.
 - [x] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5285, opened 2026-03-30
Enable profiling and migrate to UnifiedQueryParser

ahkcs:
## Summary
Implements two tasks from [#5247 comment](https://github.com/opensearch-project/sql/issues/5247#issuecomment-4156731599):

### Task 1: Enable profiling (#5268)
Add `.profiling(pplRequest.profile())` to `UnifiedQueryContext.builder()` in both `doExecute` and `doExplain`. When `profile=true` is requested, profiling metrics are collected during query planning and execution via the unified query API's auto-profiling support.

### Task 2: Migrate to UnifiedQueryParser (#5274)
Replace `StubIndexDetector` (regex-based index extraction) with `PPLQueryParser` (AST-based). Parse the query to get the `UnresolvedPlan`, walk the AST tree to find the `Relation` node, and extract the table name via `getTableQualifiedName()`. This is more robust than regex and uses the same parser infrastructure as the rest of the pipeline.

## Changed Files
| File | Change |
|---|---|
| `plugin/.../rest/RestUnifiedQueryAction.java` | Add profiling, replace regex with PPLQueryParser for index extraction, remove manual latency logging |
| `plugin/.../transport/TransportPPLQueryAction.java` | Pass Settings to RestUnifiedQueryAction, use instance method |
| `plugin/.../rest/analytics/stub/StubIndexDetector.java` | Deleted -- replaced by PPLQueryParser |
| `plugin/.../rest/RestUnifiedQueryActionTest.java` | Updated for instance method |

## Test plan
- 2 unit tests -- routing detection (parquet vs non-parquet, null/empty)
- 9 integration tests -- AnalyticsPPLIT (full pipeline)
- 6 integration tests -- AnalyticsExplainIT (explain plans)

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
@@ -31,7 +31,6 @@ import org.opensearch.sql.executor.QueryType;
 import org.opensearch.sql.executor.analytics.AnalyticsExecutionEngine;
 import org.opensearch.sql.executor.analytics.QueryPlanExecutor;
 import org.opensearch.sql.lang.LangSpec;
-import org.opensearch.sql.monitor.profile.MetricName;
 import org.opensearch.sql.plugin.rest.analytics.stub.StubSchemaProvider;
 import org.opensearch.sql.plugin.transport.TransportPPLQueryResponse;
 import org.opensearch.sql.ppl.domain.PPLQueryRequest;
@@ -59,8 +58,13 @@ public class RestUnifiedQueryAction {
   }
 
   /**
-   * Check if the query targets an analytics engine index (e.g., Parquet-backed). Creates a {@link
-   * UnifiedQueryContext} to use its parser for index name extraction, supporting both PPL and SQL.
+   * Check if the query targets an analytics engine index (e.g., Parquet-backed). Uses the context's
+   * parser for index name extraction, supporting both PPL and SQL.
+   *
+   * <p>Note: This creates a separate UnifiedQueryContext for parsing. The context cannot be shared
+   * with doExecute/doExplain because UnifiedQueryContext holds a Calcite JDBC connection that fails
+   * when used across threads (transport thread -> sql-worker thread). When real catalog metadata
+   * makes this expensive, consider moving the routing check to the sql-worker thread.
    */
   public boolean isAnalyticsIndex(String query, QueryType queryType) {
     if (query == null || query.isEmpty()) {
@@ -123,14 +127,7 @@ public class RestUnifiedQueryAction {
       CalcitePlanContext planContext = context.getPlanContext();
       plan = addQuerySizeLimit(plan, planContext);
 
-      RelNode finalPlan = plan;
-      context.measure(
-          MetricName.EXECUTE,
-          () -> {
-            analyticsEngine.execute(
-                finalPlan, planContext, createQueryListener(queryType, listener));
-            return null;
-          });
+      analyticsEngine.execute(plan, planContext, createQueryListener(queryType, listener));
     } catch (Exception e) {
       listener.onFailure(e);
     }
@@ -148,13 +145,7 @@ public class RestUnifiedQueryAction {
       CalcitePlanContext planContext = context.getPlanContext();
       plan = addQuerySizeLimit(plan, planContext);
 
-      RelNode finalPlan = plan;
-      context.measure(
-          MetricName.EXECUTE,
-          () -> {
-            analyticsEngine.explain(finalPlan, pplRequest.mode(), planContext, listener);
-            return null;
-          });
+      analyticsEngine.explain(plan, pplRequest.mode(), planContext, listener);
     } catch (Exception e) {
       listener.onFailure(e);
     }
--- a/core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
+++ b/core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
@@ -24,6 +24,9 @@ import org.opensearch.sql.data.type.ExprType;
 import org.opensearch.sql.executor.ExecutionContext;
 import org.opensearch.sql.executor.ExecutionEngine;
 import org.opensearch.sql.executor.pagination.Cursor;
+import org.opensearch.sql.monitor.profile.MetricName;
+import org.opensearch.sql.monitor.profile.ProfileMetric;
+import org.opensearch.sql.monitor.profile.QueryProfiling;
 import org.opensearch.sql.planner.physical.PhysicalPlan;
 
 /**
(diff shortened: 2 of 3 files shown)


plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -49,30 +52,59 @@ public class RestUnifiedQueryAction {
...
 
   /**
-   * Check if the query targets an analytics engine index. Delegates to {@link StubIndexDetector}
-   * which will be replaced by UnifiedQueryParser and index settings when available.
+   * Check if the query targets an analytics engine index (e.g., Parquet-backed). Uses {@link
+   * PPLQueryParser} to parse the query and extract the index name from the AST.
    */
-  public static boolean isAnalyticsIndex(String query) {
-    return StubIndexDetector.isAnalyticsIndex(query);
+  public boolean isAnalyticsIndex(String query) {
```
dai-chen: I assume this will be reused by both unified SQL and PPL. In this case, we'd better use parser from `UnifiedQueryContext`? Maybe the context object needs to be created early (before explain/execute triggered)?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -49,30 +52,59 @@ public class RestUnifiedQueryAction {
...
+
+  /** Extract the source index name by parsing the PPL AST and finding the Relation node. */
+  private String extractIndexName(String query) {
+    UnresolvedPlan plan = pplParser.parse(query);
+    Relation relation = findRelation(plan);
+    return relation != null ? relation.getTableQualifiedName().toString() : null;
+  }
+
+  /** Walk the AST to find the Relation (table scan) node. */
+  private static Relation findRelation(UnresolvedPlan plan) {
```
dai-chen: I think most of the logic of `isAnalyticsIndex`, `extractIndexName` and `findRelation` should be encapsulated into a AST visitor?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -127,14 +159,7 @@ private void doExecute(
...
 
-        long planTime = System.nanoTime();
-        LOG.info(
-            "[unified] Planning completed in {}ms for {} query",
-            (planTime - startTime) / 1_000_000,
-            queryType);
-
-        analyticsEngine.execute(
-            plan, planContext, createQueryListener(queryType, planTime, listener));
+        analyticsEngine.execute(plan, planContext, createQueryListener(queryType, listener));
```
dai-chen: wrap by `context.measure`?

ahkcs:
>For the profiling, is the result automatically populated into REST response? Is it possible to add integration test or verify manually?

The profiling result is now populated into the REST response. Switched from `JdbcResponseFormatter` to `SimpleJsonResponseFormatter` which calls `QueryProfiling.current().finish()` to include the profile field. This also aligns the response format with the existing Lucene PPL path which already uses `SimpleJsonResponseFormatter`.

The EXECUTE metric is recorded inside `AnalyticsExecutionEngine.execute()`, between the actual execution and the `listener.onResponse()` call. This is necessary because `SimpleJsonResponseFormatter` calls `QueryProfiling.finish()` inside the listener callback — if we used `context.measure()` in `RestUnifiedQueryAction`, the metric would be written after `finish()` already took the snapshot, resulting in 0ms.

Added integration test that verifies:
- `profile` field exists when `profile=true`
- `phases.analyze.time_ms > 0` (planning took real time)
- `phases.execute.time_ms > 0` (execution took real time)

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Addressed by using `context.getParser()` from `UnifiedQueryContext` for index extraction, with an IndexNameExtractor AST visitor instead of manual tree walking. `isAnalyticsIndex(query, queryType)` takes QueryType so both PPL and SQL use the right parser.

For PPL, `context.getParser().parse()` returns `UnresolvedPlan` — the visitor finds the Relation node and extracts the table name. Left a `TODO` for SQL (SqlNode table extraction).
https://github.com/opensearch-project/sql/pull/5285/changes#diff-b3354580a474ed0cfe3286d76064284ffb683bc538d3f3ff013f5e7aadf0dc5eR182

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Updated. Replaced `findRelation` with an `IndexNameExtractor` visitor extending `AbstractNodeVisitor`. The visitor overrides visitRelation() to return the table name — tree traversal is handled by the base class's default visitChildren. `extractIndexName` calls `unresolvedPlan.accept(new IndexNameExtractor()`, null) instead of manual recursion.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Updated. Wrapped both `analyticsEngine.execute()` and `analyticsEngine.explain()` with `context.measure(MetricName.EXECUTE, ...)`. Planning is auto-profiled by `UnifiedQueryPlanner` internally.

ahkcs: Updated reply: Moved EXECUTE metric recording into `AnalyticsExecutionEngine.execute()` instead of wrapping with `context.measure()`. The issue with `context.measure()` is that `SimpleJsonResponseFormatter` calls `QueryProfiling.finish()` inside the listener callback (synchronously), which snapshots all metrics before `context.measure()`'s finally block writes the EXECUTE metric — resulting in 0ms.

The fix records the metric between the actual execution (planExecutor + row conversion) and the listener.onResponse() call, so it's captured before finish() snapshots. Added IT that asserts execute phase time_ms > 0 to catch this

ahkcs:
Did verification on commands like `DESCRIBE` and `SHOW DATASOURCES` produce ASTs without a Relation node, so `extractIndexName` returns null and `isAnalyticsIndex` returns false — they fall through to the existing `Lucene/PPLService` path automatically.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
dai-chen: I see. I feel it shouldn't do that. For now, is format metric missing?

ahkcs: FORMAT metric is not missing — it's recorded by `SimpleJsonResponseFormatter.buildJsonObject()` itself, before calling `finish()`. The IT output confirms all phases:
`analyze: 13.69ms, execute: 126.32ms, format: 1.29ms, optimize: 0ms`

dai-chen: @ahkcs ^^

ahkcs: `SimpleJsonResponseFormatter` calls `finish()` inside `buildJsonObject()` to snapshot metrics. Any metric written after that (like `context.measure()`'s finally block) is missed — that's why we record EXECUTE manually before the listener fires.

The main PPL path uses the same `SimpleJsonResponseFormatter` with the same finish() pattern, so this isn't specific to our analytics path.

To fully adopt `context.measure()`, I think the formatter would need to stop calling `finish()` internally. Should we make this a follow-up PR?

PR #5290, opened 2026-03-30
Come home, dependabot

Swiddis:
### Description
Adds a quick YML config to poke it back to life. We can remove the config again once it's alive if we want to keep the default config.

Some decisions:
- Weekly updates since daily tends to flood a lot, and our stale PR review is biweekly anyway
- The group config means all the updates come in one PR, instead of dozens of PRs for every update
- Skip changelog by default

### Related Issues
N/A

### Check List

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


Swiddis:
Not sure the exact dependabot behavior, but I know you can configure it with this & if it's reconfigured it'll come back. There might be another way to revive it without a file bump but even then I'd rather push this config if we're enabling it to prevent as much PR load.

PR #5302, opened 2026-04-01
Wire analytics-engine as extendedPlugins dependency

ahkcs:
## Summary
- Wire analytics-engine as `extendedPlugins` parent classloader dependency for shared Calcite classes
- Patch Calcite CALCITE-3745 to use thread context classloader for Janino compilation, fixing `CompileException` when parent classloader can't see SQL plugin classes
- Add `bundlePlugin` excludes to prevent jar hell between parent/child classloaders
- Add analytics-engine plugin to all test clusters (integTest, yamlRestTest, security, JDBC, doctest)
- Add commons-text to analytics-engine ZIP for Calcite fuzzy matching support
- Fix `isAnalyticsIndex` to use lightweight parsing context (no cluster state needed)

## Classloader Architecture
```
analytics-engine (PARENT classloader)
├── Calcite (patched), Janino, Guava, commons-*
├── analytics-framework (QueryPlanExecutor interface)
│
└── opensearch-sql (CHILD classloader)
    ├── core, ppl, sql, opensearch modules
    └── Calcite JARs excluded (inherited from parent)
```

## Calcite Patch (CALCITE-3745)
Calcite hardcodes `EnumerableInterpretable.class.getClassLoader()` for Janino compilation. When analytics-engine is the parent, this returns the parent classloader which can't see SQL plugin classes, causing `CompileException: Cannot determine simple type name "org"`.

The patch changes 2 lines in Calcite to prefer `Thread.currentThread().getContextClassLoader()`:

```java
// Before:
ClassLoader cl = EnumerableInterpretable.class.getClassLoader();

// After:
ClassLoader contextCl = Thread.currentThread().getContextClassLoader();
ClassLoader cl = contextCl != null ? contextCl : EnumerableInterpretable.class.getClassLoader();
```

Same change in `RexExecutable`. Our code sets the context classloader before Calcite execution in 2 places:
- `CalciteToolsHelper.OpenSearchRelRunners.run()` — before `runner.prepareStatement(rel)`
- `CalciteScriptEngine.compile()` — before `new RexExecutable(code, ...)`

**How this solves the problem:** After our code sets the context classloader to `CalciteToolsHelper.class.getClassLoader()` (SQL plugin's classloader), the patched Calcite picks it up via `Thread.currentThread().getContextClassLoader()` and passes it to Janino as `compiler.setParentClassLoader(cl)`. The SQL plugin's classloader is the child, which can see both its own classes AND the parent's classes (Calcite). Janino can now resolve all class references in the generated code — SQL plugin classes like `CalciteEnumerableIndexScan` (from child) and Calcite classes like `SqlFunctions` (from parent).

Related upstream issue: https://issues.apache.org/jira/browse/CALCITE-3745
Related issue: https://github.com/opensearch-project/sql/issues/5306


penghuo (commented):
Does temporary `lib/` only for testing purpose?

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -358,7 +358,143 @@ public Type getElementType() {
...
+    /**
+     * Implements the Enumerable path with classloader fix for Janino compilation. Calcite's {@code
+     * EnumerableInterpretable.getBindable()} hardcodes {@code
+     * EnumerableInterpretable.class.getClassLoader()} as the parent classloader for Janino. When
+     * analytics-engine is the parent classloader (via extendedPlugins), this returns the
+     * analytics-engine classloader which cannot see SQL plugin classes. This method replicates the
+     * Calcite implementation but uses this class's classloader (SQL plugin, child) which can see
+     * both parent and child classes.
+     */
+    private PreparedResult implementEnumerable(RelRoot root) {
```
penghuo: The `implementEnumerable()` and `compileWithPluginClassLoader()` methods were written when analytics-engine was the parent classloader (SQL plugin extended analytics-engine). But `plugin-descriptor.properties` in the ZIP shows `extended.plugins=opensearch-sql`, meaning the SQL plugin is now the parent. With this arrangement, `EnumerableInterpretable.class.getClassLoader()` and `CalciteToolsHelper.class.getClassLoader()` both return the SQL plugin classloader — they are identical. These ~130 lines of reflection code replicate Calcite internals unnecessarily and will break silently on Calcite upgrades.

Please verify whether `super.implement(root)` works correctly with the current classloader direction. If it does, remove this code entirely. If there is a remaining scenario requiring the fix, update the Javadoc (which still describes the old "analytics-engine is parent" arrangement).

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -358,7 +358,143 @@ public Type getElementType() {
...
+            org.apache.calcite.adapter.enumerable.EnumerableCalc.create(enumerable, program);
+      }
+
+      // Access the internalParameters map via reflection. This map is shared with the
+      // DataContext so stashed values (e.g., table scan references) are available at execution.
+      java.util.Map<String, Object> parameters;
+      try {
+        java.lang.reflect.Field f =
+            CalcitePrepareImpl.CalcitePreparingStmt.class.getDeclaredField("internalParameters");
+        f.setAccessible(true);
```
penghuo: `getDeclaredField("internalParameters")` with `setAccessible(true)` accesses a private Calcite field that is not part of public API. This will break at runtime if Calcite renames or removes the field, or if the JVM enforces strict module access. If this code is truly needed after verifying the classloader direction, add the Calcite version to the error message. Prefer a non-reflective approach long-term.

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -358,7 +358,143 @@ public Type getElementType() {
...
+                + " implements "
+                + Bindable.class.getName()
+                + ", "
+                + org.apache.calcite.runtime.Typed.class.getName()
+                + " {\n"
+                + s
+                + "\n}\n";
+        compiler.getClass().getMethod("cook", String.class).invoke(compiler, fullCode);
+        ClassLoader compiledClassLoader =
+            (ClassLoader) compiler.getClass().getMethod("getClassLoader").invoke(compiler);
```
penghuo: Each call to `compileWithPluginClassLoader()` creates a new Janino `SimpleCompiler` and classloader. Never closed or cached — will accumulate classloaders in metaspace under sustained load. Consider caching compiled results or verify whether `super.implement(root)` works with the current classloader direction.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -51,9 +51,14 @@ public class RestUnifiedQueryAction {
 
   private final AnalyticsExecutionEngine analyticsEngine;
   private final NodeClient client;
```
penghuo: `OpenSearchSchemaBuilder.buildSchema(clusterService.state())` is called on every query. On clusters with many indices this could add per-query latency. Consider caching the schema and invalidating on cluster state changes via `ClusterStateListener`.

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
penghuo: Please confirm does this necessary? 
With the reversed classloader (extended.plugins=opensearch-sql in analytics-engine's descriptor), the SQL plugin is the parent. Both EnumerableInterpretable.class.getClassLoader() and  CalciteToolsHelper.class.getClassLoader() return the SQL plugin's classloader — the workaround is solving a problem that no longer exists.

ahkcs:
> Does temporary `lib/` only for testing purpose?

The core team doesn't have published artifacts, we built it locally for testing purpose

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
ahkcs: Good catch. With the reversed classloader (SQL plugin as parent), this is no longer needed, updated to remove it

ahkcs: Updated to new approach described in PR description

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
```diff
@@ -137,8 +137,17 @@ public <T> T compile(
...
-    Function1<DataContext, Object[]> function =
-        new RexExecutable(code, "generated Rex code").getFunction();
+    // Set thread context classloader so patched Calcite (CALCITE-3745) uses the SQL
+    // plugin's classloader for Janino compilation instead of the parent classloader.
+    Thread currentThread = Thread.currentThread();
+    ClassLoader originalCl = currentThread.getContextClassLoader();
+    currentThread.setContextClassLoader(CalciteScriptEngine.class.getClassLoader());
+    Function1<DataContext, Object[]> function;
+    try {
+      function = new RexExecutable(code, "generated Rex code").getFunction();
```
dai-chen: The `code` here still has classloading issue?

plugin/build.gradle
```diff
@@ -55,11 +55,51 @@ opensearchplugin {
...
-    extendedPlugins = ['opensearch-job-scheduler']
+    extendedPlugins = ['opensearch-job-scheduler', 'analytics-engine']
     licenseFile rootProject.file("LICENSE.txt")
     noticeFile rootProject.file("NOTICE")
 }
 
+// Exclude jars provided by analytics-engine plugin (shared via extendedPlugins classloader).
+// Must match exactly what's in the analytics-engine ZIP to avoid both jar hell and missing classes.
+bundlePlugin {
+    exclude 'calcite-core-*.jar'
```
dai-chen, line 66: All these need to be excluded here instead of changing dependency scope?

integ-test/src/test/resources/expectedOutput/analytics/explain_aggregation.yaml
```diff
@@ -3,5 +3,5 @@ calcite:
     LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
       LogicalProject(count()=[$1], status=[$0])
         LogicalAggregate(group=[{0}], count()=[COUNT()])
-          LogicalProject(status=[$1])
+          LogicalProject(status=[$2])
```
dai-chen, line 6: Do we know why this changed? Impact by the classloader change?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/CalciteScriptEngine.java
ahkcs: Yes, same classloader issue. Calcite uses Janino in two places — `EnumerableInterpretable` (full query plan compilation, triggered in `CalciteToolsHelper`) and `RexExecutable` (individual expression compilation, triggered here in `CalciteScriptEngine` for pushdown filter scripts). Both hardcode the parent classloader.

plugin/build.gradle
ahkcs: Tried changing dependency scope. Changing to `compileOnly` requires adding `compileOnly 'org.apache.calcite:calcite-core:1.41.0'` to 4 modules (opensearch, ppl, api, legacy) because `compileOnly` doesn't propagate to consumers like `api` does. The `bundlePlugin` excludes approach keeps all the changes in one block in one file (plugin/build.gradle). Either way works. 
Happy to switch if you prefer the `compileOnly` approach.

integ-test/src/test/resources/expectedOutput/analytics/explain_aggregation.yaml
ahkcs: Hi Chen, it's not `classloader` related. The column index changed from `$1` to `$2` because we replaced `StubSchemaProvider` (hardcoded field order) with `OpenSearchSchemaBuilder.buildSchema(clusterState)` (reads real index mappings). The real schema returns fields alphabetically (ip_addr=$0, message=$1, status=$2, ts=$3), while the old stub had a different order (ts=$0, status=$1, ...). Same change across all analytics explain YAMLs.

plugin/build.gradle
dai-chen: No problem. We can double check when we move away from this temporary bundle approach after analytics-engine published. Thanks!

integ-test/src/test/resources/expectedOutput/analytics/explain_aggregation.yaml
dai-chen: Just to confirm: is `testFilterBooleanFieldWithStringLiteral` existing test unrelated to analytics-engine and this stub API change?

ahkcs: The `SAFE_CAST('TRUE')` constant folding happens during the PPL planning phase (`UnifiedQueryPlanner.plan()`), which is not covered by our `setContextClassLoader` wrapper. On main there's no parent-child classloader, so `RexSimplify` can compile and fold `SAFE_CAST('TRUE')` to true during planning. On our branch, the parent classloader can't compile the `SAFE_CAST` UDF, so it stays unreduced in the logical plan.

This is something that's introduced by our parent-child classloader, some paths are using SQL plugin Calcite(the paths we patched) and some paths are using Analytics-Engine Calcite(unpatched paths), which introduces this difference

ahkcs: Updated:
Wrapper `explainWithCalcite` with our context classloader as well since `SAFE_CAST` is a UDF from SQL plugin and Janino can't compile it

dai-chen (approved):
Thanks for the changes! We can do more deep dive for the classloader change impact later.

plugin/build.gradle
```diff
@@ -55,11 +55,51 @@ opensearchplugin {
     name 'opensearch-sql'
     description 'OpenSearch SQL'
     classname 'org.opensearch.sql.plugin.SQLPlugin'
-    extendedPlugins = ['opensearch-job-scheduler']
+    extendedPlugins = ['opensearch-job-scheduler', 'analytics-engine']
```
mch2, line 58: We'll want to make analytics-engine an optional dependency of sql plugin imo - i think we can do this by injecting a provider for the exposed components from analytics-engine.

ahkcs: Good call — we should switch to this so the SQL plugin works standalone. Will update when merging the feature branch to main

ahkcs: Opened PR to make it optional: https://github.com/opensearch-project/sql/pull/5341

PR #5304, opened 2026-04-02
[BugFix] Fix appendcol subquery cannot resolve MAP sub-paths after spath (#5186)

### LINKED ISSUE #5186: [BUG] `appendcol` subquery cannot resolve MAP sub-paths
**What is the bug?**

When using `appendcol` after `spath` auto-extract, the subquery inside `appendcol` cannot resolve dotted MAP paths like `doc.user.city`. The subquery fails with `INTERNAL_ITEM function expects {[ARRAY,INTEGER]|[STRUCT,ANY]}, but got [STRING,STRING]`.

```
  -d '{"doc": "{\"user\":{\"name\":\"John\",\"age\":30,\"city\":\"NYC\"}}"}'
  -d '{"doc": "{\"user\":{\"name\":\"Alice\",\"age\":25,\"city\":\"LA\"}}"}'
```
2. Enable Calcite and run:
```
source=t | spath input=doc | appendcol [ stats count() as cnt by doc.user.city ] | fields doc.user.name, doc.user.city, cnt
Error: INTERNAL_ITEM function expects {[ARRAY,INTEGER]|[STRUCT,ANY]}, but got [STRING,STRING]
```

The `appendcol` subquery should be able to reference MAP sub-paths from the spath-transformed pipeline. The subquery should produce subpath columns that are appended to the main result.

 - Version: PPL Calcite V3
 - Plugins: 3.6

N/A

- `appendcol` with flat fields works correctly: appendcol [ stats count() as total ] succeeds.
- The workaround is to materialize MAP sub-paths before appendcol: eval city=doc.user.city | appendcol [ stats count() as cnt by city ].
- This is a design limitation of appendcol — the subquery always starts from the raw source relation, not from the current pipeline state. Fixing this would require either propagating the main pipeline's transformations to the subquery, or changing appendcol to pass the current RelNode as the subquery's input instead of the raw source relation.
- The root cause is that `appendcol`'s subquery operates on the raw source table, not the transformed pipeline. In `visitAppendCol`, the subsearch gets the original `Relation` (source table scan) attached as its child via `getRelation(node)` + `transformPlanToAttachChild`. The raw source table has `doc` as a STRING column. When the subquery tries to resolve `doc.user.city`, the expression analyzer creates `ITEM(doc, 'user.city')` where `doc` is STRING — but `INTERNAL_ITEM` requires STRUCT or ARRAY input.

The main pipeline's `spath` command converts `doc` from STRING to MAP, but this transformation is not inherited by the subquery.

```java
// visitAppendCol (CalciteRelNodeVisitor.java)
// 1. resolve main plan (includes spath → doc becomes MAP)
visitChildren(node, context);

// 3. build subsearch tree — attaches raw Relation(t), NOT the spath-transformed plan
UnresolvedPlan relation = getRelation(node);  // extracts Relation(t) from main plan tree
transformPlanToAttachChild(node.getSubSearch(), relation);

// 4. resolve subsearch plan — subsearch scans raw table, doc is STRING not MAP
node.getSubSearch().accept(this, context);
// → stats count() as cnt by doc.user.city
// → ITEM(doc, 'user.city') where doc is STRING → INTERNAL_ITEM rejects STRING input
```

qianheng-aws:
### Description

When `spath` auto-extract transforms a STRING field to MAP, the `appendcol` subquery could not resolve dotted MAP paths like `doc.user.city`. The subquery failed with `INTERNAL_ITEM function expects {[ARRAY,INTEGER]|[STRUCT,ANY]}, but got [STRING,STRING]`.

**Root cause**: `visitAppendCol` attached only the raw `Relation` to the subsearch via `getRelation()` + `transformPlanToAttachChild()`, missing schema-transforming commands like `spath`. Additionally, `PPLQueryDataAnonymizer.visitAppendCol` runs first and attaches a raw `Relation`, and `Relation.attach()` is a no-op, so subsequent attachment attempts in the visitor silently failed.

**Fix**: `buildAppendColSubsearchSource()` collects `SPath` nodes from the main plan chain and builds a new `SPath -> Relation` chain. The subsearch attachment walks to find the parent of any existing `Relation` and re-attaches with the full source chain including spath transformations.

### Related Issues
Resolves #5186

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -22,7 +22,6 @@ import static org.opensearch.sql.calcite.utils.PlanUtils.ROW_NUMBER_COLUMN_FOR_S
 import static org.opensearch.sql.calcite.utils.PlanUtils.ROW_NUMBER_COLUMN_FOR_SUBSEARCH;
 import static org.opensearch.sql.calcite.utils.PlanUtils.getRelation;
 import static org.opensearch.sql.calcite.utils.PlanUtils.getRexCall;
-import static org.opensearch.sql.calcite.utils.PlanUtils.transformPlanToAttachChild;
 import static org.opensearch.sql.utils.SystemIndexUtils.DATASOURCES_TABLE_NAME;
 
 import com.google.common.base.Strings;
@@ -2434,9 +2433,22 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     context.relBuilder.projectPlus(
         context.relBuilder.alias(mainRowNumber, ROW_NUMBER_COLUMN_FOR_MAIN));
 
-    // 3. build subsearch tree (attach relation to subsearch)
-    UnresolvedPlan relation = getRelation(node);
-    transformPlanToAttachChild(node.getSubSearch(), relation);
+    // 3. build subsearch tree: attach source plan that includes spath transformations
+    //    so the subsearch can resolve MAP sub-paths like doc.user.city (fixes #5186)
+    UnresolvedPlan subsearchSource = buildAppendColSubsearchSource(node);
+    // Walk down the subsearch to find the deepest node that should receive the source.
+    // The anonymizer may have already attached a raw Relation. We find the node whose
+    // child is a Relation (or is empty) and (re-)attach our subsearchSource there.
+    UnresolvedPlan parent = node.getSubSearch();
+    while (parent.getChild() != null && !parent.getChild().isEmpty()) {
+      UnresolvedPlan child = (UnresolvedPlan) parent.getChild().getFirst();
+      if (child instanceof Relation) {
+        // Parent's child is the Relation -- replace it with our source chain
+        break;
+      }
+      parent = child;
+    }
+    parent.attach(subsearchSource);
     // 4. resolve subsearch plan
     node.getSubSearch().accept(this, context);
     // 5. add row_number() column to subsearch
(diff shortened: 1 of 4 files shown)


qianheng-aws:
## Decision Log

**Root Cause**: Two interacting issues caused the bug:
1. `visitAppendCol` used `getRelation(node)` to extract only the raw `Relation` (table scan) from the main plan, discarding schema-transforming commands like `spath` that convert STRING fields to MAP.
2. `PPLQueryDataAnonymizer.visitAppendCol` runs before `CalciteRelNodeVisitor.visitAppendCol` and attaches a raw `Relation` to the subsearch via `transformPlanToAttachChild()`. Since `Relation.attach()` is a no-op (returns `this` without setting a child), any subsequent attempt to attach a different source to the subsearch's leaf silently failed — the Relation was already there but couldn't be replaced through the leaf.

**Approach**: 
- `buildAppendColSubsearchSource()` walks the main plan's child chain, collects SPath nodes, and builds a fresh `SPath(copy) -> Relation` chain.
- Instead of using `transformPlanToAttachChild()` (which finds the leaf and attaches, but fails when the leaf is an immutable Relation), the fix walks the subsearch to find the parent whose child is a Relation, then calls `parent.attach(subsearchSource)` to replace the raw Relation with the SPath-enhanced source.

**Alternatives Rejected**:
1. **Attaching the full child plan** (all commands between Relation and AppendCol): This broke existing tests because schema-reducing commands like `stats` or `eval` would restrict the available fields for the subsearch.
2. **Using `transformPlanToAttachChild` with SPath copies**: The anonymizer pre-attaches a raw Relation to the subsearch. `transformPlanToAttachChild` finds the Relation as the leaf and calls `Relation.attach(spathCopy)` — which is a no-op. The SPath was silently discarded.

**Pitfalls**: 
- `Relation.attach()` is a no-op by design (leaf nodes don't accept children). Any code that relies on `transformPlanToAttachChild` after the anonymizer has already attached a Relation will silently fail.
- The anonymizer's `visitAppendCol` mutates the AST (attaches Relation to subsearch) as a side effect. This creates an ordering dependency between the anonymizer and the visitor.

**Things to Watch**:
- If new schema-transforming commands are added (similar to spath), they would need to be added to `buildAppendColSubsearchSource()`.
- The anonymizer's `visitAppendCol` still uses `getRelation` + `transformPlanToAttachChild` — this is fine for anonymization purposes but should be noted if the anonymizer's behavior changes.

PR #5310, opened 2026-04-04
[BugFix] Fix COALESCE(null, 42) returning wrong type string instead of int (#5175)

### LINKED ISSUE #5175: [BUG] COALESCE(null, 42) returns wrong type string instead of int due to leastRestrictive type inference

qianheng-aws:
### Description
Fix COALESCE returning wrong type `string` instead of `int` when called with null and integer literals.

**Root Cause**: In PPL, `null` is parsed as a `QualifiedName` (field name), not a `Literal`. When COALESCE encounters an unresolvable field name like `null`, it falls back to `replaceWithNullLiteralInCoalesce()` which incorrectly created a null literal typed as `VARCHAR`. This caused `leastRestrictive([VARCHAR, INTEGER])` to return `VARCHAR`, making `COALESCE(null, 42)` return `string` type.

**Fix**: Two-part fix:
1. **Primary** (`QualifiedNameResolver.java`): Changed the null literal type from `SqlTypeName.VARCHAR` to `SqlTypeName.NULL` (Calcite's bottom type, compatible with any type). This makes `leastRestrictive([NULL, INTEGER])` correctly return `INTEGER`.
2. **Defense-in-depth** (`EnhancedCoalesceFunction.java`): Filter out `NULL`-typed operands in the return type inference before calling `leastRestrictive`, ensuring correct type even if NULL types slip through.

### Related Issues
Resolves #5175

### Check List
- [x] Javadoc added for new public API
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/QualifiedNameResolver.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/QualifiedNameResolver.java
@@ -317,7 +317,7 @@ public class QualifiedNameResolver {
     if (context.isInCoalesceFunction()) {
       return Optional.of(
           context.rexBuilder.makeNullLiteral(
-              context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.VARCHAR)));
+              context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.NULL)));
     }
     return Optional.empty();
   }
--- a/core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/udf/condition/EnhancedCoalesceFunction.java
@@ -7,6 +7,7 @@ package org.opensearch.sql.expression.function.udf.condition;
 
 import java.util.List;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import org.apache.calcite.adapter.enumerable.NotNullImplementor;
 import org.apache.calcite.adapter.enumerable.NullPolicy;
 import org.apache.calcite.linq4j.tree.Expression;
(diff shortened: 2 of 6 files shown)


qianheng-aws:
## Decision Log
**Root Cause**: In PPL, `null` is not a reserved literal keyword — it's parsed as a `QualifiedName` (field name). When COALESCE encounters the unresolvable field name "null", `QualifiedNameResolver.replaceWithNullLiteralInCoalesce()` creates a null literal typed as `VARCHAR` instead of `NULL`. This causes `leastRestrictive([VARCHAR, INTEGER])` to return `VARCHAR`, making `COALESCE(null, 42)` return `string` type instead of `int`.

**Approach**: Two-part fix:
1. Changed `SqlTypeName.VARCHAR` to `SqlTypeName.NULL` in `replaceWithNullLiteralInCoalesce()`. NULL is Calcite's bottom type — it's compatible with any other type, so `leastRestrictive([NULL, INTEGER])` correctly returns `INTEGER`.
2. Added defense-in-depth in `EnhancedCoalesceFunction.getReturnTypeInference()`: filter out NULL-typed operands before calling `leastRestrictive`, handle the all-NULL case explicitly, and ensure the result is always nullable.

**Alternatives Rejected**:
- Making `null` a reserved keyword in the PPL parser: This would be a much larger change affecting the grammar and many other commands. The current approach of handling unresolvable names in COALESCE is the established pattern.
- Only fixing `EnhancedCoalesceFunction` without the `QualifiedNameResolver` change: This would leave the incorrect `VARCHAR` typing for non-existent fields, which could cause issues in other contexts.

**Pitfalls**:
- Existing tests reference `null:VARCHAR` in their expected logical plans for non-existent fields inside COALESCE. These needed updating to `null:NULL`.
- The PPL response uses `PPL_SPEC.typeName()` which maps INTEGER to `"int"` (not `"integer"`), so integration/YAML test assertions must use `int`.

**Things to Watch**:
- The `nomv` command rewrites to use COALESCE internally. The `testNoMvNonExistentField` test was verified to not be affected because the non-existent field error occurs at the `array_compact` level, before COALESCE.
- The `fillnull` command also uses COALESCE but doesn't reference `null:VARCHAR` in its tests, so no impact.

PR #5315, opened 2026-04-04
[BugFix] Fix COALESCE(null, 42) returning wrong type string instead of int (#5175)

### LINKED ISSUE #5175: [BUG] COALESCE(null, 42) returns wrong type string instead of int due to leastRestrictive type inference

qianheng-aws:
### Description
Fix COALESCE returning wrong type `string` instead of `int` when called with null and integer literals.

**Root Cause**: In PPL, `null` is parsed as a `QualifiedName` (field name), not a `Literal`. When COALESCE encounters an unresolvable field name like `null`, it falls back to `replaceWithNullLiteralInCoalesce()` which incorrectly created a null literal typed as `VARCHAR`. This caused `leastRestrictive([VARCHAR, INTEGER])` to return `VARCHAR`, making `COALESCE(null, 42)` return `string` type.

**Fix (three parts)**:
1. **Primary** (`QualifiedNameResolver.java`): Changed the null literal type from `SqlTypeName.VARCHAR` to `SqlTypeName.NULL` (Calcite's bottom type, compatible with any type). This makes `leastRestrictive([NULL, INTEGER])` correctly return `INTEGER`.
2. **Defense-in-depth** (`EnhancedCoalesceFunction.java`): Filter out `NULL`-typed operands in the return type inference before calling `leastRestrictive`, with fallback to VARCHAR when all operands are NULL.
3. **Scope fix** (`CalciteRexNodeVisitor.java`): Scope the `inCoalesceFunction` context flag to direct COALESCE arguments only, preventing unresolved field names in nested function calls (e.g., `nomv` rewriting to `coalesce(mvjoin(array_compact(field), ...))`) from being silently replaced with null literals.

### Related Issues
Resolves #5175

### Check List
- [x] Javadoc added for new public API
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -499,8 +499,13 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
     List<RexNode> arguments = new ArrayList<>();
 
     boolean isCoalesce = "coalesce".equalsIgnoreCase(node.getFuncName());
+    boolean wasInCoalesce = context.isInCoalesceFunction();
     if (isCoalesce) {
       context.setInCoalesceFunction(true);
+    } else {
+      // Disable the flag for nested non-COALESCE functions so that unresolved field
+      // names inside e.g. array_compact(...) are not silently replaced with null.
+      context.setInCoalesceFunction(false);
     }
 
     List<RexNode> capturedVars = null;
@@ -529,9 +534,8 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
         }
       }
     } finally {
-      if (isCoalesce) {
-        context.setInCoalesceFunction(false);
-      }
+      // Restore the previous COALESCE context state
+      context.setInCoalesceFunction(wasInCoalesce);
     }
 
     // For transform/mvmap functions with captured variables, add them as additional arguments
--- a/core/src/main/java/org/opensearch/sql/calcite/QualifiedNameResolver.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/QualifiedNameResolver.java
@@ -317,7 +317,7 @@ public class QualifiedNameResolver {
     if (context.isInCoalesceFunction()) {
       return Optional.of(
           context.rexBuilder.makeNullLiteral(
-              context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.VARCHAR)));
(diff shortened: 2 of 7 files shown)


qianheng-aws:
## Decision Log
**Root Cause**: `QualifiedNameResolver.replaceWithNullLiteralInCoalesce()` typed null literals as `SqlTypeName.VARCHAR` instead of `SqlTypeName.NULL`. When `COALESCE(null, 42)` was processed, `null` (parsed as a `QualifiedName`) was resolved to `null:VARCHAR`, causing `leastRestrictive([VARCHAR, INTEGER])` to return `VARCHAR` — hence the incorrect `string` output type.

**Approach**: Three-part fix:
1. Use `SqlTypeName.NULL` (Calcite's bottom type) for unresolved field null literals. `NULL` is compatible with any type, so `leastRestrictive([NULL, INTEGER])` correctly returns `INTEGER`.
2. Add defense-in-depth filtering in `EnhancedCoalesceFunction.getReturnTypeInference()` — filter out NULL-typed operands before calling `leastRestrictive`.
3. Scope `inCoalesceFunction` flag in `CalciteRexNodeVisitor.visitFunction()` — save/restore the flag around nested function calls so it only applies to direct COALESCE arguments.

**Alternatives Rejected**:
- *Only fixing `QualifiedNameResolver`*: Would have caused a regression in `CalcitePPLNoMvTest.testNoMvNonExistentField` because `nomv` rewrites to `coalesce(mvjoin(array_compact(field), ...))` and the leaked `inCoalesceFunction` flag would silently replace unresolved fields inside `array_compact` with `null:NULL` (accepted as bottom type) instead of throwing an error.

**Pitfalls**:
- The `inCoalesceFunction` flag was a flat boolean that leaked into nested function calls. With `VARCHAR`, the leak was mostly harmless because `array_compact(null:VARCHAR)` would error on type mismatch. With `NULL` (bottom type), `array_compact(null:NULL)` could silently succeed — hence the need for the scoping fix.
- Existing unit tests referenced `null:VARCHAR` in logical plan expectations — these needed updating to `null:NULL`.

**Things to Watch**: The `inCoalesceFunction` flag design could benefit from a stack-based approach if more nuanced nesting is needed in the future.

Issue #5248, opened 2026-03-18
[FEATURE] SQL query support for Analytics engine integration


dai-chen:
**Is your feature request related to a problem?**

The Analytics engine integration requires ANSI SQL support to enable SQL queries against non-Lucene indices. However, the unified query API currently only supports PPL V3. The PPL integration (tracked in the sibling issue) establishes the query routing, execution engine, and response formatting infrastructure — this issue covers the additional work needed for SQL support.

Technical requirements:

- ANSI SQL queries against non-Lucene indices return correct results through `_plugins/_sql` endpoint
- SQL explain API (`_plugins/_sql/_explain`) returns the logical plan
- OpenSearch full-text search functions (`match`, `match_phrase`, `query_string`) available in ANSI SQL syntax
- Existing V2 SQL queries are unaffected (backward compatible, ANSI SQL is opt-in)
- Same response format, error handling, and observability as the PPL unified path

**What solution would you like?**

Unified query API:

1. **Enable Calcite SQL in unified query API**: Add the Calcite native SQL parser path (`SqlParser` → `SqlValidator` → `SqlToRelConverter` → `RelNode`) in UnifiedQueryPlanner with proper SQL conformance, case sensitivity, and identifier handling
2. **Register full-text search functions**: Register search functions (`match`, `match_phrase`, `query_string`) with same signatures and return types

SQL plugin: 
 
3. **Integrate with SQL REST endpoint**: Add ANSI SQL mode routing in `RestSqlAction` (via mode parameter or auto-detection) through the unified query pipeline established by the PPL issue, with explain support and backward compatibility for existing V2 SQL queries.
4. **Integration tests**: End-to-end ITs verifying ANSI SQL query, explain, full-text search functions, response format, and breaking changes in existing V2 SQL queries.**What alternatives have you considered?**

See parent issue #5246 for the design comparison of Option A (Query Delegation), Option B (Unified Query Pipeline), and Option C (Calcite Schema Adapter).

**Do you have any additional context?**

- PoC PR: [dai-chen/sql-1#10](https://github.com/dai-chen/sql-1/pull/10) — demonstrates ANSI SQL RelNode generation via Calcite's native parser in UnifiedQueryPlanner
- Depends on PPL Analytics engine integration (sibling issue) for query routing, execution engine, and response formatting infrastructure

dai-chen:
# Unified Query API — SQL Gap Analysis Report (Consolidated)

**Date:** 2026-03-30
[...]

PR #5317, opened 2026-04-06
Integrate SQL REST endpoint with analytics engine path

ahkcs:
## Summary
- Add SQL query routing through the analytics engine for Parquet-backed indices
- SQL queries targeting `parquet_*` indices are routed to `RestUnifiedQueryAction` via the unified query pipeline (Calcite SQL parser → UnifiedQueryPlanner → AnalyticsExecutionEngine)
- Non-analytics SQL queries fall through to the existing V2 engine unchanged

## Changes
- Add `SqlNode` table extraction in `RestUnifiedQueryAction.extractIndexName()` for SQL routing
- Add `executeSql()` and `explainSql()` methods in `RestUnifiedQueryAction`
- Add analytics routing in `RestSqlAction` via optional `BiFunction` router
- Wire the router in `SQLPlugin.createSqlAnalyticsRouter()`
- Add `AnalyticsSQLIT` integration tests

Resolves step 3: https://github.com/opensearch-project/sql/issues/5248

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
@@ -18,6 +18,7 @@ import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlJoin;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlSelect;
+import org.apache.calcite.sql.util.SqlBasicVisitor;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.ThreadContext;
@@ -189,8 +190,7 @@ public class RestUnifiedQueryAction {
   }
 
   /** SqlNode visitor that extracts the source table name from a SQL parse tree. */
-  private static class SqlTableNameExtractor
-      extends org.apache.calcite.sql.util.SqlBasicVisitor<String> {
+  private static class SqlTableNameExtractor extends SqlBasicVisitor<String> {
     @Override
     public String visit(SqlCall call) {
       if (call instanceof SqlSelect select) {


integ-test/src/test/java/org/opensearch/sql/sql/AnalyticsSQLExplainIT.java
```diff
@@ -0,0 +1,94 @@
...
+import java.nio.file.Paths;
+import org.junit.Test;
+import org.opensearch.client.Request;
+import org.opensearch.client.RequestOptions;
+import org.opensearch.client.Response;
+import org.opensearch.sql.legacy.SQLIntegTestCase;
+
+/**
+ * Explain integration tests for SQL queries routed through the analytics engine path (Project
+ * Mustang). Validates that SQL queries targeting "parquet_*" indices produce correct logical plans
```
dai-chen: Can we stick with public name "analytics engine" and avoid project codename everywhere?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -182,7 +187,75 @@ private static String extractIndexName(String query, UnifiedQueryContext context
     if (parseResult instanceof UnresolvedPlan unresolvedPlan) {
       return unresolvedPlan.accept(new IndexNameExtractor(), null);
     }
-    // TODO: handle SQL SqlNode for table extraction when unified SQL is enabled
+    if (parseResult instanceof SqlNode sqlNode) {
```
dai-chen: np: pass in `queryType` to avoid instanceof check

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -182,7 +187,75 @@ private static String extractIndexName(String query, UnifiedQueryContext context
...
+      return extractTableNameFromSqlNode(sqlNode);
+    }
+    return null;
```
dai-chen: np: use Optional

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -182,7 +187,75 @@ private static String extractIndexName(String query, UnifiedQueryContext context
...
+  }
+
+  /**
+   * Execute a SQL query through the unified query pipeline. Uses {@link
+   * org.opensearch.sql.plugin.transport.TransportPPLQueryResponse} as the transport response type
+   * since both PPL and SQL share the same JSON response format.
+   */
+  public void executeSql(
```
dai-chen: Why we need separate execute and explain logic for SQL path?

integ-test/src/test/java/org/opensearch/sql/sql/AnalyticsSQLExplainIT.java
```diff
@@ -0,0 +1,94 @@
...
+              + "    \"message\": {\"type\": \"keyword\"},"
+              + "    \"ip_addr\": {\"type\": \"keyword\"}"
+              + "  }"
+              + "}"
+              + "}");
+      client().performRequest(request);
+    }
+  }
+
+  private String explainSqlQuery(String sql) throws IOException {
```
dai-chen: no method in base class to reuse?

integ-test/src/test/java/org/opensearch/sql/sql/AnalyticsSQLExplainIT.java
```diff
@@ -0,0 +1,94 @@
...
+    try {
+      URI uri = Resources.getResource(filename).toURI();
+      return new String(Files.readAllBytes(Paths.get(uri)));
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Test
+  public void testExplainSelectStar() throws IOException {
```
dai-chen, line 64: Could you also check EXPLAIN statement? It's supported by Calcite SQL, e.g., `EXPLAIN PLAN FOR ...` Ref: https://calcite.apache.org/docs/reference.html

legacy/src/main/java/org/opensearch/sql/legacy/plugin/RestSqlAction.java
```diff
@@ -134,14 +150,44 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
           new SQLQueryRequest(
               sqlRequest.getJsonContent(),
               sqlRequest.getSql(),
               request.path(),
               request.params(),
               sqlRequest.cursor());
+
+      // Route to analytics engine for non-Lucene (e.g., Parquet-backed) indices.
+      // The router returns true and sends the response directly if it handled the request.
+      if (analyticsRouter != null) {
```
dai-chen: just wonder analyticsRouter can be null? Is this implementation different from PPL path?

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -175,6 +188,67 @@ public List<RestHandler> getRestHandlers(
...
+   */
+  private BiFunction<SQLQueryRequest, RestChannel, Boolean> createSqlAnalyticsRouter() {
+    RestUnifiedQueryAction unifiedQueryHandler =
+        new RestUnifiedQueryAction(client, clusterService, new StubQueryPlanExecutor());
+    return (sqlRequest, channel) -> {
+      if (!unifiedQueryHandler.isAnalyticsIndex(sqlRequest.getQuery(), QueryType.SQL)) {
+        return false;
+      }
+      if (sqlRequest.isExplainRequest()) {
+        unifiedQueryHandler.explainSql(
```
dai-chen: If we unify the SQL/PPL path by single execute/explain API as commented above, this can be deduplicated?

legacy/src/main/java/org/opensearch/sql/legacy/plugin/RestSqlAction.java
```diff
@@ -134,14 +150,44 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+                        } catch (Exception e) {
+                          handleException(ch, e);
+                        }
+                      },
+                      this::handleException)
+                  .accept(channel);
+            } catch (Exception e) {
+              handleException(channel, e);
+            }
+          }
```
dai-chen: Is this duplicate logic from somewhere?

integ-test/src/test/java/org/opensearch/sql/sql/AnalyticsSQLExplainIT.java
ahkcs: Good point, updated.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Good call, updated. Now pass in `queryType` directly instead of `instanceof` on the parse result.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Updated. `extractIndexName` now returns `Optional<String>`, and `isAnalyticsIndex` uses `.map().orElse(false)` to chain the logic.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Good point, removed the duplicate methods. SQL and PPL now share the same `execute()` and `explain()`. For SQL we pass `profiling=false` and `ExplainMode.STANDARD` since SQLQueryRequest doesn't have these fields yet — should we add them to `SQLQueryRequest` or is this fine for now?

integ-test/src/test/java/org/opensearch/sql/sql/AnalyticsSQLExplainIT.java
ahkcs: Switched to `explainQuery()` from the base class SQLIntegTestCase

ahkcs: Also update `AnalyticsSQLIT`

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -177,12 +169,30 @@ private UnifiedQueryContext buildContext(QueryType queryType, boolean profiling)
...
+    if (queryType == QueryType.PPL) {
+      UnresolvedPlan unresolvedPlan = (UnresolvedPlan) context.getParser().parse(query);
+      return Optional.ofNullable(unresolvedPlan.accept(new IndexNameExtractor(), null));
+    }
+    SqlNode sqlNode = (SqlNode) context.getParser().parse(query);
+    return Optional.ofNullable(extractTableNameFromSqlNode(sqlNode));
+  }
+
+  /** Extracts the table name from a Calcite SqlNode parse tree. */
+  private static String extractTableNameFromSqlNode(SqlNode sqlNode) {
```
dai-chen: Missed this in my previous comment. This is supposed to be visitor too (SqlNode shuttle)?

legacy/src/main/java/org/opensearch/sql/legacy/plugin/RestSqlAction.java
ahkcs: Good point — `analyticsRouter` is always provided. Removed the no-arg constructor and null check

The implementation differs from PPL because `RestSqlAction` is in the legacy module which can't depend on plugin module, so we can't inject `RestUnifiedQueryAction` directly like PPL does in `TransportPPLQueryAction`. The BiFunction bridges the module boundary — it's created in `SQLPlugin` (plugin module) where `RestUnifiedQueryAction` is visible, then passed to `RestSqlAction` (legacy module).

dai-chen: Got it. I think we may want to enable profiling, at least for unified SQL path later. @penghuo

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
ahkcs: Addressed in previous commit,  removed the separate executeSql()/explainSql() methods. SQL and PPL now share the same execute() and explain() API.

legacy/src/main/java/org/opensearch/sql/legacy/plugin/RestSqlAction.java
ahkcs: Checked — the V2 fallback logic only exists in one place (the analytics router's false branch), so there's no actual duplication. Extracted it into `delegateToV2Engine()` for readability

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Refactored to use `SqlBasicVisitor` — `SqlTableNameExtractor` extends `SqlBasicVisitor<String>` with visit(SqlCall) and visit(SqlIdentifier), consistent with how PPL's `IndexNameExtractor` extends `AbstractNodeVisitor<String, Void>`. Both use the `node.accept(visitor)` dispatch pattern.

We used `SqlBasicVisitor` for reading/extracting purpose

integ-test/src/test/java/org/opensearch/sql/sql/AnalyticsSQLExplainIT.java
ahkcs: Investigated — `EXPLAIN PLAN FOR` is parsed by Calcite as a `SqlExplain` node, but `SqlToRelConverter.convertQueryRecursive()` doesn't handle it (throws `AssertionError("not a query: ..."`) in the default case). Calcite treats EXPLAIN as a meta-statement handled at the JDBC layer (CalcitePrepareImpl), not at the `SqlToRelConverter` level.

To support it, we'd need to detect SqlExplain in `CalciteNativeStrategy.plan()`, unwrap the inner query, plan it, and return the plan as a formatted result. However, the resulting plan would be identical to what
  `/_plugins/_sql/_explain` already returns for the same query — the only difference is the response format (`EXPLAIN PLAN FOR` returns the plan as a query result row, `/_explain` returns it as a JSON explain response)
  
 Do we want to support EXPLAIN statement?

ahkcs: Error log:
```
curl -s -X POST localhost:9200/_plugins/_sql -H 'Content-Type: application/json' -d '{"query": "EXPLAIN PLAN FOR SELECT * FROM parquet_logs"}'
{
  "error": {
    "details": "Query must start with SELECT, DELETE, SHOW or DESCRIBE: EXPLAIN PLAN FOR SELECT * FROM parquet_logs",
    "type": "SQLFeatureNotSupportedException"
  },
  "status": 400
}%

dai-chen (approved):

PR #5318, opened 2026-04-06
[Feature] Add table function relation syntax to SQL grammar

mengweieric:
## Summary

Adds **generic table function relation** syntax to the SQL grammar, enabling `FROM funcName(key=value, ...) AS alias` in SQL queries. This is the foundational grammar change for the SQL vectorSearch() feature, but the syntax is not vectorSearch-specific — any table function registered with the storage engine can be invoked through this syntax.

### Changes
- **Grammar:** New `tableFunctionRelation` alternative in the `relation` rule, with `tableFunctionArgs` and `tableFunctionArg` sub-rules for named argument parsing
- **AstBuilder:** `visitTableFunctionRelation()` emits `SubqueryAlias(alias, TableFunction(name, args))` AST nodes. Argument names are canonicalized at the parser boundary (`unquoteIdentifier` + `toLowerCase`) for case-insensitive matching downstream
- **Tests:** 6 parser unit tests covering basic parse, full clause combo (WHERE + ORDER BY + LIMIT with structural AST assertion), argument reordering, case canonicalization, alias-without-AS, and alias-required negative test

## Test plan
- [x] `./gradlew :sql:test --tests "org.opensearch.sql.sql.parser.AstBuilderTest"` — all 51 tests pass
- [x] CI green

### DIFF
--- a/sql/src/main/java/org/opensearch/sql/sql/parser/AstBuilder.java
+++ b/sql/src/main/java/org/opensearch/sql/sql/parser/AstBuilder.java
@@ -201,8 +201,7 @@ public class AstBuilder extends OpenSearchSQLParserBaseVisitor<UnresolvedPlan> {
         .forEach(
             arg -> {
               String argName =
-                  StringUtils.unquoteIdentifier(arg.ident().getText())
-                      .toLowerCase(Locale.ROOT);
+                  StringUtils.unquoteIdentifier(arg.ident().getText()).toLowerCase(Locale.ROOT);
               UnresolvedExpression argValue = visitAstExpression(arg.functionArg());
               args.add(new UnresolvedArgument(argName, argValue));
             });
--- a/sql/src/test/java/org/opensearch/sql/sql/parser/AstBuilderTest.java
+++ b/sql/src/test/java/org/opensearch/sql/sql/parser/AstBuilderTest.java
@@ -168,14 +168,10 @@ class AstBuilderTest extends AstBuilderTestBase {
                                 ImmutableList.of(
                                     new UnresolvedArgument("table", stringLiteral("products")),
                                     new UnresolvedArgument("field", stringLiteral("embedding")),
-                                    new UnresolvedArgument(
-                                        "vector", stringLiteral("[0.1,0.2]")),
-                                    new UnresolvedArgument(
-                                        "option", stringLiteral("k=10"))))),
+                                    new UnresolvedArgument("vector", stringLiteral("[0.1,0.2]")),
+                                    new UnresolvedArgument("option", stringLiteral("k=10"))))),
                         function("=", qualifiedName("s", "category"), stringLiteral("shoes"))),
-                    field(
-                        qualifiedName("s", "_score"),
-                        argument("asc", booleanLiteral(false)))),
+                    field(qualifiedName("s", "_score"), argument("asc", booleanLiteral(false)))),
                 5,
                 0),
             alias("s.title", qualifiedName("s", "title")),
@@ -241,8 +237,7 @@ class AstBuilderTest extends AstBuilderTestBase {
                         new UnresolvedArgument("table", stringLiteral("products")),
                         new UnresolvedArgument("vector", stringLiteral("[0.1]"))))),
             AllFields.of()),
-        buildAST(
-            "SELECT * FROM vectorSearch(table='products', vector='[0.1]') v"));
+        buildAST("SELECT * FROM vectorSearch(table='products', vector='[0.1]') v"));
   }
 
   @Test


sql/src/main/java/org/opensearch/sql/sql/parser/AstBuilder.java
```diff
@@ -189,6 +193,24 @@ public UnresolvedPlan visitSubqueryAsRelation(SubqueryAsRelationContext ctx) {
...
 
+  @Override
+  public UnresolvedPlan visitTableFunctionRelation(TableFunctionRelationContext ctx) {
+    ImmutableList.Builder<UnresolvedExpression> args = ImmutableList.builder();
+    ctx.tableFunctionArgs()
+        .tableFunctionArg()
+        .forEach(
+            arg -> {
+              String argName =
+                  StringUtils.unquoteIdentifier(arg.ident().getText()).toLowerCase(Locale.ROOT);
```
dai-chen, line 204: Does `visitAstExpression` (visitIdent) handle this already?

sql/src/main/antlr/OpenSearchSQLParser.g4
```diff
@@ -111,6 +111,15 @@ fromClause
 relation
    : tableName (AS? alias)?                                         # tableAsRelation
    | LR_BRACKET subquery = querySpecification RR_BRACKET AS? alias  # subqueryAsRelation
+   | qualifiedName LR_BRACKET tableFunctionArgs RR_BRACKET AS? alias # tableFunctionRelation
+   ;
+
+tableFunctionArgs
```
dai-chen, line 117: Just to confirm: table function requires at least one argument?

mengweieric: Yes. it is intentional for this PR’s scope. The grammar is tableFunctionArg (COMMA tableFunctionArg)*, so table-function relations currently require one-or-more named arguments. We can relax that to zero-or-more later if we add a zero-arg table-function use case, and that would be a backward-compatible grammar expansion.

sql/src/main/java/org/opensearch/sql/sql/parser/AstBuilder.java
mengweieric: No. visitIdent() only unquotes identifiers; it does not lowercase them. In this path, the arg name is stored in UnresolvedArgument, and ExpressionAnalyzer.visitUnresolvedArgument() preserves that name when creating NamedArgumentExpression. The explicit toLowerCase(Locale.ROOT) in visitTableFunctionRelation() is therefore what canonicalizes named arg keys for table-function relations.

RyanL1997: Thanks @mengweieric  — that confirms the lowercase is the only thing canonicalizing arg keys on this path. My concern isn't *whether* it's needed, but *where* it lives.

Since `UnresolvedArgument` → `NamedArgumentExpression` is a shared pipeline that PPL also feeds (PPL's [`visitTableFunction`](https://github.com/opensearch-project/sql/blob/main/ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java#L1091) doesn't lowercase, it just uses `arg.ident().getText()`), we now have two front-ends with different casing policies converging on the same `TableFunctionResolver`. The existing Prometheus resolver compares names with a case-sensitive `HashSet.removeAll` ([`TableFunctionUtils.java#L73-L89`](https://github.com/opensearch-project/sql/blob/main/prometheus/src/main/java/org/opensearch/sql/prometheus/utils/TableFunctionUtils.java#L73-L89)), so a future `vectorSearch` resolver that declares args in camelCase (`topK`, `filterType`) will silently work in PPL and fail in SQL with `Missing arguments:[topK,filterType]`.

Two ways to keep them in sync — what's your preference?

1. Move the `toLowerCase` into [`ExpressionAnalyzer.visitUnresolvedArgument`](https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/analysis/ExpressionAnalyzer.java#L459-L461) (or `Analyzer.visitTableFunction`) so both parsers get the same canonicalization for free.
2. Keep it here, but also apply it in PPL's `visitTableFunction` and add a javadoc on [`TableFunction.java`](https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/ast/tree/TableFunction.java) documenting the contract so future resolver authors know to declare arg names in lowercase.

Not a blocker tbh — I was trying to avoid the "works in PPL, fails in SQL" footgun before vectorSearch lands on top of this.

mengweieric: My preference would be to centralize table-function arg canonicalization in Analyzer.visitTableFunction rather than ExpressionAnalyzer, so both front-ends share one table-function-specific policy without changing all named-arg handling globally. For this PR I kept the fix local to the SQL parser to minimize scope.

Issue #5246, opened 2026-03-18
[FEATURE] SQL/PPL language support for Analytics engine integration


dai-chen:
**Is your feature request related to a problem?**

The SQL/PPL plugin currently only supports querying Lucene-backed OpenSearch indices. A new Analytics engine introduces Parquet as an alternate storage format with an embedded query engine (DataFusion) for log analytics workloads. Users expect to query Parquet-backed indices using the same `_plugins/_sql` and `_plugins/_ppl` REST endpoints with the same SQL/PPL language support, but the current plugin has no integration path to route queries to the Analytics engine for execution.

### Technical Requirements

- **UX consistency**: Same REST endpoints, response format, error messages, and explain output for both Lucene and non-Lucene queries
- **Engineering complexity**: Minimal changes to existing SQL/PPL plugin code; reuse existing parsing, formatting, and metrics infrastructure
- **Plugin coupling**: Clean API boundary between SQL/PPL plugin and Analytics engine; each can evolve independently
- **Performance**: No unnecessary performance overhead or data copying in the query handoff path
- **Resource management**: Query size limits, circuit breakers, and timeouts enforced on the unified path
- **Monitoring/Observability**: Request/failure metrics, query profiling, and logging for the unified path consistent with existing paths
- **Security**: Index-level authorization during routing; security context propagated to execution thread
- **Testability**: End-to-end integration tests with the real Analytics engine plugin loaded in the test cluster
- **Long-term alignment**: Architecture supports future ANSI SQL, cross-engine JOINs, and additional storage engines

**What solution would you like?**

### High-level Design

The SQL/PPL plugin integrates with the Analytics engine through a new Unified Query Pipeline: the plugin handles query routing, parsing, and generating engine-agnostic Calcite `RelNode` logical plans, then hands them off to the Analytics engine for physical optimization, distributed planning, and execution.

<img width="1584" height="691" alt="Image" src="https://github.com/user-attachments/assets/3f7922d0-9dcd-403c-b3c2-6ce6b2c43989" />

Here is the high-level workflow:

1. `RestSqlAction`/`RestPPLQueryAction` routes queries targeting non-Lucene indices to a new `RestUnifiedQueryAction`; Lucene queries fall through to the existing V2/V3 path unchanged.
2. `RestUnifiedQueryAction` builds a Calcite schema from cluster state via the Analytics engine's `OpenSearchSchemaBuilder` (standard SQL types, no UDTs).
[...]

PR #5320, opened 2026-04-07
[Feature] Add vector search execution pipeline for vectorSearch() table function

mengweieric:
## Summary

Adds the current OpenSearch/V2 SQL execution pipeline for `vectorSearch()` table functions
on the `feature/vector-search-p0` branch. This is the execution companion to #5318, which
adds SQL relation syntax for table functions.

## What this PR adds

- **`VectorSearchTableFunctionResolver`**
  - Registers `vectorsearch` on the OpenSearch storage-engine table-function path
  - Defines the 4-argument surface: `table`, `field`, `vector`, `option`

- **`VectorSearchTableFunctionImplementation`**
  - Parses named arguments and rejects null/empty arg names defensively
  - Parses vector literals into `float[]`
  - Parses the comma-separated `option` string
  - Rejects malformed option segments
  - Rejects duplicate option keys
  - Rejects unknown option keys for P0
  - Canonicalizes P0 option values to numeric strings before DSL generation
  - Rejects empty, malformed, and non-finite vector components
  - Validates field names before interpolation into `WrapperQueryBuilder` JSON

- **`VectorSearchIndex`**
  - Seeds scans with a k-NN query via `WrapperQueryBuilder`
  - Enables score tracking so `_score` is available
  - Defaults top-k queries to `size = k` when `k` is present
  - Uses `VectorSearchQueryBuilder` for score-preserving filter pushdown

- **`VectorSearchQueryBuilder`**
  - Keeps the k-NN query in `bool.must` (scoring context)
  - Places SQL `WHERE` predicates in `bool.filter` (non-scoring context)

- **`OpenSearchDataType`**
  - Recognizes `knn_vector` and maps it to `ExprCoreType.ARRAY` as a P0 visibility shim

- **`OpenSearchStorageEngine`**
  - Registers the vectorsearch resolver in `getFunctions()`

## Scope note

This PR adds `vectorSearch()` on the current OpenSearch/V2 SQL execution path and
builds on #5318. It does not add ANSI SQL / Calcite / UnifiedQueryPlanner support.
Compatibility with the Analytics-engine unified query path is future work related
to #5248 / #5246.

The resolver is registered on the shared storage-engine table-function path via
`OpenSearchStorageEngine.getFunctions()`. PPL cannot currently reach this because
the PPL grammar's `tableSource` rule does not reference `tableFunction` — the rule
exists but is dead grammar (zero references from `fromClause`). A null-arg-name
guard is included defensively in case PPL wires up table functions in the future.

The `knn_vector` → `ARRAY` type shim is conceptually compatible with #5246's
API-contract direction of standard Calcite types without UDTs.

## Not in this PR (deferred to follow-up PRs in the feature stack)

- Mutual exclusivity validation for `k` / `max_distance` / `min_score`
- `k` range validation (1–10,000)
- Radial-query `LIMIT` enforcement
- `LIMIT > k` semantics
- Engine/method compatibility checks
- Field/type/dimension validation
- Migration from hand-built JSON to structured XContent builders
- Integration tests

## Next PRs

- **PR-3 (Validation Hardening):** Enforce exactly-one search mode (mutual exclusivity),
  k range [1, 10000], radial-without-LIMIT rejection, LIMIT > k rejection, unsupported
  shape rejection (ORDER BY non-score, GROUP BY, JOIN), integration tests
- **PR-4 (Efficient Filtering):** `filter_type=efficient|post` option, WHERE inside knn
  clause for efficient pre-filtering mode

## Test plan

- `./gradlew :opensearch:test --tests "*VectorSearchTableFunctionResolverTest"`
- `./gradlew :opensearch:test --tests "*VectorSearchTableFunctionImplementationTest"`
- `./gradlew :opensearch:test --tests "*VectorSearchQueryBuilderTest"`
- `./gradlew :opensearch:test --tests "*OpenSearchStorageEngineTest"`
- `./gradlew :opensearch:build`

## New unit coverage in this PR

- Unknown option key rejection
- Malformed option segment rejection
- Duplicate option key rejection
- Invalid field-name rejection
- Nested field-name acceptance
- Empty / malformed / non-finite vector rejection
- Non-numeric option value rejection
- Named-argument enforcement (non-named args and null arg names)
- `pushDownFilter()` preserving knn in scoring context
- Storage engine contains-check (not brittle exact-size assertion)

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
@@ -53,6 +53,13 @@ public class VectorSearchIndex extends OpenSearchIndex {
     var queryBuilder = new VectorSearchQueryBuilder(requestBuilder, buildKnnQuery());
     requestBuilder.pushDownTrackedScore(true);
 
+    // Top-k mode: default size to k so queries without LIMIT return k results
+    // instead of falling into the generic large-scan path.
+    // LIMIT pushdown will further reduce this if present.
+    if (options.containsKey("k")) {
+      requestBuilder.pushDownLimitToRequestTotal(Integer.parseInt(options.get("k")), 0);
+    }
+
     Function<OpenSearchRequestBuilder, OpenSearchIndexScan> createScanOperator =
         rb ->
             new OpenSearchIndexScan(
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.storage;
+
+import java.util.Map;
+import java.util.function.Function;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.WrapperQueryBuilder;
+import org.opensearch.sql.common.setting.Settings;
+import org.opensearch.sql.opensearch.client.OpenSearchClient;
+import org.opensearch.sql.opensearch.request.OpenSearchRequestBuilder;
+import org.opensearch.sql.opensearch.storage.scan.OpenSearchIndexScan;
+import org.opensearch.sql.opensearch.storage.scan.OpenSearchIndexScanBuilder;
+import org.opensearch.sql.opensearch.storage.scan.VectorSearchQueryBuilder;
+import org.opensearch.sql.storage.read.TableScanBuilder;
+
+/**
+ * Vector-search-aware OpenSearch index. Seeds the scan with a knn query and enables score tracking.
+ */
+public class VectorSearchIndex extends OpenSearchIndex {
+
+  private static final String VECTOR_OPTION = "vector";
+
+  private final String field;
+  private final float[] vector;
+  private final Map<String, String> options;
+
+  public VectorSearchIndex(
+      OpenSearchClient client,
+      Settings settings,
+      String indexName,
+      String field,
+      float[] vector,
+      Map<String, String> options) {
+    super(client, settings, indexName);
+    this.field = field;
+    this.vector = vector;
+    this.options = options;
+  }
+
+  @Override
+  public TableScanBuilder createScanBuilder() {
+    final TimeValue cursorKeepAlive =
+        getSettings().getSettingValue(Settings.Key.SQL_CURSOR_KEEP_ALIVE);
+    var requestBuilder = createRequestBuilder();
+
+    // Use VectorSearchQueryBuilder to keep knn in must (scoring) context.
+    // WHERE filters will be placed in filter (non-scoring) context.
+    var queryBuilder = new VectorSearchQueryBuilder(requestBuilder, buildKnnQuery());
+    requestBuilder.pushDownTrackedScore(true);
+
+    Function<OpenSearchRequestBuilder, OpenSearchIndexScan> createScanOperator =
+        rb ->
+            new OpenSearchIndexScan(
+                getClient(),
+                rb.getMaxResponseSize(),
+                rb.build(getIndexName(), cursorKeepAlive, getClient(), getFieldTypes().isEmpty()));
+    return new OpenSearchIndexScanBuilder(queryBuilder, createScanOperator);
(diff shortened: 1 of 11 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
```diff
@@ -0,0 +1,107 @@
...
+import org.opensearch.sql.opensearch.request.OpenSearchRequestBuilder;
+import org.opensearch.sql.opensearch.storage.scan.OpenSearchIndexScan;
+import org.opensearch.sql.opensearch.storage.scan.OpenSearchIndexScanBuilder;
+import org.opensearch.sql.opensearch.storage.scan.VectorSearchQueryBuilder;
+import org.opensearch.sql.storage.read.TableScanBuilder;
+
+/**
+ * Vector-search-aware OpenSearch index. Seeds the scan with a knn query and enables score tracking.
+ */
+public class VectorSearchIndex extends OpenSearchIndex {
```
dai-chen, line 24: Instead of separate `VectorSearchQueryBuilder` and `VectorSearchIndex`, just wonder can we abstract vector search translation logic by `LuceneQuery` interface?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchTableFunctionImplementation.java
```diff
@@ -0,0 +1,246 @@
...
+      if (!Float.isFinite(vector[i])) {
+        throw new ExpressionEvaluationException(
+            String.format(
+                "Invalid vector component '%s': must be a finite number", parts[i].trim()));
+      }
+    }
+    return vector;
+  }
+
+  static Map<String, String> parseOptions(String optionStr) {
```
dai-chen, line 128: Any idea to simplify the parse/validate logic? Probably double check if some is already guarded by grammar or if it can be done by existing visit method.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchTableFunctionImplementation.java
```diff
@@ -0,0 +1,246 @@
...
+    for (String key : options.keySet()) {
+      if (!ALLOWED_OPTION_KEYS.contains(key)) {
+        throw new ExpressionEvaluationException(
+            String.format("Unknown option key '%s'. Supported keys: %s", key, ALLOWED_OPTION_KEYS));
+      }
+    }
+    boolean hasK = options.containsKey("k");
+    boolean hasMaxDistance = options.containsKey("max_distance");
+    boolean hasMinScore = options.containsKey("min_score");
+    if (!hasK && !hasMaxDistance && !hasMinScore) {
```
penghuo, line 196: Missing test for "no required option provided" validation path.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
mengweieric: Good point. I split it this way because the two classes currently own different seams: `VectorSearchIndex` seeds the scan/request state (knn query + track_scores + size=k default) while `VectorSearchQueryBuilder` handles score-preserving filter pushdown. I agree a shared abstraction could make sense if we see a second consumer or want a cleaner bridge for the unified SQL path, but I'd prefer not to turn PR-2 into a broader refactor unless there's an existing LuceneQuery contract that fits both responsibilities cleanly.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchTableFunctionImplementation.java
mengweieric: I checked the grammar/visitor angle. We can lean on parser/resolver for the outer function shape and arity, but option still reaches this implementation as a single ```STRING``` argument, so the inner ```k=...,max_distance=...``` content is not visible to grammar/visit methods. The remaining checks like malformed segments, duplicate keys, unknown keys, and numeric parsing are semantic validation at this layer.

I can simplify the code shape, though: for example by extracting the option parsing/validation into a small helper/value object like VectorSearchOptions.parse(optionStr), so this class reads more like orchestration.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchTableFunctionImplementation.java
mengweieric: Added `testNoRequiredOptionThrows` in a33e68c.

penghuo (approved):

PR #5322, opened 2026-04-08
[BugFix] Fix multiple appendpipe error while revisiting parent AST

### LINKED ISSUE #5173: [BUG] Double appendpipe causes 'different planner' assertion due to AST re-visitation
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-numbers | appendpipe [stats avg(value) as avg_val] | appendpipe [stats max(value) as max_val]
```
**Expected Result:**
Query should execute and return appended aggregation output without planner failure.

**Actual Result:**
HTTP 400 assertion error:
`Relational expression ... belongs to a different planner than is currently being used.`

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "ts": { "type": "date" }
    }
  }
}
```
**Sample Data**
```json
[
  { "count": 2, "category": "A", "subcategory": "Y", "value": 20.3, "ts": "2024-01-02" },
  { "count": 10, "category": "B", "subcategory": "X", "value": 100.0, "ts": "2024-01-03" }
]
```

## Bug Description
**Issue Summary:**
Chained `appendpipe` triggers planner mismatch assertion.

**Steps to Reproduce:**
1. Create index and insert sample rows.
2. Run the query above.
3. Observe HTTP 400 assertion about different planner.

**Impact:**
Multi-stage appendpipe transformations fail and break query pipelines.

songkant-aws:
### Description
Chained appendpipe queries can produce literal-only projections during prepare-time field trimming. Calcite may simplify those projections into Values, which can trigger planner mismatch assertions during execution. Preserve the Project shape for this narrow case in OpenSearchRelFieldTrimmer.

### Related Issues
Resolves #5173 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchRelFieldTrimmer.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchRelFieldTrimmer.java
@@ -5,21 +5,31 @@
 
 package org.opensearch.sql.calcite.utils;
 
+import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import org.apache.calcite.linq4j.Ord;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Aggregate;
+import org.apache.calcite.rel.core.CorrelationId;
+import org.apache.calcite.rel.core.Project;
+import org.apache.calcite.rel.core.Values;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexPermuteInputsShuttle;
+import org.apache.calcite.rex.RexSubQuery;
+import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.rex.RexVisitor;
 import org.apache.calcite.sql.validate.SqlValidator;
 import org.apache.calcite.sql2rel.RelFieldTrimmer;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.mapping.Mapping;
+import org.apache.calcite.util.mapping.MappingType;
 import org.apache.calcite.util.mapping.Mappings;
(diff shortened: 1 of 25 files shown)


core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchRelFieldTrimmer.java
```diff
@@ -30,9 +40,94 @@
...
+        RelOptUtil.permute(project.getCluster().getTypeFactory(), rowType, mapping);
+
+    if (shouldAvoidSimplifyValues(newProjects, newInput)) {
+      return result(
+          project.copy(project.getTraitSet(), newInput, newProjects, newRowType), mapping, project);
+    }
+
+    openSearchRelBuilder.push(newInput);
+    openSearchRelBuilder.project(newProjects, newRowType.getFieldNames(), false, correlationIds);
+    return result(openSearchRelBuilder.build(), mapping, project);
```
songkant-aws: In trimFields(Project, ...), the old openSearchRelBuilder.project(...) call could trigger Calcite's internal simplifyValues() when trimming produced a literal-only projection over a fixed-row-count input. That rewrite could replace the Project with Values, which is the shape that led to the planner mismatch in chained appendpipe.

LantaoJin (dismissed):

qianheng-aws:
As discussed offline, it's root cause is inconsistent planner/cluster are being used in trim phase and optimize phase.

In original calcite, one prepare statement will be used through the whole process. While in PPL, we create 2 different prepares for analyzing and executing(including trimming, optimizing and planning, only trimming uses the planner from this new created prepare while others reuse the planner in RelNode).

The ideal solution should be reusing the same planner for trimming as well.

songkant-aws:
@qianheng-aws As suggested, prepare-time trimming now reuses the incoming RelNode's cluster so it stays aligned with the analyzed plan. The change also preserves pre-registered hint strategies required by PPL custom aggregate hints. If we find other prepare-time state that also needs to stay aligned with the incoming RelNode, we can address those cases incrementally.

qianheng-aws (approved):

qianheng-aws:
LGTM

LantaoJin (approved):

LantaoJin:
@songkant-aws the multiple `LogicalUnion`s could be optimized as discussed offline?

songkant-aws:
@LantaoJin I checked Calcite already has some optimization rules for multiple LogicalUnions. Sometimes it can merge multiple unions into one.

PR #5324, opened 2026-04-08
[Feature] Add validation hardening, sort/limit restrictions, and integration tests for vectorSearch()

mengweieric:
## Summary

Adds validation hardening, query restrictions, and integration tests on top of the
core vectorSearch() execution pipeline in #5320.

### What this PR adds

**Validation hardening**
- Option mutual exclusivity: exactly one of `k`, `max_distance`, `min_score`
- `k` range validation: `[1, 10000]`
- `LIMIT > k` rejection in top-k mode
- Sort restriction: only `ORDER BY _score DESC` allowed; rejects non-`_score`
  fields and `_score ASC`
- Radial size policy: caps `max_distance`/`min_score` queries at `maxResultWindow`
  to prevent unbounded result sets

**pushDownSort() contract preservation**
- Preserves base `OpenSearchIndexScanQueryBuilder.pushDownSort()` contract for
  `sort.getCount()` → limit pushdown. SQL always sets `count=0`, but PPL or future
  callers may combine sort+limit in one `LogicalSort` node. Focused test verifies
  `LogicalSort(count=7)` is correctly pushed down as request size.

**Test coverage**
- `VectorSearchIndexTest` (9 tests): knn JSON generation, options rendering,
  non-numeric option quoting, nested fields
- `VectorSearchIT` (16 integration tests): 4 `_explain` DSL shape verification +
  8 validation error tests + 3 ORDER BY restriction tests + 1 LIMIT success test
- Additional unit tests across existing test files: multi-sort rejection,
  case-insensitive arg lookup, sort count preservation, radial LIMIT modes,
  resolver argument count edge cases, NaN/Infinity edge cases

**Benchmark results (against live OpenSearch + k-NN)**
- Semantic correctness: **10/10** on all 8 test cases (top-k, radial, post-filter, hotels)
- SQL overhead: +1.4ms to +2.6ms p50 vs native DSL

## Test plan
- [x] `./gradlew spotlessCheck` — PASS
- [x] `./gradlew :opensearch:test :core:test :sql:test` — PASS
- [x] `./gradlew :integ-test:integTest -Dtests.class="*VectorSearchIT"` — 16/16 PASS
- [x] Correctness benchmark: 10/10 semantic match across A1, A2, A4, A5, A6a, A6b, A7-tight, A8

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
@@ -202,4 +202,73 @@ public class VectorSearchIT extends SQLIntegTestCase {
 
     assertThat(ex.getMessage(), containsString("Missing required option"));
   }
+
+  // ── Sort restriction validation ─────────────────────────────────────────
+
+  @Test
+  public void testOrderByScoreDescExplainSucceeds() throws IOException {
+    String explain =
+        explainQuery(
+            "SELECT v._id, v._score "
+                + "FROM vectorSearch(table='"
+                + TEST_INDEX
+                + "', field='embedding', "
+                + "vector='[1.0, 2.0]', option='k=5') AS v "
+                + "ORDER BY v._score DESC "
+                + "LIMIT 5");
+
+    assertTrue(
+        "Explain should succeed with ORDER BY _score DESC:\n" + explain,
+        explain.contains("wrapper"));
+  }
+
+  @Test
+  public void testOrderByNonScoreFieldRejects() throws IOException {
+    ResponseException ex =
+        expectThrows(
+            ResponseException.class,
+            () ->
+                executeQuery(
+                    "SELECT v._id FROM vectorSearch(table='"
+                        + TEST_INDEX
+                        + "', field='embedding', "
+                        + "vector='[1.0, 2.0]', option='k=5') AS v "
+                        + "ORDER BY v.firstname ASC "
+                        + "LIMIT 5"));
+
+    assertThat(ex.getMessage(), containsString("unsupported sort expression"));
+  }
+
+  @Test
+  public void testOrderByScoreAscRejects() throws IOException {
+    ResponseException ex =
+        expectThrows(
+            ResponseException.class,
+            () ->
+                executeQuery(
+                    "SELECT v._id FROM vectorSearch(table='"
+                        + TEST_INDEX
+                        + "', field='embedding', "
+                        + "vector='[1.0, 2.0]', option='k=5') AS v "
+                        + "ORDER BY v._score ASC "
+                        + "LIMIT 5"));
+
+    assertThat(ex.getMessage(), containsString("_score ASC is not supported"));
+  }
+
+  // ── LIMIT validation ───────────────────────────────────────────────────
+
+  @Test
+  public void testExplainLimitWithinKSucceeds() throws IOException {
+    String explain =
+        explainQuery(
+            "SELECT v._id, v._score "
+                + "FROM vectorSearch(table='"
+                + TEST_INDEX
+                + "', field='embedding', "
+                + "vector='[1.0, 2.0]', option='k=10') AS v "
+                + "LIMIT 5");
+
+    assertTrue("Explain should succeed with LIMIT <= k:\n" + explain, explain.contains("wrapper"));
+  }
 }
--- a/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
@@ -1,0 +1,205 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.sql;
+
+import static org.hamcrest.Matchers.containsString;
+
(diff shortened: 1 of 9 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchTableFunctionImplementation.java
```diff
@@ -197,9 +197,20 @@ private void validateOptions(Map<String, String> options) {
...
+    int modeCount = (hasK ? 1 : 0) + (hasMaxDistance ? 1 : 0) + (hasMinScore ? 1 : 0);
+    if (modeCount > 1) {
+      throw new ExpressionEvaluationException(
+          "Only one of k, max_distance, or min_score may be specified");
+    }
     // Parse and canonicalize numeric values — closes JSON injection via option values
     if (hasK) {
       parseIntOption(options, "k");
+      int k = Integer.parseInt(options.get("k"));
+      if (k < 1 || k > 10000) {
```
mengweieric, line 210: source of k in [1, 10,000]: https://docs.opensearch.org/latest/vector-search/vector-search-techniques/approximate-knn/#:~:text=The%20maximum%20value%20of%20k%20is%2010%2C000.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
```diff
@@ -44,4 +54,44 @@ public boolean pushDownFilter(LogicalFilter filter) {
...
+      if (sortItem.getLeft().getSortOrder() != Sort.SortOrder.DESC) {
+        throw new ExpressionEvaluationException(
+            "vectorSearch only supports ORDER BY _score DESC; _score ASC is not supported");
+      }
+    }
+    // _score DESC is the natural knn order — no need to push the sort itself to OpenSearch.
+    // Preserve the parent's sort.getCount() → limit pushdown contract: SQL always sets count=0,
+    // but PPL or future callers may set a non-zero count to combine sort+limit in one node.
+    if (sort.getCount() != 0) {
+      requestBuilder.pushDownLimit(sort.getCount(), 0);
```
RyanL1997, line 88: nit:

`pushDownSort` calls `requestBuilder.pushDownLimit(sort.getCount(), 0)` directly, bypassing `this.pushDownLimit()` (L59-67) which has the `LIMIT > k` validation.

I understand vectorSearch is SQL-only for P0, and in the SQL path `sort.getCount()` is always 0 so this is a no-op today. But the comment on L90-91 explicitly anticipates the PPL path ("PPL or future callers may set a non-zero count") — so just want to confirm: if/when PPL does use this path with a non-zero count, should the `LIMIT > k` guard still apply? Or is it intentionally unchecked here?

RyanL1997 (commented):
Hi @mengweieric , thanks for the change and I just left a quesiton.

integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
```diff
@@ -0,0 +1,318 @@
...
+    loadIndex(Index.ACCOUNT);
+  }
+
+  private static final String TEST_INDEX = TestsConstants.TEST_INDEX_ACCOUNT;
+
+  // ── DSL shape verification via _explain ───────────────────────────────
+
+  @Test
+  public void testExplainTopKProducesKnnQuery() throws IOException {
+    String explain =
```
ahkcs: Can we move the `Explain` test to a separate `ExplainIT` suite so that we can view the results in a yaml or json file directly
ref: CalciteExplainIT

integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
```diff
@@ -0,0 +1,318 @@
...
+
+    assertTrue("Explain should contain bool query:\n" + explain, explain.contains("bool"));
+    assertTrue(
+        "Explain should contain must clause (knn in scoring context):\n" + explain,
+        explain.contains("must"));
+    assertTrue(
+        "Explain should contain filter clause (WHERE in non-scoring context):\n" + explain,
+        explain.contains("filter"));
+  }
+
```
ahkcs: ditto for the explain ITs above

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
mengweieric: Good catch. The sort.getCount() path was not preserving the same LIMIT <= k invariant as pushDownLimit(). I fixed that by routing both paths through the same validation helper and added tests for the sort-count case. Thanks for calling it out.

integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
mengweieric: Thanks, done. The _explain cases now live in VectorSearchExplainIT, and VectorSearchIT keeps only behavior/validation tests. I kept the current SQL explain assertions for now; happy to follow up with file-backed expected outputs if we want to standardize that too.

ahkcs (approved):

RyanL1997 (approved):

PR #5330, opened 2026-04-09
Add whitelist to block non-query statements in unified SQL path

dai-chen:
### Description

The unified SQL path uses Calcite's `SqlParser.parseQuery()` which, despite its name, accepts any SQL statement type including DML and DDL. The SQL V2 grammar restricts input to `SELECT`/`SHOW`/`DESCRIBE` at the parser level, but the unified path had no equivalent protection.

## Changes

Validate the parsed `SqlNode` against `SqlKind.QUERY` in `CalciteNativeStrategy.plan()` before proceeding to planning. Non-query statements are rejected with `UnsupportedOperationException`.

| Statement type | SQL V2/Legacy | Unified SQL | Blocked by |
|---|---|---|---|
| SELECT (with JOIN, subquery, etc.) | ✅ | ✅ Allowed | — |
| UNION, EXCEPT, INTERSECT | ✅ UNION/EXCEPT | ✅ Allowed | — |
| CTE (WITH), VALUES | ❌ | ✅ Allowed | — |
| DDL (CREATE MV, etc.) | ❌ | ❌ Blocked | Calcite parser |
| SHOW / DESCRIBE | ✅ | ❌ Blocked | Calcite parser |
| EXPLAIN | Separate `/_explain` path | ❌ Blocked | Whitelist |
| INSERT, DELETE, UPDATE, MERGE | ❌ | ❌ Blocked | Whitelist |

> Note: EXPLAIN, SHOW, and DESCRIBE have separate execution paths in the current SQL plugin and are out of scope for the unified query API — they should be handled by the API caller for now.

### Related Issues
Part of https://github.com/opensearch-project/sql/issues/5248

### Check List

### DIFF
--- a/api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
+++ b/api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
@@ -14,6 +14,7 @@ import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.RelRoot;
 import org.apache.calcite.rel.core.Sort;
 import org.apache.calcite.rel.logical.LogicalSort;
+import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.tools.Frameworks;
 import org.apache.calcite.tools.Planner;
@@ -60,8 +61,7 @@ public class UnifiedQueryPlanner {
   public RelNode plan(String query) {
     try {
       return context.measure(ANALYZE, () -> strategy.plan(query));
-    } catch (SyntaxCheckException e) {
-      // Re-throw syntax error without wrapping
+    } catch (SyntaxCheckException | UnsupportedOperationException e) {
       throw e;
     } catch (Exception e) {
       throw new IllegalStateException("Failed to plan query", e);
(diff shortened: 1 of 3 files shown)


ahkcs (approved):

api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
```diff
@@ -82,6 +82,11 @@ private static class CalciteNativeStrategy implements PlanningStrategy {
     public RelNode plan(String query) throws Exception {
       try (Planner planner = Frameworks.getPlanner(context.getPlanContext().config)) {
         SqlNode parsed = planner.parse(query);
+        if (!parsed.isA(SqlKind.QUERY)) {
+          throw new UnsupportedOperationException(
+              "Only query statements are supported. Got: " + parsed.getKind());
```
Swiddis, line 87: nit (clarity): Would a user know what this means?

Intuitively I'm guessing "query" means "non-modifying & non-explain," so maybe something like "Only read-only querying statements are supported, got an {explain} query" with some case-based substitution.

Swiddis (approved):

api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
dai-chen: Good point on clarity. But I think EXPLAIN, SHOW, and DESCRIBE are also read-only and not supported here either? I'll give more careful thought for the wording in this new api module later. Thanks!

PR #5331, opened 2026-04-09
[Feature] Add efficient filtering (knn.filter) support for vectorSearch()

mengweieric:
## Summary

Adds `filter_type=post|efficient` option to `vectorSearch()` so WHERE clauses can be placed inside the knn clause (`knn.filter`) for efficient pre-filtering during ANN search, or outside as `bool.filter` for post-filtering (default). Also adds mandatory LIMIT enforcement for radial search.

### What this PR adds

**FilterType enum and option parsing**
- New `FilterType` enum (`POST`, `EFFICIENT`) with `fromString()` validation
- `filter_type` added to allowed option keys in `VectorSearchTableFunctionImplementation`
- `filter_type` is stripped from options before knn JSON generation — it's a SQL-layer directive, not a knn parameter

**Efficient filter pushdown**
- `VectorSearchQueryBuilder.pushDownFilter()` branches on filter type:
  - `POST` (default): knn in `bool.must` + WHERE in `bool.filter` (post-filtering)
  - `EFFICIENT`: rebuilds knn query with WHERE embedded in `knn.filter` via callback
- `Function<QueryBuilder, QueryBuilder>` callback keeps JSON serialization in `VectorSearchIndex`
- `buildKnnQueryJson()` collapsed to accept optional filter JSON parameter — no duplication

**Build-time validation**
- `build()` override rejects explicit `filter_type` when no filter is pushed down (either no WHERE clause at all, or the WHERE clause was not pushdownable)
- `pushDownFilter()` catches `ScriptQueryUnSupportedException` for non-pushdownable conditions:
  - With explicit `filter_type`: throws a clear error explaining the condition cannot be pushed down
  - Without explicit `filter_type`: returns `false` to fall back to in-memory filtering, matching base class behavior

**Radial search LIMIT requirement**
- Radial search (`max_distance` or `min_score`) without an explicit `LIMIT` clause is rejected at build time with a clear error message
- Prevents unbounded result sets from radial queries that could silently return up to `maxResultWindow` rows

**Engine support**
- `knn.filter` is supported for lucene and faiss engines (HNSW, IVF). Engine compatibility is not validated by the SQL plugin — unsupported engines reject at execution time.

### SQL syntax

```sql
-- Post-filtering (default, same as omitting filter_type)
SELECT v._id, v._score
FROM vectorSearch(table='my-index', field='embedding',
     vector='[1.0, 2.0, 3.0]', option='k=10,filter_type=post') AS v
WHERE v.city = 'Seattle'
LIMIT 10

-- Efficient pre-filtering (WHERE inside knn.filter)
SELECT v._id, v._score
FROM vectorSearch(table='my-index', field='embedding',
     vector='[1.0, 2.0, 3.0]', option='k=10,filter_type=efficient') AS v
WHERE v.city = 'Seattle'
LIMIT 10

-- Radial search requires LIMIT
SELECT v._id, v._score
FROM vectorSearch(table='my-index', field='embedding',
     vector='[1.0, 2.0, 3.0]', option='max_distance=10.5') AS v
LIMIT 100
```

## Test plan
- [x] `./gradlew spotlessCheck` — PASS
- [x] `./gradlew :opensearch:test` — PASS
- [x] `./gradlew :integ-test:integTest -Dtests.class="*VectorSearchIT"` — PASS
- [x] `./gradlew :integ-test:integTest -Dtests.class="*VectorSearchExplainIT"` — PASS
- [x] `./gradlew build -x integTest` — PASS (full build excluding integration tests)

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchExplainIT.java
@@ -202,8 +202,33 @@ public class VectorSearchExplainIT extends SQLIntegTestCase {
                 + "WHERE v.state = 'TX' "
                 + "LIMIT 5");
 
-    // Efficient mode: knn rebuilt with filter inside, wrapped in WrapperQueryBuilder
+    // Efficient mode: knn rebuilt with filter inside, wrapped in WrapperQueryBuilder.
+    // The knn JSON (including the embedded filter) is base64-encoded inside the wrapper,
+    // so we verify structure by: (1) wrapper present, (2) no bool/must in plaintext
+    // (that would be post-filter shape), (3) decode the base64 payload to confirm
+    // the filter and predicate field are embedded inside the knn query.
     assertTrue("Explain should contain wrapper query:\n" + explain, explain.contains("wrapper"));
+    assertFalse(
+        "Efficient mode should not produce bool query (that is post-filter shape):\n" + explain,
+        explain.contains("\"bool\""));
+    assertFalse(
+        "Efficient mode should not contain must clause:\n" + explain, explain.contains("\"must\""));
+
+    // Extract and decode the base64 knn payload to verify filter embedding.
+    // The explain output escapes quotes as \", so match both \" and " forms.
+    java.util.regex.Matcher m =
+        java.util.regex.Pattern.compile("\\\\?\"query\\\\?\":\\\\?\"([A-Za-z0-9+/=]+)\\\\?\"")
+            .matcher(explain);
+    assertTrue("Explain should contain base64-encoded knn query:\n" + explain, m.find());
+    String knnJson =
+        new String(
+            java.util.Base64.getDecoder().decode(m.group(1)),
+            java.nio.charset.StandardCharsets.UTF_8);
+    assertTrue(
+        "Efficient mode knn JSON should contain filter:\n" + knnJson, knnJson.contains("filter"));
+    assertTrue(
+        "Efficient mode knn JSON should contain the WHERE predicate field:\n" + knnJson,
+        knnJson.contains("state"));
   }
 
   @Test
--- a/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchExplainIT.java
@@ -170,4 +170,57 @@ public class VectorSearchExplainIT extends SQLIntegTestCase {
 
     assertTrue("Explain should succeed with LIMIT <= k:\n" + explain, explain.contains("wrapper"));
   }
+
+  // ── filter_type explain ─────────────────────────────────────────────
+
+  @Test
+  public void testExplainFilterTypePostProducesBoolQuery() throws IOException {
+    String explain =
+        explainQuery(
+            "SELECT v._id, v._score "
+                + "FROM vectorSearch(table='"
+                + TEST_INDEX
+                + "', field='embedding', "
+                + "vector='[1.0, 2.0, 3.0]', option='k=10,filter_type=post') AS v "
+                + "WHERE v.state = 'TX' "
+                + "LIMIT 10");
+
+    assertTrue("Explain should contain bool query:\n" + explain, explain.contains("bool"));
+    assertTrue("Explain should contain must:\n" + explain, explain.contains("must"));
+    assertTrue("Explain should contain filter:\n" + explain, explain.contains("filter"));
+  }
+
+  @Test
+  public void testExplainFilterTypeEfficientProducesKnnWithFilter() throws IOException {
+    String explain =
+        explainQuery(
+            "SELECT v._id, v._score "
+                + "FROM vectorSearch(table='"
+                + TEST_INDEX
+                + "', field='embedding', "
+                + "vector='[1.0, 2.0]', option='k=5,filter_type=efficient') AS v "
+                + "WHERE v.state = 'TX' "
+                + "LIMIT 5");
+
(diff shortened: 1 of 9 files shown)


integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchExplainIT.java
```diff
@@ -170,4 +170,57 @@ public void testExplainLimitWithinKSucceeds() throws IOException {
...
+            "SELECT v._id, v._score "
+                + "FROM vectorSearch(table='"
+                + TEST_INDEX
+                + "', field='embedding', "
+                + "vector='[1.0, 2.0]', option='k=5,filter_type=efficient') AS v "
+                + "WHERE v.state = 'TX' "
+                + "LIMIT 5");
+
+    // Efficient mode: knn rebuilt with filter inside, wrapped in WrapperQueryBuilder
+    assertTrue("Explain should contain wrapper query:\n" + explain, explain.contains("wrapper"));
```
ahkcs, line 210: The test only asserts `explain.contains("wrapper")`, which would also pass for a top-k query without any filter. It doesn't verify the filter is actually embedded inside the knn JSON.
Should we consider adding an assertion that efficient-mode explain does NOT contain `"bool" / "must"`, or positively verify the knn JSON contains the `filter` clause.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
```diff
@@ -26,18 +26,32 @@ public class VectorSearchIndex extends OpenSearchIndex {
...
+      Map<String, String> options,
+      FilterType filterType) {
     super(client, settings, indexName);
     this.field = field;
     this.vector = vector;
     this.options = options;
+    this.filterType = filterType;
+  }
+
+  /** Backward-compatible constructor — defaults to no explicit filter type. */
```
dai-chen: Backward-compatible?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
```diff
@@ -76,6 +103,15 @@ private QueryBuilder buildKnnQuery() {
...
   String buildKnnQueryJson() {
+    return buildKnnQueryJson(null);
+  }
+
+  /**
+   * Builds knn query JSON, optionally embedding a filter clause for efficient filtering.
+   *
+   * @param filterJson serialized filter JSON to embed in knn.field.filter, or null for no filter
+   */
+  String buildKnnQueryJson(String filterJson) {
```
dai-chen, line 114: Is there query builder API in OpenSearch core?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
```diff
@@ -27,37 +30,84 @@
...
     Expression queryCondition = filter.getCondition();
-    QueryBuilder whereQuery = queryBuilder.build(queryCondition);
+    QueryBuilder whereQuery;
+    try {
+      whereQuery = queryBuilder.build(queryCondition);
+    } catch (ScriptQueryUnSupportedException e) {
+      if (filterTypeExplicit) {
+        throw new ExpressionEvaluationException(
+            "filter_type requires a pushdownable WHERE clause, but the given condition"
+                + " cannot be pushed down: "
```
dai-chen: Is this exposed to users? Just afraid they won't understand this.

integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchExplainIT.java
mengweieric: Agreed. I strengthened the test. It now asserts efficient mode does not produce the outer bool / must post-filter shape, and it decodes the wrapper query payload to verify the embedded k-NN JSON contains the filter and predicate field.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
mengweieric: There is a KNNQueryBuilder, but it lives in the k-NN plugin rather than OpenSearch core. In this PR I kept JSON construction centralized in VectorSearchIndex and wrapped it with WrapperQueryBuilder instead of adding a direct SQL dependency on the plugin API. Happy to revisit that tradeoff separately if we want to move toward a shared typed-builder path.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
mengweieric: Good catch. I rewrote the message to be user-friendly: "filter_type only works when the WHERE clause can be translated to an OpenSearch filter. Rewrite the WHERE clause or omit filter_type." It no longer appends internal exception text.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchIndex.java
mengweieric: Yes. the existing constructor without filterType still exists and delegates with filterType = null. In createScanBuilder(), null is treated as the default POST behavior with filterTypeExplicit = false, so existing call sites are unaffected. Renamed the comment in both VectorSearchIndex and VectorSearchQueryBuilder for clarity.

ahkcs (approved):

dai-chen (approved):

PR #5334, opened 2026-04-10
[BugFix] Fix bin/chart NPE with null values (#5174)

### LINKED ISSUE #5174: [BUG] bin then chart with null values causes NullPointerException in compareTo
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-numbers | bin value span=50 as val_bin | chart count() over val_bin by category
```
**Expected Result:**
Rows with null `value` should be handled safely (skipped or grouped) without server crash.

**Actual Result:**
HTTP 500 with NPE:
`Cannot invoke "java.lang.Comparable.compareTo(Object)" because "v0" is null`.

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "ts": { "type": "date" }
    }
  }
}
```
**Sample Data**
```json
[
  { "count": 2, "category": "A", "subcategory": "Y", "value": 20.3, "ts": "2024-01-02" },
  { "count": 10, "category": "B", "subcategory": "X", "value": 100.0, "ts": "2024-01-03" },
  { "count": null, "category": "B", "subcategory": "Y", "value": null, "ts": "2024-01-04" }
]
```

## Bug Description
**Issue Summary:**
`bin + chart` crashes when null bin values exist.

**Steps to Reproduce:**
1. Create index and include at least one document with `value = null`.
2. Run the query above.
3. Observe HTTP 500 NPE.

**Impact:**
Charting queries become unstable on sparse real-world data.

qianheng-aws:
### Description

Fix NullPointerException when using `bin` then `chart` with null values in the binned field.

**Root cause**: The bin bucket functions (`SpanBucketFunction`, `MinspanBucketFunction`, `RangeBucketFunction`) declared a non-nullable `VARCHAR(2000)` return type via `ReturnTypes.VARCHAR_2000`, even though they return `null` for null inputs. This caused Calcite's optimizer to remove the `IS NOT NULL` filters (added by `visitAggregation`'s `nonNullGroupMask` logic) as trivially true, allowing null group keys to reach the Enumerable aggregation's `TreeMap`, which crashed with NPE in `compareTo`.

**Fix**:
1. Declare the return type as nullable using `ReturnTypes.VARCHAR_2000.andThen(SqlTypeTransforms.FORCE_NULLABLE)` so that the null-filtering logic in `visitAggregation` works correctly
2. Add `nullsLast` to chart command sort operations as a defensive measure

### Related Issues
Resolves #5174

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### Test plan
- Unit test (`CalcitePPLChartNullTest`) verifying logical plan includes null filter and NULLS LAST
- Integration test (`CalciteBinChartNullIT`) verifying bin+chart with null values works end-to-end
- YAML REST test (`5174.yml`) reproducing the exact scenario from the issue

### DIFF
--- a/.claude/harness/ppl-bugfix-reference.md
+++ b/.claude/harness/ppl-bugfix-reference.md
@@ -109,8 +109,12 @@ teardown:
       headers: { Content-Type: 'application/json' }
       ppl: { body: { query: "source=test_issue_<ISSUE> | <your PPL>" } }
   - match: { total: <expected> }
-  - length: { datarows: <expected> }
+  - match: { datarows: [ [ <row1_val1>, <row1_val2> ], [ <row2_val1>, <row2_val2> ] ] }
 ```
+
+> **Always include `datarows` assertions** — verifying only `total` and `schema` will miss
+> wrong values. Count the expected output groups carefully (e.g., for `chart ... by <col>`,
+> count distinct (row_split, col_split) groups after null filtering, not the number of input rows).
 
 ---
 
--- a/CLAUDE_GUIDE.md
+++ b/CLAUDE_GUIDE.md
@@ -30,25 +30,3 @@ Fix a PPL bug end-to-end or follow up on an existing PR.
 
 **Related files:** [`.claude/harness/ppl-bugfix-harness.md`](.claude/harness/ppl-bugfix-harness.md)
 
----
-
-## `/dedupe`
-
-Find duplicate GitHub issues for a given issue.
-
-**Usage:**
-
-```
-/dedupe 1234
-```
-
-**What it does:**
-
-1. Reads the target issue
-2. Runs 3+ parallel search strategies to find potential duplicates (only older issues)
-3. Verifies candidates by reading each one
-4. Posts a structured comment on the issue listing 1-3 confirmed duplicates (if any)
-
-**Skips:** closed issues, broad feedback issues, issues already checked
-
-**Related files:** [`scripts/comment-on-duplicates.sh`](scripts/comment-on-duplicates.sh)
--- a/.claude/commands/ppl-bugfix.md
+++ b/.claude/commands/ppl-bugfix.md
@@ -83,6 +83,33 @@ Agent(
 
 ### 2B: Follow-up
 
+Before dispatching, check if an existing worktree already has the PR branch checked out:
+
+```bash
+# List worktrees and find one on the PR branch
+for wt in .claude/worktrees/agent-*/; do
+  branch=$(git -C "$wt" branch --show-current 2>/dev/null)
+  if [ "$branch" = "<pr_branch>" ]; then
+    echo "REUSE: $wt (branch: $branch)"
+  fi
+done
+```
+
+**If existing worktree found**: Do NOT use `isolation: "worktree"`. Pass the worktree path in the prompt so the subagent works there directly.
+
+```
+Agent(
+  mode: "<resolved_mode>",
+  name: "bugfix-<issue_number>",
+  description: "PPL bugfix #<issue_number> followup",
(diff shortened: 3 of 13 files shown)


qianheng-aws:
## Decision Log
**Root Cause**: Bin bucket functions (`SpanBucketFunction`, `MinspanBucketFunction`, `RangeBucketFunction`) declared non-nullable `VARCHAR(2000)` return type via `ReturnTypes.VARCHAR_2000`. When input is null, these functions return null, but Calcite's optimizer sees the return type as NOT NULL and removes `IS NOT NULL` filters (from `visitAggregation`'s `nonNullGroupMask`) as trivially true. Null group keys then reach `TreeMap` in Enumerable aggregation, crashing with NPE in `compareTo`.

**Approach**: 
1. Changed return type to `ReturnTypes.VARCHAR_2000.andThen(SqlTypeTransforms.FORCE_NULLABLE)` on all three bucket functions — this is the root fix that lets Calcite preserve the null filter
2. Added `nullsLast` to chart command sort operations as a defensive measure against null keys reaching sort

**Alternatives Rejected**:
- Adding null checks in the bucket function implementations themselves — treats the symptom, not the root cause; Calcite would still optimize away the filter
- Adding explicit `WHERE ... IS NOT NULL` in `visitChart` — redundant if the type is correctly declared nullable, since `visitAggregation` already adds `nonNullGroupMask` filters

**Pitfalls**: The non-nullable return type was inherited from `ReturnTypes.VARCHAR_2000` which is Calcite's built-in constant. Easy to miss that UDFs returning null for null inputs need explicit nullable declaration.

**Things to Watch**: Other UDFs using `ReturnTypes.VARCHAR_2000` or similar non-nullable return types that can actually return null — they may have the same latent issue.

CLAUDE_GUIDE.md
```diff
@@ -30,25 +30,3 @@ Fix a PPL bug end-to-end or follow up on an existing PR.
 
 **Related files:** [`.claude/harness/ppl-bugfix-harness.md`](.claude/harness/ppl-bugfix-harness.md)
 
----
-
-## `/dedupe`
-
-Find duplicate GitHub issues for a given issue.
-
```
LantaoJin, line 38: Is this removing on purpose?

LantaoJin (dismissed):

.claude/harness/ppl-bugfix-reference.md
```diff
@@ -109,9 +109,13 @@ teardown:
       headers: { Content-Type: 'application/json' }
       ppl: { body: { query: "source=test_issue_<ISSUE> | <your PPL>" } }
   - match: { total: <expected> }
-  - length: { datarows: <expected> }
+  - match: { datarows: [ [ <row1_val1>, <row1_val2> ], [ <row2_val1>, <row2_val2> ] ] }
```
qianheng-aws, line 112: This change is from skill self-improving after receiving the feedback of PR_REVEW workflow comments.

CLAUDE_GUIDE.md
qianheng-aws: Yes, https://github.com/opensearch-project/sql/pull/5319 has migrate to reuse the workflow in opensearch-build. The original dedupe workflow has been migrated to that repo for reusing across openseach projects.

ahkcs (dismissed):

ahkcs:
Should we add `CalciteBinChartNullIT` to `CalciteNoPushdownIT` suite?

LantaoJin (approved):

songkant-aws (approved):

PR #5337, opened 2026-04-13
[BugFix] Fix NOT IN including null/missing rows (#5165)

### LINKED ISSUE #5165: [BUG] NOT IN includes null/missing rows due to missing exists filter in pushdown
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-types | where int_field NOT IN (42, -1, 0) | fields int_field
```
**Expected Result:**
`NOT IN` should exclude documents where `int_field` is `null`/missing.

**Actual Result:**
Result includes both `2147483647` and `null` rows.

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "int_field": { "type": "integer" }
    }
  }
}
```
**Sample Data**
```json
[
  { "int_field": 42 },
  { "int_field": -1 },
  { "int_field": 0 },
  { "int_field": 2147483647 },
  { "int_field": null }
]
```

## Bug Description
**Issue Summary:**
`NOT IN` incorrectly matches null/missing field rows.

**Steps to Reproduce:**
1. Create index with `int_field`.
2. Insert sample rows including one `null` value.
3. Run the query above.
4. Observe `null` row in output.

**Impact:**
Incorrect filter semantics and wrong result sets in production queries.

songkant-aws:
### Description
When `PredicateAnalyzer` handles a `SEARCH` expression with complemented points (`NOT IN`) and `nullAs=UNKNOWN`, the generated OpenSearch DSL query was missing an `exists` filter. This caused documents with null/missing field values to be incorrectly included in `NOT IN` results.

**Root cause**: In the `case UNKNOWN` branch of the SEARCH handler (~line 728), the expression was returned as-is without adding an `exists` check for complemented points. SQL three-valued logic dictates that `NULL NOT IN (...)` evaluates to `UNKNOWN` (not `TRUE`), so null rows must be excluded.

**Fix**: Check `isSearchWithComplementedPoints(call)` in the `UNKNOWN` branch and wrap with `AND exists` filter, consistent with the existing `FALSE` branch behavior.

### Related Issues
Resolves #5165

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteNotInNullFilterIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteNotInNullFilterIT.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.remote;
+
+import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_BANK_WITH_NULL_VALUES;
+import static org.opensearch.sql.util.MatcherUtils.rows;
+import static org.opensearch.sql.util.MatcherUtils.verifyDataRowsInOrder;
+
+import java.io.IOException;
+import org.json.JSONObject;
+import org.junit.Test;
+import org.opensearch.sql.ppl.PPLIntegTestCase;
+
+/** Integration test for NOT IN excluding null/missing rows (issue #5165). */
+public class CalciteNotInNullFilterIT extends PPLIntegTestCase {
+
+  @Override
+  public void init() throws Exception {
+    super.init();
+    enableCalcite();
+    loadIndex(Index.BANK_WITH_NULL_VALUES);
+  }
+
+  @Test
+  public void testNotInExcludesNullRows() throws IOException {
+    // age values: 32, 36, 28, 33, 36, null, 34
+    // NOT IN (32, 28) should return 36, 33, 36, 34 — excluding the null row
+    JSONObject result =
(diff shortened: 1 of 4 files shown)


songkant-aws:
## Decision Log
**Root Cause**: `PredicateAnalyzer.java` line 728 — the `case UNKNOWN` branch of the SEARCH handler returned the expression without an `exists` filter. For `NOT IN` (complemented points), this meant the generated DSL was `must_not(termsQuery)` without `exists`, so documents with null/missing fields passed through.

**Approach**: Added `isSearchWithComplementedPoints(call)` check in the `UNKNOWN` branch, wrapping with `AND exists` filter when true. This mirrors the existing correct behavior in the `FALSE` branch (lines 718-722). The fix is minimal — 7 lines changed in production code.

**Alternatives Rejected**: 
- Modifying Calcite's `nullAs` generation to always produce `FALSE` for `NOT IN` — too invasive, changes Calcite semantics upstream.
- Adding a post-processing filter in the execution layer — would defeat pushdown optimization.

**Pitfalls**: The `UNKNOWN` branch also handles non-complemented cases (positive `IN`, ranges). Those must NOT get an `exists` filter added since positive matches naturally exclude nulls. The `isSearchWithComplementedPoints` check ensures only `NOT IN` gets the fix.

**Things to Watch**: Any future Sarg patterns with `nullAs=UNKNOWN` and negation semantics should be reviewed for the same null-handling gap.

LantaoJin (approved):

qianheng-aws (approved):

PR #5338, opened 2026-04-13
[BugFix] Fix NOT LIKE includes null/missing field rows (#5169)

### LINKED ISSUE #5169: [BUG] NOT LIKE includes null/missing field rows (same root cause as A003)
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-types | where NOT keyword_field LIKE '%ello%' | fields keyword_field
```
**Expected Result:**
Rows with `keyword_field = null` should be excluded.

**Actual Result:**
Output includes `null` row in addition to non-matching string rows.

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "keyword_field": { "type": "keyword" }
    }
  }
}
```
**Sample Data**
```json
[
  { "keyword_field": "hello" },
  { "keyword_field": "world" },
  { "keyword_field": "" },
  { "keyword_field": "special chars..." },
  { "keyword_field": null }
]
```

## Bug Description
**Issue Summary:**
`NOT LIKE` incorrectly includes null/missing values.

**Steps to Reproduce:**
1. Create index with `keyword_field`.
2. Insert sample rows including `null`.
3. Run the query above.
4. Observe `null` in output.

**Impact:**
Negated string filters produce incorrect result sets.

songkant-aws:
### Description

`SimpleQueryExpression.not()` in `PredicateAnalyzer.java` generated `boolQuery().mustNot(builder())` without an `existsQuery` filter. This caused OpenSearch's `must_not` to match documents where the field doesn't exist (null/missing), leaking null rows into NOT LIKE results.

Fixed by adding `.must(existsQuery(getFieldReference()))` alongside the `mustNot`, consistent with how `notLike()` and `notEquals()` are already correctly implemented in the same class.

### Related Issues
Resolves #5169

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed


songkant-aws:
## Decision Log
**Root Cause**: `SimpleQueryExpression.not()` in `PredicateAnalyzer.java` (line ~1300) generated `boolQuery().mustNot(builder())` without an `existsQuery` filter. OpenSearch's `must_not` clause alone matches documents where the field doesn't exist, causing null/missing rows to leak through in any negated predicate that routes through `not()` (e.g., `NOT ... LIKE`, `NOT field > N`).

**Approach**: Added `.must(existsQuery(getFieldReference()))` to the `not()` method, matching the existing pattern used by `notLike()` (line ~1346) and `notEquals()` (line ~1368) in the same class. This ensures all generic negations exclude null/missing documents.

**Alternatives Rejected**: 
- Modifying the `prefix()` method to inject exists logic before calling `not()` — rejected because it would duplicate the pattern and `not()` is the correct place for this concern.
- Only fixing for LIKE specifically — rejected because `not()` is generic and affects all negations routed through it (e.g., `NOT field > N`).

**Pitfalls**: 
- The `getFieldReference()` returns the base field name (e.g., `b`), not the keyword subfield (e.g., `b.keyword`). This is correct because OpenSearch's `exists` query works with base field names.
- `NOT(field = value)` is not affected because it uses `notEquals()` directly rather than routing through `not()`.

**Things to Watch**: 
- Verify that the `not()` path doesn't cause double-exists checks when combined with other operators that already add exists filters.
- The `notEquals()` method at line ~1379 has a comment saying "NOT LIKE should return false when field is NULL" but it's in the notEquals method — this is a copy-paste comment issue, not a bug.

qianheng-aws (approved):

yuancu (approved):

Issue #5170, opened 2026-02-23
[BUG] json_append and json_extend crash with $.path JSONPath syntax (same root cause as A012)


penghuo:
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-types | eval result = json_append('{"a":[1,2]}', '$.a', 3) | fields result
source=bounty-types | eval result = json_extend('{"a":[1,2]}', '$.a', 3) | fields result
```
**Expected Result:**
Both queries should return `{"a":[1,2,3]}`.

**Actual Result:**
Both fail with HTTP 500 and malformed path error in JSON insert (`"$..meaningless_key"`).

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [ ] Simple Schema for Observability (SS4O)
- [ ] Open Cybersecurity Schema Framework (OCSF)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "int_field": { "type": "integer" }
    }
  }
}
```
**Sample Data**
```json
{ "int_field": 42 }
```

## Bug Description
**Issue Summary:**
`json_append/json_extend` fail when using standard JSONPath (`$.a`), while shorthand path (`a`) works.

**Steps to Reproduce:**
1. Ensure any source index exists (for example one row in `bounty-types`).
2. Run both queries above.
3. Observe HTTP 500 with invalid JSON insert path.

**Impact:**
JSON array update operations are broken for standard JSONPath usage.

PR #5339, opened 2026-04-13
[BugFix] Fix json_set crash and json_delete no-op with $.key paths (#5167)

### LINKED ISSUE #5167: [BUG] json_set crashes and json_delete is no-op due to JSONPath double-prefixing
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-types | where int_field = 42 | eval modified = json_set(json_data, '$.name', 'modified_alice') | fields modified
source=bounty-types | where int_field = 42 | eval deleted = json_delete(json_data, '$.name') | fields deleted
```
**Expected Result:**
- `json_set` should update `name` to `modified_alice`.
- `json_delete` should remove `name` key.

**Actual Result:**
- `json_set` returns HTTP 500 with invalid input (`kvs: ["$.","modified_alice"]`).
- `json_delete` returns original JSON unchanged.

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "json_data": { "type": "keyword" }
    }
  }
}
```
**Sample Data**
```json
{
  "int_field": 42,
  "json_data": "{\"name\":\"alice\",\"scores\":[90,85,92]}"
}
```

## Bug Description
**Issue Summary:**
JSONPath `$.key` handling in `json_set/json_delete` is broken (`json_set` crashes, `json_delete` no-op).

**Steps to Reproduce:**
1. Create index with JSON string field.
2. Insert sample row.
3. Run both queries above.
4. Observe HTTP 500 for `json_set` and unchanged output for `json_delete`.

**Impact:**
Core JSON mutation functions are unreliable with standard JSONPath syntax.

qianheng-aws:
### Description
`json_set` crashes (HTTP 500) and `json_delete` silently returns unchanged JSON when using standard JSONPath `$.key` syntax. The root cause is `convertToJsonPath()` in `JsonUtils.java`, which unconditionally prepends `$.` to the input — turning `$.name` into `$.$.name` (double-prefixed).

The fix strips any existing `$.` or `$` prefix from the input before processing, preventing double-prefixing while preserving the existing curly-brace-to-bracket conversion logic.

### Related Issues
Resolves #5167, #5170

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonUtils.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonUtils.java
@@ -23,6 +23,15 @@ public class JsonUtils {
   public static String convertToJsonPath(String input) {
     if (input == null || input.isEmpty()) return "$";
 
+    // Strip leading "$." or "$" to avoid double-prefixing (issue #5167)
+    if (input.startsWith("$.")) {
+      input = input.substring(2);
+    } else if (input.startsWith("$")) {
+      input = input.substring(1);
+    }
+
+    if (input.isEmpty()) return "$";
+
     StringBuilder sb = new StringBuilder("$.");
     int i = 0;
     while (i < input.length()) {
--- a/core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
+++ b/core/src/test/java/org/opensearch/sql/expression/json/JsonFunctionsTest.java
@@ -19,6 +19,8 @@ import org.opensearch.sql.exception.ExpressionEvaluationException;
 import org.opensearch.sql.exception.SemanticCheckException;
 import org.opensearch.sql.expression.DSL;
 import org.opensearch.sql.expression.LiteralExpression;
+import org.opensearch.sql.expression.function.jsonUDF.JsonDeleteFunctionImpl;
+import org.opensearch.sql.expression.function.jsonUDF.JsonSetFunctionImpl;
 import org.opensearch.sql.expression.function.jsonUDF.JsonUtils;
 
 @ExtendWith(MockitoExtension.class)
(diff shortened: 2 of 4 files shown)


qianheng-aws:
## Decision Log
**Root Cause**: `JsonUtils.convertToJsonPath()` unconditionally prepends `$.` to its input. When users pass standard JSONPath syntax like `$.name`, the result is `$.$.name` (double-prefixed). For `json_set`, Calcite's `JsonFunctions.jsonSet` interprets the invalid `$.` segment as a key and crashes. For `json_delete`, `JsonFunctions.jsonRemove` finds no match for `$.$.name` and returns the JSON unchanged.

**Approach**: Strip any existing `$.` or `$` prefix from the input at the start of `convertToJsonPath()` before the function proceeds with its normal processing. This is the minimal change that handles all prefix variants (`$`, `$.`, or none) and preserves the existing curly-brace-to-bracket conversion logic (e.g., `{}`→`[*]`).

**Alternatives Rejected**: 
- *Return early if input starts with `$`*: This would skip the curly-brace conversion, breaking hypothetical inputs like `$.a{}.b` that combine both syntaxes. The strip-then-process approach is more robust.
- *Regex-based normalization*: More complex for no additional benefit; the simple prefix check is clearer and handles all cases.

**Pitfalls**: None encountered — the fix is localized to one utility method. All existing tests (unit + integration + YAML) continue to pass.

**Things to Watch**: The `json_extract`, `json_append`, and `json_extend` functions also call `convertToJsonPath`, so they all benefit from this fix. Users can now use either `$.name` or `name` interchangeably for path arguments in all JSON mutation functions.

songkant-aws:
Nit: Add two more regression test cases for `json_append` and `json_extend` as the fix also covers fix in #5170

songkant-aws (approved):

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/JsonUtils.java
```diff
@@ -23,6 +23,15 @@ public class JsonUtils {
   public static String convertToJsonPath(String input) {
     if (input == null || input.isEmpty()) return "$";
 
+    // Strip leading "$." or "$" to avoid double-prefixing (issue #5167)
+    if (input.startsWith("$.")) {
+      input = input.substring(2);
+    } else if (input.startsWith("$")) {
+      input = input.substring(1);
+    }
```
LantaoJin, line 31: Could the input start with other invalid char? for example `$$.name`, etc.

qianheng-aws: > This is the minimal change that handles all prefix variants (\\$, \\$., or none)

No, $$. is not standard JSONPath.

LantaoJin (approved):

PR #5341, opened 2026-04-13
Make analytics-engine an optional dependency

ahkcs:
## Summary
- Change `extendedPlugins` to `analytics-engine;optional=true` so the SQL plugin starts without analytics-engine installed
- Conditionally exclude Calcite JARs from `bundlePlugin` based on whether analytics-engine ZIP exists at build time
- Conditionally load analytics-engine in all test clusters (integTest, yamlRestTest, security, JDBC, doctest)
- Conditionally set analytics-framework dependency scope (`compileOnly` when parent provides it, `api` when standalone)

## Behavior
- **With analytics-engine**: Calcite and analytics-framework excluded from SQL plugin (provided by parent classloader), analytics path works as before
- **Without analytics-engine**: Calcite and analytics-framework bundled in SQL plugin, analytics path skipped, regular SQL/PPL queries work normally

### With analytics-engine present
```
$ ./gradlew :opensearch-sql-plugin:bundlePlugin
BUILD SUCCESSFUL

$ unzip -l opensearch-sql-*.zip | grep calcite-core
(not found — excluded, provided by parent)

$ unzip -l opensearch-sql-*.zip | grep analytics-framework
(not found — excluded, provided by parent)

$ ./gradlew :integ-test:integTest -Dtests.class="*AnalyticsPPLIT" -Dtests.class="*AnalyticsSQLIT"
BUILD SUCCESSFUL

$ ./gradlew :doctest:doctest -PignorePrometheus
Ran 75 tests — BUILD SUCCESSFUL
```

### Without analytics-engine (ZIP removed)
```
$ mv libs/analytics-engine-*.zip libs/analytics-engine-*.zip.bak

$ ./gradlew :opensearch-sql-plugin:bundlePlugin
BUILD SUCCESSFUL

$ unzip -l opensearch-sql-*.zip | grep calcite-core
  8811320  calcite-core-1.41.0.jar (bundled)

$ unzip -l opensearch-sql-*.zip | grep analytics-framework
    11979  analytics-framework-3.6.0-SNAPSHOT.jar (bundled)

$ ./gradlew :integ-test:integTest -Dtests.class="org.opensearch.sql.calcite.CalcitePPLBasicIT"
BUILD SUCCESSFUL

$ ./gradlew :doctest:doctest -PignorePrometheus
Ran 75 tests — BUILD SUCCESSFUL
```

## Context
Per discussion with @mch2 — analytics-engine should be an optional dependency so the SQL plugin can work standalone. OpenSearch supports `optional=true` syntax for `extendedPlugins` (parses `;optional=true` suffix in `PluginInfo`).

Ref: https://github.com/opensearch-project/sql/issues/5246#issuecomment-4210068441

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -63,7 +63,13 @@ dependencies {
     }
     api 'org.apache.calcite:calcite-linq4j:1.41.0'
     api project(':common')
-    compileOnly files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
+    // When analytics-engine is present, framework classes come from parent classloader (compileOnly).
+    // When absent, bundle the framework JAR so its classes are available at runtime.
+    if (file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip").exists()) {
+        compileOnly files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
+    } else {
+        api files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
+    }
     testImplementation files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
     implementation "com.github.seancfoley:ipaddress:5.4.2"
     implementation "com.jayway.jsonpath:json-path:2.9.0"
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -65,7 +65,7 @@ dependencies {
     api project(':common')
     // When analytics-engine is present, framework classes come from parent classloader (compileOnly).
     // When absent, bundle the framework JAR so its classes are available at runtime.
-    if (file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip").exists()) {
+    if (hasAnalyticsEngine) {
         compileOnly files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
     } else {
         api files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
--- a/doctest/build.gradle
+++ b/doctest/build.gradle
@@ -195,7 +195,10 @@ testClusters {
         }))
         */
         plugin(getJobSchedulerPlugin(jsPlugin, bwcOpenSearchJSDownload))
-        plugin provider { (RegularFile) (() -> file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")) }
+        def analyticsZip = file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")
+        if (analyticsZip.exists()) {
+            plugin provider { (RegularFile) (() -> analyticsZip) }
+        }
         plugin ':opensearch-sql-plugin'
         testDistribution = 'archive'
     }
--- a/doctest/build.gradle
+++ b/doctest/build.gradle
@@ -196,7 +196,7 @@ testClusters {
         */
         plugin(getJobSchedulerPlugin(jsPlugin, bwcOpenSearchJSDownload))
         def analyticsZip = file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")
-        if (analyticsZip.exists()) {
+        if (hasAnalyticsEngine) {
             plugin provider { (RegularFile) (() -> analyticsZip) }
         }
         plugin ':opensearch-sql-plugin'
--- a/integ-test/build.gradle
+++ b/integ-test/build.gradle
@@ -270,7 +270,6 @@ def getGeoSpatialPlugin() {
     }
 }
 
-def analyticsEngineZip = file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")
 def getAnalyticsEnginePlugin() {
     provider { (RegularFile) (() -> file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")) }
 }
@@ -280,7 +279,7 @@ testClusters {
         testDistribution = 'archive'
         plugin(getJobSchedulerPlugin())
         plugin(getGeoSpatialPlugin())
-        if (analyticsEngineZip.exists()) {
+        if (hasAnalyticsEngine) {
             plugin(getAnalyticsEnginePlugin())
         }
         plugin ":opensearch-sql-plugin"
@@ -290,7 +289,7 @@ testClusters {
         testDistribution = 'archive'
         plugin(getJobSchedulerPlugin())
         plugin(getGeoSpatialPlugin())
-        if (analyticsEngineZip.exists()) {
+        if (hasAnalyticsEngine) {
             plugin(getAnalyticsEnginePlugin())
         }
         plugin ":opensearch-sql-plugin"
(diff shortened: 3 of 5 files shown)


plugin/build.gradle
```diff
@@ -55,49 +55,52 @@ opensearchplugin {
...
-    exclude 'httpcore5-5*.jar'
-    exclude 'httpcore5-h2-*.jar'
-    exclude 'httpcore5-reactive-*.jar'
-    exclude 'antlr4-runtime-4.13.2.jar'
-    exclude 'jackson-annotations-*.jar'
-    exclude 'jackson-databind-*.jar'
+// When analytics-engine is present, exclude jars it provides via extendedPlugins classloader.
+// When analytics-engine is absent, keep all jars so the SQL plugin works standalone.
+def analyticsEngineZip = file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")
+if (analyticsEngineZip.exists()) {
```
RyanL1997: Build-time file-existence check can break reproducible builds — the same source tree produces different artifacts depending on whether `libs/analytics-engine-3.6.0-SNAPSHOT.zip` happens to be on the builder's machine. CI builds could silently differ from dev builds.

Consider making it an explicit Gradle property (e.g. `./gradlew build -PbundleAnalytics=true`) with the file-exists check as a fallback. At minimum, consider printing the decision at configuration time so it's visible in build logs:

```groovy
println "Building with analytics-engine: ${analyticsEngineZip.exists()}"
```

core/build.gradle
```diff
@@ -63,7 +63,13 @@ dependencies {
     }
     api 'org.apache.calcite:calcite-linq4j:1.41.0'
     api project(':common')
-    compileOnly files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
+    // When analytics-engine is present, framework classes come from parent classloader (compileOnly).
+    // When absent, bundle the framework JAR so its classes are available at runtime.
+    if (file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip").exists()) {
```
RyanL1997: This existence check is duplicated across `core/build.gradle`, `plugin/build.gradle`, `integ-test/build.gradle`, and `doctest/build.gradle` — all looking for the same path. If the version ever changes (e.g. `3.7.0-SNAPSHOT`), all four need to be updated in lockstep.

Recommend extracting to a single root-level property, e.g. in the root `build.gradle`:

```groovy
ext {
    analyticsEngineZip = file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")
    hasAnalyticsEngine = analyticsEngineZip.exists()
}
```

Then each subproject can use `rootProject.hasAnalyticsEngine`.

core/build.gradle
```diff
@@ -63,7 +63,13 @@ dependencies {
...
+        compileOnly files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
+    } else {
+        api files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
```
RyanL1997, line 71: Question: switching from `compileOnly` to `api` means downstream modules that depend on `core` will also get `analytics-framework` on their compile classpath. Is that intentional? `runtimeOnly` might be more appropriate since the goal is just to have classes available at runtime, not to let downstream modules compile against framework APIs. Same concern applies to the similar change in `plugin/build.gradle` line 213.

plugin/build.gradle
```diff
@@ -201,7 +204,14 @@ dependencies {
...
     api project(":ppl")
     api project(':api')
-    compileOnly files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
+    // analytics-framework: compileOnly when analytics-engine provides it via parent classloader,
+    // runtimeOnly when standalone so it gets bundled in the plugin ZIP.
+    if (analyticsEngineZip.exists()) {
+        compileOnly files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
+    } else {
+        // Bundle analytics-framework when analytics-engine is absent
+        api files("${rootDir}/libs/analytics-framework-3.6.0-SNAPSHOT.jar")
```
RyanL1997, line 212: The unconditional `compileOnly files("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.jar")` a few lines below (still present, not shown in this hunk) references the analytics-**engine** JAR directly. If analytics-engine is absent, does this JAR also need to be absent — i.e. is the JAR always coupled with the ZIP? If so, that `compileOnly` line should probably also be guarded by `analyticsEngineZip.exists()`. Worth clarifying.

integ-test/build.gradle
```diff
@@ -270,6 +270,7 @@ def getGeoSpatialPlugin() {
     }
 }
 
+def analyticsEngineZip = file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")
```
RyanL1997: Nice — defining `analyticsEngineZip` at top level for reuse is cleaner than inlining. Minor: `getAnalyticsEnginePlugin()` just below still inlines its own `file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")` call. Could reuse `analyticsEngineZip` there so the path only appears once.

doctest/build.gradle
```diff
@@ -195,7 +195,10 @@ testClusters {
         }))
         */
         plugin(getJobSchedulerPlugin(jsPlugin, bwcOpenSearchJSDownload))
-        plugin provider { (RegularFile) (() -> file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")) }
+        def analyticsZip = file("${rootDir}/libs/analytics-engine-3.6.0-SNAPSHOT.zip")
```
RyanL1997, line 198: Nit: `integ-test/build.gradle` defines `analyticsEngineZip` at the top of the file for reuse, but here it's inlined inside the config block as `analyticsZip`. Not wrong, just inconsistent style across the repo after this change.

RyanL1997 (commented):
Hi @ahkcs thanks for the change. Left some comments.

plugin/build.gradle
ahkcs: Added a property `-PhasAnalyticsEngine=true/false` to override, with file-exists as fallback. Build logs show the decision: Building with `analytics-engine: true/false`.

core/build.gradle
ahkcs: Good suggestion. Moved the detection to a single root-level ext property in build.gradle — all subprojects reference `hasAnalyticsEngine` directly.

plugin/build.gradle
ahkcs: In the current feature branch it will always be true unless we specify it to false since we have the zip file locally. Logic will be updated once core is able to publish `analytics-engine` to maven

core/build.gradle
ahkcs: This is a workaround because Gradle files(...) dependencies don't propagate transitively like Maven coordinates do. Even though core declares api files("analytics-framework.jar"), it doesn't flow to plugin's classpath for bundling. Once analytics-framework is published to Maven, we can remove the declaration from `plugin/build.gradle` entirely.

ahkcs:
@RyanL1997 This PR is currently on hold, we are waiting for core team's implementation to finalize the PR. Our goal for this PR is to make SQL plugin skip the analytics path gracefully when analytics engine is not installed during runtime

Issue #5142, opened 2026-02-12
[RFC] xyseries command implementation in PPL


asifabashar:
**Problem Statement**
xyseries command is missing in PPL which is part of todo in [roadmap](https://github.com/orgs/opensearch-project/projects/261/views/1?pane=issue&itemId=152511697) 

This RFC proposes adding a new PPL transforming command, `xyseries`, with syntax and behavior aligned with SPL .

`xyseries` converts row-oriented grouped results into a wide table where:
- One field is the X axis (`x-field`) and stays as a row key.
- One field  (`y-name-field`)  provides a part of column name  from proivded  paramter values to choose from and used as pivot values in  conjuction with (`y-data-field`) field name , where data for this column are pivoted cells that are agrregated value fields.

OpenSearch PPL already supports `stats`, `chart`, and `timechart`, but there is no direct equivalent for SPL `xyseries`. 

Adding `xyseries` improves:
- Query readability for pivot-style result shaping.
- Dashboard interoperability for charts requiring wide-table series layout.

**Current State**
There is no xyseries compatiable command.

**Long-Term Goals**
Provide xyseries functionality. 

**Proposal**
Summarizes the suggested solution or improvement.

**Approach**
## User Syntax

```syntax
xyseries  [sep=<string>] [format=<string>] <x-field> <y-name-field> in (<value1>, <value2>, ...) <y-data-field>[,<y-data-field2>
```

###  Arguments
- `x-field` (required): row key in output.
- `y-name-field` (required): Number of values used to generate output series column names based on provided in paramter values such as `<value1>, <value2>` etc
- `y-data-field...` (required, at least one): value field(s) used to fill cells.

### Options
| Option | Default | Behavior |
| --- | --- | --- |
| `sep` | `":"` | Separator between `y-data-field` name and 'in' parameter values <value1>, <value2> etc.  |
| `format` | None | Naming template for output series columns represented as  `$AGG$<sep>$VAL$`  .  For each data row pivot value for `y-name-field` and `$AGG$` (`y-data-field` name). The format parameter inserts the  (`y-data-field` name)  uses “:” as the built-in separator and then pivot value. $VAL$  |

If `format` is omitted:
- Single `y-data-field`: output column is   based row number .
- Multiple `y-data-field`: output column is `$AGG$<sep>$VAL$` where $VAL$ represent provided pivot values in 'in' operator.

### Examples
```
Actual data:
ppl
[...]

PR #5343, opened 2026-04-14
Xyseries command implementation

asifabashar:
### Description
**Problem Statement**
xyseries command is missing in PPL which is part of todo in [roadmap](https://github.com/orgs/opensearch-project/projects/261/views/1?pane=issue&itemId=152511697) 

This RFC proposes adding a new PPL transforming command, `xyseries`, with syntax and behavior aligned with SPL .

`xyseries` converts row-oriented grouped results into a wide table where:
- One field is the X axis (`x-field`) and stays as a row key.
- One field  (`y-name-field`)  provides a part of column name  from proivded  paramter values to choose from and used as pivot values in  conjuction with (`y-data-field`) field name , where data for this column are pivoted cells that are agrregated value fields.

OpenSearch PPL already supports `stats`, `chart`, and `timechart`, but there is no direct equivalent for SPL `xyseries`. 

Adding `xyseries` improves:
- Query readability for pivot-style result shaping.
- Dashboard interoperability for charts requiring wide-table series layout.

**Current State**
There is no xyseries compatiable command.

**Long-Term Goals**
Provide xyseries functionality. 

**Proposal**
Summarizes the suggested solution or improvement.

**Approach**
## User Syntax

```syntax
xyseries  [sep=<string>] [format=<string>] <x-field> <y-name-field> in (<value1>, <value2>, ...) <y-data-field>[,<y-data-field2>
```

###  Arguments
- `x-field` (required): row key in output.
- `y-name-field` (required): Number of values used to generate output series column names based on provided in paramter values such as `<value1>, <value2>` etc
- `y-data-field...` (required, at least one): value field(s) used to fill cells.

### Options
| Option | Default | Behavior |
| --- | --- | --- |
| `sep` | `":"` | Separator between `y-data-field` name and 'in' parameter values <value1>, <value2> etc.  |
| `format` | None | Naming template for output series columns represented as  `$AGG$<sep>$VAL$`  .  For each data row pivot value for `y-name-field` and `$AGG$` (`y-data-field` name). The format parameter inserts the  (`y-data-field` name)  uses “:” as the built-in separator and then pivot value. $VAL$  |

If `format` is omitted:
- Single `y-data-field`: output column is   based row number .
- Multiple `y-data-field`: output column is `$AGG$<sep>$VAL$` where $VAL$ represent provided pivot values in 'in' operator.

### Examples
```
Actual data:
ppl
 source=weblogs| stats count(host) as host_cnt, count(method) as method_cnt by url, response
```                                                                                                                                                                          
| host_cnt | method_cnt | response | url    |
|-------------|---------------|----------|--------|
| 3           | 3             | 200      | /page1 |
| 1           | 1             | 404      | /page1 |
| 5           | 5             | 200      | /page2 |
| 2           | 2             | 500      | /page2 |

After xyseries transformation:
ppl
```
source=weblogs
| stats count(host) as host_cnt, count(method) as method_cnt by url, response
| xyseries url response in ("200","404","500") host_cnt, method_cnt

```
| url    | host_cnt: 200 | host_cnt: 404 | host_cnt: 500 | method_cnt: 200 | method_cnt: 404 | method_cnt: 500 |
|------|----------------|----------------|-----------------|-------------------|--------------------|--------------------|
| /page1 | 3                | 1                | null             | 3                   | 1                   | null                |
| /page2 | 5                | null             | 2                | 5                   | null                | 2                   |

If format is not specified, by default, you’d see something like:
url:count(host) and url:count(method) as column names.
 
If you provide your own separator through the format option, it overrides anything defined with sep. If for example,  sep is set to “-”, but format specifies “+”, and because format has higher priority, the “+” is used.
Here, $VAL$ and $AGG$ are placeholders represent  <y-name-field> and the <y-data-field> respectively. In the output, you can see that the name field (url) and the data field count(host) may appear in the position of $VAL$ and $AGG$, depending on how the format string arranges them.

### Input shape and type rules

Support xyseries only when pivot values are explicitly provided.

- `x-field` and `y-name-field` must exist in input schema..
- `y-name-field` values are converted to string for output column naming.
-  `y-data-field` One or more fields that contain the data to chart. If there are multiple fields specified, separate the field names with commas. 

###  Null handling
- Rows with `null` in a given `y-data-field` do not contribute a value for that generated series column.

### Implementation Plan

 Validate required fields/options and defaults.
 Project to `x`, `y_name` selected from provided pivot values in 'in' operator, and selected `y_data` fields.
 Build generated series column name using `format` or default naming with `sep` .
 Pivot to wide schema based on provided pivot values passed as parameter.

_Today our direction is to keep PPL commands translatable to SQL/Calcite plans as much as possible. We also plan to run PPL on Spark; SQL-native xyseries (with explicit in (...) values) is portable to Spark SQL, while post-processing would require separate Spark-side custom implementation.
Composability: if xyseries is implemented after query execution, it effectively must be terminal and cannot be reliably chained with downstream commands_. (Reference: Comments from [penghuo](https://github.com/penghuo) )

### out of scope
grouping option which does not apply for multifile input in OpenSearch.
**Alternative**
chart , timeseries can be used for similar use cases but not exactly same.
### Limitations
As the rows will be transposed to columns, a limit is required field as calcite planning phase number of rows are not known.

### Related Issues
Resolves [ #5142 ]

### Check List
- [ x] New functionality includes testing.
- [x ] New functionality has been documented.
 - [x ] New functionality has javadoc added.
- [x ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ x] Commits are signed per the DCO using `--signoff` or `-s`.


asifabashar:
@penghuo @LantaoJin @Swiddis  Please review

noCharger:
@asifabashar Could you please resolve the conflicts?

noCharger (commented):
Thanks @asifabashar. Reviewed at `5ddffaf` against the RFC #5142 direction. Focusing this on the high-level direction / SPL-parity calls; I'll leave the smaller robustness items as inline comments.

**Direction — looks right.** The implementation follows the v1 contract we settled on in the RFC: a SQL-native pivot with explicit `in (...)` values (`GROUP BY x-field` + `MAX(...) FILTER(y=val)`), column names built from the actual `y-name-field` values, and no post-processing. That keeps the command Spark-SQL portable and, importantly, **composable** — the explain output shows a normal rel that downstream commands can still consume, which was @penghuo's main concern about the post-processing alternative. Good.

**One thing to make explicit:** requiring `in (...)` is the main departure from SPL (Splunk derives the pivot columns dynamically from the data). This is a deliberate, RFC-approved v1 scope decision, not a defect — but let's call it out in the docs as a known limitation and open a follow-up issue to track dynamic pivot, so users/readers understand why it differs from SPL.

**SPL-parity points that still need a decision (RFC didn't pin these down):**

1. **Single `y-data-field` column naming.** SPL names the column after the `y-name` value alone (e.g. `200`) for a single data field, and only uses `<agg><sep><val>` when there are multiple data fields. This PR always emits `<agg><sep><val>` (e.g. `avg_balance: 200`). Do we want to match SPL for the single-field case? @penghuo

2. **Hardcoded `MAX` aggregation implies a `stats`-before precondition.** Using `MAX(y-data) FILTER(...)` assumes the input is already reduced to one row per (`x-field`, `y-name-field`). SPL's `xyseries` can operate on non-pre-aggregated rows. If the input isn't pre-aggregated, this silently collapses duplicates (and takes lexicographic max for strings). We should either document this precondition clearly or reconsider the collapse semantics.

3. **Forced ascending sort on `x-field`.** SPL preserves input order (unless `grouped`); this PR always sorts by `x-field`. Minor, but it's a visible output-order change from SPL — intended?

(Also two housekeeping items: please rebase on `main` for the conflicts, and `NewAddedCommandsIT#testUnionUnsupportedInV2` appears to have been overwritten by the new xyseries test — please keep the union regression test too.)

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -1696,4 +1697,36 @@ public UnresolvedPlan visitGraphLookupCommand(OpenSearchPPLParser.GraphLookupCom
...
+  @Override
+  public UnresolvedPlan visitXyseriesCommand(OpenSearchPPLParser.XyseriesCommandContext ctx) {
+    UnresolvedExpression xField = internalVisitExpression(ctx.xField);
+    UnresolvedExpression yNameField = internalVisitExpression(ctx.yNameField);
+
+    // Parse pivot values from IN (...) clause
+    List<String> pivotValues =
+        ctx.xyseriesPivotValues().stringLiteral().stream()
+            .map(s -> StringUtils.unquoteText(s.getText()))
+            .collect(Collectors.toList());
```
noCharger, line 1853: Dedup pivot values

```suggestion
            .map(s -> StringUtils.unquoteText(s.getText()))
            .distinct()
            .collect(Collectors.toList());
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3894,6 +3896,102 @@ static ChartConfig fromArguments(ArgumentMap argMap) {
...
+
+    for (String aggName : yDataFieldNames) {
+      for (String pivotVal : pivotValues) {
+        // Reference pivot output column by its generated name: {value}_{agg}
+        String pivotColName = pivotVal + "_" + aggName;
+        reorderProjections.add(b.field(pivotColName));
+        reorderNames.add(generateColumnName(aggName, pivotVal, separator, format));
+      }
+    }
+
```
noCharger:
```suggestion
    // Fail fast with a clear message if the naming scheme produced collisions
    // (e.g. a format template that omits $VAL$ or $AGG$ with multiple series).
    Set<String> seenNames = new HashSet<>();
    for (String name : reorderNames) {
      if (!seenNames.add(name)) {
        throw new IllegalArgumentException(
            "xyseries produced duplicate output column name '"
                + name
                + "'. Use a format template containing both $AGG$ and $VAL$ so column names"
                + " are unique.");
      }
    }
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3894,6 +3896,102 @@ static ChartConfig fromArguments(ArgumentMap argMap) {
...
+    RexNode axis;
+    if (!SqlTypeUtil.isCharacter(yNameRef.getType())) {
+      RelDataType varchar =
+          rx.getTypeFactory()
+              .createTypeWithNullability(
+                  rx.getTypeFactory().createSqlType(SqlTypeName.VARCHAR), true);
+      axis = rx.makeCast(varchar, yNameRef, true);
+    } else {
+      axis = yNameRef;
+    }
```
noCharger, line 4111:
```suggestion
    // Build the pivot axis - cast to VARCHAR if needed for string comparison
    RelDataType yNameType = yNameRef.getType();
    RexNode axis;
    if (!SqlTypeUtil.isCharacter(yNameRef.getType())) {
       if (!SqlTypeUtil.isAtomic(yNameType)) {
        throw new IllegalArgumentException(
            "xyseries y-name-field must be a scalar type, got: " + yNameType.getSqlTypeName());
      }
      RelDataType varchar =
          rx.getTypeFactory()
              .createTypeWithNullability(
                  rx.getTypeFactory().createSqlType(SqlTypeName.VARCHAR), true);
      axis = rx.makeCast(varchar, yNameRef, true);
    } else {
      axis = yNameRef;
    }
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3894,6 +3896,102 @@ static ChartConfig fromArguments(ArgumentMap argMap) {
...
+    List<String> reorderNames = new ArrayList<>();
+
+    reorderProjections.add(b.field(xFieldName));
+    reorderNames.add(xFieldName);
+
+    for (String aggName : yDataFieldNames) {
+      for (String pivotVal : pivotValues) {
+        // Reference pivot output column by its generated name: {value}_{agg}
+        String pivotColName = pivotVal + "_" + aggName;
+        reorderProjections.add(b.field(pivotColName));
```
noCharger:
```suggestion
        try {
          reorderProjections.add(b.field(pivotColName));
        } catch (IllegalArgumentException e) {
          throw new IllegalStateException(
              "xyseries: expected pivot output column '" + pivotColName + "' not found", e);
        }
```

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3894,6 +3896,102 @@ static ChartConfig fromArguments(ArgumentMap argMap) {
...
+    return expr.toString();
+  }
+
+  private String generateColumnName(
+      String yDataFieldName, String pivotValue, String separator, String format) {
+    if (format != null) {
+      return format.replace("$AGG$", yDataFieldName).replace("$VAL$", pivotValue);
+    }
+    return yDataFieldName + separator + pivotValue;
+  }
```
noCharger, line 4199: Need a single y-data-field names columns by the y-name value alone.

noCharger (commented):
let's also update docs/user/ppl/index.md and fix IT

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
noCharger: Sample code 
```suggestion
  private String generateColumnName(
      String yDataFieldName,
      String pivotValue,
      String separator,
      String format,
      boolean singleDataField) {
    if (format != null) {
      return format.replace("$AGG$", yDataFieldName).replace("$VAL$", pivotValue);
    }
    if (singleDataField) {
      return pivotValue;
    }
    return yDataFieldName + separator + pivotValue;
  }
```

asifabashar: applied the above changes

noCharger:
@asifabashar Can we fix the remaining CI by updating the logical and physical plan?

docs/user/ppl/index.md
```diff
@@ -35,62 +35,63 @@ source=accounts
   The following commands are available in PPL:  
   **Note:** Experimental commands are ready for use, but specific parameters may change based on feedback.
   
-| Command Name | Version Introduced | Current Status | Command Description |
```
noCharger, line 38: can we only insert one item, while keeping the rest?

asifabashar: removed extra formatting

noCharger:
### Performance note: coordinator-side pivot aggregate (non-blocking)

Recording a performance analysis for future reference. This is not a merge blocker; the design is sound for the intended use.

**Plan shape.** The upstream `stats` is pushed into OpenSearch as a composite aggregation and paged (size 1000). The pivot itself (`MAX(dataField) FILTER (y = value)`) runs as a Calcite `EnumerableAggregate` on the coordinator node. It is a hash aggregation keyed by the x-field, so all distinct x-field groups are resident in memory at once, each holding `M x P` accumulator cells (M = number of y-data-fields, P = number of pivot values). `QUERY_SIZE_LIMIT` (default 10000) sits above the sort and trims only the final output rows, so it does not bound this intermediate hash.

**Two bounded axes.**
- Width = `M x P` columns. Self-bounded because every pivot value and data field is an explicit literal the user types in the query. It cannot grow from data. The only ceiling is the query string size and a JVM 64 KB per-method codegen limit on the generated aggregate or calc (same class of limit makeresults hit around 5900 inline units; the exact xyseries threshold is untested but is likely in the low thousands of `M x P`). A very large hand-generated `in (...)` could hit this and surface an opaque Janino codegen error.
- Height = number of distinct x-field values (Gx). Data-driven and not capped at the intermediate stage.

**Worst case.** Peak coordinator memory is approximately `Gx * (M x P) * bytes_per_cell`. For example Gx = 10,000,000, M = 5, P = 20 gives 1e9 cells, roughly 16 GB on a single node.

**Backstop that already exists.** The node-level real-memory circuit breaker samples actual heap and trips before node OOM, aborting the query with a CircuitBreakingException rather than crashing the node. The SQL plugin memory monitor (plugins.query.memory_limit) is a second guard. So the realistic worst case is a failed query on high-cardinality x, not a node crash.

**Suggested follow-ups (optional, not blocking):**
1. Docs: state that xyseries targets low-cardinality x (chart-shaping), consistent with SPL usage; high-cardinality x can trip the memory breaker.
2. Optional plan-time guard on generated column count (`M x P`) that fails fast with a clear message, converting a potential Janino 64 KB codegen crash into a friendly error.
3. Engine-level improvement (separate work): a sorted-input streaming aggregate with limit push-down would cap peak memory at `10000 * (M x P)` instead of `Gx * (M x P)`. This needs input ordered by x and aggregate/limit push-down, so it is out of scope here.

noCharger (approved):
Thanks for the change!

ppl/src/test/java/org/opensearch/sql/ppl/parser/AstBuilderTest.java
```diff
@@ -1827,6 +1828,78 @@ public void testMalformedPipeProducesSyntaxError() {
     plan("source=t | invalidCmd |");
   }
 
+  // Xyseries tests
+
+  @Test
+  public void testXyseriesCommandBasic() {
```
penghuo, line 1834: Add test to verify xyseries command can translate to SQL

docs/user/ppl/index.md
```diff
@@ -89,7 +89,8 @@ source=accounts
 | [nomv command](cmd/nomv.md) | 3.6 | stable (since 3.6) | Converts a multivalue field to a single-value string by joining elements with newlines. |
 | [mvexpand command](cmd/mvexpand.md) | 3.6                | stable (since 3.6)       | Expand a multi-valued field into separate documents (one per value). |
 | [graphlookup command](cmd/graphlookup.md) | 3.6 | experimental (since 3.6) | Performs recursive graph traversal on a collection using a BFS algorithm.|
-  
+| [xyseries command](cmd/xyseries.md) | 3.8 | stable (since 3.8) | Converts row-oriented grouped results into a wide table format suitable for chart visualizations. One field serves as the X axis (row key), one field provides pivot values for generating output column names, and one or more data fields fill the pivoted cells. Only rows matching the explicitly provided pivot values in the `in` clause are included. |
```
penghuo, line 92: stable -> experimental

PR #5347, opened 2026-04-14
[BugFix] Fix SQL CLI integration test flaky failure

### LINKED ISSUE #5335: [BUG] SQL CLI integration test flaky failure due to Gradle incremental compilation race condition
## What is the bug?

The SQL CLI integration test workflow (`sql-cli-integration-test.yml`) intermittently fails with a `NoSuchFileException` during compilation. The failure is non-deterministic — re-running the job usually passes.

### Error

```
Execution failed for task ':sql:protocol:compileJava'.
> Compilation failed; see the compiler output below.
  SimpleJsonResponseFormatter.java:47: error: cannot access ProfileContext
      ProfileMetric formatMetric = QueryProfiling.current().getOrCreateMetric(MetricName.FORMAT);
                                                         ^
  bad class file: .../sql-cli/remote/sql/core/build/classes/java/main/org/opensearch/sql/monitor/profile/ProfileContext.class
      unable to access file: java.nio.file.NoSuchFileException: .../ProfileContext.class
```

### Root cause

The workflow has two steps:

1. **`publishToMavenLocal`** — builds the SQL project under `sql-cli/remote/sql/`, producing `.class` files in the build directory
2. **`./gradlew test -PuseLocalSql=true`** — runs sql-cli tests, but instead of resolving SQL dependencies from Maven Local jars, the sql-cli Gradle build references the SQL source project at `remote/sql/` directly and re-compiles it

When Gradle performs incremental compilation in step 2, it may delete stale `.class` files in `core/build/classes/` before recompiling them. Meanwhile, the `protocol` module compilation is concurrently reading those same `.class` files from the classpath. This race condition causes `NoSuchFileException`.

Re-runs pass because Gradle daemon/cache state varies between attempts — when incremental recompilation is not triggered, there is no race.

### Example failures

- https://github.com/opensearch-project/sql/actions/runs/24233000163/job/70752038338 (PR #5334, attempt 2)

### Suggested fix

Clean the SQL build directory after `publishToMavenLocal` to ensure sql-cli resolves dependencies only from Maven Local jars, not from the stale build output:

```yaml
- name: Build and publish SQL modules to Maven Local
  working-directory: sql-cli/remote/sql
  run: |
    ./gradlew publishToMavenLocal -x test -x integTest
    ./gradlew clean
```

ahkcs:
## Summary

- Fixes intermittent `NoSuchFileException` in the SQL CLI integration test workflow caused by a Gradle incremental compilation race condition
- Adds `./gradlew clean` after `publishToMavenLocal` so sql-cli resolves dependencies from Maven Local jars instead of stale build output

* Resolves #5335

## Test plan

- [x] Verify the `sql-cli-integration-test` workflow passes consistently (no more flaky `NoSuchFileException` failures)

### DIFF
--- a/.github/workflows/sql-cli-integration-test.yml
+++ b/.github/workflows/sql-cli-integration-test.yml
@@ -69,6 +69,7 @@ jobs:
           echo "Building SQL modules from current branch..."
           ./gradlew publishToMavenLocal -x test -x integTest
           echo "SQL modules published to Maven Local"
+          ./gradlew clean
 
       - name: Run SQL CLI tests with local SQL modules
         working-directory: sql-cli


mengweieric (approved):

Swiddis (approved):
i don't fully understand how this fixes the problem but i'll believe it lol

PR #5348, opened 2026-04-14
Use context.measure() for analytics engine profiling

ahkcs:
## Summary
- Use `context.measure()` for all three profiling phases (ANALYZE, EXECUTE, FORMAT) in the analytics engine path, replacing manual metric recording
- Extract `executeQuery()` from `AnalyticsExecutionEngine` that returns `QueryResponse` directly, separating execution work from the listener callback
- Extract `formatResponse()` in `RestUnifiedQueryAction`, separating formatting from the callback
- This ensures each metric is recorded before the next phase starts, so `finish()` snapshots all metrics correctly

## Context
The old approach manually recorded EXECUTE and FORMAT metrics using `ProfileMetric.set()`. This was necessary because `context.measure()` writes the metric in its `finally` block — after the action completes. But `SimpleJsonResponseFormatter` calls `finish()` inside the action (to embed the profile in the JSON response), which snapshots all metrics before the `finally` block runs, resulting in 0ms.

The fix: extract the execution and formatting work into separate methods that **return results** instead of calling callbacks. This way, `context.measure()` wraps only the work — its `finally` block writes the metric before the next phase starts. The listener callback (which triggers `finish()`) is called last, after all three metrics are already recorded:

```java
// ANALYZE — already wrapped inside UnifiedQueryPlanner.plan()
RelNode plan = planner.plan(query);

// EXECUTE — wraps extracted executeQuery() that returns QueryResponse
QueryResponse response = context.measure(EXECUTE, () -> analyticsEngine.executeQuery(plan));

// FORMAT — wraps extracted formatResponse() that returns String
String result = context.measure(FORMAT, () -> formatResponse(response, queryType));

// Callback fires AFTER all metrics are written — finish() snapshots correct values
listener.onResponse(new TransportPPLQueryResponse(result));
```

### DIFF
--- a/core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
+++ b/core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
@@ -25,9 +25,6 @@ import org.opensearch.sql.data.type.ExprType;
 import org.opensearch.sql.executor.ExecutionContext;
 import org.opensearch.sql.executor.ExecutionEngine;
 import org.opensearch.sql.executor.pagination.Cursor;
-import org.opensearch.sql.monitor.profile.MetricName;
-import org.opensearch.sql.monitor.profile.ProfileMetric;
-import org.opensearch.sql.monitor.profile.QueryProfiling;
 import org.opensearch.sql.planner.physical.PhysicalPlan;
 
 /**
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -109,11 +111,13 @@ public void execute(
...
                     CalcitePlanContext planContext = context.getPlanContext();
-                    plan = addQuerySizeLimit(plan, planContext);
-                    analyticsEngine.execute(
-                        plan, planContext, createQueryListener(queryType, listener));
+                    RelNode plan = addQuerySizeLimit(planner.plan(query), planContext);
+                    QueryResponse response =
+                        context.measure(EXECUTE, () -> analyticsEngine.executeQuery(plan));
+                    String result =
+                        context.measure(FORMAT, () -> formatResponse(response, queryType));
+                    listener.onResponse(new TransportPPLQueryResponse(result));
```
dai-chen, line 120: Is this method shared and working for SQL path too?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -219,25 +223,16 @@ private static RelNode addQuerySizeLimit(RelNode plan, CalcitePlanContext contex
...
-                    response.getSchema(), response.getResults(), response.getCursor(), langSpec));
-        transportListener.onResponse(new TransportPPLQueryResponse(result));
-      }
-
-      @Override
-      public void onFailure(Exception e) {
-        transportListener.onFailure(e);
-      }
-    };
+    return formatter.format(
```
dai-chen, line 233: I recall earlier we saw profiling code in formatter. Is that removed already?

core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
```diff
@@ -71,26 +68,25 @@ public void explain(PhysicalPlan plan, ResponseListener<ExplainResponse> listene
...
       listener.onFailure(e);
     }
   }
 
+  /**
+   * Execute a query plan and return the response directly. Separated from the listener-based {@link
+   * #execute(RelNode, CalcitePlanContext, ResponseListener)} so callers can wrap this with {@code
+   * context.measure(EXECUTE, ...)} for profiling.
+   */
+  public QueryResponse executeQuery(RelNode plan) {
```
dai-chen, line 82: Could you check if any other way to stick with the existing execute/explain API? I feel this breaks the `ExecutionEngine` interface.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Yes, `RestUnifiedQueryAction.execute()` is shared by both paths. PPL calls it from `TransportPPLQueryAction` with `QueryType.PPL`, and SQL calls it from `SQLPlugin.createSqlAnalyticsRouter()` with `QueryType.SQL`. The `context.measure()` logic inside works the same for both.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: No, the profiling code in `SimpleJsonResponseFormatter` is still there — it records FORMAT manually and calls finish() to embed the profile in the JSON response. We kept it because this formatter is shared with the existing PPL Calcite path which relies on it.
For the analytics path, both `context.measure(FORMAT)` and the formatter's manual recording write to the same metric. The `finish()` inside the formatter snapshots the correct values since EXECUTE was already recorded by context.measure(EXECUTE) before the formatter runs.

PR #5350, opened 2026-04-15
[BugFix] Fix rename with wildcard applying on hidden fields (#5099)

### LINKED ISSUE #5099: [BUG] `rename` with wildcard should not apply on hidden fields
**What is the bug?**
`search source=%s | rename * as old_*`

```java
    JSONObject result =
        executeQuery(String.format("search source=%s | rename * as old_*", TEST_INDEX_DOG_REMOTE));
    verifyColumn(
        result,
        columnName("old_dog_name"),
        columnName("old_holdersName"),
        columnName("old_age"),
        columnName("old__id"),
        columnName("old__index"),
        columnName("old__score"),
        columnName("old__maxscore"),
        columnName("old__sort"),
        columnName("old__routing"));
```

```
    JSONObject result =
        executeQuery(String.format("search source=%s | rename * as old_*", TEST_INDEX_DOG_REMOTE));
    verifyColumn(
        result,
        columnName("old_dog_name"),
        columnName("old_holdersName"),
        columnName("old_age"));
```

 - Version 3.5.0
 - Plugins

songkant-aws:
### Description
When using `rename * as old_*` (or any wildcard rename without a preceding `fields` command), metadata/hidden fields like `_id`, `_index`, `_score`, `_maxscore`, `_sort`, and `_routing` were incorrectly being matched and renamed. This happened because `WildcardRenameUtils.matchFieldNames` matched all fields in the schema including metadata fields.

**Root cause**: `CalciteRelNodeVisitor.visitRename()` passed all field names (including metadata fields) to the wildcard matching utility without filtering.

**Fix**: After wildcard matching, filter out metadata fields using the existing `isMetadataField()` check. This ensures wildcard renames only apply to user-visible fields, consistent with how `fields` and other commands handle metadata fields.

### Related Issues
Resolves #5099

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -656,6 +656,10 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
       }
 
       List<String> matchingFields = WildcardRenameUtils.matchFieldNames(sourcePattern, newNames);
+      // Exclude metadata fields from wildcard rename (issue #5099)
+      if (WildcardRenameUtils.isWildcardPattern(sourcePattern)) {
+        matchingFields.removeIf(this::isMetadataField);
+      }
 
       for (String fieldName : matchingFields) {
         String newName =
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLRenameIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLRenameIT.java
@@ -200,6 +200,41 @@ public class CalcitePPLRenameIT extends PPLIntegTestCase {
   }
 
   @Test
+  public void testRenameFullWildcardExcludesMetadataFields() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source = %s | rename * as old_*", TEST_INDEX_STATE_COUNTRY));
+    verifySchema(
+        result,
+        schema("old_name", "string"),
+        schema("old_age", "int"),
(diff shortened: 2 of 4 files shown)


songkant-aws:
## Decision Log
**Root Cause**: `CalciteRelNodeVisitor.visitRename()` calls `WildcardRenameUtils.matchFieldNames()` with the full list of field names from the RelNode schema, which includes OpenSearch metadata fields (`_id`, `_index`, `_score`, `_maxscore`, `_sort`, `_routing`). When the source pattern is a wildcard (e.g., `*`), all fields including metadata fields are matched and renamed.

**Approach**: Added a post-filter in `visitRename()` that removes metadata fields from the wildcard match results using the existing `isMetadataField()` method. The filter only applies when the source pattern contains wildcards — explicit field renames (e.g., `rename _id as something`) are unaffected (and already blocked by `Rename.validate()`). This is consistent with how `visitFields` and other visitors handle metadata fields.

**Alternatives Rejected**:
1. Modifying `WildcardRenameUtils.matchFieldNames()` to accept a filter — rejected because the utility is domain-agnostic and shouldn't know about OpenSearch metadata fields.
2. Pre-filtering `newNames` before passing to `matchFieldNames()` — rejected because `newNames` is also used for non-wildcard exact renames and for index-based operations later; filtering it would change semantics.

**Pitfalls**: None encountered. The fix is minimal and only touches the wildcard rename path.

**Things to Watch**: If new metadata fields are added to `OpenSearchConstants.METADATAFIELD_TYPE_MAP`, they will automatically be excluded from wildcard renames. No follow-up needed.

LantaoJin (approved):

qianheng-aws (approved):

PR #5351, opened 2026-04-15
[BugFix] Fix eval overwrites MAP root field when assigning multiple dotted-path names (#5185)

### LINKED ISSUE #5185: [BUG] `eval` overwrites MAP root field when assigning multiple dotted-path names

songkant-aws:
### Description

When using `eval` to materialize multiple dotted-path names from a MAP column (e.g. from `spath`), the second assignment fails with `Field [doc.user.age] not found`. The root cause sits in `shouldOverrideField` in `CalciteRelNodeVisitor.java`, which used a `newName.startsWith(originalName + ".")` prefix match to decide whether to drop the original field. When `eval` creates a column named `doc.user.name`, that check matches `doc` as a prefix and removes the MAP root column, so subsequent eval expressions fail because `doc` no longer exists in the row schema.

### Fix

Remove the prefix branch from `shouldOverrideField` entirely and return the method to an exact-name match. This aligns with SPL1 literal-name semantics: `eval foo.bar = ...` creates a brand-new field literally named `foo.bar`; it must never drop sibling or parent fields with overlapping names.

PR #4606 introduced the prefix branch to make `bin` on a nested field work without an explicit `fields` projection (issue #4482). The diagnosis was wrong — the prefix branch was a "patch on another patch" that silently clobbered any column that happened to be a prefix of an eval target. That same branch is what caused:

* Issue #5185 — a MAP root produced by `spath` got dropped when multiple dotted-path evals ran on top of it.
* Reviewer @qianheng-aws's case on this PR — `source=big5 | fields agent | eval agent.name = "test"` dropped the `agent` column entirely, leaving only `agent.name` in the response.

Both are symptoms of the same over-broad heuristic.

The original command-specific need from #4482 (letting `bin` on a nested leaf collapse the struct-parent columns so the flattened sibling leaves survive the final implicit `fields *` projection) is now handled locally inside `visitBin` via a small helper, `dropStructParentsFor`. That helper only drops parents that actually exist as struct-parent columns in the current row schema, so flat columns and MAP roots from `spath` are untouched.

### Related Issues
Resolves #5185

### Test Plan
* Kept the two spath regression tests that landed earlier on this PR (`CalcitePPLSpathTest`, `CalcitePPLSpathCommandIT`, `integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/5185.yml`).
* Added `CalciteEvalCommandIT.testEvalDottedNameDoesNotDropStructParent` and `testEvalDottedNamePreservesStructParent_ImplicitProject` to cover the reviewer's `agent` case.
* `CalciteBinCommandIT.testBinWithNestedFieldWithoutExplicitProjection` and `testBinWithEvalCreatedDottedFieldName` (from #4606) still pass — now via the scoped `dropStructParentsFor` path rather than the global prefix branch.
* `./gradlew spotlessApply spotlessCheck :core:test :ppl:test` — green.
* Relevant IT suites run locally: `CalciteBinCommandIT`, `CalcitePPLSpathCommandIT`, `CalciteEvalCommandIT`, `CalciteFlattenCommandIT`, `CalciteFlattenDocValueIT`, `CalciteObjectFieldOperateIT` — all green.

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -1295,7 +1295,7 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     List<String> originalFieldNames = context.relBuilder.peek().getRowType().getFieldNames();
     List<RexNode> toOverrideList =
         originalFieldNames.stream()
-            .filter(originalName -> shouldOverrideField(originalName, newNames))
+            .filter(originalName -> shouldOverrideField(originalName, newNames, originalFieldNames))
             .map(a -> (RexNode) context.relBuilder.field(a))
             .toList();
     // 1. add the new fields, For example "age0, country0"
@@ -1315,15 +1315,31 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     context.relBuilder.rename(expectedRenameFields);
   }
 
-  private boolean shouldOverrideField(String originalName, List<String> newNames) {
+  private boolean shouldOverrideField(
+      String originalName, List<String> newNames, List<String> allFieldNames) {
     return newNames.stream()
         .anyMatch(
             newName ->
                 // Match exact field names (e.g., "age" == "age") for flat fields
                 newName.equals(originalName)
                     // OR match nested paths (e.g., "resource.attributes..." starts with
-                    // "resource.")
-                    || newName.startsWith(originalName + "."));
+                    // "resource."), but only when the schema already contains sub-fields of
+                    // originalName. This distinguishes real struct parents (which have flattened
+                    // sub-field columns) from MAP columns produced by spath where dotted names
+                    // are just flat column names, not nested sub-fields.
+                    || (newName.startsWith(originalName + ".")
+                        && hasSubFields(originalName, allFieldNames)));
+  }
+
+  /**
+   * Checks whether the schema contains any field whose name starts with {@code parentName + "."}.
+   * This indicates that {@code parentName} is a real struct parent with flattened sub-field
+   * columns, as opposed to a standalone MAP column (e.g., from spath) that has no sub-fields in the
+   * schema.
+   */
+  private boolean hasSubFields(String parentName, List<String> allFieldNames) {
+    String prefix = parentName + ".";
+    return allFieldNames.stream().anyMatch(name -> name.startsWith(prefix));
   }
 
   private List<List<RexInputRef>> extractInputRefList(List<RelBuilder.AggCall> aggCalls) {
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -973,8 +973,45 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
 
     String alias = node.getAlias() != null ? node.getAlias() : fieldName;
     projectPlusOverriding(List.of(binExpression), List.of(alias), context);
+    dropStructParentsFor(alias, context);
 
     return context.relBuilder.peek();
+  }
+
+  /**
+   * If {@code dottedName} addresses a nested leaf inside a struct that OpenSearch has exposed
+   * through both its struct-parent columns and its flattened leaf columns (e.g. the telemetry
+   * mapping exposes {@code resource}, {@code resource.attributes}, ..., {@code
+   * resource.attributes.telemetry.sdk.version} side-by-side), drop the struct-parent prefixes from
+   * the current row. This keeps a subsequent {@link #tryToRemoveNestedFields(CalcitePlanContext)}
+   * pass from collapsing the flattened leaves back into the parents when the final implicit {@code
+   * fields *} projection runs.
+   *
+   * <p>This preserves the behaviour that issue #4482 originally required for {@code bin} on a
+   * nested field without an explicit {@code fields} projection, scoped to the command that actually
+   * needs it instead of a global prefix-override in {@link #shouldOverrideField} (which broke #5185
+   * and the reviewer's {@code eval agent.name = ...} case).
+   *
+   * <p>No-op when no such struct-parent columns exist (e.g. flat columns or MAP roots from {@code
+   * spath}).
+   */
+  private void dropStructParentsFor(String dottedName, CalcitePlanContext context) {
+    if (dottedName == null || dottedName.indexOf('.') < 0) {
+      return;
+    }
+    List<String> fieldNames = context.relBuilder.peek().getRowType().getFieldNames();
+    List<RexNode> parentsToDrop = new ArrayList<>();
+    int dotIdx = dottedName.indexOf('.');
+    while (dotIdx >= 0) {
+      String prefix = dottedName.substring(0, dotIdx);
+      if (fieldNames.contains(prefix)) {
+        parentsToDrop.add(context.relBuilder.field(prefix));
+      }
+      dotIdx = dottedName.indexOf('.', dotIdx + 1);
+    }
+    if (!parentsToDrop.isEmpty()) {
+      context.relBuilder.projectExcept(parentsToDrop);
+    }
   }
 
   @Override
@@ -1295,7 +1332,7 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     List<String> originalFieldNames = context.relBuilder.peek().getRowType().getFieldNames();
     List<RexNode> toOverrideList =
         originalFieldNames.stream()
-            .filter(originalName -> shouldOverrideField(originalName, newNames, originalFieldNames))
+            .filter(originalName -> shouldOverrideField(originalName, newNames))
             .map(a -> (RexNode) context.relBuilder.field(a))
             .toList();
     // 1. add the new fields, For example "age0, country0"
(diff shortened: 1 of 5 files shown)


songkant-aws:
## Decision Log
**Root Cause**: `shouldOverrideField` in `CalciteRelNodeVisitor.java` used a simple `newName.startsWith(originalName + ".")` prefix check to determine whether to remove the original field. When `eval` creates a column named `doc.user.name`, this check matches `doc` as a prefix, causing the MAP root column `doc` to be removed. The second eval expression `doc.user.age=doc.user.age` then fails because `doc` no longer exists in the schema.

**Approach**: Added a `hasSubFields(originalName, allFieldNames)` guard to the prefix-match heuristic. The prefix override now only triggers when the schema already contains other flattened sub-field columns of the parent (e.g., `resource.attributes.telemetry.sdk.enabled` alongside `resource`). Standalone MAP columns from `spath` have no sub-fields in the schema, so the prefix match no longer fires for them.

**Alternatives Rejected**:
1. **Type-based check (MAP vs ROW)**: Initially attempted to skip prefix matching for `SqlTypeName.MAP` fields. Failed because OpenSearch maps STRUCT types to Calcite MAP (not ROW) via `OpenSearchTypeFactory.convertExprTypeToRelDataType`, making it impossible to distinguish real struct parents from spath MAP columns by type alone.
2. **RexNode-based check**: Considered inspecting the `RexNode` of the new field to see if it accesses the original via ITEM. Too invasive and fragile.

**Pitfalls**: The initial MAP-type check broke `CalciteBinCommandIT.testBinWithNestedFieldWithoutExplicitProjection` and `testBinWithEvalCreatedDottedFieldName` because those tests depend on the prefix override for real OpenSearch struct fields (which are also MAP type in Calcite).

**Things to Watch**: The `hasSubFields` heuristic assumes that a standalone MAP column will never have flattened sub-field columns in the same schema. If a future feature adds both a MAP column and flattened sub-fields simultaneously, this heuristic could produce false positives.

songkant-aws:
This narrow fix resolves the issue with a very specific solution. In the long term, if we can resolve the data type of nested field to be interpreted by Calcite `ROW` type, it could cover more cases.

LantaoJin (dismissed):

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1329,15 +1315,31 @@ private void projectPlusOverriding(
...
                 // Match exact field names (e.g., "age" == "age") for flat fields
                 newName.equals(originalName)
                     // OR match nested paths (e.g., "resource.attributes..." starts with
-                    // "resource.")
-                    || newName.startsWith(originalName + "."));
+                    // "resource."), but only when the schema already contains sub-fields of
+                    // originalName. This distinguishes real struct parents (which have flattened
+                    // sub-field columns) from MAP columns produced by spath where dotted names
+                    // are just flat column names, not nested sub-fields.
+                    || (newName.startsWith(originalName + ".")
```
qianheng-aws: The original fix #4606 also introduced a very general bug -- adding new field with the same name as a child field name will remove the parent field. e.g.
```
source=big5 | fields agent | eval agent.name = "test" 
```
return 
```
{
    "schema": [
        {
            "name": "agent.name",
            "type": "string"
        }
    ],
    "datarows": [
        [
            "test"
        ]
    ],
    "total": 1,
    "size": 1
}
```

while we expect the `agent` field should be included in the output.

Will this fix it as well? Or, I'm thinking, is there a solution to rip-and-replace instead of making patch on another patch?

songkant-aws:
@qianheng-aws thanks — you were right. I rewrote the fix along the rip-and-replace path you suggested in the new commit `a7d35a0a4`:

* `shouldOverrideField` is now exact-name match only; the `startsWith` prefix branch (and the intermediate `hasSubFields` guard I had added) are gone. Assigning `eval foo.bar = ...` no longer touches any sibling or parent field, matching SPL1 literal-name semantics.
* Your `source=big5 | fields agent | eval agent.name = "test"` case is now covered directly by `CalciteEvalCommandIT.testEvalDottedNameDoesNotDropStructParent` — `agent` is preserved in the output.
* The original motivation for #4606 (letting `bin` on a nested leaf flatten the struct-parent so its siblings come through when no explicit `fields` projection is used) is now scoped to `visitBin` itself via `dropStructParentsFor`. The two tests that #4606 shipped (`testBinWithNestedFieldWithoutExplicitProjection`, `testBinWithEvalCreatedDottedFieldName`) still pass via that targeted path instead of via a global prefix-override.

I've updated the PR description to reflect this framing. The spath regression tests from the earlier revision are still there.

songkant-aws:
## Decision Log (update, Phase 3.4)

**New root-cause framing**: PR #4606 diagnosed issue #4482 as requiring a prefix-override in `shouldOverrideField`, so that assigning a dotted-path field would drop struct-parent columns that were a prefix of the new name. That framing was incorrect: the prefix branch also fires for any unrelated column whose name happens to be a prefix of an eval target, including MAP roots from `spath` (issue #5185) and arbitrary flat columns like the reviewer's `source=big5 | fields agent | eval agent.name = "test"`. The behavior that `bin` on a nested leaf actually needs — collapsing the struct-parent columns so the flattened sibling leaves survive the final implicit `fields *` projection — is specific to `bin`, not to `shouldOverrideField` in general.

**Removed**:
* The entire prefix branch in `shouldOverrideField` (the `newName.startsWith(originalName + ".")` match).
* The `hasSubFields` helper introduced in the previous revision of this PR.
* The third `List<String> allFieldNames` parameter on `shouldOverrideField`; `projectPlusOverriding` now calls the two-arg form.

**Added**:
* `dropStructParentsFor(String dottedName, CalcitePlanContext context)` invoked from `visitBin` only. It walks the dotted name, collects any intermediate prefixes that actually exist as columns in the current row schema, and drops them. This is a no-op for flat columns and MAP roots (spath), so it doesn't reintroduce the #5185 / reviewer's regressions.
* `CalciteEvalCommandIT.testEvalDottedNameDoesNotDropStructParent` — the exact reviewer case on the `agent` struct, verifies both `agent` and `agent.name` survive.
* `CalciteEvalCommandIT.testEvalDottedNamePreservesStructParent_ImplicitProject` — complementary coverage without the explicit trailing `fields` projection, guarding the specific regression (that `agent` is preserved).
* A tiny `test_eval_agent` fixture in `CalciteEvalCommandIT.init()` with dynamic mapping (idempotent for repeated `@Before` invocations).

**Alternatives rejected**:
* The prior `hasSubFields` heuristic (first revision of this PR). Correctly handled #5185 but did not fix the reviewer's `agent` case: `agent` has no sibling sub-field columns in the row schema, so the heuristic treated it as "not a real struct parent" and spared it — but only because of that coincidence. The reviewer pointed out this is still a patch-on-patch, and they were right.
* `SqlTypeName.MAP` vs `ROW` discrimination inside `shouldOverrideField`. Doesn't work because OpenSearch `STRUCT` maps to Calcite MAP, so the type alone cannot distinguish a struct parent from a spath MAP root.
* `RexNode`-based introspection of the eval expression to detect ITEM access on the original column. Too invasive; still doesn't address `bin`'s command-specific need.

**Tests that required command-level adjustments**:
* `CalciteBinCommandIT.testBinWithNestedFieldWithoutExplicitProjection` and `testBinWithEvalCreatedDottedFieldName` (originally from PR #4606). Without the global prefix branch, the struct-parent columns (`resource`, `resource.attributes`, …) are no longer removed during `projectPlusOverriding`. `tryToRemoveNestedFields` on the final implicit `fields *` then collapses all the flattened sibling leaves back into the struct parents and drops the freshly-computed bin value. `visitBin` now calls `dropStructParentsFor` immediately after `projectPlusOverriding` to delete those parent columns, so the downstream pass keeps the flattened siblings + the bin value. Both tests pass unchanged.

songkant-aws:
Follow-up (additive commit `ba8120ae4`): empirically verified the reviewer-flagged concern about `tryToRemoveNestedFields` eating eval overrides, and fixed it.

## Phase A — regression matrix (empirical, run against `Index.TELEMETRY`)

Telemetry mapping exposes struct parents (`resource`, `resource.attributes`, ..., `resource.attributes.telemetry.sdk`) side-by-side with the flattened leaves (e.g. `resource.attributes.telemetry.sdk.version`).

| scenario | query | result on prior HEAD (a7d35a0a4) | verdict |
|---|---|---|---|
| (a) eval override, implicit project | `source=<telemetry> \| eval resource.attributes.telemetry.sdk.version = "OVERRIDE"` | row = `[resource-struct-with-original-int, severityNumber]`; the `"OVERRIDE"` string is **gone** | **REGRESSION** |
| (a') eval override, explicit trailing `fields` | same + `\| fields resource.attributes.telemetry.sdk.version` | `"OVERRIDE"` returned as expected | passes |
| (d) bin on nested leaf (sanity) | `source=<telemetry> \| bin resource.attributes.telemetry.sdk.version span=2` | works (covered by `CalciteBinCommandIT#testBinWithNestedFieldWithoutExplicitProjection`) | passes |

(b) trendline on a nested numeric leaf and (c) rex/sed targeting a nested leaf also flow through `projectPlusOverriding`; once the eval case was confirmed as a regression, I moved straight to the fix (Option 1) rather than enumerating every callsite empirically, since the fix is unconditional across all callers.

## Phase B — fix applied (Option 1)

Moved `dropStructParentsFor` inside `projectPlusOverriding`, **gated on "the override actually matched an existing column name"** (i.e. the pre-override row schema already had that exact dotted name).

- Telemetry case: `resource.attributes.telemetry.sdk.version` already exists as a flat leaf → override fires → struct parents pruned → overridden value survives implicit `fields *`. ✅
- Reviewer's `agent` case on #5351 (`fields agent | eval agent.name = 'test'`): there's no pre-existing `agent.name` column in the row → `overriddenNames` is empty → `agent` struct parent survives untouched. ✅
- Issue #5185 (spath MAP root + multi-field eval): spath introduces new dotted names that were not already in the row schema → `overriddenNames` is empty → MAP root `doc` survives. ✅

`shouldOverrideField` is unchanged (still exact-match only). No heuristics on sub-fields or RexNode introspection were introduced.

## Phase C — regression-proof the fix

Added two new ITs in `CalciteEvalCommandIT`:

- `testEvalOverrideOfFlattenedNestedLeafSurvivesImplicitProject` — the primary regression guard; fails on `a7d35a0a4`, passes on `ba8120ae4`.
- `testEvalOverrideOfFlattenedNestedLeafWithExplicitProject` — control for the explicit-projection path.

Verified green on `ba8120ae4`:
- `./gradlew spotlessCheck` ✅
- `./gradlew :core:test :ppl:test` ✅
- `./gradlew :integ-test:integTest --tests CalciteEvalCommandIT --tests CalcitePPLSpathCommandIT --tests CalciteBinCommandIT --tests CalciteFlattenCommandIT --tests CalciteObjectFieldOperateIT --tests CalciteTrendlineCommandIT --tests CalcitePPLTrendlineIT --tests CalciteRexCommandIT --tests CalciteExpandCommandIT --tests CalciteMvExpandCommandIT --tests CalcitePPLPatternsIT` ✅

In particular:
- `testSpathAutoExtractWithMultiFieldEval` / `testSpathAutoExtractWithSeparateEvalCommands` (the #5185 guards) still pass.
- `testEvalDottedNameDoesNotDropStructParent` / `testEvalDottedNamePreservesStructParent_ImplicitProject` (the reviewer's `agent` case) still pass.
- `testBinWithNestedFieldWithoutExplicitProjection` / `testBinWithNestedFieldWithExplicitProjection` / `testBinWithEvalCreatedDottedFieldName` (the #4482 guards) still pass.

## Note on the bin-side `dropStructParentsFor` call

Left the explicit invocation from `visitBin` in place. It is redundant when the alias matches a pre-existing flat leaf (projectPlusOverriding now handles that), but still useful when `bin` is given a custom alias whose string does not match any existing field — the alias is still dotted and the parents should still be pruned. Updated the Javadoc on `dropStructParentsFor` to reflect both call sites.

qianheng-aws (approved):

yuancu (approved):

PR #5352, opened 2026-04-15
[BugFix] Fix transpose command name collision with 'value' field (#5172)

### LINKED ISSUE #5172: [BUG] Transpose command name collision: hardcoded 'value' unpivot column collides with input field named 'value'
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-numbers | where category = "A" | transpose
```
**Expected Result:**
Each transposed row should represent the original field values for that column (for example `count` row should reflect count values, `category` row should reflect category values).

**Actual Result:**
All transposed rows show duplicated values from the original `value` column (for example `10.5`, `20.3`, ...), including rows for `count`, `category`, `subcategory`, and `ts`.

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "ts": { "type": "date" }
    }
  }
}
```
**Sample Data**
```json
[
  { "count": 2, "category": "A", "subcategory": "Y", "value": 20.3, "ts": "2024-01-02" }
]
```

## Bug Description
**Issue Summary:**
`transpose` output is corrupted when input contains a field named `value`.

**Steps to Reproduce:**
1. Create index with mapping above.
2. Insert sample rows.
3. Run the query above.
4. Observe every transposed row using the same numeric series from `value`.

**Impact:**
Transpose results become unusable for downstream reporting.

songkant-aws:
### Description
The `transpose` command's internal unpivot column was hardcoded as `value`, which collides with input fields named `value`. When a field named `value` exists in the input data, all transposed rows incorrectly show duplicated values from that field instead of each column's actual values.

**Root cause**: In `CalciteRelNodeVisitor.visitTranspose()`, the unpivot step uses `ImmutableList.of("value")` as the value column alias, and the subsequent pivot step references `b.field("value")`. When the input has a `value` field, Calcite resolves the ambiguous reference to the wrong column.

**Fix**: Replaced the hardcoded `"value"` with a unique internal name `_value_transpose_` (constant `PlanUtils.VALUE_COLUMN_FOR_TRANSPOSE`), following the same naming convention as `_row_number_transpose_`.

### Related Issues
Resolves #5172

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed


songkant-aws:
## Decision Log
**Root Cause**: The `visitTranspose()` method in `CalciteRelNodeVisitor` hardcodes `"value"` as the unpivot column name (line 926: `ImmutableList.of("value")`). When input data contains a field named `value`, Calcite's field resolution picks the wrong column during the pivot step (`b.field("value")` on line 948), causing all transposed rows to show the same duplicated data from the input's `value` field.

**Approach**: Introduced a unique internal column name `_value_transpose_` (as constant `PlanUtils.VALUE_COLUMN_FOR_TRANSPOSE`) following the existing naming convention used for `_row_number_transpose_`. This avoids collision with any user-defined field name since it uses leading/trailing underscores.

**Alternatives Rejected**: 
- Dynamic name generation (e.g., appending UUID) — overly complex and would break plan determinism for explain/testing.
- Checking if `value` exists in input fields and only renaming then — more fragile and still leaves potential collision if the fallback name also clashes.

**Pitfalls**: None encountered. The fix is mechanical — replace string literal with constant in two places.

**Things to Watch**: The same `_value_transpose_` naming convention means a user field named `_value_transpose_` would still collide. This is astronomically unlikely given the naming pattern, consistent with how `_row_number_transpose_` is handled.

LantaoJin (approved):

qianheng-aws (approved):

PR #5353, opened 2026-04-15
[BugFix] Fix sort order not preserved through dedup in Calcite engine (#3922)

### LINKED ISSUE #3922: [BUG] Order not preserved when deduping columns generated with eval (in v3)
**What is the bug?**

For this query: `source=opensearch-sql_test_index_account| sort state | eval age2 = age + 2 | dedup age2 | fields state, age2`, the output is ought to be sorted by state. However, the order is not preserved in the result:

| State | Age2 |
|-------|------|
| AK    | 32   |
| AK    | 33   |
| AL    | 34   |
| AK    | 35   |
| AL    | 36   |
| AK    | 37   |
| AK    | 38   |
| AK    | 39   |
| AK    | 40   |
| AK    | 41   |
| AK    | 42   |
| AK    | 22   |
| AK    | 23   |
| CA    | 24   |
| AL    | 25   |
| AK    | 26   |
| AK    | 27   |
| AK    | 28   |
| AL    | 29   |
| AK    | 30   |
| AR    | 31   |

The problem can be seen from its physical plan:
```
{
    "calcite": {
        "logical": "LogicalSort(sort0=[$0], dir0=[ASC-nulls-first])\n  LogicalProject(state=[$7], age2=[$17])\n    LogicalFilter(condition=[<=($18, 1)])\n      LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10], _id=[$11], _index=[$12], _score=[$13], _maxscore=[$14], _sort=[$15], _routing=[$16], age2=[$17], _row_number_=[ROW_NUMBER() OVER (PARTITION BY $17 ORDER BY $17)])\n        LogicalFilter(condition=[IS NOT NULL($17)])\n          LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10], _id=[$11], _index=[$12], _score=[$13], _maxscore=[$14], _sort=[$15], _routing=[$16], age2=[+($8, 2)])\n            LogicalSort(sort0=[$7], dir0=[ASC-nulls-first])\n              CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n",
        "physical": "EnumerableCalc(expr#0..2=[{inputs}], expr#3=[1], expr#4=[<=($t2, $t3)], proj#0..1=[{exprs}], $condition=[$t4])\n  EnumerableWindow(window#0=[window(partition {1} order by [1] rows between UNBOUNDED PRECEDING and CURRENT ROW aggs [ROW_NUMBER()])])\n    EnumerableCalc(expr#0..16=[{inputs}], expr#17=[2], expr#18=[+($t8, $t17)], expr#19=[IS NOT NULL($t8)], state=[$t7], age2=[$t18], $condition=[$t19])\n      EnumerableSort(sort0=[$7], dir0=[ASC-nulls-first])\n        CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n"
    }
}
```

The order is distorted after `dedup`, but should be restored afterwards.

In contrast, if I feed in the same query **without using `eval`**, it returns the correct result:

| State | Age |
|-------|-----|
| AK    | 33  |
| AK    | 35  |
| AK    | 36  |
| AK    | 37  |
| AK    | 38  |
| AK    | 39  |
| AK    | 40  |
| AK    | 20  |
| AK    | 21  |
| AK    | 24  |
| AK    | 25  |
| AK    | 26  |
| AK    | 28  |
| AK    | 30  |
| AK    | 31  |
| AL    | 32  |
| AL    | 34  |
| AL    | 23  |
| AL    | 27  |
| AR    | 29  |
| CA    | 22  |

In its physical plan, there is an added sort after `dedup`

```
{
    "calcite": {
        "logical": "LogicalSort(sort0=[$0], dir0=[ASC-nulls-first])\n  LogicalProject(state=[$7], age=[$8])\n    LogicalFilter(condition=[<=($17, 1)])\n      LogicalProject(account_number=[$0], firstname=[$1], address=[$2], balance=[$3], gender=[$4], city=[$5], employer=[$6], state=[$7], age=[$8], email=[$9], lastname=[$10], _id=[$11], _index=[$12], _score=[$13], _maxscore=[$14], _sort=[$15], _routing=[$16], _row_number_=[ROW_NUMBER() OVER (PARTITION BY $8 ORDER BY $8)])\n        LogicalFilter(condition=[IS NOT NULL($8)])\n          LogicalSort(sort0=[$7], dir0=[ASC-nulls-first])\n            CalciteLogicalIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n",
        "physical": "EnumerableSort(sort0=[$0], dir0=[ASC-nulls-first])\n  EnumerableCalc(expr#0..17=[{inputs}], expr#18=[1], expr#19=[<=($t17, $t18)], state=[$t7], age=[$t8], $condition=[$t19])\n    EnumerableWindow(window#0=[window(partition {8} order by [8] rows between UNBOUNDED PRECEDING and CURRENT ROW aggs [ROW_NUMBER()])])\n      EnumerableCalc(expr#0..16=[{inputs}], expr#17=[IS NOT NULL($t8)], proj#0..16=[{exprs}], $condition=[$t17])\n        EnumerableSort(sort0=[$7], dir0=[ASC-nulls-first])\n          CalciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_account]])\n"
    }
}
```

I guess it has something to do with incorrect optimization.

```bash
curl --location --globoff '{{opensearch}}/_plugins/_ppl/_explain' \
--data '{
  "query" : "source=opensearch-sql_test_index_account| sort state | eval age2 = age + 2 | dedup age2 | fields state, age2"
}
'
```

Return the result with the state order being preserved.

 - OS: MacOS
 - Version: 3.1.0 (e8e081af8f251ebcfa87c36f52db24803295c231)
 - Plugins: opensearch-sql

In v2, this is not a problem since its [DedupeOperator](https://github.com/opensearch-project/sql/blob/991458063364399582651540e4d08325172504f2/core/src/main/java/org/opensearch/sql/planner/physical/DedupeOperator.java) is an order-preserving operator.

songkant-aws:
### Description

Fixes a bug where sort order was lost when a `sort` command preceded a `dedup` command in the Calcite execution engine. The root cause was in `PPLDedupConvertRule` which builds a `ROW_NUMBER() OVER (PARTITION BY ...)` window function for dedup. `EnumerableWindow` re-partitions data by the `PARTITION BY` key, destroying any upstream sort order. Simultaneously, Calcite's metadata system (`RelMdCollation`) incorrectly propagated the input's collation through the Window, causing the optimizer to eliminate a post-dedup Sort as "redundant."

The fix strips the input Sort before building the window (breaking the incorrect metadata chain), embeds the sort's collation into the window's ORDER BY (ensuring correct row selection within partitions), and adds an explicit re-sort after the dedup to restore the original order.

### Related Issues
Resolves #3922

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed


songkant-aws:
## Decision Log
**Root Cause**: `PPLDedupConvertRule` builds `ROW_NUMBER() OVER (PARTITION BY field)` without ORDER BY. `EnumerableWindow` physically re-partitions data by the PARTITION BY key, destroying upstream sort order. Meanwhile, Calcite's `RelMdCollation` metadata handler incorrectly reports that the Window preserves input collation, causing the Volcano optimizer to eliminate the outer LogicalSort as "redundant."

**Approach**: Three-pronged fix in `PPLDedupConvertRule.buildDedupNotNull()` and `buildDedupOrNull()`:
1. **Strip** the input Sort before building the window (breaks the incorrect metadata propagation chain)
2. **Embed** the stripped Sort's collation into the window's `ORDER BY` (ensures correct row selection within each partition)
3. **Re-sort** after the dedup filter/project sequence (restores the original sort order that the window destroyed)

**Alternatives Rejected**:
1. *Only add ORDER BY to the window* — Calcite metadata still sees the collation from both the Window's ORDER BY and the input Sort, so the optimizer still eliminates the outer Sort as redundant. Result: data still unsorted.
2. *Only add post-dedup Sort without stripping input Sort* — Same problem: `RelMdCollation` propagates collation metadata through Project-with-Window (reporting the input's collation is preserved), so the optimizer removes the "redundant" outer Sort.
3. *Fix Calcite's `RelMdCollation`* — This is in Calcite core code, not something we control.

**Pitfalls**:
- The metadata system's collation propagation through Window nodes is a known Calcite behavior that makes "just add a Sort" insufficient. The Sort MUST NOT have a collation-providing Sort below the Window, otherwise the optimizer will remove it.
- Stripping the Sort alone (without adding ORDER BY to the window) changes which row gets selected within each partition, since without ORDER BY the row selection is non-deterministic.

**Things to Watch**:
- If Calcite fixes their `RelMdCollation` for Window nodes upstream, the strip-and-restore pattern could be simplified to just adding ORDER BY.
- The `removeSortFromTree` helper only strips pure collation Sorts (no fetch/offset) to avoid interfering with LIMIT semantics.

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -596,6 +599,151 @@ public Void visitCorrelVariable(RexCorrelVariable correlVar) {
...
+   */
+  public static @Nullable RelCollation stripInputSort(RelBuilder relBuilder) {
+    RelNode input = relBuilder.peek();
+    RelCollation collation = findInputCollation(input);
+    if (collation != null && !collation.getFieldCollations().isEmpty()) {
+      // Remap collation field indices through any intermediate Projects between the top
+      // of the tree and the Sort. This is necessary because the Sort's collation references
+      // field indices relative to its own input, which may differ from the top-level schema
+      // after a Project narrows the output fields.
+      collation = remapCollationThroughProjects(input, collation);
```
songkant-aws: Could you check whether we can leverage the mechanism of RelMdCollation to infer the RelNode subtree's output collation first and then strip the sort away? The `remapCollationThroughProjects` seems fragile here.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -748,55 +731,8 @@ public RelNode visitHead(Head node, CalcitePlanContext context) {
...
-      }
-
-      // Continue to child node
-      if (node.getInputs().isEmpty()) {
-        break;
-      }
-      node = node.getInput(0);
-    }
-    return null;
+    return PlanUtils.findInputCollation(node);
```
songkant-aws: If this is a one line refactored util function call, we don't need backtrackForCollation method here. Just replace the backtrackForCollation method with PlanUtils.findInputCollation.

core/src/main/java/org/opensearch/sql/calcite/plan/rule/PPLDedupConvertRule.java
```diff
@@ -47,28 +51,85 @@ public void onMatch(RelOptRuleCall call) {
...
+      int newIdx = currentNames.indexOf(originalNames.get(oldIdx));
+      if (newIdx < 0) {
+        continue;
+      }
+      remapped.add(fc.withFieldIndex(newIdx));
+    }
+    if (remapped.isEmpty()) {
+      return null;
+    }
+    return org.apache.calcite.rel.RelCollations.of(remapped);
```
songkant-aws: Don't use full method name. Use imports. `return RelCollations.of(remapped)`

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
```diff
@@ -596,6 +600,113 @@ public Void visitCorrelVariable(RexCorrelVariable correlVar) {
...
+    List<RelCollation> collations = mq.collations(input);
+    RelCollation outputCollation = null;
+    if (collations != null) {
+      for (RelCollation c : collations) {
+        if (c != null && !c.getFieldCollations().isEmpty()) {
+          outputCollation = c;
+          break;
+        }
+      }
+    }
```
songkant-aws, line 673: In what case we will have multiple RelCollation? If it's not RelFieldCollation, you need to get the first one from composite.

core/src/main/java/org/opensearch/sql/calcite/utils/PPLHintUtils.java
```diff
@@ -81,4 +97,76 @@ public static boolean hasNestedAggCall(Aggregate aggregate) {
...
+
+  /**
+   * Return the dedup sort keys from aggregate hints, preserving the order from the original PPL
+   * {@code sort}. Empty list if not present.
+   */
+  public static List<DedupSortKey> getDedupSortKeys(Aggregate aggregate) {
+    return aggregate.getHints().stream()
+        .filter(hint -> hint.hintName.equals(HINT_AGG_ARGUMENTS))
+        .map(hint -> hint.kvOptions.get(KEY_DEDUP_SORT_FIELDS))
+        .filter(java.util.Objects::nonNull)
```
songkant-aws: Use imports. `Objects::nonNull`

core/src/main/java/org/opensearch/sql/calcite/utils/PPLHintUtils.java
```diff
@@ -81,4 +97,76 @@ public static boolean hasNestedAggCall(Aggregate aggregate) {
...
+        .findFirst()
+        .map(PPLHintUtils::decodeDedupSortFields)
+        .orElse(Collections.emptyList());
+  }
+
+  private static String encodeDedupSortFields(
+      RelCollation collation, java.util.List<String> fieldNames) {
```
songkant-aws: Use imports. `List<String> fieldNames`

core/src/main/java/org/opensearch/sql/calcite/utils/PPLHintUtils.java
```diff
@@ -81,4 +97,76 @@ public static boolean hasNestedAggCall(Aggregate aggregate) {
...
+    StringBuilder sb = new StringBuilder();
+    for (RelFieldCollation fc : collation.getFieldCollations()) {
+      int idx = fc.getFieldIndex();
+      if (idx < 0 || idx >= fieldNames.size()) {
+        // Skip entries we can't map back to a field name (defensive).
+        continue;
```
songkant-aws: Why fail open? It should never happen. Prefer throw exception here.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
```diff
@@ -64,6 +64,12 @@ public void init() throws Exception {
     loadIndex(Index.GRAPH_EMPLOYEES);
   }
 
+  @Override
+  public void testExplain() throws IOException {
+    enabledOnlyWhenPushdownIsEnabled();
+    super.testExplain();
+  }
+
```
songkant-aws: Why override this test here? It should still work even in NoPushdownIT.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -126,6 +143,64 @@ protected void apply(
...
+      List<RexNode> projections = project.getProjects();
+      List<org.apache.calcite.rel.RelFieldCollation> remapped = new ArrayList<>();
+      for (org.apache.calcite.rel.RelFieldCollation fc : collation.getFieldCollations()) {
+        RexNode expr = projections.get(fc.getFieldIndex());
+        if (!(expr instanceof RexInputRef ref)) {
+          return null;
+        }
+        remapped.add(fc.withFieldIndex(ref.getIndex()));
+      }
+      return org.apache.calcite.rel.RelCollations.of(remapped);
```
songkant-aws: Use imports. `return RelCollations.of(remapped)`

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -126,6 +143,64 @@ protected void apply(
...
+    RelCollation collation = dedup.getInputCollation();
+    int projectOutputSize = project.getRowType().getFieldCount();
+    int maxIdx = -1;
+    for (org.apache.calcite.rel.RelFieldCollation fc : collation.getFieldCollations()) {
+      maxIdx = Math.max(maxIdx, fc.getFieldIndex());
+    }
+    if (maxIdx < projectOutputSize) {
+      List<RexNode> projections = project.getProjects();
+      List<org.apache.calcite.rel.RelFieldCollation> remapped = new ArrayList<>();
+      for (org.apache.calcite.rel.RelFieldCollation fc : collation.getFieldCollations()) {
```
songkant-aws: use imports

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -126,6 +143,64 @@ protected void apply(
...
+   * A computed-column sort key (non-{@code RexInputRef}) is not pushable as an OS field sort, so
+   * returns {@code null} in that case. Returns {@code null} also if any sort key cannot be resolved
+   * by either path.
+   */
+  private static @Nullable RelCollation resolveCollationToScanSchema(
+      LogicalDedup dedup, LogicalProject project) {
+    RelCollation collation = dedup.getInputCollation();
+    int projectOutputSize = project.getRowType().getFieldCount();
+    int maxIdx = -1;
+    for (org.apache.calcite.rel.RelFieldCollation fc : collation.getFieldCollations()) {
```
songkant-aws: Use imports

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -126,6 +143,64 @@ protected void apply(
...
+      }
+      return org.apache.calcite.rel.RelCollations.of(remapped);
+    }
+    List<String> originalNames = dedup.getInputCollationFieldNames();
+    if (originalNames == null) {
+      return null;
+    }
+    List<String> scanNames = project.getInput().getRowType().getFieldNames();
+    List<org.apache.calcite.rel.RelFieldCollation> remapped = new ArrayList<>();
+    for (org.apache.calcite.rel.RelFieldCollation fc : collation.getFieldCollations()) {
```
songkant-aws: Use imports

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
```diff
@@ -126,6 +143,64 @@ protected void apply(
...
+      if (oldIdx < 0 || oldIdx >= originalNames.size()) {
+        return null;
+      }
+      int scanIdx = scanNames.indexOf(originalNames.get(oldIdx));
+      if (scanIdx < 0) {
+        return null;
+      }
+      remapped.add(fc.withFieldIndex(scanIdx));
+    }
+    return org.apache.calcite.rel.RelCollations.of(remapped);
```
songkant-aws: Use imports

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
songkant-aws: Addressed in commit 30622499582958dbfd77f48de037f77e88a9d483: `stripInputSort` now calls `RelMetadataQuery.collations(input)` to get the subtree's output collation, so the hand-rolled `remapCollationThroughProjects` has been removed. RelMdCollation already handles the Project index remapping via the Mappings.TargetMapping machinery.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
songkant-aws: Done — `backtrackForCollation` has been removed. The call site at `CalciteRelNodeVisitor.java:816` now invokes `PlanUtils.findInputCollation(currentNode)` directly.

core/src/main/java/org/opensearch/sql/calcite/plan/rule/PPLDedupConvertRule.java
songkant-aws: Done in 3d670ff60.

core/src/main/java/org/opensearch/sql/calcite/utils/PlanUtils.java
songkant-aws: `RelMetadataQuery.collations()` returns `ImmutableList<RelCollation>`, where each `RelCollation` is already a full composite (its own list of `RelFieldCollation`s). A RelNode can provide more than one equivalent ordering simultaneously (e.g. a scan backed by multiple secondary indexes advertises multiple RelCollations). The loop picks the first non-empty RelCollation and uses the whole composite — it does not drop any `RelFieldCollation` entries. In practice this list is almost always 0- or 1-element for our plans; the loop is just a defensive way to skip a leading empty RelCollation if one shows up.

core/src/main/java/org/opensearch/sql/calcite/utils/PPLHintUtils.java
songkant-aws: Done in 3d670ff60.

core/src/main/java/org/opensearch/sql/calcite/utils/PPLHintUtils.java
songkant-aws: Done in 3d670ff60.

core/src/main/java/org/opensearch/sql/calcite/utils/PPLHintUtils.java
songkant-aws: Agreed — changed to `IllegalStateException` in 3d670ff60. The index is resolved against scan row-type upstream in `resolveCollationToScanSchema`, so an out-of-range index here is a code bug, not a runtime condition.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
songkant-aws: You're right — the override was the wrong fix. The real issue was that `calcite_no_pushdown/explain_output.yaml` was stale (still had the pre-fix plan without the post-dedup Sort / ROW_NUMBER ORDER BY). In 3d670ff60 I dropped the override and updated that yaml fixture to the actual post-fix no-pushdown plan, so `testExplain` now runs under both pushdown and no-pushdown modes.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
songkant-aws: Done in 3d670ff60.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
songkant-aws: Done in 3d670ff60.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
songkant-aws: Done in 3d670ff60.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
songkant-aws: Done in 3d670ff60.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/DedupPushdownRule.java
songkant-aws: Done in 3d670ff60.

LantaoJin (approved):

qianheng-aws (approved):

PR #5356, opened 2026-04-15
Add profiling and explain mode support for SQL analytics path

ahkcs:
## Summary
- Add `profile` field and `mode()` method to `SQLQueryRequest`, matching PPL's existing `PPLQueryRequest.profile()` and `PPLQueryRequest.mode()` support
- Replace hardcoded `false` and `ExplainMode.STANDARD` in `SQLPlugin.createSqlAnalyticsRouter()` with `sqlRequest.profile()` and `sqlRequest.mode()`
- Add `"profile"` to `SUPPORTED_FIELDS` so SQL requests with `"profile": true` are not rejected

## Changes
| File | Change |
|------|--------|
| `SQLQueryRequest.java` | Added `profile` field (parsed from JSON body), `mode()` method (derives `ExplainMode` from format param) |
| `SQLPlugin.java` | Use `sqlRequest.profile()` and `sqlRequest.mode()` instead of hardcoded values |
| `AnalyticsSQLIT.java` | Added `testProfileResponseIncludesProfilingData` — verifies summary, analyze, execute, format are all > 0ms |

## Test plan
- [x] Existing `SQLQueryRequestTest` passes
- [x] Existing `AnalyticsSQLIT` tests pass
- [x] `testProfileResponseIncludesProfilingData` — sends `{"query": "SELECT * FROM parquet_logs", "profile": true}` and verifies all profiling metrics are > 0ms

### DIFF
--- a/sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
+++ b/sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
@@ -76,7 +76,11 @@ public class SQLQueryRequest {
     this.format = getFormat(params);
     this.sanitize = shouldSanitize(params);
     this.pretty = shouldPretty(params);
-    this.profile = jsonContent != null && jsonContent.optBoolean("profile", false);
+    this.profile =
+        jsonContent != null
+            && jsonContent.optBoolean("profile", false)
+            && !isExplainRequest()
+            && "jdbc".equalsIgnoreCase(format);
     this.cursor = cursor;
   }
 
--- a/sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
+++ b/sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
@@ -18,6 +18,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.ToString;
 import lombok.experimental.Accessors;
 import org.json.JSONObject;
+import org.opensearch.sql.ast.statement.ExplainMode;
 import org.opensearch.sql.protocol.response.format.Format;
 
 /** SQL query request. */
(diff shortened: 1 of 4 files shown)


sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
```diff
@@ -122,6 +128,11 @@ public boolean isCursorCloseRequest() {
     return path.endsWith("/close");
   }
 
+  /** Get the explain mode from the format parameter. */
+  public ExplainMode mode() {
+    return isExplainRequest() ? ExplainMode.of(format) : ExplainMode.STANDARD;
```
dai-chen, line 137: I thought format is response format and not related to explain mode?

sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
```diff
@@ -55,6 +56,10 @@ public class SQLQueryRequest {
   @Accessors(fluent = true)
   private boolean pretty = false;
 
+  @Getter
+  @Accessors(fluent = true)
+  private boolean profile = false;
```
dai-chen, line 61: Does this only work for new ANSI SQL path right? If so, later we should update doc.

sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
ahkcs: SQL already uses the format parameter as the explain mode in the V2 path: `SQLService.plan()` passes `request.getFormat()` to `AstStatementBuilder`, which creates `new Explain(query, SQL, context.format)` where format is converted to `ExplainMode` via `ExplainMode.of()`. Our mode() follows the same convention.

sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
ahkcs: Yes, only the ANSI SQL path reads `profile()` currently. The V2 SQL path doesn't use it.

dai-chen:
I think this can be closed now?

PR #5358, opened 2026-04-16
[BugFix] Fix PPL mixed text/keyword field type across wildcard indices (#4659)

### LINKED ISSUE #4659: [BUG] PPL cannot properly handle fields with mixed keyword and text data types across multiple indices
**What is the bug?**
PPL cannot properly handle fields with mixed keyword and text data types across multiple indices. Currently, when PPL merges field types, it randomly selects one. For example, if the msg field is mapped as both text and keyword across indices, PPL may select keyword. As a result, when PPL attempts to fetch the msg value from doc_values on each shard, it will return null for documents where msg is defined as a text field, since text fields do not have doc_values.

* Prepare dataset, create 2 indices, msg field type is Text in index1, and keyword in index2.
```
###
PUT {{baseUrl}}/log00001

{
  "mappings": {
    "properties": {
      "msg": {"type": "text"}
    }
  }
}

###
PUT {{baseUrl}}/log00002

{
  "mappings": {
    "properties": {
      "msg": {"type": "keyword"}
    }
  }
}

###

{"index": { "_id": 1 }}
{"msg": "status=200", "index": 1}

###
POST {{baseUrl}}/log00002/_bulk

{"index": { "_id": 1 }}
{"msg": "status=200", "index": 2} 
```

* Run query, we expected get 2 results. but only get 1 results back
```
### Expected

{
  "query": "source=log*"
}

  "datarows": [
    [
      "status=200",
      1
    ],
    [
      "status=200",
      2
    ]
  ]

### Unexpected

{
  "query": "source=log* | rex field=msg 'status=(?<statusCode>\\d+)' | where statusCode=200"
}

  "datarows": [
    [
      "status=200",
      2,
      "200"
    ]
  ]
```

Return 2 results

 - Version [e.g. 22]:  Main
 - Plugins

qianheng-aws:
### Description
When querying wildcard indices where a field has conflicting types (text vs keyword) across indices, the Calcite script engine incorrectly uses DOC_VALUE retrieval for all shards. This causes shard failures on shards where the field is text (no doc_values), resulting in silently dropped documents.

**Root cause**: The `LatestRule` merge strategy non-deterministically picks one type when merging field mappings from multiple indices. When keyword is picked, `RexStandardizer.visitInputRef()` selects DOC_VALUE retrieval via `OpenSearchTextType.toKeywordSubField()`. On shards where the field is actually text, doc_values are not available, causing the script to return null and OpenSearch to silently swallow the shard failure.

**Fix**: Added `TextKeywordConflictRule` to the index mapping merge pipeline (between `DeepMergeRule` and `LatestRule`). When text and keyword types conflict across indices, the rule merges to `OpenSearchTextType` WITHOUT keyword subfields. This forces `_source` retrieval, which works universally across all shards regardless of the actual field type.

### Related Issues
Resolves #4659

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/util/MergeRules/TextKeywordConflictRule.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/util/MergeRules/TextKeywordConflictRule.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.opensearch.util.MergeRules;
+
+import java.util.Map;
+import org.opensearch.sql.opensearch.data.type.OpenSearchDataType;
+import org.opensearch.sql.opensearch.data.type.OpenSearchDataType.MappingType;
+import org.opensearch.sql.opensearch.data.type.OpenSearchTextType;
+
+/**
+ * Merge rule for text/keyword type conflicts across indices. When a field is text in one index and
+ * keyword in another, or text-with-keyword-subfield in one and text-without in another, we merge to
+ * text WITHOUT keyword subfields. This forces _source retrieval instead of doc_values, which works
+ * universally across all shards regardless of the actual field type.
+ *
+ * <p>See GitHub issue #4659.
+ */
+public class TextKeywordConflictRule implements MergeRule {
+
+  @Override
+  public boolean isMatch(OpenSearchDataType source, OpenSearchDataType target) {
+    if (source == null || target == null) {
+      return false;
+    }
+    MappingType sourceMapping = source.getMappingType();
+    MappingType targetMapping = target.getMappingType();
+    if (sourceMapping == null || targetMapping == null) {
+      return false;
+    }
+    // Match when one is text and the other is keyword
+    if (isTextLike(sourceMapping) && isKeyword(targetMapping)) {
+      return true;
+    }
+    if (isKeyword(sourceMapping) && isTextLike(targetMapping)) {
+      return true;
+    }
+    // Match when both are text but one has keyword subfields and the other does not
+    if (isTextLike(sourceMapping) && isTextLike(targetMapping)) {
+      boolean sourceHasKeywordSub = hasKeywordSubField(source);
+      boolean targetHasKeywordSub = hasKeywordSubField(target);
+      return sourceHasKeywordSub != targetHasKeywordSub;
+    }
+    return false;
+  }
+
+  @Override
+  public void mergeInto(
+      String key, OpenSearchDataType source, Map<String, OpenSearchDataType> target) {
+    // Always merge to text WITHOUT keyword subfields.
+    // This forces _source retrieval, which works for both text and keyword fields.
+    target.put(key, OpenSearchTextType.of());
+  }
+
+  private static boolean isTextLike(MappingType mappingType) {
+    return mappingType == MappingType.Text || mappingType == MappingType.MatchOnlyText;
+  }
+
+  private static boolean isKeyword(MappingType mappingType) {
+    return mappingType == MappingType.Keyword;
+  }
+
+  private static boolean hasKeywordSubField(OpenSearchDataType type) {
+    if (type instanceof OpenSearchTextType textType) {
+      return textType.getFields().values().stream()
+          .anyMatch(f -> f.getMappingType() == MappingType.Keyword);
+    }
+    return false;
+  }
+}
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/CalciteNoPushdownIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/CalciteNoPushdownIT.java
@@ -116,6 +116,7 @@ import org.opensearch.sql.ppl.PPLIntegTestCase;
   CalciteNoMvCommandIT.class,
   CalciteMvExpandCommandIT.class,
   CalcitePPLGraphLookupIT.class,
+  CalciteMixedFieldTypeIT.class,
 })
 public class CalciteNoPushdownIT {
   private static boolean wasPushdownEnabled;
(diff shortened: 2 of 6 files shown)


qianheng-aws:
## Decision Log
**Root Cause**: When querying wildcard indices (`source=log*`) where a field (e.g., `msg`) is `text` in one index and `keyword` in another, the `LatestRule` merge strategy non-deterministically picks one type. When `keyword` wins, `RexStandardizer.visitInputRef()` selects DOC_VALUE retrieval. On shards where the field is actually `text` (no doc_values), the script fails silently, and OpenSearch swallows the shard failure with `search.default_allow_partial_results=true` (default), dropping those documents from results.

**Approach**: Added `TextKeywordConflictRule` to the merge rule chain. When text/keyword types conflict across indices, the merged type becomes `OpenSearchTextType` without keyword subfields. This causes `toKeywordSubField()` to return `null`, forcing `_source` retrieval instead of DOC_VALUE. `_source` works universally for both text and keyword fields.

**Alternatives Rejected**:
1. **ScriptDataContext fallback (DOC_VALUE -> SOURCE)**: When `getFromDocValue()` returns null, fall back to `getFromSource()`. Rejected because null from doc_values is indistinguishable from an actual null field value — this would mask genuine nulls.
2. **Fast-fail on type conflict**: Throw an error when incompatible types are detected. Rejected because it breaks valid use cases where users intentionally query across indices with different mappings.
3. **Per-shard type detection**: Have the script engine detect the actual field type on each shard and choose the retrieval strategy accordingly. Rejected as too complex for the current architecture — the script parameters (SOURCES/DIGESTS) are fixed at query planning time, not per-shard.

**Pitfalls**:
- The basic `source=log* | fields msg` query does NOT trigger the bug because it uses `_source` retrieval without script pushdown. Only queries with pushed-down script expressions (e.g., `where upper(msg) = 'X'`) trigger the DOC_VALUE retrieval path.
- OpenSearch silently swallows shard failures by default (`search.default_allow_partial_results=true`), which makes the bug appear as missing results rather than errors.

**Things to Watch**:
- The fix forces `_source` retrieval for conflicting types, which is slower than DOC_VALUE for keyword fields. This is acceptable as correctness trumps performance, and the scenario (conflicting types) is inherently problematic.
- The `LatestRule` still applies for other type conflicts (e.g., integer vs long). Those cases should be evaluated separately if similar issues arise.
- The maintainer's long-term suggestion of per-shard query plan execution would provide a more optimal solution.

opensearch/src/main/java/org/opensearch/sql/opensearch/util/MergeRules/TextKeywordConflictRule.java
```diff
@@ -0,0 +1,72 @@
...
+    if (source == null || target == null) {
+      return false;
+    }
+    MappingType sourceMapping = source.getMappingType();
+    MappingType targetMapping = target.getMappingType();
+    if (sourceMapping == null || targetMapping == null) {
+      return false;
+    }
+    // Match when one is text and the other is keyword
+    if (isTextLike(sourceMapping) && isKeyword(targetMapping)) {
```
qianheng-aws, line 34: If a field is text but has keyword subfield, and the other is keyword, should they match and use text only?

opensearch/src/main/java/org/opensearch/sql/opensearch/util/MergeRules/TextKeywordConflictRule.java
```diff
@@ -0,0 +1,72 @@
...
+      return true;
+    }
+    if (isKeyword(sourceMapping) && isTextLike(targetMapping)) {
```
qianheng-aws, line 37: ditto

opensearch/src/main/java/org/opensearch/sql/opensearch/util/MergeRules/TextKeywordConflictRule.java
qianheng-aws: Yes — this is intentional. Consider the scenario:

- **Index A**: field `msg` is `text` with a keyword subfield (`msg.keyword`)
- **Index B**: field `msg` is `keyword`

If the text-with-keyword-subfield wins the merge, `toKeywordSubField()` returns `msg.keyword`, leading to DOC_VALUE retrieval on the subfield path `msg.keyword`. But on Index B's shards, `msg` is just a `keyword` field — there is no `msg.keyword` subfield. DOC_VALUE retrieval for `msg.keyword` would fail on those shards, producing the same silent data loss as the original bug.

By matching this case and merging to plain text (without keyword subfields), we force `_source` retrieval, which works universally.

The existing test `testMatchTextAndKeyword` (line 25) covers the pure text-vs-keyword case, and this `isTextLike` check intentionally covers text-with-keyword-subfield-vs-keyword as well.

opensearch/src/main/java/org/opensearch/sql/opensearch/util/MergeRules/TextKeywordConflictRule.java
qianheng-aws: Same reasoning as above — the reverse direction (keyword source, text-like target) also needs to match to ensure we always merge to plain text without keyword subfields.

opensearch/src/main/java/org/opensearch/sql/opensearch/util/MergeRules/TextKeywordConflictRule.java
LantaoJin: Q: Curios that could this case be improved to a DSL with full DOC_VALUE retrieval?

qianheng-aws: > Curios that could this case be improved to a DSL with full DOC_VALUE retrieval?

@LantaoJin  Text only field doesn't have doc value.

LantaoJin: My question is for `If a field is text but has keyword subfield, and the other is keyword`

LantaoJin: Discussed offline, the current implementation is sufficient for bugfix. Supporting above enhancement requires additional investigation.

LantaoJin (approved):

yuancu (approved):

PR #5359, opened 2026-04-16
[BugFix] Fix chained streamstats with window causing NPE (#4800)

### LINKED ISSUE #4800: [BUG] Some complex use cases of `streamstats` failed after "Bump Calcite to 1.41.0"
**What is the bug?**
The version of Calcite was updated to 1.41.0 in #4714 .
`Streamstats` command was implemented in #4297 .

These complex use cases worked in the former version of Calcite, but now failed.

```
  "query": """
  source=stream_test_null | streamstats window=2 avg(age) as avg_age by state, country | streamstats window=2 avg(avg_age) as avg_state_age by country
  """
```

```
POST _plugins/_ppl
{
  "query": """
  source=stream_test_null | streamstats reset_before=age>20 avg(age) as avg_age | streamstats reset_before=avg_age>20 window=2 avg(avg_age) as avg_state_age by country
  """
}
```

```
POST _plugins/_ppl
{
  "query": """
  source=stream_test_null | streamstats reset_before=age>22 avg(age) as avg_age  | streamstats window=2 avg(avg_age) as avg_state_age by country
  """
```

It is caused by a NPE. But I still not find the exact reason in the Calcite code.

Set data:
```
POST stream_test_null/_bulk?refresh=true
{"index":{"_id":"1"}}
{"name":"Jake","age":70,"state":"California","country":"USA","year":2023,"month":4}
{"index":{"_id":"2"}}
{"name":"Hello","age":30,"state":"New York","country":"USA","year":2023,"month":4}
{"index":{"_id":"3"}}
{"name":"John","age":25,"state":"Ontario","country":"Canada","year":2023,"month":4}
{"index":{"_id":"4"}}
{"name":"Jane","age":20,"state":"Quebec","country":"Canada","year":2023,"month":4}
{"index":{"_id":"5"}}
{"name":null,"age":10,"state":null,"country":"Canada","year":2023,"month":4}
{"index":{"_id":"6"}}
{"name":"Kevin","year":2023,"month":4}
```
Then do the query.

 - Plugins

qianheng-aws:
### Description
Chained `streamstats` commands with `window` + `group by` caused NPE in Calcite's `RelDecorrelator.createValueGenerator` during query plan preparation. This happened because the `global=true + window + group` path generated `LogicalCorrelate` nodes, and when two such streamstats were chained, the second correlate contained the first inside its right side, creating nested correlates that Calcite's decorrelator couldn't handle.

**Root Cause**: The `buildStreamWindowJoinPlan` method created a `LogicalCorrelate` where both left and right sides shared the same plan tree. When chaining multiple streamstats, nested correlates triggered NPE in `RelDecorrelator.createValueGenerator` at line 1272 (`frame.oldToNewOutputs.get(oldCorVarOffset)` returned null).

**Fix**: Replaced the correlate-based plan with a self-join (`LogicalJoin`) plan for the `global=true + window + group` case. The self-join approach:
1. Builds a minimal right-side projection with prefixed field names to avoid collisions
2. Joins left and right on window frame + group equality conditions
3. Aggregates by all left fields with the window function applied to right-side fields
4. Sorts by the global sequence number and removes helper columns

This avoids generating `LogicalCorrelate` nodes for this code path entirely.

### Related Issues
Resolves #4800

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed


qianheng-aws:
## Decision Log
**Root Cause**: The `global=true + window + group` code path in `visitStreamWindow` used `buildStreamWindowJoinPlan` which created `LogicalCorrelate` nodes. When chaining two streamstats commands (both with window + group by), the second correlate's right side contained the first correlate, creating nested correlates. Calcite 1.41.0's `RelDecorrelator.createValueGenerator` threw NPE because `frame.oldToNewOutputs.get(oldCorVarOffset)` returned null for the nested correlation variable.

**Approach**: Replaced the correlate-based plan with a self-join (`LogicalJoin`) approach for the `global=true + window + group` case. The new `buildStreamWindowSelfJoinPlan` method builds a minimal right-side projection with prefixed field names (`__r_<name>__`), joins on window frame + group conditions, and aggregates by all left fields. This completely avoids `LogicalCorrelate` for this code path.

**Alternatives Rejected**:
- Converting nested correlates to a single correlate: Too complex, would require restructuring the entire plan
- Pre-decorrelating the inner correlate before building the outer one: Calcite's decorrelator API doesn't support partial decorrelation
- Using OVER window functions with `PARTITION BY group ORDER BY __seq__`: Gives `global=false` semantics (window within partition) rather than `global=true` (window over global sequence with group filter)
- Materializing intermediate results: Calcite doesn't support materializing within a logical plan

**Pitfalls**:
- Field name collisions: The self-join duplicates all field names. Solved by prefixing right-side fields with `__r_<name>__`.
- Alias propagation: `Alias.getAlias()` returns null for PPL streamstats (it's a SQL-only field). Must use `Alias.getName()` instead.
- Aggregate function mapping: Built a manual mapping from PPL window function names (AVG, MAX, MIN, etc.) to Calcite aggregate calls, since the existing `aggVisitor.analyze()` would resolve field references against the wrong (left) side of the join.

**Things to Watch**:
- The reset + chained window streamstats case (`streamstats reset_before=... | streamstats window=...`) still fails because the reset path uses correlate and the self-join copies it into the right side. This is tracked with a TODO in the IT class.
- The self-join plan may have different performance characteristics than the correlate plan for large datasets, since it performs a full self-join + aggregate instead of a correlated scan.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -2235,6 +2236,266 @@ private RelNode buildStreamWindowJoinPlan(
...
+          break;
+        }
+      }
+      if (rightFieldIndex == -1) {
+        throw new IllegalArgumentException(
+            "Cannot find aggregate input field '" + rightArgFieldName + "' in right side of join");
+      }
+      RexNode argRef = context.relBuilder.field(rightFieldIndex);
+
+      aggCall =
```
LantaoJin: can this code `aggCall=...` reuse the `CalciteAggCallVisitor`? Or we have to maintain multiple similar logic spots

qianheng-aws: Good call — refactored in 19d74dbb1 to reuse aggVisitor (CalciteAggCallVisitor). Instead of switching on the PPL function name to build AggCalls, the new code rewrites the window function's field references to the prefixed right-side names (SAL -> __r_SAL__), unwraps the WindowFunction to its inner Function, and delegates to aggVisitor.analyze(). This drops ~100 lines of duplicated AVG/SUM/MIN/MAX/COUNT/DC/STDDEV_*/VAR_*/EARLIEST/LATEST mapping and keeps the streamstats self-join path consistent with regular stats/eventstats aggregation handling.

LantaoJin (approved):

songkant-aws (approved):

PR #5363, opened 2026-04-17
[Feature] Add user doc page for SQL vector search

mengweieric:
## Summary
Add a user-facing reference page for the `vectorSearch()` SQL table function, covering syntax, arguments, supported option keys (`k`, `max_distance`, `min_score`, `filter_type`), filter placement semantics, scoring/sort/limit rules, and preview limitations.

## Changes
- New `docs/user/dql/vector-search.rst` with Introduction, Description, Arguments, Syntax, five examples (top-k, radial `max_distance`, radial `min_score`, implicit WHERE pushdown, explicit `filter_type=efficient`), Filtering section, Scoring/Sorting/Limits, and Limitations.

## Test plan
- [x] `./gradlew spotlessCheck` passes
- [x] Doc file intentionally not added to `docs/category.json` — reference page only, no runnable doctest (matches `beyond/fulltext.rst` precedent)
- [x] Examples match enforced syntax from `VectorSearchIT` (quoted `vector='[...]'`, quoted `option='k=5,filter_type=efficient'`)
- [x] Filter section matches `VectorSearchQueryBuilder` behavior: omitted `filter_type` attempts post-style pushdown and falls back to in-memory when the WHERE cannot be translated; explicit `filter_type` (`post` or `efficient`) requires a translatable WHERE or fails
- [x] Option keys noted as lower-case / case-sensitive (matches `ALLOWED_OPTION_KEYS` set-membership check)

### DIFF
--- a/docs/user/dql/vector-search.rst
+++ b/docs/user/dql/vector-search.rst
@@ -31,33 +31,33 @@ Description
 ``vectorSearch(table='<index>', field='<vector-field>', vector='<array>', option='<key=value[,key=value]*>')``
 
 All four arguments are required and must be passed by name as string
-literals. Positional or mixed-positional calls are rejected, for example::
+literals. Positional arguments, or a mix of positional and named
+arguments, are not supported. For example, the following is invalid::
 
-    -- invalid: positional argument
     FROM vectorSearch('my_vectors', field='embedding',
                       vector='[0.1,0.2]', option='k=5') AS v
 
-A table alias is required. Reference projected fields through the alias
-(``v._id``, ``v._score``, ``v.category``).
+A table alias is required. Projected fields are referenced through the
+alias (``v._id``, ``v._score``, ``v.category``).
 
 If the ``opensearch-knn`` plugin is not installed on the target cluster,
 query execution fails with a ``vectorSearch() requires the k-NN plugin``
-error. ``_explain`` still works without the plugin.
+error. ``_explain`` continues to work without the plugin.
 
 Arguments
 ---------
 
 - ``table`` — single concrete index or alias to search. Wildcards
   (``*``), comma-separated multi-index targets, ``_all``, ``.``, and
-  ``..`` are rejected. The target index must have ``index.knn: true`` and
-  map the target field as ``knn_vector``.
+  ``..`` are not supported. The target index must have
+  ``index.knn: true`` and map the target field as ``knn_vector``.
 - ``field`` — name of the ``knn_vector`` field.
 - ``vector`` — query vector as a JSON-style array of numbers, passed as a
   string (for example, ``'[0.1, 0.2, 0.3]'``). Components must be
-  comma-separated finite numbers; semicolon, colon, or pipe separators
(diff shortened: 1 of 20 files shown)


ahkcs:
Is it our plan to convert all the `.rst` file to `.md` file? cc @ritvibhatt

docs/user/dql/vector-search.rst
```diff
@@ -0,0 +1,236 @@
...
+Two placement strategies are available via the ``filter_type`` option:
+
+- ``post`` — the ``WHERE`` predicate is applied as a non-scoring
+  ``bool.filter`` alongside the k-NN query. The k-NN query runs first and
+  its results are then filtered.
+- ``efficient`` — the ``WHERE`` predicate is embedded directly inside the
+  k-NN query (``knn.filter``), enabling pre-filtering during the ANN search.
+  See the `k-NN filtering guide <https://docs.opensearch.org/latest/vector-search/filter-search-knn/efficient-knn-filtering/>`_
+  for engine and method requirements.
+
```
RyanL1997, line 180: The in-memory fallback described here doesn't match the current implementation. In #5331, `VectorSearchQueryBuilder.pushDownFilter()` [always returns `true`](https://github.com/opensearch-project/sql/pull/5331/files#diff-7e1e1e6f4a2e0e7033af6e5c5e4e94bce3040e8e58a34b1b14a3c0d5d08d3f76R87), so the optimizer always removes the `LogicalFilter` node from the plan. And `FilterQueryBuilder.build()` never signals "can't translate" — for expressions without a native Lucene mapping, it builds a `ScriptQueryBuilder` (painless script evaluated on the OpenSearch side), not an in-memory fallback in the SQL engine. If even the script path fails, it throws — the query fails, it doesn't fall back.

docs/user/dql/vector-search.rst
```diff
@@ -0,0 +1,236 @@
...
+- Results are returned in ``_score DESC`` order by default. The only
+  supported ``ORDER BY`` expression is ``<alias>._score DESC``.
+- In top-k mode (``k=N``), ``LIMIT n`` is optional; when present, ``n`` must
+  be ``≤ k``.
+- In radial mode (``max_distance`` or ``min_score``), ``LIMIT`` is required.
+
+Limitations
+===========
+
+The following are not part of the ``vectorSearch()`` preview contract and
```
RyanL1997: Minor coordination note: #5362 (same feature branch, also open) adds active rejection for GROUP BY / aggregations via `VectorSearchIndexScanBuilder.pushDownAggregation`. If that PR merges alongside this one, "are not validated" would be inaccurate — they'd be actively rejected with a descriptive error. May want to update to "are rejected" or "are not supported" once #5362 lands.

dai-chen (approved):
Just wonder is this feature experimental or not? I see we mention preview in the doc.

RyanL1997 (commented):
Hi @mengweieric , thanks for the doc change. Left 2 comments which are actually being carried over from some previous PRs you worked on in this feature branch.

ahkcs (approved):

PR #5365, opened 2026-04-17
Split out CI assemble and lint steps

Swiddis:
### Description
By moving assemble to its own step we can cache dependencies and will be less regularly affected by transient network issue that block dependency downloads. Linting is just nice-to-have in its own thing IMO since if we forget to fmt, we don't necessarily want to have all insights on failed build blocked.

### Related Issues
N/A

### Check List


Swiddis:
Flipping out of draft to RFC a few bits:
- Is removing the container shell ok? This takes ~2 minutes of CI time for every CI action and doesn't seem to immediately give benefit, it also just makes permissions management harder
- Dependency cache could be unstable with snapshot dependencies but not sure, tempted to just try it and see what breaks, I think it'll overall break less than the current system where we re-download dependencies for every run (seeing a lot more flaky network issues). If it's a consistent problem we have options to reduce the TTL for snapshots (e.g. to 0)
- Useful at all to separate compilation? Could just keep the same system as we have today, would prevent waiting for compilation ahead of time but potentially add more race conditions for dependency caching and make issues less obvious to debug at a glance

RyanL1997 (dismissed):
nice change.

.github/workflows/sql-test-and-build-workflow.yml
```diff
@@ -19,44 +19,98 @@ on:
...
+
+  compile-windows-macos:
+    strategy:
+      fail-fast: false
+      matrix:
+        entry:
+          - { os: windows-latest, java: 21, os_build_args: -PbuildPlatform=windows }
+          - { os: windows-latest, java: 24, os_build_args: -PbuildPlatform=windows }
+          - { os: macos-latest, java: 21, os_build_args: '' }
+          - { os: macos-latest, java: 24, os_build_args: '' }
```
penghuo: should be jdk 25 instead of 24?

RyanL1997 (approved):

PR #5373, opened 2026-04-20
[BugFix] Validate table name and reject duplicate or positional args in vectorSearch()

mengweieric:
## Summary
* Apply the same SAFE_FIELD_NAME regex validation to `table=` that already guards `field=`. Previously `table='idx"; DROP'` reached the native layer unvalidated.
* Reject duplicate named arguments with a clean 400 error. Previously passing the same name twice (e.g. `table='a', table='b'`) crashed the server with 500 ArrayIndexOutOfBoundsException.
* Reject positional arguments. Previously mixing a positional with named args (e.g. `vectorSearch('idx', field=...)`) silently returned 200 with an empty DSL and zero rows.

Closes tracker rows 4, 5, 6 in the Vector Search Tracker.

## Test plan
* [ ] Unit tests cover all three validation paths.
* [ ] Integration tests assert 400 status and user-facing error messages.
* [ ] spotlessCheck passes.

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
@@ -301,7 +301,7 @@ public class VectorSearchIT extends SQLIntegTestCase {
     assertThat(explain, containsString("wrapper"));
   }
 
-  // ── Argument shape validation (PR B) ──────────────────────────────────
+  // ── Argument shape validation ─────────────────────────────────────────
 
   @Test
   public void testInvalidTableNameRejected() throws IOException {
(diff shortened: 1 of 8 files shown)


integ-test/src/test/java/org/opensearch/sql/sql/VectorSearchIT.java
```diff
@@ -300,4 +300,138 @@ public void testExplainWithoutKnnPluginStillWorks() throws IOException {
 
     assertThat(explain, containsString("wrapper"));
   }
+
+  // ── Argument shape validation (PR B) ──────────────────────────────────
```
RyanL1997: May need to remove this.

RyanL1997 (approved):

Swiddis (approved):

PR #5374, opened 2026-04-20
[BugFix] Reject OFFSET, WHERE on _score, and script subtrees under filter_type=efficient on vectorSearch()

mengweieric:
## Summary
* Reject OFFSET on vectorSearch(). Previously OFFSET silently rewrote the search window and dropped top results.
* Reject WHERE predicates referencing the synthetic `_score` column. Previously these pushed a range query on a non-existent field and silently returned 0 rows. Users should use `option='min_score=...'` instead.
* Reject script subtrees when `filter_type=efficient`. Previously WHERE predicates that compile to script queries (arithmetic, function calls, CASE, date math) were embedded under `knn.filter`, which is incompatible with AOSS/serverless vector collections.

## Test plan
* [x] Unit tests cover all three rejection paths.
* [x] Integration tests assert 400 status and user-facing error messages.
* [x] spotlessCheck passes.

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
@@ -6,14 +6,27 @@
 package org.opensearch.sql.opensearch.storage.scan;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.ConstantScoreQueryBuilder;
+import org.opensearch.index.query.ExistsQueryBuilder;
+import org.opensearch.index.query.MatchBoolPrefixQueryBuilder;
+import org.opensearch.index.query.MatchPhrasePrefixQueryBuilder;
+import org.opensearch.index.query.MatchPhraseQueryBuilder;
+import org.opensearch.index.query.MatchQueryBuilder;
+import org.opensearch.index.query.MultiMatchQueryBuilder;
+import org.opensearch.index.query.NestedQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.index.query.QueryStringQueryBuilder;
+import org.opensearch.index.query.RangeQueryBuilder;
 import org.opensearch.index.query.ScriptQueryBuilder;
+import org.opensearch.index.query.SimpleQueryStringBuilder;
+import org.opensearch.index.query.TermQueryBuilder;
+import org.opensearch.index.query.WildcardQueryBuilder;
 import org.opensearch.sql.ast.tree.Sort;
 import org.opensearch.sql.ast.tree.Sort.SortOption;
 import org.opensearch.sql.exception.ExpressionEvaluationException;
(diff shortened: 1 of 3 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
```diff
@@ -151,6 +185,74 @@ private void validateLimitWithinK(int limit) {
...
+        if (containsScriptQuery(child)) {
+          return true;
+        }
+      }
+      return false;
+    }
+    if (qb instanceof ConstantScoreQueryBuilder) {
+      return containsScriptQuery(((ConstantScoreQueryBuilder) qb).innerQuery());
+    }
+    return false;
```
RyanL1997: The walker handles `BoolQueryBuilder` and `ConstantScoreQueryBuilder`, and falls through to `return false` for any other wrapper type. This is correct today because `FilterQueryBuilder` (the only source of `whereQuery` here) only produces those shapes plus leaf queries and `ScriptQueryBuilder`.

The concern: if `FilterQueryBuilder` later adds a new wrapper type (e.g. `NestedQueryBuilder`, `FunctionScoreQueryBuilder`, `DisMaxQueryBuilder`), a `ScriptQueryBuilder` nested inside would silently slip past this guard and be embedded under `knn.filter`, producing exactly the AOSS failure this PR is trying to prevent.

Two options to make this future-proof:
1. Fail closed on unknown wrapper types — `throw new ExpressionEvaluationException("unexpected QueryBuilder type " + qb.getClass().getSimpleName() + " in filter_type=efficient path; refusing to embed")` instead of `return false`. Verbose, but safe.
2. Leave the `return false` but add a unit test on `FilterQueryBuilder` that locks in "all shapes produced by build() must be leaf | BoolQuery | ConstantScoreQuery | ScriptQuery" so future additions fail loudly in tests.

Either is fine — just noting that the current shape carries a "quietly re-introduces the bug" risk if `FilterQueryBuilder` evolves.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
```diff
@@ -105,6 +131,14 @@ public boolean pushDownFilter(LogicalFilter filter) {
 
   @Override
   public boolean pushDownLimit(LogicalLimit limit) {
+    // OFFSET on vectorSearch() silently rewrites the search window and drops top results, which
+    // defeats the entire point of a relevance-ranked top-k query. The parent path would push
+    // `from: <offset>` into the OpenSearch request; reject it explicitly so users get a clear
+    // error instead of surprising result shifts.
+    if (limit.getOffset() != null && limit.getOffset() != 0) {
```
RyanL1997, line 139: Minor: `limit.getOffset()` — is this Integer (boxed) or int (primitive)? If `LogicalLimit.getOffset()` returns primitive `int`, the `!= null` check is always true and dead. Worth verifying: if boxed, keep the null-check for defensive coding; if primitive, drop it for readability.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/VectorSearchQueryBuilder.java
mengweieric: Agreed, updated. Efficient mode now uses a fail-closed validator instead of a best-effort script walker: scripts are rejected, nested         
  predicates are rejected for the P0 preview, known safe leaf query builders are allowed, known containers are recursively checked, and unknown
  query shapes fail closed. Added tests for nested rejection, safe leaf trees, exists, and unknown wrappers.

PR #5382, opened 2026-04-23
[BugFix] Fix COALESCE(null, int) returning string type (#5175)

### LINKED ISSUE #5175: [BUG] COALESCE(null, 42) returns wrong type string instead of int due to leastRestrictive type inference

songkant-aws:
### Description
COALESCE(null, 42) (and the mirrored COALESCE(42, null)) currently return
the string "42" with schema type `string`, instead of the integer 42.

**Root cause.** Inside a COALESCE call, any identifier that doesn't resolve
to a field is replaced with a null literal by `QualifiedNameResolver.replaceWithNullLiteralInCoalesce`.
PPL has no dedicated NULL token, so the literal `null` is parsed as an
identifier and hits the same path. The substitute literal was being created
as `SqlTypeName.VARCHAR`, so Calcite's `leastRestrictive` computation for
the COALESCE return type picked VARCHAR, and `EnhancedCoalesceFunction`
coerced every non-null operand to a string.

**Fix.** Use `SqlTypeName.NULL` for the substitute literal. `leastRestrictive`
then promotes NULL to whatever other operand type is present, so
COALESCE(null, 42) infers INTEGER and returns the numeric 42.

When every operand is missing/null, the result type is now `undefined`
(previously it was `string`), which is more accurate; the existing all-null
test was updated to reflect that.

A related unit test for `nomv does_not_exist` previously relied on the
planner throwing because VARCHAR was being passed to `ARRAY_COMPACT`. With
the null literal's type now promotable, the plan builds successfully and
the column evaluates to COALESCE's empty-string fallback; the test was
updated to assert the new plan shape instead of an exception.

### Related Issues
Resolves #5175

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed
- [ ] Integration tests passed (draft; CI to verify)

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/QualifiedNameResolver.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/QualifiedNameResolver.java
@@ -317,9 +317,12 @@ public class QualifiedNameResolver {
   private static Optional<RexNode> replaceWithNullLiteralInCoalesce(CalcitePlanContext context) {
     log.debug("replaceWithNullLiteralInCoalesce() called");
     if (context.isInCoalesceFunction()) {
+      // Use SqlTypeName.NULL so the resulting literal does not bias the least-restrictive
+      // common-type computation toward VARCHAR. See issue #5175: previously VARCHAR was used,
+      // which caused COALESCE(null, 42) to be inferred as VARCHAR and returned as "42".
       return Optional.of(
           context.rexBuilder.makeNullLiteral(
-              context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.VARCHAR)));
+              context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.NULL)));
     }
     return Optional.empty();
   }
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteNoMvCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteNoMvCommandIT.java
@@ -209,24 +209,15 @@ public class CalciteNoMvCommandIT extends PPLIntegTestCase {
   }
 
   @Test
-  public void testNoMvMissingFieldShouldReturn4xx() throws IOException {
-    ResponseException ex =
-        Assertions.assertThrows(
-            ResponseException.class,
-            () -> executeQuery("source=" + TEST_INDEX_BANK + " | nomv does_not_exist"));
+  public void testNoMvMissingField() throws IOException {
+    // After issue #5175 was fixed, unresolved identifiers inside COALESCE (the rewrite target
+    // of `nomv`) resolve to a null literal of SqlTypeName.NULL. Calcite can then promote that
+    // null to the array type expected by ARRAY_COMPACT, so the query succeeds instead of
+    // returning 4xx. The resulting column is the COALESCE empty-string fallback.
+    JSONObject result =
+        executeQuery("source=" + TEST_INDEX_BANK + " | nomv does_not_exist | head 1");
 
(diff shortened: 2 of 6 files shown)


songkant-aws:
## Decision Log
**Root Cause**: `QualifiedNameResolver.replaceWithNullLiteralInCoalesce` substituted unresolved identifiers inside COALESCE with a null literal typed as `SqlTypeName.VARCHAR`. In PPL, the keyword `null` is not a dedicated lexer token, so `coalesce(null, 42)` parses as `coalesce(QualifiedName("null"), 42)` and falls through the same path. Because VARCHAR is wider than INTEGER in `leastRestrictive`, the COALESCE return type came out as VARCHAR, and `EnhancedCoalesceFunction.enhancedCoalesceWithType` dispatched through the VARCHAR branch that wraps `String.valueOf(...)`, turning 42 into "42".

**Approach**: Change the substitute literal's type to `SqlTypeName.NULL`. Calcite's `leastRestrictive` then yields the other operand's type (e.g. INTEGER for `(NULL, INTEGER)`), so the return type — and therefore the runtime coercion — stays numeric. For the all-null case, the common type is NULL which maps to `ExprCoreType.UNDEFINED` / PPL schema "undefined"; the JSON value returned is still `null`, so this is strictly more accurate than the prior VARCHAR fiction.

**Alternatives Rejected**:
- Distinguishing "user wrote `null`" from "user wrote an unresolved identifier" in the parser (long-term fix the issue author also suggested). Rejected for this patch because it requires grammar changes and AST redesign; the short-term type fix addresses the reported bug cleanly.
- Hardcoding INTEGER as the fallback. Rejected — it would break `coalesce(missing_field, 'str')` and any other non-numeric case.
- Patching `EnhancedCoalesceFunction.getReturnTypeInference` to filter NULL-typed operands. Rejected — Calcite's `leastRestrictive` already does the right thing when given NULL-typed operands, so fixing the producer is the minimal change.

**Pitfalls**:
- `CalcitePPLNoMvTest.testNoMvNonExistentField` previously depended on the planner throwing because VARCHAR cannot feed into `ARRAY_COMPACT`. With the new NULL type, Calcite promotes the literal to `null:ANY ARRAY` and the plan builds; the test was updated to assert the new plan shape.
- `CalcitePPLEnhancedCoalesceTest.testCoalesceWithAllNonExistentFields` and the matching IT now report schema `undefined` instead of `string`. Data rows still produce `null`, so no behavioural regression for clients that only read the value.
- Schema-sensitive clients (e.g. BI drivers) that relied on all-null COALESCE columns being reported as strings will now see `undefined`. If that shows up in practice, a narrow workaround is to keep the old VARCHAR behaviour but only when every operand is NULL-typed.

**Things to Watch**:
- The long-term fix (proper NULL literal in PPL grammar / AST type system) is still warranted — the issue author called this out as a short-term fix. Once PPL parses `null` as a typed null literal, the COALESCE resolver hack here can be dropped.
- Any other place that assumes unresolved COALESCE operands are VARCHAR (I searched `isInCoalesceFunction` and only found this single production site).

LantaoJin (approved):

qianheng-aws (approved):

Issue #5346, opened 2026-04-14
[FEATURE] Unified SQL language across OpenSearch, Flint and Spark SQL


dai-chen:
**Is your feature request related to a problem?**

OpenSearch SQL is currently fragmented across two repositories and four engines. This fragmentation creates the same class of problems identified in the PPL unification effort (opensearch-project/opensearch-spark#1136): semantic divergence, engineering duplication, feature lag, and missed optimization opportunities.

| Repository | Engine | Scope |
|-----------|--------|-------|
| `opensearch-project/sql` | Legacy (NLPchina) | JOINs, UNIONs, subqueries |
| `opensearch-project/sql` | V2 | Simple queries, improved correctness |
| `opensearch-project/sql` | Calcite-based ANSI SQL (new) | Analytics engine integration, ANSI SQL compliance |
| `opensearch-project/opensearch-spark` | Spark SQL + Flint SQL | Spark queries, secondary index DDL |

> A gap analysis that runs the full V2 + Legacy integration test suite (807 queries) against the new Calcite engine shows a 67.9% pass rate, with 259 failures across 13 categories. This quantifies the distance between the current engines and a unified SQL baseline. See https://github.com/opensearch-project/sql/issues/5346#issuecomment-4247257399 below for the full breakdown.

**What solution would you like?**

Define a **Unified SQL Language** specification — an ANSI SQL core with internal language extensions that model capabilities such as custom identifier/parsing rules, relevance search and Flint DDL — and implement it once in the Calcite-based engine for reuse across both the OpenSearch and Spark.

```
Unified SQL Language
├── ANSI SQL Core
├── Search Extension (full-text search, vector search, identifier rules)
└── Secondary Index Extension (Flint DDL)
```

**Phase 1: Analytics engine integration**
- Core DQL — SELECT, subqueries, CTEs, window functions
- Explain support
- Full-text search functions

**Phase 2: Full language unification**
- V2/Legacy backward compatibility (identifier rules, type mappings, syntax gaps)
- Vector search (#5320)
- Flint DDL (skipping index, covering index, materialized view)
- Spark SQL alignment (via ANSI mode)
- Others: prepared statement, cursor/pagination

> See https://github.com/opensearch-project/sql/issues/5346#issuecomment-4247262252 below for the full design and roadmap.

**What alternatives have you considered?**

- Keep all engines permanently: Maintenance burden of 4 code paths, inconsistent behavior, no convergence.
[...]

PR #5383, opened 2026-04-23
feat(api): register SqlLibrary operators in UnifiedQueryContext

vamsimanohar:
### Description

`UnifiedQueryContext` currently registers only `SqlStdOperatorTable` and the `RELEVANCE` operator table with the Calcite validator. This means SQL queries submitted through `UnifiedQueryPlanner` cannot use library functions that live outside the SQL standard — notably the **3-argument `REGEXP_REPLACE(str, pattern, replacement)`** used by ClickBench and many real-world PostgreSQL / BigQuery queries.

Example failure before this change:

```
ClickBench Q29:
SELECT REGEXP_REPLACE(referer, '^https?://(?:www\.)?([^/]+)/.*$', '\1') AS k, ...
-> validation_exception: No match found for function signature REGEXP_REPLACE(<CHARACTER>, <CHARACTER>, <CHARACTER>)
```

### Fix

Chain `SqlLibraryOperatorTableFactory.INSTANCE.getOperatorTable(EnumSet.of(POSTGRESQL, MYSQL, BIG_QUERY, SPARK))` into the operator-table chain at validator construction time (`api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java#buildFrameworkConfig`).

The same four-library set is already used elsewhere in the codebase (`RelJsonSerializer.java` deserialization) and by the PPL path which resolves functions directly to `SqlLibraryOperators.REGEXP_REPLACE_3` (`PPLFuncImpTable.java`). This brings the SQL path in line with PPL's function surface.

### Diff scope

1 file, 10 insertions / 1 deletion.

### Functions enabled

- \`REGEXP_REPLACE(str, pattern, replacement)\` — 3-arg form
- \`SAFE_DIVIDE\`, \`REGEXP_CONTAINS\`, \`CONCAT_WS\`, \`REVERSE\`, \`RIGHT\`
- All other PostgreSQL / MySQL / BigQuery / Spark library functions already depended on by PPL

### Tests

- \`:api:test\` — \`UnifiedQueryContextTest\` (7/7) and \`UnifiedQueryPlannerTest\` / \`UnifiedQueryPlannerSqlTest\` (14/14) pass
- \`:api:assemble :core:assemble :ppl:assemble\` — BUILD SUCCESSFUL
- No ambiguous-overload errors from chained libraries

### Rationale

Calcite's \`SqlLibraryOperators.REGEXP_REPLACE_3\` lives in library tables, not in \`SqlStdOperatorTable.instance()\`. Enabling the library operator table is the standard Calcite idiom and matches established project convention.

### DIFF
--- a/api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
+++ b/api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
@@ -23,6 +23,8 @@ import org.apache.calcite.plan.RelTraitDef;
 import org.apache.calcite.rel.metadata.DefaultRelMetadataProvider;
 import org.apache.calcite.schema.Schema;
 import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.sql.fun.SqlLibrary;
+import org.apache.calcite.sql.fun.SqlLibraryOperatorTableFactory;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParser;
 import org.apache.calcite.sql.util.SqlOperatorTables;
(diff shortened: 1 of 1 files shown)


dai-chen:
Hi @vamsimanohar, I've raised https://github.com/opensearch-project/sql/pull/5419 to register the 3 missing functions (`LENGTH`, `REGEXP_REPLACE`, `DATE_TRUNC`) identified in the ClickBench test. For other functions, I think we need to be cautious about what we expose — please see my thoughts regarding the [Function Registration Criteria](https://github.com/opensearch-project/sql/issues/5346#issuecomment-4401294659) for the criteria we're following. Let me know if any missing function block critical benchmark or test. Thanks!

PR #5384, opened 2026-04-23
[BugFix] Surface dedicated error for wildcard or multi-target table in vectorSearch()

mengweieric:
## Description

Before this change, a table name containing a wildcard (`*`) or multi-target pattern (`a,b`) was rejected by `vectorSearch()`, but the user saw the generic fallback message:

> "Invalid table name '...': must contain only alphanumeric characters, dots, underscores, or hyphens"

That message hides the real constraint. `vectorSearch()` requires a single concrete index because top-k semantics, dimension checks, and the embedded filter JSON are not defined across heterogeneous shards.

This PR adds a dedicated check in `validateTableName()` that runs *before* the generic character-class regex and surfaces:

> "Invalid table name '...': vectorSearch() requires a single concrete index or alias; wildcards ('*') and multi-target patterns (comma-separated) are not supported"

No DSL or execution-path change. Accepted inputs are unchanged — simple names, dotted names, hyphens, and underscores all still pass.

## Testing

Unit coverage in `VectorSearchTableFunctionImplementationTest`:

- `applyArguments_rejectsWildcardTableWithDedicatedMessage` (`sql_vector_*`)
- `applyArguments_rejectsBareStarTableWithDedicatedMessage` (`*`)
- `applyArguments_rejectsMultiTargetTableWithDedicatedMessage` (`idx_a,idx_b`)
- `applyArguments_rejectsMidNameStarTable` (`foo*bar`)

Integration coverage in `VectorSearchIT`:

- `testWildcardTableRejectedWithDedicatedMessage`
- `testMultiTargetTableRejectedWithDedicatedMessage`

Local checks:

- `./gradlew :opensearch:test --tests "*VectorSearchTableFunctionImplementationTest*"` passes.
- `./gradlew spotlessCheck :integ-test:compileTestJava` passes.
- `./gradlew :integ-test:integTest -Dtests.class="org.opensearch.sql.sql.VectorSearchIT"` passes.

## Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchTableFunctionImplementation.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchTableFunctionImplementation.java
@@ -203,6 +203,19 @@ public class VectorSearchTableFunctionImplementation extends FunctionExpression
    * its own error message.
    */
   private void validateTableName(String tableName) {
+    // Flag wildcard and multi-target patterns with a dedicated message before the generic
+    // character-class check, so users get an actionable hint instead of "must contain only
+    // alphanumeric characters, dots, ...". vectorSearch() targets a single index because
+    // top-k semantics, dimension checks, and the embedded filter JSON are not defined across
+    // heterogeneous shards.
+    if (tableName.indexOf('*') >= 0 || tableName.indexOf(',') >= 0) {
+      throw new ExpressionEvaluationException(
+          String.format(
+              "Invalid table name '%s': vectorSearch() requires a single concrete index or alias;"
+                  + " wildcards ('*') and multi-target patterns (comma-separated) are not"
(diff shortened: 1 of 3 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/storage/VectorSearchTableFunctionImplementation.java
```diff
@@ -203,6 +203,16 @@ private void validateNamedArgs() {
    * its own error message.
    */
```
RyanL1997, line 206: non-blocking nit: Javadoc on `validateTableName` is now slightly stale — The existing javadoc only describes "alphanumeric, dots, underscores, hyphens" and mentions `_all`/`.`/`..` rejection, but doesn't mention the new wildcard/multi-target rejection.

RyanL1997 (approved):

PR #5386, opened 2026-04-24
Add integration tests for clusters without k-NN plugin

mengweieric:
## Summary
Adds integration tests that lock in vectorSearch() behavior on clusters without the opensearch-knn plugin installed: execution fails with a clear error at scan open() time, while `_explain` still returns a plan.

## Test plan
- [x] `:integ-test:integTest -Dtests.class=...VectorSearchIT -Dtests.method=testExecutionWithoutKnnPluginFailsWithClearMessage` passes
- [x] `:integ-test:integTest -Dtests.class=...VectorSearchIT -Dtests.method=testExplainWithoutKnnPluginStillSucceeds` passes
- [x] `spotlessCheck` clean


mengweieric:
Superseded by #5388, which strengthens the two existing no-kNN tests in place rather than adding near-duplicates, and is bundled with the multi-backing-index alias coverage into a single integration-test PR.

PR #5387, opened 2026-04-24
Add integration test for alias with multiple backing indices

mengweieric:
## Summary
Adds an integration test that locks in vectorSearch() planning over an alias pointing at multiple backing indices. Common multi-tenant setup where a shared alias fans out to per-tenant indices. Execution requires the k-NN plugin, so the test covers the planning path via `_explain`.

## Test plan
- [x] `:integ-test:integTest -Dtests.class=...VectorSearchIT -Dtests.method=testExplainOverAliasWithMultipleBackingIndices` passes
- [x] `spotlessCheck` clean


mengweieric:
Superseded by #5388, which bundles this alias coverage with strengthened no-kNN tests into a single integration-test PR.

Issue #5357, opened 2026-04-15
[FEATURE] Let `sql-worker` thread pool be resizeable instead of fixed


toepkerd:
**Is your feature request related to a problem?**
When running DSL vs PPL load tests on a managed OpenSearch 3.5 domain, I found that DSL's bottleneck was the entire cluster, while PPL's bottleneck was the `sql-worker` thread pool.

DSL queries execute on the resizeable `search` thread pool. When running expensive DSL queries on increasingly high volume data, it dynamically scales up to use more of the cluster's resources in response to the increased load. DSL performance only degraded when the cluster itself was strained from the load.

PPL queries execute on the fixed `sql-worker` thread pool. When running expensive PPL queries on increasingly high volume data, this thread pool eventually saturated. As a result, PPL tasks waited long times in thread pool queues, and PPL specifically degraded while the rest of the cluster moved on without issue.

**What solution would you like?**
Change the construction here to use a dynamic executor builder instead of a fixed executor builder: https://github.com/opensearch-project/sql/blob/ee28e3d106d30cd1e70fed9a35444da40683b7ec/plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java#L354-L360

**What alternatives have you considered?**
Deliberation may be needed on if this is a desirable change, and consideration of potential consequences for PPL work competing for cluster resources with other tasks.

**Do you have any additional context?**
N/A

jhansimandava27-lang:
Hi! I'd like to work on this issue. I've already implemented the change locally, will push the PR once the testing is done @RyanL1997

jhansimandava27-lang:
@RyanL1997 @toepkerd would like to work on this - can some one help in assigning?

jhansimandava27-lang:
@penghuo  can you please assign this?

PR #5389, opened 2026-04-27
Upgrade SQL worker thread pool to ScalingExecutorBuilder for dynamic thread scaling  (#5357) 

jhansimandava27-lang:
Description

Upgrades the SQL worker thread pool from FixedExecutorBuilder to ScalingExecutorBuilder to enable dynamic thread scaling based on system load.

Related Issues

Resolves #[(#5357)]

### Check List
- [ Yes] New functionality includes testing.

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
@@ -35,6 +35,7 @@ import org.opensearch.common.settings.IndexScopedSettings;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.settings.SettingsFilter;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.util.concurrent.OpenSearchExecutors;
 import org.opensearch.core.action.ActionResponse;
 import org.opensearch.core.common.io.stream.NamedWriteableRegistry;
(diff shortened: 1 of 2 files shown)


jhansimandava27-lang:
Integration test is failing and that's not related to the changes of this PR

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -340,16 +342,17 @@ public ScheduledJobParser getJobParser() {
...
+    // The background thread pool is a separate queue for asynchronous requests to other nodes.
+    // We keep them separate to prevent deadlocks during async fetches on small node counts.
+    // Tasks in the background pool should do no work except I/O to other services.
     return List.of(
-        new FixedExecutorBuilder(
-            settings,
+        new ScalingExecutorBuilder(
             SQL_WORKER_THREAD_POOL_NAME,
+            1,
             OpenSearchExecutors.allocatedProcessors(settings),
```
Swiddis, line 354: just checking: is this the same limit as what core has on scaling search threads?

If so, I'm not sure I see the benefit of making this scaling, as the scaling pool is bounded to the same value we had before anyway. But I'll benchmark it today and if it mostly functions the same, I'm fine to accept the change and maybe visit introducing the ability to scale beyond the processor limit another day

Swiddis: - https://github.com/opensearch-project/OpenSearch/blob/8804f3c29105971c20103d68b20352252028b858/server/src/main/java/org/opensearch/threadpool/ThreadPool.java#L269-L272
- https://github.com/opensearch-project/OpenSearch/blob/8804f3c29105971c20103d68b20352252028b858/server/src/main/java/org/opensearch/threadpool/ThreadPool.java#L765-L767

Neat, they use a resizing thread pool instead of a scaling one -- maybe worth following their pattern? Not going to block on that though

Issue #3751, opened 2025-06-10
[FEATURE] Support creating typed arrays & structs with Calcite


yuancu:
**Is your feature request related to a problem?**

Currently all arrays are created with type `array<any>`.

Reason: in [OpenSearchTypeFactory.java#L176](https://github.com/opensearch-project/sql/blob/ed507d728db525b558540671bfeed06b53e5c73f/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java#L176):
```
public static RelDataType convertExprTypeToRelDataType(ExprType fieldType, boolean nullable) 
...
        case ARRAY:
          return TYPE_FACTORY.createArrayType(
              TYPE_FACTORY.createSqlType(SqlTypeName.ANY, nullable), -1);
```

This raises a number of problem since we can not infer element types. E.g. when expanding an array, we can't know the element type until runtime.

**What solution would you like?**
Support creating array types with known element types.

**What alternatives have you considered?**
Not yet

**Do you have any additional context?**
- https://github.com/opensearch-project/sql/pull/3745#discussion_r2136092531
- https://github.com/opensearch-project/sql/issues/3229

PR #5390, opened 2026-04-28
[BugFix] Fix PPL nested field access inside disabled object fields (#4906)

### LINKED ISSUE #4906: [BUG] PPL query return null when access valid inner json object
```
###
PUT {{baseUrl}}/log00001

{
  "mappings": {
    "properties": {
      "log": {
        "type": "object",
        "enabled": false
      }
    }
  }
}

###

{"index": {}}
{"log": {"a":1, "c": {"d": 2}}}

###

{
  "query": "source=log00001 | fields log.c.d"
}
```
* Results
```
{
  "schema": [
    {
      "name": "log.c.d",
      "type": "undefined"
    }
  ],
  "datarows": [
    [
      null
    ]
  ],
  "total": 1,
  "size": 1
}
```

It should return 

```
{
  "schema": [
    {
      "name": "log.c.d",
      "type": "int"
    }
  ],
  "datarows": [
    [
      2
    ]
  ],
  "total": 1,
  "size": 1
}
```

qianheng-aws:
### Description
When an index mapping declares an object with `"enabled": false` (or any object without indexed properties), inner fields have no mapping entry. The data layer previously stringified the unmapped nested value via `parseContent`'s default case, so querying `fields log.c.d` returned `null` with type `"undefined"`.

**Root cause:** In `OpenSearchExprValueFactory.parse`, when a field has no type (empty `Optional<ExprType>`) and the value is a nested object, the code fell through to `parseContent` which does not handle objects — they get stringified as `ObjectNode.toString()`, losing the nested structure entirely.

**Fix:**
- Add `Content.isObject()` default method to detect map/object-like content for both JsonNode and plain Java Map sources.
- `OpenSearchExprValueFactory.parse` now recurses into unmapped objects via a new `parseUnmappedStruct` helper instead of stringifying them.
- `parseStruct` additionally exposes every descendant of an unmapped child under a dotted-path key at the current level, so the Calcite `ITEM` operator (which treats a tuple as a flat map) can resolve `ITEM(log, "c.d")` with a single lookup — this is the same shape produced by `json_extract_all`, keeping `spath`-based paths consistent.

### Related Issues
Resolves #4906

### Check List
- [x] `spotlessCheck` passed
- [x] Unit tests passed (`OpenSearchExprValueFactoryTest`)
- [x] Integration tests passed (`CalciteDisabledObjectFieldIT`, `CalcitePPLMapPathIT`, `CalciteObjectFieldOperateIT`)
- [x] YAML REST test added (`issues/4906.yml`)


qianheng-aws:
## Decision Log

**Root Cause**: In `OpenSearchExprValueFactory.parse`, when a field has no type defined in the mapping (which is the case for every descendant of an object declared with `"enabled": false`), the code fell through to `parseContent`. `parseContent` only handles scalars — if the content is a JSON object it hits the default branch and stringifies it via `objectValue().toString()`. The nested structure (`{"c": {"d": 2}}`) became a string, so any subsequent `ITEM(log, "c.d")` lookup returned `null` with type `undefined`. The bug was strictly depth-dependent: single-level paths (`log.a`) still worked because the value was a scalar; only paths of depth ≥ 2 failed.

**Approach**:
1. Add a `Content.isObject()` default method (handles both `JsonNode` and `Map`). Clean and opt-in — existing subclasses inherit the correct behavior.
2. In `parse`, when type is empty/UNDEFINED **and** content is an object, recurse via a new `parseUnmappedStruct` helper instead of stringifying.
3. `parseUnmappedStruct` builds a tuple whose keys include both the direct child names and every descendant under a dotted-path key (e.g. `{"a": 1, "c": tuple{d:2, ...}, "c.d": 2}`). This matches the shape produced by `json_extract_all` (the existing `spath` code path), so `ITEM(tuple, "c.d")` — which performs a single `map.get(key)` — resolves the leaf with one lookup.
4. `parseStruct` (used for mapped objects) only injects dotted-path entries for **unmapped** children. Mapped struct children keep their original nested shape, preserving behavior for normal mapped objects (e.g. `city.location.latitude`).

**Alternatives Rejected**:
- *Chain ITEM calls in `QualifiedNameResolver.resolveFieldAccess`* — would break the `spath` / `json_extract_all` path which already produces flat-keyed maps; `ITEM(ITEM(doc, "user"), "name")` returns null because `doc` has no key "user", only "user.name".
- *Store only the flat dotted-path form* — would break `fields log.c` (intermediate struct access) and any downstream command that iterates the tuple as a struct.
- *Fix at the query layer (`resolveFieldAccess`)* — requires runtime type introspection to pick flat vs nested access, fragile and harder to reason about.
- *Modify `json_extract_all` to produce nested maps* — larger refactor and would reopen already-fixed issues (e.g. #5185) that rely on the flat-key shape.

**Pitfalls**:
- First fix attempt only updated `parse` (added `parseUnmappedStruct`), but this only flattened keys at the unmapped level. The parent `parseStruct` (for the mapped `log` object) still produced a nested tuple without a top-level `c.d` key, so `ITEM(log, "c.d")` still returned null. Required adding `exposeDescendantsAsDottedKeys` in `parseStruct` so flat keys bubble up one level — to the tuple that Calcite's ITEM operator actually sees.
- The leaf schema type is still `undefined` for disabled-object paths because the mapping truly has no type information. Only the **value** can be recovered. The integration test asserts `verifyDataRows(..., rows(2))` (not `verifySchema`), matching this limitation.
- Malformed field names (`""`, `".a"`, etc.) — already handled by `isFieldNameMalformed`, the new helper inherits that guard.

**Things to Watch**:
- The fix applies whenever a parent field lacks a type in the mapping. Same logic handles disabled objects, dynamic-mapping-disabled objects, and sub-fields of generated objects (the TODO for #3751 remains valid).
- If future code relies on a tuple value having **only** its direct children as keys (i.e. no dotted-path flat keys), the injection in `parseStruct` could surprise it. I scanned callers of `ExprTupleValue.tupleValue()` and they all use `get(exactKey)` or iterate entries for serialization — both are safe with extra flat keys.
- The schema type for disabled-object descendants remains `undefined` because we cannot infer types without a mapping. A follow-up enhancement could infer types from the first observed value, but that's out of scope here.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/utils/Content.java
```diff
@@ -53,6 +53,18 @@ public interface Content {
   /** Is array value. */
   boolean isArray();
 
+  /** Is an object / map-like value. */
+  default boolean isObject() {
```
qianheng-aws: The implementation should be split into the extended calss.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -383,15 +445,40 @@ private ExprValue parseStruct(Content content, String prefix, boolean supportArr
...
-                    new JsonPath(fieldKey),
-                    parse(entry.getValue(), fullFieldPath, type(fullFieldPath), supportArrays));
+                Optional<ExprType> childType = type(fullFieldPath);
+                ExprValue childValue =
+                    parse(entry.getValue(), fullFieldPath, childType, supportArrays);
+                populateValueRecursive(result, new JsonPath(fieldKey), childValue);
+                // If the child's type is unmapped (e.g. inside a disabled object — see #4906),
+                // also expose every descendant under its dotted-path key so single-key ITEM
+                // lookups at this level resolve nested paths like log.c.d.
+                if (isUnmappedType(childType) && childValue instanceof ExprTupleValue) {
```
qianheng-aws, line 447: Could this be moved into parse method?

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
```diff
@@ -203,6 +203,12 @@ private ExprValue parse(
     //  is resolved
     if (fieldType.isEmpty()
         || fieldType.get().equals(OpenSearchDataType.of(ExprCoreType.UNDEFINED))) {
+      // If the content is an object (e.g. inside a disabled object field, see issue #4906),
+      // recurse into it as a struct so downstream path access (log.c.d) works. Without this,
+      // parseContent would fall through to stringify the object and nested keys would be lost.
+      if (content.isObject()) {
```
qianheng-aws: Could this be moved into parseContent method?

qianheng-aws:
Addressed bot review feedback in the follow-up commit:

**Applied:**
- *Leaky abstraction in `Content.isObject()`* — made the method abstract and implemented it directly in `OpenSearchJsonContent` (via `JsonNode.isObject()`) and `ObjectContent` (via `instanceof Map`). The Jackson reference no longer leaks into the interface.
- *Redundant `isArray()` / `else` branches in `parseUnmappedStruct`* — collapsed into a single `else` with a comment clarifying scalars and arrays both go through `parseContent`.

**Not applied (with reasoning):**
- *Skip intermediate tuple nodes when exposing dotted keys* — this is intentional per the Decision Log. Intermediate paths (e.g. `log.c` mapping to a tuple) must remain addressable so `fields log.c` or any downstream struct access still works; only dotted leaf lookups are the new addition. Suppressing tuple descendants would regress intermediate-struct access.
- *Null guard in `isObject()`* — moot after making the method abstract. Each implementation handles its own null semantics: `OpenSearchJsonContent` explicitly null-checks `value`, `ObjectContent` relies on `instanceof Map` which is null-safe.
- *Flat key explosion / mutable state* — verified: the flat-key shape mirrors what `json_extract_all` already produces (consistent with the `spath` code path), and all callers of `ExprTupleValue.tupleValue()` use exact-key lookups or serialization iteration. Both are safe with extra dotted-path entries. The existing eval/map code already mutates the returned map in place, so no abstraction is broken.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
qianheng-aws: Thanks, done in dd1e295 — `parseContent` now owns the `content.isObject()` branch and dispatches to `parseUnmappedStruct` itself. `parse()`'s unmapped-type branch is now a single `return parseContent(content, field, supportArrays);`, so all content-shape dispatch lives in one method. `parseUnmappedStruct`'s child recursion also goes through `parseContent`, collapsing the previous null/object/scalar if-else into one call.

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
qianheng-aws: Thanks — partially done in dd1e295. The per-child body of `parseStruct` (the `populateValueRecursive` + conditional `exposeDescendantsAsDottedKeys`) is now extracted into a private helper `parseStructEntry` that sits next to `parse()`. `parseStruct` itself is reduced to:

```java
content.map().forEachRemaining(entry -> parseStructEntry(result, prefix, entry, supportArrays));
```

Moving the dotted-key exposure *literally inside* `parse()` would require `parse()` to accept the parent `ExprTupleValue result` + key-in-parent so it can write flat entries directly. That widens `parse()`'s signature across all call sites (`parseArray`, `parseInnerArrayValue`, the two `construct()` overloads, and the recursive call from `parseStructEntry`) with arguments that are meaningless for non-struct-child contexts. It also breaks `parse()`'s current contract of "return an `ExprValue`, no side effects on caller state".

An alternative — having `parse()` mark its return with a sentinel so `parseStruct` can uniformly flatten — also fails because mapped-struct children *can* contain unmapped subtrees and do need selective flattening based on the declared field type, not just on content shape.

The current shape (`parseStruct` → trivial delegation to `parseStructEntry` which encapsulates parse + populate + conditional expose) keeps `parse()`'s signature clean while still hiding the "awkward conditional" behind a single helper. Happy to go deeper (e.g. widen `parse()` to take the sink) if you prefer that trade-off — let me know.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDisabledObjectFieldIT.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  @Test
+  public void testSelectNestedFieldFromDisabledObject() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source=%s | fields log.c.d", DISABLED_OBJECT_INDEX));
+    // Schema is undefined because disabled object fields have no field mapping.
+    // The important check is that the value is extracted correctly from _source.
+    verifyDataRows(result, rows(2));
```
qianheng-aws, line 57: Should also verifySchema

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDisabledObjectFieldIT.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Test
+  public void testSelectTopLevelFieldFromDisabledObject() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source=%s | fields log.a", DISABLED_OBJECT_INDEX));
+    // Schema is undefined because disabled object fields have no field mapping.
+    // The important check is that the value is extracted correctly from _source.
+    verifyDataRows(result, rows(1));
```
qianheng-aws, line 65: ditto

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDisabledObjectFieldIT.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Test
+  public void testSelectEntireDisabledObject() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source=%s | fields log", DISABLED_OBJECT_INDEX));
+    verifySchema(result, schema("log", "struct"));
```
qianheng-aws, line 72: Should also verifyDataRows

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDisabledObjectFieldIT.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Test
+  public void testSelectIntermediateFieldFromDisabledObject() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source=%s | fields log.c", DISABLED_OBJECT_INDEX));
+    verifySchema(result, schema("log.c", "struct"));
```
qianheng-aws, line 80: ditto

opensearch/src/main/java/org/opensearch/sql/opensearch/data/value/OpenSearchExprValueFactory.java
qianheng-aws: Update: in ac4afb0 I reverted the `parseStructEntry` extraction and inlined the per-child loop body back into `parseStruct`. On reflection, the extraction didn't really address the intent of your comment — moving the dotted-key exposure into `parse()` itself — it just moved it one call deeper.

A literal move into `parse()` would require either (a) widening its signature with a parent-tuple sink + key-in-parent across 8+ call sites, breaking its current pure-return contract and carrying meaningless parameters through non-struct-child paths, or (b) returning a richer shape that `parseStruct` then picks apart — either trades simplicity for a false sense of locality.

So the cleanest place for the mutation is `parseStruct` itself, which is the only caller with the parent tuple in scope. Keeping the `parseContent` unification from your line-209 comment and leaving `parseStruct` as the single owner of the dotted-key exposure. Thanks again for the review.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDisabledObjectFieldIT.java
qianheng-aws: Added `verifySchema(result, schema("log.c.d", "int"))` in commit 8aefdab. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDisabledObjectFieldIT.java
qianheng-aws: Added `verifySchema(result, schema("log.a", "int"))` in commit 8aefdab. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDisabledObjectFieldIT.java
qianheng-aws: Added `verifyDataRows(result, rows(new JSONObject("{\"a\":1,\"c\":{\"d\":2},\"c.d\":2}")))` in commit 8aefdab. Thanks!

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDisabledObjectFieldIT.java
qianheng-aws: Added `verifyDataRows(result, rows(new JSONObject("{\"d\":2}")))` in commit 8aefdab. Thanks!

LantaoJin (dismissed):

penghuo (approved):

PR #5392, opened 2026-04-29
Preserve original expression text as column name in unified SQL

### LINKED ISSUE #5332: UnifiedQueryPlanner should preserve original SQL expression names instead of EXPR$N
## Problem

When executing SQL queries with unnamed aggregate expressions through `UnifiedQueryPlanner`, Calcite's default column naming produces `EXPR$0`, `EXPR$1`, etc. instead of the original SQL expression text.

**Example:**
```sql
SELECT COUNT(*) FROM hits
```
Returns column name `EXPR$0` instead of `count(*)`.

```sql
SELECT SUM(advengineid), COUNT(*), AVG(resolutionwidth) FROM hits
```
Returns `EXPR$0`, `EXPR$1`, `EXPR$2` instead of `sum(advengineid)`, `count(*)`, `avg(resolutionwidth)`.

**Note:** Explicitly aliased expressions work correctly — `SELECT COUNT(*) AS cnt` returns `cnt`.

## Expected Behavior

Match PostgreSQL/MySQL convention — use the original SQL expression text as the column name when no alias is provided:

| Engine | `SELECT COUNT(*)` column name |
|--------|-------------------------------|
| PostgreSQL | `count` |
| MySQL | `COUNT(*)` |
| DuckDB | `count_star()` |
| Spark SQL | `count(1)` |
| Trino/Presto | `_col0` |
| **Current (Calcite)** | **`EXPR$0`** |

## Root Cause

In `UnifiedQueryPlanner.plan()`, the flow is:
1. Parse SQL → `SqlNode` (has original expression text like `COUNT(*)`)
2. Validate → `SqlNode` (still has expression text)
3. Convert to `RelNode` → **Calcite replaces unnamed expressions with `EXPR$N`**

The original expression text is available in the `SqlNode` (specifically `SqlSelect.getSelectList()`) but is lost during `SqlToRelConverter`.

## Proposed Fix

In `UnifiedQueryPlanner`, after parsing but before/after converting to `RelNode`:

1. Extract the select-list expression names from the `SqlNode` (e.g., `SqlCall.toString()` gives `COUNT(*)`)
2. After `SqlToRelConverter` produces the `RelNode`, rename any `EXPR$N` fields back to the original expression text (lowercased to match PostgreSQL convention)

This could be done by wrapping the resulting `RelNode` in a `LogicalProject` that renames the fields, or by overriding the `RelDataType` field names.

## Context

This affects the lakehouse SQL endpoint (`_lakehouse/sql`) in the analytics-engine plugin. The `UnifiedQueryService` calls `UnifiedQueryPlanner.plan()` and uses `RelNode.getRowType().getFieldNames()` to get column names for the response. Since `UnifiedQueryPlanner` is in the `org.opensearch.sql.api` package (compiled dependency in dsl-query-executor), downstream consumers cannot fix this without modifying the planner.

## Workaround

Downstream consumers can post-process column names (e.g., replace `EXPR$N` with `_colN`), but this loses the original expression text.

dai-chen:
### Description

This PR preserves the original expression text as column names by adding a SqlNode rewriter to the existing `LanguageSpec.postParseRules` hook, instead of Calcite's default synthetic `EXPR$0`, `EXPR$1`, etc.

Examples:

```
SELECT COUNT() FROM t       -> COUNT()   (was EXPR$0)
SELECT SUM(MyCol)            -> SUM(MyCol)
SELECT UPPER(name), age + 1  -> UPPER(name), age + 1
SELECT name, *, x AS y       -> unchanged
```

Implementation notes:

1. **Column-naming conventions vary across engines**: See [comparison table](https://github.com/opensearch-project/sql/issues/5332#issuecomment-4355686910). This PR produces canonical built-in function names with bare user identifiers (`COUNT(*)`, `SUM(MyCol)`, `UPPER(name)`).

2. **Verbatim preservation (V2/PPL V3/MySQL) is debatable**: SQL treats unquoted function names as case-insensitive (`COUNT` == `count`), so canonical labels are spec-consistent. Verbatim text also couples column names to source formatting — `SUM(x)+1` and `SUM(x) + 1` would produce different labels for the same query. We'll revisit verbatim preservation in https://github.com/opensearch-project/sql/issues/5346.

### Related Issues

Resolves #5332

### Check List

### DIFF
--- a/api/src/main/java/org/opensearch/sql/api/spec/search/SearchExtension.java
+++ b/api/src/main/java/org/opensearch/sql/api/spec/search/SearchExtension.java
@@ -22,6 +22,8 @@ public class SearchExtension implements LanguageSpec.LanguageExtension {
 
   @Override
   public List<SqlVisitor<SqlNode>> postParseRules() {
-    return List.of(NamedArgRewriter.INSTANCE);
+    return List.of(
+        SelectItemAliasRewriter.INSTANCE, // TODO: non-search-specific and temporarily placed here
+        NamedArgRewriter.INSTANCE);
   }
 }
--- a/api/src/main/java/org/opensearch/sql/api/spec/search/SelectItemAliasRewriter.java
+++ b/api/src/main/java/org/opensearch/sql/api/spec/search/SelectItemAliasRewriter.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.api.spec.search;
+
+import java.util.function.UnaryOperator;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlSelect;
+import org.apache.calcite.sql.SqlWriterConfig;
+import org.apache.calcite.sql.dialect.AnsiSqlDialect;
+import org.apache.calcite.sql.util.SqlShuttle;
+import org.apache.calcite.sql.validate.SqlValidatorUtil;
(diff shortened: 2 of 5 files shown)


dai-chen:
Closing as we decided to use SQL V2 parser for now.

Issue #203, opened 2021-09-04
Support for converting sql to Elasticsearch query on the client side itself


newUserForTesting:
**Is your feature request related to a problem? Please describe.**
Our hundreds of Production Clusters are running on OpenDistro version of Elasticsearch. We recently observed the new OpenSearch initiative. We can't wait to upgrade to OpenSearch version!
But since we have lots of existing Production Clusters and enormous amount of data within it, it wouldn't be happening in immediate near future. This means we can't make use of all the fixes(some example includes fuzzy support, querying object in nested structures) which are coming in OpenSearch but not in OpenDistro.

**Describe the solution you'd like**
A possible solution for the above issue could be that on Application Side (client side), we use https://github.com/opensearch-project/sql which has latest features (and upcoming new features as well), convert our sql query to Elasticsearch query **on the client side itself** (without hitting Elasticsearch since its still on OpenDistro version), and then use the Elasticsearch query to send request to Elasticsearch. Could someone please guide which methods of this project we can invoke to achieve this? We really appreciate your help!

chloe-zh:
Hi @newUserForTesting thanks for reaching out to us! Glad to hear that you and your team like our software and would like to migrate your product to OpenSearch. 

For the solution you proposed, I'm not sure if this is workable to convert SQL query to Elasticsearch DSL from OpenSearch of the client side, since the DSL generated in the client side would be OpenSearch query DSL but might not be compatible with ES DSL at least at the query API level. 

And I guess you might want the direct DSL script that is obtained from our explain endpoint, but actually this script is not exactly the same with the executable query DSL that used in search, but we have also included the in memory execution plan in the explanations. 

So for now probably the best way to do it is to migrate the entire cluster to OpenSearch if possible. Will discuss it with the team to see what we can do to help and keep you updated. Thanks!

chloe-zh:
Here is a procedure to upgrade ODFE cluster to OpenSearch: https://opensearch.org/docs/upgrade-to/upgrade-to/ Hope it helps you well!

Yury-Fridlyand:
@newUserForTesting,
You can use [explain](https://github.com/opensearch-project/sql/blob/main/docs/user/interfaces/endpoint.rst#explain) request to get DSL query executed on the server without actual execution. Keep in mind, that SQL plugin does post-processing of the results, which is not reflected in the DSL query.

PR #5398, opened 2026-05-01
Draft AnalyticsFrontEndExtension SPI for analytics-engine integration

ahkcs:
## TL;DR

Drafts an SPI interface so the SQL plugin can integrate with `analytics-engine` without taking a hard install-time dependency on it. **This PR alone does NOT fix the install failure** (`Missing plugin [analytics-engine]`); it's the first of three coordinated changes. See the rollout sequence below.

## Background

Today, `plugin/build.gradle` declares `extendedPlugins = ['..., analytics-engine']` as a HARD dependency, so `opensearch-sql` fails to install on stock OpenSearch distros that don't ship `analytics-engine` (observed in `feature-build-opensearch` #203). Marking it `;optional=true` is necessary but not sufficient: `TransportPPLQueryAction` Guice-injects `QueryPlanExecutor`, and Guice cannot satisfy that binding when the providing plugin is absent — SQL still fails to construct.

The fix is to invert the dependency via an SPI, the same pattern `opensearch-job-scheduler` uses. SQL declares its capability via the SPI; analytics-engine discovers consumers and pushes its services through the hook. SQL boots fine standalone, and analytics paths stay inert when analytics-engine is absent.

## What's in this PR

A single new file: `plugin/src/main/java/org/opensearch/analytics/spi/AnalyticsFrontEndExtension.java`. The interface has two setters:

- `setQueryPlanExecutor(QueryPlanExecutor<RelNode, Iterable<Object[]>>)`
- `setSchemaProvider(SchemaProvider)`

JavaDoc spells out the lifecycle (discovery via `ExtensiblePlugin#loadExtensions`, push after Guice builds the node injector, exactly once per consumer per node).

The file is placed under its eventual FQN (`org.opensearch.analytics.spi`) so `SQLPlugin`'s future `implements` clause does not change once the interface lands in the `analytics-framework` JAR.

## What's NOT in this PR

- No `;optional=true` flip
- No `SQLPlugin implements AnalyticsFrontEndExtension`
- No changes to `TransportPPLQueryAction` / `AnalyticsExecutorHolder`
- No vendoring of a new `analytics-framework` JAR

These all land in the follow-up SQL PR (step 3 below).

## Rollout sequence (what unblocks the install)

| Step | Repo | Owner | What it does | Effect on install |
|------|------|-------|--------------|-------------------|
| 1 | this PR | this | drafts the SPI contract for review | none |
| 2 | `opensearch-project/OpenSearch` ([#21449](https://github.com/opensearch-project/OpenSearch/pull/21449)) + analytics-engine team | analytics-engine team | adds the SPI to `analytics-framework` + Guice listener in `AnalyticsPlugin` to discover consumers and push `DefaultPlanExecutor` + a `SchemaProvider` | none |
| 3 | this repo | this | vendors the rebuilt `analytics-framework` JAR, deletes this placeholder file, adds `SQLPlugin implements AnalyticsFrontEndExtension`, flips `extendedPlugins;optional=true` | **fixes the install failure** |

The temporary install-script workaround that force-installs `analytics-engine` before `opensearch-sql` is still required until step 3 merges.

## Test plan

- [x] `./gradlew :opensearch-sql-plugin:spotlessCheck`
- [x] `./gradlew :opensearch-sql-plugin:compileJava :opensearch-sql-plugin:compileTestJava`
- [x] Existing `RestUnifiedQueryActionTest` — passes (no behavioral change)

The interface has no implementations or callers in this PR, so there are no new behavioral tests.


ahkcs:
Closing in favor of the core PR opensearch-project/OpenSearch#21449 which establishes the contract directly in analytics-framework where it belongs. The SQL-side follow-up will pick up the SPI from the rebuilt analytics-framework JAR once that PR merges — no need for a placeholder draft in this repo.

PR #5403, opened 2026-05-04
Make analytics-engine an optional dependency 

bowenlan-amzn:
## Summary

Makes `analytics-engine` an optional dependency of `opensearch-sql`. Fixes the install failure on distributions without `analytics-engine`:

```
Missing plugin [analytics-engine], dependency of [opensearch-sql]
```

Without `analytics-engine`, PPL/SQL queries route through the in-plugin v2 Calcite engine (Lucene). With it, the existing Mustang path (parquet indices → `QueryPlanExecutor`) still works.

## Changes

1. `extendedPlugins = ['opensearch-job-scheduler', 'analytics-engine;optional=true']`
2. `QueryPlanExecutor` moved from a required constructor parameter to an `@Inject(optional = true)` setter (Guice rejects optional on constructors). Guice skips the setter when the binding is absent; `unifiedQueryHandler` stays null; PPL routing falls through to the Calcite→OpenSearch path.
3. `analytics-framework.jar`: `compileOnly` → `api`. Needed so sql's classpath resolves `QueryPlanExecutor`/`SchemaProvider` standalone. `analytics-engine.jar` stays `compileOnly` — its classes (e.g. `OpenSearchSchemaBuilder`) are only reached through `RestUnifiedQueryAction`, which doesn't load when the setter is skipped.
4. Dropped the `bundlePlugin { exclude '*.jar' }` block so sql carries guava/calcite/jackson/etc. for standalone install.

## Validation

Live 2-node cluster, both configurations:

**With `analytics-engine` installed:** legacy PPL, analytics PPL (20 rows, parquet-backed), and SQL all work ✓

**Without `analytics-engine`** (only `opensearch-job-scheduler` + `opensearch-sql`): install succeeds, PPL/SQL against Lucene indices work, `parquet_*` lookup returns a clean `IndexNotFoundException` (not NPE or `NoClassDefFoundError`) ✓

## Not in this PR

- The routing heuristic (table prefix `parquet_`) is unchanged; a setting- or data-format-based trigger is a separate discussion.
- No changes in `analytics-engine` or core OpenSearch.
- No CI guard for transitive version consistency between sql and analytics-engine — worth adding as a follow-up.

---

<details>
<summary><b>Why bundling jars is safe</b> (addresses the diff analyzer findings)</summary>

The duplicated transitives (guava, calcite, jackson, etc.) are tolerated because OpenSearch explicitly supports this shape for optional deps:

- `;optional=true` is standard syntax (`PluginInfo.java:229`).
- Jar-hell cross-check between sql and analytics-engine is skipped for optional deps (`PluginsService.java:763`).
- At runtime, parent-first classloader delegation routes every shared-class lookup to analytics-engine's copy when it's installed; sql's bundled copies sit idle.

**Assumption:** sql and analytics-engine track the same transitive versions. Today both inherit from the OpenSearch root `versions.*` catalog, so they match by construction. If they drift, parent-first would silently pick analytics-engine's version — a CI guard comparing transitives would catch it.

**Alternatives considered:** a conditional exclude by deployment shape isn't viable (Gradle doesn't know at build time whether the target cluster has analytics-engine). Shipping two zips isn't supported by the OpenSearch plugin distribution model.

</details>

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java
@@ -62,9 +62,12 @@ public class TransportPPLQueryAction
 
   private final Supplier<Boolean> pplEnabled;
 
-  private final RestUnifiedQueryAction unifiedQueryHandler;
+  /** Null when analytics-engine plugin is absent; set via {@link #setQueryPlanExecutor}. */
+  private volatile RestUnifiedQueryAction unifiedQueryHandler;
 
-  /** Constructor of TransportPPLQueryAction. */
+  private final NodeClient clientRef;
+  private final ClusterService clusterServiceRef;
+
   @Inject
   public TransportPPLQueryAction(
       TransportService transportService,
@@ -72,9 +75,10 @@ public class TransportPPLQueryAction
       NodeClient client,
       ClusterService clusterService,
       DataSourceServiceImpl dataSourceService,
-      org.opensearch.common.settings.Settings clusterSettings,
-      QueryPlanExecutor<RelNode, Iterable<Object[]>> queryPlanExecutor) {
+      org.opensearch.common.settings.Settings clusterSettings) {
     super(PPLQueryAction.NAME, transportService, actionFilters, TransportPPLQueryRequest::new);
+    this.clientRef = client;
+    this.clusterServiceRef = clusterService;
 
     ModulesBuilder modules = new ModulesBuilder();
     modules.add(new OpenSearchPluginModule());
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java
```diff
@@ -98,6 +108,21 @@ public TransportPPLQueryAction(
...
   }
 
+  /**
+   * Optional method injection target. Guice invokes this ONLY when a binding for {@code
+   * QueryPlanExecutor<RelNode, Iterable<Object[]>>} exists — i.e. when analytics-engine plugin
+   * is installed and its {@code createGuiceModules} has bound {@code DefaultPlanExecutor}.
+   * Absent analytics-engine, Guice silently skips the call; {@link #unifiedQueryHandler}
+   * stays null and all PPL routing falls through to the legacy engine.
+   */
+  @Inject(optional = true)
```
mch2, line 103: does this work?

ahkcs (approved):

PR #5412, opened 2026-05-06
Add include_metadata request parameter for PPL queries #5235

### LINKED ISSUE #5235: [FEATURE] Add include_metadata request parameter for PPL queries

Yes. When using PPL queries, there's no way to include metadata fields like `_id` and `_index` in the results without explicitly listing them. This creates problems for common workflows like searching around documents.

**Use Case: Search Around Documents**

A typical workflow is:
1. Query to find N interesting documents (e.g., errors, anomalies, specific patterns)
2. Use the returned `_id` values to search for surrounding context:
   - Log entries before/after an error event
   - Related events in the same time window
   - Documents from the same session/user
   - Time-series data around an anomaly

**Current problem:** The wildcard `*` excludes metadata fields by default, so `source=logs | fields *` returns all regular fields but no `_id`. Users must either list all fields explicitly (tedious and breaks with dynamic schemas) or run separate queries (double the latency).

Add a **request-level parameter** `include_metadata` to the PPL query API:

```bash
POST /_plugins/_ppl?include_metadata=true
{
  "query": "source=logs | where level='ERROR' | fields * | head 10"
}
```

**Result:** All regular fields PLUS metadata fields (`_id`, `_index`, `_score`, etc.)

**Benefits:**
- ✅ Get document IDs in a single query - no double roundtrip
- ✅ No query text changes needed - just a request parameter
- ✅ Easy to use in SDKs and client applications
- ✅ Works with any field selection (wildcards, patterns, explicit fields)
- ✅ Non-breaking - default behavior unchanged (metadata excluded)
- ✅ Enables search-around-documents and correlation workflows

1. **Explicit listing** - `source=logs | fields *, _id, _index`
   - Requires knowing metadata field names
   - Only works with Calcite enabled
   - Undocumented

2. **Changing default behavior** - Make `fields *` include metadata
   - Breaking change for existing queries

This feature would make PPL more practical for observability and log analysis workflows where document correlation is essential. The request parameter approach is simple, discoverable, and aligns with how similar features work in other query APIs.

**Metadata fields to include when enabled:**
- `_id` - Document ID
- `_index` - Index name
- `_score` - Search relevance score
- `_routing` - Routing value
- `_sort` - Sort values

ishag4:
### Description
Add a request-level parameter include_metadata to the PPL query API:

POST /_plugins/_ppl?include_metadata=true
{
"query": "source=logs | where level='ERROR' | fields * | head 10"
}
Result: All regular fields PLUS metadata fields (_id, _index, _score, etc.)

### Related Issues
Resolves #5235 

### Check List


ishag4:
Hi @LantaoJin @penghuo @RyanL1997 @Swiddis Could you please review?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLFlattenTest.java
```diff
@@ -64,16 +64,16 @@ public void testFlatten() {
...
-        "SELECT `DEPTNO`, `EMP`, `EMP.EMPNAME` `EMPNAME`, `EMP.EMPNO` `EMPNO`\nFROM `scott`.`DEPT`";
+        "SELECT `DEPTNO`, `EMP`, `EMP.EMPNAME`, `EMP.EMPNO`, `EMP.EMPNAME` `EMPNAME`, `EMP.EMPNO` `EMPNO`\nFROM `scott`.`DEPT`";
     verifyPPLToSparkSQL(root, expectedSparkSql);
     String expectedResult =
-        "DEPTNO=10; EMP={7369, SMITH}; EMPNAME=SMITH; EMPNO=7369\n"
-            + "DEPTNO=20; EMP={7499, ALLEN}; EMPNAME=ALLEN; EMPNO=7499\n"
-            + "DEPTNO=30; EMP={7521, WARD}; EMPNAME=WARD; EMPNO=7521\n";
+        "DEPTNO=10; EMP={7369, SMITH}; EMP.EMPNAME=SMITH; EMP.EMPNO=7369; EMPNAME=SMITH; EMPNO=7369\n"
+            + "DEPTNO=20; EMP={7499, ALLEN}; EMP.EMPNAME=ALLEN; EMP.EMPNO=7499; EMPNAME=ALLEN; EMPNO=7499\n"
+            + "DEPTNO=30; EMP={7521, WARD}; EMP.EMPNAME=WARD; EMP.EMPNO=7521; EMPNAME=WARD; EMPNO=7521\n";
```
LantaoJin: wait, why the `flatten` output changed

LantaoJin (commented):
Please add integration tests for this enhancement and update documentation (add a new section in [endpoint.md](https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/interfaces/endpoint.md)

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLFlattenTest.java
ishag4: This was an unintended regression from metadata handling changes.
The flatten operator internally creates intermediate nested fields (`EMP.EMPNAME`, `EMP.EMPNO`) which should be cleaned up before the final projection. 
Reverted.

ishag4:
Hi @LantaoJin @penghuo @RyanL1997 @Swiddis Could you please re-review?

integ-test/src/test/java/org/opensearch/sql/ppl/IncludeMetadataIT.java
```diff
@@ -0,0 +1,248 @@
...
+    Response response = client().performRequest(request);
+    assertEquals(200, response.getStatusLine().getStatusCode());
+    return jsonify(TestUtils.getResponseBody(response, true));
+  }
+
+  private JSONObject executeQueryWithParams(String query, String paramName, String paramValue)
+      throws IOException {
+    String endpoint = String.format("/_plugins/_ppl?%s=%s", paramName, paramValue);
+    Request request = new Request("POST", endpoint);
+    request.setJsonEntity(String.format(Locale.ROOT, "{\n" + "  \"query\": \"%s\"\n" + "}", query));
```
Swiddis: issue: this doesn't handle string escaping

Prefer using a proper json serializer (jackson is available in the classpath)

core/src/main/java/org/opensearch/sql/executor/execution/QueryPlan.java
```diff
@@ -50,12 +52,25 @@ public QueryPlan(
...
+
+  /** Constructor with highlight config and include metadata flag. */
+  public QueryPlan(
+      QueryId queryId,
+      QueryType queryType,
+      UnresolvedPlan plan,
+      QueryService queryService,
+      ResponseListener<ExecutionEngine.QueryResponse> listener,
+      HighlightConfig highlightConfig,
+      boolean includeMetadata) {
```
Swiddis, line 67: suggestion: It may be worth merging this with HighlightConfig and renaming it to something like QueryConfig (with `fieldNames` -> `highlightFieldNames` and `toFieldMap` -> `highlightFieldMap`)

Semantically they're similar in that they provide different ways the query might run, and this means we don't now need to be passing two parameters everywhere. It'd overall reduce the one-liner changes of "the same code but add a field" and make it easier to add more config in the future.

docs/user/ppl/interfaces/endpoint.md
```diff
@@ -325,3 +325,138 @@ Expected output (trimmed):
...
+
+### Example 1: Default behavior (exclude metadata)
+
+```bash ppl
+curl -sS -H 'Content-Type: application/json' \
+-X POST localhost:9200/_plugins/_ppl \
+-d '{"query" : "source=accounts | fields * | head 1"}'
+```
+
+Expected output (metadata fields excluded):
```
Swiddis: polish: i would remove the non-example, there's plenty of these throughout the rest of the docs

docs/user/ppl/interfaces/endpoint.md
```diff
@@ -325,3 +325,138 @@ Expected output (trimmed):
   "rulesToVisit": [200, 201, "..."]
 }
 ```
+
+## Include Metadata
+
+### Description
+
+You can add an `include_metadata` parameter to the PPL request URL to control whether metadata fields (such as `_id`, `_index`, `_score`, etc.) are included in wildcard field selections. This parameter only affects implicit field selections using `fields *` and does not impact explicit field selections.
```
Swiddis: suggestion (if-minor): It might be worthwhile to also support setting it via the request body? (as `{"query": "...", "include_metadata": true}`).

We support that type of duality for `format`, for example. It might simplify some types of request builder code, e.g. last I checked for most of the frontend we already work with request bodies directly and don't change the URL as much.

In a perfect world we'd have a shared request parser that uses both locations as sources for config with some well-defined precedence. I'm not sure how different the two parsing paths are today.

Swiddis (changes requested):
One issue, a few suggestions & polish

core/src/main/java/org/opensearch/sql/executor/execution/QueryPlan.java
ishag4: Good suggestion.
I agree a shared `QueryConfig` (or similar) would likely scale better than continuing to thread individual execution flags through constructors/methods.

We can avoid doing that in this PR mainly to keep the scope focused on `include_metadata` behavior itself and minimize the amount of unrelated refactoring/review surface.

That said, now that we have both `HighlightConfig` and `includeMetadata`, I can definitely see the benefit of consolidating execution-time options before more flags get added. 

Happy to follow up with a separate cleanup/refactor PR if that direction makes sense.

integ-test/src/test/java/org/opensearch/sql/ppl/IncludeMetadataIT.java
ishag4: Thank you for the suggestion.
Updated this to use Jackson serialization

docs/user/ppl/interfaces/endpoint.md
ishag4: Removed, thanks!

docs/user/ppl/interfaces/endpoint.md
ishag4: Great suggestion! Actually, this feature is already implemented. The include_metadata parameter supports both URL parameter and request body JSON with proper precedence handling:

URL Parameter:

```
POST /_plugins/_ppl?include_metadata=true
{
  "query": "source=accounts | fields * | head 5"
}

```

Request Body JSON:

```
POST /_plugins/_ppl
{
  "query": "source=accounts | fields * | head 5",
  "include_metadata": true
}
```

Precedence: Request body takes precedence over URL parameter.

ishag4: Updated docs.

Swiddis (dismissed):

ishag4:
Hi @Swiddis @LantaoJin @penghuo, could you please re-review? A few pipelines were failing, and I’ve pushed the necessary fixes. The workflows are now awaiting approval.

mengweieric:
@ishag4 please check failing CIs

ishag4:
Hi @mengweieric Can you please re-trigger the CIs?

ishag4:
Hi @mengweieric @Swiddis @LantaoJin @penghuo Could you please review and approve this PR?

docs/user/ppl/interfaces/endpoint.md
```diff
@@ -441,3 +325,214 @@ Expected output (trimmed):
...
+  ],
+  "total": 1,
+  "size": 1
+}
+```
+
+### Notes
+
+- The `include_metadata` parameter only affects wildcard field selections (`fields *`).
+- **Current limitation**: Explicit metadata field selection (e.g., `fields firstname, _id`) currently requires `include_metadata=true` to work properly.
```
dai-chen: Is this new limitation or a breaking change on existing behavior? e.g., perviously explicitly select meta field works but now it requires this parameter?

ishag4: That note was inaccurate and I've removed it.
Explicitly selected metadata fields are returned regardless of the parameter, so there's no change to existing behaviour. 
Verified on a live cluster:
```
 POST /_plugins/_ppl?include_metadata=false
{"query": "source=my_index | fields field1, _id"}
 → schema: ["field1", "_id"]
```

In fairness to your instinct: an earlier revision of this PR did introduce a regression here.

integ-test/src/test/java/org/opensearch/sql/ppl/IncludeMetadataIT.java
```diff
@@ -0,0 +1,423 @@
...
+import java.util.Map;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+import org.opensearch.client.Request;
+import org.opensearch.client.RequestOptions;
+import org.opensearch.client.Response;
+import org.opensearch.sql.legacy.TestUtils;
+
+public class IncludeMetadataIT extends PPLIntegTestCase {
```
LantaoJin, line 32: please add a class `CalciteIncludeMetadataIT extends IncludeMetadataIT` in `integ-test/src/test/java/org/opensearch/sql/calcite/remote/`. And add the `CalciteIncludeMetadataIT.class` to `CalciteNoPushdownIT.java`

ishag4: added

ishag4:
Hi @mengweieric @Swiddis @LantaoJin @penghuo Please trigger CIs. Also, Could you please review and approve this PR?

Swiddis (approved):

PR #5413, opened 2026-05-06
Default plugins.calcite.enabled=true on the unified query path

ahkcs:
## Summary

Companion to [opensearch-project/OpenSearch#21521](https://github.com/opensearch-project/OpenSearch/pull/21521) (sort/aggregation/eval-predicate scalar surface) and [opensearch-project/OpenSearch#21498](https://github.com/opensearch-project/OpenSearch/pull/21498) (eval string concat / CAST). Two changes — both **strictly scoped to PPL `fields` / `rename` / `head` / `sort`**; aggregation work (`AVG` / `STDDEV_*` / `VAR_*` / `stats` surface) is explicitly **out of scope**.

### 1. `UnifiedQueryContext` — default `plugins.calcite.enabled=true`

The unified PPL parser reuses the v2 `AstBuilder`, which gates `table` / `regex` / `rex` / `convert` on this setting. The unified context built empty maps without it, so analytics-routed `table` queries 500'd with `Table command is supported only when plugins.calcite.enabled=true` even when the cluster setting was true.

Per [@dai-chen's review](https://github.com/opensearch-project/sql/pull/5413#discussion_r...): the default goes in the `Map.of(...)` initializer at line 123 (alongside `QUERY_SIZE_LIMIT`/`PPL_SUBSEARCH_MAXOUT`/`PPL_JOIN_SUBSEARCH_MAXOUT`), not as a special case in `getSettingValue`.

### 2. `CalcitePPLRenameIT` — column-name-keyed row matcher

Previously, `verifyStandardDataRows` and a handful of bespoke `verifyDataRows` calls compared rows positionally — they assumed the engine emits columns in the v2 / Lucene `_source` iteration order. Under the analytics-engine route the parquet-backed reader returns columns in storage order, so the same 4 canonical state_country rows came back as `[70,"USA",4,"Jake","California",2023]` instead of `[Jake,USA,California,4,2023,70]` and 12 of 22 tests failed despite the data being identical.

Replace with `verifyStandardDataRows(result, canonicalColumns...)` plus a generic `verifyDataRowsByColumn(result, Map<String, Object>...)` helper. Both read the actual schema from the response, look up each canonical value by column name, and place it at the corresponding schema position before comparing — the test no longer leaks the engine's emission order into the assertion.

Verified working on **both paths**: 20 / 22 pass on each. The 2 remaining failures (`testRenameInAgg`, `testRenameWithBackticksInAgg`) are the `Unable to find binding for call AVG` aggregation issue — **out of scope for this PR**; tracked in [#21521](https://github.com/opensearch-project/OpenSearch/pull/21521)'s "Out of scope" section.

## Test plan

- [x] `./gradlew :api:check` green
- [x] Manual: with `plugins.calcite.analytics.force_routing=true`, `POST /_plugins/_ppl {\"query\":\"source=test_eval | table name\"}` no longer 500s with the `plugins.calcite.enabled=true` error
- [x] `:integ-test:integTestRemote` against an analytics cluster carrying #21521 + #21498, both with and without `-Dtests.analytics.force_routing=true -Dtests.analytics.parquet_indices=true`:

| IT | Analytics path | v2 / Calcite path | Notes |
|---|---|---|---|
| `CalciteFieldsCommandIT` | 39 / 39 ✅ | 39 / 39 ✅ | scope of this PR |
| `CalciteRenameCommandIT` | 2 / 2 ✅ | 2 / 2 ✅ | scope of this PR |
| `CalciteHeadCommandIT` | 4 / 4 ✅ (2 skipped) | 4 / 4 ✅ | scope of this PR |
| `CalcitePPLRenameIT` | 20 / 22 | 20 / 22 | column-order helper added by this PR; remaining 2 are AVG-binding (out of scope) |
| `CalciteEvalCommandIT` | 4 / 4 ✅ | 4 / 4 ✅ | scope of #21498 |
| `CalcitePPLSortIT` | 14 / 18 | 18 / 18 ✅ | 4 analytics-side fails on tie-break ordering / date-format diff — separate follow-up |
| `CalciteSortCommandIT` | 26 / 30 | 30 / 30 ✅ | 4 analytics-side fails on `weblogs` IP-type scan — separate follow-up |

## Companion routing-and-coverage workflow docs

The `docs/dev/ppl-analytics-engine-{routing,coverage-sop}.md` files originally bundled with the first revision of this PR were dropped per [@dai-chen's review](https://github.com/opensearch-project/sql/pull/5413#discussion_r...) — they belong on a separate doc PR.

## Out of scope

Aggregation surface — `AVG`, `SUM`, `COUNT`, `STDDEV_*`, `VAR_*`, the `stats` family, `count(eval(...))`, etc. — is not addressed by this PR. The `Unable to find binding for call AVG($N)` Substrait isthmus issue is a cross-plugin design discussion (the SQL plugin's `NullableSqlAvgAggFunction` extends `SqlAggFunction` directly rather than `SqlAvgAggFunction`, so isthmus' built-in translator skips it). Tracked in [#21521](https://github.com/opensearch-project/OpenSearch/pull/21521).


api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
```diff
@@ -225,6 +225,14 @@ private Settings buildSettings() {
         @Override
         @SuppressWarnings("unchecked")
         public <T> T getSettingValue(Key key) {
+          // The unified query path is always Calcite-based — every query reaching this context
+          // goes through Calcite's planner, never the v2 engine. Default
+          // {@code plugins.calcite.enabled} to true so AstBuilder gates (e.g. visitTableCommand)
+          // that protect v2-only code paths don't block valid analytics-engine queries when the
+          // underlying setting hasn't been propagated through Builder#setting.
+          if (key == Key.CALCITE_ENGINE_ENABLED && !settings.containsKey(key)) {
```
dai-chen: Either add it to default setting list or pass via `setting()` API since only unified PPL require this?

Ref:
1. https://github.com/opensearch-project/sql/blob/main/api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java#L123
2. https://github.com/opensearch-project/sql/blob/main/integ-test/src/test/java/org/opensearch/sql/api/UnifiedQueryOpenSearchIT.java#L51

ahkcs: Updated

PR #5418, opened 2026-05-07
Bridge PPL_REX_MAX_MATCH_LIMIT into UnifiedQueryContext on the unified query path

RyanL1997:
## Description

The PPL `rex` command's `AstBuilder` reads `Settings.Key.PPL_REX_MAX_MATCH_LIMIT` unconditionally and unboxes the result to `int`. On the unified query path, `UnifiedQueryContext` builds its `Settings` map with only a small whitelist of planning-required keys; for any unregistered key, `getSettingValue` returns null and the auto-unbox NPEs the planner before any operator-level capability check runs. Every `rex` query through `/_analytics/ppl` hits this NPE today.

This PR ships two changes that together let the unified path execute `rex` correctly:

### 1. Default `PPL_REX_MAX_MATCH_LIMIT=10` in `UnifiedQueryContext`

Adds the key to the static settings map so `AstBuilder.visitRexCommand` no longer NPEs. The value mirrors the cluster-side default of `10` registered by `OpenSearchSettings.PPL_REX_MAX_MATCH_LIMIT_SETTING`, so unified-path behavior matches v2-path behavior when neither has an explicit cluster override. Mirrors the precedent Kai introduced for `CALCITE_ENGINE_ENABLED` in #5413.

### 2. Bridge live cluster settings for `PPL_REX_MAX_MATCH_LIMIT` only

Without this, every key in the static map resolves to its hardcoded default and `_cluster/settings` updates are invisible to the unified path. `CalciteRexCommandIT.testRexMaxMatchConfigurableLimit` exercises this: it sets the cluster-side limit to 5 and asserts that `max_match=0` caps at 5, but on the unified path the static `10` keeps winning.

Adds a `Builder.liveSettings(Settings)` hook so the REST handler can inject the cluster's live `OpenSearchSettings` instance. At `build()` time the Builder snapshots the live value of `PPL_REX_MAX_MATCH_LIMIT` into the static map, overriding the hardcoded default when the operator has set a cluster value. Snapshot-at-build matches the per-HTTP-request lifecycle of `UnifiedQueryContext` and avoids per-call lookup overhead.

`RestUnifiedQueryAction` gains a `pluginSettings` field (the same `OpenSearchSettings` instance bound in the Guice module) and forwards it to the Builder in both `buildContext` and `buildParsingContext`. Both construction sites — `SQLPlugin.createSqlAnalyticsRouter` and `TransportPPLQueryAction.<init>` — are updated.

#### Why scoped to `PPL_REX_MAX_MATCH_LIMIT` only

The same architectural gap exists for every key in the static map (`QUERY_SIZE_LIMIT`, `PPL_SUBSEARCH_MAXOUT`, `PPL_JOIN_SUBSEARCH_MAXOUT`, `CALCITE_ENGINE_ENABLED`). For three of those, the static defaults are fine in practice (no test overrides them mid-run; `head N` covers `QUERY_SIZE_LIMIT` per-query). `CALCITE_ENGINE_ENABLED` is intentionally pinned to `true` for the unified path — a cluster override toggling it off would defeat the point of routing here. So this PR widens only the one key that demonstrably needs it; widening the snapshot to the rest is a future scope decision tied to whichever new IT first depends on it.

## Companion PR

[opensearch-project/OpenSearch#21550](https://github.com/opensearch-project/OpenSearch/pull/21550) — onboards PPL `rex` to DataFusion via the analytics-engine path. Without this PR's fixes, every rex query through `/_analytics/ppl` NPEs at parse time and never reaches the planner.

## Test results

`CalciteRexCommandIT` through the analytics-engine route (every PPL query forced through `/_analytics/ppl` via `tests.analytics.force_routing=true`):

* Without this PR: 0/18 — every test NPEs in `AstBuilder.visitRexCommand`.
* With change 1 only (default 10): 17/18 — `testRexMaxMatchConfigurableLimit` fails with `expected:<5> but was:<10>`.
* With both changes: **18/18, 100%** — all `testRexMaxMatch*` variants honor the cluster setting.

Signed-off-by: Jialiang Liang <jiallian@amazon.com>

### DIFF
--- a/api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
+++ b/api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
@@ -131,10 +131,12 @@ public class UnifiedQueryContext implements AutoCloseable {
      *
      * <p>{@link Settings.Key#PPL_REX_MAX_MATCH_LIMIT} defaults to {@code 10} here because {@code
      * AstBuilder.visitRexCommand} reads it unconditionally and unboxes to {@code int} — a {@code
-     * null} return from {@code getSettingValue} NPEs the planner before any operator- level
+     * null} return from {@code getSettingValue} NPEs the planner before any operator-level
      * capability check runs. The value mirrors the cluster-side default of {@code 10} registered by
-     * {@code OpenSearchSettings.PPL_REX_MAX_MATCH_LIMIT_SETTING}, so unified-path behavior matches
-     * v2-path behavior when neither has an explicit cluster override.
+     * {@code OpenSearchSettings.PPL_REX_MAX_MATCH_LIMIT_SETTING}. Cluster-side overrides reach this
+     * map via {@link #setting(String, Object)} — the REST handler reads the live value from {@code
+     * OpenSearchSettings} and routes it through that existing API, keeping {@link
+     * UnifiedQueryContext} decoupled from any specific {@link Settings} implementation.
      */
     private final Map<Settings.Key, Object> settings =
         new HashMap<Settings.Key, Object>(
@@ -232,6 +234,7 @@ public class UnifiedQueryContext implements AutoCloseable {
             case SQL -> UnifiedSqlSpec.extended();
             case PPL -> UnifiedPplSpec.create();
           };
+
       Settings settings = buildSettings();
       CalcitePlanContext planContext =
           CalcitePlanContext.create(
--- a/api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
+++ b/api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
@@ -7,6 +7,7 @@ package org.opensearch.sql.api;
 
 import static org.opensearch.sql.common.setting.Settings.Key.CALCITE_ENGINE_ENABLED;
 import static org.opensearch.sql.common.setting.Settings.Key.PPL_JOIN_SUBSEARCH_MAXOUT;
+import static org.opensearch.sql.common.setting.Settings.Key.PPL_REX_MAX_MATCH_LIMIT;
 import static org.opensearch.sql.common.setting.Settings.Key.PPL_SUBSEARCH_MAXOUT;
 import static org.opensearch.sql.common.setting.Settings.Key.QUERY_SIZE_LIMIT;
 
@@ -127,6 +128,13 @@ public class UnifiedQueryContext implements AutoCloseable {
      * org.opensearch.sql.api.parser.PPLQueryParser} reuses the v2 {@code AstBuilder}, which gates
      * Calcite-only commands (e.g. {@code visitTableCommand}) on this setting; without the default,
      * those commands fail at parse time even when the cluster setting is true.
+     *
+     * <p>{@link Settings.Key#PPL_REX_MAX_MATCH_LIMIT} defaults to {@code 10} here because {@code
+     * AstBuilder.visitRexCommand} reads it unconditionally and unboxes to {@code int} — a {@code
+     * null} return from {@code getSettingValue} NPEs the planner before any operator- level
+     * capability check runs. The value mirrors the cluster-side default of {@code 10} registered by
+     * {@code OpenSearchSettings.PPL_REX_MAX_MATCH_LIMIT_SETTING}, so unified-path behavior matches
+     * v2-path behavior when neither has an explicit cluster override.
      */
     private final Map<Settings.Key, Object> settings =
         new HashMap<Settings.Key, Object>(
@@ -134,7 +142,8 @@ public class UnifiedQueryContext implements AutoCloseable {
                 QUERY_SIZE_LIMIT, SysLimit.DEFAULT.querySizeLimit(),
                 PPL_SUBSEARCH_MAXOUT, SysLimit.DEFAULT.subsearchLimit(),
                 PPL_JOIN_SUBSEARCH_MAXOUT, SysLimit.DEFAULT.joinSubsearchLimit(),
-                CALCITE_ENGINE_ENABLED, true));
+                CALCITE_ENGINE_ENABLED, true,
+                PPL_REX_MAX_MATCH_LIMIT, 10));
 
     /**
      * Sets the query language frontend to be used.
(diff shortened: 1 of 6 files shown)


ahkcs (approved):

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
```diff
@@ -210,6 +220,25 @@ public Builder setting(String name, Object value) {
...
+     *
+     * <p>Scoped to {@link Settings.Key#PPL_REX_MAX_MATCH_LIMIT} for now — required so
+     * {@code rex max_match=0} caps at the operator-configured limit (CalciteRexCommandIT's
+     * testRexMaxMatchConfigurableLimit). The other keys in {@link #settings} keep their
+     * static defaults; widening the snapshot to all keys is a separate scope.
+     *
+     * @param liveSettings the cluster's live {@code OpenSearchSettings} instance
+     * @return this builder instance
+     */
+    public Builder liveSettings(Settings liveSettings) {
```
penghuo: liveSettings -> settings.

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
```diff
@@ -223,6 +252,19 @@ public UnifiedQueryContext build() {
...
+      // semantics (not per-call lookup) — UnifiedQueryContext is built per HTTP
+      // request, so a one-shot read at build() is sufficient. Other keys are
+      // intentionally not bridged here; that's a follow-up scope decision.
+      if (liveSettings != null) {
+        Object liveRexLimit = liveSettings.getSettingValue(PPL_REX_MAX_MATCH_LIMIT);
+        if (liveRexLimit != null) {
+          settings.put(PPL_REX_MAX_MATCH_LIMIT, liveRexLimit);
+        }
+      }
+
```
penghuo: Merge logic should in buildSettings()

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
RyanL1997: good suggestion, renamed.

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
RyanL1997: moved.

penghuo (approved):

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
```diff
@@ -240,16 +269,27 @@ private UnifiedQueryParser<?> createParser(CalcitePlanContext planContext, Setti
...
+      // mid-test `_cluster/settings` PUT takes effect on the unified path. Snapshot
+      // semantics (not per-call lookup) — UnifiedQueryContext is built per HTTP
+      // request, so a one-shot read at build() is sufficient. Other keys are
+      // intentionally not bridged here; that's a follow-up scope decision.
+      if (settings != null) {
+        Object liveRexLimit = settings.getSettingValue(PPL_REX_MAX_MATCH_LIMIT);
+        if (liveRexLimit != null) {
+          defaults.put(PPL_REX_MAX_MATCH_LIMIT, liveRexLimit);
+        }
+      }
```
dai-chen: Is this required since `PPL_REX_MAX_MATCH_LIMIT` already in the `defaults`?

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
```diff
@@ -206,7 +216,25 @@ public Builder setting(String name, Object value) {
...
+     *
+     * <p>Scoped to {@link Settings.Key#PPL_REX_MAX_MATCH_LIMIT} for now — required so {@code rex
+     * max_match=0} caps at the operator-configured limit (CalciteRexCommandIT's
+     * testRexMaxMatchConfigurableLimit). The other keys in {@link #defaults} keep their static
+     * defaults; widening the snapshot to all keys is a separate scope.
+     *
+     * @param settings the cluster's live {@code OpenSearchSettings} instance
+     * @return this builder instance
+     */
+    public Builder settings(Settings settings) {
```
dai-chen: If possible, can we avoid this new API because of `Settings` in argument? We want to keep it internal and may decouple from it later.

api/src/test/java/org/opensearch/sql/api/UnifiedQueryContextTest.java
```diff
@@ -83,6 +87,99 @@ public void testInvalidDefaultNamespace() {
...
+            .build();
+
+    assertEquals(
+        "Cluster override should win over the static default",
+        Integer.valueOf(5),
+        context.getSettings().getSettingValue(PPL_REX_MAX_MATCH_LIMIT));
+  }
+
+  @Test
+  public void testLiveSettingsAbsentFallsBackToStaticDefault() {
```
dai-chen: Test default value in existing `testContextCreationWithDefaults`?

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
RyanL1997: Good catch — removed in latest push. The REST handler now writes through the existing `setting(String, Object)` API directly, which overwrites the default-10 entry in the same map when the operator has a cluster value.

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
RyanL1997: Agreed — dropped `settings(Settings)` entirely. REST handler routes the cluster value through the existing `setting(String, Object)` API:
```
  Object rexLimit = pluginSettings.getSettingValue(PPL_REX_MAX_MATCH_LIMIT);
  if (rexLimit != null) {
    builder.setting(PPL_REX_MAX_MATCH_LIMIT.getKeyValue(), rexLimit);
  }
```
UnifiedQueryContext stays decoupled from any specific Settings impl. Also makes @penghuo's earlier "merge logic in buildSettings()" comment moot since there's no merge logic anymore.

api/src/test/java/org/opensearch/sql/api/UnifiedQueryContextTest.java
RyanL1997: Done — folded into existing tests:
* `testContextCreationWithDefaults` now also asserts PPL_REX_MAX_MATCH_LIMIT defaults to 10.
* `testContextCreationWithCustomConfig` now also asserts that `.setting("plugins.ppl.rex.max_match.limit", 5)` resolves to 5.

Dropped the four standalone tests — they exercised the removed API.

dai-chen (approved):

PR #5421, opened 2026-05-08
Default empty array() return type to ARRAY<VARCHAR>

ahkcs:
### Description

PPL's `array()` no-arg form currently returns `ARRAY<NULL>` (and `ARRAY<UNKNOWN>` when all operands are typeless nulls). The Calcite engine's in-process executor tolerates either: `ArrayImplementor.internalCast` only consumes the element type when there are elements to cast, so an empty list flows straight through to `ExprCollectionValue` regardless of the declared element type.

The **analytics-engine route** is stricter. When isthmus walks a RexCall like `mvjoin(array(), '-')`, it eventually feeds the first operand's type to `io.substrait.isthmus.TypeConverter.toSubstrait`, which throws `UnsupportedOperationException: Unable to convert the type UNKNOWN`. Substrait has no on-wire encoding for NULL/UNKNOWN element types, so the planner can't serialize the call at all. Two ITs hit this directly:

- `CalciteArrayFunctionIT.testMvjoinWithEmptyArray`
- `CalciteArrayFunctionIT.testMvdedupWithEmptyArray`

Substituting `VARCHAR` when the inferred element type is `NULL` or `UNKNOWN` gives the call a substrait-serializable type without affecting any value computation — the result list is empty either way.

### Test plan
- `:core:test --tests "*ArrayFunction*"` — passes (no existing test asserted on the empty-array element type).
- `CalciteArrayFunctionIT` force-routed through the analytics-engine path via opensearch-project/OpenSearch#21554's plugin set — pass-rate **26/60 → 28/60**, with `testMvjoinWithEmptyArray` and `testMvdedupWithEmptyArray` newly green.

Companion to opensearch-project/OpenSearch#21554.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
@@ -50,14 +50,7 @@ public class ArrayFunctionImpl extends ImplementorUDF {
         RelDataType originalType =
             SqlLibraryOperators.ARRAY.getReturnTypeInference().inferReturnType(sqlOperatorBinding);
         RelDataType innerType = originalType.getComponentType();
-        // For empty `array()` Calcite infers element type as NULL, which downstream
-        // serializers (notably the analytics-engine route's substrait converter)
-        // reject with "Unable to convert the type UNKNOWN". Default to VARCHAR — the
-        // result is empty either way, so the chosen scalar element type doesn't
-        // affect any value computation, but it gives the call a substrait-serializable
-        // type. Existing v2-engine tests (which feed Object lists straight through to
-        // ExprCollectionValue) are unaffected because the empty list contains no
-        // elements that need to be cast.
+        // Default empty/unknown element type to VARCHAR — see PR description for why.
         if (innerType == null || isUnknownLikeType(innerType.getSqlTypeName())) {
           innerType = typeFactory.createSqlType(SqlTypeName.VARCHAR);
         }
--- a/core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
@@ -50,6 +50,17 @@ public class ArrayFunctionImpl extends ImplementorUDF {
         RelDataType originalType =
             SqlLibraryOperators.ARRAY.getReturnTypeInference().inferReturnType(sqlOperatorBinding);
         RelDataType innerType = originalType.getComponentType();
+        // For empty `array()` Calcite infers element type as NULL, which downstream
+        // serializers (notably the analytics-engine route's substrait converter)
+        // reject with "Unable to convert the type UNKNOWN". Default to VARCHAR — the
+        // result is empty either way, so the chosen scalar element type doesn't
+        // affect any value computation, but it gives the call a substrait-serializable
+        // type. Existing v2-engine tests (which feed Object lists straight through to
+        // ExprCollectionValue) are unaffected because the empty list contains no
+        // elements that need to be cast.
+        if (innerType == null || isUnknownLikeType(innerType.getSqlTypeName())) {
+          innerType = typeFactory.createSqlType(SqlTypeName.VARCHAR);
+        }
         return createArrayType(
             typeFactory, typeFactory.createTypeWithNullability(innerType, true), true);
       } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


dai-chen (commented):
Can we add unit test in `ArrayFunctionImplTest`?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
```diff
@@ -50,6 +50,17 @@ public SqlReturnTypeInference getReturnTypeInference() {
         RelDataType originalType =
             SqlLibraryOperators.ARRAY.getReturnTypeInference().inferReturnType(sqlOperatorBinding);
         RelDataType innerType = originalType.getComponentType();
+        // For empty `array()` Calcite infers element type as NULL, which downstream
```
dai-chen: np: If possible, can we make the comment more concise. I think it's okay to leave most context/details in PR description.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
```diff
@@ -50,6 +50,17 @@ public SqlReturnTypeInference getReturnTypeInference() {
...
+        // serializers (notably the analytics-engine route's substrait converter)
+        // reject with "Unable to convert the type UNKNOWN". Default to VARCHAR — the
+        // result is empty either way, so the chosen scalar element type doesn't
+        // affect any value computation, but it gives the call a substrait-serializable
+        // type. Existing v2-engine tests (which feed Object lists straight through to
+        // ExprCollectionValue) are unaffected because the empty list contains no
+        // elements that need to be cast.
+        if (innerType == null || isUnknownLikeType(innerType.getSqlTypeName())) {
+          innerType = typeFactory.createSqlType(SqlTypeName.VARCHAR);
+        }
```
penghuo, line 56: The idea is CAST(ARRAY[] to VARCHAR[])? Add an UT to cover this assumption.
Question, Is it limitation of Substrait, becuase it does not supoort UNKNOW?

penghuo: Refer PostgreSQL, it does not allow construct empty ARRAY[]. 

https://www.postgresql.org/docs/current/sql-expressions.html#SQL-SYNTAX-ARRAY-CONSTRUCTORS

```
You can construct an empty array, but since it's impossible to have an array with no type, you must explicitly cast your empty array to the desired type. For example:

SELECT ARRAY[]::integer[];
 array
-------
 {}
(1 row)
```

But in PPL, we allow create emtpy array(). So which means we implict covert to varchar[] in future, right?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
ahkcs: Trimmed in 22d0cce67 — comment now just points to the PR description.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ArrayFunctionImpl.java
ahkcs: Yes — semantically equivalent to `CAST(ARRAY[] AS VARCHAR[])`. The fallback path adds an empty-typed-array case Calcite/Postgres handle via explicit cast, but PPL surface allows bare `array()` so we do it implicitly.

Yes, Substrait limitation — Substrait's `Type` proto has no encoding for `UNKNOWN` (or Calcite's `NULL` SqlTypeName as an element type). When isthmus' `TypeConverter.toSubstrait` walks an `ARRAY<UNKNOWN>` it can't produce a valid wire type and throws `UnsupportedOperationException: Unable to convert the type UNKNOWN`. The Calcite-engine local executor is more lenient because it never serializes the type — it just iterates the empty `List<Object>` and never reads the element type.

UT added in aa82704d9 — `testReturnTypeForEmptyCallIsVarcharArray` and `testReturnTypeForAllNullOperandsIsVarcharArray` cover both fallback paths (empty operand list and typeless-NULL operand). `testReturnTypeForIntegerOperandPreservesType` is the regression guard that confirms concrete element types pass through unchanged.

ahkcs: Right — that's the trade-off. Postgres requires the explicit cast (`SELECT ARRAY[]::integer[]`); PPL doesn't have that surface today, so any caller writing bare `array()` would otherwise hit an analytics-engine-side error. The implicit VARCHAR default keeps the bare form working, and concrete-element calls (`array(1, 2)`, `array('a')`) keep their original element type — `testReturnTypeForIntegerOperandPreservesType` in aa82704d9 is the regression guard for that.

Long-term we could expose a PPL `cast(array() as <T>[])` syntax that mirrors Postgres and remove the implicit default, but that's a separate language-surface change.

ahkcs:
Pushed aa82704d9 (UTs) and 22d0cce67 (trim inline comment).

Replies to the open questions:

**@dai-chen — "with this change, we actually define VARCHAR[] array(), right?"**

Yes, but only for the bare/empty form. Concrete-element calls keep the inferred element type:
- `array()`           → `ARRAY<VARCHAR>` (was `ARRAY<NULL>`)
- `array(NULL)`       → `ARRAY<VARCHAR>` (was `ARRAY<UNKNOWN>`)
- `array(1)`          → `ARRAY<INTEGER>` (unchanged)
- `array('a', 'b')`   → `ARRAY<VARCHAR>` (unchanged)
- `array(1, 1.5)`     → `ARRAY<DOUBLE>` (unchanged — leastRestrictive widening from `SqlLibraryOperators.ARRAY` still wins)

`testReturnTypeForIntegerOperandPreservesType` in aa82704d9 is the regression guard for the third case.

**@penghuo — "CAST(ARRAY[] to VARCHAR[])? Substrait limitation?"**

Yes to both, replied inline.

**@penghuo — "Postgres requires `ARRAY[]::integer[]`, PPL implicit VARCHAR[]?"**

Yes — PPL surface allows bare `array()` and we silently default to `VARCHAR[]`. Replied inline with the rationale and a follow-up note about adding explicit `cast(array() as <T>[])` syntax later.

penghuo (approved):

PR #5426, opened 2026-05-08
Switch sandbox deps to analytics-api (JDK 21 surface)

bowenlan-amzn:
## Summary

sql depends on two interfaces from analytics-engine: `QueryPlanExecutor` and `OpenSearchSchemaBuilder`. Previously these were vendored as local JARs. This PR:

1. **Replaces vendored JARs with `analytics-api`** — a new JDK-21-targeted Maven module that exposes only those two interfaces. Depends on opensearch-project/OpenSearch#21574.
2. **Removes analytics-engine from standard test clusters** — it was never needed for sql's own integration tests. Standard `integTest` runs without it.
3. **Adds `analyticsEngineCompatIT`** — a dedicated smoke test that boots sql alongside analytics-engine + arrow-flight-rpc to verify the three plugins load together without classloader errors.

## Root cause of the original JDK mismatch

`analytics-framework` and `analytics-engine` target JDK 25 (they use `java.lang.foreign`). sql targets JDK 21. Consuming their JARs directly caused `class file has wrong version 69.0, should be 65.0` on the JDK-21 CI matrix. The fix is `analytics-api`: a thin module containing only the interfaces sql needs, compiled at JDK 21.

## Why `analytics-api` is `implementation` (not `compileOnly`) in plugin

OpenSearch scans plugin classes at startup by calling `getMethods()`. sql's handler classes reference `QueryPlanExecutor` in their method signatures. Without `analytics-api` bundled in the plugin zip, OpenSearch throws `NoClassDefFoundError: QueryPlanExecutor` at startup even when no analytics-engine code path runs. When analytics-engine IS installed, its classloader provides the class via `extendedPlugins` parent-first delegation — the bundled copy is unused but harmless.

## Verification

Tested locally with `analytics-api` published to `mavenLocal` from the OpenSearch PR branch:

- [x] `./gradlew :core:compileJava` — BUILD SUCCESSFUL
- [x] `./gradlew :opensearch-sql-plugin:bundlePlugin` — BUILD SUCCESSFUL  
- [x] `./gradlew :integ-test:analyticsEngineCompatIT` — BUILD SUCCESSFUL (cluster boots, all three plugins load)
- [ ] CI will turn green once OpenSearch#21574 merges and `analytics-api-3.7.0-SNAPSHOT` is published to the snapshot Maven repo.

## Depends on

- opensearch-project/OpenSearch#21574 — Extract `analytics-api` module targeting JDK 21
- #5425 — Switch analytics-engine/framework resolution from vendored libs to remote Maven


build.gradle
```diff
@@ -183,6 +210,30 @@ subprojects {
...
+            }
+            withModule('org.opensearch.sandbox:analytics-engine') { details ->
+                details.allVariants {
+                    attributes {
+                        attribute(org.gradle.api.attributes.java.TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE, 21)
+                    }
+                }
+            }
+        }
+    }
```
bowenlan-amzn: @peterzhuamazon

build.gradle
```diff
@@ -117,6 +118,26 @@ spotless {
     }
 }
 
+// Remote URL for the analytics-engine plugin ZIP. The OpenSearch CI maven snapshot repo only
+// publishes the jar/pom/sources/javadoc, not the plugin distribution zip — so we fetch the zip
```
peterzhuamazon:
```suggestion
// publishes the jar/pom/sources/javadoc, not the plugin distribution zip for core plugins — so we fetch the zip
```

ahkcs (approved):

RyanL1997 (approved):

PR #5429, opened 2026-05-10
Route analytics queries by index setting, not table-name prefix

bowenlan-amzn:
### Description

Today \`RestUnifiedQueryAction.isAnalyticsIndex\` routes queries to the analytics engine when the source index name starts with \`parquet_\`. That's brittle — it conflates naming convention with storage type: an index created without the prefix but with pluggable dataformat enabled is silently sent to the Lucene path, and an index named \`parquet_foo\` without the setting is mis-dispatched.

Switch to the authoritative signal: the \`index.pluggable.dataformat.enabled\` cluster-state setting. This is the same flag integration tests (\`CoordinatorReduceIT\`, \`CompositeCommitDeletionIT\`) use to create analytics-backed indices, and what \`FieldStorageResolver\` reads to resolve field storage.

**Routing behavior**

| \`index.pluggable.dataformat.enabled\` | Dispatch |
| --- | --- |
| \`true\` | analytics engine (DataFusion via \`QueryPlanExecutor\`) |
| absent / \`false\` / index missing | sql-plugin Calcite → OpenSearch DSL path |

### Issues Resolved

Mustang rollout pre-work — aligns PPL/SQL routing with the index-setting-based model already used elsewhere in the engine.

### Check List

- [x] New functionality has been documented. (javadoc on \`isAnalyticsIndex\`)

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
@@ -75,8 +75,9 @@ public class RestUnifiedQueryAction {
 
   /**
    * Returns true iff the target index has {@link
-   * IndexSettings#PLUGGABLE_DATAFORMAT_ENABLED_SETTING} set, routing it to DataFusion instead of
-   * the Calcite→DSL path.
+   * IndexSettings#PLUGGABLE_DATAFORMAT_ENABLED_SETTING} set and {@link
+   * IndexSettings#PLUGGABLE_DATAFORMAT_VALUE_SETTING} is {@code "parquet"}, routing it to
+   * DataFusion instead of the Calcite→DSL path.
    *
    * <p>Note: This creates a separate UnifiedQueryContext for parsing. The context cannot be shared
    * with doExecute/doExplain because UnifiedQueryContext holds a Calcite JDBC connection that fails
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -87,18 +89,25 @@ public boolean isAnalyticsIndex(String query, QueryType queryType) {
...
 
+  private String stripSchemaPrefix(String indexName) {
+    int lastDot = indexName.lastIndexOf('.');
+    return lastDot >= 0 ? indexName.substring(lastDot + 1) : indexName;
+  }
+
+  private boolean isPluggableDataformatIndex(String indexName) {
+    var indexMetadata = clusterService.state().metadata().index(indexName);
+    return indexMetadata != null
+        && IndexSettings.PLUGGABLE_DATAFORMAT_ENABLED_SETTING.get(indexMetadata.getSettings());
```
expani:
```
    "index.pluggable.dataformat.enabled": true,
    "index.pluggable.dataformat": "composite",
    "index.composite.primary_data_format": "parquet"
```

I think we also need to check on  `index.composite.primary_data_format` which would indicate if this is Parquet backed index. 

If the pluggable dataformat is enabled `index.pluggable.dataformat.enabled` , but `index.composite.primary_data_format : lucene` , then we can still go the old route until AnalyticsPlugin supports pure Lucene Indices and DocValues.

expani (approved):

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -87,18 +90,29 @@ public boolean isAnalyticsIndex(String query, QueryType queryType) {
...
+  }
+
+  private boolean isPluggableDataformatIndex(String indexName) {
+    var indexMetadata = clusterService.state().metadata().index(indexName);
+    if (indexMetadata == null) {
+      return false;
+    }
+    var settings = indexMetadata.getSettings();
+    return IndexSettings.PLUGGABLE_DATAFORMAT_ENABLED_SETTING.get(settings)
+        && "parquet".equals(IndexSettings.PLUGGABLE_DATAFORMAT_VALUE_SETTING.get(settings));
```
bowenlan-amzn, line 113: PLUGGABLE_DATAFORMAT_VALUE_SETTING should be `composite`

PR #5433, opened 2026-05-11
Close gaps from top/rare analytics-engine wiring

ahkcs:
## Description

Two SQL-plugin–side fixes that close the three out-of-scope failures called out in [opensearch-project/OpenSearch#21593](https://github.com/opensearch-project/OpenSearch/pull/21593) (window-function wiring for `top`/`rare` on the analytics-engine route). With that PR plus these fixes, `CalciteTopCommandIT` and `CalciteRareCommandIT` go from **8/11 → 11/11** against the force-routed analytics-engine path.

### Fix 1 — Forward `PPL_SYNTAX_LEGACY_PREFERRED` through the unified context

`RestUnifiedQueryAction.applyClusterOverrides` previously only forwarded `PPL_REX_MAX_MATCH_LIMIT` into the per-request `UnifiedQueryContext`. As a result, cluster-side updates to `plugins.ppl.syntax.legacy.preferred` were ignored on the analytics-engine route — `PPLQueryParser` → `AstBuilder` → `ArgumentFactory` reads the legacy-preferred flag from the unified context's settings map, which never received the override. Queries like `top age` with the cluster setting flipped to `false` behaved as if `usenull=true` (legacy default) on the analytics route only.

Refactored the override builder into a small `forwardClusterSetting` helper and forward both `PPL_REX_MAX_MATCH_LIMIT` and `PPL_SYNTAX_LEGACY_PREFERRED`. Future keys are now one-liners.

Unblocks `testTopCommandLegacyFalse` and `testRareCommandLegacyFalse`.

### Fix 2 — Stable tie-break for `RareTopN` ROW_NUMBER

`CalciteRelNodeVisitor.visitRareTopN` lowers `rare`/`top` to `ROW_NUMBER() OVER (PARTITION BY ... ORDER BY count [DESC])`. With only the count column in the ORDER BY clause, ties at the same count resolved via the upstream operator's insertion order, which differed between backends (in-process Calcite vs. analytics-engine vs. Lucene pushdown). `testRareWithGroup` failed on the analytics route because ROW_NUMBER picked `NV` at count=8 while the test expected `AR`.

Appended the rare/top field columns as secondary ASC keys so ties resolve alphabetically and deterministically across backends. This matches the existing OpenSearch terms-aggregation pushdown, which already tie-breaks on `_key:asc`.

`RareTopPushdownRule` now accepts the new shape: 1 or 2 order keys, where the optional second key must be the rare/top target field in ASC direction. **The pushdown's wire payload is unchanged** — same OS terms-agg request as before.

Unblocks `testRareWithGroup`.

## Pass-rate impact

Measured on the SQL plugin's `:integTestRemote` task against an externally-managed cluster running [opensearch-project/OpenSearch#21593](https://github.com/opensearch-project/OpenSearch/pull/21593) (`feature/toprare-analytics-verify` @ `114e8bf8e3a`) with `-Dtests.analytics.force_routing=true -Dtests.analytics.parquet_indices=true`:

| IT | Before (PR #21593 alone) | After (PR #21593 + this PR) |
|---|---|---|
| `CalciteTopCommandIT` | 5/6 | **6/6** |
| `CalciteRareCommandIT` | 3/5 | **5/5** |
| **Combined** | **8/11** | **11/11** |

### Regression sweep (all green)

- In-process Calcite (`integTestRemote` without `force_routing`): Top + Rare → 11/11 (no regression)
- Legacy v2 (`RareCommandIT`, `TopCommandIT`): 5/5
- `:ppl:test :core:test :opensearch:test :api:test`: clean
- `CalciteExplainIT` (consumes the 5 updated explain YAML fixtures): clean

## Files

- `plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java` — Fix 1
- `core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java` — Fix 2 (tie-break in lowering)
- `opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/RareTopPushdownRule.java` — Fix 2 (accept the new 2-key shape)
- `ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLRareTopNTest.java` — updated 11 expected RelNode / result / SparkSQL strings to reflect deterministic tie-break order
- `integ-test/src/test/resources/expectedOutput/calcite/explain_{rare,top}_usenull_{true,false}.yaml`, `explain_nested_agg_top_push.yaml` — updated 5 explain fixtures

### Issues resolved
N/A — closes the out-of-scope follow-ups documented in [opensearch-project/OpenSearch#21593](https://github.com/opensearch-project/OpenSearch/pull/21593).

### Check List
- [x] Functionality includes testing.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.


docs/user/ppl/cmd/top.md
```diff
@@ -166,10 +166,10 @@ fetched rows / total rows = 4/4
 +-----------------------+-------+
 | email                 | count |
 |-----------------------+-------|
-| null                  | 1     |
 | amberduke@pyrami.com  | 1     |
 | daleadams@boink.com   | 1     |
 | hattiebond@netagy.com | 1     |
+| null                  | 1     |
```
RyanL1997: just for my knowledge are we change the result set order?

RyanL1997 (commented):
LGTM. Just notice that the same CI problem is also happening in core with my newly created PR for dedup https://github.com/opensearch-project/OpenSearch/pull/21622.

docs/user/ppl/cmd/top.md
ahkcs: We are updating the test cases since we are supporting stable tie-break for RareTopN ROW_NUMBER in this PR

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/RareTopPushdownRule.java
```diff
@@ -49,22 +50,39 @@ protected void onMatchImpl(RelOptRuleCall call) {
...
               .toList();
       if (targetList.size() != 1) {
         return;
       }
       String targetName = targetList.getFirst();
-      digest = new RareTopDigest(targetName, byList, number, orderKey.getDirection());
+      if (orderKeys.size() == 2) {
+        RexFieldCollation tieBreakKey = orderKeys.get(1);
+        if (tieBreakKey.getDirection() != RelFieldCollation.Direction.ASCENDING) {
+          return;
```
Swiddis: suggestion: Maybe move some of these early-return checks to a helper?

This whole `orderKeys.size() == 2` block could be split out into `if (hasInvalidTieBreaks(orderKeys)) return;` which makes the intent clearer

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3069,6 +3083,15 @@ public RelNode visitRareTopN(RareTopN node, CalcitePlanContext context) {
     return context.relBuilder.peek();
   }
 
+  private static boolean isComparableOrderKey(RexNode key) {
+    SqlTypeName typeName = key.getType().getSqlTypeName();
```
Swiddis: not sure if it's applicable here but `getType()` should have a `getComparability` method or similar that does this

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/RareTopPushdownRule.java
```diff
@@ -49,22 +50,39 @@ protected void onMatchImpl(RelOptRuleCall call) {
...
+        }
+        List<Integer> tieBreakIndices = PlanUtils.getSelectColumns(List.of(tieBreakKey.getKey()));
+        List<String> tieBreakList = tieBreakIndices.stream().map(fieldNameList::get).toList();
+        if (tieBreakList.size() != 1 || !tieBreakList.getFirst().equals(targetName)) {
+          return;
+        }
+      }
+      digest = new RareTopDigest(targetName, byList, number, primaryOrderKey.getDirection());
```
Swiddis, line 77: q: how does this actually apply the secondary sorting? I don't see anything taking the tiebreak keys just the primaryOrderKey...

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/RareTopPushdownRule.java
ahkcs: Good call — extracted into `hasIncompatibleTieBreak(orderKeys, fieldNameList, targetName)`. The call site now reads `if (hasIncompatibleTieBreak(...)) return;` and the helper handles both the direction check and the target-field match on `orderKeys.get(1)`

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/RareTopPushdownRule.java
ahkcs: It isn't applied by us — the OpenSearch terms aggregation already tie-breaks deterministically on `_key:asc` natively. When the second order key is "rare/top target field, ASC", the OS request is exactly the same as the single-key shape: `terms { field: targetName, order: { _count: <direction> } }`, where the implicit `_key:asc` secondary order is what the new Calcite tie-break key encoded. The wire payload is unchanged from before this PR.

The validation block exists to guard against pushing down a 2-key collation that isn't this safe shape (e.g. a user-added secondary sort on some unrelated column would silently drop on the OS side). When the shape matches, we only forward the primary key into `RareTopDigest` and let OS handle the secondary deterministically.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Looked into it — `RelDataType#getComparability()` does exist and returns `RelDataTypeComparability.{NONE,UNORDERED,ALL}`. The problem is that in Calcite 1.41 the default `RelDataTypeImpl#getComparability` returns `ALL`, and none of the concrete subclasses we care about (`ArraySqlType`, `MultisetSqlType`, `MapSqlType`, `RelRecordType`) override it — verified by `javap` on the bundled jar. So that predicate would happily say `ALL` for `ARRAY<int>` etc., and not filter out the cases that actually trip `ArrayList cannot be cast to Comparable` at runtime.

I did refactor the predicate to use Calcite's structural helpers though — `SqlTypeUtil.isCollection` (covers ARRAY + MULTISET) + `RelDataType#isStruct` (ROW + STRUCTURED) + an explicit MAP check, in place of the verbatim `SqlTypeName` list. Comment in the code now flags why `getComparability` isn't load-bearing here. (commit 91b20950b)

Swiddis (approved):

mengweieric (approved):

PR #5435, opened 2026-05-12
Add tiebreaker to stats sort-on-measure IT queries

ahkcs:
## Summary

Six tests in `StatsCommandIT` assert a specific top-K cutoff for queries like `| stats … by X | sort - count() | head 5`. When `count()` has ties at the cutoff, PPL spec doesn't define a tiebreaker, so different execution engines (legacy DSL terms-agg, Calcite enumerable sort, analytics-engine hash sort) return different tied rows. The strict `containsInAnyOrder` assertion only happened to match whichever member the original engine picked.

Address by adding the group-by column(s) as an explicit secondary sort key. The query is now deterministic regardless of engine, and the existing strict assertion can stay. Per @penghuo's review feedback on an earlier revision, this fixes the root cause rather than masking it with a tie-tolerant matcher.

## Tests changed (in `StatsCommandIT`)

| Test | Query change |
|---|---|
| `testStatsSortOnMeasure` Q1 + Q2 | `sort - count(), state \| head 5` / `sort count(), state \| head 5` |
| `testStatsSortOnMeasureWithScript` Q1 + Q2 | `sort - count(), new_state \| head 5` / `sort count(), new_state \| head 5` |
| `testStatsSpanSortOnMeasure` Q1 + Q2 | `sort - cnt, \`span(birthdate,1month)\` \| head 5` / `sort cnt, \`span(birthdate,1month)\` \| head 5` |
| `testStatsSpanSortOnMeasureMultiTerms` Q1 + Q2 | `sort - count(), gender, state \| head 5` / `sort count(), gender, state \| head 5` |
| `testStatsSpanSortOnMeasureMultiTermsWithScript` Q1 + Q2 | `sort - count(), new_gender, new_state \| head 5` / `sort count(), new_gender, new_state \| head 5` |
| `testStatsSortOnMeasureComplex` Q1 + Q2 | `sort - c, state \| head 5` / `sort - d, gender, new_state \| head 5` |

Where the original test had separate pushdown / no-pushdown / non-Calcite branches asserting different tied alternatives, all branches collapse into a single expected ordering once the tiebreaker is added.

## Verification

`CalciteStatsCommandIT` (default Calcite-DSL path) against the local cluster:

```
  -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9300 -Dtests.clustername=runTask \
  --tests "org.opensearch.sql.calcite.remote.CalciteStatsCommandIT"
→ 63 / 63 pass
```

## Test plan

- [x] `./gradlew :integ-test:compileTestJava` clean.
- [x] Spotless clean.
- [x] `CalciteStatsCommandIT` against a default-routed (non-analytics) cluster: 63 / 63 pass.
- [ ] CI on this PR.

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/StatsCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/StatsCommandIT.java
@@ -811,37 +811,26 @@ public class StatsCommandIT extends PPLIntegTestCase {
   public void testStatsSortOnMeasure() throws IOException {
     try {
       setQueryBucketSize(5);
+      // count=25 is tied across multiple states (AL/ME/TN/WY); add `state` as an explicit
+      // tiebreaker so the top-K cutoff is deterministic. PPL spec doesn't define a
+      // tiebreaker for tied measure values, so without this the result varies by engine.
       JSONObject response =
           executeQuery(
               String.format(
-                  "source=%s | stats bucket_nullable=false count() by state | sort - `count()` |"
-                      + " head 5",
+                  "source=%s | stats bucket_nullable=false count() by state | sort - `count()`,"
+                      + " state | head 5",
                   TEST_INDEX_ACCOUNT));
       verifyDataRows(
-          response, rows(30, "TX"), rows(28, "MD"), rows(27, "ID"), rows(25, "ME"), rows(25, "AL"));
+          response, rows(30, "TX"), rows(28, "MD"), rows(27, "ID"), rows(25, "AL"), rows(25, "ME"));
+      // Ascending: count=13 (NV/SC) and count=14 (many states); state asc as tiebreaker.
       response =
           executeQuery(
               String.format(
-                  "source=%s | stats bucket_nullable=false count() by state | sort `count()` | head"
-                      + " 5",
+                  "source=%s | stats bucket_nullable=false count() by state | sort `count()`,"
+                      + " state | head 5",
                   TEST_INDEX_ACCOUNT));
-      if (!isPushdownDisabled()) {
-        verifyDataRows(
-            response,
-            rows(13, "NV"),
-            rows(13, "SC"),
-            rows(14, "CO"),
-            rows(14, "AZ"),
-            rows(14, "DE"));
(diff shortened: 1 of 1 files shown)


integ-test/src/test/java/org/opensearch/sql/ppl/StatsCommandIT.java
```diff
@@ -817,31 +818,33 @@ public void testStatsSortOnMeasure() throws IOException {
                   "source=%s | stats bucket_nullable=false count() by state | sort - `count()` |"
                       + " head 5",
                   TEST_INDEX_ACCOUNT));
-      verifyDataRows(
-          response, rows(30, "TX"), rows(28, "MD"), rows(27, "ID"), rows(25, "ME"), rows(25, "AL"));
+      // count=25 is tied across multiple states (AL/ME/TN/WY); accept any tied pair at the
+      // cutoff — legacy DSL terms-agg, Calcite enumerable sort, and analytics-engine each
+      // break the tie deterministically but with different orderings.
+      verifyDataRowsAllowingTiesAt(
```
penghuo: Why not just modify the query and add another sort column as a tiebreaker? PPL spec does not define tiebreaker.

When the IT was first written, the author likely didn't realize the PPL query produces non-deterministic results when measure values tie. Adding a deterministic tiebreaker (e.g. sort - count(), state) fixes the root cause

ahkcs: Good call — applied. Rewrote the PR to drop the `verifyDataRowsAllowingTiesAt` matcher and instead added an explicit tiebreaker (the group-by column) to each of the six queries' `sort` clauses, restoring the strict `verifyDataRows` assertions with the deterministic expected rows.

For the queries that had separate pushdown / no-pushdown / non-Calcite branches asserting different tied alternatives, all branches now collapse to a single expected ordering. `CalciteStatsCommandIT` against the default Calcite-DSL path: 63 / 63 pass.

penghuo (approved):

dai-chen (approved):

Issue #5420, opened 2026-05-07
Datetime output cast format diverges between Calcite and DataFusion engines


ahkcs:
### Description

The `DatetimeOutputCastRule` introduced in #5408 wraps datetime
output columns in `CAST(... AS VARCHAR)` so PPL responses are
rendered as ISO strings. The string produced by that cast is
engine-dependent:

- **PPL Calcite path** (`UnifiedQueryCompiler`) — emits ANSI SQL
  format `2024-01-15 12:00:00` (space separator, no `T`),
  consistent with SparkSQL, PostgreSQL, MySQL, Oracle, and
  SQL Server.
- **Analytics-engine path** (DataFusion native runtime) — emits
  ISO 8601 format `2024-01-15T12:00:00` (with `T` separator).

This means the same PPL query against the same data returns
two different string formats depending on which execution
engine handled it, breaking the wire-format contract that
#5408 set out to enforce.

### Reproduction

```
source=events | fields created_at
```

- Calcite engine → `"2024-01-15 12:00:00"`
- DataFusion engine → `"2024-01-15T12:00:00"`

### Context

Called out as Note 2 of #5408. Filed as a follow-up so the
divergence is tracked outside that already-merged PR.

### Possible directions

1. Replace the implicit `CAST(... AS VARCHAR)` with an explicit
   `DATE_FORMAT(..., '<pattern>')` (or `to_char` /
   `format_datetime`) that pins the wire format on both engines.
2. Push the formatting into the response formatter rather than
   the logical plan, so engine-side cast semantics never leak
   to the wire.
3. Add a DataFusion-side cast adapter that emits the ANSI SQL
   string format to match Calcite.

### Related

- #5408 — introduced `DatetimeOutputCastRule`
- #5250 — parent issue tracking unified-query datetime work

ahkcs:
Clarification on scope — this applies to **TIMESTAMP** (and likely `TIMESTAMP_WITH_LOCAL_TIME_ZONE`), not pure DATE / TIME:

| Calcite type | Calcite (ANSI SQL) cast string | DataFusion (ISO 8601) cast string | Diverges? |
| --- | --- | --- | --- |
| DATE | `2024-01-15` | `2024-01-15` | no |
| TIME | `12:00:00` | `12:00:00` | no |
| TIMESTAMP | `2024-01-15 12:00:00` | `2024-01-15T12:00:00` | **yes** (separator) |
[...]

PR #5436, opened 2026-05-12
Add tests.analytics.parquet_indices toggle for analytics-engine IT coverage

ahkcs:
## Summary

Adds a small, opt-in test infrastructure slice so the PPL integration test suite can run end-to-end against the analytics-engine backend without per-test rewiring. Default behavior is unchanged: with the flag unset, every IT runs through the existing V2 / Calcite path exactly as before.

The single knob is `-Dtests.analytics.parquet_indices=true`. When set, `TestUtils.createIndexByRestClient` backs every test-created index with single-shard composite/parquet storage. `RestUnifiedQueryAction.isAnalyticsIndex` (post-[#5432](https://github.com/opensearch-project/sql/pull/5432)) reads those settings and routes the query to the analytics-engine planner (DataFusion). **No additional cluster setting, REST short-circuit, or routing override** — production routing is the single source of truth.

## Commits

**1. `Add tests.analytics.parquet_indices test toggle`** — the infra slice (3 files, +68 LOC):

| File | Change |
|---|---|
| `integ-test/.../TestUtils.java` | `ANALYTICS_PARQUET_INDICES_PROP` constant; `createIndexByRestClient` injects composite/parquet settings when the flag is on; `loadDataByRestClient` uses `refresh=true` to work around `LuceneCommitter.getSafeCommitInfo` stub |
| `integ-test/.../PPLIntegTestCase.java` | `isAnalyticsParquetIndicesEnabled()` predicate for per-test branching on engine semantics |
| `integ-test/build.gradle` | Forwards `-Dtests.analytics.parquet_indices` to `:integTestRemote` |

**2. `Branch 7 stats tests on isAnalyticsParquetIndicesEnabled`** — demonstrates the helper (1 file, +40/-8 LOC):

### SQL-spec semantics (DataFusion follows the spec; legacy DSL returns 0)

| Test | What changed |
|---|---|
| `testSumWithNull` | `SUM` of all-null returns `null` on analytics; legacy DSL pushdown returns `0`. OR analytics into the existing `isPushdownDisabled()` branch |
| `testStatsPercentileByNullValue` + `NonNullBucket` | `percentile` of empty/all-null group returns `null` on analytics. Same OR pattern |

### Non-deterministic head ordering

| Test | What changed |
|---|---|
| `testStatsWithLimit` Q1 + Q2 | `head 5` over a 6-bucket result picks `(null,36)` instead of `(null,null)` on analytics (DataFusion hash-bucket order differs from Calcite enumerable order). Size-only assertion via the existing pushdown branch |

### TDigest interpolation divergence

| Test | What changed |
|---|---|
| `testStatsPercentileWithNull` | DataFusion's TDigest interpolates p50 to `35413` vs OpenSearch's `39225`; both are valid approximations within TDigest's compression-bound error. Per-engine expected value |
| `testStatsPercentileBySpan` | Same TDigest diff on the age=30 bucket (`33194` vs `39225`). Coordinated with [opensearch-project/OpenSearch#21584](https://github.com/opensearch-project/OpenSearch/pull/21584) commit 5 which fixes the upstream planner-side type mismatch so the query reaches the data assertion at all |

### Semantic divergence pending team decision

| Test | What changed |
|---|---|
| `testDisableLegacyPreferred` | Under `PPL_SYNTAX_LEGACY_PREFERRED=false`, V2 / Calcite-DSL drops the null-age bucket (5 rows) while DataFusion keeps it (6 rows). Skipped on the analytics path via `Assume.assumeFalse` until the team decides which behavior is the intended new-syntax semantics. |

## Why no force-routing cluster setting

An earlier revision of this PR added a `plugins.calcite.analytics.force_routing` cluster setting plus a short-circuit in `RestUnifiedQueryAction.isAnalyticsIndex`. With [#5432](https://github.com/opensearch-project/sql/pull/5432) merged (routing now keys on `index.pluggable.dataformat.enabled` + `pluggable.dataformat=composite`), that override is redundant:

1. `parquet_indices=true` makes every test-created index carry the right settings.
2. `CalciteStatsCommandIT.init()` calls `enableCalcite()`, so the Calcite path is active.
3. `isAnalyticsIndex()` returns `true` via the index-setting check → query routes to analytics-engine.

Verified end-to-end: removing the force-routing cluster setting / IT toggle / REST short-circuit gives the **exact same** pass-rate as the earlier revision.

## Verification

Against an analytics-engine cluster (opensearch-project/OpenSearch#21584 with commit 5 + main rebase):

```bash
  -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9300 -Dtests.clustername=runTask \
  -Dtests.analytics.parquet_indices=true \
  --tests "org.opensearch.sql.calcite.remote.CalciteStatsCommandIT"
```

### `CalciteStatsCommandIT` pass-rate

| Test | Before this PR | After this PR |
|---|---|---|
| `testStatsWithLimit` | FAIL | **PASS** |
| `testSumWithNull` | FAIL | **PASS** |
| `testStatsPercentileByNullValue` | FAIL | **PASS** |
| `testStatsPercentileByNullValueNonNullBucket` | FAIL | **PASS** |
| `testStatsPercentileWithNull` | FAIL (TDigest diff) | **PASS** (per-engine value) |
| `testStatsPercentileBySpan` | FAIL (500 / split rule + TDigest diff) | **PASS** (per-engine value, paired with #21584 commit 5) |
| `testDisableLegacyPreferred` | FAIL | **SKIPPED** (semantic divergence pending team decision) |
| `testStatsBySpanTimeWithNullBucket` | FAIL (multi-unit span) | FAIL (intentionally — out of scope for this PR) |
| **Overall** | **46 / 63 pass, 17 fail** | **52 / 63 pass, 10 fail, 1 skipped** |

### Default behavior unchanged

With the flag unset, the cluster's persistent settings are untouched, indices stay Lucene-backed, and every IT runs through the existing V2 / Calcite path. CI that doesn't opt in sees zero behavior change.

## Remaining 10 failures on `CalciteStatsCommandIT` (analytics path)

| Bucket | Count | Notes |
|---|---|---|
| Sort-on-measure ties | 6 | Fixed by [opensearch-project/sql#5435](https://github.com/opensearch-project/sql/pull/5435) (tie-tolerant matcher). When both PRs merge: combined **58 / 63 + 1 skipped + 4 fail**. |
| `testStatsTimeSpan` datetime output-cast format | 1 | DataFusion's `CAST(timestamp AS VARCHAR)` emits ISO 8601 with `T` separator while PPL's Calcite path emits space-separated ANSI SQL format. Tracked at [opensearch-project/sql#5420](https://github.com/opensearch-project/sql/issues/5420). |
| `testStatsBySpanTimeWithNullBucket` multi-unit time span | 1 | `span(@timestamp, 12h)` falls through `SpanAdapter` (only `interval=1` is rewritten to `DATE_TRUNC`); needs DataFusion's `date_bin`. Left failing on the analytics path so the gap is visible. |
| `LogicalJoin` planner gap | 2 | Out of scope — analytics-engine planner doesn't support joins. |

## Test plan

- [x] `./gradlew :integ-test:compileTestJava` clean.
- [x] Spotless clean.
- [x] `CalciteStatsCommandIT` against the analytics-engine cluster with only `-Dtests.analytics.parquet_indices=true`: 46 → 52 passes + 1 documented skip.
- [ ] CI on this PR — default branches untouched, no regression expected on V2 / Calcite paths.

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/legacy/TestUtils.java
+++ b/integ-test/src/test/java/org/opensearch/sql/legacy/TestUtils.java
@@ -38,6 +38,18 @@ public class TestUtils {
   private static final String MAPPING_FILE_PATH = "src/test/resources/indexDefinitions/";
 
   /**
+   * System property that makes every test-created index parquet-backed (composite primary data
+   * format = parquet) with a single shard. When set, {@link
+   * RestUnifiedQueryAction#isAnalyticsIndex} (which routes based on {@code
+   * index.pluggable.dataformat.enabled} / {@code index.pluggable.dataformat=composite}, see #5432)
+   * will return {@code true} for every test-created index — exercising the analytics-engine route
+   * end-to-end without per-test rewiring.
+   *
+   * <p>Off by default; normal CI is untouched.
+   */
+  public static final String ANALYTICS_PARQUET_INDICES_PROP = "tests.analytics.parquet_indices";
+
+  /**
    * Create test index by REST client.
    *
    * @param client client connection
@@ -48,6 +60,9 @@ public class TestUtils {
     Request request = new Request("PUT", "/" + indexName);
     JSONObject jsonObject = isNullOrEmpty(mapping) ? new JSONObject() : new JSONObject(mapping);
     setZeroReplicas(jsonObject);
+    if (Boolean.parseBoolean(System.getProperty(ANALYTICS_PARQUET_INDICES_PROP, "false"))) {
+      makeParquetBacked(jsonObject);
+    }
     request.setJsonEntity(jsonObject.toString());
     performRequest(client, request);
   }
@@ -65,6 +80,25 @@ public class TestUtils {
     JSONObject indexSettings =
         settings.has("index") ? settings.getJSONObject("index") : new JSONObject();
     indexSettings.put("number_of_replicas", 0);
+    settings.put("index", indexSettings);
+    jsonObject.put("settings", settings);
+  }
+
+  /**
+   * Switches the test index to a parquet-backed composite store with a single shard so the
+   * analytics-engine path has a backend that can scan it. Routing is then driven entirely by index
+   * settings (#5432) — no other test plumbing required.
+   */
+  private static void makeParquetBacked(JSONObject jsonObject) {
+    JSONObject settings =
+        jsonObject.has("settings") ? jsonObject.getJSONObject("settings") : new JSONObject();
+    JSONObject indexSettings =
+        settings.has("index") ? settings.getJSONObject("index") : new JSONObject();
+    indexSettings.put("number_of_shards", 1);
+    indexSettings.put("pluggable.dataformat.enabled", true);
+    indexSettings.put("pluggable.dataformat", "composite");
+    indexSettings.put("composite.primary_data_format", "parquet");
+    indexSettings.put("composite.secondary_data_formats", new org.json.JSONArray());
     settings.put("index", indexSettings);
     jsonObject.put("settings", settings);
   }
@@ -116,8 +150,17 @@ public class TestUtils {
   public static void loadDataByRestClient(
       RestClient client, String indexName, String dataSetFilePath) throws IOException {
     Path path = Paths.get(getResourceFilePath(dataSetFilePath));
-    Request request =
-        new Request("POST", "/" + indexName + "/_bulk?refresh=wait_for&wait_for_active_shards=all");
+    // Workaround: parquet-backed indices in the analytics-backend-lucene composite engine
+    // do not yet implement LuceneCommitter.getSafeCommitInfo (UnsupportedOperationException
+    // "TODO:: with index deleter"), which hangs refresh=wait_for until the test framework
+    // request timeout (~60s). Force-refresh sidesteps the safe-commit-info path while still
+    // making the bulk-loaded docs immediately searchable. Drop this branch once
+    // LuceneCommitter.getSafeCommitInfo is implemented.
+    String refreshParam =
+        Boolean.parseBoolean(System.getProperty(ANALYTICS_PARQUET_INDICES_PROP, "false"))
+            ? "refresh=true"
+            : "refresh=wait_for&wait_for_active_shards=all";
+    Request request = new Request("POST", "/" + indexName + "/_bulk?" + refreshParam);
     request.setJsonEntity(new String(Files.readAllBytes(path)));
     performRequest(client, request);
   }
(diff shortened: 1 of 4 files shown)


integ-test/src/test/java/org/opensearch/sql/legacy/TestUtils.java
```diff
@@ -69,6 +84,25 @@ private static void setZeroReplicas(JSONObject jsonObject) {
     jsonObject.put("settings", settings);
   }
 
+  /**
+   * Switches the test index to a parquet-backed composite store with a single shard so the
+   * analytics-engine path has a backend that can scan it. Routing is then driven entirely by index
+   * settings (#5432) — no other test plumbing required.
+   */
+  private static void makeParquetBacked(JSONObject jsonObject) {
+    JSONObject settings =
```
dai-chen: np: make both these index settings and refresh config into a test index config in single place. We have this problem in os-spark repo which has aoss conditional check and configs in many places.

integ-test/src/test/java/org/opensearch/sql/ppl/StatsCommandIT.java
```diff
@@ -508,7 +512,9 @@ public void testSumWithNull() throws IOException {
     // TODO: Fix -- temporary workaround for the pushdown issue:
     //  The current pushdown implementation will return 0 for sum when getting null values as input.
     //  Returning null should be the expected behavior.
-    Integer expectedValue = isPushdownDisabled() ? null : 0;
+    // The analytics-engine backend (DataFusion) follows the SQL spec like Calcite-no-pushdown —
+    // SUM of all-null is null, not 0.
+    Integer expectedValue = (isPushdownDisabled() || isAnalyticsParquetIndicesEnabled()) ? null : 0;
```
dai-chen, line 517: Just thinking does all these assertion changes means semantic/behavior change? Or we can simply fix the assertion in this way.

integ-test/src/test/java/org/opensearch/sql/legacy/TestUtils.java
ahkcs: Done — consolidated into `TestUtils.AnalyticsIndexConfig` (b1c8502).

integ-test/src/test/java/org/opensearch/sql/ppl/StatsCommandIT.java
ahkcs: Depends on the test:

- **Existing pattern (`isPushdownDisabled() ? null : 0`)** — extend it (`testSumWithNull`, `testStatsWithLimit`).
- **Semantic divergence** — `Assume.assumeFalse` (`testDisableLegacyPreferred`).
- **Percentile tests (`testStatsPercentileBySpan` and friends)** — since #21584 no longer wires `percentile_approx` on the analytics path, those queries fail at execution before the assertion. Reverted those branches in efa97de — they were doing nothing.

penghuo (approved):

dai-chen (approved):

PR #5439, opened 2026-05-14
[PPL][analytics-engine-compatibility] Lower isempty/isblank to CHAR_LENGTH = 0 instead of multiset IS_EMPTY

songkant-aws:
## Description

PPL's `isempty(x)` / `isblank(x)` were lowered in `PPLFuncImpTable` using `SqlStdOperatorTable.IS_EMPTY` against string operands. That operator is the SQL:2003 multiset emptiness predicate — its `OperandTypeChecker` is `OperandTypes.COLLECTION_OR_MAP` and its enumerable-runtime binding calls `java.util.Collection.isEmpty()` reflectively.

Passing a string slipped through only because:
1. `RexBuilder.makeCall` bypasses the operand checker.
2. Calcite's Linq4j codegen emits `target.isEmpty()` (a bare method-name call), which Janino then resolves against the actual operand type. `String.isEmpty()` happens to exist with the same signature, so the script-path runtime works by coincidence.

This shape is fragile:
- **Outside the script-pushdown path** — most importantly Substrait emitters such as the analytics-engine route — there is no isthmus mapping for `SqlStdOperatorTable.IS_EMPTY`, and the plan is rejected with `"unresolved function reference IS EMPTY"`.
- The `PredicateAnalyzer` had to special-case-block DSL pushdown for the `OR(IS_NULL, IS_EMPTY)` shape because OpenSearch's non-short-circuiting `bool.should` would NPE evaluating `name.isEmpty()` on a null operand. So the entire OR was wrapped in a single script_query — a strict pessimization compared to a mixed bool.should.

## Issues Resolved

This PR replaces the lowering with the predicate that actually matches the documented semantics:

```
isempty(x) -> OR(IS_NULL(x), CHAR_LENGTH(x) = 0)
isblank(x) -> OR(IS_NULL(x), CHAR_LENGTH(TRIM(BOTH ' ' FROM x)) = 0)
```

`CHAR_LENGTH` is a standard string operator with explicit type semantics, backed by `SqlFunctions.charLength(String)` in Calcite's enumerable runtime, and round-trips through Substrait's default extension catalog. Same observable behavior on existing PPL paths; correct behavior on Substrait-based paths.

The `containIsEmptyFunction` guard in `PredicateAnalyzer.andOr` is removed: `CHAR_LENGTH(null)` follows Calcite's `NullPolicy.STRICT` and returns null rather than NPE, so DSL's non-short-circuiting `should` evaluation is safe even when the field is null. This also unlocks better pushdown — the `IS_NULL` arm is now a native `bool.must_not.exists` clause and only the `CHAR_LENGTH=0` arm remains a script. Null docs no longer touch the script engine.

The `isNullOr_ScriptPushDown` test was renamed to `isEmpty_pushesIsNullArmAsExistsAndCharLengthArmAsScript` and rewritten to assert the actual structural shape of the pushdown (2-arm bool.should with the expected types per arm) instead of just substring-matching the script lang marker.

Plan changes are reflected in the regenerated explain golden files for both `calcite/` and `calcite_no_pushdown/` variants.

### Check List


qianheng-aws (approved):

yuancu (approved):
LGTM

A minor appeal 👀: the same problem (and many other type problems) are also fixed in #4892 with a similar approach

- `isempty(x)` -> `OR(IS_NULL(x), x = '')`
- `isblank(x)` -> `OR(IS_NULL(x), TRIM(BOTH ' ' FROM x) ='')`

PR #5441, opened 2026-05-14
spath: parquet-backed test indices for analytics-engine route

ahkcs:
Pairs with [opensearch-project/OpenSearch#21664](https://github.com/opensearch-project/OpenSearch/pull/21664). Both PRs are required to move `CalcitePPLSpathCommandIT` off 0 / 16 on the analytics-engine route.

## What the change does

`CalcitePPLSpathCommandIT.init()` was creating its four test indices by raw `PUT /<idx>/_doc/N` requests, which auto-creates the index via the default Lucene path. The analytics-engine compatibility run (`-Dtests.analytics.parquet_indices=true`) only injects parquet/composite settings **inside** `TestUtils.createIndexByRestClient`, so the raw-PUT indices were Lucene-only — and DataFusion fails with `UnsupportedOperationException: acquireReader is not supported in EngineBackedIndexer` on any Lucene-only index.

The fix is one line per index: create the empty index up-front through the helper so the parquet toggle gets a chance to inject its settings, then let the existing doc PUTs populate it via dynamic mapping.

```
Before: PUT /test_spath/_doc/1 {…}                     → auto-creates Lucene-backed index
After:  PUT /test_spath {}     (via helper, no mapping) → empty index inherits parquet settings
        PUT /test_spath/_doc/1 {…}                     → doc lands in parquet-backed index
```

No mapping is declared (`null` mapping argument) — DataFusion handles dynamic mapping on parquet-backed composite indices just fine. Same pattern as `CalciteEvalCommandIT` and `CalciteFieldFormatCommandIT`. No change for the v2 / Calcite path; the helper is a no-op when the parquet toggle isn't set.

## Pass rate

| IT | Route | Before | After |
|---|---|---|---|
| `CalcitePPLSpathCommandIT` | analytics-engine (`-Dtests.analytics.force_routing=true -Dtests.analytics.parquet_indices=true`) | 0 / 16 | **16 / 16** |
| `CalcitePPLSpathCommandIT` | default v2 / Calcite (no flags) | 16 / 16 | 16 / 16 (no regression) |

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLSpathCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLSpathCommandIT.java
@@ -15,9 +15,17 @@ import java.io.IOException;
 import org.json.JSONObject;
 import org.junit.jupiter.api.Test;
 import org.opensearch.client.Request;
+import org.opensearch.sql.legacy.TestUtils;
 import org.opensearch.sql.ppl.PPLIntegTestCase;
 
 public class CalcitePPLSpathCommandIT extends PPLIntegTestCase {
+  // Pre-create each test index through TestUtils.createIndexByRestClient so the
+  // analytics-engine compatibility run (tests.analytics.parquet_indices=true)
+  // provisions them as parquet-backed composite stores. Raw `PUT /<idx>/_doc/N`
+  // bypasses the helper, yielding a Lucene-only index that DataFusion cannot
+  // acquireReader on (`UnsupportedOperationException: acquireReader is not
+  // supported in EngineBackedIndexer`). Mapping passed as null — dynamic
+  // mapping infers the doc fields from the subsequent PUTs.
   @Override
   public void init() throws Exception {
     super.init();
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/opensearch/sql/data/model/ExprValueUtils.java
```diff
@@ -111,6 +111,17 @@ public static ExprValue nullValue() {
...
+  private static final String ARROW_TEXT_CLASS_NAME = "org.apache.arrow.vector.util.Text";
+
+  /**
+   * Whether {@code o} is an Arrow {@code Text} (the UTF-8 byte-buffer wrapper that arrow's Map /
+   * Struct / List vectors emit for string values). FQN match keeps {@code core/} free of an Arrow
+   * dependency.
+   */
+  private static boolean isArrowText(Object o) {
+    return o != null && ARROW_TEXT_CLASS_NAME.equals(o.getClass().getName());
+  }
```
penghuo: ExprValueUtils should do not know ARROW data type. Why ExprValueUtils been used on execution code path?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLSpathCommandIT.java
```diff
@@ -15,9 +15,29 @@
...
+      "{\"mappings\":{\"properties\":{\"doc\":{\"type\":\"keyword\"}}}}";
+
+  private static final String AUTO_DOC_MAPPING =
+      "{\"mappings\":{\"properties\":{"
+          + "\"nested_doc\":{\"type\":\"keyword\"},"
+          + "\"array_doc\":{\"type\":\"keyword\"},"
+          + "\"merge_doc\":{\"type\":\"keyword\"},"
+          + "\"stringify_doc\":{\"type\":\"keyword\"},"
+          + "\"empty_doc\":{\"type\":\"keyword\"},"
+          + "\"malformed_doc\":{\"type\":\"keyword\"}}}}";
```
penghuo: Why add auto_doc_mapping? Becuae  Analytics Eengine does not support dynamic mapping?

core/src/main/java/org/opensearch/sql/data/model/ExprValueUtils.java
ahkcs: Good catch, updated to remove the change

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLSpathCommandIT.java
ahkcs: Updated to remove explicit mapping,  currently our IT creates the index using the lazy way, which makes it a default lucene-backed index, the change is to create the empty index up-front through the helper so the parquet toggle gets a chance to inject its settings

penghuo (approved):

dai-chen (approved):
Thanks for the changes!

PR #5443, opened 2026-05-14
Added coercion rules and placeholder UDF to handle VARBINARY

vinaykpud:
### Description
Adds coercion support so PPL queries can compare a `VARBINARY` column with a `VARCHAR` literal.

eg: `source=logs | where client_ip = '192.168.1.1'`

For `ip` and `binary` fields types [OpenSearchSchemaBuilder.java‎](https://github.com/opensearch-project/OpenSearch/pull/21631/changes#diff-ff824465bf640034c87540d3ce045e50311d308fc1fa4baf15fba37968dc431bR119) maps to `SqlTypeName.VARBINARY`. 

 The change introduces a `BINARY(varchar)` placeholder UDF and threads VARBINARY through the type/coercion pipeline so that:

1. A `VARCHAR` literal compared to a `VARBINARY` column is recognized as coercible to `BINARY`.
2. `CoercionUtils.cast` triggers `ExtendedRexBuilder.makeCast`, which emits `BINARY(varchar)` instead of a generic Calcite CAST.  the on-disk encoding of `ip` / `binary` fields.
3. The placeholder `BINARY(varchar)` `RexCall` is intended to be rewritten by a backend adapter (OpenSearch repo, separate PR) into a `VARBINARY` `RexLiteral` whose bytes match the on-disk format.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
@@ -414,8 +414,32 @@ public class OpenSearchTypeFactory extends JavaTypeFactoryImpl {
         }
         return first;
       }
+      // When the list has a VARBINARY column plus VARCHAR literals, treat VARBINARY
+      // as the common type so IN / BETWEEN can insert casts.
+      RelDataType varbinaryResult = leastRestrictiveVarbinaryVarchar(types);
+      if (varbinaryResult != null) {
+        return varbinaryResult;
+      }
     }
     return super.leastRestrictive(types);
+  }
+
+  private @Nullable RelDataType leastRestrictiveVarbinaryVarchar(List<RelDataType> types) {
+    boolean hasVarbinary = false;
+    boolean anyNullable = false;
+    for (RelDataType t : types) {
+      SqlTypeName name = t.getSqlTypeName();
+      if (name == SqlTypeName.VARBINARY) {
+        hasVarbinary = true;
+      } else if (name != SqlTypeName.VARCHAR && name != SqlTypeName.CHAR) {
+        return null;
+      }
+      anyNullable |= t.isNullable();
+    }
+    if (!hasVarbinary) {
+      return null;
+    }
+    return createTypeWithNullability(createSqlType(SqlTypeName.VARBINARY), anyNullable);
   }
 
   /**
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
@@ -174,6 +174,8 @@ public class OpenSearchTypeFactory extends JavaTypeFactoryImpl {
           return TYPE_FACTORY.createUDT(ExprUDT.EXPR_TIME, nullable);
         case TIMESTAMP:
           return TYPE_FACTORY.createUDT(ExprUDT.EXPR_TIMESTAMP, nullable);
+        case BINARY:
+          return TYPE_FACTORY.createSqlType(SqlTypeName.VARBINARY, nullable);
         case ARRAY:
           return TYPE_FACTORY.createArrayType(
               TYPE_FACTORY.createSqlType(SqlTypeName.ANY, nullable), -1);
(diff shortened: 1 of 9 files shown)


penghuo (commented):
Code looks good. Please failed Fix IT/UT.

dai-chen (commented):
The CI is failing due to spotless check. Please fix it. Thanks!

ahkcs (dismissed):

core/src/main/java/org/opensearch/sql/calcite/utils/OpenSearchTypeFactory.java
```diff
@@ -411,10 +414,34 @@ public static boolean isNumericType(RelDataType fieldType) {
...
+      // as the common type so IN / BETWEEN can insert casts.
+      RelDataType varbinaryResult = leastRestrictiveVarbinaryVarchar(types);
+      if (varbinaryResult != null) {
+        return varbinaryResult;
+      }
     }
     return super.leastRestrictive(types);
   }
 
+  private @Nullable RelDataType leastRestrictiveVarbinaryVarchar(List<RelDataType> types) {
```
penghuo, line 427: Add UT to cover this.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -908,6 +912,29 @@ void populate() {
...
+                  byte[][] range = parseCidrToIpv6Range(lit.getValueAs(String.class));
+                  RelDataType varbinary =
+                      builder.getTypeFactory().createSqlType(SqlTypeName.VARBINARY);
+                  RexNode low = builder.makeLiteral(new ByteString(range[0]), varbinary, false);
+                  RexNode high = builder.makeLiteral(new ByteString(range[1]), varbinary, false);
+                  // makeCall(AND, ...) auto-flattens at construction, so no Filter.isFlat issue.
+                  return builder.makeCall(
+                      SqlStdOperatorTable.AND,
+                      builder.makeCall(SqlStdOperatorTable.GREATER_THAN_OR_EQUAL, col, low),
+                      builder.makeCall(SqlStdOperatorTable.LESS_THAN_OR_EQUAL, col, high));
```
penghuo, line 933: Why rewrite the function in RexNode?  It should be delayed in optimization stage?

vinaykpud: Since CIDR is specific to PPL, we rewrite it using standard SQL comparison operators (<= and >=) during preprocessing. This ensures the optimizer and planner only need to understand standard Calcite, without needing special handling for PPL-specific functions.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -908,6 +912,29 @@ void populate() {
...
+                }
+                return builder.makeCall(PPLBuiltinOperators.CIDRMATCH, col, cidr);
+              },
+          PPLTypeChecker.family(SqlTypeFamily.BINARY, SqlTypeFamily.STRING));
```
penghuo, line 937: The CIDR should only support IP/STRING type. not binary.
The binary UDF already cast the BINARY TYPE TO STRING/IP, right?

vinaykpud: BINARY UDF casts given string literal. ex: BINARY(VARCHAR): VARBINARY

This CIDR match function applicable only when we have query like 
```
source=ft_ip_filters | where cidrmatch(val, '192.168.0.0/16')"
```
here `val` is of type  VARBINARY from the schema
`'192.168.0.0/16'` is a string literal of type VARCHAR.

IP fields schema type is VARBINARY

vinaykpud: exising `registerOperator(CIDRMATCH, PPLBuiltinOperators.CIDRMATCH);` will continue to work. This adds a overload for (VARBINARY, VARCHAR) type.

penghuo (approved):

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
vinaykpud: For your question on 
> How we define the cidrmatch argument type in doc? https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/ip.md

No documentation change is needed — the user-facing contract stays the same: cidrmatch(ip, string_literal).

This is an implementation-level detail. In the analytics-engine schema, both IP and Binary fields are exposed as `SqlTypeName.VARBINARY`, and the PPL parser / RelBuilder consumes that schema directly. As a result, the SQL plugin sees an ip column as `VARBINARY`. That's why we added the coercion rule that casts the string literal via BINARY(VARCHAR) -> VARBINARY , analogous to how we handle TIMESTAMP today.

vinaykpud: Ideally, the analytics-engine should expose distinct UDTs for IP and Binary in the schema so PPL can distinguish them at the type level. That's a larger change and out of scope here; we can tackle it as a follow-up.

ahkcs (approved):

PR #5438, opened 2026-05-13
Route unified SQL path through V2 ANTLR parser

dai-chen:
### Description

Switch the unified SQL from Calcite's standard SQL pipeline to the V2 ANTLR parser + `CalciteRelNodeVisitor`. Existing Calcite-native SQL tests are ignored; add three new V2 test classes covering basic queries and relevance search functions. More tests will be added as we continue fixing `CalciteRelNodeVisitor` to handle the full V2 AST.

#### Planned PRs

- [ ] PR 1 (current): Enable V2 parser in unified SQL path
- [ ] PR 2: Add missing relevance functions to Calcite function table
- [ ] PR 3: Extend V2 grammar/parser for missing statements, JOIN/UNION/MINUS
- [ ] PR 4: Fix remaining `CalciteRelNodeVisitor` issues identified during V2 AST integration

### Related Issues
Part of https://github.com/opensearch-project/sql/issues/5248

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
 - [x] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5448, opened 2026-05-15
Extend V2 SQL parser with `IN`/`EXISTS` subqueries for unified query path

dai-chen:
### Description

Extend the V2 SQL ANTLR grammar and add `ExtendedAstExpressionBuilder` to support `IN` and `EXISTS` subqueries in the unified Calcite-based query path. The base `AstExpressionBuilder` continues to throw `SyntaxCheckException` for these statements, preserving the legacy engine fallback in production.

#### Notes

- Fix Alias handling in `CalciteRelNodeVisitor` here to unblock tests using SELECT with specific fields
- `EXISTS` subquery only works for PartiQL nested fields in the legacy engine, thus no IT added here

#### Planned PRs

- [x] PR 1: https://github.com/opensearch-project/sql/pull/5438
- [x] PR 2: https://github.com/opensearch-project/sql/pull/5440
- [x] PR 3a: https://github.com/opensearch-project/sql/pull/5446
- [ ] PR 3b (current): Extend V2 grammar/parser for IN and EXISTS subqueries
- [ ] PR 4: Fix remaining `CalciteRelNodeVisitor` issues identified during V2 AST integration

### Related Issues

Part of https://github.com/opensearch-project/sql/issues/5248

### Check List

### DIFF
--- a/api/src/main/java/org/opensearch/sql/api/parser/SqlV2QueryParser.java
+++ b/api/src/main/java/org/opensearch/sql/api/parser/SqlV2QueryParser.java
@@ -5,10 +5,13 @@
 
 package org.opensearch.sql.api.parser;
 
+import static org.opensearch.sql.ast.dsl.AstDSL.existsSubquery;
+import static org.opensearch.sql.ast.dsl.AstDSL.inSubquery;
 import static org.opensearch.sql.ast.dsl.AstDSL.join;
 
 import java.util.Optional;
 import org.antlr.v4.runtime.tree.ParseTree;
+import org.opensearch.sql.ast.expression.Not;
 import org.opensearch.sql.ast.expression.UnresolvedExpression;
 import org.opensearch.sql.ast.statement.Query;
 import org.opensearch.sql.ast.statement.Statement;
@@ -16,8 +19,11 @@ import org.opensearch.sql.ast.tree.Join.JoinType;
 import org.opensearch.sql.ast.tree.UnresolvedPlan;
 import org.opensearch.sql.sql.antlr.SQLSyntaxParser;
 import org.opensearch.sql.sql.antlr.parser.OpenSearchSQLParser;
+import org.opensearch.sql.sql.antlr.parser.OpenSearchSQLParser.ExistsSubqueryExpressionAtomContext;
+import org.opensearch.sql.sql.antlr.parser.OpenSearchSQLParser.InSubqueryPredicateContext;
 import org.opensearch.sql.sql.antlr.parser.OpenSearchSQLParser.JoinClauseContext;
 import org.opensearch.sql.sql.parser.AstBuilder;
+import org.opensearch.sql.sql.parser.AstExpressionBuilder;
 import org.opensearch.sql.sql.parser.AstStatementBuilder;
 
 /** SQL query parser that produces {@link UnresolvedPlan} using the V2 ANTLR grammar. */
@@ -53,6 +59,11 @@ public class SqlV2QueryParser implements UnifiedQueryParser<UnresolvedPlan> {
     }
 
     @Override
+    protected AstExpressionBuilder createExpressionBuilder() {
+      return new ExtendedAstExpressionBuilder();
+    }
+
+    @Override
     public UnresolvedPlan visitJoinClause(JoinClauseContext ctx) {
       JoinType joinType = toJoinType(ctx);
       UnresolvedPlan right = visit(ctx.relation());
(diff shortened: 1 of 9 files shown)


sql/src/main/java/org/opensearch/sql/sql/parser/AstExpressionBuilder.java
```diff
@@ -668,4 +671,17 @@ private List<UnresolvedExpression> getExtractFunctionArguments(ExtractFunctionCa
...
+  public UnresolvedExpression visitInSubqueryPredicate(InSubqueryPredicateContext ctx) {
+    throw new SyntaxCheckException(
+        "IN subquery is not supported in the V2 SQL engine. Falling back to legacy engine.");
+  }
+
+  @Override
+  public UnresolvedExpression visitExistsSubqueryExpressionAtom(
+      ExistsSubqueryExpressionAtomContext ctx) {
+    throw new SyntaxCheckException(
+        "EXISTS subquery is not supported in the V2 SQL engine. Falling back to legacy engine.");
```
Swiddis, line 685: suggestion: Change these to `NotSupportedException` or similar to make downstream handling more precise.

It's not technically a syntax error to use an unsupported operation and I think it'd be good to distinguish "the engine can't handle your query" from "your query is bad."

Might need to make a new exception type if we don't have one. If we use the design from https://github.com/opensearch-project/sql/pull/5448#discussion_r3260784412, a clearer option might be `NotEnabledException`.

Swiddis, line 685: suggestion: I would prefer we don't explicitly state "Falling back to legacy engine" in the exception since that's really the exception handler's concern

api/src/main/java/org/opensearch/sql/api/parser/SqlV2QueryParser.java
```diff
@@ -69,5 +80,27 @@ private JoinType toJoinType(JoinClauseContext ctx) {
         default -> JoinType.INNER;
       };
     }
+
+    /**
+     * Expression builder with IN/EXISTS subquery support. Accesses the enclosing AstBuilder to
+     * visit subquery plan nodes. Must be created via {@link #createExpressionBuilder()} because the
+     * enclosing {@code this} reference is not available during {@code super()} construction.
+     */
+    private class ExtendedAstExpressionBuilder extends AstExpressionBuilder {
```
Swiddis, line 89: issue: If this is only relevant to legacy we should call this `LegacyExpressionBuilder`.

We hope that eventually the primary expression builder will have more features than legacy. I also suspect without proof that if we start splitting out the builder, we'll eventually have yet another builder with a different set of features.

Swiddis, line 89: suggestion (architecture): I would prefer if we put these additional predicates behind configured flags instead of making a whole new builder for them.

This makes it easier to find where AST visit code is defined since it's all in the same place, and makes it easy to see at a glance under what conditions different features are enabled. It also means that we can easily have multiple types of builders share functionality: if we eventually need another subclass of `AstExpressionBuilder` we may end up needing to duplicate visit code too.

One way might be to have `AstExpressionBuilder` take a `enableSubqueryPredicates` flag as part of its construction (or `enableExtendedFeatures`, or `enableLegacyFeatures`...). Another could be supplying an `Engine` enum and using that.

api/src/test/java/org/opensearch/sql/api/UnifiedQueryPlannerSqlV2Test.java
```diff
@@ -142,4 +142,80 @@ public void testJoinWithFilterAndOrderBy() {
...
+  public void testInSubquery() {
+    givenQuery(
+            """
+            SELECT name FROM catalog.employees
+              WHERE age IN (SELECT age FROM catalog.departments WHERE dept_name = 'Engineering')
+            """)
+        .assertPlan(
+            """
+            LogicalProject(name=[$1])
+              LogicalFilter(condition=[IN($2, {
```
penghuo, line 156: Does this logical plan can be rewrite as Join? PPL has similar execution path?

api/src/main/java/org/opensearch/sql/api/parser/SqlV2QueryParser.java
dai-chen: To give you some context: this design is a temporary workaround to enable SQL V2 + Calcite on the Analytics Engine path with minimal impact on existing SQL. Given that, I'd like to keep the change surface as small as possible and avoid introducing a flag-based mechanism.

dai-chen: V2's expression builder is `AstExpressionBuilder` so this one is `ExtendedAstExpressionBuilder` which extends V2's?

sql/src/main/java/org/opensearch/sql/sql/parser/AstExpressionBuilder.java
dai-chen: I thought this is internal and won't be returned. Let me double check. Thanks!

api/src/main/java/org/opensearch/sql/api/parser/SqlV2QueryParser.java
Swiddis: I was going off the exceptions saying 'falling back to legacy'

sql/src/main/java/org/opensearch/sql/sql/parser/AstExpressionBuilder.java
dai-chen: yeah, it make senses but that introduces many changes for this temporary workaround. Probably we can defer this refactor?

api/src/test/java/org/opensearch/sql/api/UnifiedQueryPlannerSqlV2Test.java
dai-chen: Yes, confirmed that PPL generates similar RelNode plan: https://github.com/opensearch-project/sql/blob/main/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLInSubqueryTest.java#L38

penghuo (approved):

Swiddis (approved):

api/src/main/java/org/opensearch/sql/api/parser/SqlV2QueryParser.java
dai-chen: - Execution path 1: V2 grammar/AstBuilder + Legacy (will be migrating to path 2)
- Execution path 2: V2 extended grammar/AstBuilder + CalciteRelNodeVisitor

PR #5453, opened 2026-05-19
Fix SQL query routing to analytics engine

finnegancarroll:
### Description

SQL plugin appears to be falling back to the V2 engine instead of using the analytics engine due to the recent move to V2. It seems like when we are making our routing decision to see if we can use analytics engine we [fetch a parser ](https://github.com/opensearch-project/sql/blob/main/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java#L222)from the context such that we can parse the raw query and inspect the index (to determine if it's analytics plugin compatible).

However, this no longer returns a `SqlNode` as we are now using the `SqlV2QueryParser`. This causes a casting exception to be thrown and caught upstream, where we fallback to the regular SQL V2 engine.

Added a test which fails on main but passes with this PR.

### Related Issues
N/A

### Check List


finnegancarroll:
Fix in: https://github.com/opensearch-project/sql/pull/5456/

PR #5454, opened 2026-05-19
Remove DatetimeOutputCastRule on the analytics-engine route (#5420)

### LINKED ISSUE #5420: Datetime output cast format diverges between Calcite and DataFusion engines

mengweieric:
## Problem

On the analytics-engine route, every datetime root column is wrapped in `CAST(<DATE/TIME/TIMESTAMP> AS VARCHAR)` by `DatetimeOutputCastRule`. A matching `DatetimeOutputCastRewriter` on the engine side then translates that cast into DataFusion's `to_char` extension. Whenever the rewriter's format string and the PPL formatter disagree (e.g. trailing `Z`, `T` separator), users see wire-format divergence — issue #5420.

## Solution

Drop the cast rule and let the engine return real datetime cells. The PPL response pipeline already handles datetime → string conversion natively at the formatter layer:

- `AnalyticsExecutionEngine.convertRows` feeds engine cells (`LocalDateTime` / `LocalDate` / `LocalTime`) through `ExprValueUtils.fromObjectValue`, which produces `ExprTimestampValue` / `ExprDateValue` / `ExprTimeValue`.
- Their `value()` methods render the documented PPL space-separated format (`yyyy-MM-dd HH:mm:ss[.SSSSSSSSS]` etc.).

The companion change in [opensearch-project/OpenSearch#21748](https://github.com/opensearch-project/OpenSearch/pull/21748) removes the matching `DatetimeOutputCastRewriter` and its callsites.

## Changes

- Delete `DatetimeOutputCastRule`.
- Drop the unused `UdtMapping.isDatetimeType` helper.
- Rewrite the four cast-shape assertions in `DatetimeExtensionTest` to assert the post-removal RelNode (no `CAST(... VARCHAR)` wrapper, schema reports `DATE` / `TIME` / `TIMESTAMP`).
- Add `CalciteAnalyticsDatetimeWireFormatIT` as the wire-format regression net.

### 1. Unit tests — RelNode shape + return-type assertions

`api/src/test/java/org/opensearch/sql/api/spec/datetime/DatetimeExtensionTest.java` (PPL V3 parser path) covers:

| Test | What it pins |
|---|---|
| `testUdfResultNormalized` | `DATE`/`TIME`/`TIMESTAMP` UDFs produce standard Calcite types, no `CAST(... VARCHAR)` wrapper |
| `testNestedUdfCallsNormalized` | UDT normalization survives through nested calls (`DATEDIFF(DATE(name), DATE(name))`) |
| `testDateLiteralNormalized` | Literal `DATE('2024-01-01')` returns `DATE`, not `VARCHAR` |
| `testFilterWithTimestampLiteral` | `WHERE created_at > "..."` keeps `TIMESTAMP` end-to-end |
| `testComparisonWithDatetimeUdf` | Mixed-type comparison promotes `DATE` → `TIMESTAMP`, never `VARCHAR` |
| `testStandardDatetimeFieldsNotWrapped` | Direct datetime field projection — no projection wrapper at all |
| `testNonDatetimeFieldsNotWrapped` | Non-datetime fields are untouched (regression guard) |
| `testDatetimeFieldsPreserveStandardTypes` | **End-to-end JDBC schema check** — compiles + executes a plan, asserts `ResultSetMetaData` reports `java.sql.Types.DATE` / `TIME` / `TIMESTAMP`, **not** `VARCHAR` |

`DatetimeExtensionSqlTest` covers the same shape on the SQL V2 parser path (3 cases).

`CalcitePlannerConcurrencyIT` (added in #5458) is preserved — its `testSequentialPlanCallsDoNotCorruptShuttleStack` case is a regression guard for the singleton-shuttle bug and exercises the post-analysis pipeline that this PR slims down.

### 2. Integration test — `CalciteAnalyticsDatetimeWireFormatIT`

`integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteAnalyticsDatetimeWireFormatIT.java` provisions a parquet-backed index on the force-routed analytics-engine cluster and asserts:

- Schema labels are `timestamp` / `date` / `time` (never `string`).
- Row values are PPL space-separated (`2024-03-15 10:30:00`), never ISO `T`-separator or trailing `Z`.

### 3. Curl wire-format probes (manual, pre-PR)

Driver: `/tmp/ae-curl-verify/run-verify.py` against a force-routed runTask cluster on `parquet_wire_format_dt`. Coverage:

- **Routing proofs** — `_explain` shows `LogicalTableScan` + lowercase `opensearch` (AE route) on the parquet index, and `CalciteLogicalIndexScan` + capital `OpenSearch` on the legacy probe index, confirming routing is binary.
- **No `CAST(... VARCHAR)`** — `_explain` of `fields ts`, `fields d, t`, and `sort ts | fields ts` contain neither `CAST(... VARCHAR)` nor `to_char`.
- **11 pinned wire-format queries** — equality filter, projection, eval-rebind, `min`/`max`/`stats`, `>` filter, `sort`, nanosecond precision, `dc()`. Each asserts schema label and PPL space-separator value.
- **7 broader queries** — `where`+`fields`, `head`+`fields`, `stats min/max`, `group by date`, `dc(d)`, multi-column projection sorted by `ts`, nanos `>` filter.

All 18 wire-format probes return `timestamp`/`date`/`time` schema labels with space-separated values; no `T` separator, no trailing `Z`, no `string` schema labels.

### 4. Local CI gates (run before each push)

- [x] `:api:test --tests DatetimeExtensionTest` → 8/8 pass
- [x] `:api:test --tests DatetimeExtensionSqlTest` → 3/3 pass
- [x] `:api:spotlessJavaCheck`
- [x] `:integ-test:compileTestJava`

## Test plan

- [x] UTs (`DatetimeExtensionTest`, `DatetimeExtensionSqlTest`)
- [x] Wire-format curl sweep (18/18 PASS)
- [x] `CalciteAnalyticsDatetimeWireFormatIT` shape
- [ ] CI gates
- [ ] `:integ-test:integTestRemote` against force-routed AE cluster

Fixes #5420


dai-chen (commented):
Thanks for the fix! Please check the CI failure.

api/src/test/java/org/opensearch/sql/api/spec/datetime/DatetimeExtensionTest.java
dai-chen: np: some test queries look wrong using `name`. Could you help fix it too? Thanks

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteAnalyticsDatetimeWireFormatIT.java
```diff
@@ -0,0 +1,291 @@
...
+  @Test
+  public void testTimestampOrderByTemporalSemantics() throws IOException {
+    String query = "source=" + INDEX + " | sort ts | fields ts";
+    assertRoutedToAnalyticsEngine(query);
+    JSONObject result = executeQuery(query);
+    verifySchema(result, schema("ts", "timestamp"));
+    verifyDataRows(result, rows("2024-03-15 10:30:00"), rows("2024-03-16 23:59:59"));
+  }
+
+  /* ---------- 3. Precision (Chendai's "losing precision" concern) ---------- */
```
dai-chen: Could you do some cleanup on the comments :)

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteAnalyticsDatetimeWireFormatIT.java
```diff
@@ -0,0 +1,291 @@
...
+ * labels ({@code timestamp}/{@code date}/{@code time}, never {@code string}).
+ *
+ * <p>Before the fix, {@code DatetimeOutputCastRule} (SQL plugin) wrapped every datetime root column
+ * in {@code CAST(... AS VARCHAR)} and {@code DatetimeOutputCastRewriter} (analytics sandbox)
+ * translated that to {@code to_char(ts, '%Y-%m-%d %H:%M:%S%.f')} server-side. Both are removed by
+ * sql#5454 / OpenSearch#21748; the response pipeline now relies on AE returning real datetime cells
+ * and {@code ExprValueUtils.fromObjectValue} → {@code ExprTimestampValue.value()} producing the
+ * documented format. These tests are the regression net for that contract.
+ */
+public class CalciteAnalyticsDatetimeWireFormatIT extends PPLIntegTestCase {
```
dai-chen, line 28: just wonder is what we test here same as existing datetime IT or any special case/assertion here?

dai-chen (dismissed):
Add some minor comments. Thanks for the fix!

dai-chen:
CI seems failing on spotless check

api/src/test/java/org/opensearch/sql/api/spec/datetime/DatetimeExtensionTest.java
mengweieric: Good catch — renamed the column `name` → `event_str` (still VARCHAR) in bot `DatetimeExtensionTest` and `DatetimeExtensionSqlTest`. Plan shape unchanged ($1 ref); the queries now read sensibly.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteAnalyticsDatetimeWireFormatIT.java
mengweieric: Done — collapsed the class Javadoc, dropped the section dividers, and trimmed per-method Javadocs to one line each to match the style of the other `Calcite*IT` files.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteAnalyticsDatetimeWireFormatIT.java
mengweieric: The unique value here is the routing pin (`assertRoutedToAnalyticsEngine`) plus the `assumeTrue(parquet)` skip. Existing datetime ITs assert wire format but not which backend served the query, so a silent Calcite fallback would leave those green. This IT fails fast in that case. I've reframed the class Javadoc to make that explicit.

dai-chen (approved):

ahkcs (approved):

RyanL1997 (approved):

PR #5455, opened 2026-05-20
Fix distribution build: pin analytics-api dep to 3.7.0-SNAPSHOT

### LINKED ISSUE #5434: [AUTOCUT] Distribution Build Failed for sql-3.7.0
***Build Failed Error***: **sql failed during the distribution build for version: 3.7.0.**
                    Please see build log at https://build.ci.opensearch.org/job/distribution-build-opensearch/11942/display/redirect.
                    The failed build stage will be marked as unstable :warning: . Please see ./build.sh step for more details.
                    Checkout the [wiki](https://github.com/opensearch-project/opensearch-build/wiki/Building-an-OpenSearch-and-OpenSearch-Dashboards-Distribution) to reproduce the failure locally.

lezzago:
## Summary

Pin `analytics-api` to `3.7.0-SNAPSHOT` in `plugin/build.gradle`. The previous `${opensearch_version}` resolves to `3.7.0` in release builds, but sandbox artifacts are never published with a release coordinate, so the build fails:

```
Could not find org.opensearch.sandbox:analytics-api:3.7.0
```

Same hardcoded snapshot pattern is already used in `core/build.gradle:67` for the same artifact.

Resolves #5434

## Test plan

- [x] `./gradlew assemble -DskipTests=true` (snapshot mode) — passes
- [x] `bash scripts/build.sh -v 3.7.0 -p linux -a arm64 -s false -o builds` reproduced locally — passes (was failing on `main`)
- [x] Plugin loads cleanly into a 3.7.0 distribution alongside 10 sibling plugins; cluster reaches GREEN
- [x] CI distribution build passes

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -160,7 +160,7 @@ dependencies {
 
     api project(":ppl")
     api project(':api')
-    implementation 'org.opensearch.sandbox:analytics-api:3.7.0-SNAPSHOT'
+    implementation "org.opensearch.sandbox:analytics-api:${opensearch_version.tokenize('-')[0]}-SNAPSHOT"
     api project(':legacy')
     api project(':opensearch')
     api project(':prometheus')
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -160,7 +160,7 @@ dependencies {
 
     api project(":ppl")
     api project(':api')
-    implementation("org.opensearch.sandbox:analytics-api:${opensearch_version}")
+    implementation 'org.opensearch.sandbox:analytics-api:3.7.0-SNAPSHOT'
     api project(':legacy')
     api project(':opensearch')
     api project(':prometheus')


lezzago:
Re: Code-Diff-Analyzer findings — both flags are about the hardcoded `3.7.0-SNAPSHOT` coordinate. Adding context for reviewers:

**Why a literal `3.7.0-SNAPSHOT` instead of `${opensearch_version}`:**

`opensearch_version` resolves to `3.7.0` in the distribution build (`./build.sh -s false`). For most artifacts that's correct, but `org.opensearch.sandbox:analytics-api` is a sandbox subproject of OpenSearch core, and OpenSearch's [`sandbox/build.gradle`](https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/build.gradle) **deliberately disables publish tasks for sandbox artifacts unless `-Dsandbox.enabled=true` AND `build.snapshot=true`** — explicitly to "prevent accidental inclusion of these artifacts in a release distribution." Result: `analytics-api:3.7.0` (release coord) does not exist anywhere; only `3.7.0-SNAPSHOT` is published. That mismatch is the cause of #5434.

**Why this isn't a new pattern:** [`core/build.gradle:67`](https://github.com/opensearch-project/sql/blob/main/core/build.gradle#L67) already declares `'org.opensearch.sandbox:analytics-api:3.7.0-SNAPSHOT'` for the same reason. This PR brings the plugin module in line.

**Re: SNAPSHOT supply-chain risk:** the SQL plugin already pulls `analytics-api:3.7.0-SNAPSHOT` transitively via `core/build.gradle`, so this PR doesn't change the trust surface. The artifact is hosted on the OpenSearch project's own snapshot repo (`ci.opensearch.org/ci/dbc/snapshots/maven/`), not Maven Central. Pinning to a snapshot here is consistent with how the rest of the project handles sandbox dependencies. This will be removable once OpenSearch core promotes `analytics-api` out of `sandbox/` (or once the plugin can use the `AnalyticsFrontEndExtension` SPI tracked in [opensearch-project/OpenSearch#21449](https://github.com/opensearch-project/OpenSearch/issues/21449)).

plugin/build.gradle
```diff
@@ -160,7 +160,7 @@ dependencies {
 
     api project(":ppl")
     api project(':api')
-    implementation("org.opensearch.sandbox:analytics-api:${opensearch_version}")
+    implementation 'org.opensearch.sandbox:analytics-api:3.7.0-SNAPSHOT'
```
ps48: Why is the env var not working?  https://github.com/oepnsearch-project/sql/blob/9ae998347aa4cd3d4a074f954a30eba9fa7d721a/build.gradle#L9

lezzago: Good catch — bare `${opensearch_version}` doesn't work because that variable swings between two values:

| Mode | `-Dopensearch.version` | resolves to |
|---|---|---|
| Local dev (default) | not passed | `3.7.0-SNAPSHOT` ✅ |
| CI distribution build | `3.7.0` | `3.7.0` ❌ — sandbox artifact never published as a release coord |

OpenSearch core's [`sandbox/build.gradle`](https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/build.gradle) explicitly disables release publishing for sandbox artifacts ("prevent accidental inclusion of these artifacts in a release distribution"), so `analytics-api:3.7.0` doesn't exist anywhere — only `3.7.0-SNAPSHOT` does. That's the actual #5434 failure.

Appending `-SNAPSHOT` directly (`${opensearch_version}-SNAPSHOT`) doesn't work either: in local dev that produces `3.7.0-SNAPSHOT-SNAPSHOT`.

Pushed [8b62fc2](https://github.com/opensearch-project/sql/pull/5455/commits/8b62fc220) which uses the same tokenize pattern `build.gradle:12` already uses for `opensearch_build`:

```gradle
implementation "org.opensearch.sandbox:analytics-api:${opensearch_version.tokenize('-')[0]}-SNAPSHOT"
```

Strips any existing `-SNAPSHOT` suffix, then appends one. Resolves to `3.7.0-SNAPSHOT` in both modes; auto-tracks version bumps. Verified locally with both `-Dopensearch.version=3.7.0 -Dbuild.snapshot=false` (CI release) and the default snapshot mode.

RyanL1997 (approved):

ps48 (approved):

PR #5458, opened 2026-05-20
Fix singleton stack-corruption NPE in DatetimeUdtNormalizeRule

ahkcs:
## What this fixes

Some PPL queries to parquet-backed indices return HTTP 500 with
`NoSuchElementException` when several queries hit the cluster at the same time
— the failure mode reported on dashboard "field statistics" panels that probe
many fields in parallel. Affected query shape:

```
source = my_index | stats count() as field_count, distinct_count(<datetime_field>) as distinct_count
```

The failure is intermittent: same query passes or fails depending on what
other queries run at the same time.

## Why it happens

The unified query API runs two Calcite rewrites on every plan — one normalizes
datetime UDT types (`DatetimeUdtNormalizeRule`), one casts datetime output to
varchar (`DatetimeOutputCastRule`). Both extend `RelHomogeneousShuttle`, which
in Calcite inherits a `Deque<RelNode> stack` field used internally for
push/pop during tree traversal. **That stack is a plain non-thread-safe
`ArrayDeque`.**

`DatetimeExtension.postAnalysisRules()` returned a static `INSTANCE` of each
rule, so every `plan()` call shared the same shuttle — and the same stack.
When dashboard issues N field-stats queries at once, N cluster threads call
`plan()` concurrently. Their `push` and `pop` operations on the shared
`ArrayDeque` interleave, leaving residual entries on the stack. A subsequent
traversal then tries to `pop()` an entry that isn't there, and the cluster
returns 500 with this stack trace:

```
java.util.NoSuchElementException
  at java.util.ArrayDeque.pop(ArrayDeque.java:591)
  at org.apache.calcite.rel.RelShuttleImpl.visitChild(RelShuttleImpl.java:67)  ← finally{ stack.pop() }
  at org.apache.calcite.rel.RelShuttleImpl.visit(RelShuttleImpl.java:151)
  at org.opensearch.sql.api.spec.datetime.DatetimeUdtNormalizeRule.visit(DatetimeUdtNormalizeRule.java:33)
  at org.apache.calcite.rel.RelHomogeneousShuttle.visit(RelHomogeneousShuttle.java:43)
  at org.apache.calcite.rel.logical.LogicalAggregate.accept(LogicalAggregate.java:159)
```

`RelShuttleImpl` was never designed to be a singleton: the stack is per-shuttle
mutable state. Sharing it across queries is unsafe regardless of timing — the
race just amplifies the chance of a visible failure.

## What this PR changes

`api/src/main/java/org/opensearch/sql/api/spec/datetime/DatetimeExtension.java`
now returns a **fresh instance per `plan()` call**:

```diff
   @Override
   public List<RelShuttle> postAnalysisRules() {
-    return List.of(DatetimeUdtNormalizeRule.INSTANCE, DatetimeOutputCastRule.INSTANCE);
+    // Fresh instances per plan() because RelHomogeneousShuttle inherits a stateful stack.
+    return List.of(new DatetimeUdtNormalizeRule(), new DatetimeOutputCastRule());
   }
```

The `INSTANCE` constants and the Lombok `@NoArgsConstructor(PRIVATE)` are
removed from both rules. Class JavaDoc on each rule now explicitly warns
against making it a singleton again.

## Test plan

| Cluster state | `testConcurrentStatsDistinctCountOverDatetime` | `testConcurrentMixedDatetimePlans` |
|---|---|---|
| **Without fix** (singleton `INSTANCE`)         | **2/80 queries fail** with `NoSuchElementException` (HTTP 500) | **3/80 fail** with same error |
| **With fix** (fresh instances per `plan()`)    | **80/80 pass**                                                | **80/80 pass** |

Reproduced and verified end-to-end via the analytics-engine path with
parquet-backed indices:

```bash
  -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9300 \
  -Dtests.clustername=runTask \
  -Dtests.analytics.force_routing=true \
  -Dtests.analytics.parquet_indices=true \
  --tests "*CalciteDatetimeUdtNormalizeRegressionIT"
```

The IT (`CalciteDatetimeUdtNormalizeRegressionIT`) fires 80 queries across an
8-thread pool against a parquet-backed `DATE_FORMATS` index with multiple
datetime columns — mirroring the dashboard field-stats workload.

- [x] `:api:test` and `:core:test` pass
- [x] `:api:spotlessCheck` passes
- [x] Concurrent IT reliably reproduces the failure against unfixed code
- [x] Same IT passes 0/80 failures against fixed code
- [x] Unit test `testSequentialPlanCallsDoNotCorruptShuttleStack` in
      `DatetimeExtensionTest` guards against future reintroduction of the
      singleton

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDatetimeUdtNormalizeRegressionIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDatetimeUdtNormalizeRegressionIT.java
@@ -5,11 +5,15 @@
 
 package org.opensearch.sql.calcite.remote;
 
-import static org.junit.Assert.assertNotNull;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_DATE_FORMATS;
 
-import java.io.IOException;
-import org.json.JSONObject;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.Test;
 import org.opensearch.sql.ppl.PPLIntegTestCase;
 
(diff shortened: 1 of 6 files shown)


RyanL1997 (dismissed):

ritvibhatt (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDatetimeUdtNormalizeRegressionIT.java
```diff
@@ -0,0 +1,153 @@
...
+ * <pre>{@code
+ * ./gradlew :integ-test:integTestRemote \
+ *   -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9300 \
+ *   -Dtests.clustername=runTask \
+ *   -Dtests.analytics.force_routing=true \
+ *   -Dtests.analytics.parquet_indices=true \
+ *   --tests org.opensearch.sql.calcite.remote.CalciteDatetimeUdtNormalizeRegressionIT
+ * }</pre>
+ */
+public class CalciteDatetimeUdtNormalizeRegressionIT extends PPLIntegTestCase {
```
dai-chen: np: we can rename this as general test for our planner.

dai-chen (dismissed):
Thanks for the fix!

RyanL1997 (dismissed):

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteDatetimeUdtNormalizeRegressionIT.java
ahkcs: Renamed to `CalcitePlannerConcurrencyIT` and reframed the JavaDoc as a general bucket for planner-level concurrency / state-isolation regressions. The current test methods are noted as the motivating regression — future planner-concurrency tests can land alongside. Pushed as `0c1b7017a`.

mengweieric (approved):

RyanL1997 (approved):

dai-chen (approved):

PR #5463, opened 2026-05-22
Added UDT for IP and Binary support

vinaykpud:
### Description

Companion change for [opensearch-project/OpenSearch#21807](https://github.com/opensearch-project/OpenSearch/pull/21807) which adds Calcite UDTs (`IpType`, `BinaryType`) for OpenSearch `ip` and `binary` columns to the analytics-engine path. This PR teaches the SQL plugin to recognize the new UDTs at the response-schema boundary so:

- `byte[]` cells from the analytics backend get rendered as canonical IP strings / base64 strings (was: `unsupported object class [B`).
- The response schema reports `"type": "ip"` / `"type": "binary"` correctly (was: both labelled `"binary"`).
- Existing comparison and `CIDRMATCH` query shapes keep working unchanged — the UDT is invisible to Calcite's planner-internal coercion machinery.

#### Background

`OpenSearchSchemaBuilder` (in the OpenSearch sandbox) used to map both `ip` and `binary` field types to plain `SqlTypeName.VARBINARY`. That collapse caused three downstream bugs in this plugin's PPL path through the analytics engine:

1. The DataFusion backend ships back a 16-byte ipv4-mapped-ipv6 buffer; `AnalyticsExecutionEngine.convertRows` calls `ExprValueUtils.fromObjectValue(byte[])` which has no `byte[]` case → `unsupported object class [B`.
2. The response schema reports `"type": "binary"` for an IP column (because `convertRelDataTypeToExprType(VARBINARY) → BINARY` ExprType).
3. CIDRMATCH against a column was registered with `PPLTypeChecker.family(SqlTypeFamily.BINARY, SqlTypeFamily.STRING)`, with the byte-range expansion living inline at the registration site.

The companion OpenSearch PR introduces `IpType` / `BinaryType` (Calcite UDTs that extend `AbstractSqlType` with `VARBINARY` underneath) and moves CIDRMATCH dispatch into a backend `ScalarFunctionAdapter`.

#### Approach

Two response-boundary changes plus one cleanup, all minimal:

**1. Result-column UDT recognition (`OpenSearchTypeFactory`).**
New sibling function `convertAnalyticsEngineRelDataTypeToExprType` does an `instanceof` dispatch:
```java
if (type instanceof IpType) return IP;
if (type instanceof BinaryType) return BINARY;
return convertRelDataTypeToExprType(type);
```
The original `convertRelDataTypeToExprType` is **deliberately unchanged** — Calcite's coercion machinery round-trips through it, so returning `IP` ExprType for a VARBINARY column synthesizes `IP(string)` casts that DataFusion can't resolve. Keeping UDT recognition in a sibling function isolates it to the response-schema path.

**2. Per-column UDT dispatch in row conversion (`AnalyticsExecutionEngine.convertRows`).**
New static helper `toExprValue(Object value, RelDataType type)`:
- `byte[]` + `IpType` → `InetAddresses.toAddrString(InetAddress.getByAddress(bytes))` (matches `IpFieldMapper.valueFetcher` semantics: dotted-quad for IPv4 / IPv4-mapped, RFC 5952 for pure IPv6).
- `byte[]` + `BinaryType` → `Base64.getEncoder().encodeToString(bytes)` (matches the OpenSearch `binary` field wire contract).
- Otherwise → `ExprValueUtils.fromObjectValue(value)` unchanged.

`UnknownHostException` from a malformed IP buffer falls through to the default handler so the user sees a clear error rather than a malformed address string.

**3. CIDRMATCH cleanup (`PPLFuncImpTable`).**
- Removes `udf/ip/CidrMatchAdapter` and its inline registration. CIDRMATCH dispatch now lives in `CidrMatchFunctionAdapter` on the backend (in the companion OpenSearch PR), which means it serves both the production SQL-plugin variant and the sandbox test front-end with a single implementation.
- Collapses the two-line registration (typecheck + `withRexBuilderShim`) to a single typecheck-only `registerOperator(CIDRMATCH, PPLBuiltinOperators.CIDRMATCH)`.
- The runtime `CidrMatchFunction` UDF stays as the dynamic last-resort fallback.

### Testing

**Unit tests added:**
- `OpenSearchTypeFactoryTest`:
  - `testConvertResultColumnIpTypeReturnsIpExprType` — `IpType` → `ExprCoreType.IP`.
  - `testConvertResultColumnBinaryTypeReturnsBinaryExprType` — `BinaryType` → `ExprCoreType.BINARY`.
  - `testConvertResultColumnPlainVarbinaryFallsBackToBinary` — plain `VARBINARY` (no UDT) keeps returning `BINARY`.
  - `testConvertResultColumnDelegatesParityForNonUdtTypes` — for every non-UDT `RelDataType`, the result-column variant must agree with the planner-internal variant. Drift here would mean response-schema labels diverge from what Calcite's coercion sees.

- `AnalyticsExecutionEngineTest`:
  - `executeRelNode_ipColumnRendersAsAddressString` — both ipv4-mapped IPv6 (`::ffff:1.2.3.4` → `"1.2.3.4"`) and pure IPv6 (`::1`) buffers render to the right canonical form, schema reports `"ip"`.
  - `executeRelNode_binaryColumnRendersAsBase64` — `byte[]` payload base64-encodes to match OpenSearch wire format, schema reports `"binary"`.

Existing tests still pass. `[B`-class regression caught at: `physicalPlanExecute_callsOnFailure` (already in the file, exercises the same converter dispatch).

**End-to-end** (against a single-node `gradle run` cluster with the companion OpenSearch PR applied):
- `cast(host as STRING)` returns `"1.2.3.4"` / `"::1"` (was a 16-byte garbage buffer).
- `cast(blob as STRING)` matches `| fields blob` (base64-encoded).
- `where host = '1.2.3.4'` coerces cleanly with no `Unable to convert call IP(string)` regression.
- `cidrmatch(host, '1.2.3.0/24')` (column form) and `CIDRMATCH('1.2.3.4', '1.2.3.0/24')` (literal form) both return correct row counts via the backend adapter.
- Response schema labels: `"type": "ip"` for `ip` columns, `"type": "binary"` for `binary` columns.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
+++ b/core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
@@ -5,7 +5,10 @@
 
 package org.opensearch.sql.executor.analytics;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -14,6 +17,9 @@ import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.opensearch.analytics.exec.QueryPlanExecutor;
+import org.opensearch.analytics.schema.BinaryType;
+import org.opensearch.analytics.schema.IpType;
+import org.opensearch.common.network.InetAddresses;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.sql.ast.statement.ExplainMode;
 import org.opensearch.sql.calcite.CalcitePlanContext;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsExecutionEngine.java
```diff
@@ -123,15 +129,47 @@ private List<ExprValue> convertRows(Iterable<Object[]> rows, List<RelDataTypeFie
...
+   *       the OpenSearch {@code binary} field wire format).
+   *   <li>Anything else &rarr; existing {@link ExprValueUtils#fromObjectValue} path.
+   * </ul>
+   *
+   * <p>Without this dispatch, {@code fromObjectValue} throws {@code unsupported object class
+   * [B} on byte[] cells, and IP buffers leak through as raw 16-byte ipv4-mapped-ipv6 garbage.
+   */
+  private static ExprValue toExprValue(Object value, RelDataType type) {
+    if (value instanceof byte[] bytes) {
+      if (type instanceof IpType) {
```
penghuo, line 158: AE report schema should include IpType [OpenSearchSchemaBuilder.java‎](https://github.com/opensearch-project/OpenSearch/pull/21807/changes#diff-ff824465bf640034c87540d3ce045e50311d308fc1fa4baf15fba37968dc431b), otherwise, Calciter Planer does not aware of the column is IpType, right?

vinaykpud: Yes, and that's exactly what `OpenSearchSchemaBuilder.buildLeafType` does.  `buildLeafType` wraps ip columns in IpType and binary columns in BinaryType. So the row type emitted by the schema builder carries IpType and BinaryType on ip/binary columns.

penghuo (dismissed):

penghuo (approved):

dai-chen (approved):

PR #5466, opened 2026-05-23
Register DISTINCT_COUNT_APPROX logical marker in PPLFuncImpTable

songkant-aws:
### Description

PPL parser fails with `Cannot resolve function: DISTINCT_COUNT_APPROX` on any execution path that does not construct `OpenSearchExecutionEngine` (unified-query / analytics-engine / DataFusion), because the UDAF was only registered to `PPLFuncImpTable.aggExternalFunctionRegistry` as a side effect of that constructor (`registerOpenSearchFunctions`).

This change adds a logical marker UDAF in `core` that lets PPL parser succeed on every path. Backends are expected to push down or rewrite the operator before execution.

#### Layered registration

- **`core` (this PR)**: register a marker `DistinctCountApproxLogicalAggFunction` in `PPLFuncImpTable.AggBuilder.populate()` via the new `PPLBuiltinOperators.DISTINCT_COUNT_APPROX`. Lookup precedence in `PPLFuncImpTable.getImplementation()` is `aggExternalFunctionRegistry` first, then `aggFunctionRegistry`.
- **`opensearch` (unchanged)**: `OpenSearchExecutionEngine.registerOpenSearchFunctions()` still installs the real `DistinctCountApproxAggFunction` (HyperLogLog++) into `aggExternalFunctionRegistry`. Because external is consulted first, the V3 path always sees the real implementation once the constructor has run, and the marker is effectively overridden.

This is the same pattern already used by `RelevanceQueryFunction.RelevanceQueryImplementor` for `match`/`match_phrase`/etc — relevance functions register a no-op operator whose `RelevanceQueryImplementor.implement()` throws `UnsupportedOperationException`, because their semantics live entirely on the OpenSearch index side. `DISTINCT_COUNT_APPROX` is in the same situation: real semantics on the OpenSearch side (cardinality aggregation) and on backends like DataFusion (`approx_count_distinct`).

#### Marker class behavior

`init` / `add` / `result` and the accumulator's `value` all throw `UnsupportedOperationException` with the message:

> DISTINCT_COUNT_APPROX logical marker reached Enumerable execution; an engine-specific implementation must be registered or rewritten before execution.

Reaching the body means a backend either failed to push down or did not register an adapter — surfacing this as a clear error rather than silently producing wrong results is the intended behavior.

#### V3 path behavior is preserved

- `AggregateAnalyzer` translates `DISTINCT_COUNT_APPROX` to OpenSearch `cardinality` aggregation through a `BuiltinFunctionName` switch, independent of the wrapped `SqlAggFunction` instance — unchanged by this PR.
- When V3 falls back to Calcite enumerable execution (push-down failed for some sub-plan), `RexImpTable.getAggImplementor` reads `SqlUserDefinedAggFunction.function`, which is the real `DistinctCountApproxAggFunction` registered by `OpenSearchExecutionEngine` (external takes precedence) — unchanged.
- Operand metadata for the marker is `null` to match the existing external registration's permissive type policy. No new type rejection is introduced.

#### How this unblocks unified-query / DataFusion path

`RestUnifiedQueryAction` does not construct `OpenSearchExecutionEngine`; before this change, `dc()` / `distinct_count()` / `distinct_count_approx()` failed at PPL parse stage on that path. After this change, the parser succeeds via the marker, and the downstream backend (e.g. analytics-engine's DataFusion adapter) rewrites `RexOver(DISTINCT_COUNT_APPROX)` to `APPROX_COUNT_DISTINCT` before substrait emission. End-to-end verified locally with the analytics-engine REST IT in the OpenSearch sandbox.

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/udf/udaf/DistinctCountApproxLogicalAggFunction.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/udf/udaf/DistinctCountApproxLogicalAggFunction.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.sql.calcite.udf.udaf;
+
+import org.opensearch.sql.calcite.udf.UserDefinedAggFunction;
+
+/**
+ * Logical marker UDAF for {@code DISTINCT_COUNT_APPROX}. Lets the PPL parser produce a RelNode that
+ * contains the operator without committing to a JVM execution path; backends are expected to push
+ * it down or rewrite it before execution:
+ *
+ * <ul>
+ *   <li>OpenSearch V3 path: {@code OpenSearchExecutionEngine#registerOpenSearchFunctions} registers
+ *       a real HyperLogLog++ implementation in {@code PPLFuncImpTable.aggExternalFunctionRegistry},
+ *       which overrides this marker (external registry has lookup precedence in {@code
+ *       getImplementation}). {@code AggregateAnalyzer} then translates the operator to OpenSearch
+ *       cardinality DSL.
+ *   <li>Unified-query / DataFusion / analytics-engine path: backend planner rewrites the RexOver to
+ *       {@code APPROX_COUNT_DISTINCT} (Calcite stdop) before substrait emission; the DataFusion
+ *       substrait reader consumes that natively.
+ * </ul>
+ *
+ * <p>This class deliberately throws on every method. Reaching a method body means a backend either
+ * failed to push down or did not register an adapter — that is a configuration bug, not a runtime
+ * fallback. {@code RelevanceQueryFunction.RelevanceQueryImplementor} (used by {@code match}, {@code
+ * match_phrase}, etc.) follows the same pattern for relevance search functions that have no JVM
+ * execution semantics.
+ */
+public class DistinctCountApproxLogicalAggFunction
+    implements UserDefinedAggFunction<DistinctCountApproxLogicalAggFunction.MarkerAccumulator> {
+
+  private static final String NOT_EXECUTABLE =
+      "DISTINCT_COUNT_APPROX logical marker reached Enumerable execution; "
+          + "an engine-specific implementation must be registered or rewritten before execution.";
+
+  @Override
+  public MarkerAccumulator init() {
+    throw new UnsupportedOperationException(NOT_EXECUTABLE);
+  }
+
+  @Override
+  public MarkerAccumulator add(MarkerAccumulator acc, Object... values) {
+    throw new UnsupportedOperationException(NOT_EXECUTABLE);
+  }
+
+  @Override
+  public Object result(MarkerAccumulator accumulator) {
+    throw new UnsupportedOperationException(NOT_EXECUTABLE);
+  }
+
+  /** Placeholder accumulator. Never actually constructed because {@link #init()} throws. */
+  public static class MarkerAccumulator implements Accumulator {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/opensearch/sql/calcite/udf/udaf/DistinctCountApproxLogicalAggFunction.java
```diff
@@ -0,0 +1,61 @@
...
+ * </ul>
+ *
+ * <p>This class deliberately throws on every method. Reaching a method body means a backend either
+ * failed to push down or did not register an adapter — that is a configuration bug, not a runtime
+ * fallback. {@code RelevanceQueryFunction.RelevanceQueryImplementor} (used by {@code match}, {@code
+ * match_phrase}, etc.) follows the same pattern for relevance search functions that have no JVM
+ * execution semantics.
+ */
+public class DistinctCountApproxLogicalAggFunction
+    implements UserDefinedAggFunction<DistinctCountApproxLogicalAggFunction.MarkerAccumulator> {
```
LantaoJin, line 33: how many udf/udaf are skipping the JVM execution path now? How about add a new parent class for it, such as
```java
class DistinctCountApproxLogicalAggFunction
  implements NativeSpecificFunction, UserDefinedAggFunction<DistinctCountApproxLogicalAggFunction.MarkerAccumulator>
```

songkant-aws: There are only a couple of functions in this category today:

- GEOIP is a UDF that depends on NodeClient / OpenSearch-side lookup. (Datafusion won't support it for now)
- DISTINCT_COUNT_APPROX is a UDAF whose OpenSearch implementation uses the HLL++ core search cardinality aggregation path.

Both are registered from OpenSearchExecutionEngine.registerOpenSearchFunctions():
https://github.com/opensearch-project/sql/blob/main/opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java#L329-L355

That registration only happens on the SQL/PPL V3 OpenSearchExecutionEngine path. The unified-query / analytics-engine path does not construct OpenSearchExecutionEngine; it uses UnifiedQueryPlanner to produce a Calcite RelNode via CalciteRelNodeVisitor, then passes that RelNode to the backend planner. As a result, PPLFuncImpTable cannot rely on aggExternalFunctionRegistry being populated on that path.

songkant-aws: @LantaoJin Could we hold on adding a new parent marker class? Because there is only a single UDAF case impacted. 

I just checked that a NativeSpecificFunction marker would mostly only serve as a self-documenting type hint here — sandbox/analytics plugins don't depend on sql plugin's concrete classes (they dispatch on SqlOperator.getName() to their own enums via Calcite RelNode as a neutral ABI), and inside sql plugin there's no centralized hook (e.g., validator pass or RexImpTable) that would actually leverage instanceof on it today. Happy to revisit the abstraction in a follow-up once a third instance and a concrete consumer (validation pass /centralized rejection point) emerge.

LantaoJin (approved):

core/src/main/java/org/opensearch/sql/calcite/udf/udaf/DistinctCountApproxLogicalAggFunction.java
LantaoJin: not a blocker. LGTM

yuancu (approved):

PR #5467, opened 2026-05-26
feat(ppl): wire patterns command for analytics-engine dashboard route

ahkcs:
## What

Wires PPL `patterns` (BRAIN + SIMPLE methods, label + aggregation modes) through the analytics-engine route, including this OpenSearch Dashboards query:

```ppl
source = test_logs_parquet
| where `@timestamp` >= '...' and `@timestamp` <= '...'
| patterns message method=brain mode=label
| stats count() as pattern_count, take(message, 1) as sample_logs by patterns_field
| sort - pattern_count
| fields patterns_field, pattern_count, sample_logs
```

Companion to **opensearch-project/OpenSearch#21797** (DataFusion UDFs + analytics-engine planner work + DataFusion `regexp_replace` `'g'` flag in `RegexpReplaceAdapter`). Both PRs are required.

## Changes

| Commit | Why |
|---|---|
| `feat(api)`: add `PATTERN_*` defaults to `UnifiedQueryContext` | Without these, `patterns` queries that omit explicit `method=`/`mode=` died on `PatternMethod.valueOf("NULL")` in `AstBuilder.visitPatternsCommand` before reaching any capability check. |
| `test(api)`: `UnifiedQueryPlannerTest.testPPLPatternsPicksUpDefaults` | Pins that the PATTERN_* defaults flow through AstBuilder so a bare `patterns <field>` produces a SIMPLE / LABEL RelNode with `REGEXP_REPLACE` + a `patterns_field` projection. |
| `test(integ-test)`: `CalcitePPLDashboardPatternsIT` + suite inclusion | Pins the Dashboards BRAIN-label panel query end-to-end with `verifyDataRows` against `HDFS_LOGS`. Added to `CalciteNoPushdownIT` so the no-pushdown variant runs too. |

## Results

| Suite | Before | After |
|---|---|---|
| `CalcitePPLDashboardPatternsIT` (new) | — | **1 / 1** |
| `CalcitePPLPatternsIT` (analytics-engine) | 3 / 15 | **10 / 15** |
| `CalcitePPLPatternsIT` (Calcite/V2) | 15 / 15 | **15 / 15** |

Verified on all four configurations: Calcite/V2 (pushdown on + off) and analytics-engine route (pushdown on + off).

Remaining 5 / 15 patterns failures — analytics-engine-route-only, tracked separately:
- 3 × `BrainAggregationMode_*` → planner needs `LogicalCorrelate` post-aggregate.
- 2 × `Brain*Mode_ShowNumberedToken` → substrait type wiring for `_ShowNumberedToken`.


api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
```diff
@@ -145,12 +150,24 @@ public static class Builder {
...
+                Map.entry(
+                    PPL_JOIN_SUBSEARCH_MAXOUT, SysLimit.UNLIMITED_SUBSEARCH.joinSubsearchLimit()),
+                Map.entry(CALCITE_ENGINE_ENABLED, true),
+                Map.entry(PPL_REX_MAX_MATCH_LIMIT, 10),
+                // PPL `patterns` command defaults — mirror the cluster-side defaults registered in
+                // OpenSearchSettings (DEFAULT_PATTERN_METHOD_SETTING etc.). Without these the
+                // analytics-engine path's AstBuilder.visitPatternsCommand reads null from
+                // `settings.getSettingValue(Key.PATTERN_METHOD)`, fails with
+                // `PatternMethod.valueOf("NULL")` IllegalArgumentException, and every query that
+                // omits an explicit `method=` / `mode=` argument is rejected.
```
penghuo: delete all these "AI" comments.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -4301,9 +4301,25 @@ private void buildParseRelNode(Parse node, CalcitePlanContext context) {
...
-          PPLFuncImpTable.INSTANCE.resolve(
-              context.rexBuilder, ParseUtils.BUILTIN_FUNCTION_MAP.get(parseMethod), rexNodeList);
+      RexNode innerRex;
+      if (ParseMethod.PATTERNS.equals(parseMethod)) {
+        // Emit `regexp_replace(field, pattern, replacement, "g")` directly so the replacement
+        // is global (every match replaced). DataFusion's `regexp_replace` defaults to FIRST
+        // match only without the "g" flag — using the 3-arg form via the REPLACE handler
+        // produces `<*>@pyrami.com` instead of `<*>@<*>.<*>` on the analytics-engine route.
+        // Calcite's REGEXP_REPLACE_PG_4 with "g" matches what `replaceAll` does, so V2 /
+        // Calcite-path semantics are preserved.
```
penghuo: Currently, PPL use regex_replace(field, pattern, replacement) without 'g' , right?
the propsoal is add 'g' paramater? It is does not change PPL behavoiur, right, becuase default is 'g'?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/RexStandardizer.java
```diff
@@ -73,6 +74,27 @@ public RexNode visitCall(final RexCall call, ScriptParameterHelper helper) {
...
+        && call.operands.size() == 4
+        && call.operands.get(3) instanceof RexLiteral flagsLit
+        && "g".equals(flagsLit.getValueAs(String.class))) {
+      return helper.rexBuilder
+          .makeCall(
+              call.getType(),
+              SqlLibraryOperators.REGEXP_REPLACE_3,
+              call.operands.subList(0, 3))
+          .accept(this, helper);
+    }
```
penghuo: Why only this function need sepecial handling?

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
ahkcs: Updated, trimmed verbose comments

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLDashboardPatternsIT.java
```diff
@@ -0,0 +1,46 @@
...
+                "source=%s"
+                    + " | patterns content method=BRAIN mode=label"
+                    + "   max_sample_count=5 variable_count_threshold=5"
+                    + "   frequency_threshold_percentage=0.2"
+                    + " | stats count() as pattern_count, take(content, 1) as sample_logs"
+                    + "   by patterns_field"
+                    + " | sort - pattern_count"
+                    + " | fields patterns_field, pattern_count, sample_logs",
+                TEST_INDEX_HDFS_LOGS));
+    verifySchemaInOrder(
```
dai-chen, line 43: Is verifying schema sufficient?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -4301,9 +4301,21 @@ private void buildParseRelNode(Parse node, CalcitePlanContext context) {
     }
     List<RexNode> newFields = new ArrayList<>();
     for (String groupCandidate : groupCandidates) {
-      RexNode innerRex =
-          PPLFuncImpTable.INSTANCE.resolve(
-              context.rexBuilder, ParseUtils.BUILTIN_FUNCTION_MAP.get(parseMethod), rexNodeList);
+      RexNode innerRex;
+      if (ParseMethod.PATTERNS.equals(parseMethod)) {
```
dai-chen: Is it possible to merge this if-else into the one below it?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: >Currently, PPL use regex_replace(field, pattern, replacement) without 'g' , right?

Yes

Calcite's 3-arg `REGEXP_REPLACE` is already replace-all, however data fusion's `REGEXP_REPLACE` replaces only the first match, which causes mismatch. My proposal is to add this `g` parameter to make the behavior replace-all for both engines.

So our PPL behavior would not change because we will later collapse the 4-arg to 3-arg in Calcite-path in `RexStandardizer.java`

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/RexStandardizer.java
ahkcs: The purpose is specifically to let the 4-arg `REGEXP_REPLACE` survive the pushdown path.

3-arg has a matching `SqlFunctions.regexpReplace(String, String, String)` Java method — Janino binds, pushdown works. 4-arg-with-flags doesn't: Calcite's SqlFunctions ships `regexpReplace(String, String, String, int)` (position) but no `(String, String, String, String)` (flags-as-string) variant. The operator is registered, the Java impl is missing — that's the gap we rewrite around. We collapse the 4-arg to 3-arg in this case

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLDashboardPatternsIT.java
ahkcs: Added `verifyDataRows`

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: Good catch — done. Fused the two if branches into a single if/else

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
ahkcs: After taking another look, this approach can only cover the pushdown path, the no-pushdown path will be broken because of our 4-arg change

Updated the approach, the 'g' flag is now appended in OpenSearch's `RegexpReplaceAdapter` only when sending to DataFusion

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/serde/RexStandardizer.java
ahkcs: After taking another look, this approach can only cover the pushdown path, the no-pushdown path will be broken because of our 4-arg change

Updated the approach, the `g` flag is now appended in OpenSearch's RegexpReplaceAdapter only when sending to DataFusion

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
```diff
@@ -145,12 +150,18 @@ public static class Builder {
      */
     private final Map<Settings.Key, Object> settings =
         new HashMap<Settings.Key, Object>(
-            Map.of(
-                QUERY_SIZE_LIMIT, SysLimit.DEFAULT.querySizeLimit(),
-                PPL_SUBSEARCH_MAXOUT, SysLimit.UNLIMITED_SUBSEARCH.subsearchLimit(),
-                PPL_JOIN_SUBSEARCH_MAXOUT, SysLimit.UNLIMITED_SUBSEARCH.joinSubsearchLimit(),
-                CALCITE_ENGINE_ENABLED, true,
-                PPL_REX_MAX_MATCH_LIMIT, 10));
+            Map.ofEntries(
```
dai-chen, line 153: np: we can add a test in `UnifiedQueryPlannerTest` to verify the RelNode generated if needed.

dai-chen (dismissed):

api/src/main/java/org/opensearch/sql/api/UnifiedQueryContext.java
ahkcs: Added in b83058237619c627c91b1ef708b5f8b09bcdb319

penghuo (approved):

dai-chen (approved):

PR #5473, opened 2026-05-27
Propagate correct HTTP status for security exceptions on SQL analytic…

finnegancarroll:
### Description
The SQL plugin's unified query handler (/_plugins/_sql) returned HTTP 500 for all exceptions from the analytics engine path, including security exceptions that should be 403 Forbidden. This caused authorization denials to appear as internal server errors to users.

Fix: Check if the exception is an OpenSearchException and extract its proper HTTP status (e.g., 403 for OpenSearchSecurityException) instead of hardcoding 500. Both the explain and execute paths are fixed.

Also updates AnalyticsEngineSecurityIT SQL deny tests to assert 403 directly, removing the previous workaround that accepted either 403 or 500.

### Related Issues
N/A

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
@@ -263,8 +263,10 @@ public class SQLPlugin extends Plugin
 
               @Override
               public void onFailure(Exception e) {
-                channel.sendResponse(
-                    new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, e.getMessage()));
+                RestStatus status = (e instanceof org.opensearch.OpenSearchException osEx)
+                    ? osEx.status()
+                    : RestStatus.INTERNAL_SERVER_ERROR;
+                channel.sendResponse(new BytesRestResponse(status, e.getMessage()));
               }
             });
       } else {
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -263,8 +263,11 @@ protected Object buildJsonObject(ExplainResponse resp) {
 
               @Override
               public void onFailure(Exception e) {
-                channel.sendResponse(
-                    new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, e.getMessage()));
+                RestStatus status =
+                    (e instanceof org.opensearch.OpenSearchException osEx)
+                        ? osEx.status()
+                        : RestStatus.INTERNAL_SERVER_ERROR;
```
dai-chen: Could you reuse this https://github.com/opensearch-project/sql/blob/main/legacy/src/main/java/org/opensearch/sql/legacy/plugin/RestSqlAction.java#L213?

dai-chen: @ahkcs Could you help review and check if we publish metrics from this handler. I recall we've added it. Ref: https://github.com/opensearch-project/sql/blob/main/legacy/src/main/java/org/opensearch/sql/legacy/plugin/RestSqlAction.java#L209

ahkcs: Checked, it seems like no metrics published here for analytics path, `SQLPlugin.createSqlAnalyticsRouter` bypasses `RestSqlAction`, so `logAndPublishMetrics` never runs for analytics path, should we do a follow-up for analytics-routed SQL query failures?

dai-chen: Sure, we can prioritize in follow up PR. Because this handler is in SQLPlugin, we need to differentiate SQL/PPL metrics name.

ahkcs (approved):

dai-chen (approved):
Thanks for the fix!

PR #5475, opened 2026-05-27
Support SQL query profiling with the analytics engine

dai-chen:
### Description

Adds a `profile` boolean field to the `/_plugins/_sql` request body to capture per-stage query timings, mirroring PPL's existing profile pattern. Also propagates the profile context across the Analytics Engine's async listener thread so profile output is visible in the response.

- **Enabled** when `"profile": true` is set on a SQL query that routes to the analytics engine.
- **Ignored for V2 SQL**: queries handled by the V2 SQL engine silently drop the flag.

### Related Issues
Part of https://github.com/opensearch-project/sql/issues/5246.

### Check List

### DIFF
--- a/docs/user/interfaces/endpoint.rst
+++ b/docs/user/interfaces/endpoint.rst
@@ -266,6 +266,58 @@ Result set::
       "status": 200
     }
 
+Profile [Experimental]
+======================
+
+Description
+-----------
+
+Profiling captures per-stage timings (in milliseconds) for SQL query
+execution. To enable profiling, set ``"profile": true`` in the request
+body alongside ``"query"``.
+
+.. note::
+   The ``profile`` parameter only takes effect when the query runs on
+   the Analytics Engine. In all other cases the flag is silently ignored.
+
+   Profile output is returned only for regular query execution (not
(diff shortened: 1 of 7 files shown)


mengweieric (approved):

docs/user/interfaces/endpoint.rst
```diff
@@ -266,6 +266,58 @@ Result set::
...
+
+.. note::
+   The ``profile`` parameter only takes effect when the query runs on
+   the Analytics Engine. In all other cases the flag is silently ignored.
+
+   Profile output is returned only for regular query execution (not
+   ``_explain``) and only with the default ``format=jdbc``.
+
+Example
+-------
```
ahkcs, line 287: Is this example an integration test? I am wondering if we can have an integration test for profiling(assert each metric to be greater than 0)?

penghuo (approved):

docs/user/interfaces/endpoint.rst
dai-chen: I did this manually against OS cluster with SQL + AE. I'm also thinking can we add some dedicated IT for AE. Let me check after this.

PR #5479, opened 2026-05-28
Fix transpose VARCHAR length drift in unpivot/pivot lowering

songkant-aws:
## Summary

PPL `transpose` lowers via `RelBuilder.unpivot() + pivot()`. Calcite's `unpivot` synthesizes a `LogicalValues` leaf carrying axis literals (the input field names), and types them as `CHAR(maxAxisLiteralLen)` — the longest literal wins on column-level type inference. That precision propagates through the rest of the lowering and breaks the analytics-engine route end-to-end.

This PR makes the lowering own its type contract:

1. **Build every axis literal at the same `CHAR(maxAxisLiteralLen)` type.** Calcite then space-pads the shorter literals at value-construction time, so the runtime CHAR vector and the declared schema both have the same fixed length. The downstream TRIM strips the padding.
2. **Wrap the trimmed-axis group key in an explicit `CAST(... AS VARCHAR)` to unbounded VARCHAR.** This makes the group key type match `_value_transpose_`'s unbounded VARCHAR end-to-end, so the aggregate's row-type check sees consistent types.

## Why this has to be in sql plugin (not analytics-engine planner)

The typing decisions happen inside Calcite's `RelBuilder.unpivot()` when it constructs the VALUES leaf — long before any analytics-engine rule sees the plan. By the time the plan reaches the analytics-engine route, the precision drift is already baked into the `RelDataType` chain. Fixing it downstream would require pattern-matching on transpose-shaped sub-trees inside the planner, which is fragile and mis-attributes the root cause. **The lowering author owns the type contract for the operators it emits.**

## What goes wrong on the analytics-engine route without this fix

Concrete plan from `source=accounts | head N | fields firstname, age, balance | transpose`:

```
LogicalAggregate(group=[{1}], row 1_null=[MAX($0) FILTER $2], ...)
  LogicalProject(_value_transpose_=[CAST($5):VARCHAR NOT NULL],   -- $0
                 $f6=[TRIM(BOTH, ' ', $4)],                       -- $1  ← group key
                 $f7=[=($3, 1)], ...)                             -- $2..
    LogicalProject(... column=$4, _value_transpose_=[CASE(...)])
      LogicalJoin(condition=[true])                               -- cross join from pivot
        LogicalProject(..., ROW_NUMBER() OVER ())
          LogicalSort(fetch=N)
            LogicalScan(t1)
        LogicalValues(tuples=[['firstname'},{'age'},{'balance'}]) -- axis literals
```

Type chain:
- `column` (group key): `CHAR(9)` from longest literal `'firstname'` → after TRIM → `VARCHAR(9)`
- `_value_transpose_`: `CAST(... AS VARCHAR)` — unbounded
- `MAX(_value_transpose_).type`: unbounded VARCHAR (inferred from arg 0 at construction)

Two failure modes downstream:

1. **`Aggregate.<init>` rejects the plan.** Non-prefix `groupSet={1}` aggregates split into PARTIAL+FINAL on the analytics path. PARTIAL hoists group keys to its output prefix, so FINAL's `argList=[0]` ends up reading the group-key slot (`VARCHAR(9)`) instead of the agg-state slot. Calcite's `typeMatchesInferred` check then fires:
   ```
   type mismatch:
     aggCall type:    VARCHAR
     inferred type:   VARCHAR(9)
   ```
2. **Substrait/Arrow `FixedChar` length mismatch at execution time.** Even if the aggregate validation passes, the `CHAR(maxAxisLiteralLen)` propagates through isthmus → DataFusion as a `FixedChar(maxAxisLiteralLen)` schema. Runtime arrays carry the actual values whose lengths are shorter (e.g. `"age"` is length 3, not 9), so DataFusion rejects:
   ```
   Row field type (FixedChar{length=3}) does not match schema field type (FixedChar{length=9})
   ```

Both failures trace back to the same root cause — uneven typing in the unpivot/pivot lowering — and disappear once the lowering emits consistent types.

## What changes

`core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java` — `visitTranspose`:

```diff
+    int axisLiteralLength = fieldNames.stream().mapToInt(String::length).max().orElse(0);
+    RelDataType axisLiteralType =
+        rx.getTypeFactory().createSqlType(SqlTypeName.CHAR, axisLiteralLength);
     // unpivot axisMap entries:
-                        ImmutableList.of(rx.makeLiteral(f)),
+                        ImmutableList.of(
+                            (RexLiteral) rx.makeLiteral(f, axisLiteralType, false, false)),
     // group key for pivot:
-    RexNode trimmedColumnName =
-        context.rexBuilder.makeCall(SqlStdOperatorTable.TRIM, ...);
+    RexNode trimmedColumnName =
+        context.rexBuilder.makeCast(
+            varchar,
+            context.rexBuilder.makeCall(SqlStdOperatorTable.TRIM, ...),
+            true);
```

## Tests

- `CalcitePPLTransposeTest` — added `testTransposeColumnAxisUsesUnboundedVarchar` regression assertion that pins the output `column` field's type to unbounded VARCHAR. This catches any future change that re-introduces axis-literal precision into the group key.
- Updated plan-shape assertions across the existing transpose tests to reflect:
  - Padded axis literals (`'cnt    '`, `'COMM '`, `'JOB  '`, `'SAL  '`, etc.)
  - `CAST(TRIM(...) AS VARCHAR)` shape on the group key
  - SparkSQL output `GROUP BY CAST(TRIM(\`column\`) AS STRING)`

End-to-end verification: `CalciteTransposeCommandIT` 5/5 pass with `-Dtests.analytics.parquet_indices=true`.

## Test plan

- [x] Unit: `./gradlew :ppl:test --tests "*CalcitePPLTransposeTest"`
- [x] Spotless: `./gradlew spotlessCheck`
- [x] V2 engine path (non-parquet): no regression — unchanged plan shape produces identical results
- [x] Analytics-engine path (parquet): `CalciteTransposeCommandIT` 5/5 (vs 0/5 before)


RyanL1997 (approved):

songkant-aws:
Hold on for this PR as the latest datafusion main branch could have other decent fix.

songkant-aws:
Refreshed against latest \`opensearch-project/sql:main\` (\`acd4437c3\`) and re-tested locally; this PR is **still required** for \`CalciteTransposeCommandIT\` to pass on the analytics-engine route. Earlier I had tentatively suggested upstream OpenSearch #21794 (\`Convert varchar to str in analytics engine Project operations\`) might make this redundant — that turned out to be wrong on closer testing.

### Today's test setup
- OpenSearch sandbox: \`transpose-final-agg-fix\` branch (= 21804 head + the \`filterArg=-1\` and \`adaptJoin\` fixes from opensearch-project/OpenSearch#21911), so the marking-phase + filter-arg + join-condition issues are all already resolved upstream of where this PR's lowering change kicks in.
- sql plugin: clean \`main\` (\`acd4437c3\`), **without** the changes in this PR.
- Test: \`./gradlew :integ-test:integTestRemote -Dtests.analytics.parquet_indices=true --tests "*CalciteTransposeCommandIT"\`

### Result without this PR
**0/5 pass; all 5 fail with FixedChar length mismatch:**
\`\`\`
testTranspose                            : Row field type (FixedChar{nullable=false, length=3}) does not match schema field type (FixedChar{nullable=false, length=9})
testTransposeColumnName                  : Row field type (FixedChar{nullable=false, length=3}) does not match schema field type (FixedChar{nullable=false, length=9})
testTransposeLimit                       : Row field type (FixedChar{nullable=false, length=3}) does not match schema field type (FixedChar{nullable=false, length=9})
testTransposeLowerLimit                  : Row field type (FixedChar{nullable=false, length=3}) does not match schema field type (FixedChar{nullable=false, length=9})
testTransposeWithValueFieldNameCollision : Row field type (FixedChar{nullable=false, length=5}) does not match schema field type (FixedChar{nullable=false, length=7})
\`\`\`

This is exactly the error this PR was originally opened to fix:

- length=9 (schema) = max axis literal length, e.g. \`firstname\` (the longest of \`firstname\`/\`age\`/\`balance\` in the test fixture)
- length=3 (data) = the actual short string in the column at runtime, e.g. \`age\`
- Schema is a fixed-length \`CHAR(9)\` because Calcite's \`RelBuilder.unpivot()\` types the synthesized \`LogicalValues\` axis literals as \`CHAR(maxAxisLiteralLen)\` — the longest literal wins on column-level inference. Runtime data carries the actual per-literal length, so DataFusion rejects the schema/data mismatch.

This isn't covered by upstream sandbox fixes. OpenSearch #21794 substituted Utf8 for Utf8View in \`OpenSearchProject\` operations, but the Substrait \`FixedChar(N)\` schema produced by transpose's \`LogicalValues\` axis leaf flows through unchanged into the partial-aggregate StageInputScan and runtime decode.

### What this PR does
Fixes the typing at the lowering site so the schema is consistent with runtime values:

1. Builds every axis literal at the same \`CHAR(maxAxisLiteralLen)\`. Calcite then space-pads shorter literals at value-construction time, so the runtime CHAR vector and the declared schema both have the same fixed length. The downstream TRIM strips the padding.
2. Wraps the trimmed-axis group key in an explicit \`CAST(... AS VARCHAR)\` to unbounded VARCHAR. This makes the group key type match \`_value_transpose_\`'s unbounded VARCHAR end-to-end, so the aggregate's row-type check sees consistent types.

### Verified
With this PR applied (\`./gradlew :integ-test:integTestRemote -Dtests.analytics.parquet_indices=true --tests "*CalciteTransposeCommandIT"\`):
- 5/5 pass, against the same OpenSearch \`transpose-final-agg-fix\` cluster used above.

### Branch state
Force-pushed to rebase onto current \`opensearch-project/sql:main\` (\`acd4437c3\`). PR head is now \`ccb0f7c37\`.

LantaoJin (approved):

PR #5495, opened 2026-06-02
[Enhancement] Push eventstats down by rewriting RexOver to Join + Aggregate (#5483)

### LINKED ISSUE #5483: [BUG] eventstats does not push down to OpenSearch (RexOver excluded from aggregation pushdown)
## Query Information
**PPL Command/Query:**
```ppl
source=logs-* 
| where `@timestamp` > DATE_SUB(NOW(), INTERVAL 5 MINUTE)
| where severityText = 'ERROR'
| eventstats count() as total
| head 10
| fields severityText, total
```

**Expected Result:**
The global `eventstats count()` should push down to OpenSearch as `track_total_hits` (the same way `stats count()` does), so the shard returns one number and the coordinator scales it across documents — instead of streaming every matching document to the coordinator just to count them.

**Actual Result:**
Pushdown does not fire. The scan runs with `requestedTotalSize=2147483647` and `_source` includes — i.e., **every matching document streams to the coordinator**, where a `COUNT() OVER ()` window function counts them in memory. On a 47B-document index this exceeds query timeouts (the reported "FAIL on PI") and on smaller indices it returns 0 rows when the streaming exceeds limits ("0 rows on CAPE/WI"). The query returns correct results on tiny test data but the plan is unviable at production scale.

## Dataset Information
**Dataset/Schema Type**
- [x] OpenTelemetry (OTEL)

**Index Mapping**
```json
{
  "index_patterns": ["logs-*"],
  "template": {
    "mappings": {
      "properties": {
        "@timestamp":   { "type": "date" },
        "body":         { "type": "text", "norms": false },
        "severityText": { "type": "keyword" }
      }
    }
  }
}
```

**Sample Data**
```json
{ "@timestamp": "2026-05-28T17:44:20.000Z", "severityText": "ERROR", "body": "..." }
{ "@timestamp": "2026-05-28T17:44:21.000Z", "severityText": "INFO",  "body": "..." }
{ "@timestamp": "2026-05-28T17:44:22.000Z", "severityText": "ERROR", "body": "..." }
{ "@timestamp": "2026-05-28T17:44:23.000Z", "severityText": "ERROR", "body": "..." }
```

## Bug Description
**Issue Summary:**

`eventstats` (no `BY`, no partition) — equivalent to a global `stats count()` for purposes of the count value — produces a `COUNT() OVER ()` window function instead of a flat aggregation, and the existing pushdown rules deliberately skip it.

**Explain output for both shapes (Calcite engine, pushdown enabled):**

`stats count() as total` (works, pushed down):
```
CalciteEnumerableIndexScan(
  PushDownContext=[[FILTER->..., AGGREGATION->...COUNT(), LIMIT->10000]],
  sourceBuilder={"size":0, ..., "track_total_hits":2147483647})
```
DSL has `"size":0` and `track_total_hits` — the shard returns one number.

`eventstats count() as total` (fails to push down):
```
EnumerableLimit(fetch=[10000])
  EnumerableLimit(fetch=[10])
    EnumerableWindow(window#0=[window(aggs [COUNT()])])
      CalciteEnumerableIndexScan(
        PushDownContext=[[PROJECT->[@timestamp, severityText],
                          FILTER->AND(>(@ts,...), =(severityText,'ERROR')),
                          PROJECT->[severityText]]],
        sourceBuilder={"_source":{"includes":["severityText"]},
                       "query":{"bool":{"must":[{"range":...},{"term":{"severityText":"ERROR"}}]}}},
        requestedTotalSize=2147483647)
```
DSL has `_source` includes and `requestedTotalSize=MAX_INT` — every matching document streams up to the coordinator. `EnumerableWindow` then counts them.

**Root cause:**
PPL `eventstats` is parsed in `AstBuilder.visitEventstatsCommand` (`ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java:498`) as a `Window` node, which Calcite lowers to a `RexOver`. The aggregation pushdown path in `AggregateIndexScanRule` (`opensearch/.../planner/rules/AggregateIndexScanRule.java:122,171`) explicitly excludes `RexOver` from pushdown — windows are treated as opaque coordinator-side computations.

This treatment is correct for windows that need partitioning or ordering. But `eventstats <agg>` without a `BY` clause produces an unpartitioned, unordered window — `OVER ()` — which is semantically equivalent to: "compute a single global aggregate, broadcast it to every input row." The first half of that (the global aggregate) is exactly what OpenSearch can do natively in the shard.

The result is pathological at scale: instead of one round-trip returning a single number, the coordinator pulls every matching document.

**Steps to Reproduce:**
1. Apply the index template above; index a few sample documents.
2. Set `plugins.calcite.enabled: true`, `plugins.calcite.pushdown.enabled: true`.
3. Run `EXPLAIN` on the failing query — observe `EnumerableWindow` and the absence of an `aggregations` block in the DSL.
4. Replace `eventstats` with `stats` and re-run `EXPLAIN` — observe the correct `aggregations` + `track_total_hits` shape.

## Environment Information
**OpenSearch Version:** 3.7.0-SNAPSHOT (reproduced on `main`, commit `ec433f4ff`).

**Additional Details:**
- Calcite engine: `plugins.calcite.enabled = true`
- Pushdown: `plugins.calcite.pushdown.enabled = true`
- Code locations:
  - `ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java:498` — `eventstats` becomes a `Window` node
  - `opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/AggregateIndexScanRule.java:122,171` — pushdown rule excludes `RexOver`

**Possible fix:**
Add a rule that recognizes `RexOver` with empty partition/order/frame as a global aggregate and rewrites it to:
```
Project(input.*, broadcast(globalAgg))
  Aggregate(group=[], aggs=[<agg>])
    <input>
```
The aggregate is then pushable through the existing `AggregateIndexScanRule`. The broadcast step on the coordinator is one extra column join that costs nothing — far cheaper than streaming the entire matched set.

Same fix applies to `eventstats count() by <field>`: `OVER (PARTITION BY <field>)` → `Aggregate(group=[<field>], aggs=[<agg>])` joined back on `<field>`. Both shapes are common in observability queries.

RyanL1997:
### Description

PPL `eventstats` lowers to `LogicalProject(RexOver(...))` directly above the scan. No rule in `OpenSearchIndexRules.OPEN_SEARCH_PUSHDOWN_RULES` matches that shape — every `AggregateIndexScanRule` config requires `LogicalAggregate` at the operand root, and `RareTopPushdownRule` requires a `ROW_NUMBER` window with a `LESS_THAN_OR_EQUAL` filter above. The plan therefore reaches Volcano with `RexOver` intact, becomes `EnumerableWindow`, and the scan beneath it stays in `_source` + `requestedTotalSize=MAX_INT` mode. On 47B-doc indices the coordinator times out streaming every matching document just to count it. Same pathological behavior with `BY` (the production query in the issue).

This PR rewrites `Window` AST nodes in `CalciteRelNodeVisitor.visitWindow` into `Project → Join → (input, Aggregate(input))`. The right-side `Aggregate` sits directly over a re-pushed copy of the input, matching `AggregateIndexScanRule.AGGREGATE_SCAN` (no-`BY`) or `DEFAULT` / `BUCKET_NON_NULL_AGG` (`BY`). OpenSearch sees the same shape as `stats count()` — `size:0 + track_total_hits` or a `terms` aggregation — instead of an unsized row fetch. The left side returns rows as before; the join broadcasts the aggregate value(s) onto each row, preserving the `[original cols, agg cols]` row type so downstream consumers (`limit`, `head`, `fields`) see no change.

### Design

Follows the precedent in `buildStreamWindowSelfJoinPlan` (`CalciteRelNodeVisitor.java:2348-2362`):

- **Join, not `LogicalCorrelate`**, because `LogicalCorrelate` causes NPE in `RelDecorrelator` per the existing comment.
- **NULL bucket handling** mirrors lines 2442-2449 of the streamstats self-join: `bucketNullable=true` → `INNER` join with `IS NOT DISTINCT FROM` (`(left.k = right.k) OR (both NULL)`); `bucketNullable=false` → `LEFT` join with simple equality, `IS NOT NULL` filter pushed below the right aggregate to match `BUCKET_NON_NULL_AGG`. NULL-keyed left rows are preserved with `NULL` aggregate values, matching the previous CASE-wrapped behavior.
- **`aggregateWithTrimming` is reused** for the right-side aggregate construction, so agg-resolution semantics are identical to `stats`/`streamstats`.

Rewriteability predicate (`canRewriteWindowAsAggregateJoin`) rejects: non-aggregate window functions (`ROW_NUMBER` / `LAG` / etc.), non-empty `sortList`, non-default frame, non-bare-field partition keys. Anything outside the eventstats shape falls through to `visitWindowAsRexOver`, preserving existing behavior for any future `Window` producer.

### Coverage against the issue's 7-requirement comment

[Comment](https://github.com/opensearch-project/sql/issues/5483#issuecomment-4595172847):

1. **No-`BY` + `BY` both addressed** in a single rewrite; `hasGroup` toggles cross-join vs equi-join. ✓
2. **All aggregate functions** — uses `aggregateWithTrimming` (same path `stats` uses), supports COUNT/SUM/AVG/MIN/MAX/STDDEV/VAR/etc. uniformly. ✓
3. **Multiple aggregates per call** packed into one right-side `Aggregate` with multiple agg calls, one broadcast join. ✓
4. **NULL semantics on `BY` join-back** — `IS NOT DISTINCT FROM` for `bucketNullable=true`; `LEFT` join + right-side `IS NOT NULL` filter for `bucketNullable=false`. Equivalent to the existing Window+CASE behavior. ✓
5. **Tight predicate** rejects ordered/framed/non-aggregate/non-bare-field cases. ✓
6. **No decorrelation dependency** — rewrite directly produces the operand chain `AggregateIndexScanRule` expects. ✓
7. **No symptom-masking shortcuts** — `Predicate.not(LogicalProject::containsOver)` guards in `AggregateIndexScanRule` are untouched. Real Window→Aggregate rewrite at the AST level. ✓

### Verification — status

**Draft** because the following needs CI to finalize, blocked locally by a pre-existing `core:compileJava` failure on `AnalyticsExecutionEngine.java:92` (unrelated `analytics-api:3.7.0-SNAPSHOT` symbol issue that also reproduces on `main` with no edits):

- `CalcitePPLEventstatsTest` `verifyLogical` expectations have been updated to my pattern-derived prediction of the new lowered shape. May need 1-2 small adjustments (trim-project corner cases) based on actual output. `verifyPPLToSparkSQL` calls are temporarily removed pending observation of the `SparkSqlDialect` output for the join+aggregate form.
- Existing eventstats EXPLAIN expected-output files (`explain_eventstats_avg.json`, `_dc.json`, `_distinct_count.json`, `_earliest_latest.json`, `_earliest_latest_custom_time.json`, `_earliest_latest_no_group.json`, `_null_bucket.yaml`) will diff against the new plan shape. Regeneration from CI logs incoming as a follow-up commit.
- A new positive-pushdown EXPLAIN assertion (`testEventstatsPushdownExplain` mirroring `stats count()` — asserting `AGGREGATION->...COUNT()` in `PushDownContext` and `track_total_hits` in `sourceBuilder`) incoming.
- NULL-bucket `BY` integration tests in `CalcitePPLEventstatsIT` against `TEST_INDEX_STATE_COUNTRY_WITH_NULL` (both `bucketNullable=true` and `=false`) incoming.

Adjacent unit suites (`CalcitePPLAggregationTest`, `CalcitePPLStreamstatsTest`, `CalcitePPLRareTest`, `CalcitePPLTopTest`) should all be unchanged — different visitor paths or `StreamWindow`/`RareTopN` AST nodes.

### Related Issues

Resolves opensearch-project/sql#5483

### Check List

- [x] Commits are signed per the DCO using \`--signoff\` or \`-s\`.


RyanL1997:
## Perf finding & fix on the no-BY rewrite

An internal review flagged a concern about the cost of the two-scan + coordinator-join shape vs the original window form. Ran a local A/B against a 20k-doc accounts index to put numbers on it.

### Initial measurement uncovered a real problem

| Scenario | Before rewrite (legacy window) | With rewrite (commit before this fix) |
|---|---|---|
| `... | head 10` (10-row result) | ~234ms, ~13 OS calls | **~132ms, ~13 OS calls** ✓ |
| no `head` (10k-row result) | ~169ms, ~1 OS call | **~1055ms, ~10,004 OS calls** ❌ |

For the unbounded result-set case, the no-BY rewrite was triggering **one OpenSearch search request per left-side row**.

### Root cause

The no-BY case emitted:
```
EnumerableNestedLoopJoin(condition=[true], joinType=[inner])
  leftScan   ← returns N rows
  rightScan  ← returns 1 row (the COUNT() scalar)
```

`EnumerableNestedLoopJoin` calls `Enumerable.enumerator()` on the right side **per left tuple**. Each enumerator open on a `CalciteEnumerableIndexScan` triggers a fresh `_search` request. For a 10k-row left side that means 10k OS calls. On a remote cluster (1–10ms RTT per call) that would blow up to tens of seconds.

The BY case never hit this — it already has an equi-join condition (or `IS NOT DISTINCT FROM` for `bucketNullable=true`) which lets Calcite pick `EnumerableMergeJoin` and drain one side once.

### Fix (commit `6f333fc22`)

For the no-BY case, project a literal-0 key column onto both sides and join on equality:
- Left: append the key after the original columns.
- Right: append the key after the aggregate outputs.
- Join condition: `left.__eventstats_join_key__ = right.__eventstats_join_key__`.

The equi-join condition makes the planner pick **`EnumerableHashJoin`**, which drains the single-row right side once into a hash table and probes per left row in O(1). Final projection strips the two key columns so the output row type is unchanged.

### After-fix measurement (same data, same node, same queries)

| Scenario | After fix |
|---|---|
| `... | head 10` | ~210ms warm, ~13 OS calls |
| no `head` (10k-row result) | **~174ms warm, 4 OS calls across 5 runs** |

That's ~6× faster wall and ~2,500× fewer OS round-trips for the unbounded case. Latency for the head-bounded case ticks up slightly (~30ms) because of the extra projection + hash-join overhead at small scale, but the absolute numbers are well under one round-trip.

### Pushdown verification

Confirmed via EXPLAIN that the right-side aggregate still pushes down with the new shape:

```
EnumerableHashJoin(condition=[=($1, $4)], joinType=[inner])
  EnumerableCalc(... LITERAL(0) as join_key)
    CalciteEnumerableIndexScan(
      PushDownContext=[[AGGREGATION->...COUNT()]],
      sourceBuilder={"size":0,"track_total_hits":2147483647})
  EnumerableCalc(... LITERAL(0) as join_key)
    CalciteEnumerableIndexScan(... full _source path ...)
```

The literal-0 projection wraps the aggregate but doesn't disrupt the `Aggregate → Scan` operand chain that `AggregateIndexScanRule.AGGREGATE_SCAN` matches — pushdown fires normally and the right side still issues exactly one `size:0 + track_total_hits` (or `cardinality` for `dc()`, or `composite + top_hits` for `earliest`/`latest`) request.

### Test impact

Unit-test expectations updated for 2 tests in `CalcitePPLEventstatsTest` and 4 in `CalcitePPLEventstatsEarliestLatestTest` (the no-BY variants now show an outer `LogicalProject` because the literal-0 key columns must be stripped). 3 EXPLAIN expected files regenerated (`explain_eventstats_dc.json` pushdown, `explain_eventstats_earliest_latest_no_group.json` in both modes).

BY-case tests are unchanged — they already use equi-join conditions.

PR #5497, opened 2026-06-02
Add STATS and EXTENDED_STATS compound aggregate expansion

varun-st:
## Description
  
  Adds parser-level expansion for the `STATS` and `EXTENDED_STATS` compound aggregates, matching legacy V1 SQL plugin output. A query like `SELECT STATS(price) FROM orders` now expands to its primitive
  components at AST-building time, so the analyzer/planner sees only standard primitives.
  
  | Compound | Expanded columns |
  |---|---|
  | `STATS(f)` | `count, sum, avg, min, max` |
  | `EXTENDED_STATS(f)` | above + `sumOfSquares, variance, stdDeviation` |
  
  `variance` and `stdDeviation` use population variants (`var_pop`, `stddev_pop`), matching V1. `sumOfSquares` lowers to `SUM(f * f)`.
  
  Display-name format matches V1: `STATS(price).count`, or `p.count` when an explicit `AS p` is given. `FILTER (WHERE …)` propagates to every primitive.

### DIFF
--- a/language-grammar/src/main/antlr4/OpenSearchSQLParser.g4
+++ b/language-grammar/src/main/antlr4/OpenSearchSQLParser.g4
@@ -500,6 +500,8 @@ aggregationFunctionName
    | STDDEV
    | STDDEV_POP
    | STDDEV_SAMP
+   | STATS
+   | EXTENDED_STATS
    ;
 
 mathematicalFunctionName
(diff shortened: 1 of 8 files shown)


ahkcs (changes requested):
Grammar changes shouldn't be made in `language-grammar` directory, that directory is used for shared grammar with spark repo.

dai-chen (commented):
Please fix DCO and add IT to verify. I assume this will enable in SQL V2 engine too (previously query with such function falls back to legacy engine), I recall many legacy IT was ignored after 3.0

PR #5498, opened 2026-06-02
Increment version to 3.8.0-SNAPSHOT

mengweieric:
Increment version to 3.8.0-SNAPSHOT.


ahkcs (approved):

dai-chen (commented):
Should we wait for geospatial and ml-common version bump in CI failure, otherwise main branch CI will be broken?

RyanL1997:
> Should we wait for geospatial and ml-common version bump in CI failure, otherwise main branch CI will be broken?

+1

Swiddis:
https://github.com/opensearch-project/geospatial/pull/866
https://github.com/opensearch-project/ml-commons/pull/4854
https://github.com/opensearch-project/common-utils/pull/969
https://github.com/opensearch-project/opensearch-remote-metadata-sdk/pull/408

why do none of these repos get auto-bumps lol

RyanL1997:
> why do neither of these repos get auto-bumps lol

Cuz we are kinda early still. Even the core version bump PR has been merged into main, it seems like it is not been publish to whatever needs to be published here yet:

https://github.com/opensearch-project/sql/blob/8394e5cedf198d31335502275a907e53a4281504/.github/workflows/sql-test-and-build-workflow.yml#L23

RyanL1997 (approved):

Issue #5505, opened 2026-06-02
[RFC] Better Command-Chain Testing with Index Views


Swiddis:
## Problem Statement
We see a lot of bugs that boil down to "Command A works, command B works, but A + B doesn't work." Recent examples are #5482 and #5483. These show clear testing gaps with combining multiple commands in one query, some of which are easy to catch (e.g. prepending a `where` clause) but not currently tested.

## Current State
We have a lot of integration correctness and performance tests that use specific test indices. These do somewhat thoroughly exercise all individual commands, but these tests rarely involve carefully checking multiple commands. We also don't have great general visibility on whether certain optimization rules are applied, so `explain` tests are locked in scope to only simple cases or hand-checked complex cases.

## Long-Term Goals
* Catch more bugs with common cross-command combinations. For example, I think all commands should work after simple-predicate `where` clauses, up to some pathological counterexample, since `where` is very easy to push down.
* Keep it simple: easy to implement, largely reuse our current testing infrastructure without complex introductions like property-based tests.
* Better model real-world use cases: complex commands are rarely used directly, usually there are some pre-filtering steps.

## Proposal
I propose implementing two general migrations. Both can be incrementally done, and I believe they're mechanical enough to largely delegate to agents:
1. Introduce Views as additional options for test indices
2. Add optimization annotations to `_explain` output

### Test Views

We currently implement all of our integ tests on top of test indices. The largest indices are `TEST_INDEX_ACCOUNT` (accounting for 19.4% of all E2E query tests), followed by `TEST_INDEX_BANK` (15.7%) and `TEST_INDEX_STATE_COUNTRY` (5.5%). This means that, in order to test any given command with another command, we need to manually work out a query that exercises the new command, as well as its output.

Taking the idea of a View in SQL, we can instead use an approach where we store and reuse query parts as test indices.

Suppose we have this index for tests, with many tests relying on it:

```json
{"a": 1, "b": "cat"}
{"a": 2, "b": "dog"}
```

Now let's extend it with some fake additional fields and records:

```json
{"a": 1, "b": "cat", "_is_real": true}
{"a": 2, "b": "dog", "_is_real": true}
{"a": 3, "b": "duck", "_is_real": false}
{"a": 4, "b": "goose", "_is_real": false}
```

And we create a view: `source=extended_idx | where _is_real | fields - _is_real`

[...]

Issue #5482, opened 2026-05-28
[BUG] Dedup pushdown disabled when combined with WHERE clause (FilterMergeRule defeats PPLSimplifyDedupRule)


penghuo:
## Query Information
**PPL Command/Query:**
```ppl
source=logs-* 
| where `@timestamp` > DATE_SUB(NOW(), INTERVAL 5 MINUTE)
| dedup `resource.attributes.k8s.namespace.name`
| fields `resource.attributes.k8s.namespace.name`
```

**Expected Result:**
The dedup should push down to the shard as a `composite` aggregation + `top_hits` (the same pushdown that already works without a `where` clause), returning one row per distinct namespace value.

**Actual Result:**
With **any** `where` clause upstream, dedup pushdown is disabled. The scan returns every matching document to the coordinator and dedup runs in memory via a `ROW_NUMBER` window. On large indices (47B+ docs reported by users) this exceeds timeouts or triggers shard failures.

The same query **without** `where` pushes down correctly.

## Dataset Information
**Dataset/Schema Type**
- [x] OpenTelemetry (OTEL)

**Index Mapping**
```json
{
  "index_patterns": ["logs-*"],
  "template": {
    "mappings": {
      "dynamic_templates": [
        { "resource_attributes": { "path_match": "resource.attributes.*",
                                   "mapping": {"type": "keyword"},
                                   "match_mapping_type": "string" } }
      ],
      "properties": {
        "@timestamp":     { "type": "date" },
        "body":           { "type": "text", "norms": false },
        "severityText":   { "type": "keyword" }
      }
    }
  }
}
```

**Sample Data**
```json
{ "@timestamp": "2026-05-28T16:18:40.000Z", "severityText": "INFO",
  "resource": {"attributes": {"k8s.namespace.name": "kube-system"}}, "body": "..." }
{ "@timestamp": "2026-05-28T16:18:41.000Z", "severityText": "INFO",
  "resource": {"attributes": {"k8s.namespace.name": "default"}},     "body": "..." }
{ "@timestamp": "2026-05-28T16:18:42.000Z", "severityText": "INFO",
  "resource": {"attributes": {"k8s.namespace.name": "monitoring"}},  "body": "..." }
{ "@timestamp": "2026-05-28T16:18:43.000Z", "severityText": "INFO",
  "resource": {"attributes": {"k8s.namespace.name": "kube-system"}}, "body": "..." }
```

## Bug Description
**Issue Summary:**

[...]

PR #5508, opened 2026-06-03
feat(test): add WHERE-prefix view infrastructure for cross-command testing (#5505)

gingeekrishna:
Part of #5505

## What does this PR do?

Implements phase 1 of the RFC: foundational infrastructure for WHERE-prefix view testing, with a proof-of-concept parametrization of `FieldsCommandIT`.

The core idea: extend the top test indices with a `_is_real` boolean flag, then run the same test assertions through:
```
source=<extended_index> | where _is_real | fields - _is_real | <original query>
```
This catches bugs where commands fail only when preceded by a `where` filter — a gap identified in issues like #5482 and #5483.

### New extended test datasets
- `accounts_extended.json` — 1000 real account records (`_is_real: true`) + 3 synthetic records (`_is_real: false`)
- `bank_extended.json` — 7 real bank records (`_is_real: true`) + 3 synthetic records (`_is_real: false`)
- Corresponding index mappings with `_is_real: boolean`

### Infrastructure
| File | Change |
|------|--------|
| `TestsConstants.java` | `TEST_INDEX_ACCOUNT_EXTENDED`, `TEST_INDEX_BANK_EXTENDED` constants |
| `TestUtils.java` | `getAccountExtendedIndexMapping()`, `getBankExtendedIndexMapping()` helpers |
| `SQLIntegTestCase.Index` | `ACCOUNT_EXTENDED`, `BANK_EXTENDED` enum entries |
| `PPLIntegTestCase.java` | `sourceView(extendedIndex, query)` helper method |

### Proof-of-concept: `FieldsCommandIT`
Three tests converted to `@ParameterizedTest` running on both the direct index and the WHERE-prefix view:
- `testBasicFieldSelection` (schema assertions)
- `testMultipleFieldSelection` (row assertions with `head 3`)
- `testSpecialDataTypes` (all 7 bank birthdate rows)

### Why `fields - fieldname` works without Calcite
`hasEnhancedFieldFeatures()` in `AstBuilder.java` gates only wildcards and space-delimited fields — not the MINUS/exclusion form — so the view chain is safe in the default (non-Calcite) test mode.

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/legacy/SQLIntegTestCase.java
+++ b/integ-test/src/test/java/org/opensearch/sql/legacy/SQLIntegTestCase.java
@@ -535,7 +535,7 @@ public abstract class SQLIntegTestCase extends OpenSearchSQLRestTestCase {
         "src/test/resources/accounts.json"),
     ACCOUNT_EXTENDED(
         TestsConstants.TEST_INDEX_ACCOUNT_EXTENDED,
-        "account",
+        "account_extended",
         getAccountExtendedIndexMapping(),
         "src/test/resources/accounts_extended.json"),
     PHRASE(
@@ -633,7 +633,7 @@ public abstract class SQLIntegTestCase extends OpenSearchSQLRestTestCase {
         "src/test/resources/bank.json"),
     BANK_EXTENDED(
         TestsConstants.TEST_INDEX_BANK_EXTENDED,
-        "account",
+        "bank_extended",
         getBankExtendedIndexMapping(),
         "src/test/resources/bank_extended.json"),
     BANK_TWO(
--- a/integ-test/src/test/java/org/opensearch/sql/legacy/SQLIntegTestCase.java
+++ b/integ-test/src/test/java/org/opensearch/sql/legacy/SQLIntegTestCase.java
@@ -533,6 +533,11 @@ public abstract class SQLIntegTestCase extends OpenSearchSQLRestTestCase {
         "account",
         getAccountIndexMapping(),
         "src/test/resources/accounts.json"),
+    ACCOUNT_EXTENDED(
+        TestsConstants.TEST_INDEX_ACCOUNT_EXTENDED,
+        "account",
+        getAccountExtendedIndexMapping(),
+        "src/test/resources/accounts_extended.json"),
     PHRASE(
         TestsConstants.TEST_INDEX_PHRASE,
         "phrase",
@@ -626,6 +631,11 @@ public abstract class SQLIntegTestCase extends OpenSearchSQLRestTestCase {
         "account",
         getBankIndexMapping(),
         "src/test/resources/bank.json"),
+    BANK_EXTENDED(
+        TestsConstants.TEST_INDEX_BANK_EXTENDED,
+        "account",
+        getBankExtendedIndexMapping(),
+        "src/test/resources/bank_extended.json"),
     BANK_TWO(
         TestsConstants.TEST_INDEX_BANK_TWO,
         "account_two",
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/FieldsCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/FieldsCommandIT.java
@@ -38,18 +38,12 @@ public class FieldsCommandIT extends PPLIntegTestCase {
 
   // --- Parameterized sources ---
 
-  /**
-   * Provides a direct account index source and a WHERE-prefix view source that produces the same
-   * rows. Shared by all account-based parameterized tests.
-   */
   static Stream<String> accountIndexSources() {
-    return Stream.of(
-        String.format("source=%s", TEST_INDEX_ACCOUNT), sourceView(TEST_INDEX_ACCOUNT_EXTENDED));
+    return sourceViews(TEST_INDEX_ACCOUNT);
   }
 
   static Stream<String> bankIndexSources() {
-    return Stream.of(
-        String.format("source=%s", TEST_INDEX_BANK), sourceView(TEST_INDEX_BANK_EXTENDED));
+    return sourceViews(TEST_INDEX_BANK);
   }
 
   // --- Tests ---
--- a/integ-test/src/test/java/org/opensearch/sql/ppl/FieldsCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/ppl/FieldsCommandIT.java
@@ -16,9 +16,13 @@ import static org.opensearch.sql.util.MatcherUtils.verifyErrorMessageContains;
 import static org.opensearch.sql.util.MatcherUtils.verifySchema;
 
 import java.io.IOException;
+import java.util.stream.Stream;
 import org.json.JSONObject;
 import org.junit.Ignore;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 public class FieldsCommandIT extends PPLIntegTestCase {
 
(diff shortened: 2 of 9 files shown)


integ-test/src/test/resources/bank_extended.json
```diff
@@ -0,0 +1,20 @@
+{"index":{"_id":"1"}}
+{"account_number":1,"balance":39225,"firstname":"Amber JOHnny","lastname":"Duke Willmington","age":32,"gender":"M","address":"880 Holmes Lane","employer":"Pyrami","email":"amberduke@pyrami.com","city":"Brogan","state":"IL","male":true,"birthdate":"2017-10-23","_is_real":true}
+{"index":{"_id":"6"}}
+{"account_number":6,"balance":5686,"firstname":"Hattie","lastname":"Bond","age":36,"gender":"M","address":"671 Bristol Street","employer":"Netagy","email":"hattiebond@netagy.com","city":"Dante","state":"TN","male":true,"birthdate":"2017-11-20","_is_real":true}
+{"index":{"_id":"13"}}
+{"account_number":13,"balance":32838,"firstname":"Nanette","lastname":"Bates","age":28,"gender":"F","address":"789 Madison Street","employer":"Quility","email":"nanettebates@quility.com","city":"Nogal","state":"VA","male":false,"birthdate":"2018-06-23","_is_real":true}
+{"index":{"_id":"18"}}
+{"account_number":18,"balance":4180,"firstname":"Dale","lastname":"Adams","age":33,"gender":"M","address":"467 Hutchinson Court","employer":"Boink","email":"daleadams@boink.com","city":"Orick","state":"MD","male":true,"birthdate":1542152000000,"_is_real":true}
```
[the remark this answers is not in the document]
gingeekrishna: The epoch-millis value for `_id:18` (`1542152000000`) is intentionally carried over from the original `bank.json` — that record has always stored `birthdate` as epoch millis in the source data, and the test already asserts `rows("2018-11-13 23:33:20")` which confirms OpenSearch normalises it correctly to a timestamp string regardless of input format. Changing it to a date string would diverge from the baseline dataset and could mask issues where the date mapping behaves differently for the two formats.

integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
```diff
@@ -148,6 +148,28 @@ protected static String source(String index, String query) {
...
+   */
+  protected static String sourceView(String extendedIndex) {
+    return String.format("source=%s | where _is_real | fields - _is_real", extendedIndex);
+  }
+
+  /**
+   * Convenience overload that appends {@code query} to the WHERE-prefix view chain. Equivalent to
+   * {@code sourceView(extendedIndex) + " | " + query}.
+   */
+  protected static String sourceView(String extendedIndex, String query) {
```
Swiddis: we don't actually use this, right? looks like we just concat everything in the tests

integ-test/src/test/java/org/opensearch/sql/ppl/FieldsCommandIT.java
```diff
@@ -16,35 +16,56 @@
...
-        executeQuery(String.format("source=%s | fields firstname, lastname", TEST_INDEX_ACCOUNT));
+  // --- Parameterized sources ---
+
+  /**
+   * Provides a direct account index source and a WHERE-prefix view source that produces the same
+   * rows. Shared by all account-based parameterized tests.
+   */
+  static Stream<String> accountIndexSources() {
+    return Stream.of(
+        String.format("source=%s", TEST_INDEX_ACCOUNT), sourceView(TEST_INDEX_ACCOUNT_EXTENDED));
```
Swiddis: I think this stream should be defined in a shared source somewhere, I don't want to need to change every onboarded test file when we add more views later.

This per-suite view is better if we end up in a situation where some tests support different parametrizations than other tests, but I think the ideal state we're aiming for is that all tests support all parametrizations, and "not parametrized" is the fallback state on the way there.

integ-test/src/test/java/org/opensearch/sql/legacy/SQLIntegTestCase.java
```diff
@@ -533,6 +533,11 @@ public enum Index {
         "account",
         getAccountIndexMapping(),
         "src/test/resources/accounts.json"),
+    ACCOUNT_EXTENDED(
+        TestsConstants.TEST_INDEX_ACCOUNT_EXTENDED,
+        "account",
```
Swiddis: it might be worth changing the index name to avoid collisions?

Swiddis (commented):
nice!

integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
```diff
@@ -148,6 +148,28 @@ protected static String source(String index, String query) {
...
+   * Returns a WHERE-prefix view source string over an extended index. The extended index is
+   * expected to contain all rows from the base index with {@code "_is_real":true}, plus additional
+   * synthetic rows with {@code "_is_real":false}. Use this as a drop-in replacement for {@code
+   * "source=<index>"} in parameterized tests to verify commands work correctly when preceded by a
+   * {@code where} filter.
+   *
+   * <p>The returned string ends at {@code fields - _is_real} with no trailing pipe, so callers can
+   * append additional PPL commands with {@code " | <command>"}.
+   */
+  protected static String sourceView(String extendedIndex) {
```
Swiddis, line 170: In line with the "centralize stream logic" comment, we could change the overall design of this to instead return a `List<String>` or `Stream<String>`, which makes it easy to add new views later without changing tests.

So the parametrization setup for each test just looks like `return sourceViews(index)` and this expands out into all of `source=index`, `source=index_extended | where _is_real | fields - _is_real`, ...

Pseudocode:
```java
protected static Stream<String> sourceViews(String index) {
  switch (index) {
    case ACCOUNTS -> Stream.of("source=accounts", "source=accounts_extended | ...");
    case BANK -> Stream.of("source=bank", "source=bank_extended | ...");
    default -> throw new IllegalArgumentException("views not supported for %s", index);
  }
}
```

gingeekrishna:
Thanks for the review @Swiddis! Addressed all four comments in the latest commit:

- **Removed** the unused `sourceView(extendedIndex, query)` overload
- **Centralized** stream generation into `PPLIntegTestCase.sourceViews(baseIndex)` backed by an `INDEX_VIEWS` map — adding new views in future only requires updating that one map
- **Updated** `FieldsCommandIT` source methods to delegate to `sourceViews()`
- **Fixed** `ACCOUNT_EXTENDED` and `BANK_EXTENDED` type names (`"account_extended"` and `"bank_extended"` respectively) to avoid collision with existing entries

integ-test/src/test/java/org/opensearch/sql/ppl/PPLIntegTestCase.java
gingeekrishna: You're right, it wasn't being used anywhere — removed it in the latest commit.

integ-test/src/test/java/org/opensearch/sql/ppl/FieldsCommandIT.java
gingeekrishna: Agreed on the design goal. Moved the stream generation into `PPLIntegTestCase.sourceViews(baseIndex)`, backed by an `INDEX_VIEWS` map. Each test file's source method is now just a one-liner delegating to `sourceViews()`, so adding new views in future only requires updating the map in one place

integ-test/src/test/java/org/opensearch/sql/legacy/SQLIntegTestCase.java
gingeekrishna: Good catch — both ACCOUNT_EXTENDED and BANK_EXTENDED were incorrectly using "account" as the type. Updated them to "account_extended" and "bank_extended" respectively in the latest commit.

Swiddis (approved):
needs dco

gingeekrishna:
> needs dco

Hi @Swiddis ,

I just fixed the DCO

PR #5514, opened 2026-06-04
Add SQL histogram and date_histogram bucket functions

varun-st:
## Summary
  
  Adds parse-time support for `histogram` and `date_histogram` in V2 SQL with named-argument invocation. Each call lowers to existing AST primitives (`Span`, `COALESCE`, `DATE_FORMAT`, `TIMESTAMPADD`).
  
  ## Supported parameters
  
  - `histogram`: `field`, `interval`, `offset`, `missing`
  - `date_histogram`: `field`, `interval` / `fixed_interval` / `calendar_interval`, `format`, `time_zone`, `missing`
  
  ## Deferred to follow-up PR
  
  - `min_doc_count`, `order`, `alias` — need parser-side plumbing to mutate the surrounding query (HAVING / ORDER BY / SELECT-list alias).
  - `date_histogram` `offset` — needs a duration-string parser distinct from `time_zone`'s `ZoneOffset` format.

### DIFF
--- a/language-grammar/src/main/antlr4/OpenSearchSQLParser.g4
+++ b/language-grammar/src/main/antlr4/OpenSearchSQLParser.g4
@@ -411,6 +411,12 @@ scalarFunctionName
    | flowControlFunctionName
    | systemFunctionName
    | nestedFunctionName
+   | bucketFunctionName
+   ;
+
+bucketFunctionName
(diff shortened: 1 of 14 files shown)


RyanL1997:
closing this due to the merge of #5700

PR #5515, opened 2026-06-04
[BugFix] Key date_time pushdown on field type, not literal UDT (#5481)

### LINKED ISSUE #5481: [BUG] Timestamp range pushdown emits unparseable date format when AND'd with IN/OR clause
## Query Information
**PPL Command/Query:**
```ppl
source=logs-* 
| where `@timestamp` > DATE_SUB(NOW(), INTERVAL 5 MINUTE)
| where severityText in ('ERROR', 'WARN') 
| head 10
```

**Expected Result:**
The query should return up to 10 documents from the last 5 minutes whose `severityText` is `ERROR` or `WARN`.

**Actual Result:**
HTTP 500. The shard fails to parse the date in the pushed-down `range` clause.

```
SearchPhaseExecutionException: Failed to execute phase [query], all shards failed
  shardFailures {[...][logs-000001][0]:
    OpenSearchParseException[failed to parse date field [2026-05-28 16:18:43]
      with format [strict_date_optional_time||epoch_millis]];
    IllegalArgumentException[failed to parse date field [2026-05-28 16:18:43]
      with format [strict_date_optional_time||epoch_millis]];
    DateTimeParseException[Failed to parse with all enclosed parsers]; }
```

The same query without the `IN` clause succeeds:
```ppl
source=logs-* | where `@timestamp` > DATE_SUB(NOW(), INTERVAL 5 MINUTE) | head 10
```
Likewise replacing `IN (...)` with a single equality (`severityText != 'INFO'`) succeeds. The bug only triggers when the timestamp range is AND'd with an `IN` or chained `OR` of equalities.

## Dataset Information
**Dataset/Schema Type**
- [x] OpenTelemetry (OTEL)

**Index Mapping**
```json
{
  "index_patterns": ["logs-*"],
  "template": {
    "mappings": {
      "properties": {
        "@timestamp":     { "type": "date" },
        "body":           { "type": "text", "norms": false },
        "severityText":   { "type": "keyword" },
        "severityNumber": { "type": "integer" },
        "traceId":        { "type": "keyword" },
        "spanId":         { "type": "keyword" }
      }
    }
  }
}
```
The `@timestamp` field uses the default `strict_date_optional_time||epoch_millis` parser.

**Sample Data**
```json
{ "@timestamp": "2026-05-28T16:18:40.000Z", "severityText": "ERROR", "body": "Error finding unassigned IPs for ENI xyz" }
{ "@timestamp": "2026-05-28T16:18:41.000Z", "severityText": "WARN",  "body": "high latency observed during HTTP request" }
{ "@timestamp": "2026-05-28T16:18:42.000Z", "severityText": "INFO",  "body": "request completed HTTP/1.1 200 in 24ms" }
```

## Bug Description
**Issue Summary:**

DSL emitted for the failing case (note the timestamp value and missing `format` field):
```json
{"bool":{"must":[
  {"range":{"@timestamp":{"from":"2026-05-28 16:18:43","include_lower":false,"include_upper":true,"boost":1.0}}},
  {"terms":{"severityText":["ERROR","WARN"]}}
]}}
```
DSL emitted for the working case (no `IN`):
```json
{"range":{"@timestamp":{"from":"2026-05-28T16:18:43.000Z","include_lower":false,"include_upper":true,"format":"date_time","boost":1.0}}}
```

**Root cause:**
With Calcite pushdown enabled, `severityText IN ('ERROR','WARN')` is folded into a `Sarg` literal during `RexSimplify`. Because the merged `AND` is simplified as a whole, `RexSimplify` re-types sibling literals — the constant-folded result of `DATE_SUB(NOW(), INTERVAL 5 MINUTE)` loses its `EXPR_TIMESTAMP` UDT and is emitted as `:VARCHAR` (visible in `PushDownContext` as `'2026-05-28 16:18:43':VARCHAR`). Downstream, `PredicateAnalyzer.LiteralExpression.isDateTime()` checks for `ExprUDT.EXPR_TIMESTAMP` on the literal, sees `VARCHAR`, returns false. As a result, `SimpleQueryExpression.gt()` skips both `addFormatIfNecessary` (no `.format("date_time")`) and `timestampValueForPushDown` (no ISO-8601 normalization). The raw `2026-05-28 16:18:43` ships to the shard, where the default date parser rejects it. A defensive fix is to key the format/encoding decision on the field's type (already known to be `EXPR_TIMESTAMP`) rather than the literal's surviving UDT.

**Steps to Reproduce:**
1. Apply the index template above; index a few sample documents into `logs-pr172502-000001`.
2. Set `plugins.calcite.enabled: true` and `plugins.calcite.pushdown.enabled: true`.
3. Run the failing PPL query above.
4. Observe HTTP 500 with `SearchPhaseExecutionException` / `DateTimeParseException`.
5. Run the same query with the `IN` clause removed → succeeds. Run with `severityText != 'INFO'` → succeeds.

## Environment Information
**OpenSearch Version:** 3.7.0-SNAPSHOT (reproduced locally on `main`, commit `ec433f4ff`).

RyanL1997:
### Description

When a timestamp comparison is AND'd with a SARG-eligible sibling (`IN`, chained `OR` of equalities), Calcite's `RexSimplify` folds the sibling into a `Sarg` and **re-types the timestamp literal** during simplification. `DATE_SUB(NOW(), INTERVAL 5 MINUTE)` loses its `EXPR_TIMESTAMP` UDT and arrives at the predicate analyzer as `VARCHAR`.

That breaks the DSL emission downstream in `SimpleQueryExpression`:

1. `LiteralExpression.value()` takes the `isString()` branch and returns the raw `RexLiteral.stringValue()` — the space-separated `"2026-05-28 16:18:43"` form, never normalized to ISO-8601.
2. `addFormatIfNecessary` skips `.format("date_time")` because `literal.isDateTime()` reads `false`.

Both checks key off the **literal's** surviving UDT, which is unreliable after `RexSimplify`. The shard receives `"2026-05-28 16:18:43"` against the default `strict_date_optional_time||epoch_millis` parser and returns HTTP 500.

### Fix

The five non-Sarg comparison paths in `SimpleQueryExpression` (`equals`, `notEquals`, `gt`, `gte`, `lt`, `lte`) now key off `rel.isTimeStampType()` — the **field's** type via `NamedFieldExpression`. This is the same defensive pattern the Sarg path in `constructQueryExpressionForSearch` (PredicateAnalyzer.java:794-796) already uses.

When the field is a timestamp:
- The endpoint value is routed through `timestampValueForPushDown` to canonicalize to ISO-8601 regardless of the literal's surviving type.
- The range query always carries `format("date_time")`.

The non-timestamp-field branch is unchanged — `termQuery` / `boolQuery` behavior for keyword/text/numeric fields is preserved exactly.

### Verification

End-to-end on a live local node with the index template and sample data from the issue:

**Before:**
```
"range":{"@timestamp":{"from":"2026-06-04 16:53:50",...}}    // no format, raw string
→ failed to parse date field [2026-06-04 16:53:50] with format [strict_date_optional_time||epoch_millis]
→ HTTP 500
```

**After:**
```
"range":{"@timestamp":{"from":"2026-06-04T17:14:01.000Z",..., "format":"date_time", ...}}
→ STATUS: OK — 2 rows returned (ERROR + WARN; INFO filtered out)
```

`:opensearch:test` passes with no changes to existing assertions; the existing `date_time` format assertions in `PredicateAnalyzerTest.java` continue to hold for the literal-typed-correctly path.

### Related Issues

Resolves opensearch-project/sql#5481

### Check List

### DIFF
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
@@ -1601,12 +1601,12 @@ public class PredicateAnalyzer {
     }
 
     private Object convertEndpointValue(Object value, boolean isTimeStamp) {
-      value = sargPointValue(value);
-      // Shared normalization entry point: a Sarg endpoint can normalize to null, and
-      // timestampValueForPushDown(value.toString()) would otherwise NPE.
+      // Shared normalization entry point: guard a null endpoint so the timestamp branch's
+      // value.toString() cannot NPE. sargPointValue never produces null from a non-null input.
       if (value == null) {
         return null;
       }
+      value = sargPointValue(value);
       return isTimeStamp ? timestampValueForPushDown(value.toString()) : value;
     }
   }
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
@@ -1378,10 +1378,10 @@ public class PredicateAnalyzer {
 
     @Override
     public QueryExpression equals(LiteralExpression literal) {
-      Object value = literal.value();
-      if (literal.isDateTime()) {
-        builder =
-            addFormatIfNecessary(literal, rangeQuery(getFieldReference()).gte(value).lte(value));
+      boolean isTimeStamp = isFieldOrLiteralDateTime(literal);
+      Object value = endpointValue(literal, isTimeStamp);
+      if (isTimeStamp) {
+        builder = rangeQuery(getFieldReference()).gte(value).lte(value).format("date_time");
       } else {
         builder = termQuery(getFieldReferenceForTermQuery(), value);
       }
@@ -1390,12 +1390,13 @@ public class PredicateAnalyzer {
 
     @Override
     public QueryExpression notEquals(LiteralExpression literal) {
-      Object value = literal.value();
-      if (literal.isDateTime()) {
+      boolean isTimeStamp = isFieldOrLiteralDateTime(literal);
+      Object value = endpointValue(literal, isTimeStamp);
+      if (isTimeStamp) {
         builder =
             boolQuery()
-                .should(addFormatIfNecessary(literal, rangeQuery(getFieldReference()).gt(value)))
-                .should(addFormatIfNecessary(literal, rangeQuery(getFieldReference()).lt(value)));
+                .should(rangeQuery(getFieldReference()).gt(value).format("date_time"))
+                .should(rangeQuery(getFieldReference()).lt(value).format("date_time"));
       } else {
         builder =
             boolQuery()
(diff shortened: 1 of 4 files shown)


opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1408,32 +1409,57 @@ public QueryExpression notEquals(LiteralExpression literal) {
...
+    private boolean isFieldOrLiteralDateTime(LiteralExpression literal) {
+      return literal.isDateTime() || (rel != null && rel.isTimeStampType());
+    }
+
+    private Object endpointValue(LiteralExpression literal, boolean isTimeStamp) {
+      if (!isTimeStamp || literal.isDateTime()) {
+        return literal.value();
+      }
+      return timestampValueForPushDown(literal.value().toString());
+    }
```
penghuo: Can reuse LiteralExperssion.value() method?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1408,32 +1409,57 @@ public QueryExpression notEquals(LiteralExpression literal) {
...
 
     @Override
     public QueryExpression gte(LiteralExpression literal) {
-      Object value = literal.value();
-      builder = addFormatIfNecessary(literal, rangeQuery(getFieldReference()).gte(value));
+      boolean isTimeStamp = isFieldOrLiteralDateTime(literal);
+      Object value = endpointValue(literal, isTimeStamp);
+      RangeQueryBuilder rq = rangeQuery(getFieldReference()).gte(value);
+      if (isTimeStamp) rq.format("date_time");
+      builder = rq;
```
penghuo: The same code duplicated in lt(), lte(). Can reuse addFormatIfNecessary?

penghuo (commented):
Add yamlRestIT

opensearch/src/test/java/org/opensearch/sql/opensearch/request/PredicateAnalyzerTest.java
```diff
@@ -1147,6 +1147,37 @@ void notEquals_generatesBoolQueryForDateTime() throws ExpressionNotAnalyzableExc
...
   }
 
+  /**
+   * RexSimplify can strip the EXPR_TIMESTAMP UDT off a literal when a sibling clause is folded into
+   * a Sarg (e.g. {@code @timestamp > X AND severityText IN (...)}), leaving the literal as plain
+   * VARCHAR. The comparison must still emit a {@code format("date_time")} range query keyed off the
+   * field's type so the shard's default date parser accepts the value.
+   */
+  @Test
+  void gt_normalizesVarcharLiteralAgainstTimestampField() throws ExpressionNotAnalyzableException {
```
dai-chen, line 1157: Besides this UT, could you add a simple IT test to verify end-to-end?

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
RyanL1997: Fixed - now uses `convertEndpointValue(literal.value(), isTimeStamp)`, reusing `value()` plus the shared Sarg-path normalizer; dropped the `endpointValue` helper.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
RyanL1997: I revived `addFormatIfNecessary` (re-keyed on the `isTimeStamp` boolean); all six methods now wrap their range in it instead of inlining `.format("date_time")`.

opensearch/src/test/java/org/opensearch/sql/opensearch/request/PredicateAnalyzerTest.java
RyanL1997: - Added `CalcitePPLBasicIT.testTimestampRangeWithInClausePushDown` — and proved it catches the bug: with the fix neutered, it fails with the original issue shard error (`failed to parse date field [2018-06-01 00:00:00]`); with the fix, it passes. cc @dai-chen 
- Added the yamlRest test issues/5481.yml and confirmed passing. cc @penghuo

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -1581,6 +1602,11 @@ public QueryExpression between(Range<?> range, boolean isTimeStamp) {
 
     private Object convertEndpointValue(Object value, boolean isTimeStamp) {
       value = sargPointValue(value);
+      // Shared normalization entry point: a Sarg endpoint can normalize to null, and
+      // timestampValueForPushDown(value.toString()) would otherwise NPE.
+      if (value == null) {
+        return null;
+      }
```
penghuo: if non-null check is required, it should at beginning

penghuo (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
RyanL1997: Moved to the top — `sargPointValue` never returns null for a non-null input, so guarding the incoming value first fully covers the NPE. Thanks.

dai-chen (approved):
Thanks for the fix!

Issue #4026, opened 2025-08-12
[FEATURE] Multivalue Statistics Functions for PPL Calcite Engine (v3)


ps48:
## Summary

This RFC proposes the addition of two new multivalue statistical functions to the PPL (Piped Processing Language) Calcite engine (v3): `list()` and `values()`. These functions will enhance PPL's analytical capabilities by providing efficient aggregation of multiple values into structured collections, enabling users to perform advanced data analysis and visualization tasks.

## Motivation

Current PPL statistics commands provide scalar aggregation functions (count, sum, avg, min, max, etc.) that reduce multiple values to single results. However, many analytical use cases require collecting and preserving multiple values within groups, such as:

* Building lists of unique values per category for data exploration
* Creating collections of all values for downstream processing
* Generating structured data for visualization and reporting
* Maintaining relationships between grouped data points

The proposed multivalue functions will fill this gap by providing native support for collecting values during aggregation operations.

#### **Purpose**: Collects all values of a field into an ordered array, preserving duplicates and maintaining insertion order.

**Syntax**: `list(field)`
**Parameters**:

* `field`: Expression or field name to collect values from

**Return Type**: `ARRAY` of the input field's data type
**Behavior**:

* Collects all non-null values in the order they are processed
* Preserves duplicate values
* Returns empty array if no non-null values exist
* Null values are ignored (not included in the result)
* If more than 100 values are in a field, only the first 100 are returned.

**Examples**:

```
source=logs | stats list(user_id) by status
source=events | stats list(timestamp) by event_type
source=products | stats count(*), list(category) by brand
```

#### **Purpose**: Collects unique values of a field into an ordered array, removing duplicates while maintaining order of first occurrence.

**Syntax**: `values(field)`
**Parameters**:

* `field`: Expression or field name to collect unique values from

**Return Type**: `ARRAY` of the input field's data type
**Behavior**:

* Collects unique non-null values in order of first occurrence
* Removes duplicate values
* Returns empty array if no non-null values exist
* Null values are ignored (not included in the result)
* By default there is no limit to the number of values returned. In other similar solutions, users with the appropriate permissions can specify a limit in the `limits.conf` file. You specify the limit in the [stats | sistats] stanza using the `maxvalues` setting.

**Examples**:

```
source=logs | stats values(user_id) by status
source=events | stats values(source_ip) by hour
source=products | stats count(*), values(category) by brand
```

#### **Grammar Changes** (`OpenSearchPPLParser.g4`):

```
[...]

PR #5516, opened 2026-06-04
Fix/array type ip binary

vinaykpud:
### Description

Companion fix for the LIST nullability change in opensearch-project/OpenSearch#21997. After that fix, `stats list(<field>)` returns HTTP 200 for 10 of 12 element types — but **`list(ip_value)`** and **`list(binary_value)`** still fail with `HTTP 400 ExpressionEvaluationException: unsupported object class [B`. This PR fixes those two.

### Root cause

Two layers, each correct in isolation:

1. `PPLBuiltinOperators.LIST` was registered with return type `PPLReturnTypes.STRING_ARRAY`, which **always** returns `ARRAY<VARCHAR>` regardless of input element type — type information is lost.
2. The UDT-aware response-boundary dispatch from #5463 (`AnalyticsExecutionEngine.toExprValue`) checks `type instanceof IpType` / `BinaryType` per cell. With the column type collapsed to `ARRAY<VARCHAR>` it never matches, so per-element `byte[]` values fall through to `ExprValueUtils.fromObjectValue`, which has no `byte[]` case and throws.

### Fix

Two small changes:

1. **`PPLBuiltinOperators.LIST`**: switch return-type inference from `STRING_ARRAY` to `ARG0_ARRAY`, so the column type at the response boundary becomes `ARRAY<IpType>` / `ARRAY<BinaryType>` (matching the existing `TAKE` registration).
2. **`AnalyticsExecutionEngine.toExprValue`**: when the column is `ARRAY<IpType>` / `ARRAY<BinaryType>`, recurse element-wise and apply the same canonical-IP-string / Base64 conversion already used for scalar `IpType` / `BinaryType` cells. Extracted the two byte[] converters into helpers (`ipBytesToExprValue`, `binaryBytesToExprValue`) so the scalar and array paths share them.

### Testing

| Query | Before | After |
|---|---|---|
| `stats list(ip_value)` | HTTP 400 `unsupported object class [B` | `[["127.0.0.1"]]` |
| `stats list(binary_value)` | HTTP 400 `unsupported object class [B` | `[["U29tZSBiaW5hcnkgYmxvYg=="]]` |
| `stats list(<other 10 types>)` | HTTP 200 | HTTP 200 (no regression) |
| \`\| fields ip_value, binary_value\` | HTTP 200 (#5463 path) | HTTP 200 (no regression) |

### Out of scope

- `VALUES` aggregate registration still uses `STRING_ARRAY`. Same fix would apply if anyone exercises `values(ip_value)` / `values(binary_value)`; flagging for follow-up.
- `ListAggFunction.java` still does `String.valueOf(byte[])` — latent on the V3 (non-analytics-engine) path. Analytics path uses DataFusion's `LOCAL_ARRAY_AGG_OP` instead, so it doesn't fire here.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
@@ -499,9 +499,10 @@ public class PPLBuiltinOperators extends ReflectiveSqlOperatorTable {
           "pattern",
           ReturnTypes.explicit(UserDefinedFunctionUtils.nullablePatternAggList),
           null);
+  // ARG0_ARRAY preserves UDT element type so AnalyticsExecutionEngine can render IP/binary byte[].
   public static final SqlAggFunction LIST =
       createUserDefinedAggFunction(
-          ListAggFunction.class, "LIST", PPLReturnTypes.STRING_ARRAY, PPLOperandTypes.ANY_SCALAR);
+          ListAggFunction.class, "LIST", PPLReturnTypes.ARG0_ARRAY, PPLOperandTypes.ANY_SCALAR);
   public static final SqlAggFunction VALUES =
       createUserDefinedAggFunction(
           ValuesAggFunction.class,
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultiValueStatsIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultiValueStatsIT.java
@@ -43,7 +43,7 @@ public class CalciteMultiValueStatsIT extends PPLIntegTestCase {
                 "source=%s | stats list(boolean_value) as bool_list",
                 TEST_INDEX_DATATYPE_NONNUMERIC));
     verifySchema(response, schema("bool_list", "array"));
-    verifyDataRows(response, rows(List.of("true")));
+    verifyDataRows(response, rows(List.of(true)));
   }
 
   @Test
@@ -53,7 +53,7 @@ public class CalciteMultiValueStatsIT extends PPLIntegTestCase {
             String.format(
                 "source=%s | stats list(byte_number) as byte_list", TEST_INDEX_DATATYPE_NUMERIC));
     verifySchema(response, schema("byte_list", "array"));
-    verifyDataRows(response, rows(List.of("4")));
+    verifyDataRows(response, rows(List.of(4)));
   }
 
   @Test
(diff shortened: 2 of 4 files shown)


ahkcs:
Please take a look at the CI failures

core/src/main/java/org/opensearch/sql/expression/function/PPLBuiltinOperators.java
```diff
@@ -499,9 +499,10 @@ public class PPLBuiltinOperators extends ReflectiveSqlOperatorTable {
           "pattern",
           ReturnTypes.explicit(UserDefinedFunctionUtils.nullablePatternAggList),
           null);
+  // ARG0_ARRAY preserves UDT element type so AnalyticsExecutionEngine can render IP/binary byte[].
   public static final SqlAggFunction LIST =
       createUserDefinedAggFunction(
-          ListAggFunction.class, "LIST", PPLReturnTypes.STRING_ARRAY, PPLOperandTypes.ANY_SCALAR);
+          ListAggFunction.class, "LIST", PPLReturnTypes.ARG0_ARRAY, PPLOperandTypes.ANY_SCALAR);
```
penghuo, line 505: Does ALL IT passed? e.g CalciteMultiValueStatsIT

vinaykpud: I have covered them now

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteMultiValueStatsIT.java
```diff
@@ -43,7 +43,7 @@ public void testListFunctionWithBoolean() throws IOException {
                 "source=%s | stats list(boolean_value) as bool_list",
                 TEST_INDEX_DATATYPE_NONNUMERIC));
     verifySchema(response, schema("bool_list", "array"));
-    verifyDataRows(response, rows(List.of("true")));
+    verifyDataRows(response, rows(List.of(true)));
```
penghuo, line 46: @ps48  Do u remember any reasons we enforce return ARRAY(STRING) in PPL?

ps48: This was based on industry standards and workings of multivalue stats commands.

vinaykpud: Can you provide example on this why the return value is always kept string?

ps48: here's the RFC and the decisions we made through the initial implementation https://github.com/opensearch-project/sql/issues/4026

vinaykpud: Thanks @penghuo @ps48 , I fixed it by rewriting the plan which will keep the contract same. 
https://github.com/opensearch-project/OpenSearch/pull/21997

PR #5517, opened 2026-06-04
[Enhancement] Support bare-field join criteria `join on <field>` shorthand

### LINKED ISSUE #5484: [FEATURE] Support implicit field joins when fields match across indices
Consider a query like:
```kql
source = idx1 | join left=l right=r on l.fieldname = r.fieldname idx2
```

It would be nice if, because `fieldname` is the same in both indices, we could write:
```kql
source = idx1 | join left=l right=r on fieldname idx2
```

This currently returns:
```json
{
  "error": {
    "details": "Ambiguous field: fieldname",
    "reason": "Invalid Query",
  },
  "status": 400
}
```

Implement this type of auto-field-join

At least make the error more obvious

N/A

Hanyu-W:
### Description

Lets `join` `on`/`where` criteria accept a **bare field name** (or an `AND` chain of bare fields) as shorthand for the qualified equality on that common field. Before this change, the only way to join on a shared column was to spell out both qualifiers:

```
source=t1 | inner join on a t2          == on t1.a = t2.a
source=t1 | inner join on a AND b t2     == on t1.a = t2.a AND t1.b = t2.b
source=t1 | join on a t2                 == on t1.a = t2.a   (no prefix/alias)
```

The shorthand behaves **exactly** like the explicit `on l.f = r.f` it sugars: it **keeps both key columns** (the right key surfaced as `<rightAlias>.f`, or `<rightTable>.f` when no alias is given). This is intentionally different from the field-list join `join f t2`, which merges the duplicate key into a single column.

Two things made this more than a one-line grammar tweak, and shaped the design:

1. **Grammar ambiguity.** `on` is a keyword that can also be an identifier, so the no-prefix form `join on a t2` was greedily consumed by the field-list alternative — `on` parsed as a *field name*, failing with `field [on] not found`. Reordering the `joinCommand` alternatives so the criteria alternative is listed first makes ANTLR ALL(*) prefer reading `on` as the criteria keyword on genuine ambiguity, with zero change to the field-list form's behavior.
2. **Qualifier resolution.** A bare `on a` carries no table qualifier, but the keep-both planner branch needs `left.a = right.a`. Resolution is done in the **planner** (`CalciteRelNodeVisitor.visitJoin`), where both inputs are already on the RelBuilder stack: each bare field is resolved **by stack position** (`LEFT.f = RIGHT.f`) via the existing `buildJoinConditionByFieldName` helper. The unresolved AST keeps the criteria exactly as the user wrote it (`on a`), so explain/anonymizer reflect the original query. Because matching is positional rather than by qualifier, a self-join `join on f t` is a genuine cross-scan equi-join (`=($7, $15)`), not the `f = f` tautology a name-based rewrite would collapse to.

The keep-both column-rename logic is **unchanged** — it works off the output row types, independent of how the join condition was built. A mixed condition that isn't all bare fields (e.g. `on a AND b.x = c.y`) is detected as not-bare and falls through to the normal `analyzeJoinCondition` path untouched.

Resolves #5484.

### Grammar precedence change

This PR reorders the two `joinCommand` alternatives so the **criteria** alternative is listed first:

```antlr
joinCommand
   // Criteria alt listed first - so `join on a` reads `on` as the criteria keyword, not a field.
   : sqlLikeJoinType? JOIN (joinOption)* sideAlias joinHintList? joinCriteria right = tableOrSubqueryClause
   | JOIN (joinOption)* (fieldList)? right = tableOrSubqueryClause
   ;
```

**This is broader than the new shorthand.** ANTLR's ALL(*) picks the first-listed alternative on genuine ambiguity, so the reorder changes the parse path of **all** join queries, not just the new bare-field form. Without it, the no-prefix form `join on a <right>` would let the field-list alternative greedily consume `on` as a field name instead of as the criteria keyword.

**Existing behavior was re-verified, not assumed.** Two guard tests pin the unchanged paths:

- `testJoinFieldListStillParsesAsFieldList` — `source=t1 | join a t2` still parses as a field-list join (`getJoinFields() == [a]`, no condition).
- `testJoinPrefixWithoutCriteriaKeywordIsSyntaxError` — `source=t1 | inner join a t2` (a prefix with a bare field but no `on`/`where`) is still a `SyntaxCheckException`.

**Edge case (verified).** `ON` and `WHERE` are both reachable as bare-field identifiers via `keywordsCanBeId`, and `joinCriteria: (ON | WHERE) logicalExpression` treats them identically. So a field literally named `on` or `where` in the no-prefix form (`join on t2` / `join where t2`) shifts from a field-list parse to a criteria parse. This is acceptable: `on`/`where` are keyword-like, and their use as bare field-list identifiers in a join prefix was already ambiguous/undocumented.

### Changes

| Area | File | Change |
| --- | --- | --- |
| Grammar | `ppl/src/main/antlr/OpenSearchPPLParser.g4` | Reorder `joinCommand` so the criteria alternative precedes the field-list alternative, so `join on a` reads `on` as the criteria keyword instead of a field |
| Parser | `ppl/.../ppl/parser/AstBuilder.java` | Leave a bare single-part field (or `AND` chain) as the join condition rather than treating it as a field list — the unresolved AST reflects what the user wrote |
| Planner | `core/.../calcite/CalciteRelNodeVisitor.java` | `visitJoin` criteria branch detects a bare-field condition and builds the equi-join by stack position (`LEFT.f = RIGHT.f`) via the existing `buildJoinConditionByFieldName`; otherwise falls through to `analyzeJoinCondition` unchanged |
| Util | `core/.../calcite/utils/JoinAndLookupUtils.java` | `collectBareFields` detects a bare single-part field / AND-chain and returns `Optional<List<String>>` (the name list is assembled only when every node is a bare field) |
| Docs | `docs/user/ppl/cmd/join.md` | Shorthand examples in both syntax sections, `<joinCriteria>` parameter description, and the field-dedup limitation note |
| Tests | `ppl/.../calcite/CalcitePPLJoinTest.java`, `ppl/.../parser/AstBuilderTest.java`, `ppl/.../utils/PPLQueryDataAnonymizerTest.java`, `core/.../calcite/utils/JoinAndLookupUtilsTest.java` | Plan-shape, AST-parse, anonymizer, and bare-field-detector coverage |
| QA (IT) | `integ-test/.../calcite/remote/CalcitePPLJoinIT.java` | End-to-end coverage on a live cluster |

### Testing

`CalcitePPLJoinIT` on a live cluster, exercising the new shorthand against pre-change behavior:

| Test | Exercises | Before | After |
| --- | --- | --- | --- |
| `testJoinWithImplicitField` | `inner join on name` — keep-both; asserts shorthand output is identical to explicit `on l.name = r.name`; result asserted in sort order via `verifyDataRowsInOrder` | ❌ | ✅ |
| `testJoinNoPrefixImplicitField` | `join on name` — no prefix; `on` parsed as the criteria keyword, identical to the explicit form (not the field-list merge) | ❌ | ✅ |
| `testLeftJoinWithImplicitField` | `left join on name` — unmatched-left rows keep the left key (no NULL key; keep-both does not coalesce) | ❌ | ✅ |
| `testAliasedSelfJoinWithImplicitField` | self-join on a unique key — a real equi-join (8 rows), not the 64-row cross product a tautology would give | ❌ | ✅ |

> Before = the same query run against pre-change code (parse/plan error — the bare field was read as a field name, e.g. `field [on] not found`).

Unit suites (all green, run locally):

- `CalcitePPLJoinTest`: **57** — plan shape for single/multi-field, alias, `where`, no-prefix, left/semi/anti, self-join (real equi-join), `max=N` dedup wrapper, mixed-condition fall-through, and not-on-both-sides cases.
- `AstBuilderTest`: **158** — rewrite-vs-passthrough decisions (bare field kept; qualified/comparison criteria untouched; field-list still parses as a field list; prefix-without-criteria is a syntax error).
- `PPLQueryDataAnonymizerTest`: **119** — shorthand anonymizes correctly (renders `on identifier`).
- `JoinAndLookupUtilsTest`: **5** — `collectBareFields` over single field, AND-chain, qualified field, mixed condition, and the no-partial-state case.

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -1890,12 +1890,13 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     } else {
       // The join-with-criteria grammar doesn't allow empty join condition
       RexNode joinCondition;
-      List<String> bareFields = new ArrayList<>();
-      if (JoinAndLookupUtils.collectBareFields(node.getJoinCondition().get(), bareFields)) {
+      Optional<List<String>> bareFields =
+          JoinAndLookupUtils.collectBareFields(node.getJoinCondition().get());
+      if (bareFields.isPresent()) {
         // Bare-field shorthand `on a [AND b ...]`. Resolving by stack position rather than by
         // qualifier keeps a self-join `join on f t` a real cross-scan equi-join, not f = f.
         joinCondition =
-            bareFields.stream()
+            bareFields.get().stream()
                 .map(f -> buildJoinConditionByFieldName(context, f))
                 .reduce(context.rexBuilder::and)
                 .orElse(context.relBuilder.literal(true));
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -1889,10 +1889,22 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
       }
     } else {
       // The join-with-criteria grammar doesn't allow empty join condition
-      RexNode joinCondition =
-          node.getJoinCondition()
-              .map(c -> rexVisitor.analyzeJoinCondition(c, context))
-              .orElse(context.relBuilder.literal(true));
+      RexNode joinCondition;
+      List<String> bareFields = new ArrayList<>();
+      if (JoinAndLookupUtils.collectBareFields(node.getJoinCondition().get(), bareFields)) {
+        // Bare-field shorthand `on a [AND b ...]`. Resolving by stack position rather than by
+        // qualifier keeps a self-join `join on f t` a real cross-scan equi-join, not f = f.
+        joinCondition =
+            bareFields.stream()
+                .map(f -> buildJoinConditionByFieldName(context, f))
+                .reduce(context.rexBuilder::and)
+                .orElse(context.relBuilder.literal(true));
+      } else {
+        joinCondition =
+            node.getJoinCondition()
+                .map(c -> rexVisitor.analyzeJoinCondition(c, context))
+                .orElse(context.relBuilder.literal(true));
+      }
       if (node.getJoinType() == SEMI || node.getJoinType() == ANTI) {
         // semi and anti join only return left table outputs
         context.relBuilder.join(
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/JoinAndLookupUtils.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/JoinAndLookupUtils.java
@@ -13,6 +13,10 @@ import java.util.stream.Collectors;
 import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.util.Pair;
+import org.opensearch.sql.ast.expression.And;
+import org.opensearch.sql.ast.expression.Field;
+import org.opensearch.sql.ast.expression.QualifiedName;
+import org.opensearch.sql.ast.expression.UnresolvedExpression;
 import org.opensearch.sql.ast.tree.Join;
 import org.opensearch.sql.ast.tree.Lookup;
 import org.opensearch.sql.calcite.CalcitePlanContext;
@@ -35,6 +39,24 @@ public interface JoinAndLookupUtils {
       default:
         return JoinRelType.INNER;
     }
+  }
+
+  /**
+   * Collects the names of bare single-part-field join criteria (e.g. `on a AND b` -> [a, b]).
+   * Returns false for anything else (qualified field, comparison, OR); {@code out} is only valid
+   * when this returns true, so callers must check the return before reading it.
+   */
+  static boolean collectBareFields(UnresolvedExpression expr, List<String> out) {
+    if (expr instanceof And and) {
+      return collectBareFields(and.getLeft(), out) && collectBareFields(and.getRight(), out);
+    }
+    if (expr instanceof Field field
+        && field.getField() instanceof QualifiedName qn
+        && qn.getParts().size() == 1) {
+      out.add(qn.getParts().get(0));
+      return true;
+    }
+    return false;
   }
 
   /* ------For Lookup------ */
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/JoinAndLookupUtils.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/JoinAndLookupUtils.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rex.RexNode;
@@ -42,21 +43,29 @@ public interface JoinAndLookupUtils {
   }
 
   /**
-   * Collects the names of bare single-part-field join criteria (e.g. `on a AND b` -> [a, b]).
-   * Returns false for anything else (qualified field, comparison, OR); {@code out} is only valid
-   * when this returns true, so callers must check the return before reading it.
+   * Collects the names of bare single-part-field join criteria (e.g. {@code on a AND b} -> {@code
+   * Optional.of(["a","b"])}). Returns empty for anything else (qualified field, comparison, OR).
+   * The list is only constructed when every node in the AND-tree is a bare field.
    */
-  static boolean collectBareFields(UnresolvedExpression expr, List<String> out) {
+  static Optional<List<String>> collectBareFields(UnresolvedExpression expr) {
     if (expr instanceof And and) {
-      return collectBareFields(and.getLeft(), out) && collectBareFields(and.getRight(), out);
+      return collectBareFields(and.getLeft())
+          .flatMap(
+              left ->
+                  collectBareFields(and.getRight())
+                      .map(
+                          right -> {
+                            List<String> merged = new ArrayList<>(left);
+                            merged.addAll(right);
+                            return merged;
+                          }));
     }
     if (expr instanceof Field field
         && field.getField() instanceof QualifiedName qn
-        && qn.getParts().size() == 1) {
-      out.add(qn.getParts().get(0));
-      return true;
+        && qn.getPrefix().isEmpty()) {
+      return Optional.of(new ArrayList<>(List.of(qn.getSuffix())));
     }
-    return false;
+    return Optional.empty();
   }
 
   /* ------For Lookup------ */
--- a/docs/user/ppl/cmd/join.md
+++ b/docs/user/ppl/cmd/join.md
@@ -29,6 +29,9 @@ source = table1 | left anti join left = l right = r on l.a = r.a table2
 source = table1 | join left = l right = r [ source = table2 | where d > 10 | head 5 ]
 source = table1 | inner join on table1.a = table2.a table2 | fields table1.a, table2.a, table1.b, table1.c
 source = table1 | inner join on a = c table2 | fields a, b, c, d
+source = table1 | inner join on a table2 | fields a, table2.a, b, c
+source = table1 | inner join on a AND b table2 | fields a, table2.a, b, table2.b
+source = table1 | join on a table2 | fields a, table2.a, b, c
 source = table1 as t1 | join left = l right = r on l.a = r.a table2 as t2 | fields l.a, r.a
 source = table1 as t1 | join left = l right = r on l.a = r.a table2 as t2 | fields t1.a, t2.a
 source = table1 | join left = l right = r on l.a = r.a [ source = table2 ] as s | fields l.a, s.a
@@ -40,7 +43,7 @@ The basic `join` syntax supports the following parameters.
 
 | Parameter | Required/Optional | Description |
 | --- | --- | --- |
-| `<joinCriteria>` | Required | A comparison expression specifying how to join the datasets. Must be placed after the `on` or `where` keyword in the query. |
+| `<joinCriteria>` | Required | A comparison expression specifying how to join the datasets. Must be placed after the `on` or `where` keyword in the query. A bare field name `f` (or `f AND g ...`) is shorthand for an equi-join on the common field(s) and keeps both key columns. |
 | `<right-dataset>` | Required | The right dataset, which can be an index or a subsearch, with or without an alias. |
 | `joinType` | Optional | The type of join to perform. Valid values are `left`, `semi`, `anti`, and performance-sensitive types (`right`, `full`, and `cross`). Default is `inner`. |
 | `left` | Optional | An alias for the left dataset (typically a subsearch) used to avoid ambiguous field names. Specify as `left = <leftAlias>`. |
@@ -71,7 +74,7 @@ The extended `join` syntax supports the following parameters.
 
 | Parameter | Required/Optional | Description |
 | --- | --- | --- |
-| `<joinCriteria>` | Required | A comparison expression specifying how to join the datasets. Must be placed after the `on` or `where` keyword in the query. |
+| `<joinCriteria>` | Required | A comparison expression specifying how to join the datasets. Must be placed after the `on` or `where` keyword in the query. A bare field name `f` (or `f AND g ...`) is shorthand for an equi-join on the common field(s); it keeps both key columns and differs from `<join-field-list>`, which merges duplicates. |
 | `<right-dataset>` | Required | The right dataset, which can be an index or a subsearch, with or without an alias. |  
 | `type` | Optional | The join type when using extended syntax. Valid values are `left`, `outer` (same as `left`), `semi`, `anti`, and performance-sensitive types (`right`, `full`, and `cross`). Default is `inner`. |
 | `<join-field-list>` | Optional | A list of fields used to build the join criteria. These fields must exist in both datasets. If not specified, all fields common to both datasets are used as join keys. |
--- a/ppl/src/main/antlr/OpenSearchPPLParser.g4
+++ b/ppl/src/main/antlr/OpenSearchPPLParser.g4
@@ -729,8 +729,9 @@ sourceFilterArg
 
 // join
 joinCommand
-   : JOIN (joinOption)* (fieldList)? right = tableOrSubqueryClause
-   | sqlLikeJoinType? JOIN (joinOption)* sideAlias joinHintList? joinCriteria right = tableOrSubqueryClause
+   // Criteria alt listed first - so `join on a` reads `on` as the criteria keyword, not a field.
+   : sqlLikeJoinType? JOIN (joinOption)* sideAlias joinHintList? joinCriteria right = tableOrSubqueryClause
+   | JOIN (joinOption)* (fieldList)? right = tableOrSubqueryClause
    ;
 
 sqlLikeJoinType
--- a/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLJoinTest.java
+++ b/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLJoinTest.java
@@ -1050,6 +1050,38 @@ public class CalcitePPLJoinTest extends CalcitePPLAbstractTest {
   }
 
   @Test
+  public void testJoinWithImplicitFieldMaxGreaterThanZero() {
+    // Bare-field shorthand combined with a max=N dedup wrapper. Keep-both renames the
+    // unaliased right-side key to `DEPT.DEPTNO`, and the dedup subquery wraps the right input.
+    String ppl = "source=EMP | inner join max=1 on DEPTNO DEPT";
+    RelNode root = getRelNode(ppl);
+    String expectedLogical =
+        "LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5],"
+            + " COMM=[$6], DEPTNO=[$7], DEPT.DEPTNO=[$8], DNAME=[$9], LOC=[$10])\n"
+            + "  LogicalJoin(condition=[=($7, $8)], joinType=[inner])\n"
+            + "    LogicalTableScan(table=[[scott, EMP]])\n"
+            + "    LogicalProject(DEPTNO=[$0], DNAME=[$1], LOC=[$2])\n"
+            + "      LogicalFilter(condition=[<=($3, 1)])\n"
+            + "        LogicalProject(DEPTNO=[$0], DNAME=[$1], LOC=[$2],"
+            + " _row_number_dedup_=[ROW_NUMBER() OVER (PARTITION BY $0)])\n"
+            + "          LogicalTableScan(table=[[scott, DEPT]])\n";
+    verifyLogical(root, expectedLogical);
+    verifyResultCount(root, 14);
+
+    String expectedSparkSql =
+        "SELECT `EMP`.`EMPNO`, `EMP`.`ENAME`, `EMP`.`JOB`, `EMP`.`MGR`, `EMP`.`HIREDATE`,"
+            + " `EMP`.`SAL`, `EMP`.`COMM`, `EMP`.`DEPTNO`, `t1`.`DEPTNO` `DEPT.DEPTNO`,"
+            + " `t1`.`DNAME`, `t1`.`LOC`\n"
+            + "FROM `scott`.`EMP`\n"
+            + "INNER JOIN (SELECT `DEPTNO`, `DNAME`, `LOC`\n"
+            + "FROM (SELECT `DEPTNO`, `DNAME`, `LOC`, ROW_NUMBER() OVER (PARTITION BY `DEPTNO`)"
+            + " `_row_number_dedup_`\n"
+            + "FROM `scott`.`DEPT`) `t`\n"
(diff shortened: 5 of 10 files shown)


core/src/main/java/org/opensearch/sql/ast/tree/Join.java
```diff
@@ -70,9 +77,92 @@ public UnresolvedPlan attach(UnresolvedPlan child) {
...
+        .map(
+            f ->
+                (UnresolvedExpression)
+                    new Compare(
+                        "=",
+                        new Field(QualifiedName.of(leftQual, f)),
+                        new Field(QualifiedName.of(rightQual, f))))
+        .reduce((l, r) -> new And(l, r))
+        .orElseThrow(
+            () -> new SemanticCheckException("join 'on' shorthand requires at least one field"));
```
Swiddis: is this the only error case that can throw here?

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
```diff
@@ -70,9 +77,92 @@ public UnresolvedPlan attach(UnresolvedPlan child) {
...
+  }
+
+  /**
+   * Derives the qualifier (table name) for an unaliased side by finding the single underlying
+   * Relation. Throws when no single named table is reachable (e.g. an unaliased subquery).
+   */
+  private static String qualifierOf(UnresolvedPlan plan, String side) {
+    Relation relation = findRelation(plan);
+    if (relation == null) {
+      throw new SemanticCheckException(
```
Swiddis: nice detail on the error message!

consider using the `ReportBuilder` to split out suggestions from the problem

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
```diff
@@ -70,9 +77,92 @@ public UnresolvedPlan attach(UnresolvedPlan child) {
     if (rightAlias.isEmpty() && this.right instanceof SubqueryAlias alias) {
       rightAlias = Optional.of(alias.getAlias());
     }
+    rewriteBareFieldCriteria();
```
Swiddis: Is putting it here the right boundary over resolving these fields in the constructor (or even frontloading it to the AST builder)?

Putting the rewrite here introduces a weird boundary, where now `Join` plays the role of dynamically parsing itself during attachment based directly on AST grammar choices. That multiple places comment "we actually resolve these fields in attach()" also hints that this is unintuitive.

It might be clearer to put this in the constructor. Alternatively, it's a little more work, but we could consider moving all of `leftAlias`, `rightAlias`, `joinFields` and similar into a dedicated `JoinClause` data type that supports operations like `clause.expandBareFields()` and `clause.resolveAliases()`

docs/user/ppl/cmd/join.md
```diff
@@ -220,6 +223,6 @@ The `join` command has the following limitations:
   | `source=table1 \| join on table1.id=t2.id table2 as t2 \| eval a = 1` | `table1.id, t2.id, a` |
   | `source=table1 \| join right=tt on table1.id=t2.id [ source=table2 as t2 \| eval b = id ] \| eval a = 1` | `table1.id, tt.id, tt.b, a` |
 
-* **Field deduplication in extended syntax** – When using the extended syntax with a field list, duplicate field names in the output are deduplicated according to the `overwrite` option.
+* **Field deduplication in extended syntax** – When using the extended syntax with a field list (`<join-field-list>`), duplicate field names are deduplicated according to the `overwrite` option. The bare-field `on`/`where` shorthand keeps both key columns, unlike the field-list form.
```
Swiddis: the query grammar at the top of the doc wasn't updated to show what `join-field-list` is

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
```diff
@@ -70,9 +77,92 @@ public UnresolvedPlan attach(UnresolvedPlan child) {
...
     return this;
   }
 
+  /**
+   * Rewrite bare-field criteria `on a` to qualified equality `on l.a = r.a`, keeping both key
+   * columns. Runs in attach() since that is the first point both inputs (and the left qualifier)
+   * are known; qualifier is the side alias else the table name. Non-bare criteria (qualified, OR,
+   * comparisons) are left untouched.
+   */
+  private void rewriteBareFieldCriteria() {
```
RyanL1997: This method resolves the left and right qualifiers independently, with no check that they differ. When both sides are the same unaliased table (e.g. `source=EMP | inner join on DEPTNO EMP`), `qualifierOf` returns `"EMP"` for both, and the conjunction it builds becomes EMP.DEPTNO = EMP.DEPTNO — a tautology that Calcite simplifies to `IS NOT NULL(...)`, not an equi-join. The user gets a near-cross-product with no error or warning.

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
RyanL1997: Performing semantic rewriting inside the AST node's `attach()` mixes tree construction with analysis, which usually lives in the analyzer layer here. The unresolved tree no longer reflects what the user wrote, which can make explain output and debugging confusing. Was an analyzer pass considered?

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
```diff
@@ -70,9 +77,92 @@ public UnresolvedPlan attach(UnresolvedPlan child) {
...
+    return fieldNames.stream()
+        .map(
+            f ->
+                (UnresolvedExpression)
+                    new Compare(
+                        "=",
+                        new Field(QualifiedName.of(leftQual, f)),
+                        new Field(QualifiedName.of(rightQual, f))))
+        .reduce((l, r) -> new And(l, r))
+        .orElseThrow(
```
RyanL1997: This `orElseThrow` looks unreachable: `rewriteBareFieldCriteria` early-returns when no bare fields are collected, and `collectBareFields` only returns true after adding at least one. Can you double check?

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
```diff
@@ -70,9 +77,92 @@ public UnresolvedPlan attach(UnresolvedPlan child) {
...
+          "the `on <field>` join shorthand requires the "
+              + side
+              + " side to be a named table or carry an alias; use the qualified `on l.f = r.f` form"
+              + " or add a left=/right=/AS alias");
+    }
+    return relation.getTableQualifiedName().getSuffix();
+  }
+
+  /** Walk the single-child chain down to the underlying Relation. */
+  private static Relation findRelation(UnresolvedPlan plan) {
```
RyanL1997: This bails as soon as it hits a node with !=1 child. Please add a test for an aliased subquery with a bare field (`source=t1 | join on a [ source=t2` ] as s) to confirm it uses the alias and never reaches `qualifierOf`.

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
```diff
@@ -28,7 +34,8 @@ public class Join extends UnresolvedPlan {
   private Optional<String> leftAlias;
   private Optional<String> rightAlias;
   private final JoinType joinType;
-  private final Optional<UnresolvedExpression> joinCondition;
+  // Not final: bare-field `on <field>` shorthand is rewritten to qualified equality in attach().
+  private Optional<UnresolvedExpression> joinCondition;
```
RyanL1997: Dropping `final` to allow mutation in `attach()` weakens this AST node's immutability. If `attach()` ever runs twice or the node is shared, the rewrite re-runs against already-rewritten state. It's idempotent in practice (the qualified result fails `collectBareFields`), but the invariant is fragile — consider a guard flag or doing this rewrite in an analyzer pass instead of mutating here.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLJoinTest.java
```diff
@@ -1163,4 +1163,207 @@ public void testSqlLikeJoinWithSpecificJoinType() {
...
+  @Test
+  public void testJoinWithImplicitFieldMultiField() {
+    String ppl = "source=EMP | inner join on DEPTNO AND SAL EMP";
+    RelNode root = getRelNode(ppl);
+    String expectedLogical =
+        "LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5],"
+            + " COMM=[$6], DEPTNO=[$7], EMP.EMPNO=[$8], EMP.ENAME=[$9], EMP.JOB=[$10],"
+            + " EMP.MGR=[$11], EMP.HIREDATE=[$12], EMP.SAL=[$13], EMP.COMM=[$14],"
+            + " EMP.DEPTNO=[$15])\n"
+            + "  LogicalJoin(condition=[AND(IS NOT NULL($15), IS NOT NULL($13))],"
```
RyanL1997: This asserts `AND(IS NOT NULL($15), IS NOT NULL($13))` as expected output, which locks in the self-join tautology as if intended. Should flip to an expected-exception test once the guard is added.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLJoinTest.java
```diff
@@ -1163,4 +1163,207 @@ public void testSqlLikeJoinWithSpecificJoinType() {
...
+  @Test
+  public void testJoinWithImplicitFieldSelfJoinIsTautology() {
+    String ppl = "source=EMP | inner join on DEPTNO EMP";
+    RelNode root = getRelNode(ppl);
+    String expectedLogical =
+        "LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5],"
+            + " COMM=[$6], DEPTNO=[$7], EMP.EMPNO=[$8], EMP.ENAME=[$9], EMP.JOB=[$10],"
+            + " EMP.MGR=[$11], EMP.HIREDATE=[$12], EMP.SAL=[$13], EMP.COMM=[$14],"
+            + " EMP.DEPTNO=[$15])\n"
+            + "  LogicalJoin(condition=[IS NOT NULL($15)], joinType=[inner])\n"
```
RyanL1997: Same concern as the above— this enshrines `IS NOT NULL($15)` as correct. The test name even admits it's a tautology. Convert to expected-exception, or revisit once self-join handling is decided.

RyanL1997 (commented):
Hi @Hanyu-W , thanks for the change, and just left some reviews.

docs/user/ppl/cmd/join.md
Hanyu-W: Good catch! I'll cut it because it isn't adding anything valuable to the docs.

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
RyanL1997: +1

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
RyanL1997: +1, same concern I called out [here](https://github.com/opensearch-project/sql/pull/5517#discussion_r3359127682)

Hanyu-W: Yes — that's the fix I'm making. I'm going to move the expansion into visitJoin instead of attach(), so the AST 
 keeps query how wrote it and the analysis happens in the planner where it belongs. Explain output will now match the original query too.

Hanyu-W: > Yes — that's the fix I'm making. I'm going to move the expansion into visitJoin instead of attach(), so the AST keeps query how wrote it and the analysis happens in the planner where it belongs. Explain output will now match the original query too.

Yeah, I'll make the change Ryan suggested to fix this.

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
Hanyu-W: I think it should be fixed now I'm moving it into the planner, as the planner already knows which input is left and which is right by position, so doing the match there produces a correct equi-join even for join on DEPTNO EMP.

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
Hanyu-W: You're right, it's unreachable. The method will be deleted in the redesign

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
Hanyu-W: This method will get deleted too, but I'll still add the aliased-subquery test to confirm.

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
Hanyu-W: Good point. I'll restore final as the redesign should no longer mutate the AST here

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLJoinTest.java
Hanyu-W: I'd actually go the other way: instead of making the self-join an error, the redesign makes it work correctly. So I'll update these to assert the real join condition rather than the tautology — the query is valid, it was just being built wrong

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLJoinTest.java
Hanyu-W: > I'd actually go the other way: instead of making the self-join an error, the redesign makes it work correctly. So I'll update these to assert the real join condition rather than the tautology — the query is valid, it was just being built wrong

same response as above

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
Hanyu-W: That error path is going away — once matching is done by position in the planner, there's no "must be a named table" restriction, so the exception it was attached to is deleted.

core/src/main/java/org/opensearch/sql/ast/tree/Join.java
Hanyu-W: There were two (this one, and an empty-list guard above it). Both are removed in the redesign, so neither throws anymore.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLJoinTest.java
RyanL1997: agree.

Hanyu-W:
📝 **Implementation note — the description above predates the redesign.**

Heads up for anyone reading top-to-bottom: after the review feedback, I moved the bare-field expansion **out of the AST (`Join.attach()`) and into the planner**, so the original description's "Changes" table and its "planner 100% unchanged" line no longer match the committed code. I'm leaving the original post as-is for review-thread continuity; here's the accurate summary of what's on the branch now (`19e6693`):

**Design (as committed):**
- The bare-field condition is kept **verbatim** in the unresolved AST, so explain/anonymizer reflect what the user wrote. `Join.java` is **unchanged** vs `main`.
- Expansion happens in `CalciteRelNodeVisitor.visitJoin`: the criteria branch detects a bare single-part field / `AND`-chain via a new `JoinAndLookupUtils.collectBareFields`, then resolves each field **positionally** through the existing `buildJoinConditionByFieldName` — `field(2,0,f) = field(2,1,f)`, i.e. left-input.f = right-input.f **by stack position, not by qualifier**.
- Because resolution is positional, a self-join `join on f t` is a real cross-scan equi-join (`=($7,$15)`), **not** the `f = f` tautology a name-qualified rewrite would collapse to — this is the issue @RyanL1997 flagged, now fixed by construction. There is no `SemanticCheckException` / `qualifierOf` / `findRelation` / `rewriteBareFieldCriteria` (all removed along with the AST rewrite).

**Files actually changed** (`git diff main...HEAD`):

| File | Change |
| --- | --- |
| `ppl/src/main/antlr/OpenSearchPPLParser.g4` | Reorder `joinCommand` so the criteria alt precedes the field-list alt (so `join on a` reads `on` as the criteria keyword, not a field) |
| `ppl/.../ppl/parser/AstBuilder.java` | Keep the bare `on <field>` (or `AND` chain) verbatim instead of folding it into the field list |
| `core/.../calcite/CalciteRelNodeVisitor.java` | `visitJoin` criteria branch: build the equi-join positionally for a bare-field condition; otherwise fall through to `analyzeJoinCondition` unchanged |
| `core/.../calcite/utils/JoinAndLookupUtils.java` | New `collectBareFields` helper (true only for a bare single-part `Field` or an `AND`-chain of them) |
| `docs/user/ppl/cmd/join.md` | Shorthand examples + `<joinCriteria>` parameter wording |
| test files | `CalcitePPLJoinTest`, `AstBuilderTest`, `PPLQueryDataAnonymizerTest`, `CalcitePPLJoinIT` |

**`Join.java` is not touched** — contrary to the "Rewrite" row in the original Changes table.

**Tests (re-run locally just now, all green):**
- `CalcitePPLJoinTest`: **53/53** (13 new) — plan shape for single/multi-field, alias, `where`, no-prefix, left/semi, the **self-join real equi-join** `=($7,$15)`, aliased/unaliased subquery, and not-on-both-sides error.
- `AstBuilderTest`: **158/158** (11 new) — asserts the bare field is **kept verbatim** (not rewritten); qualified/comparison criteria untouched; field-list still parses as a field list; prefix-without-criteria is a syntax error.
- `PPLQueryDataAnonymizerTest`: **119/119** (1 new) — anonymizes the un-rewritten condition (`on identifier`).
- `CalcitePPLJoinIT`: **4 new** methods, green in CI's `build-linux (…, integration)` jobs on this commit.

core/src/main/java/org/opensearch/sql/calcite/utils/JoinAndLookupUtils.java
```diff
@@ -37,6 +41,24 @@ static JoinRelType translateJoinType(Join.JoinType joinType) {
     }
   }
 
+  /**
+   * Collects the names of bare single-part-field join criteria (e.g. `on a AND b` -> [a, b]).
+   * Returns false for anything else (qualified field, comparison, OR); {@code out} is only valid
+   * when this returns true, so callers must check the return before reading it.
+   */
+  static boolean collectBareFields(UnresolvedExpression expr, List<String> out) {
```
RyanL1997: The `out` list is mutated as a side effect and is only valid when the method returns true — the Javadoc says so, which is good. But this is the classic boolean-return-plus-out-param pattern that's easy to misuse: a caller that reads `out` after a `false` return gets a partially-populated list (the recursion adds left-side fields before the right side returns false).

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -1889,10 +1889,22 @@ public RelNode visitJoin(Join node, CalcitePlanContext context) {
       }
     } else {
       // The join-with-criteria grammar doesn't allow empty join condition
-      RexNode joinCondition =
-          node.getJoinCondition()
-              .map(c -> rexVisitor.analyzeJoinCondition(c, context))
-              .orElse(context.relBuilder.literal(true));
+      RexNode joinCondition;
```
RyanL1997, line 1892: What happens with a mixed condition like `on a AND b.x = c.y` ?

ppl/src/main/antlr/OpenSearchPPLParser.g4
```diff
@@ -729,8 +729,9 @@ sourceFilterArg
 
 // join
 joinCommand
-   : JOIN (joinOption)* (fieldList)? right = tableOrSubqueryClause
-   | sqlLikeJoinType? JOIN (joinOption)* sideAlias joinHintList? joinCriteria right = tableOrSubqueryClause
+   // Criteria alt listed first - so `join on a` reads `on` as the criteria keyword, not a field.
+   : sqlLikeJoinType? JOIN (joinOption)* sideAlias joinHintList? joinCriteria right = tableOrSubqueryClause
```
RyanL1997, line 733: Same as before: the alternative reorder affects every join query's parse, not just the shorthand. `testJoinFieldListStillParsesAsFieldList` and `testJoinPrefixWithoutCriteriaKeywordIsSyntaxError` are good guards. 

Please still call the grammar precedence change out explicitly in the PR description so reviewers don't miss it — the new comment covers the why well, but a note that existing queries were re-verified would close it out.

Hanyu-W: Will do-adding a note to the PR description that the reorder affects all join queries' parse path, and that existing behavior is verified by testJoinFieldListStillParsesAsFieldList + testJoinPrefixWithoutCriteriaKeywordIsSyntaxError (158 AstBuilder + 55 CalcitePPLJoin tests).

core/src/main/java/org/opensearch/sql/calcite/utils/JoinAndLookupUtils.java
Hanyu-W: That's a good point. I refactored the code to Optional<List<String>>, so therefore the list is only assembled when every node in the AND-tree is a bare field, so partial state can't exist. 

I also added JoinAndLookupUtilsTest with 5 cases including mixedLeftSideReturnsEmptyNoPartialState, confirming that And(Compare(...), Field("b")) returns Optional.empty(), not a list containing "b".

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
Hanyu-W: I tested this, collectBareFields returns Optional.empty(), since the Compare node isn't a bare field), so the whole expression falls through to analyzeJoinCondition unchanged. I added two tests (testJoinMixedConditionFallsThrough and testJoinMixedConditionWithInequality) to prove it produces the correct plan via the normal path.

RyanL1997 (approved):

Swiddis (approved):

PR #5534, opened 2026-06-10
update datetime tests to stay within ae bounds

Swiddis:
### Description
Simplifies the test indices & changes some values, preserves all the same core semantics of the test. Parquet/Datafusion use 64-bit microsecond epochs which cap at year 2262.

### Related Issues
N/A

### Check List


integ-test/src/test/java/org/opensearch/sql/ppl/DateTimeComparisonIT.java
```diff
@@ -28,7 +28,7 @@ public class DateTimeComparisonIT extends PPLIntegTestCase {
   @Override
   public void init() throws Exception {
     super.init();
-    loadIndex(Index.DATA_TYPE_NONNUMERIC);
+    loadIndex(Index.DATETIME_SIMPLE);
```
RyanL1997, line 31: qq: does the sql variant's `testCompare` actually need this index loaded? scrolling down to the method body, it is just `select <expr> AS <name>` with no `FROM` clause, so i don't think this load is reaching any query.

RyanL1997 (approved):
Thanks for the change @Swiddis. 

one small nit on the description — the i64 cap that kicks in at year 2262 is from nanosecond precision, not microsecond (μs would actually let us go way past year 290k). doesn't change anything about the fix itself since 2242 is under 2262 either way.

integ-test/src/test/java/org/opensearch/sql/ppl/DateTimeComparisonIT.java
Swiddis: lol, probably. i think it's just copied from ppl maybe since ppl doesn't support bare queries without indices. won't rock the boat with it though

PR #5541, opened 2026-06-11
[analytics-engine] Strip AE-unsupported fields from test data and exclude the ITs doomed by it

mengweieric:
### Problem

On the analytics-engine route (`-Dtests.analytics.parquet_indices=true`), a test dataset whose mapping contains a field type the parquet/composite store can't handle (`nested`, `geo_point`, `geo_shape`, `alias`, or a store-disabled `binary`) fails index creation / ingestion. The failure surfaces later as an unrelated `expected:<1> but was:<0>` in a downstream IT rather than at the offending index/field — masking real assertions.

### Fix

**1. Strip unsupported-typed fields at load time** — gated on `tests.analytics.parquet_indices`; a byte-for-byte no-op off the AE route.

- `createIndexByRestClient` removes unsupported `{nested, geo_point, geo_shape, alias}` properties **recursively** and reports the **exact dropped paths** (e.g. `[location, point]`), not just top-level names.
- `loadIndex` passes those paths to `loadDataByRestClient`, which strips each path from every bulk source doc **recursively — through objects and arrays of objects** — preserving siblings (`location.city` etc.), so mapping and data agree. Untouched docs stay byte-for-byte verbatim.
- `loadDataByRestClient` now **fails loudly on bulk `errors=true`** (names the index + first item errors), for all loads — silent partial ingestion no longer slips through a 200.
- The trigger is the mapping's field **types**, not the test's identity, so any IT — existing or newly added — loading such a dataset is handled out of the box. No per-index variant files. `UNSUPPORTED_FIELD_TYPES` is the single source of truth, imported by the verifier so the two can't drift.

`binary` is **conditional**: kept when `store: true` (engine reads it as `VARBINARY`), but stripped when `store: false` — the parquet store can't create a store-disabled binary field (`Unable to derive source for [X] with store disabled`, confirmed on the AE cluster).

**2. Physically exclude PPL ITs doomed by the strip** — a test whose query references a stripped field can never pass once that field is gone, so excluding it is correct, not a workaround. Centralized in `integ-test/build.gradle`, gated on the parsed `analyticsEnabled` boolean (so `=false` excludes nothing), method-level wherever possible so the rest of each class still runs. Verified field-by-field against the index mappings:

| Excluded | Reason | Level |
|---|---|---|
| `GeoPointFormatsIT` | every test reads a `geo_point` field | class |
| `DataTypeIT.test_nonnumeric_data_types` | asserts `nested_value` + `geo_point_value` | method |
| `DataTypeIT.test_alias_data_type` | `where alias_col` (`type: alias`) | method |
| `SystemFunctionIT.typeof_opensearch_types` | `typeof(geo_point_value)` | method |

Kept running (verified **not** doomed): `GeoIpFunctionsIT` (`geoip()` over `text` ip/name), `StatsCommandIT.testStatsNested*` (nested *function calls*, not a nested field), `ConvertCommandIT`/`TrendlineCommandIT`/`ExplainIT` alias usages (query result aliases, not `alias`-typed fields).

### Verification

Verified locally — CI's `integTest` jobs run the **legacy/default route** (no `tests.analytics.parquet_indices`), so they never exercise this code; the AE route needs a force-routed 9-plugin cluster CI doesn't spin up.

**Goal achieved (AE route, force-routed cluster, `_explain` confirmed `LogicalTableScan(table=[[opensearch` = real AE, not Calcite fallback):**
- `AnalyticsUnsupportedFieldStripVerifyIT` **passes over every `Index` dataset** — per dataset it forces a clean reload, then asserts: clean live mapping, no stripped path left in sampled `_source`, and doc-count == source-doc count. (1 dataset, `DATASOURCES`, logs a transparent skip of the count leg only — its count query touches a system index; legs 1–3 still ran.)
- This is the fix proving itself: the verifier *caught* the `binary store:false` case, which the conditional strip now handles.

**Unit (pure-logic, no cluster) — `AnalyticsFieldStripTests` 7/7:** recursive path strip, `complex_geo`-shape (`location.point` removed, siblings kept, deep `nested_locations.primary.office`), arrays of objects, byte-identity of untouched docs, `binary` store-conditional, no-op-when-disabled.

**No regression when AE is off (the v2/v3 path):** strip code early-returns on `!isEnabled()` (mapping/data byte-identical); excludes are gated on the parsed `analyticsEnabled`. Confirmed empirically — ran `StatsCommandIT`/`DataTypeIT`/`FieldsCommandIT`/`SearchCommandIT` **legacy (no parquet flag): 123 pass / 0 fail**, and the new bulk-error check never tripped.

### Check List


mengweieric:
Failing CIs are unrelated to this PR

ahkcs (approved):

dai-chen (commented):
Is all our IT CI broken? Just wonder how to verify these changes..

mengweieric:
> Is all our IT CI broken? Just wonder how to verify these changes..

we should fix the CIs besides this PR but I verified locally of

Run AnalyticsUnsupportedFieldStripVerifyIT on a force-routed AE cluster with -Dtests.analytics.parquet_indices=true. It loads every Index dataset through the real load path and asserts per dataset: clean mapping, no stripped path left in _source, doc count == source doc count. It passes, and fails loudly/attributably if any dataset still ingests an unsupported field, so it's the durable proof artifact in the PR.

Also: unit tests 7/7 (nested paths + arrays of objects), and legacy route with no parquet flag 123 pass / 0 fail so nothing regresses when AE is off.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteAliasFieldAggregationIT.java
```diff
@@ -31,6 +32,12 @@ public class CalciteAliasFieldAggregationIT extends PPLIntegTestCase {
   @Override
   public void init() throws Exception {
     super.init();
+    // Alias fields are unsupported on the analytics-engine (parquet/composite) route — the index
+    // can't even be created, and these tests query the alias fields directly. There's nothing to
+    // salvage by stripping, so skip the whole class on the AE route. Runs normally otherwise.
+    Assume.assumeFalse(
+        "Alias-field aggregation is unsupported on the analytics-engine route",
+        isAnalyticsParquetIndicesEnabled());
```
dai-chen: we may want to maintain exclude list in single place like build.gradle?

integ-test/src/test/java/org/opensearch/sql/legacy/TestUtils.java
```diff
@@ -143,6 +145,185 @@ static String bulkLoadRefreshParam() {
...
+     *
+     * <p>Paths are returned as ordered part lists, e.g. {@code ["location", "point"]} for a {@code
+     * geo_point} sub-field of an object {@code location}, so the bulk-source strip can remove the
+     * exact nested value while preserving unaffected siblings ({@code location.city} etc.). A
+     * top-level unsupported field comes back as a single-element list, e.g. {@code
+     * ["nested_value"]}.
+     *
+     * @return the dropped field paths (empty when disabled or nothing matched)
+     */
+    static Set<List<String>> stripUnsupportedMappingFields(JSONObject jsonObject) {
```
dai-chen, line 215: I may miss it somewhere. Could you point me where we gate this by AE enabled setting?

dai-chen:
> > Is all our IT CI broken? Just wonder how to verify these changes..
> 
> we should fix the CIs besides this PR but I verified locally of
> 
> Run AnalyticsUnsupportedFieldStripVerifyIT on a force-routed AE cluster with -Dtests.analytics.parquet_indices=true. It loads every Index dataset through the real load path and asserts per dataset: clean mapping, no stripped path left in _source, doc count == source doc count. It passes, and fails loudly/attributably if any dataset still ingests an unsupported field, so it's the durable proof artifact in the PR.
> 
> Also: unit tests 7/7 (nested paths + arrays of objects), and legacy route with no parquet flag 123 pass / 0 fail so nothing regresses when AE is off.

I thought our CI is failing on dependency after version bump. But it seems due to this single test:

```
2026-06-11T22:45:47.0923183Z Tests with failures:
2026-06-11T22:45:47.0924323Z  - org.opensearch.sql.sql.QueryValidationIT.testAliasToKeywordMultiFieldFailsWithBadRequest
```

mengweieric:
> > > Is all our IT CI broken? Just wonder how to verify these changes..
> > 
> > 
> > we should fix the CIs besides this PR but I verified locally of
> > Run AnalyticsUnsupportedFieldStripVerifyIT on a force-routed AE cluster with -Dtests.analytics.parquet_indices=true. It loads every Index dataset through the real load path and asserts per dataset: clean mapping, no stripped path left in _source, doc count == source doc count. It passes, and fails loudly/attributably if any dataset still ingests an unsupported field, so it's the durable proof artifact in the PR.
> > Also: unit tests 7/7 (nested paths + arrays of objects), and legacy route with no parquet flag 123 pass / 0 fail so nothing regresses when AE is off.
> 
> I thought our CI is failing on dependency after version bump. But it seems due to this single test:
> 
> ```
> 2026-06-11T22:45:47.0923183Z Tests with failures:
> 2026-06-11T22:45:47.0924323Z  - org.opensearch.sql.sql.QueryValidationIT.testAliasToKeywordMultiFieldFailsWithBadRequest
> ```

yea fixed in this PR: https://github.com/opensearch-project/sql/pull/5545

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteAliasFieldAggregationIT.java
mengweieric: Good Catch. Moved.

integ-test/src/test/java/org/opensearch/sql/legacy/TestUtils.java
mengweieric: the call site is intentionally unconditional, but the actual gate is inside `AnalyticsIndexConfig`

PR #5532, opened 2026-06-09
[BugFix] Fix invalid field or index error misclassified as internal failures

dai-chen:
### Description

The unified query path returned a 500 instead of a client error for queries referencing a non-existent index or field, because `ErrorReport` (field-not-found) and `CalciteException` (table-not-found) fell through to the catch-all and were rethrown as `IllegalStateException`. This PR fixes that by propagating `ErrorReport` unwrapped and mapping `CalciteException` to `SemanticCheckException`.

#### Example: Non-existent Index

```
  Request:
  POST /_plugins/_sql
  {"query": "SELECT * FROM nonexistent_index_xyz LIMIT 1"}

  Before:
  HTTP/1.1 500 Internal Server Error
  {
    "error": {
      "reason": "There was internal problem at backend",
      "details": "Failed to plan query: unexpected error",
      "type": "IllegalStateException"
    },
    "status": 500
  }

  After:
  HTTP/1.1 400 Bad Request
  {
    "error": {
      "reason": "Invalid SQL query",
      "details": "Table 'nonexistent_index_xyz' not found",
      "type": "SemanticCheckException"
    },
    "status": 400
  }
```
 

#### Example: Non-existent Field

```
  Request:
  POST /_plugins/_sql
  {"query": "SELECT nonexistent_field FROM sanity_test"}

  Before:
  HTTP/1.1 500 Internal Server Error
  {
    "error": {
      "reason": "There was internal problem at backend",
      "details": "Failed to plan query: unexpected error",
      "type": "IllegalStateException"
    },
    "status": 500
  }

  After:
  HTTP/1.1 400 Bad Request
  {
    "error": {
      "reason": "Invalid SQL query",
      "details": "Field [nonexistent_field] not found.",
      "type": "IllegalArgumentException"
    },
    "status": 400
  }
```

### Related Issues
Part of https://github.com/opensearch-project/sql/issues/5248

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
 - [x] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
[...]

PR #5545, opened 2026-06-11
fix(integ): repair two pre-existing IT failures on main (QueryValidationIT error type, NoMv explain flake)

mengweieric:
### Description

Repairs two pre-existing IT failures on `main` (both in `integ-test`, test-only, no production change).

**1. `QueryValidationIT.testAliasToKeywordMultiFieldFailsWithBadRequest`** — #5532 changed the SQL error formatter to unwrap `ErrorReport` to its cause, so the reported `type` is now the underlying `SemanticCheckException`, not the `ErrorReport` wrapper. The query is still correctly rejected with 400; only the error-envelope `type` changed. Updated the expected type to match.

**2. `CalciteExplainIT.testNoMvBasic` / `testNoMvWithEval`** — the assertion scanned the full logical+physical explain YAML for `ARRAY_JOIN`, but the **physical** section renders differently when pushdown is disabled (`CalciteNoPushdownIT`), making the test flaky on some runners. `nomv` lowers to `MVJOIN → ARRAY_JOIN`, which is stable in the **logical** plan regardless of pushdown (verified with pushdown on and off). Assert on the logical section only.

### Testing

Ran against an OpenSearch cluster:
- `QueryValidationIT` — full class passes (target test + siblings).
- `CalciteExplainIT.testNoMvBasic` / `testNoMvWithEval` — pass; logical plan confirmed to contain `ARRAY_JOIN` with pushdown both enabled and disabled.

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteExplainIT.java
@@ -2888,10 +2888,13 @@ public class CalciteExplainIT extends ExplainIT {
             "source=%s | fields firstname, age | eval names = array(firstname) | nomv names |"
                 + " fields names",
             TEST_INDEX_BANK);
-    var result = explainQueryYaml(query);
+    // Assert on the LOGICAL plan only: nomv lowers to MVJOIN -> ARRAY_JOIN, which is stable in the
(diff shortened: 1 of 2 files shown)


dai-chen (approved):
Thanks for the fix!

ahkcs (approved):

PR #5546, opened 2026-06-12
Bring CalciteWhereCommandIT to parity on the analytics-engine route

ahkcs:
### Description

Brings `CalciteWhereCommandIT` to parity on the analytics-engine route (`:integ-test:integTestRemote -Dtests.analytics.parquet_indices=true`): 9 failing → 0, by branching the tests that exercise behavior the route inherently can't support behind `isAnalyticsParquetIndicesEnabled()`. All branches are no-ops off the analytics route, so the v2 path is unchanged.

### Pass rate

| Route | Before | After |
|---|---|---|
| analytics-engine | 9 failing / 41 | **0 failing — 32 pass, 9 documented skips** |
| v2 (`:integ-test:integTest`, fresh cluster) | 41 pass | **41 pass, 0 skip** |

### Documented skips (inherent analytics-route divergences)

| Test(s) | Reason |
|---|---|
| `testFilterOnComputedNestedFields`, `testFilterOnNestedAndRootFields`, `testFilterOnNestedFields`, `testFilterOnMultipleCascadedNestedFields`, `testScriptFilterOnDifferentNestedHierarchyShouldThrow`, `testAggFilterOnNestedFields` | These query nested fields (`address.city`, `projects.name`, `author.books.*`). The parquet/composite store has no nested-document support, so nested fields are stripped from the dataset at load (#5541) and the fields don't exist on this route. |
| `testWhereWithMetadataFields`, `testWhereWithMetadataFields2` | Use `where _id='1'`. The analytics-engine route doesn't expose the `_id` metadata field — parquet-backed scans surface only mapped document fields. |
| `testDoubleEqualWithSpecialCharacters` | `email` is *dynamically* mapped to a `text` field **without** a `.keyword` sub-field on the parquet route (the composite dataformat's dynamic mapping omits the keyword sub-field that standard OpenSearch adds). Exact equality on an analyzed text field then returns no rows on the DataFusion scan. `firstname` (explicit `text`+`keyword`) still exercises `==` on the analytics route. |

### Note for reviewers

The `testDoubleEqualWithSpecialCharacters` root cause is broader than this one test: **any exact-match (`=` / `==`) on a dynamically-mapped string field silently returns 0 rows on the analytics-engine route**, because dynamic strings get no `.keyword` sub-field there. Worth tracking as a potential analytics-engine fix (have dynamic string mapping add the keyword sub-field, or make text-field equality fall back to the stored value) rather than relying on every fixture to map string fields explicitly.

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
@@ -5,10 +5,10 @@
 
 package org.opensearch.sql.calcite.remote;
 
-import static org.junit.Assume.assumeFalse;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_CASCADED_NESTED;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_DEEP_NESTED;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_NESTED_SIMPLE;
+import static org.opensearch.sql.util.AnalyticsRouteLimitation.NESTED_FIELDS;
 import static org.opensearch.sql.util.MatcherUtils.rows;
 import static org.opensearch.sql.util.MatcherUtils.schema;
 import static org.opensearch.sql.util.MatcherUtils.verifyDataRows;
@@ -24,11 +24,6 @@ import org.opensearch.sql.common.utils.StringUtils;
 import org.opensearch.sql.ppl.WhereCommandIT;
 
 public class CalciteWhereCommandIT extends WhereCommandIT {
-
-  private static final String NESTED_UNSUPPORTED_ON_AE =
-      "Nested-field queries can't run on the analytics-engine route: the parquet/composite store"
-          + " has no nested-document support, so nested fields are stripped from the dataset at"
-          + " load.";
 
   @Override
   public void init() throws Exception {
@@ -47,7 +42,7 @@ public class CalciteWhereCommandIT extends WhereCommandIT {
 
   @Test
   public void testFilterOnComputedNestedFields() throws IOException {
-    assumeFalse(NESTED_UNSUPPORTED_ON_AE, isAnalyticsParquetIndicesEnabled());
+    assumeNotAnalytics(NESTED_FIELDS);
     JSONObject result =
         executeQuery(
             String.format(
@@ -60,7 +55,7 @@ public class CalciteWhereCommandIT extends WhereCommandIT {
 
   @Test
   public void testFilterOnNestedAndRootFields() throws IOException {
-    assumeFalse(NESTED_UNSUPPORTED_ON_AE, isAnalyticsParquetIndicesEnabled());
+    assumeNotAnalytics(NESTED_FIELDS);
     JSONObject result =
         executeQuery(
             String.format(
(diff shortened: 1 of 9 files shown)


integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
```diff
@@ -40,6 +47,7 @@ protected String getIncompatibleTypeErrMsg() {
 
   @Test
   public void testFilterOnComputedNestedFields() throws IOException {
+    assumeFalse(NESTED_UNSUPPORTED_ON_AE, isAnalyticsParquetIndicesEnabled());
```
dai-chen: I thought we'd like to maintain the exclude list in gradle. Or any other benefits with this?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
```diff
@@ -23,6 +24,12 @@
 import org.opensearch.sql.ppl.WhereCommandIT;
 
 public class CalciteWhereCommandIT extends WhereCommandIT {
+
+  private static final String NESTED_UNSUPPORTED_ON_AE =
```
Swiddis: can this be public since it's relevant to other tests too

integ-test/src/test/java/org/opensearch/sql/ppl/WhereCommandIT.java
```diff
@@ -216,6 +217,10 @@ public void testIsNotNullPredicate() throws IOException {
 
   @Test
   public void testWhereWithMetadataFields() throws IOException {
+    assumeFalse(
+        "The analytics-engine route doesn't expose the _id metadata field (parquet-backed scans"
```
Swiddis: For all the AE assumptions, if we're sticking with this (instead of chen's gradle exclude), we should maybe put this in one global enum so it's all trackable at once, without doing a bunch of string copying + searching. We can also just feed agents that enum as context for bulk-debugging.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
ahkcs: Done — I rolled it (and the other route reasons) into a shared `AnalyticsRouteLimitation` enum in `org.opensearch.sql.util`, so it's reusable across `ppl/` and `calcite/` without each test owning its own string. The nested-field reason is now `AnalyticsRouteLimitation.NESTED_FIELDS` and this test reads `assumeNotAnalytics(NESTED_FIELDS)`.

integ-test/src/test/java/org/opensearch/sql/ppl/WhereCommandIT.java
ahkcs: pushed a follow-up that adds a single `AnalyticsRouteLimitation` enum holding every analytics-route skip reason, plus an `assumeNotAnalytics(LIMITATION)` helper on `PPLIntegTestCase`. Skips now read as `assumeNotAnalytics(ID_METADATA)` / `assumeNotAnalytics(DYNAMIC_STRING_NO_KEYWORD)` instead of inline strings.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteWhereCommandIT.java
ahkcs: Added the exclusion list

Swiddis (approved):

PR #5561, opened 2026-06-17
Stabilize PPL ITs on the analytics-engine route (case/string/full-text/like/appendpipe/datatype)

ahkcs:
### Description

Analytics-engine route (`-Dtests.analytics.parquet_indices=true`) parity pass across several PPL IT classes. All changes are test-only; the v2/Calcite route is unchanged (every gated test still runs there — the `assumeNotAnalytics(...)` guards are no-ops off-route and the gradle excludes apply only to `integTestRemote`).

#### CalcitePPLCaseFunctionIT (3/9 → 8/9 pass, 1 excluded)
- **Accumulation:** `appendDataForBadResponse()` raw-PUTs 4 weblogs docs unconditionally in `init()`; the append-only AE store can't replace on same-`_id` PUT, so they accumulated per method. Guarded on a pre-`loadIndex` `isIndexExist` check.
- **otel_logs multi-value reject:** the dataset has a multi-value `attributes.email.invalid_recipients` array the parquet store rejects (`Cannot accept multiple values for field ... of type keyword`), aborting `init()`. Only `testNestedCaseAggWithAutoDateHistogram` uses it, so the load is skipped on the AE route.
- **`BIN_TIME_FIELD_BUCKETING`** (existing): `bin @timestamp | stats by @timestamp` returns the bucket column typed `string` not `timestamp`. Skips `testNestedCaseAggWithAutoDateHistogram`.

#### CalcitePPLStringBuiltinFunctionIT (7 tests skipped — `DOC_MUTATION`)
`testConcatWithField/ConcatWs/Reverse/Right/Trim/RTrim/LTrim` re-PUT a shared `_id` (5/6/7) with **different** data, relying on PUT-replace. On the AE store same-`_id` PUT **appends** and **DELETE is unsupported**, so these accumulate and cross-contaminate (counts are order-dependent). Skipped via the existing `DOC_MUTATION` limitation.

#### Full-text relevance functions (3 tests skipped — new `FULLTEXT_RELEVANCE_FUNC`)
`MultiMatchIT.test_wildcard_multi_match`, `QueryStringIT.wildcard_test`, `SimpleQueryStringIT.test_wildcard_simple_query_string` use `multi_match` / `query_string` / `simple_query_string` — Lucene relevance functions with no DataFusion equivalent (they return no rows).

#### CalciteLikeQueryIT.test_the_default_3rd_option (skipped — new `LIKE_CASE_SENSITIVITY`)
The v3 branch expects case-sensitive `LIKE` (0 rows for `'test Wildcard%'` vs lowercase data); the AE route's `LIKE` is case-insensitive (DataFusion) and returns 7.

#### CalcitePPLAppendPipeCommandIT.testDoubleAppendPipeWithFilter (skipped — new `APPENDPIPE_MAIN_RESULT_DROPPED`)
`appendpipe [subpipe]` drops the main pipeline's rows on the AE route: the subpipe's filter is applied to the main result instead of being appended, so the originals are lost (verified: `stats ... | appendpipe [where gender='F']` returns only the F rows, not the originals plus the filtered copy).

#### DataTypeIT exclude globs (coverage fix)
`test_nonnumeric_data_types`, `test_alias_data_type`, and `SystemFunctionIT.typeof_opensearch_types` were excluded with `org.opensearch.sql.ppl.*` globs that did **not** match the Calcite subclasses. Broadened to `*` so `CalciteDataTypeIT` / `CalciteSystemFunctionIT` are covered too.

### Results (analytics route, per-listed-test)

| Test | Before | After |
|---|---|---|
| CalcitePPLCaseFunctionIT.testCaseWhen{InFilter,InSubquery,WithCast,WithIn} | fail (accumulation) | pass |
| CalcitePPLStringBuiltinFunctionIT.test{LTrim,Reverse,Trim} | fail (accumulation) | excluded (DOC_MUTATION) |
| CalciteMultiMatchIT.test_wildcard_multi_match | fail | excluded |
| CalciteQueryStringIT.wildcard_test | fail | excluded |
| CalciteSimpleQueryStringIT.test_wildcard_simple_query_string | fail | excluded |
| CalciteLikeQueryIT.test_the_default_3rd_option | fail | excluded |
| CalcitePPLAppendPipeCommandIT.testDoubleAppendPipeWithFilter | fail | excluded |
| CalciteDataTypeIT.test_nonnumeric_data_types | fail | excluded (glob fix) |

v2/Calcite route: all edited classes 100% pass, 0 skips.

> **Not included** (deferred — investigated but not cleanly fixable here): `CalciteChartCommandIT` has 5 analytics-route failures (its own multi-bucket triage, partly the known otel_logs ip-scan defect), and `NfwPplDashboardIT.testTopLongLivedTCPFlows` returns 0 rows via the IT harness but the identical query returns the correct 2 rows in a direct probe — its cause isn't reproducible/classifiable from outside the harness, so it's left for separate investigation rather than shipping a mislabeled skip.

### Check List

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/util/Capability.java
+++ b/integ-test/src/test/java/org/opensearch/sql/util/Capability.java
@@ -192,7 +192,35 @@ public enum Capability {
   EVAL_MAX_MIN_INT_WIDENING(
       "eval max()/min() over integer operands reports the result column as bigint on the"
           + " analytics-engine route (DataFusion widens integers to Int64), whereas the v2/Calcite"
-          + " path reports int.");
+          + " path reports int."),
+
+  /**
+   * Lucene full-text relevance functions ({@code match}, {@code multi_match}, {@code query_string},
+   * {@code simple_query_string}, …) are unsupported on the analytics-engine route — DataFusion has
+   * no relevance scorer, so a query that filters on one returns no rows.
+   */
+  FULLTEXT_RELEVANCE_FUNC(
+      "Full-text relevance functions (match/multi_match/query_string/simple_query_string) are"
+          + " unsupported on the analytics-engine route: DataFusion has no relevance scorer, so the"
+          + " filter returns no rows."),
+
+  /**
+   * LIKE is case-insensitive on the analytics-engine route (DataFusion), whereas the v2/Calcite
+   * path treats {@code LIKE} as case-sensitive (only {@code ILIKE} is case-insensitive).
+   */
+  LIKE_CASE_SENSITIVITY(
+      "LIKE is case-insensitive on the analytics-engine route (DataFusion), whereas the v2/Calcite"
+          + " path treats LIKE as case-sensitive."),
+
+  /**
+   * {@code appendpipe [subpipe]} drops the main pipeline's rows on the analytics-engine route: the
+   * subpipe's filter is applied to the main result instead of its output being appended to it, so
+   * the original rows are lost (e.g. {@code stats ... | appendpipe [where gender='F']} returns only
+   * the filtered F rows, not the originals plus the filtered copy).
+   */
+  APPENDPIPE_MAIN_RESULT_DROPPED(
+      "appendpipe drops the main pipeline's rows on the analytics-engine route: the subpipe filter"
+          + " is applied to the main result instead of appended, so the originals are lost.");
 
   private final String reason;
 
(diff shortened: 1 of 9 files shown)


integ-test/src/test/java/org/opensearch/sql/util/Capability.java
```diff
@@ -192,7 +192,35 @@ public enum Capability {
...
+      "LIKE is case-insensitive on the analytics-engine route (DataFusion), whereas the v2/Calcite"
+          + " path treats LIKE as case-sensitive."),
+
+  /**
+   * {@code appendpipe [subpipe]} drops the main pipeline's rows on the analytics-engine route: the
+   * subpipe's filter is applied to the main result instead of its output being appended to it, so
+   * the original rows are lost (e.g. {@code stats ... | appendpipe [where gender='F']} returns only
+   * the filtered F rows, not the originals plus the filtered copy).
+   */
+  APPENDPIPE_MAIN_RESULT_DROPPED(
```
dai-chen, line 221: For AE, is this a bug or missing capability?

ahkcs: It's a delegation bug for AE side

Swiddis (approved):

Issue #5500, opened 2026-06-02
[RFC] Query Insights: Profile and Analyze for PPL/SQL


anasalkouz:
## Problem Statement

PPL/SQL users have no way to understand **why** a query is slow or how to optimize it. Customers report slow queries, browser crashes, and stuck searches with no self-service diagnostics. The raw building blocks exist independently (estimated plan costs via `explain cost`, actual execution metrics via the `profile` API), but they are separate API calls returning different formats that only engine developers can interpret. There is no unified view comparing estimated vs. actual metrics, no human-readable interpretation, and no optimization suggestions.

This RFC directly addresses [#4343 (Support analyze alongside explain)](https://github.com/opensearch-project/sql/issues/4343).

## Current State

- **`explain cost`** provides estimated row counts and plan costs via Calcite's cost model, but only as raw plan text.
- **The `profile` API** ([#4983](https://github.com/opensearch-project/sql/pull/4983), [#5044](https://github.com/opensearch-project/sql/pull/5044)) provides actual execution time and row counts per operator as JSON.
- These two data sources are completely disconnected -- users must call them separately and mentally correlate operator nodes between different formats.
- No UI integration exists in Discover to surface this information.
- Competing tools (PostgreSQL `EXPLAIN ANALYZE`, BigQuery plan explanations, Kusto best-practices analyzer) provide unified estimated-vs-actual views natively.

## Long-Term Goals

- **Unified query diagnostics**: A single API call that returns both estimated and actual metrics per operator, enabling users to identify bottlenecks without needing engine-level expertise.
- **Self-service optimization**: Rule-based heuristic recommendations that surface actionable suggestions (e.g., "add a time filter", "this aggregation runs in-memory on 50M rows").
- **Extensible rule framework**: An `InsightRule` interface that allows contributors to add new optimization heuristics without modifying core analyzer logic.
- **UI integration**: A visual panel in Discover that renders the operator tree and recommendations, making query profiling accessible to all users.

## Proposal

Add an **EXPLAIN ANALYZE** endpoint that merges the existing `explain cost` and `profile` code paths into a unified operator tree, augmented by a rule-based **QueryInsightsAnalyzer** that produces human-readable optimization recommendations.

### Response Schema

```json
{
  "query": "source=my_index | where age > 30 | stats count() by city | sort count",
  "total_time_ms": 342,
  "operator_tree": [
    {
      "id": 1,
      "node_name": "OpenSearchIndexScan",
[...]

PR #5568, opened 2026-06-19
Overriding profile endpoint with analyze endpoint with operator tree and profiling

### LINKED ISSUE #4343: [FEATURE] Support `analyze` alongside `explain`
In most SQL databases, there are actually two flavors of explanation: `EXPLAIN` and `EXPLAIN ANALYZE`.

Explain:

```sql
EXPLAIN SELECT *
FROM tenk1 t1, tenk2 t2
WHERE t1.unique1 < 100 AND t1.unique2 = t2.unique2
ORDER BY t1.fivethous;
```

```
                                     QUERY PLAN
------------------------------------------------------------------------------------------------
 Sort  (cost=717.34..717.59 rows=101 width=488)
   Sort Key: t1.fivethous
   Sort Method: quicksort  Memory: 77kB
   ->  Hash Join  (cost=230.47..713.98 rows=101 width=488)
         Hash Cond: (t2.unique2 = t1.unique2)
         ->  Seq Scan on tenk2 t2  (cost=0.00..445.00 rows=10000 width=244)
         ->  Hash  (cost=229.20..229.20 rows=101 width=244)
               Buckets: 1024  Batches: 1  Memory Usage: 28kB
               ->  Bitmap Heap Scan on tenk1 t1  (cost=5.07..229.20 rows=101 width=244)
                     Recheck Cond: (unique1 < 100)
                     ->  Bitmap Index Scan on tenk1_unique1  (cost=0.00..5.04 rows=101 width=0)
                           Index Cond: (unique1 < 100)
```

Analyze:

```sql
EXPLAIN ANALYZE SELECT *
FROM tenk1 t1, tenk2 t2
WHERE t1.unique1 < 100 AND t1.unique2 = t2.unique2
ORDER BY t1.fivethous;
```

```
                                                                 QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=717.34..717.59 rows=101 width=488) (actual time=7.761..7.774 rows=100 loops=1)
   Sort Key: t1.fivethous
   Sort Method: quicksort  Memory: 77kB
   ->  Hash Join  (cost=230.47..713.98 rows=101 width=488) (actual time=0.711..7.427 rows=100 loops=1)
         Hash Cond: (t2.unique2 = t1.unique2)
         ->  Seq Scan on tenk2 t2  (cost=0.00..445.00 rows=10000 width=244) (actual time=0.007..2.583 rows=10000 loops=1)
         ->  Hash  (cost=229.20..229.20 rows=101 width=244) (actual time=0.659..0.659 rows=100 loops=1)
               Buckets: 1024  Batches: 1  Memory Usage: 28kB
               ->  Bitmap Heap Scan on tenk1 t1  (cost=5.07..229.20 rows=101 width=244) (actual time=0.080..0.526 rows=100 loops=1)
                     Recheck Cond: (unique1 < 100)
                     ->  Bitmap Index Scan on tenk1_unique1  (cost=0.00..5.04 rows=101 width=0) (actual time=0.049..0.049 rows=100 loops=1)
                           Index Cond: (unique1 < 100)
 Planning time: 0.194 ms
 Execution time: 8.008 ms
```

The difference is that `ANALYZE` also provides actual timings, making it possible to benchmark the internals of queries without needing to set up an external profiler.

The ability to analyze queries in explanation.

Keep using external profilers. (Generally necessary for larger-scale throughput benchmarking anyway, but for smaller cases, individually queries can be debugged.)

N/A

Krish-Gandhi:
### Description

- Introduces the `analyze` endpoint for PPL queries, which can be activated by passing `"analyze": true` as a request body parameter. 
- Overrides the existing `profile` endpoint to run `analyze` functionality.
- Propagate PPL`analyze` flag through transport and request parsing.
- Combines logical plan and physical plan nodes by walking each tree and grouping nodes into an `operator_tree`.
- Maps PPL query segments and estimated/actual rows to `operator_tree` nodes.
- Combined current `profile` `plan` response with `operator_tree` nodes to calculate time taken by each node.

Note: This PR provides a simple implementation of creating the `operator_tree`, which works for simpler queries. The logic doesn't hold for queries that produce non-linear physical plan trees (for example, `JOIN`s). This only affects the `operator_tree`, meaning the subset of the response that corresponds to `profile` is 100% functional. Therefore, in this scenario, the `analyze` endpoint will "fallback" to the `profile` output by returning a response mirroring the `profile` endpoint (by only including the `profile`, `schema`, `datarows`, `total`, and `size` fields in the response). This `operator_tree` issue will be addressed in a later PR. 

---
### Important

This PR overrides the existing `profile` endpoint by routing all requests with either `"analyze": true` or `"profile": true` (or both) to `pplService.analyze()` in `TransportPPLQueryAction.java`. For current end-users of the `profile` endpoint, this will make little difference, as the response from `profile` is a subset of the response from `analyze`. In simpler terms, ***current end-users can make no changes*** and have similar results. 

All of the existing code for `profile` still exists and is in place. This makes it ***extremely simple to separate `analyze` and `profile`*** again. This can be done by removing the ` || transformedRequest.profile()` part of the boolean expression in line 200 of `TransportPPLQueryAction.java`.

`plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java:193-209`:
```.java
    if (transformedRequest.isExplainRequest()) {
      pplService.explain(
          transformedRequest, createExplainResponseListener(transformedRequest, clearingListener));
    /**
     * Removing  `|| transformedRequest.profile()` from line 200 will
     * separate the `profile` and `analyze` endpoints. See PR #5568.
     */
    } else if (transformedRequest.analyze() || transformedRequest.profile()) {
      pplService.analyze(
        transformedRequest, createAnalyzeResponseListener(transformedRequest, clearingListener));
    } else {
      pplService.execute(
          transformedRequest,
          createListener(transformedRequest, clearingListener),
          createExplainResponseListener(transformedRequest, clearingListener));
    }
  }
```

---
### Example Query and Response

The following `curl` command will run the query `source=accounts | where age < 30 | eval full_name = firstname + \" \" + lastname | fields full_name, email, age` on the `analyze` endpoint:
```
  -d '{"query": "source=accounts | where age < 30 | eval full_name = firstname + \" \" + lastname | fields full_name, email, age", "analyze": true}'
```
The response of this will be as follows. (NOTE: The `"logicalPlan"` and `"physicalPlan"` fields are included for debugging purposes and should not be included in the final version of this endpoint.)
```
{
  "query": "source=accounts | where age < 30 | eval full_name = firstname + \" \" + lastname | fields full_name, email, age",
  "querySegments": [
    {
      "nodeType": "SearchFrom",
      "source": "source=accounts"
    },
    {
      "nodeType": "WhereCommand",
      "source": "where age < 30"
    },
    {
      "nodeType": "EvalCommand",
      "source": "eval full_name = firstname + \" \" + lastname"
    },
    {
      "nodeType": "FieldsCommand",
      "source": "fields full_name, email, age"
    }
  ],
  "logicalPlan": [
    "LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT]): rowcount = 5000.0, cumulative cost = {114000.0 rows, 145000.0 cpu, 0.0 io}, id = 103",
    "LogicalProject(full_name=[||(||($0, ' '), $4)], email=[$3], age=[$2]): rowcount = 5000.0, cumulative cost = {109000.0 rows, 25000.0 cpu, 0.0 io}, id = 102",
    "LogicalFilter(condition=[<($2, 30)]): rowcount = 5000.0, cumulative cost = {104000.0 rows, 10000.0 cpu, 0.0 io}, id = 100",
    "CalciteLogicalIndexScan(table=[[OpenSearch, accounts]]): rowcount = 10000.0, cumulative cost = {99000.0 rows, 0.0 cpu, 0.0 io}, id = 99"
  ],
  "physicalPlan": [
    "EnumerableCalc(expr#0..3=[{inputs}], expr#4=[' '], expr#5=[||($t0, $t4)], expr#6=[||($t5, $t3)], full_name=[$t6], email=[$t2], age=[$t1]): rowcount = 5000.0, cumulative cost = {22996.4 rows, 50000.0 cpu, 0.0 io}, id = 193",
    "CalciteEnumerableIndexScan(table=[[OpenSearch, accounts]], PushDownContext=[[PROJECT->[firstname, age, email, lastname], FILTER-><($1, 30), LIMIT->10000], OpenSearchRequestBuilder(sourceBuilder={\"from\":0,\"size\":10000,\"timeout\":\"1m\",\"query\":{
  ],
  "profile": {
    "summary": {
      "total_time_ms": 777.18
    },
    "phases": {
      "analyze": {
        "time_ms": 477.57
      },
      "optimize": {
        "time_ms": 225.46
      },
      "execute": {
        "time_ms": 73.41
      },
      "format": {
        "time_ms": 0.11
      }
    },
    "plan": {
      "node": "EnumerableCalc",
      "time_ms": 64.19,
      "rows": 3,
      "children": [
        {
          "node": "CalciteEnumerableIndexScan",
          "time_ms": 64.0,
          "rows": 3
        }
      ]
    }
  },
  "operator_tree": [
    {
      "source": "source=accounts | where age < 30",
      "node_type": [
        "SearchFrom",
        "WhereCommand"
      ],
      "description": [
        "CalciteLogicalIndexScan(table=[[OpenSearch, accounts]]): rowcount = 10000.0, cumulative cost = {99000.0 rows, 0.0 cpu, 0.0 io}, id = 99",
        "LogicalFilter(condition=[<($2, 30)]): rowcount = 5000.0, cumulative cost = {104000.0 rows, 10000.0 cpu, 0.0 io}, id = 100"
      ],
      "estimated_rows": 5000,
      "actual_time_ms": "64.00 ms",
      "actual_rows": 3,
      "is_pushed_down": true
    },
    {
      "source": "eval full_name = firstname + \" \" + lastname | fields full_name, email, age",
      "node_type": [
        "EvalCommand",
        "FieldsCommand"
      ],
      "description": [
        "LogicalProject(full_name=[||(||($0, ' '), $4)], email=[$3], age=[$2]): rowcount = 5000.0, cumulative cost = {109000.0 rows, 25000.0 cpu, 0.0 io}, id = 102"
      ],
      "estimated_rows": 5000,
      "actual_time_ms": "0.19 ms",
      "actual_rows": 3
    }
  ],
  "recommendations": [
    {
      "serverity": "INFO",
      "rule": "Pushdown visibility",
      "message": "1 of 2 stages pushed down; 1 ran in-memory"
    },
    {
      "serverity": "INFO",
      "rule": "Bottleneck stage",
      "message": "87% of time is in the *SearchFrom, WhereCommand* stage",
      "affected_node": "source=accounts | where age < 30",
      "suggestion": "Consider optimizing the SearchFrom, WhereCommand operation"
    }
  ],
  "schema": [
    {
      "name": "full_name",
      "type": "STRING"
    },
    {
      "name": "email",
      "type": "STRING"
    },
    {
      "name": "age",
      "type": "LONG"
    }
  ],
  "datarows": [
    [
      "Jane Smith",
      "jane@example.com",
      28
    ],
    [
      "Kyle Miller",
      "goat@example.com",
      22
    ],
    [
      "Joette Kap",
      "coast2coast@example.com",
      22
    ]
  ],
  "total": 3,
  "size": 3
}
```
### Performance of `analyze`

After writing a benchmarking script to run 20 queries on a sample 10 GB dataset, the results were as follows. This shows that analyze has a small enough overhead to justify running it by default on the Discover tab on the OpenSearch Dashboard.
```
=== PPL Query Benchmark ===

--- Results ---

normal      avg=4489.8ms  p50=206.9ms  p95=1083.1ms  min=28.5ms  max=84080.9ms
profile     avg=4407.9ms  p50=197.9ms  p95=749.1ms  min=19.1ms  max=83598.1ms
analyze     avg=4410.2ms  p50=213.3ms  p95=801.3ms  min=38.3ms  max=83146.3ms
```

### Related Issues
#5500 
Resolves #4343  

### Check List


ahkcs:
Let's update the `endpoint.md`(https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/interfaces/endpoint.md) and `integ-test/src/yamlRestTest/resources/rest-api-spec/test/api/ppl.profile.yml` file for our profile API change

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java
```diff
@@ -209,6 +193,13 @@ protected void doExecute(
     if (transformedRequest.isExplainRequest()) {
       pplService.explain(
           transformedRequest, createExplainResponseListener(transformedRequest, clearingListener));
+      /**
+       * Removing `|| transformedRequest.profile()` from line 200 will separate the `profile` and
+       * `analyze` endpoints. See PR #5568.
+       */
+    } else if (transformedRequest.analyze() || transformedRequest.profile()) {
```
ahkcs, line 219: We might need to be careful about a silent API break here

Previously `profile:true` fell through to `pplService.execute()` and was formatted by `SimpleJsonResponseFormatter`, now the same request returns the new `AnalyzeResponse` shape

Existing consumers(OSD and AE) may experience breaking change

core/src/main/java/org/opensearch/sql/executor/execution/AnalyzePlan.java
```diff
@@ -0,0 +1,54 @@
...
+    this.query = query;
+    this.querySegments = querySegments;
+    this.plan = plan;
+    this.queryService = queryService;
+    this.listener = listener;
+  }
+
+  @Override
+  public void execute() {
+    queryService.analyzeWithCalcite(query, querySegments, plan, getQueryType(), listener);
```
ahkcs, line 46: This line calls `analyzeWithCalcite` directly with no `shouldUseCalcite(getQueryType())` gate, we need to make sure that it has proper fallback when calcite is not enabled(maybe we preserve the original profile API for V2 path)

ahkcs:
Let's also document the current limitation for `join`, also we can provide fallback option to original profile API when we meet limitations

core/src/main/java/org/opensearch/sql/executor/execution/AnalyzePlan.java
Krish-Gandhi: Added `shouldUseCalcite` to `core/src/main/java/org/opensearch/sql/executor/QueryService.java`. Will now return exception when Calcite is disabled. This is acceptable because `profile` only runs when Calcite is enabled, according to `endpoint.md`.

Added "fallback" to `profile` endpoint on complex queries, see note in PR description for more detail.

Krish-Gandhi:
Updated `docs/user/ppl/interfaces/endpoint.md` to add documentation for current state of `analyze` and update note for `profile` endpoint, as well as a note about when the logic of `analyze` falls apart and fallback to `profile`. Also updated `integ-test/src/yamlRestTest/resources/rest-api-spec/test/api/ppl.profile.yml` and added `integ-test/src/yamlRestTest/resources/rest-api-spec/test/api/ppl.analyze.yml`.

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java
Krish-Gandhi: Should be okay, as the output of the previous `profile` endpoint is intentionally a subset of the `analyze` response.

ahkcs (approved):

dai-chen (commented):
High level questions:

1. Is it possible to avoid adding more top-level fields like `querySegments`, `logicalPlan`, `physicalPlan`, `operator_tree` and `recommendations`. Any idea how to group them better? Like the first 3 more like explain output and the last 2 belong to `profile`?
2. How to extend this to SQL language and Analytics engine backend? From what I see, `querySegments` may not work for SQL and `operator_tree` is mostly for OpenSearch execution?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -219,15 +219,47 @@ public CalciteRelNodeVisitor(DataSourceService dataSourceService) {
   }
 
   public RelNode analyze(UnresolvedPlan unresolved, CalcitePlanContext context) {
+    if (context.isTrackingEnabled()) {
+      int idBefore = context.relBuilder.size() > 0 ? context.relBuilder.peek().getId() : -1;
+      RelNode result = unresolved.accept(this, context);
+      int idAfter = context.relBuilder.peek().getId();
+      java.util.List<Integer> producedIds = new java.util.ArrayList<>();
```
dai-chen: np: avoid FQN

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -219,15 +219,47 @@ public CalciteRelNodeVisitor(DataSourceService dataSourceService) {
   }
 
   public RelNode analyze(UnresolvedPlan unresolved, CalcitePlanContext context) {
+    if (context.isTrackingEnabled()) {
```
dai-chen, line 235: Since both `analyze` and `visitChildren` are critical and may impact all commands, is it possible to make it well unit tested?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
Krish-Gandhi: Thanks, removing FQNs in lines 226, 241, 249

Krish-Gandhi:
> High level questions:
> 
> 1. Is it possible to avoid adding more top-level fields like `querySegments`, `logicalPlan`, `physicalPlan`, `operator_tree` and `recommendations`. Any idea how to group them better? Like the first 3 more like explain output and the last 2 belong to `profile`?

Agreed, `querySegments`, `logicalPlan` and `physicalPlan` will be removed in a future PR. They are redundant, but left in for now because they are useful for debugging.

> 2. How to extend this to SQL language and Analytics engine backend? From what I see, `querySegments` may not work for SQL and `operator_tree` is mostly for OpenSearch execution?

As far as I can tell, extending to SQL will be interesting for a few reasons, however it (should be) much easier. The main complexity in my implementation for PPL ANALYZE is building the `operator_tree`, i.e. mapping query operations to their physical execution node. Theoretically, this should be much simpler in SQL, because the push down logic is simpler to track and doesn't use Calcite for regular OpenSearch indices. After the `operator_tree` is built, the same recommendations/rules can be applied. This changes if Calcite is integrated more with SQL in the future, but, in that scenario, the current PPL ANALYZE solution is easier to utilize with SQL. SQL on analytics-engine indices already uses Calcite today, so the current implementation should work with minimal changes. SQL segments would be clause-based rather than pipe-based.

For the Analytics engine, we need a profile-to-operator-tree adapter, but the data is already there and it provides more insights than Calcite. Again, after the `operator_tree` is built, we can use the same rules/recommendations.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
Krish-Gandhi: Just pushed some more tests.

dai-chen:
> > High level questions:
> > 
> > 1. Is it possible to avoid adding more top-level fields like `querySegments`, `logicalPlan`, `physicalPlan`, `operator_tree` and `recommendations`. Any idea how to group them better? Like the first 3 more like explain output and the last 2 belong to `profile`?
> 
> Agreed, `querySegments`, `logicalPlan` and `physicalPlan` will be removed in a future PR. They are redundant, but left in for now because they are useful for debugging.
> 
> > 2. How to extend this to SQL language and Analytics engine backend? From what I see, `querySegments` may not work for SQL and `operator_tree` is mostly for OpenSearch execution?
> 
> As far as I can tell, extending to SQL will be interesting for a few reasons, however it (should be) much easier. The main complexity in my implementation for PPL ANALYZE is building the `operator_tree`, i.e. mapping query operations to their physical execution node. Theoretically, this should be much simpler in SQL, because the push down logic is simpler to track and doesn't use Calcite for regular OpenSearch indices. After the `operator_tree` is built, the same recommendations/rules can be applied. This changes if Calcite is integrated more with SQL in the future, but, in that scenario, the current PPL ANALYZE solution is easier to utilize with SQL. SQL on analytics-engine indices already uses Calcite today, so the current implementation should work with minimal changes. SQL segments would be clause-based rather than pipe-based.
> 
> For the Analytics engine, we need a profile-to-operator-tree adapter, but the data is already there and it provides more insights than Calcite. Again, after the `operator_tree` is built, we can use the same rules/recommendations.

Sure, IMO, the response format should be clear and extensible for different language (SQL/PPL) + backend (OS/AE) without breaking changes. It will be very helpful if we can take all these into account in the proposal.

Also could you confirm if we need feature branch if there is a series of PR for this work? Currently we cut release branch from `main` branch directly. Just want to make sure there is no half-done or impact on existing functionalities. Thanks!

PR #5569, opened 2026-06-19
[Enhancement] Classify unsupported-feature errors as client errors (4xx) on the SQL path

RyanL1997:
### Description
On the analytics-engine (Calcite) route, requests for features the engine does **not** implement leak to clients as **HTTP 500 "internal problem at backend"** — for example a SQL `vectorSearch(...)` table function (`CalciteUnsupportedException: Table function is unsupported in Calcite`). A 500 wrongly signals a server fault (prompting client retries) and pollutes the server-error metric / ops alarms, when the truth is "the client asked for something this engine does not support" — a **client error (4xx)**.

The PPL path (`RestPPLQueryAction`) already classifies these as 400; the SQL path (`RestSqlAction`) was missing the equivalent, so the two engines were inconsistent for the **same** exception (e.g. PPL `ad`/`kmeans` → 400, SQL `vectorSearch()` → 500).

This PR makes unsupported-feature errors return a clean 4xx on the SQL path, consistent with PPL, and leverages the existing `ErrorReport`/`ErrorCode` framework so the classification is structured rather than purely exception-type based.

**Changes**
- `RestSqlAction.isClientError()`: add `QueryEngineException` (the parent of `CalciteUnsupportedException`), mirroring `RestPPLQueryAction`.
- `RestSqlAction` / `RestPPLQueryAction` `getRawErrorCode()`: honor an `ErrorReport`'s structured `ErrorCode` (`UNSUPPORTED_OPERATION`, `FIELD_NOT_FOUND`, `PERMISSION_DENIED`, …) when mapping to an HTTP status, instead of always unwrapping to the cause. Backend/unknown codes fall back to the existing cause-classification, so nothing regresses. (This is the `ErrorCode`-based classification the PPL path's existing TODO anticipated.)
- `CalciteRelNodeVisitor.visitTableFunction()`: surface the unsupported table function as a structured `ErrorReport` coded `UNSUPPORTED_OPERATION` with a user-facing suggestion, preserving the `CalciteUnsupportedException` as the cause so `QueryService`'s v2-fallback detection still recognizes it.

### Related Issues

N/A

### Check List


dai-chen (commented):
Please check if we can simply add Calcite unsupported exception to the exception catch in `UnifiedQueryPlanner`. If it works, we can catch all such cases.

RyanL1997:
> Please check if we can simply add Calcite unsupported exception to the exception catch in UnifiedQueryPlanner. If it works, we can catch all such cases.

Hi @dai-chen, good point.

The current approach fits the semantic check because the codebase already defines `SemanticCheckException` to include unsupported-feature errors, and because plan()'s existing contract is to funnel all "invalid for this engine" outcomes into it.

ahkcs (approved):

dai-chen (approved):
Thanks for the fix!

PR #5599, opened 2026-06-30
[Feature] Add PPL `rest` command

### LINKED ISSUE #5597: [RFC] PPL `rest` command
## Problem Statement

PPL can read documents from indices, but it has no way to bring a cluster's
operational and topology state into a query pipeline. Information such as
cluster health, node resource usage, shard placement, cluster settings,
installed plugins, index resolution, and the caller's identity lives behind
dedicated management and `_cat` REST endpoints. To inspect that state today a
user must leave PPL, call the endpoint directly, and post-process the JSON by
hand. There is no way to filter, sort, aggregate, or project management-endpoint
data with the rest of the language (`where`, `stats`, `sort`, `fields`).

## Current State

- On the Calcite path, PPL row sources are limited to index scans
  (`visitRelation`), index subqueries, and literal values.
- `describe <index>` and `show datasources` already expose metadata as tabular
  rows through a reserved-name system source that resolves via `visitRelation`.
  They cover only an index's field metadata and the datasource connection
  catalog; neither reaches cluster or `_cat` operational endpoints.
- The Calcite table-function seam (`visitTableFunction`) is unsupported on the
  primary path: it throws `CalciteUnsupportedException`. A `source=fn(...)`
  style therefore cannot introduce a new row source today.
- Net result: operational endpoints are reachable only outside PPL, and their
  responses cannot be composed with downstream pipeline operators or inspected
  with `EXPLAIN`.

## Long-Term Goals

**Ideal outcome.** A first-class, leading PPL command that turns a curated set
of read-only management endpoints into fixed-schema rows that compose naturally
with the rest of the language.

**Primary objectives.**
1. Read-only, safe access to operational and topology endpoints from within a
   PPL pipeline.
2. A fixed, plan-time-known schema per endpoint, so downstream `where`, `stats`,
   `sort`, and `fields` work and `EXPLAIN` is meaningful.
3. A default-deny allow-list, so only vetted read-only endpoints are reachable. The command ships **disabled by default** (the operator-facing allow-list setting is empty); a deployment explicitly opts endpoints in, so open source is closed out of the box while a managed service enables only what it supports.
4. Caller-context authorization and secret-field redaction.

**Sustainability and scalability.** Endpoints are expressed as data in a
registry, so adding one is a reviewed registry entry rather than new operators
or grammar. The command rides the existing system-row-source seam, so it
inherits the optimizer and execution machinery already used by `describe` and
the system-index family.

**Does it address the root problem?** Yes. It closes the gap that operational
data is not query-able in PPL, without standing up a new engine or execution
model.

**Confidence.** High. It reuses a shipped, proven seam (the reserved-name
system source behind `describe`/`show`) and a fixed-schema scan; both the
mechanism and the execution path are already established in the codebase.

## Proposal

Add a leading command:

```
| rest <endpoint-path> [count=<int>] [<arg>=<value> ...]
```

It resolves an allow-listed, read-only management endpoint into a fixed-schema
table on the Calcite path, modeled as a system row source. It supports
row-count capping, per-endpoint server-side filter arguments with explicit
value validation, and deterministic plan-time validation that produces
client-side (400-class) errors. Endpoint responses are normalized into flat,
typed rows.

Example:

```
| rest '/_cat/indices' | where health = 'yellow' | sort index | fields index, health, pri
```

## Approach

- **Grammar and AST.** A new `REST` lexer token and `restCommand` parser rule;
  a `RestRelation` AST node. The AST builder validates the endpoint spec and
  encodes it into a single reserved table name.
- **Resolution.** The storage engine decodes the reserved name into a
  `RestSourceTable`, exactly as `describe` resolves to a system index. This
  bridges through `visitRelation` and never reaches the unsupported
  table-function seam.
- **Execution.** `RestSourceTable` -> `CalciteLogicalRestScan` ->
  `CalciteEnumerableRestScan`. A central registry maps each endpoint to its
  read-only transport action, its fixed output schema, the query arguments it
  accepts (with allowed value domains), and a secret-field filter. Dispatch runs
  under the caller's security context through the node client (with a standalone
  REST-client path for the datasource mode).
- **Initial endpoint set.** `/_cluster/health`, `/_cluster/state`,
  `/_cluster/settings`, `/_cat/indices`, `/_cat/nodes`, `/_cat/cluster_manager`,
  `/_cat/plugins`, `/_cat/shards`, `/_resolve/index`,
  `/_plugins/_security/authinfo`.
- **Arguments.** `count` caps emitted rows. Server-side filter arguments are
  applied per endpoint with explicit value validation: `local` on
  `/_cluster/health`, `health` on `/_cat/indices`, `expand_wildcards` on
  `/_resolve/index`. A `timeout` token is reserved in the grammar but rejected
  with a 400 in the initial release, since a single timeout cannot map uniformly
  across the endpoints.
- **Output shaping.** Each endpoint normalizes its response into the fixed
  schema: numeric type normalization, identifier-to-name resolution (for example
  the cluster-manager node id rendered as its node name), role-name expansion,
  structural flattening of nested responses into uniform rows, secret-field
  filtering, and graceful null-valued rows when an optional plugin is absent.

## Alternative

- **Table-function source `source=rest('<endpoint>')`.** Cleaner if the endpoint
  set later becomes open-ended and value-parameterized, and it would unify with
  other parameterized sources under one seam. It requires first building a
  generic Calcite table-function-source capability, because `visitTableFunction`
  is unsupported on the primary path. Deferred; the leading-command form delivers
  the same result today without that prerequisite.
- **Dynamic, response-driven schema (`_MAP` / schema-on-read).** Needed only for
  endpoints whose response shape is keyed by data values. Deferred to a future
  release that depends on the schema-on-read work; the initial endpoint set is
  fully representable with fixed schemas.
- **Raw JSON pass-through.** Returning the endpoint's JSON unchanged is rejected:
  a non-fixed shape cannot publish a plan-time row type, which breaks downstream
  operators and `EXPLAIN`.

## Implementation Discussion

- **Capability gate: table-function source.** `visitTableFunction` throws on the
  Calcite path, so introducing a row source through a function call is not
  available today. The leading-command plus reserved-name approach sidesteps this
  by reusing the system-index scan seam, which is already supported.
- **Capability gate: plan-time schema.** A Calcite scan must publish its row type
  before execution. This is why each endpoint declares a fixed schema in the
  registry, and why dynamic, response-driven endpoints are out of the initial
  scope.
- **Security.** Default-deny allow-list; the command is disabled by default via
  an empty `plugins.ppl.rest.allowed_endpoints` (node-level, non-dynamic) and a
  deployment opts specific endpoints in; only read-only endpoints are
  registered; dispatch runs under the caller's security context; secret-bearing
  fields are filtered during row shaping; and user-supplied argument values are
  validated against per-argument domains rather than passed unchecked into the
  underlying request.
- **No pushdown.** Rows originate from a management transport action, not a
  Lucene index, so `count` caps rows after the call and downstream `where`,
  `sort`, and `stats` run in Calcite. `EXPLAIN` shows the scan with downstream
  operators composed above it.
- **Testing.** Unit tests cover the registry (declared schema, allow-list,
  argument key and value validation, type coercion). Integration tests cover each
  endpoint plus negative cases that assert a 400 for a non-allow-listed endpoint,
  an empty path, a disallowed argument, an out-of-domain argument value, and a
  negative count.
- **Deferred items.** Dynamic, response-driven endpoints (`_MAP`); the
  `include_defaults` argument on `/_cluster/settings` and the `level` argument on
  `/_cluster/health`, which need additional plumbing or a different output shape;
  and a generic table-function source. Each is a follow-on, not a blocker for the
  initial command.

noCharger:
### Description
Add a leading `rest <endpoint>` command that exposes a curated, read-only, fixed-schema set of in-cluster management endpoints as a PPL table, modeled as a system row source bridged through visitRelation (the same seam as describe and the system-index family), so it runs on the Calcite engine without the unsupported table-function path.

- Grammar/AST: REST/TIMEOUT tokens, restCommand rule, RestRelation, AstBuilder visit, query anonymizer.
- Execution: RestSourceTable -> CalciteLogicalRestScan / CalciteEnumerableRestScan; RestEndpointRegistry (read-only allow-list + fixed schema + accepted args); RestEnumerator/RestRequest dispatch via OpenSearchClient (NodeClient in-cluster, RestClient standalone).
- 9 endpoints: cluster health/state/settings, cat indices/nodes/cluster_manager/ plugins/shards, resolve/index.
- Output shaping: numeric type normalization, id-to-name resolution, role-name expansion, structural flattening, graceful null degrade.
- Args: count caps emitted rows; timeout reserved but rejected with 400; get-args applied server-side with per-arg value validation (local on health, health on cat/indices, expand_wildcards on resolve/index). Undeclared arg or out-of-domain value is rejected with a 400. level and include_defaults are deferred to a later release; flat_settings is dropped as redundant.
- Error handling: blank endpoint, negative count, disallowed arg, and uncoercible value all surface clean 400s rather than 500s.

### Response redaction, endpoint allow-list, and access control

Two node-level settings (`NodeScope`, set in `opensearch.yml`) gate the command. The allow-list defaults to **empty, so the command is disabled by default** — a deployment must explicitly opt endpoints in. The redaction setting defaults to native (off). They are intentionally **not** dynamically updatable, so the gate cannot be turned off (or the allow-list widened) at runtime via `_cluster/settings` or `_plugins/_query/settings` once an operator has set it.

- **`plugins.ppl.rest.redaction.enabled`** (Boolean, default `false`): when enabled, masks network identifiers in `/_cat/*` output — IPv4, IPv6, `inet[/...]` addresses, EC2-style host names (`ip-a-b-c-d`), and availability-zone names — and availability-zone names in `/_cluster/settings` values. Default `false` preserves the native `_cat` output (real values).
- **`plugins.ppl.rest.allowed_endpoints`** (List, default `[]`): the command is **disabled by default**; a deployment opts endpoints in explicitly. `[]` (the default) disables the command entirely; an explicit subset allows only those endpoints (others return a clear "not enabled on this cluster" 400); `["*"]` allows every registered endpoint. This lets open source ship closed while a managed service (for example AOS) enables only the endpoints it supports, and AOSS leaves it empty. Enforced at a single choke point in `OpenSearchStorageEngine#getTable`.

**Access control.** The command is already subject to the security plugin's fine-grained access control with no special integration. Each endpoint dispatches a standard transport action (for example `cluster:monitor/nodes/stats`, `cluster:monitor/state`, `indices:admin/resolve/index`) through the node client under the caller's identity, so the security `ActionFilter` authorizes every call by action name. A caller lacking the privilege is rejected, exactly as on the native endpoint, and `/_resolve/index` requires the `indices:admin/resolve/index` privilege because the command resolves all indices. Document- and field-level security do not apply because the command issues no document searches, which matches the native `_cat` and `_cluster` APIs. Consequently, once a deployment opts endpoints in, behavior matches native 1:1 — real IPs and the enabled endpoints are visible only to callers who already hold the corresponding `cluster:monitor/*` privilege — while the empty-by-default allow-list and optional redaction are additional hardening for managed or restricted deployments.

### Related Issues
Resolves https://github.com/opensearch-project/sql/issues/5597

### Check List


language-grammar/src/main/antlr4/OpenSearchPPLLexer.g4
```diff
@@ -13,6 +13,8 @@ options { caseInsensitive = true; }
 SEARCH:                             'SEARCH';
 DESCRIBE:                           'DESCRIBE';
 SHOW:                               'SHOW';
+REST:                               'REST';
+TIMEOUT:                            'TIMEOUT';
 FROM:                               'FROM';
```
ahkcs, line 16: We should leave language-grammar directory untouched since it's the grammar dependency for spark repo

language-grammar/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -113,6 +114,16 @@ describeCommand
...
+
+restCommand
+   : REST stringLiteral (restArgument)*
+   ;
+
+restArgument
+   : COUNT EQUAL integerLiteral
+   | TIMEOUT EQUAL stringLiteral
+   | ident EQUAL literalValue
+   ;
```
ahkcs: ditto

ahkcs:
Did verification on a live composite cluster with analytics path, it seems like `rest` breaks on composite-default clusters.

Repro: composite cluster (cluster.pluggable.dataformat=composite) + analytics stack + this PR, run 
`| rest '/_cluster/health'`

  `TransportPPLQueryAction`  Routing PPL query to analytics engine
  `SemanticCheckException: Table 'REST…__REST_SOURCE' not found`
Same query works on a non-composite cluster. `testRestCommandUnaffectedByAnalyticsEngine` doesn't catch it because its cluster isn't composite-default.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestEndpointRegistry.java
```diff
@@ -0,0 +1,396 @@
...
+        }
+        out.add(new ExprTupleValue(tuple));
+      }
+      return out;
+    }
+  }
+
+  private static final Map<String, Endpoint> REGISTRY = buildRegistry();
+
+  private static Map<String, Endpoint> buildRegistry() {
```
dai-chen, line 118: Could you confirm if all exposed API here are allowed in managed service / serverless. Just concerned if any security issue later. Ref: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-operations.html

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestSourceTable.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * The {@code rest} command's row source: a fixed-schema, server-side dispatch table modeled on
+ * {@link org.opensearch.sql.opensearch.storage.system.OpenSearchSystemIndex}. It resolves the
+ * validated endpoint spec against {@link RestEndpointRegistry} (which enforces the read-only
+ * allow-list), exposes the endpoint's fixed schema, and produces a {@link CalciteLogicalRestScan}
+ * on the Calcite path.
+ */
+@Getter
+public class RestSourceTable extends AbstractOpenSearchTable {
```
dai-chen: Design question: Just wonder is it possible to follow our V2 engine approach to simplify and make this generic, like enable system table in Calcite instead of such specific table type. I recall in V2 we convert SHOW/DESC statement into regular SELECT on system table which is backed by cat index API.

noCharger:
> Did verification on a live composite cluster with analytics path, it seems like `rest` breaks on composite-default clusters.
> 
> Repro: composite cluster (cluster.pluggable.dataformat=composite) + analytics stack + this PR, run `| rest '/_cluster/health'`
> 
> `TransportPPLQueryAction` Routing PPL query to analytics engine `SemanticCheckException: Table 'REST…__REST_SOURCE' not found` Same query works on a non-composite cluster. `testRestCommandUnaffectedByAnalyticsEngine` doesn't catch it because its cluster isn't composite-default.

Good catch, revised on 2db942f

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestSourceTable.java
noCharger: > Design question: Just wonder is it possible to follow our V2 engine approach to simplify and make this generic, like enable system table in Calcite instead of such specific table type. I recall in V2 we convert SHOW/DESC statement into regular SELECT on system table which is backed by cat index API.

Yes, we can introduce an "system table" abstraction to cover both. However, the actual logic are quite diverged so the benefit is minimal. @dai-chen  could you please confirm if https://github.com/opensearch-project/sql/commit/08191587ed45190a49f56c9807f8128b80919e65 is what you meant or there's a better approach for refactor?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestEndpointRegistry.java
songkant-aws: +1, Is there anyway we could verify it by adding some security ITs? Not quite sure whether it's feasible in current plugin.

noCharger: @dai-chen @songkant-aws introduced both allow-listing and column redaction 95f0618ce17654b58ddd7514166bcba7d037d8e3

language-grammar/src/main/antlr4/OpenSearchPPLLexer.g4
noCharger: resolved 06ff820dedb45843da9f30d4b00eb24a951c5540

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestEndpointRegistry.java
dai-chen: I think it's still risky. Managed service customer can toggle the PPL_REST_REDACTION_ENABLED setting via our SQL setting endpoint. Please double check.

noCharger: > I think it's still risky. Managed service customer can toggle the PPL_REST_REDACTION_ENABLED setting via our SQL setting endpoint. Please double check.

Good catch. Both settings were registered Dynamic, so a managed-service customer could flip PPL_REST_REDACTION_ENABLED off at runtime via _plugins/_query/settings or _cluster/settings. https://github.com/opensearch-project/sql/commit/aef46635798e65c9a3c7962bfec4d38f3d844eb0 updated it by dropping Setting.Property.Dynamic so both are now NodeScope-only (set in opensearch.yml). OpenSearch core rejects runtime updates on both paths.

opensearch/src/main/java/org/opensearch/sql/opensearch/client/OpenSearchClient.java
```diff
@@ -114,4 +114,105 @@ public interface OpenSearchClient {
...
+
+  /**
+   * Read-only cat-indices listing for the {@code rest} command (backs {@code /_cat/indices}). One
+   * map per index. Runs under the caller's security thread-context; read-only.
+   *
+   * @param params endpoint query args (already allow-list-validated)
+   * @return one map of column name to value per index
+   */
+  default List<Map<String, Object>> catIndices(Map<String, String> params) {
+    throw new UnsupportedOperationException("catIndices is not supported by this client");
```
dai-chen, line 138: np: remove all these?

docs/user/ppl/cmd/rest.md
```diff
@@ -0,0 +1,82 @@
...
+| `/_cluster/state` | `cluster_name` (string), `state_uuid` (string), `version` (long), `cluster_manager_node` (string) | (none) |
+| `/_cluster/settings` | `setting` (string), `value` (string), `tier` (string) | (none) |
+| `/_cat/indices` | `index` (string), `health` (string), `pri` (integer), `rep` (integer), `active_shards` (integer) | `health` |
+| `/_cat/nodes` | `name` (string), `ip` (string), `node_role` (string), `heap_percent` (integer), `ram_percent` (integer), `cpu` (integer) | (none) |
+| `/_cat/cluster_manager` | `id` (string), `host` (string), `ip` (string), `node` (string) | (none) |
+| `/_cat/plugins` | `name` (string), `component` (string), `version` (string) | (none) |
+| `/_cat/shards` | `index` (string), `shard` (integer), `prirep` (string), `state` (string), `node` (string) | (none) |
+| `/_resolve/index` | `name` (string), `type` (string) | `expand_wildcards` |
```
dai-chen, line 52: I only see `RestResponseRedactor` focus on IP address redaction. But in my test (Dev Tools in AOS domain), I can see many other masked values, e.g., version in _cat/plugins, cluster.routing.allocation.awareness.force.zone in _cluster/settings, host in _cat/cluster_manager. Are all these redacted automatically? If not, any better way to avoid such risk, like sending what's in REST command go through the exact same process as a REST request sent to the domain by users?

noCharger: > I only see `RestResponseRedactor` focus on IP address redaction. But in my test (Dev Tools in AOS domain), I can see many other masked values, e.g., version in _cat/plugins, cluster.routing.allocation.awareness.force.zone in _cluster/settings, host in _cat/cluster_manager. Are all these redacted automatically? If not, any better way to avoid such risk, like sending what's in REST command go through the exact same process as a REST request sent to the domain by users?

The redaction replicates the AOS-side logic exactly. As called out in the PR description, it masks more than IPv4 addresses — IPv6, inet[/…], EC2-style hostnames (ip-a-b-c-d), and availability-zone names are all covered. Every case is asserted in [RestResponseRedactorTest](https://github.com/opensearch-project/sql/pull/5599/changes#diff-674a8678503ada26a7ab6c2e319bc3e66da9898aca6567f66f2a36acc9fe533b).

opensearch/src/main/java/org/opensearch/sql/opensearch/client/OpenSearchClient.java
noCharger: > np: remove all these?

I tried this and had to revert it. It turns out these can't be removed. OpenSearchClient is also implemented downstream by opensearch-project/sql-cli (client/http5/OpenSearchRestClientImpl), which doesn't implement the rest command methods.

Ref https://github.com/opensearch-project/sql/actions/runs/29349113963/job/87140469074?pr=5599

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestResponseRedactor.java
```diff
@@ -0,0 +1,61 @@
...
+  private static final Pattern IPV4 =
+      Pattern.compile("\\b" + OCTET + "\\." + OCTET + "\\." + OCTET + "\\." + OCTET + "\\b");
+  private static final Pattern INET = Pattern.compile("inet\\[/[\\d.:]+\\]");
+  private static final Pattern EC2_HOST =
+      Pattern.compile("\\bip-" + OCTET + "-" + OCTET + "-" + OCTET + "-" + OCTET + "\\b");
+  private static final Pattern IPV6 =
+      Pattern.compile("([0-9a-f]{1,4}:){7}[0-9a-f]{1,4}", Pattern.CASE_INSENSITIVE);
+  private static final Pattern AZ_NAME =
+      Pattern.compile(
+          "(us(-(gov|iso[a-z]?))?|af|ap|ca|cn|eu|sa|me|il)-(central|(north|south)?(east|west)?)-\\d[a-z]",
```
dai-chen: we need to update this whenever there is a new region?

noCharger: Enhanced this logic on [44c9bb0](https://github.com/opensearch-project/sql/pull/5599/commits/44c9bb08557ca75a25b49857d2cbd520e9cdf2f6)

ahkcs (approved):

songkant-aws (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/client/OpenSearchNodeClient.java
```diff
@@ -285,4 +288,274 @@ public void deletePit(DeletePitRequest deletePitRequest) {
...
+    ClusterHealthResponse response = client.admin().cluster().health(request).actionGet();
+    return flattenHealth(response);
+  }
+
+  @Override
+  public List<Map<String, Object>> catIndices(Map<String, String> params) {
+    ClusterHealthResponse response =
+        client.admin().cluster().health(new ClusterHealthRequest()).actionGet();
+    List<Map<String, Object>> rows = new java.util.ArrayList<>();
+    for (Map.Entry<String, ClusterIndexHealth> entry : response.getIndices().entrySet()) {
```
penghuo, line 307: Does all indices visible? e.g. system indices, hidden indices?

opensearch/src/main/java/org/opensearch/sql/opensearch/client/OpenSearchNodeClient.java
```diff
@@ -285,4 +288,274 @@ public void deletePit(DeletePitRequest deletePitRequest) {
...
+            .admin()
+            .cluster()
+            .state(new org.opensearch.action.admin.cluster.state.ClusterStateRequest())
+            .actionGet();
+    List<Map<String, Object>> rows = new java.util.ArrayList<>();
+    org.opensearch.common.settings.Settings persistent =
+        filter.filter(response.getState().metadata().persistentSettings());
+    org.opensearch.common.settings.Settings transientSettings =
+        filter.filter(response.getState().metadata().transientSettings());
+    collectSettings(persistent, "persistent", rows);
```
penghuo, line 476: Please review with security engineer. Some setting should be masked from user.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestEndpointRegistry.java
```diff
@@ -0,0 +1,428 @@
...
+      throw new IllegalArgumentException(
+          "rest endpoint must be a non-empty path. Supported read-only endpoints: "
+              + REGISTRY.keySet());
+    }
+    Endpoint endpoint = REGISTRY.get(path);
+    if (endpoint == null) {
+      throw new IllegalArgumentException(
+          "rest endpoint ["
+              + path
+              + "] is not allow-listed. Only read-only in-cluster endpoints are supported: "
```
penghuo, line 271: Only read-only in-cluster endpoints are supported -> Only follwing endpoints are supported: [_cat/nodes, _cluster/health]

PR #5587, opened 2026-06-25
Return 4xx instead of 500 for unsupported window functions

OVI3D0:
### Description

Window functions outside `WINDOW_FUNC_MAPPING` (e.g. `RANK`) were escaping the analytics-engine route as HTTP 500. The throw site in `CalciteRexNodeVisitor#visitWindowFunction` emitted a raw `UnsupportedOperationException`, which `UnifiedQueryPlanner.plan` rethrows unchanged, so it never got reclassified as a client error.

Switching the throw to `CalciteUnsupportedException` lets the 4xx wrapper from #5569 take over and normalize it to `SemanticCheckException`.

Repro:
```sql
SELECT RegionID, COUNT(*) AS cnt,
       RANK() OVER (ORDER BY COUNT(*) DESC) AS rnk
FROM clickbench GROUP BY RegionID LIMIT 5
```

### Test plan
- [x] New `UnifiedQueryPlannerTest#unsupportedWindowFunctionIsRethrownAsSemanticCheckException`: fails on `main`, passes with the fix
- [x] Existing `CalcitePPLEventstatsIT#testUnsupportedWindowFunctions` still green
- [x] Manual repro on a local cluster:
```
mikeovi@80a997164b98 ~ % curl -sS -w "\nHTTP %{http_code}\n" -X POST "http://localhost:9200/_plugins/_ppl" \
    -H 'Content-Type: application/json' \
    -d '{"query":"source=demo | eventstats rank() | head 5"}'
{
  "error": {
    "context": {
      "stage": "analyzing",
      "stage_description": "Parsing and validating the query"
    },
    "reason": "Unexpected window function: rank",
    "details": "Unexpected window function: rank",
    "location": ["while preparing and validating the query plan"],
    "code": "UNKNOWN",
    "type": "CalciteUnsupportedException"
  },
  "status": 400
}
HTTP 400
mikeovi@80a997164b98 ~ %
```

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
- [x] Commits are signed per the DCO using --signoff.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5600, opened 2026-06-30
Improve error message for unsupported window functions in eventstats/streamstats

### LINKED ISSUE #5168: [BUG] eventstats/streamstats reject window functions that grammar accepts
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-types | eventstats row_number() by str_field
source=bounty-types | streamstats rank() by int_field
source=bounty-types | eventstats dense_rank() by str_field
```
**Expected Result:**
Supported window functions should execute in `eventstats/streamstats`.

**Actual Result:**
Each query fails with HTTP 500 and `Unexpected window function` for `row_number`, `rank`, and `dense_rank`.

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "str_field": { "type": "keyword" }
    }
  }
}
```
**Sample Data**
```json
[
  { "int_field": 42, "str_field": "alpha" },
  { "int_field": -1, "str_field": "alpha" },
  { "int_field": 0, "str_field": "beta" }
]
```

## Bug Description
**Issue Summary:**
`eventstats/streamstats` rejects common window functions even though grammar accepts them.

**Steps to Reproduce:**
1. Create index with `int_field` and `str_field`.
2. Insert sample rows.
3. Run each query above.
4. Observe HTTP 500 `Unexpected window function` errors.

**Impact:**
Window-based analytics pipelines cannot run using these standard functions.

gingeekrishna:
## Description

Fixes #5168

`eventstats`/`streamstats` reject window functions outside `WINDOW_FUNC_MAPPING` (e.g. `rank()`, `dense_rank()`, `nth_value()`) with a bare `"Unexpected window function: X"` error.

As confirmed in the [issue discussion](https://github.com/opensearch-project/sql/issues/5168#issuecomment-2967588622) by @songkant-aws, this is expected behavior, not a bug — `rank`/`dense_rank`/`nth_value` require `ORDER BY` semantics, but `eventstats`/`streamstats` only support `partition by`. The issue was relabeled `error-experience`: the fix is to make the error message clearer, not to add support for these functions.

### Changes

- `CalciteRexNodeVisitor#visitWindowFunction`: replaced the generic `"Unexpected window function: X"` message with one that names the rejected function and lists the functions `eventstats`/`streamstats` do support (sourced from `WINDOW_FUNC_MAPPING`), so users get actionable guidance instead of a bare error.
- Updated the existing unit test (`UnifiedQueryPlannerTest`) and integration tests (`CalcitePPLEventstatsIT`, `CalciteStreamstatsCommandIT`) to assert against the new message.
- Added new integration tests covering `rank`/`dense_rank` specifically, since the issue's repro queries used those functions.

Note: this PR builds on top of #5587 (already merged), which fixed the same throw site to return a 4xx instead of a 500. This PR only changes the message text, not the exception type or HTTP status.

## Test plan

- [x] Unit test `UnifiedQueryPlannerTest#unsupportedWindowFunctionIsRethrownAsSemanticCheckException` passes
- [x] Integration tests added/updated for `eventstats` and `streamstats` (require a live cluster to run in CI)

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -705,7 +705,14 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
                       node.getWindowFrame());
             })
         .orElseThrow(
-            () -> new CalciteUnsupportedException("Unexpected window function: " + funcName));
+            () ->
+                new CalciteUnsupportedException(
+                    "Window function '"
+                        + funcName
+                        + "' is not supported in eventstats/streamstats."
+                        + " Supported functions: avg, count, dc, distinct_count, earliest,"
+                        + " latest, max, min, row_number, stddev_pop, stddev_samp, sum,"
+                        + " var_pop, var_samp."));
   }
 
   private List<RexNode> translateOrderKeys(
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -808,7 +808,14 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
                       node.getWindowFrame());
             })
         .orElseThrow(
-            () -> new CalciteUnsupportedException("Unexpected window function: " + funcName));
+            () ->
+                new CalciteUnsupportedException(
+                    "Window function '"
+                        + funcName
+                        + "' is not supported in eventstats/streamstats."
+                        + " Supported functions: avg, count, dc, distinct_count, earliest,"
+                        + " latest, max, min, row_number, stddev_pop, stddev_samp, sum,"
+                        + " var_pop, var_samp."));
   }
 
   private List<RexNode> translateOrderKeys(
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -95,6 +95,7 @@ import org.opensearch.sql.data.type.ExprType;
 import org.opensearch.sql.exception.CalciteUnsupportedException;
 import org.opensearch.sql.exception.ExpressionEvaluationException;
 import org.opensearch.sql.exception.SemanticCheckException;
+import org.opensearch.sql.executor.QueryType;
 import org.opensearch.sql.expression.function.BuiltinFunctionName;
 import org.opensearch.sql.expression.function.CoercionUtils;
 import org.opensearch.sql.expression.function.PPLBuiltinOperators;
@@ -763,6 +764,17 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
             .toList();
     List<RexNode> orderKeys = translateOrderKeys(node.getSortList(), context);
     return BuiltinFunctionName.ofWindowFunction(funcName)
+        // rank/dense_rank were added to WINDOW_FUNC_MAPPING to support SQL's RANK()/DENSE_RANK()
+        // OVER (...), but PPL only ever reaches this visitor through eventstats/streamstats
+        // (there's no other PPL syntax that builds a WindowFunction node), which don't support
+        // them - PPL has no ORDER BY syntax for eventstats/streamstats, so ranking would be
+        // meaningless there. Excluding them here for PPL keeps that restriction despite the
+        // shared mapping, without needing a PPL-specific grammar change.
+        .filter(
+            functionName ->
+                !(context.queryType == QueryType.PPL
+                    && (functionName == BuiltinFunctionName.RANK
+                        || functionName == BuiltinFunctionName.DENSE_RANK)))
         .map(
             functionName -> {
               RexNode field = arguments.isEmpty() ? null : arguments.getFirst();
--- a/api/src/test/java/org/opensearch/sql/api/UnifiedQueryPlannerTest.java
+++ b/api/src/test/java/org/opensearch/sql/api/UnifiedQueryPlannerTest.java
@@ -168,7 +168,8 @@ public class UnifiedQueryPlannerTest extends UnifiedQueryTestBase {
     givenInvalidQuery("source = catalog.employees | eventstats rank()")
         .assertErrorType(SemanticCheckException.class)
         .assertCauseType(CalciteUnsupportedException.class)
-        .assertErrorMessageContains("Window function 'rank' is not supported in eventstats/streamstats");
+        .assertErrorMessageContains(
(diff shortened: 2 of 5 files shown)


dai-chen (commented):
Hi, I'm trying to add support for `rank` and `dense_rank` function in Calcite for unified SQL for analytics engine. If we confirm both are not supported by PPL evenstats and streamstats, shall we remove them from PPL grammar directly? I assume that's the only place window function can be used. Otherwise it will be tricky to check language type in Calcite planner.

gingeekrishna:
@dai-chen Good question — I dug into this a bit.

Your assumption holds for PPL specifically: in the PPL grammar, `windowFunction` (the parenthesized call syntax) is only ever invoked from `eventstatsAggTerm`/`streamstatsAggTerm` ([OpenSearchPPLParser.g4:922-943](https://github.com/opensearch-project/sql/blob/main/ppl/src/main/antlr/OpenSearchPPLParser.g4#L922-L943)), so that is the only place a window function can appear in PPL today.

But it's not the only place in the engine overall. SQL has its own `windowFunctionClause`/`overClause` grammar ([OpenSearchSQLParser.g4:150-159](https://github.com/opensearch-project/sql/blob/main/language-grammar/src/main/antlr4/OpenSearchSQLParser.g4#L150-L159)) that already accepts `RANK()`/`DENSE_RANK()` in `... OVER (...)`. Both SQL and PPL build the same `WindowFunction` AST node and are resolved by the same `CalciteRexNodeVisitor#visitWindowFunction` — there's even a comment there noting "SQL emits AggregateFunction for aggregate-as-window (e.g., SUM(x) OVER); PPL emits Function." `WINDOW_FUNC_MAPPING` in `BuiltinFunctionName` (used by both paths) simply has no `rank`/`dense_rank` entries yet, so both languages fail identically right now.

Practical implication: if you add `rank`/`dense_rank` to `WINDOW_FUNC_MAPPING` and implement them in the shared visitor, that enables them for PPL eventstats/streamstats too, since the mapping/visitor is shared — not just SQL.

So yes, I think removing `RANK`/`DENSE_RANK` from PPL's `scalarWindowFunctionName` rule (lines 933-934) is the right move, and it avoids the language-type check you were worried about: PPL's parser would reject the syntax before it ever reaches the AST/Calcite layer, while SQL's grammar (which already accepts it) passes through untouched. That pushes the language separation to parse time instead of planner time.

One side note on this PR itself: the new error message I added ("Window function 'x' is not supported in eventstats/streamstats") is thrown from that same shared visitor, so it would also currently fire — misleadingly — for a plain SQL `RANK() OVER (...)` query, since the check isn't actually PPL-specific. I may follow up with a tweak to make that message language-agnostic.

dai-chen:
@gingeekrishna Thanks for digging in! I've put up PR https://github.com/opensearch-project/sql/pull/5720 for the SQL side. As you confirmed, removing them from grammar should make PPL reject at parse time and leaves my change SQL-only. With both PRs merged, I think we get the language separation at parse time rather than in the planner, which is where it belongs as expected. Thanks!

gingeekrishna:
@dai-chen Nice, thanks for #5720. Pushed `da13116` here: removed `RANK`/`DENSE_RANK` from PPL's `scalarWindowFunctionName` rule, so `eventstats rank()`/`streamstats dense_rank()` now fail at parse time (`SyntaxCheckException`) instead of reaching the shared visitor.

Updated the tests this PR added for that case (`testRankingWindowFunctionsUnsupportedInEventstats`/`...InStreamstats`) to expect the parse-time rejection, and switched `unsupportedWindowFunctionIsRethrownAsSemanticCheckException` from `rank()` to `percent_rank()` since it needs a token that's still in the grammar but outside `WINDOW_FUNC_MAPPING` to exercise that path.

One caveat: I don't have a Java toolchain in my current environment, so I couldn't run `generateGrammarSource`/the actual test suite to confirm the exact ANTLR error message — the assertions only check for the generic `"is not a valid term at this part of the query"` substring rather than the full expected-tokens list, to stay robust either way, but it'd be good to have CI confirm this once it runs.

dai-chen:
@gingeekrishna My PR was already merged. You can rebase and verify yours by CI if you're unable locally. Btw, I recall changes in `unsupportedWindowFunctionIsRethrownAsSemanticCheckException` is already covered in my PR.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -808,7 +808,14 @@ public RexNode visitWindowFunction(WindowFunction node, CalcitePlanContext conte
...
         .orElseThrow(
-            () -> new CalciteUnsupportedException("Unexpected window function: " + funcName));
+            () ->
+                new CalciteUnsupportedException(
+                    "Window function '"
+                        + funcName
+                        + "' is not supported in eventstats/streamstats."
+                        + " Supported functions: avg, count, dc, distinct_count, earliest,"
+                        + " latest, max, min, row_number, stddev_pop, stddev_samp, sum,"
+                        + " var_pop, var_samp."));
```
dai-chen, line 830: maybe revert this change because this list is dynamic and defined in grammar.

dai-chen (approved):
Minor comment. Thanks for the changes!

dai-chen:
I retried Linter CI but it still failed. Probably worth taking another look.

gingeekrishna:
Found it - pushed `58fda664f` to fix it.

The failing check was the linter, specifically:
```
[ppl-lint-grammar] FAILED unsupported-window-function-in-eventstats/eventstats-rank: expected 1, got 0
```

That's OpenSearch-Dashboards' own PPL linter (fetched via `yarn osd bootstrap` in that job) being run against this repo's grammar bundle, using the cases in `scripts/ppl-lint/grammar-cases.json`. Its `unsupported-window-function-in-eventstats` rule expects `eventstats rank()` to **parse successfully** and then get flagged by a semantic-layer diagnostic - that's how it surfaces a friendly, specific error in the IDE.

My earlier commit here (`cf4cb1bd3`, removing `RANK`/`DENSE_RANK` from the PPL grammar so `eventstats rank()` fails at parse time instead of reaching the shared visitor) broke exactly that: the query now dies with a raw syntax error before OSD's semantic rule ever gets a chance to run, so its diagnostic never fires. Since that rule lives in the OpenSearch-Dashboards repo, not here, fixing it properly would need a coordinated cross-repo change - out of scope for this PR.

Reverted that commit. This PR's actual point - the improved `"Window function 'x' is not supported in eventstats/streamstats"` message in `CalciteRexNodeVisitor` - is a semantic-layer check, so it's untouched by the revert and still works exactly as intended. The parse-time language-separation idea from our earlier discussion doesn't hold up against the existing linter contract, so I'm dropping it rather than pursuing it further here.

dai-chen:
Hi @gingeekrishna Thanks for the deep dive! I thought this means we should remove `rank`/`dense_rank` from both PPL grammar and linter config? cc @RyanL1997 do you have some context for our PPL linter?

RyanL1997 (approved):

Issue #5164, opened 2026-02-23
[BUG] Silent integer and long overflow wrapping in Calcite arithmetic path


penghuo:
## Query Information
**PPL Command/Query:**
```ppl
source=bounty-types | where int_field = 2147483647 | eval overflow = int_field + 1 | fields int_field, overflow
source=bounty-types | where long_field = 9223372036854775807 | eval overflow = long_field + 1 | fields long_field, overflow
```
**Expected Result:**
Overflow should not silently wrap. Engine should either return validated error or safe overflow handling.

**Actual Result:**
- `2147483647 + 1` returns `-2147483648`
- `9223372036854775807 + 1` returns `-9223372036854775808`
Both return HTTP 200 with wrapped values.

## Dataset Information
**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [ ] Simple Schema for Observability (SS4O)
- [ ] Open Cybersecurity Schema Framework (OCSF)
- [x] Custom (details below)

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "int_field": { "type": "integer" },
      "long_field": { "type": "long" }
    }
  }
}
```
**Sample Data**
```json
{
  "int_field": 2147483647,
  "long_field": 9223372036854775807
}
```

## Bug Description
**Issue Summary:**
Integer and long arithmetic in `eval` wraps on overflow without warning.

**Steps to Reproduce:**
1. Create index with `int_field` and `long_field`.
2. Insert a max-value record.
3. Run both queries above.
4. Observe wrapped negative results with HTTP 200.

**Impact:**
Silent numeric corruption in analytical results.

penghuo:
### Investigation Summary

**Root cause:** This is a Calcite engine behavior, not specific to the PPL/SQL plugin. Standard Calcite arithmetic operators (`SqlStdOperatorTable.PLUS/MINUS/MULTIPLY`) generate plain Java `+`/`-`/`*` in the Enumerable code path, which silently wrap on integer/long overflow. Confirmed by reproducing the same wrapping behavior in Calcite's own CsvTest — `SELECT iv, iv+1 FROM STATUS` with `iv = 2147483647` returns `-2147483648`.

The V2 (legacy) engine does not have this issue — it already uses `Math.addExact`/`subtractExact`/`multiplyExact`.

### Prototype Fix

[...]

PR #5603, opened 2026-07-01
Widen narrow integer operands in PPL +/-/* to prevent overflow

ahkcs:
### Description

Integer arithmetic (`+`, `-`, `*`) on narrow integer types silently overflowed because the operators kept Calcite's `LEAST_RESTRICTIVE` result type (`SMALLINT * SMALLINT -> SMALLINT`, `INTEGER * INTEGER -> INTEGER`). For example `eval area = ResolutionWidth * ResolutionHeight` (both `short`) computed `28,777,836` but wrapped into the 16-bit range, so `where area > 2000000` returned 0 rows. The reported workaround was `cast(... as long)`.

This widens narrow integer operands before the operation so the result cannot overflow the (mis-inferred) narrow type:

| operands | result |
|---|---|
| `byte`/`short` | `int` |
| any `int`/`long` involved | `long` (`bigint`) |
| any `float`/`double`/`decimal`/datetime | unchanged (defers to Calcite) |

The widening is a CAST injected on the operands in the SQL-plugin lowering (`PPLFuncImpTable`), applied to `+`, `-`, `*` and the named `add`/`subtract`/`multiply` functions. The string-concat `+` and the `DATETIME - DATETIME` variant are unchanged. Arithmetic inside higher-order-function lambda bodies (`reduce`/`mvmap`) is left unwidened — those operands are lambda-parameter placeholders resolved by the HOF's own type inference, and they execute JVM-side where operands already promote to int.

#### Why widening (and not error-on-overflow)

An alternative is to detect overflow and error (Calcite's `CHECKED_PLUS`/`CHECKED_MINUS`/`CHECKED_MULTIPLY` → `Math.addExact`, explored in #5604). Widening was chosen because it is the only approach that fixes **both** execution engines from a single change:

- **The CAST is baked into the RelNode before the engine fork.** PPL runs on two paths — the Calcite/Enumerable engine and the analytics-engine (DataFusion) backend. Because widening happens in `PPLFuncImpTable` (shared lowering, upstream of the fork), the widened operand types flow through to *both* — including through Substrait to DataFusion. Verified live: `short * short` (30000×30000) returns `int` `900000000` on the Calcite path **and** on the DataFusion route (profile confirms `*(CAST($n):INTEGER, ...)` with `backends=[datafusion]`); the reported `| where area > 2000000` now returns the matching row on both.
- **Error-on-overflow cannot be done consistently across engines.** The checked operators are a Calcite-codegen mechanism (`Math.addExact` in generated Java). `CHECKED_*` has **no Substrait representation**, so on the analytics-engine route the arithmetic executes natively in DataFusion and the checked rewrite never reaches it — leaving the overflow unfixed there. Making DataFusion error would require separate backend work (a custom Rust checked-arithmetic UDF). So erroring would need two independent implementations (Calcite checked-arith + a DataFusion UDF) and still diverge; widening needs one and is consistent.

Widening fully resolves the reported `short`/`int` overflow on both engines. It does not address `long * long` overflow (no wider primitive type) — that residual is out of scope here and, if needed, is a separate error-on-overflow effort.

#### Verification

| query | before | after (both engines) |
|---|---|---|
| `eval area = short * short` (30000²) | wraps (`smallint`, negative) | `int` `900000000` |
| `eval area = short * short \| where area > 2000000` | 0 rows | 1 row |
| `eval x = int * int` (2e9²) | wraps | `bigint`, correct |

| test | result |
|---|---|
| `CalcitePPLBuiltinFunctionIT` widening ITs (new) | pass |
| `CalcitePPLEvalTest`/`ArrayFunctionTest`/`DedupTest`/etc. + `UnifiedQueryPlannerSqlV2Test` (plan snapshots) | pass (updated for widening CAST) |
| `CalciteExplainIT` / `CalciteNoPushdownIT` (explain fixtures) | pass (regenerated) |
| `collection.md` doctest (reduce/mvmap unaffected) | pass |

### Related Issues

Resolves the "eval integer arithmetic overflow (no type widening)" bug. See #5164 / #5604 for the alternative error-on-overflow approach (Calcite-path only).

### Check List


core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -741,23 +804,25 @@ void populate() {
       registerOperator(OR, SqlStdOperatorTable.OR);
       registerOperator(NOT, SqlStdOperatorTable.NOT);
 
-      // Register ADDFUNCTION for numeric addition only
-      registerOperator(ADDFUNCTION, SqlStdOperatorTable.PLUS);
-      registerOperator(
+      // Register ADDFUNCTION for numeric addition only. Widen narrow integer operands so the sum
+      // cannot overflow the (mis-)inferred SMALLINT/TINYINT result type; see
+      // registerWideningIntegerOperator.
+      registerWideningIntegerOperator(ADDFUNCTION, SqlStdOperatorTable.PLUS);
```
RyanL1997, line 893: Does this means the behavior-change note in the PR description applies to SQL too, not just PPL eval?

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
```diff
@@ -104,7 +104,7 @@ public void testEvalSum() {
     String ppl = "source=EMP | eval total = sum(1, 2, 3) | fields EMPNO, total";
     RelNode root = getRelNode(ppl);
     String expectedLogical =
-        "LogicalProject(EMPNO=[$0], total=[+(1, +(2, 3))])\n"
+        "LogicalProject(EMPNO=[$0], total=[+(1, +(2:BIGINT, 3:BIGINT))])\n"
```
RyanL1997, line 107: This snapshot is the one I couldn't reconcile: the inner `+(2:BIGINT, 3:BIGINT)` widens both literals, but the outer operand stays plain `1` (INTEGER), whereas `selectExpressionWithoutFrom`'s `1 + 1` widens both to `1:BIGINT`. If the outer `+` in the `sum` reduction goes through the same widening operator, I'd expect `+(1:BIGINT, +(2:BIGINT, 3:BIGINT))`. Can you confirm whether the outer add is intentionally not widened (and that a single `+` mixing INTEGER + BIGINT operands is expected)?

RyanL1997 (commented):
Hi @ahkcs , thanks for the change, and here are somer reviews.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: No — PPL only. `QueryService.shouldUseCalcite()` gates the Calcite path on `queryType == QueryType.PPL`, so SQL is untouched.

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLEvalTest.java
ahkcs: It *is* widened — the bare `1` is a rendering artifact. Dumping the tree shows `outer.op0 = 1:BIGINT (type=BIGINT)`; every operand is BIGINT, no mixed add. Calcite's `RexCall` digest drops a literal's `:BIGINT` suffix when its sibling has the same type (here `+(2,3)` is BIGINT), so `1:BIGINT` prints as `1`.

RyanL1997 (approved):

PR #5604, opened 2026-07-01
Detect long (BIGINT) arithmetic overflow instead of silently wrapping (#5164)

### LINKED ISSUE #5164: [BUG] Silent integer and long overflow wrapping in Calcite arithmetic path

ahkcs:
### Description

`long` (BIGINT) arithmetic (`+`, `-`, `*`) in PPL/SQL `eval`/`SELECT` expressions silently **wrapped** on overflow in the Calcite engine — e.g. `long_field * 9999999999` at the top of the 64-bit range returned a negative value with HTTP 200. `SqlStdOperatorTable.PLUS/MINUS/MULTIPLY` generate plain Java `+`/`-`/`*` in the Enumerable code path, which wrap.

This rewrites `long` `+`/`-`/`*` to their overflow-checked variants (`CHECKED_PLUS` / `CHECKED_MINUS` / `CHECKED_MULTIPLY`), which generate `Math.addExact` / `subtractExact` / `multiplyExact` and throw `ArithmeticException` on overflow. The exception is caught in `QueryService` and surfaced as a **4xx** client error (`NonFallbackCalciteException`) instead of wrapping or falling back to the V2 engine. (The V2 engine already uses `Math.*Exact`.)

#### Scope: BIGINT-only

The checked rewrite is gated to `long` (BIGINT) operands. Narrower integer arithmetic (`byte`/`short`/`int`) is **widened** to a type that cannot overflow *before* this rewrite runs — `PPLFuncImpTable` promotes `byte`/`short` → `int` and any `int`/`long` operand → `long` for `+`/`-`/`*` (#5603) — so `long`, which has no wider integer type to widen into, is the only remaining overflow case on the Calcite engine. `float`/`double`/`decimal` follow IEEE 754 / decimal semantics and have no `CHECKED_*` runtime, so they are left untouched.

The rewrite runs **after** the analytics-engine fork in `convertToCalcitePlan`, so only the Calcite/Enumerable path is affected. The DataFusion analytics backend handles its own `long` overflow (checked arrow kernels, opensearch-project/OpenSearch#22378) — the two engines now converge on erroring for `long` overflow.

It runs **before** pushdown so both coordinator-executed and pushed-down (script) arithmetic are checked. `PPLAggregateConvertRule`, `OpenSearchRelOptUtil`, and `ExtendedRelJson` recognize the `CHECKED_*` kinds so aggregate-pushdown, sort-expression pushdown, and script serialization keep working for pushed-down checked arithmetic.

#### Stacked on #5603

This PR is stacked on #5603 (operand widening) and depends on it: without the widening underneath, `short`/`int` overflow would slip through un-checked on the Calcite path. The first commit here is #5603's widening; it drops out of this diff once #5603 merges. Together the two PRs close integer-arithmetic overflow on the Calcite engine (#5603 widens the narrow tiers; this errors the `long` tier), matching the analytics-engine behavior.

### Testing

| test | behavior |
|---|---|
| `issues/5164.yml` (REST) | `int`/`short` overflow → widened correct value (200); `long` overflow (`+`/`-`/`*`) → **4xx**; normal arithmetic unaffected |
| `CalciteExplainIT` (pushdown) | 258/258 pass — explain fixtures regenerated to show `CHECKED_*` in pushed scripts |
| `CalciteNoPushdownIT` (CalciteExplainIT) | 258/258 pass — pushdown-only arithmetic tests skipped as expected |

### Related Issues
Resolves #5164. Stacked on #5603. Analytics-engine counterpart: opensearch-project/OpenSearch#22378.

### Check List


core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -383,6 +400,91 @@ private boolean isCalciteEnabled(Settings settings) {
...
+
+  /**
+   * Checked arithmetic is applied to BIGINT ({@code long}) operands only. Narrower integer
+   * arithmetic (byte/short/int) is already widened to a type that cannot overflow before this
+   * rewrite runs — {@code PPLFuncImpTable} promotes byte/short to INT and any int/long operand to
+   * BIGINT for {@code +}/{@code -}/{@code *} — so the sole remaining overflow case that reaches the
+   * Calcite engine is {@code long} arithmetic, which has no wider integer type to widen into.
+   * Float/double/decimal follow IEEE 754 (or decimal semantics) and have no {@code CHECKED_*}
+   * runtime (e.g. {@code SqlFunctions.checkedMultiply(double, double)} does not exist), so they are
+   * left untouched. Require both the result and every operand to be BIGINT.
```
penghuo: simply comments. it duplicate with Line 431-434.

core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -187,7 +195,15 @@ public void executeWithCalcite(
                 },
                 QueryService.class);
           } catch (Throwable t) {
-            if (isCalciteFallbackAllowed(t) && !(t instanceof NonFallbackCalciteException)) {
+            ArithmeticException overflow = findArithmeticOverflow(t);
```
penghuo: Is there better way to handle expected calcite exception? can it wrapped with NonFallbackcalciteException?

core/src/main/java/org/opensearch/sql/executor/QueryService.java
ahkcs: simplified

core/src/main/java/org/opensearch/sql/executor/QueryService.java
ahkcs: Good suggestion. I moved overflow handling to the Calcite execution boundary, where the wrapped `ArithmeticException` is translated into a top-level `NonFallbackCalciteException`. The outer catch now uses the existing generic fallback logic. Addressed in 729438cd8.

penghuo (approved):

PR #5607, opened 2026-07-05
Fix SQL explain API rejecting ?format=json parameter

### LINKED ISSUE #4373: [BUG] Deprecation of OpenSearch DSL format has affected the behavior of the explain API
**What is the bug?**
Deprecation of OpenSearch DSL format has affected the behavior of the explain API
https://github.com/opensearch-project/sql/issues/3280

Prior to OpenSearch 3.0 we could request DSL syntax using the following query:
```json
POST /_plugins/_sql/_explain?format=json
{"query":"SELECT Shape, _id FROM railroads LIMIT 0,10000"}
```
result:
```json
{
  "from": 0,
  "size": 10000,
  "_source": {
    "includes": [
      "Shape",
      "_id"
    ],
    "excludes": []
  }
}
```
With 3.0 we have noticed a couple of issues with our workflow.
1.  using `format=json` returns an error when used with explain API
   ```json
   {
     "error": {
       "details": "Failed to create executor due to unknown response format: json",
     },
     "status": 400
   }
   ```
2.  if we dont use `format=json` then we run into issues with including a geoshape column
   ```json
   POST /_plugins/_sql/_explain
   {"query":"SELECT Shape, _id FROM railroads LIMIT 0,10000"}
   ```
   ```json
   {
     "error": {
       "details": "can't resolve Symbol(namespace=FIELD_NAME, name=Shape) in type env",
       "type": "SemanticCheckException"
     },
     "status": 400
   }
   ```

gingeekrishna:
### Description

Fixes #4373

**Problem:** Since OpenSearch 3.0, calling `POST /_plugins/_sql/_explain?format=json` returns:
```json
{
  "error": {
    "details": "Failed to create executor due to unknown response format: json",
  },
  "status": 400
}
```

**Root cause (corrected):** My first attempt at this fix (still included below) only added `"json"` to `isSupportedExplainFormat()` in `SQLQueryRequest.java`, the V2/`sql`-module engine's own validation. That turned out to be insufficient - that method is never reached for this bug. `RestSqlAction#prepareRequest` calls `SqlRequestParam.getFormat(request.params())` **unconditionally, for every request**, before `SQLQueryRequest` is even constructed. That call throws `IllegalArgumentException` for any format outside the legacy `Format` enum (`jdbc`/`csv`/`raw`/`table`) - `"json"` included - which is exactly the error above. So the request was failing before either engine (legacy or V2) ever got a chance to look at it.

**Fix:**
1. `RestSqlAction#resolveFormat` (new): when `SqlRequestParam.getFormat` rejects the value AND the request is an explain request AND the rejected value is `"json"`, fall back to a default `Format` instead of throwing. The explain response body is JSON regardless of this parameter (`executeSqlRequest`'s explain branch calls `queryAction.explain().explain()` directly, never through a `Format`-specific executor), so ignoring it here changes nothing else about the response.
2. `SQLQueryRequest#isSupportedExplainFormat()` (kept from the original attempt): once a request gets past (1), this makes the V2 engine treat `format=json` as a supported explain format, so it handles the explain natively instead of always falling back to the legacy engine.

### Changes

- `legacy/src/main/java/.../plugin/RestSqlAction.java` — add `resolveFormat()`, the actual fix for the reported crash
- `sql/src/main/java/.../SQLQueryRequest.java` — add `"json"` to `isSupportedExplainFormat()` (keeps the V2 engine as the primary path once the format-parsing crash is fixed)
- `sql/src/test/java/.../SQLQueryRequestTest.java` — unit test for `isSupportedExplainFormat()`
- `integ-test/.../legacy/ExplainIT.java` — integration test hitting the real `/_explain?format=json` endpoint end-to-end (per review - a unit test of `isSupportedExplainFormat()` alone doesn't exercise `RestSqlAction`, so it wouldn't have caught that the real bug was one layer up)

### Testing

- `SQLQueryRequestTest`: unit coverage for the V2 engine's format validation.
- `ExplainIT#testExplainAcceptsJsonFormatForBackwardCompatibility`: hits `POST /_plugins/_sql/_explain?format=json` for real and asserts a 200 with no `error` key - this is the test that actually exercises the code path the bug lived in.

Note: I don't have a Java/Gradle toolchain available in my current environment, so I couldn't run these locally - would appreciate CI confirming.

### DIFF
--- a/sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
+++ b/sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
@@ -159,7 +159,11 @@ public class SQLQueryRequest {
   }
 
   private boolean isSupportedExplainFormat() {
-    return Stream.of("simple", "standard", "extended", "cost").anyMatch(format::equalsIgnoreCase);
+    // "json" is accepted for backward compatibility: the explain endpoint always returns JSON
+    // regardless of this parameter, so treating it as valid avoids the 400 regression
+    // introduced in OpenSearch 3.0. See https://github.com/opensearch-project/sql/issues/4373
+    return Stream.of("simple", "standard", "extended", "cost", "json")
+        .anyMatch(format::equalsIgnoreCase);
   }
 
   private String getFormat(Map<String, String> params) {
--- a/integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/legacy/ExplainIT.java
@@ -13,11 +13,13 @@ import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_LOCATION;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_NESTED_TYPE;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_PEOPLE;
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_PHRASE;
+import static org.opensearch.sql.legacy.plugin.RestSqlAction.EXPLAIN_API_ENDPOINT;
 
 import com.google.common.io.Files;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import org.json.JSONObject;
 import org.junit.Assert;
 import org.junit.Ignore;
 import org.junit.Test;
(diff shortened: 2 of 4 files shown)


gingeekrishna:
The `enforce-label` check is failing — a release label is required. Since this PR improves the error message for unsupported window functions (a UX/error-experience fix), the `bugFix` label would be appropriate. Could a maintainer add it? Thanks!

sql/src/main/java/org/opensearch/sql/sql/domain/SQLQueryRequest.java
```diff
@@ -159,7 +159,11 @@ private boolean isSupportedFormat() {
   }
 
   private boolean isSupportedExplainFormat() {
-    return Stream.of("simple", "standard", "extended", "cost").anyMatch(format::equalsIgnoreCase);
+    // "json" is accepted for backward compatibility: the explain endpoint always returns JSON
```
dai-chen, line 162: Could you add yaml IT to verify and this may be moved there as context? Ref: https://github.com/opensearch-project/sql/blob/main/integ-test/src/yamlRestTest/resources/rest-api-spec/test/issues/5702.yml

gingeekrishna: Thanks for asking for this - it turned up something important. While putting together an integration test I traced the actual request flow and found the original fix (adding "json" to `isSupportedExplainFormat()`) doesn't actually fix the bug: that method is in `SQLQueryRequest` (the V2 engine), but `RestSqlAction#prepareRequest` calls `SqlRequestParam.getFormat(request.params())` unconditionally, for every request, *before* `SQLQueryRequest` is ever constructed. That call throws for any format outside `jdbc`/`csv`/`raw`/`table` - "json" included - which is exactly the "Failed to create executor due to unknown response format: json" error from #4373. So the original fix was never reached; the request was already failing a layer up.

Pushed `2a4d6a4c0`, which adds the real fix in `RestSqlAction` (falls back instead of throwing when the rejected value is "json" on an explain request), and kept the original `SQLQueryRequest` change since it still matters once that's fixed (makes the V2 engine handle it natively instead of always falling back to legacy).

For the test: there's no YAML REST-spec coverage for any SQL endpoint yet (only `ppl`/`ppl.explain`/`ppl.grammar`/`query.settings` action specs exist), so introducing a new `sql.explain` action spec felt like a lot of new surface for one regression test. Added `26ce5c514` instead, an integration test in `legacy/ExplainIT.java` (SQL's existing IT pattern) that hits `POST /_plugins/_sql/_explain?format=json` for real. It's the test that would have actually caught this - my original unit test on `isSupportedExplainFormat()` alone gave false confidence since it never touched `RestSqlAction` at all. Let me know if you'd still prefer a YAML test / new action spec added instead.

PR #5611, opened 2026-07-07
[Enhancement] Forward cluster planning settings to the Analytics Engine unified query path

RyanL1997:
### Description

Several cluster settings were **silently ignored on the Analytics Engine (unified query) path** — the AE route always planned against `UnifiedQueryContext.Builder`'s hardcoded seed value regardless of the configured cluster value. The default (non-AE) pipeline honored them correctly.

**Root cause.** `RestUnifiedQueryAction.applyClusterOverrides()` forwarded only a hand-picked subset of cluster settings into the `UnifiedQueryContext`, while `UnifiedQueryContext.Builder` independently seeds a default settings map. Any planning setting present in the seed map but absent from the forward list silently regressed to its default. This is a *two-lists-drift* defect: the two lists are maintained independently, so the drift is invisible until someone configures the setting and it does nothing.

### Settings fixed

| Setting | Value the AE path actually used | Effect |
|---|---|---|
| `plugins.query.size_limit` | pinned to `10000` | configured result cap ignored |
| `plugins.ppl.pattern.method` | pinned to `SIMPLE_PATTERN` | configured `patterns` default ignored |
| `plugins.ppl.pattern.mode` | pinned to `LABEL` | ditto |
| `plugins.ppl.pattern.max.sample.count` | pinned to `10` | ditto |
| `plugins.ppl.pattern.buffer.limit` | pinned to `100000` | ditto |
| `plugins.ppl.pattern.show.numbered.token` | pinned to `false` | ditto |

Every one of these has a cluster-side default identical to the seeded value, so **behavior is unchanged unless an operator explicitly configured the setting** — at which point the configured value now takes effect. No query that works today changes behavior on defaults.

### Fix

- Replace the hand-maintained `forwardClusterSetting` calls with a single `FORWARDED_CLUSTER_SETTINGS` allow-list — one source of truth for which cluster settings the unified path honors. This matches the existing shape of the sibling handler `RestQuerySettingsAction`, which declares its own settings allow/deny lists the same way.
- Add a **drift guard** (`everySeededPlanningSettingIsClassified`) asserting that every key the builder seeds is either forwarded or explicitly listed as deliberately excluded, with a reason. Adding a key to the builder's seed map without classifying it now fails a test instead of silently regressing. The guard reads the seeded keys through the existing public `Settings#getSettings()` API — no reflection, no new production API.

### Verified end to end on a live Analytics Engine cluster

Single-node cluster with `composite-engine`, `parquet-data-format`, `analytics-engine`, `analytics-backend-datafusion`, `analytics-backend-lucene` + this plugin, querying parquet-backed composite indices. Same cluster, same data, **baseline build vs this PR's build** — every forwarded key this PR adds:

| Setting probed | Baseline (setting ignored) | This PR (setting honored) |
|---|---|---|
| `query.size_limit=2` | 6 rows | **2 rows** |
| `query.size_limit=4` | 6 rows | **4 rows** |
| `pattern.mode=AGGREGATION` | 6 rows, schema `[age, name, patterns_field]` | **1 row, schema `[patterns_field, pattern_count, sample_logs]`** |
| `pattern.method=BRAIN` | `<*> <*> <*> <*> <*> <*>.<*>.<*>.<*> <*> <*> <*>` | **`user <*> logged in from <*IP*> at port <*>`** |
| `pattern.show.numbered.token=true` | `<*> <*> <*> …` | **`<token1> <token2> <token3> …`** |
| `pattern.max.sample.count=2` / `=5` | n/a — `pattern_count` field absent, since `pattern.mode` was ignored too | **`len(sample_logs)` = 2 / 5 respectively** |

`plugins.ppl.pattern.buffer.limit` is forwarded by the same mechanism but is **not** covered by a live-cluster probe: its minimum is `50000`, so distinguishing values requires more buffered patterns than a probe dataset produces. It is covered only by the unit test asserting the value reaches the plan context.

### Deliberately not forwarded

- **`plugins.calcite.enabled`** — the unified path is Calcite-based by definition and must force it on regardless of the cluster value.
- **`plugins.ppl.values.max.limit`** — forwarding this one makes the AE route **strictly worse**. The cap is applied by attaching a `limit` argument to `values()`, which lowers to `array_agg(DISTINCT x, limit)`; the DataFusion backend has no binding for that two-argument form. Confirmed on the cluster above: with the key forwarded, `stats values(f)` fails with `UnsupportedOperationException: Unable to find binding for call array_agg(DISTINCT $0, $1)`, surfaced as HTTP 500 — where today it merely ignores the cap. Turning a silent no-op into a hard failure is a regression, so the key stays out until the backend can bind the limited form. Tracked in #5736; the gap is already recorded in-repo as `Capability.VALUES_LIMIT_NOT_HONORED`.
- **`plugins.ppl.subsearch.maxout` / `plugins.ppl.join.subsearch_maxout`** — the builder seeds these to `0` (unlimited) *on purpose*, to keep `LogicalSystemLimit` out of plans built by external consumers of the unified query API. Note these **do diverge from the cluster defaults (`10000` / `50000`) even when unconfigured**, so the AE path currently runs subsearches unbounded. Whether the in-cluster REST path should override that documented choice is a separate behavioral decision, tracked in #5735.

One further gap found during this work is also **out of scope here**, tracked in #5734: `plugins.calcite.all_join_types.allowed` is not seeded, so `AstBuilder.validateJoinType` reads `null` and its `config != null && !config` check skips validation entirely — the high-cost-join guardrail is inactive on the AE path. Restoring it is a user-visible tightening, so it belongs in its own PR.

### Check List
- [x] New functionality has been documented in code.

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
@@ -10,9 +10,11 @@ import static org.opensearch.sql.lang.PPLLangSpec.PPL_SPEC;
 import static org.opensearch.sql.opensearch.executor.OpenSearchQueryManager.SQL_WORKER_THREAD_POOL_NAME;
 import static org.opensearch.sql.protocol.response.format.JsonResponseFormatter.Style.PRETTY;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.calcite.rel.RelNode;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -346,26 +349,60 @@ private static QueryRequestContext withParentTask(QueryRequestContext ctx, Task
     return new QueryRequestContext(ctx.clusterState(), ctx.schema(), ctx.querySource(), parentTask);
   }
 
+  /**
```
dai-chen, line 352: Could you simplify the javadoc here? Also are these 3 the only ones excluded?

RyanL1997: Both done in ab19073 — and your second question found a real gap, thanks.

**No, 3 was not the full list.** I re-derived it properly by tracing every `getSettingValue` reachable from the unified context's `Settings` (`SysLimit.fromSettings`; the `AstBuilder` / `AstExpressionBuilder` / `AstBuildGuard` the context's parser is built from; `UnresolvedPlanHelper`; `CalcitePlanContext`). The AE path reads **14** keys — 9 forwarded, **5** deliberately not:

| Not forwarded | Why |
|---|---|
| `CALCITE_ENGINE_ENABLED` | unified path is Calcite by definition |
| `PPL_SUBSEARCH_MAXOUT` | seeded unlimited on purpose for external API consumers (#5735) |
| `PPL_JOIN_SUBSEARCH_MAXOUT` | same (#5735) |
| `PPL_VALUES_MAX_LIMIT` | forwarding it **500s** the route (#5736) |
| `CALCITE_SUPPORT_ALL_JOIN_TYPES` | never seeded, so the high-cost-join guard is inactive here (#5734) |

dai-chen (approved):

PR #5612, opened 2026-07-07
[BugFix] Detect BIGINT overflow in SUM and fix wrong AVG on the Calcite engine

ahkcs:
### Description

`SUM`/`AVG` over integral columns near the `BIGINT` boundary could silently return wrong results on the Calcite engine:

- Local `SUM` could wrap its running `long` accumulator.
- Local `AVG(BIGINT)` could overflow its intermediate long sum before division.
- Native OpenSearch aggregation uses `double`, so pushed-down sums can lose low-order precision for large values.

### Fix

- Integral `SUM` inputs (`TINYINT`, `SMALLINT`, `INTEGER`, and `BIGINT`) use `CHECKED_LONG_SUM`.
- With pushdown disabled, `CheckedLongSumAggFunction` calls `Math.addExact` during every accumulation step.
- `CHECKED_LONG_SUM` reports `SqlKind.SUM`, preserving SUM planner rewrites and native pushdown.
- Pushdown uses OpenSearch's existing native `sum(field)` or scripted `sum(expression)` aggregation. No new OpenSearch aggregation is registered.
- The pushed result is range-checked and narrowed to `BIGINT`; backend precision already lost in `double` cannot be recovered.
- Bare `AVG(BIGINT)` uses a reflective `double sum + long count` aggregate locally and reports `SqlKind.AVG`, allowing pushdown to remain native `avg(field)` without a cast-generated script.
- FLOAT/DOUBLE arithmetic retains standard IEEE 754 behavior.

### Architecture

```text
PPL: stats sum(value), avg(value)
                 |
                 v
        Type-aware operator selection
                 |
       +---------+----------+
       |                    |
 pushdown disabled     pushdown enabled
       |                    |
       v                    v
 Calcite Enumerable     Planner sees standard kinds
 execution             CHECKED_LONG_SUM -> SqlKind.SUM
                       BIGINT_AVG      -> SqlKind.AVG
       |                    |
       |                    v
       |             OpenSearch AggregateAnalyzer
       |                    |
       |             +------+----------------+
       |             |                       |
       |        SUM(field)              SUM(expression)
       |        native field sum        native scripted sum
       |             |
       |        AVG(field)
       |        native field avg
       |
       +-- integral SUM
       |   Math.addExact(runningSum, value)
       |   -> exact long or overflow error
       |
       +-- bare BIGINT AVG
           double sum + long count
           -> sum / count
```

This separates execution implementation from planner identity: Calcite invokes the reflective checked aggregates locally, while `SqlKind.SUM`/`AVG` lets planner rules retain native OpenSearch pushdown.

### Precision behavior

The two execution paths intentionally have different precision guarantees:

```text
No pushdown:
  long accumulation with Math.addExact
  -> exact while every intermediate result fits in BIGINT

Pushdown:
  native OpenSearch double accumulation
  -> preserves existing DSL performance and semantics
  -> can lose low-order bits for large integers
```

Example:

```text
Values: 2^62 and 1
True sum: 4611686018427387905

No pushdown: 4611686018427387905
Pushdown:    may return 4611686018427387904
```

At that magnitude, adjacent `long` values cannot all be represented by `double`. `CheckedLongSumParser` can detect a non-finite or out-of-BIGINT-range result, but it cannot reconstruct bits already rounded by the backend.

`AVG` returns `DOUBLE` on both paths, so it follows double precision semantics. The local BIGINT-specific implementation prevents long overflow; it does not provide arbitrary-precision averaging.

### Overflow semantics

Local overflow is checked during accumulation, not only against the final mathematical result:

```text
Values: Long.MAX_VALUE, 1, -1

0 + Long.MAX_VALUE = Long.MAX_VALUE
Long.MAX_VALUE + 1 = overflow -> error
```

The later `-1` is not processed. A sequence whose final mathematical sum fits can therefore fail when an intermediate running sum exceeds the BIGINT range. This fail-fast behavior is intentional and follows `Math.addExact` accumulation semantics.

### Before / After

| Query/path | Before | After |
|---|---|---|
| Local integral `SUM` overflow | wrapped/wrong value | overflow error |
| Local integral `SUM` in range | could wrap during accumulation | exact `BIGINT` |
| Pushed integral `SUM` | native double semantics | unchanged native double semantics |
| Local `AVG(BIGINT)` near long overflow | wrong result | valid `DOUBLE` result |
| Pushed `AVG(BIGINT)` | cast-generated script in an intermediate revision | native `avg(field)` |
| FLOAT/DOUBLE `SUM` | IEEE 754 | unchanged |

### Testing

- Covered exact local sums, `Long.MAX_VALUE`, intermediate overflow, null handling, and all integral input types.
- Covered local BIGINT AVG without intermediate long overflow.
- Covered native field AVG, native field SUM, and scripted SUM request generation.
- Verified pushdown plans retain `AGGREGATION->` and bare field AVG no longer creates a script.
- Updated Calcite, no-pushdown, Big5, ClickBench, and explain fixtures.
- Verified the focused unit suites and all 265 `CalciteExplainIT` tests with pushdown disabled.

### Related

Addresses the aggregate half of #5164 and builds on #5604's checked scalar arithmetic.

### Check List


Swiddis (approved):

docs/user/ppl/functions/expressions.md
```diff
@@ -11,6 +11,10 @@ Arithmetic expressions are formed by combining numeric literals and binary arith
 4. `/`: Division. When [`plugins.ppl.syntax.legacy.preferred`](../admin/settings.md) is `true` (default), integer operands follow the legacy truncating result. When the setting is `false`, the operands are promoted to floating-point, preserving the fractional part. Division by zero returns `NULL`.
 5. `%`: Modulo. This operator can only be used with integers and returns the remainder of the division.
 
+### Overflow behavior
+
+Integer and long arithmetic operations (`+`, `-`, `*`) in `eval` expressions detect overflow and return an error instead of silently wrapping. For example, `eval x = int_field + 1` where `int_field` is `2147483647` (integer max) returns an error rather than `-2147483648`. Floating-point (`float`, `double`) arithmetic follows IEEE 754 and does not produce overflow errors.
+
```
penghuo, line 24: call-out alg difference between pushdown enabled vs disabled?

Two docs indexed: v = 4611686018427387904 (2^62) and v = 1. True sum = 4611686018427387905 (fits BIGINT exactly).

The no-pushdown PPL path (via widenBigintColumnToDecimal + CHECKED_LONG_NARROW) returns the exact bigint. The pushdown path silently loses precision because OpenSearch computes the sum in double, and 2^62 + 1 isn't representable

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1609,10 +1610,68 @@ void registerOperator(BuiltinFunctionName functionName, SqlAggFunction aggFuncti
...
+     * (the enumerable {@code SumImplementor} generates a plain {@code long + long}); accumulating
+     * in DECIMAL is exact. {@link CalciteRelNodeVisitor} casts the DECIMAL result back to BIGINT so
+     * the output type is unchanged and a genuine overflow surfaces as a client error instead of a
+     * wrong value.
+     *
+     * <p>Only a bare BIGINT column reference is widened. Expressions like {@code sum(field + 1)}
+     * are left as-is so {@link org.opensearch.sql.calcite.plan.rule.PPLAggregateConvertRule} can
+     * still rewrite them to pushdown-friendly {@code sum(field) OP literal} form.
+     */
+    void registerSumOperator() {
```
penghuo, line 1620: #5604 use Math.addExact to detect overflow. Why Sum choose another approach?

docs/user/ppl/functions/expressions.md
ahkcs: Added this distinction to the overflow documentation.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: #5604 and this PR now use the same checked-addition primitive, but at different planner levels.

#5604 handles scalar arithmetic such as `a + b`. Calcite exposes it as `RexCall(PLUS, a, b)`, so it can be rewritten to `CHECKED_PLUS`, which calls `Math.addExact` once.

`SUM` is an `AggregateCall`; its internal additions are not visible `PLUS` nodes, so the scalar rewrite cannot intercept them. We now solve that with a custom `CHECKED_LONG_SUM` aggregate:

- Non-pushdown: `CheckedLongSumAggFunction` calls `Math.addExact(accumulator, value)` for every row.
- Pushdown: native `checked_long_sum` calls `Math.addExact` during shard accumulation and coordinator reduction, and transports exact `long` values.
- `TINYINT`, `SMALLINT`, `INTEGER`, and `BIGINT` inputs all use this checked BIGINT accumulator.

This intentionally fails on an intermediate overflow even if later values would bring the final mathematical sum back into range. For example, `Long.MAX_VALUE, 1, -1` throws at `Long.MAX_VALUE + 1`. We consider that fail-fast, order-dependent behavior acceptable and consistent with checked Java/Spark-style accumulation.

plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
```diff
@@ -172,6 +176,17 @@ public String description() {
     return "Use sql to query OpenSearch.";
   }
 
+  @Override
+  public List<AggregationSpec> getAggregations() {
+    return List.of(
```
penghuo: Add an new sum aggregation in DSL? I do not think it is required. I am OK post-processing and DSL-pushdown has percision difference (this is DSL existing feature).

ahkcs: Agreed. I removed the custom DSL aggregation and getAggregations() registration. Pushdown now uses native OpenSearch sum with post-processing for BIGINT range validation, while accepting the existing native-double precision behavior. The non-pushdown path still uses Math.addExact during accumulation.

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/CheckedLongSumParser.java
```diff
@@ -0,0 +1,47 @@
...
+    double value = ((NumericMetricsAggregation.SingleValue) aggregation).value();
+    Long narrowed = Double.isNaN(value) ? null : narrow(value);
+    return Collections.singletonList(
+        new HashMap<>(Collections.singletonMap(aggregation.getName(), narrowed)));
+  }
+
+  static long narrow(double value) {
+    if (!Double.isFinite(value) || value > TWO_POW_63 || value < -TWO_POW_63) {
+      throw new ArithmeticException("BIGINT overflow in SUM");
+    }
```
penghuo, line 44: Any test cover this?

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1630,7 +1659,15 @@ void populate() {
 
       register(
           AVG,
-          (distinct, field, argList, ctx) -> ctx.relBuilder.avg(distinct, null, field),
+          (distinct, field, argList, ctx) -> {
+            if (field instanceof RexInputRef
```
penghuo: Why limit to RexInputRef? field could be RexCall, e.g. `avg(toInt(value))`.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1608,10 +1609,38 @@ void registerOperator(BuiltinFunctionName functionName, SqlAggFunction aggFuncti
...
+          };
+      register(SUM, handler, typeChecker);
+    }
+
+    private static boolean isIntegral(SqlTypeName typeName) {
+      return switch (typeName) {
+        case TINYINT, SMALLINT, INTEGER, BIGINT -> true;
+        default -> false;
+      };
+    }
```
penghuo, line 1632: Use "SqlTypeName.INT_TYPES.contains(...)"

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
```diff
@@ -1608,10 +1609,38 @@ void registerOperator(BuiltinFunctionName functionName, SqlAggFunction aggFuncti
...
+      AggHandler handler =
+          (distinct, field, argList, ctx) -> {
+            List<RexNode> newArgList =
+                argList.stream().map(PlanUtils::derefMapCall).collect(Collectors.toList());
+            boolean checkedLongSum = isIntegral(field.getType().getSqlTypeName());
+            SqlAggFunction sumOperator =
+                checkedLongSum ? PPLBuiltinOperators.CHECKED_LONG_SUM : SqlStdOperatorTable.SUM;
+            return UserDefinedFunctionUtils.makeAggregateCall(
+                sumOperator, List.of(field), newArgList, ctx.relBuilder);
+          };
```
penghuo: Code is duplicate with registerOperator. simplify it.

penghuo (changes requested):

opensearch/src/main/java/org/opensearch/sql/opensearch/response/agg/CheckedLongSumParser.java
ahkcs: Added unit and IT coverage for finite/infinite values, both boundaries, and positive/negative overflow.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: Agreed, removed it. All BIGINT expressions now use BIGINT_AVG, with RexCall IT coverage.

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: Updated

core/src/main/java/org/opensearch/sql/expression/function/PPLFuncImpTable.java
ahkcs: Updated

penghuo (approved):

PR #5613, opened 2026-07-09
Support PPL foreach command

songkant-aws:
### Description

Adds the PPL `foreach` command (Splunk-compatible), which runs a templated `eval` for each field in a field list, each element of a multivalue (array) field, or each element of a JSON array.

```
# multifield: double every matching field
source=idx | foreach value_* [ eval <<FIELD>>_double = <<FIELD>> * 2 ]

# multivalue: fold array elements into an accumulator
source=idx | eval nums = array(1,2,3), total = 0
           | foreach mode=multivalue nums [ eval total = total + <<ITEM>> ]

# json_array: parse JSON array text (literal or field) and iterate
source=idx | eval total = 0
           | foreach mode=json_array '[1,2,3]' [ eval total = total + <<ITEM>> ]
```

All four Splunk modes are supported: `multifield` (default), `multivalue`, `json_array`, and `auto_collections` (runtime detection). Placeholders `<<FIELD>>`/`<<MATCHSTR>>` (multifield) and `<<ITEM>>`/`<<ITER>>` (collection modes) are supported, with `fieldstr`/`matchstr`/`itemstr`/`iterstr` rename options.

#### Design

- **Multifield mode** expands each eval clause once per matching field via placeholder bindings on `CalcitePlanContext`; `<<FIELD>>` resolves to the current row field, `<<MATCHSTR>>` to the wildcard-captured segment.
- **Collection modes** rewrite each eval clause into a `reduce` call over the collection. Elements are packed into internal pairs `[item, iter, captured-field...]` by a `foreach_pair_collection` UDF so the two-parameter reduce lambda can reach the loop item, the loop index, and any referenced row fields; `foreach_pair_item(pair, slot)` extracts a slot with its plan-time type attached. Bindings are staged on the context and only activate inside the cloned lambda context, so the reduce call's own arguments resolve against the row normally.
- **json_array element typing**: `json_array(...)` calls and string literals are inspected at plan time (mixed string/number arrays rejected); for opaque expressions (a field holding JSON text — the primary Splunk use), the type is inferred from usage: arithmetic on `<<ITEM>>` selects DOUBLE, otherwise VARCHAR.
- Planning lives in a dedicated `ForeachPlanner` class rather than growing `CalciteRelNodeVisitor`.

#### Behavior verified against Splunk 10.4.0

| Scenario | Splunk | This PR |
| --- | --- | --- |
| mv field + multivalue | 6 | 6 ✓ |
| JSON-text field + json_array (sum of `"[10,20,30]"`) | 60 | 60 ✓ |
| json_array mode fed a real array | silent no-op | iterates (more permissive, pinned in IT) |
| multivalue mode fed JSON text | silent no-op | plan-time error (louder, pinned in IT) |

Known limitation (documented in tests and user doc): fields whose mapping is a scalar type (e.g. `long`) holding arrays are rejected by multivalue mode, because OpenSearch mappings do not declare array-ness and the plan-time type is scalar. Nested-typed fields map to `ARRAY<ANY>` and iterate correctly.

### Related Issues
N/A (new PPL command)

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -53,6 +53,7 @@ import org.opensearch.sql.ast.expression.Case;
 import org.opensearch.sql.ast.expression.Cast;
 import org.opensearch.sql.ast.expression.Compare;
 import org.opensearch.sql.ast.expression.EqualTo;
+import org.opensearch.sql.ast.expression.ForeachPlaceholder;
 import org.opensearch.sql.ast.expression.Function;
 import org.opensearch.sql.ast.expression.HighlightFunction;
 import org.opensearch.sql.ast.expression.In;
@@ -400,6 +401,19 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
   @Override
   public RexNode visitQualifiedName(QualifiedName node, CalcitePlanContext context) {
     return QualifiedNameResolver.resolve(node, context);
+  }
+
+  @Override
+  public RexNode visitForeachPlaceholder(ForeachPlaceholder node, CalcitePlanContext context) {
+    String name = node.getName().toUpperCase(Locale.ROOT);
+    String value = context.getForeachBindings().get(name);
+    if (value == null) {
+      throw new SemanticCheckException("Unresolved foreach placeholder <<" + node.getName() + ">>");
+    }
+    if ("FIELD".equals(name)) {
+      return context.relBuilder.field(value);
+    }
+    return context.rexBuilder.makeLiteral(value);
   }
 
   @Override
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -433,9 +433,10 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
         }
         return lambdaRef;
       case PAIR_ITEM:
-        return foreachPairItem(binding.value(), 0, name, context);
+      case PAIR_EXTRA:
+        return foreachPairItem(binding, name, context);
       case PAIR_ITER:
-        return foreachPairItem(binding.value(), 1, name, context);
+        return foreachPairItem(binding, name, context);
       case LITERAL:
       default:
         return context.rexBuilder.makeLiteral(binding.value());
@@ -443,16 +444,21 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
   }
 
   private RexNode foreachPairItem(
-      String pairName, int index, String placeholderName, CalcitePlanContext context) {
-    RexLambdaRef pair = context.getRexLambdaRefMap().get(pairName);
+      ForeachBinding binding, String placeholderName, CalcitePlanContext context) {
+    RexLambdaRef pair = context.getRexLambdaRefMap().get(binding.value());
     if (pair == null) {
       throw new SemanticCheckException("Unresolved foreach lambda placeholder " + placeholderName);
     }
     return PPLFuncImpTable.INSTANCE.resolve(
         context.rexBuilder,
-        BuiltinFunctionName.MVINDEX,
+        BuiltinFunctionName.FOREACH_PAIR_ITEM,
         pair,
-        context.rexBuilder.makeExactLiteral(BigDecimal.valueOf(index)));
+        context.rexBuilder.makeExactLiteral(BigDecimal.valueOf(binding.pairIndex())),
+        context.rexBuilder.makeLiteral(binding.typeName()),
+        binding.componentTypeName() == null
+            ? context.rexBuilder.makeNullLiteral(
+                context.rexBuilder.getTypeFactory().createSqlType(SqlTypeName.NULL))
+            : context.rexBuilder.makeLiteral(binding.componentTypeName()));
   }
 
   @Override
@@ -631,7 +637,8 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
 
     List<RexNode> capturedVars = null;
     try {
-      for (UnresolvedExpression arg : args) {
+      for (int argIndex = 0; argIndex < args.size(); argIndex++) {
+        UnresolvedExpression arg = args.get(argIndex);
         if (arg instanceof LambdaFunction) {
           CalcitePlanContext lambdaContext =
               prepareLambdaContext(
@@ -650,6 +657,14 @@ public class CalciteRexNodeVisitor extends AbstractNodeVisitor<RexNode, CalciteP
           arguments.add(lambdaNode);
           // Capture any external variables that were referenced in the lambda
           capturedVars = lambdaContext.getCapturedVariables();
+        } else if (node.getFuncName().equalsIgnoreCase("reduce") && argIndex == 0) {
+          Map<String, ForeachBinding> foreachBindings = new HashMap<>(context.getForeachBindings());
+          context.clearForeachBindings();
+          try {
+            arguments.add(analyze(arg, context));
+          } finally {
+            context.pushForeachBindings(foreachBindings);
+          }
         } else {
           arguments.add(analyze(arg, context));
         }
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
@@ -94,6 +94,7 @@ import org.opensearch.sql.exception.ExpressionEvaluationException;
 import org.opensearch.sql.exception.SemanticCheckException;
 import org.opensearch.sql.expression.function.BuiltinFunctionName;
 import org.opensearch.sql.expression.function.CoercionUtils;
+import org.opensearch.sql.expression.function.PPLBuiltinOperators;
 import org.opensearch.sql.expression.function.PPLFuncImpTable;
 
 @RequiredArgsConstructor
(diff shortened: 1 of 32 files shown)


songkant-aws:
Note for reviewers: two items from the new-command checklist are still pending and I'll push them to this PR shortly:
- `PPLQueryDataAnonymizer.visitForeach` (query anonymization)
- v2 `Analyzer.visitForeach` throwing the standard "only for Calcite" exception (currently the v2 path behavior is undefined)

songkant-aws:
Both pending checklist items are now in (`0e74b5d34`):
- `Analyzer.visitForeach` throws the standard only-for-Calcite `UnsupportedOperationException` on the v2 path, with a UT asserting the message.
- `PPLQueryDataAnonymizer.visitForeach` renders the command with masked options/targets/eval clauses (collection targets can embed literals such as JSON array strings); `ForeachPlaceholder` masks like a column reference. Covered by three anonymizer UTs (multifield, multivalue with options, json_array with a literal target).

The new-command checklist is now fully satisfied.

songkant-aws:
CI notes on the two failures from run 29075374167:
- **build-linux (25, unit)**: infrastructure — `Got socket exception during request. It might be caused by SSL misconfiguration` while fetching from Maven; unrelated to this change (same suites pass locally and on the fork's CI).
- **build-linux (21, integration)**: `testNoMvBasic`/`testNoMvWithEval` failed under the randomized `az-Cyrl` test locale — their default-locale `toLowerCase()` turns `ARRAY_JOIN` into `array_joın` (dotless i). Pre-existing bug in those tests (added by the timewrap PR, #5241); reproduced locally with the CI seed and fixed in `e2934613a` by using `Locale.ROOT`.

core/src/main/java/org/opensearch/sql/calcite/ForeachPlanner.java
```diff
@@ -0,0 +1,430 @@
...
+  private static final String PLACEHOLDER_FIELD = "FIELD";
+  private static final String PLACEHOLDER_MATCHSTR = "MATCHSTR";
+  private static final String PLACEHOLDER_MATCHSEG = "MATCHSEG";
+  private static final String PLACEHOLDER_ITEM = "ITEM";
+  private static final String PLACEHOLDER_ITER = "ITER";
+
+  /** Name of the lambda variable holding the internal pair inside generated reduce calls. */
+  private static final String PAIR_VAR = "__foreach_pair";
+
+  private static final Set<String> ARITHMETIC_OPERATORS = Set.of("+", "-", "*", "/", "%");
```
noCharger, line 81: 1. do we need comparisons (>)?
2. for numeric funcs, JSON element typed VARCHAR, would numeric compare becomes string compare? Ex: abs(<<ITEM>>), it treats the element as a string, so "9" > "10"

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/ForeachJsonArrayFunctionImpl.java
```diff
@@ -0,0 +1,91 @@
...
+      return null;
+    }
+  }
+
+  private static Object cast(Object value, SqlTypeName elementType) {
+    if (value == null) {
+      return null;
+    }
+    return switch (elementType) {
+      case DOUBLE -> ((Number) value).doubleValue();
```
noCharger, line 101: is this cast only on top level or reductive?

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ForeachPairCollectionFunctionImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return pairs;
+  }
+
+  private static List<?> toList(Object value) {
+    if (value instanceof List<?> list) {
+      return list;
+    }
+    if (value instanceof Object[] array) {
+      return List.of(array);
```
noCharger: List.of(array) throws NPE on null elements; JSON arrays can contain nulls. Use Arrays.asList

core/src/main/java/org/opensearch/sql/calcite/ForeachPlanner.java
```diff
@@ -0,0 +1,430 @@
...
+              List.of(context.relBuilder.alias(expr, alias)), List.of(alias), context);
+        }
+      } finally {
+        context.clearForeachBindings();
+      }
+    }
+    return context.relBuilder.peek();
+  }
+
+  private Map<String, ForeachBinding> multifieldBindings(
```
noCharger: Currently it is order-dependent. Later wildcard patterns never contribute.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
```diff
@@ -399,9 +402,51 @@ private boolean isBooleanLiteral(RexNode node) {
   /** Resolve qualified name. Note, the name should be case-sensitive. */
   @Override
   public RexNode visitQualifiedName(QualifiedName node, CalcitePlanContext context) {
+    ForeachBinding binding = foreachBinding(node.toString(), context);
+    if (binding != null) {
+      return foreachBindingToRexNode(node.toString(), binding, context);
+    }
```
noCharger: what would happen if a column named item/iter/field?

noCharger:
`multivalue` fed JSON text hard-errors while Splunk silently no-op. Is this on purpose?

core/src/main/java/org/opensearch/sql/calcite/ForeachPlanner.java
songkant-aws: Addressed in 273e330b3. Field-backed JSON element inference now handles numeric comparisons and numeric-only function arguments using the existing PPL function-signature metadata. `<<ITEM>> > 9` and `abs(<<ITEM>>)` therefore parse elements as DOUBLE rather than comparing strings. Covered in CalcitePPLForeachTest and by real execution tests in ForeachFieldJsonIT.

core/src/main/java/org/opensearch/sql/expression/function/jsonUDF/ForeachJsonArrayFunctionImpl.java
songkant-aws: This is top-level only, matching Splunk behavior. ForeachJsonArrayFunctionImpl now parses to Gson JsonArray and iterates only its direct children; nested arrays/objects remain serialized JSON values rather than being recursively traversed. ForeachFunctionImplTest covers nested array, nested object, and JSON null in one input.

core/src/main/java/org/opensearch/sql/expression/function/CollectionUDF/ForeachPairCollectionFunctionImpl.java
songkant-aws: Fixed in cc8fef4c3: Object[] is converted with Arrays.asList, so null elements are preserved without List.of throwing NPE. ForeachFunctionImplTest now explicitly covers a null collection element.

core/src/main/java/org/opensearch/sql/calcite/ForeachPlanner.java
songkant-aws: Fixed in cc8fef4c3. Exact matching patterns are considered before wildcard patterns for MATCHSTR/MATCHSEG capture, independent of input order; wildcard expansion order for the fields themselves remains stable. CalcitePPLForeachTest covers `*NO EMPNO` and verifies the exact EMPNO match wins.

core/src/main/java/org/opensearch/sql/calcite/CalciteRexNodeVisitor.java
songkant-aws: Fixed in cc8fef4c3. Default ITEM/ITER/FIELD names bind only `<<...>>` placeholders, so bare identifiers with those names continue to resolve as row fields. A bare identifier is shadowed only when the corresponding option explicitly renames the placeholder (for example, itemstr=ITEM), matching Splunk. Both collision cases are covered in CalcitePPLForeachTest.

songkant-aws:
@noCharger This was not intentional and is fixed in cc8fef4c3. The planner now checks the collection shape before wrapping it: `mode=multivalue` on JSON text and `mode=json_array` on a native array both leave the row unchanged, matching Splunk no-op behavior. The two directions are covered in CalcitePPLForeachTest, CalciteForeachCommandIT, and ForeachFieldJsonIT, and the affected foreach IT classes also pass with pushdown disabled.

noCharger (approved):
Thanks for the change!

docs/user/ppl/cmd/foreach.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Notes
+
+The following considerations apply when using the `foreach` command:
+
+* In collection modes, the bracketed `eval` acts as an accumulator: each target field must already exist (typically initialized with a preceding `eval`), and the expressions are applied once per element. Multiple assignments run from left to right, so a later assignment in the same iteration sees an earlier assignment's updated value.
+* In `json_array` mode the element type is inferred: a `json_array(...)` call or JSON string literal is inspected at plan time; for a field holding JSON text, elements are treated as numbers when `<<ITEM>>` is used in arithmetic and as strings otherwise. Mixed string/number JSON arrays are rejected.
+* Placeholders are also substituted inside string literals. For example, `eval <<FIELD>> = '<<FIELD>>'` replaces each selected field value with that field's name.
+* As in Splunk, `multivalue` mode applied to a non-array value and `json_array` mode applied to a native array are no-ops. A field whose mapping is scalar cannot be identified as multivalue at plan time even when a document stores several values, so `multivalue` mode also no-ops for that mapping.
```
penghuo, line 46: "As in Splunk". Remove it

docs/category.json
```diff
@@ -20,6 +20,7 @@
     "user/ppl/cmd/fieldformat.md",
     "user/ppl/cmd/fields.md",
     "user/ppl/cmd/fillnull.md",
+    "user/ppl/cmd/foreach.md",
```
penghuo, line 23: Update index.md, mark as experiment command.

docs/user/ppl/cmd/foreach.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+## Example 5: Iterate over a JSON array string
+
+The following query parses a JSON array literal and accumulates its elements:
+
+```ppl
+source=accounts
+| eval total = 0
+| foreach mode=json_array '[1,2,3]' [ eval total = total + <<ITEM>> ]
```
penghuo, line 147: Does TypeCoercion works?

Indexed a doc where jsonstrs is a text field holding ["a","b","c"] 

The crossed query

```
source=foreach_test | eval total = 0
| foreach mode=json_array jsonstrs [ eval total = total + <<ITEM>> ]
| fields total
```

Result — HTTP 500:
{"error":{"reason":"class java.lang.String cannot be cast to class java.lang.Number",
  "type":"ClassCastException"},"status":500}

songkant-aws: This opens up a broader runtime type-coercion problem. With fully dynamic semantics, we would ideally support all of the following cases:

| JSON array | `total = total + <<ITEM>>` | Runtime type |
| --- | --- | --- |
| `[1,2,3]` | `6` | Number |
| `["1","2","3"]` | `6` | Number |
| `[1,"2",3]` | `6` | Number |
| `["a","b","c"]` | `0abc` | String |
| `["1","a","2"]` | `1a2` | String |

At a high level, there are three possible approaches:

1. Use `ANY` or `VARIANT`. This would require explicit casts at typed operator boundaries and broader support across Calcite Enumerable, result-schema handling, and DataFusion/Substrait. Calcite Enumerable can compare some `ANY` values, but sorting by `VARIANT` is not fully supported without an explicit cast.
2. Follow the `spath` convention and return `VARCHAR` containing the stringified value. However, this introduces correctness risks for type-sensitive operators. For example, numeric values could be sorted lexicographically by their string representations.
3. Preserve strong static typing and require users to disambiguate operators and functions explicitly, for example by using numeric conversion or `concat`. This keeps the implementation scoped, but does not provide the dynamic behavior shown above.

For this follow-up, I suggest keeping the scope narrow: infer a concrete element type from the expression context and perform safe conversion in the JSON-array UDF. Incompatible values should produce `null` instead of an HTTP 500. Supporting dynamic `+` semantics should be handled separately as a broader runtime type-coercion design.

PR #5621, opened 2026-07-14
[Feature] Add PPL outputlookup command (synchronous terminal write sink)

noCharger:
### Description

Implements the PPL `outputlookup` command from **RFC #5625** — a synchronous, terminal write sink that materializes the current pipeline result into a lookup and returns a single `rows_written` count.

```
... | outputlookup [append=<bool>] [override_if_empty=<bool>] [key_field=<f1>(,<f2>)*] [max=<int>] <name>
```

Semantics, substrate, consistency contract, `<name>` resolution/migration, permissions, and alternatives are all in the RFC and are **not** restated here.

This PR also adds the operator ceiling **`plugins.ppl.outputlookup.max_rows`** (`NodeScope`, `Dynamic`, default `1_000_000`): a single call exceeding it fails with 400 and writes nothing (fail-loud, no truncated slice), orthogonal to the per-query `max=<int>` truncation.

**Tests**: `CalcitePPLOutputLookupIT` (18, incl. `testMaxRowsSettingRejectsExceeding`), `OutputLookupPermissionsIT` (2, incl. a read-privileged user reading the lookup through its alias), and unit tests.

**Benchmark** (3-node `m5.xlarge`): write-bound; per-batch refresh is ~2.5x of plain bulk (the main follow-up lever); 1M rows written un-truncated with no OOM; same-name concurrent overwrite is last-writer-wins with 13,532 reads and 0 torn/partial across 30 atomic repoints; crash-window self-heals on re-run. Full perf + chaos results and charts are in the comment below.

### Related Issues
Addresses #5625

### Check List


noCharger:
## Benchmark: performance and resilience

3-node `m5.xlarge`, node-side, single run per cell (directional).

### Performance (current build)

Measured on the current build: per-batch `RefreshPolicy.NONE` + one refresh before the alias repoint, plus build-time slice load settings (0 replicas, async translog, no auto refresh during the load, all restored to serving settings before publish) and bulk batch size 5000.

- Write-bound, not read-bound: the source scan is 0.01 to 0.2s at every size; the cost is the write.
- `outputlookup` write throughput is now on par with or faster than a plain `_bulk` load of the same rows. Overwrite, rows/s:

| write | `outputlookup` | plain `_bulk` | ratio |
|---|---|---|---|
| 100K / 2 cols | 25,802 | 27,533 | 0.94x |
| 100K / 5 cols | 30,980 | 19,984 | 1.55x |
| 100K / 20 cols | 11,145 | 11,329 | 0.98x |
| 1M / 2 cols | 44,358 | 31,275 | 1.42x |
| 1M / 5 cols | 32,982 | 24,043 | 1.37x |
| 1M / 20 cols | 13,774 | 10,976 | 1.25x |

- An earlier build using per-batch `RefreshPolicy.IMMEDIATE` was about 2.5x slower than plain bulk. Removing the per-batch refresh and loading the pre-publish slice with build-time settings eliminates that tax.
- Why it can match or beat plain bulk: the slice is invisible until the atomic alias repoint, so it is loaded like a build-time index (0 replicas, async translog, no auto refresh) and restored to serving settings before publish. The plain `_bulk` baseline used default settings (1 replica, request durability), so the ratios reflect each path at its realistic write config rather than identical durability.
- overwrite is approximately equal to append and to keyed upsert at scale: the alias repoint and the uuid-salted `_id` are minor next to the write.
- 1M rows written un-truncated via PIT, no OOM even at 1M x 20 cols (20M cells) on an 8GB heap.
- Read tax negligible: a 100-row lookup reads in about 11 to 15ms with 0 / 9 / 99 orphan slices in the backing index; the `__lookup` term filter is efficient, so the reaper is storage-justified, not read-latency-justified, at this scale.

### Resilience (chaos)
- Same-name concurrent overwrite, last-writer-wins, reads never torn: 6 writers x 5 rounds (30 overwrites), a dense poller over the alias ran 13,532 reads with 0 violations (no empty, partial, or mixed-writer read) across 30 observed atomic repoints; the alias settled on exactly one uuid with complete data.
- Crash-window self-heal: reconstructing the exact post-crash state (slice written, alias not yet repointed or added) and re-running the real command heals in both windows (overwrite-before-repoint, create-before-alias); during the window a read returns only the complete old slice, or does not resolve, the orphan is invisible through the filter and enumerable for the reaper.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -926,6 +926,67 @@ public RelNode visitReverse(
...
+                  + "] is not a field of the result; available fields: "
+                  + resultFields);
+        }
+      }
+    }
+    org.apache.calcite.plan.RelOptTable sourceTable = findSourceTable(child);
+    if (sourceTable == null) {
+      throw new IllegalArgumentException(
+          "outputlookup requires an OpenSearch source scan in the pipeline");
+    }
```
songkant-aws: This check seems weird because upstream could be `makeresults` or multi-source joined results from binary inputs

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,312 @@
...
+        break;
+      }
+      if (rows.size() > maxRows) {
+        throw new IllegalArgumentException(
+            "outputlookup wrote nothing because the result has more than "
+                + maxRows
+                + " rows, the maximum allowed for a single write. To write fewer rows, add"
+                + " max=<n> to your query. To allow more, raise the"
+                + " plugins.ppl.outputlookup.max_rows setting.");
+      }
```
songkant-aws, line 164: In case of `max=maxRows + 1`, it will write one extra row beyond the operator ceiling. `max=0` or a negative value also writes one row. Could we validate 1 <= max <= maxRows during planning?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupsIndex.java
```diff
@@ -0,0 +1,76 @@
...
+ * dynamic template maps every string field to {@code text} plus {@code keyword} and disables
+ * content-based detection, so lookups sharing the index never collide on field type and typing is
+ * deterministic.
+ */
+public final class LookupsIndex {
+
+  private LookupsIndex() {}
+
+  /** Fixed, single, not-configurable shared lookup store. */
+  public static final String INDEX_NAME = ".lookups";
```
songkant-aws: Does it have to be a hidden index? It could probably lead to confusion here. How about a business index with specific prefix(like index pattern mechanism) that has permission control? I'm not quite familiar with the security part here.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupsIndex.java
```diff
@@ -0,0 +1,76 @@
...
+    return Map.of(
+        "dynamic_templates",
+        List.of(Map.of("strings_as_text_and_keyword", stringAsTextAndKeyword)),
+        "date_detection",
+        false,
+        "numeric_detection",
+        false,
+        "properties",
+        Map.of(LOOKUP_FIELD, Map.of("type", "keyword")));
+  }
```
songkant-aws, line 52: The shared mapping is concerning. What happens when two different lookup destinations contain a field with the same name but incompatible types? Filtered aliases isolate documents, but not index mappings, so both lookups still share the .lookups mapping and one of the writes may fail with a mapping conflict. The same issue seems to apply when overwriting a lookup with a changed schema.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupsIndex.java
songkant-aws: Since the bulk request targets the shared .lookups index directly, how can Security permissions distinguish a user who may write lookup A but not lookup B? Granting bulk permission on .lookups appears to grant write access to every lookup slice, and filtered aliases do not constrain writes. index.hidden only affects discovery and is not an authorization boundary. Should each lookup use a dedicated backing index?

core/src/main/java/org/opensearch/sql/calcite/OutputLookupTableModify.java
```diff
@@ -0,0 +1,94 @@
...
+ * Terminal write node for outputlookup, modeled as a Calcite {@link TableModify} INSERT so the
+ * optimizer treats it as a mandatory table-modifying side effect (never dropped or reordered) and
+ * it exposes the standard rowcount row type. The referenced table is the source scan, used only so
+ * the lowering rule can reach the in-cluster client; the real destination lookup index is created
+ * and written at execution time by the physical operator, so a preexisting destination table is not
+ * required. Subclasses {@link TableModify} rather than LogicalTableModify so the built-in
+ * EnumerableTableModifyRule (which needs a ModifiableTable) does not fire on it.
+ */
+@Getter
+public class OutputLookupTableModify extends TableModify {
```
songkant-aws, line 32: You probably need to override https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/rel/core/TableModify.java#L217-L220 to change schema to be `rows_written`.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupIdEncoder.java
```diff
@@ -0,0 +1,92 @@
...
+  private static String canonical(Object value) {
+    if (value instanceof Boolean) {
+      return value.toString();
+    }
+    if (value instanceof Float || value instanceof Double) {
+      return Double.toString(((Number) value).doubleValue());
+    }
+    if (value instanceof Number) {
+      return Long.toString(((Number) value).longValue());
+    }
```
songkant-aws, line 74: BigDecimal is also a Number. Its longValue() method will discard any fractional part.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,312 @@
...
+    LookupsIndex.ensureExists(client);
+
+    Target target = resolveTarget(client, name);
+    switch (target.kind()) {
+      case ABSENT:
+        {
+          String uuid = newUuid();
+          writeSlice(client, LookupsIndex.INDEX_NAME, fields, mode, keyFields, rows, uuid);
+          addFilteredAlias(client, name, LookupsIndex.INDEX_NAME, uuid);
+          break;
```
songkant-aws: The class has no shared mutable state, but the lookup lifecycle is not concurrency-safe. In particular, two concurrent appends to an absent lookup can both resolve ABSENT, write separate UUID slices, and race to install the alias. Both requests may return success, while only the last alias target remains visible. Concurrent append and overwrite has a similar lost-visibility problem. What concurrency contract do we want for same-name writes?

For example, similar concurrent queries like `... | outputlookup append=true my_lookup` generate the following sequence, resulting orphan slice in the backing index:

Timestamp | Request A | Request B | alias status
-- | -- | -- | --
T1 | resolve → ABSENT |   | non-existent
T2 |   | resolve → ABSENT | non-existent
T3 | write uuid-A |   | non-existent
T4 |   | write uuid-B | non-existent
T5 | add alias → uuid-A |   | my_lookup → uuid-A
T6 | return success | add alias → uuid-B | my_lookup → uuid-B
T7 |   | return success | my_lookup → uuid-B

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
noCharger: > This check seems weird because upstream could be `makeresults` or multi-source joined results from binary inputs

Removed the source-scan requirement. The write now resolves the client from the schema instead, so makeresults|outputlookup and join inputs work. Added a sourceless IT.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
noCharger: > In case of `max=maxRows + 1`, it will write one extra row beyond the operator ceiling. `max=0` or a negative value also writes one row. Could we validate 1 <= max <= maxRows during planning?

Fixed. Validated 1 <= max <= max_rows before any write.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupsIndex.java
noCharger: > Does it have to be a hidden index? It could probably lead to confusion here. How about a business index with specific prefix(like index pattern mechanism) that has permission control? I'm not quite familiar with the security part here.

Fixed. Dropped the hidden index; a lookup is now an ordinary non-hidden per-lookup index (<name>__lookup).

> Since the bulk request targets the shared .lookups index directly, how can Security permissions distinguish a user who may write lookup A but not lookup B? Granting bulk permission on .lookups appears to grant write access to every lookup slice, and filtered aliases do not constrain writes. index.hidden only affects discovery and is not an authorization boundary. Should each lookup use a dedicated backing index?

Fixed as suggested. Each lookup is now a dedicated backing index, so write authz is per-index; index.hidden is no longer relied on as a boundary.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupsIndex.java
noCharger: > The shared mapping is concerning. What happens when two different lookup destinations contain a field with the same name but incompatible types? Filtered aliases isolate documents, but not index mappings, so both lookups still share the .lookups mapping and one of the writes may fail with a mapping conflict. The same issue seems to apply when overwriting a lookup with a changed schema.

Fixed. There is no shared backing anymore: each lookup has its own index, so the cross-lookup mapping conflict cannot arise. Overwrite replaces the whole slice, so a changed schema is fine.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
noCharger: > The class has no shared mutable state, but the lookup lifecycle is not concurrency-safe. In particular, two concurrent appends to an absent lookup can both resolve ABSENT, write separate UUID slices, and race to install the alias. Both requests may return success, while only the last alias target remains visible. Concurrent append and overwrite has a similar lost-visibility problem. What concurrency contract do we want for same-name writes?
> 
> For example, similar concurrent queries like `... | outputlookup append=true my_lookup` generate the following sequence, resulting orphan slice in the backing index:
> 
> Timestamp	Request A	Request B	alias status
> T1	resolve → ABSENT	 	non-existent
> T2	 	resolve → ABSENT	non-existent
> T3	write uuid-A	 	non-existent
> T4	 	write uuid-B	non-existent
> T5	add alias → uuid-A	 	my_lookup → uuid-A
> T6	return success	add alias → uuid-B	my_lookup → uuid-B
> T7	 	return success	my_lookup → uuid-B

Contract defined as append-to-absent useing a deterministic per-lookup discriminant so concurrent first-appends converge into one slice (no lost write); overwrite is last-writer-wins on the atomic repoint. Added a concurrent IT.

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -563,6 +563,12 @@ public LogicalPlan visitNoMv(NoMv node, AnalysisContext context) {
     throw getOnlyForCalciteException("nomv");
   }
 
+  @Override
+  public LogicalPlan visitOutputLookup(
+      org.opensearch.sql.ast.tree.OutputLookup node, AnalysisContext context) {
```
songkant-aws: nit: fix this import?

core/src/main/java/org/opensearch/sql/ast/tree/OutputLookup.java
```diff
@@ -0,0 +1,61 @@
...
+  /** false (default) overwrites the destination; true appends to it. */
+  private boolean append = false;
+
+  /** true (default) clears the destination on empty results; false keeps it. */
+  private boolean overrideIfEmpty = true;
+
+  /**
+   * Fields whose values form the document {@code _id} for upsert; empty means auto-generated id.
+   */
+  private List<String> keyFields = java.util.List.of();
```
songkant-aws: nit: fix imports

core/src/main/java/org/opensearch/sql/ast/AbstractNodeVisitor.java
```diff
@@ -305,6 +305,10 @@ public T visitReverse(Reverse node, C context) {
     return visitChildren(node, context);
   }
 
+  public T visitOutputLookup(org.opensearch.sql.ast.tree.OutputLookup node, C context) {
```
songkant-aws: nit: fix imports

core/src/main/java/org/opensearch/sql/calcite/plan/AbstractOpenSearchTable.java
```diff
@@ -22,4 +22,11 @@ public abstract class AbstractOpenSearchTable extends AbstractTable
   public RelDataType getRowType(RelDataTypeFactory relDataTypeFactory) {
     return OpenSearchTypeFactory.convertSchema(this);
   }
+
+  /**
+   * Write-lowering rules a write node registers itself when the pipeline has no OpenSearch scan.
+   */
+  public java.util.List<org.apache.calcite.plan.RelOptRule> getWriteConversionRules() {
+    return java.util.List.of();
```
songkant-aws: nit: fix imports

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -926,6 +926,59 @@ public RelNode visitReverse(
     return context.relBuilder.peek();
   }
 
+  @Override
+  public RelNode visitOutputLookup(
+      org.opensearch.sql.ast.tree.OutputLookup node, CalcitePlanContext context) {
```
songkant-aws: nit: fix imports

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -926,6 +926,59 @@ public RelNode visitReverse(
...
+    String name = node.getIndexName();
+    if (name.startsWith(".")) {
+      throw new IllegalArgumentException(
+          "outputlookup lookup name [" + name + "] must not be dot-prefixed");
+    }
+    RelNode child = context.relBuilder.build();
+    // Validate key_field names against the result schema: a missing/misspelled key field would
+    // otherwise encode identically for every row and collapse them into a single document.
+    if (!node.getKeyFields().isEmpty()) {
+      java.util.List<String> resultFields = child.getRowType().getFieldNames();
```
songkant-aws: nit: fix imports

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -926,6 +926,59 @@ public RelNode visitReverse(
...
+    org.apache.calcite.plan.RelOptSchema relOptSchema = context.relBuilder.getRelOptSchema();
+    if (!(relOptSchema instanceof org.apache.calcite.prepare.Prepare.CatalogReader catalogReader)) {
+      throw new IllegalStateException("outputlookup could not obtain a Calcite catalog reader");
+    }
+    // Resolve a table from the schema (not a scan) so sourceless pipelines reach the client; any
+    // OpenSearch name yields the same client, and the supplied row type avoids a mapping fetch.
+    OpenSearchSchema openSearchSchema =
+        context.config.getDefaultSchema().unwrap(OpenSearchSchema.class);
+    org.apache.calcite.schema.Table clientTable = openSearchSchema.getTable(name);
+    org.apache.calcite.plan.RelOptTable destTable =
```
songkant-aws: nit: fix imports

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -926,6 +926,59 @@ public RelNode visitReverse(
...
+          throw new IllegalArgumentException(
+              "outputlookup key_field ["
+                  + keyField
+                  + "] is not a field of the result; available fields: "
+                  + resultFields);
+        }
+      }
+    }
+    org.apache.calcite.plan.RelOptSchema relOptSchema = context.relBuilder.getRelOptSchema();
+    if (!(relOptSchema instanceof org.apache.calcite.prepare.Prepare.CatalogReader catalogReader)) {
```
songkant-aws: nit: fix imports

core/src/main/java/org/opensearch/sql/calcite/OutputLookupTableModify.java
```diff
@@ -0,0 +1,115 @@
...
+ * LogicalTableModify so the built-in EnumerableTableModifyRule (which needs a ModifiableTable) does
+ * not fire.
+ */
+@Getter
+public class OutputLookupTableModify extends TableModify {
+
+  private final String indexName;
+  private final boolean append;
+  private final boolean overrideIfEmpty;
+  private final java.util.List<String> keyFields;
```
songkant-aws: nit: fix imports

core/src/main/java/org/opensearch/sql/calcite/OutputLookupTableModify.java
```diff
@@ -0,0 +1,115 @@
...
+        .builder()
+        .add("rows_written", typeFactory.createSqlType(SqlTypeName.BIGINT))
+        .build();
+  }
+
+  /** Registers the write-lowering rule when no OpenSearch scan in the pipeline does. */
+  @Override
+  public void register(org.apache.calcite.plan.RelOptPlanner planner) {
+    org.opensearch.sql.calcite.plan.AbstractOpenSearchTable osTable =
+        table.unwrap(org.opensearch.sql.calcite.plan.AbstractOpenSearchTable.class);
```
songkant-aws: nit: fix imports

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchIndex.java
```diff
@@ -54,6 +54,11 @@
 /** OpenSearch table (index) implementation. */
 public class OpenSearchIndex extends AbstractOpenSearchTable {
 
+  @Override
+  public java.util.List<org.apache.calcite.plan.RelOptRule> getWriteConversionRules() {
+    return org.opensearch.sql.opensearch.planner.rules.OpenSearchIndexRules.WRITE_CONVERSION_RULES;
```
songkant-aws: nit: fix imports

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,358 @@
...
+      NodeClient client,
+      String name,
+      List<String> fields,
+      WriteMode mode,
+      List<String> keyFields,
+      @Nullable Integer max,
+      int maxRows,
+      boolean overrideIfEmpty,
+      boolean append,
+      Enumerator<@Nullable Object> input) {
```
songkant-aws, line 56: The input enumerator needs to be closed when bulk write finishes

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,358 @@
...
+    List<String> sliceKeys = keyFields;
+    if (mode == WriteMode.UPSERT) {
+      sliceKeys = new ArrayList<>(keyFields);
+      sliceKeys.add(LookupsIndex.LOOKUP_FIELD);
+    }
+
+    // The slice is not yet published, so drop redundancy/durability/auto-visibility during the load
+    // and restore them after the single refresh below (before the alias op): no replica writes, no
+    // per-request translog fsync, no background refresh churn while filling the slice.
+    applyLoadSettings(client, index);
```
songkant-aws: Each writeSlice will overwrite index level settings. Will it impact other regular indices?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,358 @@
...
+        System.arraycopy(row, 0, tagged, 0, row.length);
+        tagged[row.length] = uuid;
+        writer.add(tagged);
+      }
+    }
+    // Refresh once after the whole slice is written rather than per bulk batch: overwrite publishes
+    // the slice through the atomic alias repoint that follows, and append through this refresh, so
+    // per-batch refresh is redundant and dominates write cost at scale.
+    client.admin().indices().refresh(new RefreshRequest(index)).actionGet();
+    restoreServeSettings(client, index);
```
songkant-aws: It will never restore settings if writer or client throws exception

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupIdEncoder.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private static String canonical(Object value) {
+    if (value instanceof Boolean) {
+      return value.toString();
+    }
+    if (value instanceof Float || value instanceof Double) {
+      return Double.toString(((Number) value).doubleValue());
+    }
+    if (value instanceof java.math.BigDecimal bigDecimal) {
```
songkant-aws: nit: fix imports

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupIdEncoder.java
```diff
@@ -0,0 +1,106 @@
...
+      // longValue() would truncate the fraction and collide distinct decimals onto one id
+      return bigDecimal.stripTrailingZeros().toPlainString();
+    }
+    if (value instanceof java.math.BigInteger bigInteger) {
```
songkant-aws: nit: fix imports

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,358 @@
...
+
+    String backingIndex = name + LookupsIndex.BACKING_SUFFIX;
+
+    List<Object[]> rows = drain(input, max, maxRows);
+
+    if (!append && rows.isEmpty() && !overrideIfEmpty) {
+      return 0;
+    }
+
+    LookupsIndex.ensureExists(client, backingIndex);
```
songkant-aws: It will create an empty index when there is an existing lookup index, right? Then the empty index is never used.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,358 @@
...
+          } else {
+            String uuid = newUuid();
+            writeSlice(client, backingIndex, fields, mode, keyFields, rows, uuid);
+            repointFilteredAlias(client, name, target.aliasIndices(), backingIndex, uuid);
+            // TODO(reaper, separate PR): the atomic repoint leaves the previous slice as an
+            // orphan (a __lookup uuid in the backing index referenced by no alias). Crash-before-
+            // repoint and concurrent same-name overwrite produce the same orphan shape. A reaper
+            // reclaims them per backing index: enumerate distinct __lookup uuids, subtract the set
+            // referenced by any filtered alias, delete_by_query the remainder. This info log is the
+            // reaper's observability seam until then.
```
songkant-aws: I’m still concerned that maintaining multiple UUID slices in one backing index introduces unnecessary lifecycle complexity. Before the alias is repointed, a newly written slice is indistinguishable from an orphan to the proposed reaper, so safe cleanup would require additional coordination or a grace period. The generations also share mappings, index-level settings, and the reserved __lookup field, and cleanup requires delete_by_query.
Would one physical index per generation plus an atomic unfiltered-alias switch be simpler? Orphan generations would then be unreferenced managed indices and could be removed by deleting the entire index.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
noCharger: Removed this path entirely.

Two points:
1. Scope: the update targeted the single named backing index via `new UpdateSettingsRequest(index)`, so its effect stayed local to that one index.
2. The restore step wrote back hardcoded values (replicas=1, durability=request, refresh=default), which would clobber a backing index that carried custom settings.

To align with the merged data importer (OpenSearch-Dashboards#11303), which writes and ingests using default settings, I dropped `applyLoadSettings`/`restoreServeSettings`. outputlookup now writes with default settings, matching the importer.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
noCharger: Addressed in the current revision. `ensureExists` runs only on the two branches that write into `<name>__lookup`:
- new-lookup: create the backing index, then write.
- overwrite: create-if-absent, then write the new slice (the index is used immediately).

The append branch writes into the alias's existing primary index, so that path leaves backing-index creation out. The empty-shell case is resolved.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
noCharger: Aligned this with the merged precedent.

OpenSearch-Dashboards#11303 (the data importer) uses the same model: one index holding multiple slices keyed by a `__lookup=<uuid>` field, with a filtered alias `{term:{__lookup:<uuid>}}` published on a clean load. Field name, type, and alias-filter convention match, so outputlookup and the importer stay wire-compatible on the read side.

On orphan handling: #11303's alias update is add-only (a single `add` action), so a re-import repoints the alias and leaves the previous slice in place; it keeps the slice model and defers reclamation. I matched that here by removing the reaper/delete_by_query design. Overwrite repoints the filtered alias to the new slice in one atomic aliases request; reclaiming the superseded slice is deferred to a follow-up, consistent with the importer.

One deliberate difference, in the safer direction: the backing index is per-lookup (`<name>__lookup`), giving each lookup its own mapping and write boundary while keeping the identical `__lookup` + filtered-alias convention.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupIdEncoder.java
noCharger: Imports are ordered and cleaned up in the current revision.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupIdEncoder.java
noCharger: Imports are ordered and cleaned up in the current revision.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
noCharger: Contract for concurrent same-name writes:

- Concurrent first-time appends to an absent lookup converge on one slice via a deterministic per-lookup discriminant, so every appended row persists and the alias resolves to that slice. Covered by `testConcurrentAppendToAbsentDoesNotLoseWrites`.
- A concurrent overwrite is last-writer-wins on the atomic repoint. An append that races an overwrite may write into the slice the overwrite orphans, so those appended rows are acknowledged while staying outside the published lookup. Concrete interleaving on an existing lookup `hosts` (alias -> U0):

| T | A: outputlookup append=true hosts | B: outputlookup hosts (overwrite) | alias |
| --- | --- | --- | --- |
| T1 | resolve -> ALIAS (U0, primary hosts__lookup) | | hosts -> U0 |
| T2 | | resolve -> ALIAS (U0) | hosts -> U0 |
| T3 | | writeSlice(hosts__lookup, U1) | hosts -> U0 |
| T4 | | repoint -> U1 | hosts -> U1 (U0 orphaned) |
| T5 | writeSlice(hosts__lookup, U0) | | hosts -> U1 |
| T6 | return success | | hosts -> U1 |

  A wrote into U0, which the overwrite orphaned, so A's rows stay outside the published lookup while A reports success.

- When both writers observe the lookup as absent (one append, one overwrite), the alias resolves to whichever installs it last, so the surviving data comes from that writer.

This matches the merged data importer (OpenSearch-Dashboards#11303), which uses add-only alias updates and leaves cross-request coordination to the caller. I documented this contract in `docs/user/ppl/cmd/outputlookup.md` (Concurrency section) and recommend serializing writes to a given lookup, for example a single scheduled refresh per lookup, for deterministic results under contention. A per-lookup lock (control-index doc with seqNo/primaryTerm CAS) would eliminate the append-vs-overwrite window; I have it as a follow-up rather than in this PR to stay aligned with the importer precedent.

docs/user/ppl/index.md
```diff
@@ -84,6 +84,7 @@ source=accounts
 | [explain command](cmd/explain.md) | 3.1 | stable (since 3.1) | N/A | Explain the plan of query. |
 | [show datasources command](cmd/showdatasources.md) | 2.4 | stable (since 2.4) | N/A | Query datasources configured in the PPL engine. |
 | [makeresults command](cmd/makeresults.md) | 3.8 | experimental (since 3.8) | No | Generate in-memory rows for testing and seeding, optionally from inline CSV/JSON data. |
+| [outputlookup command](cmd/outputlookup.md) | 3.8 | experimental (since 3.8) | No | Write pipeline results into a lookup, read back with source=<name> or the lookup command. |
```
songkant-aws: nit: need to be changed to 3.9

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/LookupsIndex.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public final class LookupsIndex {
+
+  private LookupsIndex() {}
+
+  /** Per-row lookup slice discriminant. */
+  public static final String LOOKUP_FIELD = "__lookup";
+
+  /** Suffix of the dedicated per-lookup backing index derived from the lookup name. */
+  public static final String BACKING_SUFFIX = "__lookup";
```
songkant-aws, line 28: nit: A better namespace control could be metadata enrichment to avoid index naming conflicts with other business indices. Curious if we could add reserved metadata info to `_meta` to strengthen it?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,315 @@
...
+              writeSlice(
+                  client,
+                  target.primaryIndex(),
+                  fields,
+                  mode,
+                  keyFields,
+                  rows,
+                  target.lookupUuid());
+            } else {
+              String uuid = newUuid();
```
songkant-aws, line 120: Could we validate the target overwrite index to be valid lookup index?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
```diff
@@ -0,0 +1,315 @@
...
+    WriteConfig cfg =
+        new WriteConfig(index, sliceFields, mode, sliceKeys, BATCH_SIZE, RefreshPolicy.NONE);
+    try (OpenSearchBulkWriter writer = new OpenSearchBulkWriter(client, cfg)) {
+      for (Object[] row : rows) {
+        Object[] tagged = new Object[row.length + 1];
+        System.arraycopy(row, 0, tagged, 0, row.length);
+        tagged[row.length] = uuid;
+        writer.add(tagged);
+      }
+    }
```
songkant-aws, line 196: What would be partial write success/failure behavior? In case of append mode, could it generate partial duplicate rows? If we don't expect to handle it, add it to limitations.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/write/OutputLookupWriteExec.java
songkant-aws: Ok, I'm not quite familiar with data importer background. If our goal is to align with previous behavior, then we‘d better clarify all of limitations. Please also get the signoff from data importer author or other developer whoever knows the previous design decisions.

PR #5622, opened 2026-07-14
[Feature] Add PPL `makeresults` command

### LINKED ISSUE #3629: [FEATURE] Support data generation command
Yes. Currently, PPL does not support a built-in command for generating data. In documentation or integration tests, we must explicitly define a dataset in the index, which adds unnecessary overhead.

Introduce a data generation command in PPL. This command should behave similarly to [values](https://www.postgresql.org/docs/current/queries-values.html) / [Makeresults](https://docs.splunk.com/Documentation/SplunkCloud/latest/SearchReference/Makeresults) / [datatable](https://learn.microsoft.com/en-us/kusto/query/datatable-operator?view=microsoft-fabric). It would allow users to define inline rows for testing, examples, and quick queries without requiring an actual data source.

noCharger:
### Description

Adds the `makeresults` leading command on the Calcite (v3) path. It generates in-memory rows with no index scan.

**Count path**
- `| makeresults` / `| makeresults count=N` produces N rows (default 1), each with a single `@timestamp` (TIMESTAMP) column set to query time (`now()`, one value for the whole query).
- `@timestamp` is OpenSearch PPL's implicit time field (`OpenSearchConstants.IMPLICIT_FIELD_TIMESTAMP`), so the generated rows compose with the time-aware command family (`sort`, `stats ... by span`, `timechart`, `reverse`, `timewrap`). This intentionally uses `@timestamp` rather than a synthetic `_time` column.

**format/data path**
- `format=csv|json data="..."` parses an inline literal into typed rows. Column types are synthesized at plan time following OpenSearch dynamic-mapping semantics and surfaced through the same type path an index scan uses:
  - JSON integer → `long` (`bigint`), JSON decimal → `float` (`REAL`), JSON boolean → `boolean`, JSON string → `keyword`/`string`.
  - A typed CSV header token `name:type` uses the same vocabulary as `cast` (e.g. `age:int`); a bare CSV header token defaults to `string`.
  - Per-column heterogeneity widens to a common type; a cell that fails to parse to its declared type is a plan-time error.

Grammar lives in the `ppl/` copies only. The command adds the `MakeResults` AST node, `AstBuilder` wiring with `MakeResultsDataParser`, `CalciteRelNodeVisitor.visitMakeResults` (builds `LogicalValues` + `Project`), a V2 `Analyzer` reject stub (V2 returns the standard "supported only when `plugins.calcite.enabled=true`" error), and anonymizer rendering.

### Robustness / hardening

- `count=0` and header-only CSV build an empty typed `LogicalValues(tuples=[[]])` with the parsed rowType (previously threw a raw Calcite "Value count must be a positive multiple of field count"). A negative `count` also yields zero rows.
- All-null JSON column is rejected with a clean `SyntaxCheckException` (previously NPE via `Map.merge`).
- JSON big integers that overflow `long` keep full precision as a string column instead of silently wrapping.
- Nested JSON object/array values are rejected with a clear message (previously silently became `""`).
- CSV values containing commas are handled by a quote-aware splitter (RFC4180 `""` escape).
- Caps: `count` ≤ 1,000,000; `data` ≤ 29,999 chars.

### Notes / intentional divergences

- A JSON integer column reports its Calcite type name `bigint` in the response schema, because makeresults rows carry no index mapping to map `bigint` back to the PPL type name `long`.
- Inline object/array values and the `date`/`time`/`timestamp`/`ip`/`json` inline CSV types are not supported on this path (use `string` + `cast`).
- The count path output (`@timestamp = now()`) is non-deterministic, so it is not registered for doctest; the deterministic `data=` examples can be added after cluster validation.

### Related Issues
Closes #3629
RFC: #5626

### Check List
- [x] New functionality includes testing (unit + integration).
- [x] New functionality has been documented (`docs/user/ppl/cmd/makeresults.md`).
- [x] `spotlessCheck` passed.
- [x] Unit tests passed (`CalcitePPLMakeResultsTest` 13/13, `AstBuilderTest`, `PPLQueryDataAnonymizerTest`).
- [x] Integration tests passed on a local cluster (`CalcitePPLMakeResultsIT` 6/6, `NewAddedCommandsIT.testMakeResults`).


ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+    out.add(cur.toString());
+    return out;
+  }
+
+  private static ExprCoreType inferJsonType(JsonNode v) {
+    if (v.isNull()) {
+      return null;
+    }
+    if (v.isObject() || v.isArray()) {
```
songkant-aws, line 270: Do we expect to reject those nested field types? Or just fallback to serialized JSON strings? I think spath command could resolve the json string later.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLMakeResultsIT.java
```diff
@@ -0,0 +1,76 @@
...
+  @Test
+  public void testJson() throws IOException {
+    JSONObject result =
+        executeQuery(
+            "makeresults format=json data='[{\"name\":\"John\",\"age\":35,\"score\":3.5},"
+                + "{\"name\":\"Sarah\",\"age\":39,\"score\":4.0}]'");
+    verifySchema(
+        result, schema("name", "string"), schema("age", "bigint"), schema("score", "float"));
+    verifyDataRows(result, rows("John", 35, 3.5), rows("Sarah", 39, 4.0));
+  }
```
songkant-aws, line 83: We should generate `@timestamp` default time fields as well for specified json data.

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLMakeResultsIT.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Test
+  public void testTypedCsv() throws IOException {
+    JSONObject result =
+        executeQuery("makeresults format=csv data='name:string,age:int\nJohn,35\nSarah,39'");
+    verifySchema(result, schema("name", "string"), schema("age", "int"));
+    verifyDataRows(result, rows("John", 35), rows("Sarah", 39));
+  }
```
songkant-aws, line 91: Same as above

integ-test/src/test/java/org/opensearch/sql/calcite/standalone/CalcitePPLMakeResultsIT.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Test
+  public void testBareCsv() throws IOException {
+    JSONObject result = executeQuery("makeresults format=csv data='name,age\nJohn,35\nSarah,39'");
+    verifySchema(result, schema("name", "string"), schema("age", "string"));
+    verifyDataRows(result, rows("John", "35"), rows("Sarah", "39"));
+  }
```
songkant-aws, line 98: Same as above

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -287,6 +289,55 @@ public UnresolvedPlan visitRestCommand(OpenSearchPPLParser.RestCommandContext ct
...
+        format = "csv";
+      }
+    }
+    if (data != null || format != null) {
+      if (data == null || format == null) {
+        throw new SyntaxCheckException("makeresults format and data must be provided together");
+      }
+      if (data.length() > 29999) {
+        throw new SyntaxCheckException("makeresults data must not exceed 29999 characters");
+      }
```
songkant-aws: Nit: Claim this constraint in the md doc.

ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java
```diff
@@ -0,0 +1,340 @@
...
+    if (value == null) {
+      return null;
+    }
+    String s = String.valueOf(value);
+    try {
+      switch (type) {
+        case STRING:
+          return s;
+        case BOOLEAN:
+          return value instanceof Boolean ? value : Boolean.parseBoolean(s);
```
songkant-aws: Boolean.parseBoolean() will treat any invalid boolean value string to false. For example, Boolean.parseBoolean("not true") -> false.

songkant-aws:
> ## PR Code Suggestions ✨
> Latest suggestions up to [58ab1c4](https://github.com/opensearch-project/sql/commit/58ab1c4a0bde132b5878b3fb57c69af06e454db5)
> 
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> General	
> Handle empty first row edge case
> **When `rows` is not empty but contains an empty first row, accessing `rows.get(0).size()` could return 0 unexpectedly. Add validation to ensure the first row is not empty when inferring column count from rows, or handle this edge case explicitly.**
> 
> [core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java [4495-4502]](https://github.com/opensearch-project/sql/pull/5622/files#diff-518ddbde8cc139f2be35ffaeccd6dca4468147cc06f24f3458792e05584a61daR4495-R4502)
> 
> ```diff
>  if (explicitTypes != null) {
>    nc = explicitTypes.size();
>  } else if (explicitNames != null) {
>    nc = explicitNames.size();
> +} else if (!rows.isEmpty() && !rows.get(0).isEmpty()) {
> +  nc = rows.get(0).size();
>  } else {
> -  nc = rows.isEmpty() ? 0 : rows.get(0).size();
> +  nc = 0;
>  }
> ```
> 
> Suggestion importance[1-10]: 7
> __
> 
> Why: The suggestion addresses a potential edge case where `rows` is not empty but contains an empty first row. This could prevent unexpected behavior when `rows.get(0).size()` returns 0, improving robustness.
> 
> Low
> #### Previous suggestions
> Suggestions up to commit [1257c59](https://github.com/opensearch-project/sql/commit/1257c595c3ac45f5904a8208201224f253c4c030)
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> General	
> Detect unterminated CSV quoted fields
> **The CSV parser doesn't handle the case where a quoted field is not properly closed (unterminated quote). If `inQuotes` is still true after processing the entire line, it indicates malformed CSV data that should be rejected rather than silently accepted.**
> 
> [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [162-190]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR162-R190)
> 
> ```diff
>  private static List<String> splitCsvLine(String line) {
>    List<String> out = new ArrayList<>();
>    StringBuilder cur = new StringBuilder();
>    boolean inQuotes = false;
>    for (int i = 0; i < line.length(); i++) {
>      char c = line.charAt(i);
>      if (inQuotes) {
>        if (c == '"') {
>          if (i + 1 < line.length() && line.charAt(i + 1) == '"') {
>            cur.append('"');
>            i++;
>          } else {
>            inQuotes = false;
>          }
>        } else {
>          cur.append(c);
>        }
>      } else if (c == '"') {
>        inQuotes = true;
>      } else if (c == ',') {
>        out.add(cur.toString());
>        cur.setLength(0);
>      } else {
>        cur.append(c);
>      }
>    }
> +  if (inQuotes) {
> +    throw new SyntaxCheckException("makeresults CSV has unterminated quoted field");
> +  }
>    out.add(cur.toString());
>    return out;
>  }
> ```
> 
> Suggestion importance[1-10]: 8
> __
> 
> Why: Valid bug fix for malformed CSV input. The current implementation silently accepts unterminated quoted fields, which could lead to incorrect data parsing. Adding validation for `inQuotes` after the loop ensures proper CSV format and provides clear error messages for malformed input.
> 
> Medium
> Security	
> Validate row count during parsing
> **The row count validation occurs after parsing the entire data string, which could consume significant memory for large inputs. Consider validating the row count during parsing to fail fast and prevent potential memory exhaustion from malicious or malformed inputs.**
> 
> [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [57-59]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR57-R59)
> 
> ```diff
> -if (result.getRows() != null && result.getRows().size() > 5000) {
> -  throw new SyntaxCheckException("makeresults data must not exceed 5000 rows");
> -}
> +// Validate row count during parsing in parseJson/parseCsv methods
> +// to fail fast before materializing all rows in memory
> ```
> 
> Suggestion importance[1-10]: 7
> __
> 
> Why: Valid security improvement to prevent memory exhaustion by validating row count during parsing rather than after. However, the implementation would require significant refactoring of the parsing methods, and the current approach with a 29999 character limit already provides some protection.
> 
> Medium
> Suggestions up to commit [3a8654b](https://github.com/opensearch-project/sql/commit/3a8654b4ff29f5194a261b689b784d79c26e6309)
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Reject oversized integers explicitly
> **The fallback to STRING for integers exceeding long range may cause unexpected type inconsistencies. Consider throwing a SyntaxCheckException instead to alert users that their integer value is too large, ensuring data integrity and preventing silent type changes.**
> 
> [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [193-196]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR193-R196)
> 
> ```diff
>  if (v.isIntegralNumber()) {
> -  // A JSON integer wider than long keeps full precision as a string rather than overflowing.
> -  return v.canConvertToLong() ? ExprCoreType.LONG : ExprCoreType.STRING;
> +  if (!v.canConvertToLong()) {
> +    throw new SyntaxCheckException(
> +        "makeresults JSON integer value exceeds long range: " + v.asText());
> +  }
> +  return ExprCoreType.LONG;
>  }
> ```
> 
> Suggestion importance[1-10]: 5
> __
> 
> Why: The suggestion proposes throwing an exception for integers exceeding `long` range instead of falling back to `STRING`. While this would prevent silent type changes, the current behavior of preserving precision as a string is a reasonable design choice documented in the code comment. The suggestion improves consistency but isn't critical.
> 
> Low
> General	
> Validate empty CSV lines properly
> **Empty lines are silently skipped, which may hide malformed CSV data. Consider validating that skipped lines contain only whitespace, or throw an exception for truly empty lines to ensure data quality and prevent accidental data loss.**
> 
> [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [132-135]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR132-R135)
> 
> ```diff
>  for (int li = 1; li < lines.length; li++) {
> -  if (lines[li].isEmpty()) {
> +  if (lines[li].trim().isEmpty()) {
>      continue;
>    }
>    List<String> cells = splitCsvLine(lines[li]);
>    if (cells.size() > names.size()) {
>      throw new SyntaxCheckException(
>          "makeresults CSV row has more columns than the header: " + lines[li]);
>    }
> ```
> 
> Suggestion importance[1-10]: 3
> __
> 
> Why: The suggestion to use `trim().isEmpty()` instead of `isEmpty()` is a minor improvement that handles lines with only whitespace more consistently. However, the current behavior is acceptable for typical CSV parsing scenarios and the impact is minimal.
> 
> Low
> Suggestions up to commit [8e6ecc8](https://github.com/opensearch-project/sql/commit/8e6ecc8c20959e96dd2ebc8cd5a57be7a34f544b)
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Fix unreachable empty array condition
> **The condition `lines.length == 0` can never be true because `split()` with a limit of -1 always returns at least one element (even for an empty string). This makes the first part of the condition unreachable. Consider checking if the array is empty or if the first element is empty after splitting.**
> 
> [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [96-99]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR96-R99)
> 
> ```diff
>  String[] lines = data.split("\r?\n", -1);
> -if (lines.length == 0 || lines[0].trim().isEmpty()) {
> +if (lines.length == 0 || lines[0].isEmpty()) {
>    throw new SyntaxCheckException("makeresults CSV data must start with a header line");
>  }
> ```
> 
> Suggestion importance[1-10]: 7
> __
> 
> Why: The suggestion correctly identifies that `split()` with limit -1 always returns at least one element, making `lines.length == 0` unreachable. However, the improved code removes `.trim()` from the second condition, which changes the behavior from checking if the trimmed first line is empty to checking if the raw first line is empty. This could introduce inconsistency with the intended validation logic.
> 
> Medium
> Security	
> Add upper bound for count parameter
> **When `count` is zero, an empty array is created and the loop doesn't execute, which is correct. However, when `count` is very large (e.g., Integer.MAX_VALUE), this could cause an OutOfMemoryError. Consider adding a reasonable upper bound check to prevent memory exhaustion attacks.**
> 
> [core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java [4506-4512]](https://github.com/opensearch-project/sql/pull/5622/files#diff-518ddbde8cc139f2be35ffaeccd6dca4468147cc06f24f3458792e05584a61daR4506-R4512)
> 
> ```diff
>  int count = node.getCount();
> +if (count > 10000) {
> +  throw new IllegalArgumentException("makeresults count exceeds maximum allowed value of 10000");
> +}
>  // Build `count` one-column dummy rows, then project them away with a single _time column set
>  // to query time. The dummy column only carries row multiplicity; count=0 yields zero rows.
>  Object[] dummy = new Object[count];
>  for (int i = 0; i < count; i++) {
>    dummy[i] = i;
>  }
> ```
> 
> Suggestion importance[1-10]: 6
> __
> 
> Why: The suggestion raises a valid concern about potential memory exhaustion with very large `count` values. However, the specific limit of 10000 is arbitrary and not justified by the PR context. A more appropriate approach would involve configuration-based limits or resource-aware validation rather than a hardcoded constant.
> 
> Low
> General	
> Handle whitespace-only lines consistently
> **Empty lines are skipped but trimmed empty lines are not, which could lead to inconsistent behavior. A line containing only whitespace will be processed and may cause unexpected results. Consider using `trim().isEmpty()` for consistency with the header check.**
> 
> [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [119-127]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR119-R127)
> 
> ```diff
>  for (int li = 1; li < lines.length; li++) {
> -  if (lines[li].isEmpty()) {
> +  if (lines[li].trim().isEmpty()) {
>      continue;
>    }
>    String[] cells = lines[li].split(",", -1);
>    if (cells.length > names.size()) {
>      throw new SyntaxCheckException(
>          "makeresults CSV row has more columns than the header: " + lines[li]);
>    }
> ```
> 
> Suggestion importance[1-10]: 5
> __
> 
> Why: The suggestion identifies a minor inconsistency where whitespace-only lines are not skipped like empty lines. While adding `.trim()` would improve consistency, this is a relatively minor issue that doesn't affect correctness significantly, as whitespace-only lines would still be parsed (potentially as rows with empty cells).
> 
> Low

Also, these edge cases have some point as well.

ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java
```diff
@@ -0,0 +1,340 @@
...
+        ExprCoreType declared = resolveType(maybeType);
+        if (declared != null) {
+          names.add(t.substring(0, colon).trim());
+          types.add(declared);
+          continue;
+        }
+      }
+      names.add(t);
+      types.add(ExprCoreType.STRING);
+    }
```
songkant-aws, line 196: How about dedupe the field names while processing csv headers? Or we can check if Project can handle them. I remember `SqlValidatorUtil.uniquify` could append identity numbers to avoid them. For example, 'field, field' names will be uniquified as 'field, field0'

core/src/main/java/org/opensearch/sql/analysis/Analyzer.java
```diff
@@ -561,6 +561,12 @@ public LogicalPlan visitNoMv(NoMv node, AnalysisContext context) {
     throw getOnlyForCalciteException("nomv");
   }
 
+  @Override
+  public LogicalPlan visitMakeResults(
+      org.opensearch.sql.ast.tree.MakeResults node, AnalysisContext context) {
```
songkant-aws: Nit: avoid this full path `org.opensearch.sql.ast.tree.MakeResults`, use import

ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java
songkant-aws: And how about the blank field handling? Like ',field'?

ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java
noCharger: revised to fallback to serialized JSON strings

noCharger:
> > ## PR Code Suggestions ✨
> > Latest suggestions up to [58ab1c4](https://github.com/opensearch-project/sql/commit/58ab1c4a0bde132b5878b3fb57c69af06e454db5)
> > Explore these optional code suggestions:
> > **Category**	**Suggestion                                                                                                                                    **	**Impact**
> > General
> > Handle empty first row edge case
> > **When `rows` is not empty but contains an empty first row, accessing `rows.get(0).size()` could return 0 unexpectedly. Add validation to ensure the first row is not empty when inferring column count from rows, or handle this edge case explicitly.**
> > [core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java [4495-4502]](https://github.com/opensearch-project/sql/pull/5622/files#diff-518ddbde8cc139f2be35ffaeccd6dca4468147cc06f24f3458792e05584a61daR4495-R4502)
> > ```diff
> >  if (explicitTypes != null) {
> >    nc = explicitTypes.size();
> >  } else if (explicitNames != null) {
> >    nc = explicitNames.size();
> > +} else if (!rows.isEmpty() && !rows.get(0).isEmpty()) {
> > +  nc = rows.get(0).size();
> >  } else {
> > -  nc = rows.isEmpty() ? 0 : rows.get(0).size();
> > +  nc = 0;
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestion importance[1-10]: 7
> > __
> > Why: The suggestion addresses a potential edge case where `rows` is not empty but contains an empty first row. This could prevent unexpected behavior when `rows.get(0).size()` returns 0, improving robustness.
> > Low
> > #### Previous suggestions
> > Suggestions up to commit [1257c59](https://github.com/opensearch-project/sql/commit/1257c595c3ac45f5904a8208201224f253c4c030)
> > **Category**	**Suggestion                                                                                                                                    **	**Impact**
> > General
> > Detect unterminated CSV quoted fields
> > **The CSV parser doesn't handle the case where a quoted field is not properly closed (unterminated quote). If `inQuotes` is still true after processing the entire line, it indicates malformed CSV data that should be rejected rather than silently accepted.**
> > [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [162-190]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR162-R190)
> > ```diff
> >  private static List<String> splitCsvLine(String line) {
> >    List<String> out = new ArrayList<>();
> >    StringBuilder cur = new StringBuilder();
> >    boolean inQuotes = false;
> >    for (int i = 0; i < line.length(); i++) {
> >      char c = line.charAt(i);
> >      if (inQuotes) {
> >        if (c == '"') {
> >          if (i + 1 < line.length() && line.charAt(i + 1) == '"') {
> >            cur.append('"');
> >            i++;
> >          } else {
> >            inQuotes = false;
> >          }
> >        } else {
> >          cur.append(c);
> >        }
> >      } else if (c == '"') {
> >        inQuotes = true;
> >      } else if (c == ',') {
> >        out.add(cur.toString());
> >        cur.setLength(0);
> >      } else {
> >        cur.append(c);
> >      }
> >    }
> > +  if (inQuotes) {
> > +    throw new SyntaxCheckException("makeresults CSV has unterminated quoted field");
> > +  }
> >    out.add(cur.toString());
> >    return out;
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestion importance[1-10]: 8
> > __
> > Why: Valid bug fix for malformed CSV input. The current implementation silently accepts unterminated quoted fields, which could lead to incorrect data parsing. Adding validation for `inQuotes` after the loop ensures proper CSV format and provides clear error messages for malformed input.
> > Medium
> > Security
> > Validate row count during parsing
> > **The row count validation occurs after parsing the entire data string, which could consume significant memory for large inputs. Consider validating the row count during parsing to fail fast and prevent potential memory exhaustion from malicious or malformed inputs.**
> > [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [57-59]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR57-R59)
> > ```diff
> > -if (result.getRows() != null && result.getRows().size() > 5000) {
> > -  throw new SyntaxCheckException("makeresults data must not exceed 5000 rows");
> > -}
> > +// Validate row count during parsing in parseJson/parseCsv methods
> > +// to fail fast before materializing all rows in memory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestion importance[1-10]: 7
> > __
> > Why: Valid security improvement to prevent memory exhaustion by validating row count during parsing rather than after. However, the implementation would require significant refactoring of the parsing methods, and the current approach with a 29999 character limit already provides some protection.
> > Medium
> > Suggestions up to commit [3a8654b](https://github.com/opensearch-project/sql/commit/3a8654b4ff29f5194a261b689b784d79c26e6309)
> > **Category**	**Suggestion                                                                                                                                    **	**Impact**
> > Possible issue
> > Reject oversized integers explicitly
> > **The fallback to STRING for integers exceeding long range may cause unexpected type inconsistencies. Consider throwing a SyntaxCheckException instead to alert users that their integer value is too large, ensuring data integrity and preventing silent type changes.**
> > [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [193-196]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR193-R196)
> > ```diff
> >  if (v.isIntegralNumber()) {
> > -  // A JSON integer wider than long keeps full precision as a string rather than overflowing.
> > -  return v.canConvertToLong() ? ExprCoreType.LONG : ExprCoreType.STRING;
> > +  if (!v.canConvertToLong()) {
> > +    throw new SyntaxCheckException(
> > +        "makeresults JSON integer value exceeds long range: " + v.asText());
> > +  }
> > +  return ExprCoreType.LONG;
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestion importance[1-10]: 5
> > __
> > Why: The suggestion proposes throwing an exception for integers exceeding `long` range instead of falling back to `STRING`. While this would prevent silent type changes, the current behavior of preserving precision as a string is a reasonable design choice documented in the code comment. The suggestion improves consistency but isn't critical.
> > Low
> > General
> > Validate empty CSV lines properly
> > **Empty lines are silently skipped, which may hide malformed CSV data. Consider validating that skipped lines contain only whitespace, or throw an exception for truly empty lines to ensure data quality and prevent accidental data loss.**
> > [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [132-135]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR132-R135)
> > ```diff
> >  for (int li = 1; li < lines.length; li++) {
> > -  if (lines[li].isEmpty()) {
> > +  if (lines[li].trim().isEmpty()) {
> >      continue;
> >    }
> >    List<String> cells = splitCsvLine(lines[li]);
> >    if (cells.size() > names.size()) {
> >      throw new SyntaxCheckException(
> >          "makeresults CSV row has more columns than the header: " + lines[li]);
> >    }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestion importance[1-10]: 3
> > __
> > Why: The suggestion to use `trim().isEmpty()` instead of `isEmpty()` is a minor improvement that handles lines with only whitespace more consistently. However, the current behavior is acceptable for typical CSV parsing scenarios and the impact is minimal.
> > Low
> > Suggestions up to commit [8e6ecc8](https://github.com/opensearch-project/sql/commit/8e6ecc8c20959e96dd2ebc8cd5a57be7a34f544b)
> > **Category**	**Suggestion                                                                                                                                    **	**Impact**
> > Possible issue
> > Fix unreachable empty array condition
> > **The condition `lines.length == 0` can never be true because `split()` with a limit of -1 always returns at least one element (even for an empty string). This makes the first part of the condition unreachable. Consider checking if the array is empty or if the first element is empty after splitting.**
> > [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [96-99]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR96-R99)
> > ```diff
> >  String[] lines = data.split("\r?\n", -1);
> > -if (lines.length == 0 || lines[0].trim().isEmpty()) {
> > +if (lines.length == 0 || lines[0].isEmpty()) {
> >    throw new SyntaxCheckException("makeresults CSV data must start with a header line");
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestion importance[1-10]: 7
> > __
> > Why: The suggestion correctly identifies that `split()` with limit -1 always returns at least one element, making `lines.length == 0` unreachable. However, the improved code removes `.trim()` from the second condition, which changes the behavior from checking if the trimmed first line is empty to checking if the raw first line is empty. This could introduce inconsistency with the intended validation logic.
> > Medium
> > Security
> > Add upper bound for count parameter
> > **When `count` is zero, an empty array is created and the loop doesn't execute, which is correct. However, when `count` is very large (e.g., Integer.MAX_VALUE), this could cause an OutOfMemoryError. Consider adding a reasonable upper bound check to prevent memory exhaustion attacks.**
> > [core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java [4506-4512]](https://github.com/opensearch-project/sql/pull/5622/files#diff-518ddbde8cc139f2be35ffaeccd6dca4468147cc06f24f3458792e05584a61daR4506-R4512)
> > ```diff
> >  int count = node.getCount();
> > +if (count > 10000) {
> > +  throw new IllegalArgumentException("makeresults count exceeds maximum allowed value of 10000");
> > +}
> >  // Build `count` one-column dummy rows, then project them away with a single _time column set
> >  // to query time. The dummy column only carries row multiplicity; count=0 yields zero rows.
> >  Object[] dummy = new Object[count];
> >  for (int i = 0; i < count; i++) {
> >    dummy[i] = i;
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestion importance[1-10]: 6
> > __
> > Why: The suggestion raises a valid concern about potential memory exhaustion with very large `count` values. However, the specific limit of 10000 is arbitrary and not justified by the PR context. A more appropriate approach would involve configuration-based limits or resource-aware validation rather than a hardcoded constant.
> > Low
> > General
> > Handle whitespace-only lines consistently
> > **Empty lines are skipped but trimmed empty lines are not, which could lead to inconsistent behavior. A line containing only whitespace will be processed and may cause unexpected results. Consider using `trim().isEmpty()` for consistency with the header check.**
> > [ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java [119-127]](https://github.com/opensearch-project/sql/pull/5622/files#diff-d6bcaef0c6646273d60e95d7435b2a3712458ff6250ba4b6a0f620e13488abdcR119-R127)
> > ```diff
> >  for (int li = 1; li < lines.length; li++) {
> > -  if (lines[li].isEmpty()) {
> > +  if (lines[li].trim().isEmpty()) {
> >      continue;
> >    }
> >    String[] cells = lines[li].split(",", -1);
> >    if (cells.length > names.size()) {
> >      throw new SyntaxCheckException(
> >          "makeresults CSV row has more columns than the header: " + lines[li]);
> >    }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestion importance[1-10]: 5
> > __
> > Why: The suggestion identifies a minor inconsistency where whitespace-only lines are not skipped like empty lines. While adding `.trim()` would improve consistency, this is a relatively minor issue that doesn't affect correctness significantly, as whitespace-only lines would still be parsed (potentially as rows with empty cells).
> > Low
> 
> Also, these edge cases have some point as well.

Thanks, went through all of these. Taken:

* Empty first-row guard (buildLiteralValues): nc = 0 when the first row is empty instead of reading it.
* Unterminated CSV quote: splitCsvLine throws if inQuotes is still open after the line.
* Whitespace-only CSV lines: data lines are now skipped with trim().isEmpty(), consistent with the header check.

Considered and kept as-is:

* Oversized JSON integer to string: deliberate, a value wider than long keeps full precision as a string rather than throwing (documented in code).
* Validate row count during parsing: the data string is already capped at 29999 chars before parsing, so a single parse cannot exhaust memory; the 5000-row check is a codegen-safety bound, not a memory guard.
* count upper bound: already enforced, count is capped at 5000 (beyond that the generated Calcite bytecode exceeds the JVM 64 KB per-method limit), stricter than the suggested 10000.
* Unreachable lines.length == 0: harmless defensive branch next to the real lines[0].trim().isEmpty() header check. Left as-is.

ppl/src/main/java/org/opensearch/sql/ppl/utils/MakeResultsDataParser.java
noCharger: > How about dedupe the field names while processing csv headers? Or we can check if Project can handle them. I remember `SqlValidatorUtil.uniquify` could append identity numbers to avoid them. For example, 'field, field' names will be uniquified as 'field, field0'

> And how about the blank field handling? Like ',field'?

both resolved 9dc627785

docs/user/ppl/index.md
```diff
@@ -80,6 +80,7 @@ source=accounts
 | [describe command](cmd/describe.md) | 2.1 | stable (since 2.1) | Query the metadata of an index. |
 | [explain command](cmd/explain.md) | 3.1 | stable (since 3.1) | Explain the plan of query. |
 | [show datasources command](cmd/showdatasources.md) | 2.4 | stable (since 2.4) | Query datasources configured in the PPL engine. |
+| [makeresults command](cmd/makeresults.md) | 3.5 | experimental | Generate in-memory rows for testing and seeding, optionally from inline CSV/JSON data. |
```
songkant-aws: Nit: version 3.8?

songkant-aws (approved):

PR #5573, opened 2026-06-19
Pin GitHub Actions to commit SHAs

ritvibhatt:
### Description
Pin all GitHub Action tag references to their corresponding commit SHAs.

Tags are mutable references that can be force-pushed to point to different commits, making them vulnerable to supply chain attacks. Commit SHAs are immutable and guarantee that the exact reviewed code is executed in CI workflows. This change pins all third-party actions to their current commit SHAs to prevent potential tampering.

Adding this change from main: https://github.com/opensearch-project/sql/pull/5464

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
 - [ ] New functionality has javadoc added.
 - [ ] New functionality has a user manual doc added.
- [ ] New PPL command [checklist](https://github.com/opensearch-project/sql/blob/main/docs/dev/ppl-commands.md) all confirmed.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5624, opened 2026-07-14
[Backport 2.19-dev] Pin async-query-core publish actions to commit SHAs

ahkcs:
### Description

Pins all actions in `publish-async-query-core.yml` to full commit SHAs, matching the versions pinned on `main` by #5464.

The 2.19-dev pinning backport #5617 was based on #5573, whose source branch did not contain this workflow, so these five action references remained unpinned. This caused the July 10 async-query-core publication to fail during action preparation: https://github.com/opensearch-project/sql/actions/runs/29123519082

### Testing

- `git diff --check`
- Verified every `uses:` entry in the workflow is pinned to a full commit SHA

Related: #5464, #5573, #5617

### DIFF
--- a/.github/workflows/publish-async-query-core.yml
+++ b/.github/workflows/publish-async-query-core.yml
@@ -33,15 +33,15 @@ jobs:
       contents: write
 
     steps:
-      - uses: actions/setup-java@v3
+      - uses: actions/setup-java@17f84c3641ba7b8f6deff6309fc4c864478f5d62 # v3
         with:
           distribution: temurin
           java-version: 17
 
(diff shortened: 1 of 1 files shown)


RyanL1997 (approved):

noCharger (approved):
While scanning `2.19-dev`, `delete_backport_branch.yml` still has one unpinned action:

    uses: SvanBoxel/delete-merged-branch@main

Branch ref will hit the same "must be pinned to a full-length commit SHA" rejection. Worth a follow-up PR to pin it. Everything else on 2.19-dev is SHA-pinned.

PR #5628, opened 2026-07-15
feat: complex query thread pool

### LINKED ISSUE #5608: [RFC] Some Performance/Stability Ideas for Rex
## Problem Statement

There's general demand to allow more performant and stable `rex` (and by similar token, `parse`) support over Lucene tables. `rex` is a command which can take a regex pattern with capture groups and extract it into fields. For example, `rex field=raw_log "\"(?<method>\\w+) (?<path>\\S+) HTTP"` will take a standard HTTP request line like `GET /api HTTP/1.1` and extract its method and path.

The current implementation of `rex` is prone to expensive table scans that linearly scale based on how much the tables can be filtered on alternative clauses (such as timestamp filters). Table scans themselves are not as performant in Lucene due to Lucene's optimization for search-type queries that emphasizes batch key-value retrieval. Because table scans are themselves not performant, it's also very easy to saturate the SQL plugin's worker queue with many `rex` queries, causing other queries to hang for some time.
## Current State

Rex queries are unable to leverage Lucene indices due to a general lack of automata-based text support. There are [`regexp` Queries](https://docs.opensearch.org/latest/query-dsl/term/regexp/) that function on keyword fields, but they search over terms, not entire strings. This means that `regexp` can't, for example, run patterns over multiple terms such as `/test pattern/`.

The current implementation is reliant on scripts, which mean individually loading and processing all documents over a shard. In addition to making the query itself go slowly, this also takes resources from the shard which can cause other queries to take more time. It also saturates the worker queue.

There are two general things that can be done for this situation:
1. Make `rex` faster.
2. Place limits on how much cluster resources `rex` and other slow operators can take.

Between these, the former is hard because table scans are somewhat unavoidable given what the command needs to do (even writing hand-optimized DSL requires scripting). The latter is hard because there isn't a great way to mark search tasks as associated with specific pools and prioritize accordingly.

I've constructed a custom dataset of 35 rex- and parse-driven queries based on customer usage and common analytics query patterns, and profiled the slowest in the batch on a dataset of 100k records. The timings are as such:

|Query|Time (s)|_source %|JSON parse %|LZ4 %|Regex %|Filter %|
|---|---|---|---|---|---|---|
|10_request_id_tracking|7.68|81%|54%|25%|9%|37%|
|27_rex_stacked_filter|5.68|82%|53%|28%|9%|78%|
|05_ip_geoblocking|5.3|83%|52%|30%|6%|0%|
|18_sampling_rates|2.07|22%|15%|6%|5%|0%|
|31_rex_eval_timechart|1.82|78%|54%|22%|10%|0%|
|06_rest_api_usage|1.76|54%|37%|16%|35%|0%|
|01_nginx_timing|1.52|59%|42%|17%|31%|0%|
|20_comprehensive_parsing|1.48|63%|43%|18%|29%|0%|
|12_timestamp_ordering|1.14|81%|56%|23%|12%|0%|
|04_user_agent_bots|1.12|82%|56%|24%|12%|0%|

The majority of the time for all queries is spent in JSON parsing and LZ4 decompression. These are unavoidable costs with Lucene's document model and inherent to any table-scan process, the only thing we can optimize from the PPL plugin's side is 'regex' and potentially 'filtering'.

So the performance proposals are minor, the stability proposal emphasizes containing the damage of runaway table scans.
## Long-Term Goals

- Achieve near-Painless performance for most PPL queries involving table scans.
- Enable table scans at production scales without hanging the SQL worker.
- Maintain query correctness.
## Proposals

I have 2 performance proposals, 1 stability proposal, and 1 feature proposal, of which we can accept any subset.
#### 1. Pattern Pre-filtering

This is a good optimization when a regex has specific keyword-driven patterns. It's similar in spirit to the approach I shared for joins in [#5093: 200x faster joins with key pre-filtering](https://github.com/opensearch-project/sql/issues/5093).

The idea is that some kinds of patterns (e.g. key-value patterns or json key matching) will have long fixed clauses. For example, this parse command:

```
parse metadata_json '"sampling_rate":(?<samplingrate>[\\d.]+)'
```

This is only applicable to logs that actually have a `sampling_rate` json field. We could turn that condition into a direct text filter that looks for `"sampling_rate":` and only evaluate the script on matching documents.

The savings from this are proportional to how varied the unstructured text types are. If `sampling_rate` is only in ~10% of documents, we can save ~90% of the table scan costs. If this key is in every document, we won't see any savings from doing this. We also won't see much savings in cases of very broad patterns like `.*`. (But if the pattern is in every document, it makes a lot more sense to index it instead of relying on dynamic parsing, so introducing an optimization like this could encourage only using rex for very unstructured content.)

Implementing this shouldn't be too bad -- need to scan over a regex pattern at query compile time and filter out special characters, from what's left over we can detect the longest pattern and throw in a filter if it's >3 characters or so.
#### 2. Merged Scripts

Using a query from a custom parse dataset, `01_nginx_timing`:

```
source=logs
| rex field=raw_log "(?<clientip>\\d+\\.\\d+\\.\\d+\\.\\d+) .+ \"(?<method>\\w+) (?<path>\\S+) HTTP"
| rex field=raw_log "\" (?<status>\\d{3}) (?<bytes>\\d+) .+ rt=(?<reqtime>[\\d.]+)"
| eval reqtime_num = cast(reqtime as double)
| stats avg(reqtime_num) as avg_rt, percentile(reqtime_num, 95) as p95_rt, count() as requests by method, status
| sort - requests
```

This creates a plan with 4 scripts:

```
> cat ~/workplace/utils/locust-sql/logs-parseset/01_nginx_timing.ppl | ppl --explain
= Calcite Plan =
== Logical ==
LogicalSystemLimit(sort0=[$2], dir0=[DESC-nulls-last], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalSort(sort0=[$2], dir0=[DESC-nulls-last])
    LogicalProject(avg_rt=[$2], p95_rt=[$3], requests=[$4], method=[$0], status=[$1])
      LogicalAggregate(group=[{0, 1}], avg_rt=[AVG($2)], p95_rt=[percentile_approx($2, $3, $4)], requests=[COUNT()])
        LogicalProject(method=[REX_EXTRACT($48, '(?<clientip>\d+\.\d+\.\d+\.\d+) .+ "(?<method>\w+) (?<path>\S+) HTTP', 'method')], status=[REX_EXTRACT($48, '" (?<status>\d{3}) (?<bytes>\d+) .+ rt=(?<reqtime>[\d.]+)', 'status')], reqtime_num=[SAFE_CAST(REX_EXTRACT($48, '" (?<status>\d{3}) (?<bytes>\d+) .+ rt=(?<reqtime>[\d.]+)', 'reqtime'))], $f4=[95], $f5=[FLAG(DOUBLE)])
          CalciteLogicalIndexScan(table=[[OpenSearch, logs]])


== Physical ==
CalciteEnumerableTopK(sort0=[$2], dir0=[DESC-nulls-last], fetch=[10000])
  CalciteEnumerableIndexScan(table=[[OpenSearch, logs]], PushDownContext=[[AGGREGATION-rel#2346:LogicalAggregate.NONE.[](input=RelSubset#2345,group={0, 1},avg_rt=AVG($2),p95_rt=percentile_approx($2, $3, $4),requests=COUNT()), PROJECT-[avg_rt, p95_rt, requests, method, status]], OpenSearchRequestBuilder(sourceBuilder={
    "aggregations": {
      "composite_buckets": {
        "aggregations": {
          "avg_rt": {
            "avg": {
              "script": {
                "lang": "opensearch_compounded_script",
                "params": {
                  "DIGESTS": [
                    "raw_log",
                    "\" (?<status>\\d{3}) (?<bytes>\\d+) .+ rt=(?<reqtime>[\\d.]+)",
                    "reqtime"
                  ],
                  "SOURCES": [
                    1,
                    2,
                    2
                  ],
                  "utcTimestamp": 1783347776578257955
                },
                "source": "{\"langType\":\"calcite\",\"script\":\"rO0ABXQEPHsKICAib3AiOiB7CiAgICAibmFtZSI6ICJTQUZFX0NBU1QiLAogICAgImtpbmQiOiAiU0FGRV9DQVNUIiwKICAgICJzeW50YXgiOiAiU1BFQ0lBTCIKICB9LAogICJvcGVyYW5kcyI6IFsKICAgIHsKICAgICAgIm9wIjogewogICAgICAgICJuYW1lIjogIlJFWF9FWFRSQUNUIiwKICAgICAgICAia2luZCI6ICJPVEhFUl9GVU5DVElPTiIsCiAgICAgICAgInN5bnRheCI6ICJGVU5DVElPTiIKICAgICAgfSwKICAgICAgIm9wZXJhbmRzIjogWwogICAgICAgIHsKICAgICAgICAgICJkeW5hbWljUGFyYW0iOiAwLAogICAgICAgICAgInR5cGUiOiB7CiAgICAgICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgICAgIH0KICAgICAgICB9LAogICAgICAgIHsKICAgICAgICAgICJkeW5hbWljUGFyYW0iOiAxLAogICAgICAgICAgInR5cGUiOiB7CiAgICAgICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgICAgIH0KICAgICAgICB9LAogICAgICAgIHsKICAgICAgICAgICJkeW5hbWljUGFyYW0iOiAyLAogICAgICAgICAgInR5cGUiOiB7CiAgICAgICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgICAgIH0KICAgICAgICB9CiAgICAgIF0sCiAgICAgICJjbGFzcyI6ICJvcmcub3BlbnNlYXJjaC5zcWwuZXhwcmVzc2lvbi5mdW5jdGlvbi5Vc2VyRGVmaW5lZEZ1bmN0aW9uQnVpbGRlciQxIiwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgICAicHJlY2lzaW9uIjogMjAwMAogICAgICB9LAogICAgICAiZGV0ZXJtaW5pc3RpYyI6IHRydWUsCiAgICAgICJkeW5hbWljIjogZmFsc2UKICAgIH0KICBdLAogICJ0eXBlIjogewogICAgInR5cGUiOiAiRE9VQkxFIiwKICAgICJudWxsYWJsZSI6IHRydWUKICB9Cn0=\"}"
              }
            }
          },
          "p95_rt": {
            "percentiles": {
              "keyed": true,
              "percents": [
                95.0
              ],
              "script": {
                "lang": "opensearch_compounded_script",
                "params": {
                  "DIGESTS": [
                    "raw_log",
                    "\" (?<status>\\d{3}) (?<bytes>\\d+) .+ rt=(?<reqtime>[\\d.]+)",
                    "reqtime"
                  ],
                  "SOURCES": [
                    1,
                    2,
                    2
                  ],
                  "utcTimestamp": 1783347776578257955
                },
                "source": "{\"langType\":\"calcite\",\"script\":\"rO0ABXQEPHsKICAib3AiOiB7CiAgICAibmFtZSI6ICJTQUZFX0NBU1QiLAogICAgImtpbmQiOiAiU0FGRV9DQVNUIiwKICAgICJzeW50YXgiOiAiU1BFQ0lBTCIKICB9LAogICJvcGVyYW5kcyI6IFsKICAgIHsKICAgICAgIm9wIjogewogICAgICAgICJuYW1lIjogIlJFWF9FWFRSQUNUIiwKICAgICAgICAia2luZCI6ICJPVEhFUl9GVU5DVElPTiIsCiAgICAgICAgInN5bnRheCI6ICJGVU5DVElPTiIKICAgICAgfSwKICAgICAgIm9wZXJhbmRzIjogWwogICAgICAgIHsKICAgICAgICAgICJkeW5hbWljUGFyYW0iOiAwLAogICAgICAgICAgInR5cGUiOiB7CiAgICAgICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgICAgIH0KICAgICAgICB9LAogICAgICAgIHsKICAgICAgICAgICJkeW5hbWljUGFyYW0iOiAxLAogICAgICAgICAgInR5cGUiOiB7CiAgICAgICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgICAgIH0KICAgICAgICB9LAogICAgICAgIHsKICAgICAgICAgICJkeW5hbWljUGFyYW0iOiAyLAogICAgICAgICAgInR5cGUiOiB7CiAgICAgICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgICAgIH0KICAgICAgICB9CiAgICAgIF0sCiAgICAgICJjbGFzcyI6ICJvcmcub3BlbnNlYXJjaC5zcWwuZXhwcmVzc2lvbi5mdW5jdGlvbi5Vc2VyRGVmaW5lZEZ1bmN0aW9uQnVpbGRlciQxIiwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgICAicHJlY2lzaW9uIjogMjAwMAogICAgICB9LAogICAgICAiZGV0ZXJtaW5pc3RpYyI6IHRydWUsCiAgICAgICJkeW5hbWljIjogZmFsc2UKICAgIH0KICBdLAogICJ0eXBlIjogewogICAgInR5cGUiOiAiRE9VQkxFIiwKICAgICJudWxsYWJsZSI6IHRydWUKICB9Cn0=\"}"
              },
              "tdigest": {
                "compression": 100.0
              }
            }
          }
        },
        "composite": {
          "size": 10000,
          "sources": [
            {
              "method": {
                "terms": {
                  "missing_bucket": true,
                  "missing_order": "first",
                  "order": "asc",
                  "script": {
                    "lang": "opensearch_compounded_script",
                    "params": {
                      "DIGESTS": [
                        "raw_log",
                        "(?<clientip>\\d+\\.\\d+\\.\\d+\\.\\d+) .+ \"(?<method>\\w+) (?<path>\\S+) HTTP",
                        "method"
                      ],
                      "SOURCES": [
                        1,
                        2,
                        2
                      ],
                      "utcTimestamp": 1783347776578257955
                    },
                    "source": "{\"langType\":\"calcite\",\"script\":\"rO0ABXQC63sKICAib3AiOiB7CiAgICAibmFtZSI6ICJSRVhfRVhUUkFDVCIsCiAgICAia2luZCI6ICJPVEhFUl9GVU5DVElPTiIsCiAgICAic3ludGF4IjogIkZVTkNUSU9OIgogIH0sCiAgIm9wZXJhbmRzIjogWwogICAgewogICAgICAiZHluYW1pY1BhcmFtIjogMCwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgfQogICAgfSwKICAgIHsKICAgICAgImR5bmFtaWNQYXJhbSI6IDEsCiAgICAgICJ0eXBlIjogewogICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICJudWxsYWJsZSI6IHRydWUsCiAgICAgICAgInByZWNpc2lvbiI6IC0xCiAgICAgIH0KICAgIH0sCiAgICB7CiAgICAgICJkeW5hbWljUGFyYW0iOiAyLAogICAgICAidHlwZSI6IHsKICAgICAgICAidHlwZSI6ICJWQVJDSEFSIiwKICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICJwcmVjaXNpb24iOiAtMQogICAgICB9CiAgICB9CiAgXSwKICAiY2xhc3MiOiAib3JnLm9wZW5zZWFyY2guc3FsLmV4cHJlc3Npb24uZnVuY3Rpb24uVXNlckRlZmluZWRGdW5jdGlvbkJ1aWxkZXIkMSIsCiAgInR5cGUiOiB7CiAgICAidHlwZSI6ICJWQVJDSEFSIiwKICAgICJudWxsYWJsZSI6IHRydWUsCiAgICAicHJlY2lzaW9uIjogMjAwMAogIH0sCiAgImRldGVybWluaXN0aWMiOiB0cnVlLAogICJkeW5hbWljIjogZmFsc2UKfQ==\"}"
                  }
                }
              }
            },
            {
              "status": {
                "terms": {
                  "missing_bucket": true,
                  "missing_order": "first",
                  "order": "asc",
                  "script": {
                    "lang": "opensearch_compounded_script",
                    "params": {
                      "DIGESTS": [
                        "raw_log",
                        "\" (?<status>\\d{3}) (?<bytes>\\d+) .+ rt=(?<reqtime>[\\d.]+)",
                        "status"
                      ],
                      "SOURCES": [
                        1,
                        2,
                        2
                      ],
                      "utcTimestamp": 1783347776578257955
                    },
                    "source": "{\"langType\":\"calcite\",\"script\":\"rO0ABXQC63sKICAib3AiOiB7CiAgICAibmFtZSI6ICJSRVhfRVhUUkFDVCIsCiAgICAia2luZCI6ICJPVEhFUl9GVU5DVElPTiIsCiAgICAic3ludGF4IjogIkZVTkNUSU9OIgogIH0sCiAgIm9wZXJhbmRzIjogWwogICAgewogICAgICAiZHluYW1pY1BhcmFtIjogMCwKICAgICAgInR5cGUiOiB7CiAgICAgICAgInR5cGUiOiAiVkFSQ0hBUiIsCiAgICAgICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICAgICAicHJlY2lzaW9uIjogLTEKICAgICAgfQogICAgfSwKICAgIHsKICAgICAgImR5bmFtaWNQYXJhbSI6IDEsCiAgICAgICJ0eXBlIjogewogICAgICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgICAgICJudWxsYWJsZSI6IHRydWUsCiAgICAgICAgInByZWNpc2lvbiI6IC0xCiAgICAgIH0KICAgIH0sCiAgICB7CiAgICAgICJkeW5hbWljUGFyYW0iOiAyLAogICAgICAidHlwZSI6IHsKICAgICAgICAidHlwZSI6ICJWQVJDSEFSIiwKICAgICAgICAibnVsbGFibGUiOiB0cnVlLAogICAgICAgICJwcmVjaXNpb24iOiAtMQogICAgICB9CiAgICB9CiAgXSwKICAiY2xhc3MiOiAib3JnLm9wZW5zZWFyY2guc3FsLmV4cHJlc3Npb24uZnVuY3Rpb24uVXNlckRlZmluZWRGdW5jdGlvbkJ1aWxkZXIkMSIsCiAgInR5cGUiOiB7CiAgICAidHlwZSI6ICJWQVJDSEFSIiwKICAgICJudWxsYWJsZSI6IHRydWUsCiAgICAicHJlY2lzaW9uIjogMjAwMAogIH0sCiAgImRldGVybWluaXN0aWMiOiB0cnVlLAogICJkeW5hbWljIjogZmFsc2UKfQ==\"}"
                  }
                }
              }
            }
          ]
        }
      }
    },
    "from": 0,
    "size": 0,
    "timeout": "1m"
  }, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```

Respectively, the scripts:
- Extract the method field
- Extract the status field
- Aggregate request_time for the average
- Aggregate request_time for p95 (identical to previous)

We could consider merging the scripts to facilitate more work sharing. I'm mentioning it because the optimization looks tempting and I spent a bit of time researching/testing it.

From the above profiles, we know the majority of the time is spent in LZ4 decompression from the Lucene index, and JSON parsing of Lucene docs. This is cached between the scripts. Once that's done, a new script context is created for each invocation to compile and execute the regex, which involves a minor amount of allocation, but the regex compilation is also LRU-cached. The vast majority of all script execution here is just evaluating the regex.

In this special case, finding a way to merge all the scripts might save ~1/4 regex executions due to the duplicated request time aggregation (which only applies to queries that have this type of extract-then-multi-aggregate setup). In the general case, I roughly estimate the script context construction savings for n>2 scripts to be 5-10% of the total script execution time.
#### 1. Dedicated slow-query thread pool

We can leverage Calcite's cost model to predict queries that are likely to be slow, and put slow queries in a separated IO thread pool so they share resources better with fast queries. This still has some limitations in that e.g. ultimately all queries will share the same `search` thread pool, but it avoids hanging the cluster for normal cases when there are lots of `rex` queries in progress.

We can also allow users to configure separate timeouts for slow queries and fast queries, if they want to ensure certain request/second lower bounds.
#### 1. PPL-native reindexing

We can revisit implementing an indexing command in PPL that's able to dynamically index specific extraction patterns, i.e. a wrapper for reindex that works with rex-like syntax. This can lower some complexity for indexing pipelines but generally will still have the same costs (and consistency issues since new data is presumably also coming in between invocations of the index command).
## Approach

Skipping over the specifics of each, roughly:

- Fixed pattern extraction should be easy, just a string scan with a lightweight lexer that we turn into a `match_phrase` or similar. I've already proof-of-concepted the core extraction so I just need to wire it to DSL. ~2 days including testing.
- Script merging is harder -- in cases where it works we can just glue scripts together, but a lot of DSL contexts expect scripts to emit one value (e.g. there isn't a clear way to combine filter and aggregation scripts, and aggregation ValueSource contexts require one scalar value per document call). We'd need to introduce cross-script memoization with regex extraction on specific fields. More complex, probably ~2 weeks.
- The slow query thread pool should be easy, I already did something similar in #4345. 1 or 2 days.
- The reindex feature is significantly more complex, and if customers don't want to put consistently accessed data directly in the indexing pipeline, it's unlikely they want to do it but worse at query time.
## Alternatives

N/A, RFC itself has lots of alternatives, open to other suggestions.
## Implementation Discussion

N/A

Incidentally, the analytics engine should save a lot of table scan cost.

Swiddis:
### Description
Following discussion on #5608, this PR implements a dedicated thread pool for slow queries. This helps keep the cluster responsive in cases of specific expensive queries being run, as the main worker thread pool will still be available to process fast queries. This comes with a `plugins.sql.slow_worker_pool.enabled` setting (default true).

The pool exists as a separate resource that the worker can hand off to. So the normal fast query path is unchanged, we only hand off in known special cases. This also means we only pay rescheduling overhead for queries that are already slow.

I've tested this by saturating a cluster with >20sec queries (until all slow threads are active and requests are queued), and then verified that normal cheap queries are still responsive even if slow queries are at capacity.

**Guide for reviewers:**
- The detection rules live in `ScriptDetector.java`. We primary look for specific expensive UDFs (rex, parse), window functions, joins, or any script pushdown contexts.
  - The biggest class of false-negatives is aggregations with high cardinality, we can't determine from the query plan alone that an aggregation field is very large. We'd need some sort of statistics tracking on the index. I don't want to put all aggregations in the slow pool since they're very common for dashboarding and many aggregations are low-cardinality.
- Execution dispatch is in `ThreadPoolExecutionDispatcher`, if we hit the `ScriptDetector` check we do a handoff involving a lot of thread context propagation and cleanup. This is the riskiest part of the code, I've been carefully checking that the context propagation is working correctly and that we handle things like auth context (field level security) and cancellation.

### Related Issues
Resolves #5608 

### Check List


plugin/build.gradle
```diff
@@ -305,8 +305,27 @@ def getJobSchedulerPlugin() {
...
 
+def getWorkloadManagementPlugin() {
+    provider(new Callable<RegularFile>() {
+        @Override
+        RegularFile call() throws Exception {
+            return new RegularFile() {
+                @Override
+                File getAsFile() {
+                    // ponytail: assumes ~/workplace/opensearch is built
+                    def wlmPath = System.getProperty('user.home') + '/workplace/opensearch/plugins/workload-management/build/distributions'
```
RyanL1997: do we actually need this?

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
```diff
@@ -516,8 +516,6 @@ private static void enrichErrorsForSpecialCases(ErrorReport.Builder report, SQLE
     public static PreparedStatement run(CalcitePlanContext context, RelNode rel) {
       ProfileMetric optimizeTime = QueryProfiling.current().getOrCreateMetric(OPTIMIZE);
       long startTime = System.nanoTime();
```
penghuo, line 518: optimize time collector should move?

core/src/main/java/org/opensearch/sql/monitor/profile/QueryProfile.java
```diff
@@ -24,14 +24,17 @@ public final class QueryProfile {
   /** Execution-engine-specific plan profile: a {@link PlanNode} tree, or a pre-rendered object. */
   private final Object plan;
 
+  @SerializedName("thread_pool")
+  private final String threadPool;
+
```
penghuo, line 29: profile results will include threadpool info? if yes, update doc. https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/interfaces/endpoint.md#profile-experimental

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
```diff
@@ -326,7 +327,8 @@ public void execute(
       RelNode rel, CalcitePlanContext context, ResponseListener<QueryResponse> listener) {
     client.schedule(
         () -> {
-          try (PreparedStatement statement = OpenSearchRelRunners.run(context, rel)) {
+          try (PreparedStatement statement =
+              OpenSearchRelRunners.run(context, CalciteToolsHelper.optimize(rel, context))) {
```
penghuo: Optimize already been called in QueryService. Why call optimize again?

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/ThreadPoolExecutionDispatcher.java
```diff
@@ -0,0 +1,122 @@
...
+                RelMetadataQueryBase.THREAD_PROVIDERS.set(metadataProvider);
+              }
+              if (currentTime >= 0) {
+                hookHandle =
+                    Hook.CURRENT_TIME.addThread(
+                        (Consumer<org.apache.calcite.util.Holder<Long>>) h -> h.set(currentTime));
+              }
+              CalcitePlanContext.stripNullColumns.set(stripNullCols);
+              CalcitePlanContext.timewrapUnitName.set(twUnitName);
+              CalcitePlanContext.timewrapSeries.set(twSeries);
```
penghuo: why L87-L94 change required? seems not releated to threadpool change.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/ThreadPoolExecutionDispatcher.java
```diff
@@ -0,0 +1,122 @@
...
+  public void dispatchTask(RelNode plan, CalcitePlanContext context, Runnable task) {
+    dispatchInternal(plan, context, task, null);
+  }
+
+  private void dispatchInternal(
+      RelNode optimizedPlan,
+      CalcitePlanContext context,
+      Runnable task,
+      @Nullable ResponseListener<?> failureListener) {
+    if (isComplexPoolEnabled() && ScriptDetector.hasScripts(optimizedPlan)) {
```
penghuo, line 67: When query schedule in complex-worker thread, the complex query timeout control OpenSearchQueryManager does not take effect.

core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -207,9 +237,15 @@ private void executeCalcitePlan(
...
+      // Optimize before dispatch so the dispatcher's ScriptDetector
+      // sees the post-optimization plan for accurate routing.
+      RelNode optimizedPlan = CalciteToolsHelper.optimize(calcitePlan, context);
+
+      // Wrap execution with EXECUTING stage tracking — dispatch via
+      // ExecutionDispatcher which may route to a complex worker pool
       StageErrorHandler.executeStageVoid(
           QueryProcessingStage.EXECUTING,
-          () -> executionEngine.execute(calcitePlan, context, listener),
+          () -> executionDispatcher.dispatch(optimizedPlan, context, listener, executionEngine),
```
penghuo, line 263: Calcite fallback exception will not catched if it is complex query. The reasons is dispatch() returns success (the schedule call didn't throw), so executeStageVoid records the EXECUTING stage as successful. The real error surfaces later, on the complex thread, and is delivered raw via listener.onFailure(e). It never passes through executeStageVoid's wrapping (no EXECUTING stage/location on the ErrorReport), never reaches the overflow-translation catch, and never triggers V2 fallback.

core/src/main/java/org/opensearch/sql/monitor/profile/DefaultProfileContext.java
```diff
@@ -63,7 +64,8 @@ public synchronized QueryProfile finish() {
     double totalMillis = ProfileUtils.roundToMillis(endNanos - startNanos);
     Object planSnapshot =
         enginePlan != null ? enginePlan : (planRoot == null ? null : planRoot.snapshot());
-    profile = new QueryProfile(totalMillis, snapshot, planSnapshot);
+    String threadPool = CalcitePlanContext.executionPool.get();
```
penghuo, line 67: finish() is invoked from the response formatter. QueryProfile.thread_pool is read on the wrong thread → always null/wrong for complex-pool queries

penghuo (changes requested):

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/ThreadPoolExecutionDispatcher.java
Swiddis: They're thread-local state associated with the calcite context that needs to be copied or they get dropped. In particular the `current_time` hook bits are related to integ test failures with time queries involving `NOW()`, and the others are just copied for following the same pattern.

To make the intent of this clearer I moved all of this copy logic into `snapshot` methods in CalcitePlanContext. Should also make it so if the thread context is modified in the future, this branch is less likely to be missed.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/ThreadPoolExecutionDispatcher.java
Swiddis: It's applying but slowly, I poked more and it's because script-based complex queries can spend very long delays in DSL (compared to e.g. joins where individual DSL pages are fast). I added some polling so the PPL thread interrupts faster in these cases. Notably we still leave zombie search tasks behind, but this is a preexisting issue and they're still usually much faster than the overall query (involving multiple pages -- in testing I got no sub-searches beyond 10 seconds or so).

I can follow up doing more aggressive cancelation that propagates to the search task if we need it but I believe the polling impl is enough for this PR.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
Swiddis: ah yeah, my bad -- I was moving optimize out of 'execute' since we need the optimized plan to make complex branch decisions, and did a ctrl+f to find old execute/run call sites and make sure they called optimize.

core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
Swiddis: The majority of the optimization time is in the volcano planner done below this call, the HepPlanner optimization is sub-millisecond.

I think it makes sense to keep this here and count HepPlanner optimization as part of analysis. I moved the analysis timing gate to after we call optimization in that path. (This means we don't technically have the "fully optimized plan" for the routing decision but empirically I don't believe we need it, HEP is enough for our current heuristics. We'd need to move this if we want to incorporate fancier cost metrics in the decision later.)

So from here I think analysis is more clearly defined at the end as "everything necessary for the query routing decision including lightweight pre-optimization," while optimization does more heavy-duty cost analysis and final script codegen.

core/src/main/java/org/opensearch/sql/executor/QueryService.java
Swiddis: yeah I didn't think about that... It's hard to add handling here because the fallback calls to another path back in the simple pool, we'd I think need some sort of callback that sends the request back over, or reimplement that side of the logic in the complex pool.

I think it's acceptable to not handle this case because we currently have no logic to fallback after execution, those decisions are all made at planning time before the thread handoff. And I don't expect we'll add any new execution-time fallback logic in the future, seems like these sites should be trending toward 0 over time.

core/src/main/java/org/opensearch/sql/monitor/profile/DefaultProfileContext.java
Swiddis: This execution pool context is set with the thread state copy over near https://github.com/opensearch-project/sql/pull/5628/changes#r3616809250, empirically this works. I added an IT for complex profiling correctly marking its thread pool.

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
penghuo: I did not get u. We should run optimize 2 times?

Swiddis: I fixed

penghuo (approved):

noCharger:
Since both sql-worker and sql-complex-worker default to allocatedProcessors, a saturated complex pool can add another N CPU-heavy executions on the same JVM. What guarantees that this doesn't simply move starvation from the SQL worker queue to CPU/heap/search-thread contention? Should the complex pool instead have a smaller configurable concurrency limit?

PR #5631, opened 2026-07-16
Surface PIT-context exhaustion with an actionable error message

ahkcs:
### Description

The Calcite engine opens an internal Point-in-Time (PIT) context to page over a query it cannot push down to OpenSearch. A common case: a `stats`/aggregation that groups by a text field with no `keyword` sub-field — aggregations only push down on keyword/numeric/date fields, so the engine falls back to scanning the docs, which opens a PIT.

A PIT allocates **one reader context per shard**, and the node caps the total at `search.max_open_pit_context` (default 300). So a single such query over a many-shard index can exhaust the budget on its own; concurrent load makes it more likely.

Previously that failure surfaced to the user as an opaque internal message:

```json
{
  "error": {
    "reason": "java.sql.SQLException: exception while executing query: Error occurred while creating PIT for internal plugin operation",
    "type": "RuntimeException",
    "code": "UNKNOWN"
  },
  "status": 500
}
```

— no hint of the cause or the remedy.

### What it does

Detect the PIT-context-limit rejection in the execute-time cause chain (`OpenSearchExecutionEngine.execute`) and rethrow it as a `PointInTimeLimitExceededException` wrapped in an `ErrorReport`. The user now gets:

```json
{
  "error": {
    "reason": "Too many open Point-In-Time (PIT) contexts on this node.",
    "details": "This query opened a Point-In-Time (PIT) context on each shard and reached the limit set by [search.max_open_pit_context]. Increase that setting, or optimize the query to reduce the number of contexts it needs.",
    "code": "RESOURCE_LIMIT_EXCEEDED",
    "type": "PointInTimeLimitExceededException"
  },
  "status": 500
}
```

- **`reason`** is a short title; the explanation and the two remedies (raise the setting, or optimize the query) live in **`details`**.
- **`code`** is `RESOURCE_LIMIT_EXCEEDED` for programmatic handling.

**Why match on the message, not the exception class:** `OpenSearchRejectedExecutionException` is also raised for scroll-context and thread-pool rejections, so the class alone can't identify PIT-context exhaustion. The `too many Point In Time contexts` substring (authored by OpenSearch `SearchService`) is the reliable discriminator.

**Why walk the whole cause chain:** at execute time the rejection surfaces several layers deep (`SQLException` → `RuntimeException` → `ExecutionException` "all shards failed" → per-shard `OpenSearchRejectedExecutionException`), so detection scans every cause. The walk guards against self-referential cause loops.

**Why the Calcite path specifically:** Calcite is the default engine, and its execution errors are wrapped in an `ErrorReport` that the error formatter (`ErrorMessageFactory`) short-circuits before it reaches the OpenSearch-exception formatter — so the fix belongs where the `ErrorReport` is built, not in the shared formatter.

### Scope note

This is a diagnostics-only change. It does not alter PIT routing, the create/delete PIT logic, pushdown behavior, or any other exception handling — every non-exhaustion failure still falls through to the existing path unchanged. The only difference is that this one common, previously-opaque failure now produces a message a user can act on.

### Testing

**Unit — `OpenSearchExecutionEngineTest`:**
- PIT-context limit detected when the marker is nested several layers deep in the cause chain.
- An unrelated "all shards failed" failure is not misclassified.
- Detection terminates on a self-referential cause (no infinite loop).

**Manual end-to-end:** built the plugin, created a 5-shard index with a text-only field (no `keyword`), set `search.max_open_pit_context` to `0`, and ran `source=<index> | stats count() by <text-field>` on the Calcite path. The response is the actionable error above (confirmed `reason`/`details`/`code`/`type`); a pushdown-able aggregation on a keyword field still returns 200 with no PIT.

`:opensearch:test` for the touched class passes; `spotlessJavaCheck` is clean on `:opensearch`.

### Check List
- [ ] New functionality has been documented. (error-message/diagnostics change; no user-facing doc)


opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
```diff
@@ -336,11 +338,48 @@ public void execute(
...
+              // explanation and remedies in details.
+              PointInTimeLimitExceededException pitException =
+                  new PointInTimeLimitExceededException(
+                      "Too many open Point-In-Time (PIT) contexts on this node.", e);
+              throw ErrorReport.wrap(pitException)
+                  .code(ErrorCode.RESOURCE_LIMIT_EXCEEDED)
+                  .details(
+                      "This query opened a Point-In-Time (PIT) context on each shard and reached"
+                          + " the limit set by [search.max_open_pit_context]. Increase that"
+                          + " setting, or optimize the query to reduce the number of contexts it"
```
penghuo: Remove "or optimize the query to reduce the number of contexts it needs"

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/PointInTimeLimitExceededException.java
```diff
@@ -0,0 +1,20 @@
...
+package org.opensearch.sql.opensearch.executor;
+
+/**
+ * Raised when the engine cannot open the Point-In-Time (PIT) context it needs to run a query
+ * because the node has hit {@code search.max_open_pit_context}. The engine opens a PIT (one reader
+ * context per shard) to page over a query it cannot push down -- e.g. a {@code stats} that groups
+ * by a text field with no {@code keyword} sub-field -- so a single query over a many-shard index
+ * can exhaust the per-node budget on its own. Its message is the customer-facing {@code reason}.
+ */
+public class PointInTimeLimitExceededException extends RuntimeException {
```
penghuo: [non blocker] Is it best practice to define PointInTimeLimitExceededException for each error message type?

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/OpenSearchExecutionEngine.java
ahkcs: updated

opensearch/src/main/java/org/opensearch/sql/opensearch/executor/PointInTimeLimitExceededException.java
ahkcs: response type changes from `PointInTimeLimitExceededException` → `ResourceLimitExceededException` to be more general

penghuo (approved):

PR #5633, opened 2026-07-16
Decouple Calcite PPL planning from ExprType

penghuo:
### Summary

The Calcite planning surface for PPL currently round-trips every operand through the v2 `ExprType` enum: type checkers, coercion, UDF operand metadata, and cast targets all bounce between `RelDataType` and `ExprType`. This makes the planning layer coupled to the v2 type system and blocks future work that only needs `RelDataType`.

This PR decouples the planning surface: 
* `PPLTypeChecker` exposes signatures as `List<List<RelDataType>>`
* `PPLOperandTypes` publishes `RelDataType` constants 
* UDF metadata / IP / temporal functions branch on `RelDataType` directly. 

Out of scope
* UDTs (`ExprDateType`, `ExprTimeType`, `ExprTimeStampType`, `ExprIPType`, `ExprBinaryType`) still flow through logical plans unchanged; only the planning-time abstraction is affected. 
* `CoercionUtils` keeps main-branch behavior — only the entry to `castArguments` converts `RelDataType` back to `ExprType` at the boundary.

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
@@ -572,10 +572,10 @@ public interface PPLTypeChecker {
   }
 
   /**
-   * Compares two RelDataTypes for signature matching. Two UDTs match if they share the same
-   * {@link ExprUDT} tag — comparing {@code getClass()} is unsafe because addCharsetAndCollation
-   * collapses ExprDateType/ExprTimeType/ExprTimeStampType/ExprBinaryType down to ExprSqlType, so
-   * different UDTs would appear equal. Plain types match by SqlTypeName.
+   * Compares two RelDataTypes for signature matching. Two UDTs match if they share the same {@link
+   * ExprUDT} tag — comparing {@code getClass()} is unsafe because addCharsetAndCollation collapses
+   * ExprDateType/ExprTimeType/ExprTimeStampType/ExprBinaryType down to ExprSqlType, so different
+   * UDTs would appear equal. Plain types match by SqlTypeName.
    */
   private static boolean typesMatch(RelDataType expected, RelDataType actual) {
     if (expected instanceof AbstractExprRelDataType<?> expUdt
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
```diff
@@ -299,18 +297,38 @@ private static boolean isComparable(RelDataType type1, RelDataType type2) {
...
       }
 
+      // Plain types are comparable when their SqlTypeFamily matches (CHAR vs VARCHAR, INTEGER vs
+      // BIGINT, BOOLEAN vs BOOLEAN, etc.). UDTs are excluded since they share the underlying
+      // VARCHAR family but should not be transparently comparable to plain strings here.
+      if (!(type1 instanceof AbstractExprRelDataType<?>)
+          && !(type2 instanceof AbstractExprRelDataType<?>)) {
+        SqlTypeFamily family1 = type1.getSqlTypeName().getFamily();
+        SqlTypeFamily family2 = type2.getSqlTypeName().getFamily();
+        if (family1 != null && family1 == family2) {
```
RyanL1997: [non blocking] 

I was scanning the code and I think the following 2 edge pairs are become more restricter of the code change: 

> 1. plain BINARY/VARBINARY vs the EXPR_BINARY UDT — the !(instanceof AbstractExprRelDataType) guard skips this block, so it's no longer comparable (old code matched them via BINARY == BINARY).  
                                    
> 2. day-time interval vs year-month interval — family1 == family2 is false since Calcite splits them into INTERVAL_DAY_TIME / INTERVAL_YEAR_MONTH, where the old code collapsed both to INTERVAL.

no unit test on `isComparable` maybe we can add some tests on these later.

RyanL1997 (approved):
[nit] I was scanning the code - maybe we can add more unit test coverage on

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
dai-chen: yeah, I don't see UT for `PPLTypeChecker`. Probably we need to confirm the behavior is the same as previous ExprType.shouldCast() or changed as we expect.

dai-chen (approved):
Thanks for the changes!

core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
penghuo: Add UT and revert the code change in isComparable() method. As it only use ExprType in private method.

PR #5636, opened 2026-07-19
[Feature] Integrate SQL/PPL with query-insights plugin

### LINKED ISSUE #5677: [FEATURE] Integrate SQL/PPL with query-insights plugin for query source tracking

The query-insights plugin captures all search queries as generic DSL queries. When SQL/PPL queries are executed through the _plugins/_sql or _plugins/_ppl endpoints, they generate one or more underlying DSL searches, but there is no way to trace these back to the original SQL/PPL statement. Operators cannot identify which SQL/PPL queries are consuming the most resources, causing latency spikes, or need optimization. The Top N Queries view shows only raw DSL bodies, requiring manual correlation.

Propagate SQL/PPL query metadata to the query-insights plugin via thread context headers so that:

- Each DSL search generated by SQL/PPL is tagged with x-query-source (sql/ppl), x-original-query (the SQL/PPL text), and x-query-execution-id (a UUID correlating all DSL searches from one SQL/PPL execution)
- SQL/PPL processing phase timings (parse, analyze, plan) are captured via x-query-phases header and exposed alongside the query record
- Headers are registered as task headers via getTaskHeaders() so they propagate through the transport layer to the query-insights listener
- The Calcite execution path (used by PPL) includes phase tracking alongside the legacy path (used by SQL)

Pattern-matching the DSL source in the query-insights listener to detect SQL/PPL-generated queries — rejected as fragile, unreliable, and unable to recover the original SQL/PPL text.
Building a separate SQL-specific query tracking system — rejected as duplicative; the existing query-insights infrastructure already provides top-N ranking, storage, and dashboards.
Adding metadata at the REST layer only — rejected because it wouldn't propagate to child search tasks across nodes in a distributed cluster.

Implementation PR: https://github.com/opensearch-project/sql/pull/5636

Companion PRs in other repos:

- Query Insights plugin (aggregation, storage): https://github.com/opensearch-project/query-insights/pull/647
- Query Insights Dashboards (UI): https://github.com/opensearch-project/query-insights-dashboards/pull/577

KishoreKicha14:
### Description
Propagate SQL/PPL query metadata to OpenSearch's query-insights plugin so that DSL queries generated by the SQL engine are identifiable and traceable back to their originating SQL/PPL statement.

Changes:
- Add thread context headers (x-query-source, x-original-query, x-query-execution-id, x-query-phases) set before DSL execution
- Register headers via getTaskHeaders() so TaskManager copies them into SearchTask for query-insights to read
- Add QueryPhaseTracker to instrument parse/analyze/plan phases with wall-clock time, CPU time, and memory allocation per phase
- Execution ID links multiple DSL queries from a single SQL/PPL execution (e.g., JOINs, pagination)

### Related Issues
Resolves #5677 

### Check List
- [X] New functionality has been documented.
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).


core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -279,12 +285,15 @@ public void executeWithLegacy(
       ResponseListener<ExecutionEngine.QueryResponse> listener,
       Optional<Throwable> calciteFailure) {
     try {
-      executePlan(analyze(plan, queryType), PlanContext.emptyPlanContext(), listener);
+      org.opensearch.sql.common.utils.QueryPhaseTracker tracker =
```
ansjcy: please do proper import, don't use org.opensearch.sql.common.utils.QueryPhaseTracker..

core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -279,12 +285,15 @@ public void executeWithLegacy(
...
+      tracker.beginPhase("analyze");
+      LogicalPlan analyzed = analyze(plan, queryType);
+      tracker.beginPhase("plan");
+      executePlan(analyzed, PlanContext.emptyPlanContext(), listener);
     } catch (Exception e) {
       if (calciteFailure.isPresent()) {
-        // This happens if Calcite fell back to V2 due to some issue, and then V2 also failed.
-        // Prefer the Calcite error.
-        // https://github.com/opensearch-project/sql/issues/5060
+        // Calcite fell back to V2 which also failed — prefer Calcite error (#5060)
```
ansjcy: nit: why is this needed?

ansjcy (commented):
I need to take a closer look into the core logistics but please make sure you format this PR before submitting.

core/src/main/java/org/opensearch/sql/executor/QueryService.java
```diff
@@ -201,6 +202,8 @@ public void executeWithCalcite(
     CalcitePlanContext.run(
         () -> {
           try {
+            QueryPhaseTracker tracker = QueryPhaseTracker.startOrRestore();
+            tracker.beginPhase("analyze");
```
Swiddis: Is it better to add these phase markers here instead of part of `executeStage`?

Can possibly rename StageErrorHandler if it's not broad enough to cover all stage-covering monitoring. Same thing with possibly merging the whole queryphasetracker into what we're already using to define these phases.

dzane17:
Instead of injecting four new request headers, can you add the parent id to the DSL search tasks created by sql/ppl? Then every normal DSL search task will have a reference back to its originating sql/ppl query. Query Insights can simply look up whatever info it needs, like the original request body.

The gap right now is that not all DSL search tasks carry the parent PPL/SQL id. PPL v2 was instrumented recently in #5628 (`applyParentTask` in `OpenSearchNodeClient`). If the same were done for the v3 PPL and SQL paths (SQL also needs a coordinator task created first), it would avoid the new headers, `getTaskHeaders()` wiring, thread-context propagation, and raw query text on the wire that the header approach requires.

semantic-review/2025-07-15-162000-pr-sql-query-insights.md
Swiddis, line 1: did you mean to put this ai review folder in the project root

plugin/src/main/java/org/opensearch/sql/plugin/transport/TransportPPLQueryAction.java
```diff
@@ -179,6 +181,25 @@ protected void doExecute(
 
     // in order to use PPL service, we need to convert TransportPPLQueryRequest to PPLQueryRequest
     PPLQueryRequest transformedRequest = transportRequest.toPPLQueryRequest();
+
+    // Tag the thread context so query-insights can identify this as a PPL-derived query.
+    org.opensearch.common.util.concurrent.ThreadContext threadContext =
```
Swiddis: are these headers correctly getting propagated after a complex thread pool handoff?

Swiddis (commented):
preliminary review, haven't traced through the low-level details of thread marking and such / why it matters whether or not we're on a sun jvm

Swiddis:
> Instead of injecting four new request headers, can you add the parent id to the DSL search tasks created by sql/ppl? Then every normal DSL search task will have a reference back to its originating sql/ppl query. Query Insights can simply look up whatever info it needs, like the original request body.
> 
> The gap right now is that not all DSL search tasks carry the parent PPL/SQL id. PPL v2 was instrumented recently in #5628 (`applyParentTask` in `OpenSearchNodeClient`). If the same were done for the v3 PPL and SQL paths (SQL also needs a coordinator task created first), it would avoid the new headers, `getTaskHeaders()` wiring, thread-context propagation, and raw query text on the wire that the header approach requires.

That pr is mostly focused on v3, not sure where v2 is from. But other than that, I like this proposal, we should be roughly aligning to the existing stage tracking and thread propagation we already have as part of complex/background thread handoff. Particularly because historically thread propagation is a pain to modify and guard on edge cases, I'd be very in favor of a single component that owns "create an instrumented thread context that runs $LAMBDA and emits the right metrics to the right places."

dzane17:
My bad, then the parent task approach is even simpler. @penghuo @dai-chen do you have any ideas or concerns?

penghuo:
@KishoreKicha14 

- Does both live_queries and TopN queries will be supported in query insigts?
- I suggest seperete PR. One for PPL, one for SQL.

KishoreKicha14:
@penghuo Yes both live and TopN queries will be support SQL and PPL queries. I will update this into 2 separate PRs (SQL and PPL).

penghuo:
> @penghuo Yes both live and TopN queries will be support SQL and PPL queries. I will update this into 2 separate PRs (SQL and PPL).
- Could u update a PR description to explain how PPL works in live_queries and topN queries.
- Could u include a screenshot to demo how PPL looks like in Query insights?

PR #5637, opened 2026-07-20
Fix PPL foreach JSON array type coercion

songkant-aws:
### Description

This is a follow-up to #5613 for JSON arrays stored in text fields.

For field-backed JSON, element values are not visible while the query is being planned. Numeric use of `<<ITEM>>` therefore selects a concrete numeric element type from the expression context, but `ForeachJsonArrayFunctionImpl` previously assumed every parsed value was already a `Number`. A JSON string in that path caused a `ClassCastException` and an HTTP 500 response.

This change:

- converts numeric JSON elements with Calcite's runtime conversion helpers, including numeric strings;
- returns `null` when a value cannot be converted to the inferred concrete type instead of throwing;
- keeps `ARRAY<DOUBLE>` and `ARRAY<VARCHAR>` return types concrete rather than introducing `ANY` or runtime-dependent operator coercion;
- adds unit and indexed-field integration coverage for failed numeric conversion;
- marks `foreach` as experimental and removes product-comparison wording from plugin documentation and test comments.

### User impact

Field-backed JSON arrays used by numeric `foreach` expressions no longer fail the request when an element is not numeric. Valid numeric values and numeric strings continue to use numeric evaluation, while incompatible values produce `null`.

### Validation

```text
./gradlew :core:test --tests org.opensearch.sql.expression.function.CollectionUDF.ForeachFunctionImplTest :ppl:test --tests org.opensearch.sql.ppl.calcite.CalcitePPLForeachTest
```

The change also adds the affected `ForeachFieldJsonIT` coverage to the no-pushdown integration suite.

Related review: https://github.com/opensearch-project/sql/pull/5613#discussion_r3599082074

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/ForeachPlanner.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/ForeachPlanner.java
@@ -432,9 +432,9 @@ class ForeachPlanner {
    * VARCHAR.
    *
    * <p>For {@code json_array()} calls and string literals the content is visible at plan time;
-   * mixed content is rejected. For opaque expressions — typically a field holding JSON text, the
-   * primary Splunk use of json_array mode — content is unknowable, so infer from usage: an item
-   * placeholder consumed by arithmetic means numeric elements, anything else means strings.
+   * mixed content is rejected. For opaque expressions, typically a field holding JSON text, content
+   * is unknowable, so infer from usage: an item placeholder consumed by arithmetic means numeric
+   * elements, anything else means strings.
    */
   private SqlTypeName jsonElementType(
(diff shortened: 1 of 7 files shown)


songkant-aws:
@penghuo As per merged PR disucssion: https://github.com/opensearch-project/sql/pull/5613#discussion_r3600694347, I've changed document phrase and coercion behavior from parsing unknown dynamic type directly from index storage. For now, I haven't find a good solution to immediately solve that. 

Additionally, none of the user’s example queries use `json_str` mode. All of them involve `multifield` mode, where the type is known in advance and the field value can be normalized accordingly. Therefore, I decided to fall back to null when the type is ambiguous in this case.

noCharger (approved):

penghuo (approved):

Issue #5640, opened 2026-07-21
[RFC] PPL multikv command — extract rows from a field


noCharger:
# [RFC] PPL `multikv` command — extract rows from a field

## Problem

Operational and log data frequently arrives with a table packed inside a single field, or with
records nested inside a document, and PPL has no terse way to turn either into rows:

- **Table-formatted text.** Command output such as `ps`, `top`, `netstat`, or `df` is captured as
  one text blob (for example a `message` field). It has a header row and one record per line, but
  PPL sees it as a single opaque string. Extracting the columns today means hand-rolling `rex` /
  `parse` regexes per tool, which is brittle against variable whitespace.
- **Structured records.** An array of objects (`procs = [{pid,cpu}, ...]`) or a single object must
  be exploded and projected with a `mvexpand` + `eval field.sub` + `fields` chain, which is verbose
  and easy to get wrong.

Splunk solves the first case with `multikv`. There is no PPL equivalent, and no single command that
covers both the text and the structured shapes with one interface.

## Proposal

Add a streaming (mid-pipeline) command `multikv` that reads an input field and emits **one row per
source record** — a one-to-many, row-multiplying command. It runs on the coordinating node and
requires the Calcite (v3) engine.

```
multikv [field=<name>] [fields <col>[:<type>]...] [forceheader=<int>] [noheader=<bool>]
```

- `field=<name>` selects the input field. Defaults to `_raw`. OpenSearch has no implicit `_raw`
  field, so callers either pass `field=<name>` or place text in `_raw` first with `eval _raw=<f>`.
- `fields <col>[:<type>]...` declares the output columns by name; an optional `:<type>`
  (`string|boolean|int|integer|long|float|double`) types the column at plan time (see Column
  typing). This is the only form that yields named columns, and it is what makes v1 fixed-schema
  (see Scope).
- `forceheader=<int>` uses the given 1-based line as the header, skipping banner lines above it.
- `noheader=<bool>` performs row explosion only (no named columns), for example to count data lines.

### Input modes (dispatch on the input field's type)

`multikv` inspects the plan-time type of `field=<name>` and routes to one of three paths. This is
the key design point: the same `field= ... fields ...` interface works whether the field is raw
text or already structured.

| Input field type | Calcite type | Behavior |
|---|---|---|
| Text | `VARCHAR` | Parse the table text: a header row names the columns, each following line becomes a row. Extracted values are `string`. |
[...]

PR #5641, opened 2026-07-21
Add PPL multikv command (fixed-schema)

noCharger:
### Description

Adds the PPL `multikv` command: a streaming, row-multiplying command that extracts fields from an input field and emits one row per source record. It requires the Calcite (v3) engine.

The input field is selected with `field=<name>` (defaults to `_raw`); output columns are declared with `fields <col>...`. `multikv` dispatches on the input field's plan-time type:

- **Text** (`VARCHAR`): parse table-formatted text (for example `ps` / `top` / `netstat` / `df` output) into columns. Extracted values are typed `string`.
- **Array of objects** (`ARRAY<ANY>`): explode into one row per element, reading each declared column from the element and preserving its type.
- **Single object** (`MAP<VARCHAR,ANY>`): read each declared column, preserving its type; emits one row.

Nested container values are returned serialized (matching the merged `makeresults` convention); extract deeper fields downstream with `spath` or another `multikv field=<subfield>`.

Design, semantics, and deferred scope (runtime auto-header, aligned-offset parsing, `filter`/`rmorig`, parse-to-MAP optimization) are in the RFC: **#5640**.

### Changes
- Grammar + `Multikv` AST + Calcite lowering, including the `field=<name>` input selector, `forceheader`, and `noheader`.
- Type-dispatch in `visitMultikv`: text → `MULTIKV_SPLIT` / `mvexpand` / `MULTIKV_EXTRACT`; array-of-objects → `mvexpand` + `INTERNAL_ITEM`; single object → `INTERNAL_ITEM` only.
- Bare `multikv` (no `fields`, no `noheader`) is rejected at the semantic layer with actionable guidance (v1 is fixed-schema).
- V2 (non-Calcite) engine returns an unsupported-command error, matching the merged `makeresults` convention.
- User docs: `docs/user/ppl/cmd/multikv.md` and an `index.md` row.

### Test coverage
- Unit: `CalcitePPLMultikvTest`, `AstBuilderTest#testMultikvCommand` (incl. `field=`), `PPLQueryDataAnonymizerTest`.
- IT: `CalcitePPLMultikvCommandIT` (covering `field=`, text, array-of-objects, and single-object modes), `NewAddedCommandsIT#testMultikv`.

### Check List


core/src/main/java/org/opensearch/sql/ast/tree/Multikv.java
```diff
@@ -0,0 +1,89 @@
...
+ * is set. Runtime header auto-detection (no fields, no forceheader, no noheader) is not supported;
+ * such a query is rejected at the field-resolution phase with a message directing the author to add
+ * a {@code fields} clause.
+ */
+@ToString
+@EqualsAndHashCode(callSuper = false)
+@Getter
+public class Multikv extends UnresolvedPlan {
+
+  /** Default Splunk input field for multikv. */
```
songkant-aws: Avoid Splunk wording in comments.

core/src/main/java/org/opensearch/sql/ast/tree/Multikv.java
```diff
@@ -0,0 +1,89 @@
...
+  public static final String DEFAULT_INPUT_FIELD = "_raw";
```
songkant-aws, line 38: We don't have such metadata field yet. Is it expected to be explicitly generated by user in `makeresults` command?

docs/user/ppl/cmd/multikv.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Example 1: Extract a single column
+
+The following query reads the table text from the `raw` field with `field=` and extracts the `pctIdle` column:
+
+```ppl
+source=metrics
+| multikv field=raw fields pctIdle
+| fields pctIdle
+```
```
songkant-aws, line 41: We need to print real output for each example ppl queries.

docs/user/ppl/index.md
```diff
@@ -84,6 +84,7 @@ source=accounts
 | [explain command](cmd/explain.md) | 3.1 | stable (since 3.1) | N/A | Explain the plan of query. |
 | [show datasources command](cmd/showdatasources.md) | 2.4 | stable (since 2.4) | N/A | Query datasources configured in the PPL engine. |
 | [makeresults command](cmd/makeresults.md) | 3.8 | experimental (since 3.8) | No | Generate in-memory rows for testing and seeding, optionally from inline CSV/JSON data. |
+| [multikv command](cmd/multikv.md) | 3.8 | experimental (since 3.8) | No | Extract fields from table-formatted text in a field, emitting one row per table data row. |
```
songkant-aws: nit: version 3.9?

ppl/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -1107,6 +1108,38 @@ public UnresolvedPlan visitMvexpandCommand(OpenSearchPPLParser.MvexpandCommandCo
     return new MvExpand(field, limit);
   }
 
+  @Override
+  public UnresolvedPlan visitMultikvCommand(OpenSearchPPLParser.MultikvCommandContext ctx) {
+    List<Field> fields = null;
+    String inField = Multikv.DEFAULT_INPUT_FIELD;
+    Integer forceHeader = null;
+    boolean noHeader = false;
+    boolean rmOrig = true; // Splunk default
```
songkant-aws: Avoid Splunk wording in comments.

docs/user/ppl/cmd/multikv.md
```diff
@@ -0,0 +1,97 @@
+
+# multikv
```
songkant-aws, line 2: add it to docs/category.json to enable docs test.

docs/user/ppl/cmd/multikv.md
```diff
@@ -0,0 +1,97 @@
...
+When `field=` points at an array of objects, `multikv` emits one row per element and reads each declared column from the element, preserving the element value:
+
+```ppl
+source=hosts
+| multikv field=procs fields pid cpu
+```
+
+For a document with `procs = [{"pid":1,"cpu":0.5},{"pid":42,"cpu":9.1}]`, the query returns two rows: `(1, 0.5)` and `(42, 9.1)`. When `field=` points at a single object rather than an array, one row is returned. Nested container values are returned as-is; extract deeper fields downstream with `spath` or another `multikv field=<subfield>`.
+
+## Limitations
```
songkant-aws, line 132: Per the RFC discussion, we should call out that users must explicitly cast expanded key-value fields from VARCHAR or ANY to the expected types when type-sensitive processing is required. 

Please confirm that this reflects an outcome discussed and agreed upon with users.

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalcitePPLMultikvCommandIT.java
```diff
@@ -0,0 +1,153 @@
...
+
+  @Test
+  public void testMultikvFields() throws IOException {
+    // Declared single column: auto-detected header maps pctIdle -> its column.
+    JSONObject result =
+        executeQuery(
+            "source=test_multikv | eval _raw = raw | multikv fields pctIdle | fields pctIdle");
+    verifySchema(result, schema("pctIdle", "string"));
+    verifyDataRows(result, rows("90"), rows("92"));
+  }
```
songkant-aws, line 84: Add more complex ITs to let downstream commands correctly consume multikv results after casting

ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLMultikvTest.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Logical-plan tests for the {@code multikv} command (fixed-schema).
+ *
+ * <p>multikv lowers to a three-layer rewrite: an {@code eval} that calls {@code MULTIKV_SPLIT} into
+ * a helper record column, an mvexpand (Uncollect over a Correlate) that explodes one row per table
+ * data row, and a per-declared-column {@code MULTIKV_EXTRACT} eval followed by a narrowing project.
+ * These assertions pin the presence and absence of those operators rather than the full Rex
+ * rendering, which is exercised end-to-end by CalcitePPLMultikvCommandIT.
+ */
+public class CalcitePPLMultikvTest extends CalcitePPLAbstractTest {
```
songkant-aws, line 26: Please also add tests to CalciteExplainIT to assert printed logical plans are correct.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -4663,6 +4665,143 @@ public RelNode visitMvExpand(MvExpand mvExpand, CalcitePlanContext context) {
...
+    // its schema; the structured branch reuses that build, the text branch discards it.
+    boolean savedProjectVisited = context.isProjectVisited();
+    RelNode probe = node.getChild().get(0).accept(this, context);
+    RelDataTypeField probeField = probe.getRowType().getField(node.getInField(), true, false);
+    boolean structuredArray =
+        probeField != null
+            && (SqlTypeUtil.isArray(probeField.getType())
+                || SqlTypeUtil.isMultiset(probeField.getType()));
+    boolean structuredMap = probeField != null && SqlTypeUtil.isMap(probeField.getType());
+    if (structuredArray || structuredMap) {
```
songkant-aws, line 4719: For input types other than ARRAY and MAP, the current implementation silently falls back to the text path and treats them as strings. Should we validate the input type here and reject unsupported types during planning?

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -4663,6 +4665,143 @@ public RelNode visitMvExpand(MvExpand mvExpand, CalcitePlanContext context) {
...
+        projected.add(item);
+        names.add(col);
+      }
+      relBuilder.project(projected, names);
+      context.setProjectVisited(true);
+      return relBuilder.peek();
+    }
+    // Text input: discard the probe build and run the split pipeline on a fresh build.
+    context.relBuilder.build();
+    context.setProjectVisited(savedProjectVisited);
```
songkant-aws: The type-based dispatch is reasonable, but the current implementation visits the child twice. The first visit mutates the shared RelBuilder and CalcitePlanContext; text mode then calls build() and only restores isProjectVisited before rebuilding the child. This is not a full rollback and may affect features such as correlation binding or plan-node tracking. Could we build the child once and lower text mode from the resulting RelNode? Is there a specific reason to discard the build in text mode? The schema is known at stack peek and it seems feasible to build text mode specific MvExpand Relnode on top of it.

PR #5642, opened 2026-07-21
[Feature] `percentfield` and `showperc` with `top` and `rare` commands.

### LINKED ISSUE #5530: [FEATURE] `showperc` with `top`
With `top` (and `rare`) commands, it'd be nice to conveniently get percentages. Currently they need to be computed manually with `eventstats`: `stats count() as url_count by url | eventstats sum(url_count) as total | eval perc = 100.0 * url_count / total`.

A simple implementation would rewrite `top` with `showperc` to its equivalent.

```kql
source=whatever | top url showperc=true
```

Some more complex eval involving computing counts as a special-case without an `eventstats` node.

N/A

AjimelecGonzalez:
### Description
Add `percentfield` and `showperc` options to the `top` and `rare` PPL commands. When `showperc=true` is specified, a percent column is added to the output showing each value's percentage of the total count within its partition (group-by clause). The `percentfield` option allows customizing the name of the percentage column (default: percent).

The percentage is computed as `ROUND(100.0 * count / SUM(count) OVER (PARTITION BY group_fields), 6)` and is rounded to 6 decimal places. When used with a by clause, percentages are calculated within each partition (summing to 100% per group).

#### Syntax
`source=index | top [percentfield='<name>'] showperc=true field_name [by group_field]`
`source=index | rare [percentfield='<name>'] showperc=true field_name [by group_field]`

#### `Top` Example
`source=otellogs | top showperc=true severityText`

Returns:

| severityText | count | percent |
| :---         | :---: |    ---: |
| ERROR        | 7     | 35.0    |
| INFO         | 6     | 30.0    |
| WARN         | 4     | 20.0    |
| DEBUG        | 3     | 15.0    |

#### `Rare` Example
`source=otellogs | rare showperc=true severityText`

Returns:

| severityText | count | percent |
| :---         | :---: |    ---: |
| DEBUG        | 3     | 15.0    |
| WARN         | 4     | 20.0    |
| INFO         | 6     | 30.0    |
| ERROR        | 7     | 35.0    |

### Custom Percentage Field Name Example
`source=otellogs | top percentfield='pct' showperc=true severityText`

Returns:

| severityText | count | pct |
| :---         | :---: |    ---: |
| ERROR        | 7     | 35.0    |
| INFO         | 6     | 30.0    |
| WARN         | 4     | 20.0    |
| DEBUG        | 3     | 15.0    |

### Related Issues
Resolves #5530 

### Check List

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -223,6 +223,8 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
   private final MapPathPreMaterializer mapPathMaterializer;
   private final ForeachPlanner foreachPlanner;
 
+  private static final int PERCENT_DECIMAL_PLACES = 6;
+
   public CalciteRelNodeVisitor(DataSourceService dataSourceService) {
     this.rexVisitor = new CalciteRexNodeVisitor(this);
     this.aggVisitor = new CalciteAggCallVisitor(rexVisitor);
@@ -3264,6 +3266,39 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     orderKeys.add(countField);
     orderKeys.addAll(tieBreakKeys);
 
+    // 3. compute percentage if showperc=true
+    Boolean showPerc = (Boolean) argumentMap.get(RareTopN.Option.showPerc.name()).getValue();
+    if (showPerc) {
+      String percentFieldName =
+          (String) argumentMap.get(RareTopN.Option.percentField.name()).getValue();
+
+      RexNode totalWindowOver =
+          PlanUtils.makeOver(
+              context,
+              BuiltinFunctionName.SUM,
+              context.relBuilder.field(countFieldName),
+              List.of(),
+              partitionKeys,
+              List.of(),
+              WindowFrame.rowsUnbounded());
+
+      RexNode hundred = context.relBuilder.literal(new BigDecimal("100.0"));
+      RexNode countCast =
+          context.relBuilder.cast(context.relBuilder.field(countFieldName), SqlTypeName.DOUBLE);
+      RexNode totalCast = context.relBuilder.cast(totalWindowOver, SqlTypeName.DOUBLE);
+      RexNode numerator = context.relBuilder.call(SqlStdOperatorTable.MULTIPLY, hundred, countCast);
+      RexNode percValue = context.relBuilder.call(SqlStdOperatorTable.DIVIDE, numerator, totalCast);
+
+      // Round the percent value 6 decimal places
+      RexNode roundedPerc =
+          context.relBuilder.call(
+              SqlStdOperatorTable.ROUND,
+              percValue,
+              context.relBuilder.literal(PERCENT_DECIMAL_PLACES));
+
+      context.relBuilder.projectPlus(context.relBuilder.alias(roundedPerc, percentFieldName));
+    }
+
     RexNode rowNumberWindowOver =
         PlanUtils.makeOver(
             context,
@@ -3276,14 +3311,14 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
     context.relBuilder.projectPlus(
         context.relBuilder.alias(rowNumberWindowOver, ROW_NUMBER_COLUMN_FOR_RARE_TOP));
 
-    // 3. filter row_number() <= k in each partition
+    // 4. filter row_number() <= k in each partition
     int k = node.getNoOfResults();
     context.relBuilder.filter(
         context.relBuilder.lessThanOrEqual(
             context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_RARE_TOP),
             context.relBuilder.literal(k)));
 
-    // 4. project final output. the default output is group by list + field list
+    // 5. project final output: group by list + field list, optionally count and percent
     Boolean showCount = (Boolean) argumentMap.get(RareTopN.Option.showCount.name()).getValue();
     if (showCount) {
       context.relBuilder.projectExcept(context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_RARE_TOP));
--- a/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTopCommandIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTopCommandIT.java
@@ -6,7 +6,9 @@
 package org.opensearch.sql.calcite.remote;
 
 import static org.opensearch.sql.legacy.TestsConstants.TEST_INDEX_BANK_WITH_NULL_VALUES;
+import static org.opensearch.sql.util.MatcherUtils.rows;
 import static org.opensearch.sql.util.MatcherUtils.schema;
+import static org.opensearch.sql.util.MatcherUtils.verifyDataRows;
 import static org.opensearch.sql.util.MatcherUtils.verifyNumOfRows;
 import static org.opensearch.sql.util.MatcherUtils.verifySchemaInOrder;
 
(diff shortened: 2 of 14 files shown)


integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTopCommandIT.java
```diff
@@ -41,6 +41,25 @@ public void testTopCommandUseNullFalse() throws IOException {
     verifyNumOfRows(result, 5);
   }
 
+  @Test
+  public void testTopCommandShowPerc() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format("source=%s | top showperc=true age", TEST_INDEX_BANK_WITH_NULL_VALUES));
+    verifySchemaInOrder(
+        result, schema("age", "int"), schema("count", "bigint"), schema("percent", "double"));
```
ahkcs, line 52: Can we also have `verifyDataRows` for the ITs

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTopCommandIT.java
```diff
@@ -41,6 +41,25 @@ public void testTopCommandUseNullFalse() throws IOException {
...
+  }
+
+  @Test
+  public void testTopCommandShowPercWithoutShowCount() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | top showperc=true showcount=false age",
+                TEST_INDEX_BANK_WITH_NULL_VALUES));
+    verifySchemaInOrder(result, schema("age", "int"), schema("percent", "double"));
```
ahkcs, line 71: ditto

integ-test/src/test/java/org/opensearch/sql/ppl/TopCommandIT.java
```diff
@@ -58,4 +58,19 @@ public void testTopNWithGroup() throws IOException {
...
+  @Test
+  public void testTopWithShowPerc() throws IOException {
+    JSONObject result =
+        executeQuery(String.format("source=%s | top showperc=true gender", TEST_INDEX_ACCOUNT));
+    if (isCalciteEnabled()) {
+      verifySchemaInOrder(
+          result,
+          schema("gender", "string"),
+          schema("count", "bigint"),
+          schema("percent", "double"));
```
ahkcs, line 71: ditto

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3251,7 +3249,35 @@ public RelNode visitRareTopN(RareTopN node, CalcitePlanContext context) {
...
+      RexNode totalCast = context.relBuilder.cast(totalWindowOver, SqlTypeName.DOUBLE);
+      RexNode numerator = context.relBuilder.call(SqlStdOperatorTable.MULTIPLY, hundred, countCast);
+      RexNode percValue = context.relBuilder.call(SqlStdOperatorTable.DIVIDE, numerator, totalCast);
+
+      // Round the percent value 2 decimal places
+      RexNode roundedPerc =
+          context.relBuilder.call(
+              SqlStdOperatorTable.ROUND, percValue, context.relBuilder.literal(2));
+
+      context.relBuilder.projectPlus(context.relBuilder.alias(roundedPerc, "percent"));
```
ahkcs: The percentage column is hardcoded to the name percent with no collision check, so if a count field or grouped field already uses that name, Calcite silently renames the percentage column to percent0.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3251,7 +3249,35 @@ public RelNode visitRareTopN(RareTopN node, CalcitePlanContext context) {
             context.relBuilder.field(ROW_NUMBER_COLUMN_FOR_RARE_TOP),
             context.relBuilder.literal(k)));
 
-    // 4. project final output. the default output is group by list + field list
+    // 4. compute percentage if showperc=true
+    Boolean showPerc = (Boolean) argumentMap.get(RareTopN.Option.showPerc.name()).getValue();
+    if (showPerc) {
+      RexNode totalWindowOver =
```
ahkcs, line 3275: Percentages are computed after the top/rare-N limit is applied, so they're taken relative to only the surviving rows instead of the full dataset (did a local run on top 1 showperc=true reports 100% instead of 28.57%).

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTopCommandIT.java
AjimelecGonzalez: Will add

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
AjimelecGonzalez: Will add check and exception

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
AjimelecGonzalez: Fixed the problem. Was calculating percentages after filtering. And add tests to check the logic.

ahkcs (approved):

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTopCommandIT.java
ahkcs: Can we have `verifyDataRows` instead of `verifyNumOfRows` for clarity?

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTopCommandIT.java
ahkcs: same here

integ-test/src/test/java/org/opensearch/sql/calcite/remote/CalciteTopCommandIT.java
AjimelecGonzalez: will have both

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3264,6 +3232,45 @@ public RelNode visitRareTopN(RareTopN node, CalcitePlanContext context) {
...
+    Boolean showPerc = (Boolean) argumentMap.get(RareTopN.Option.showPerc.name()).getValue();
+    if (showPerc) {
+      String percentFieldName =
+          (String) argumentMap.get(RareTopN.Option.percentField.name()).getValue();
+
+      if (context.relBuilder.peek().getRowType().getFieldNames().contains(percentFieldName)) {
+        throw new IllegalArgumentException(
+            "Field `"
+                + percentFieldName
+                + "` already exists in the output. Consider renaming with percentfield='xyz'");
```
Swiddis: suggestion: Would it be better user experience if we provided a new percent field automatically?

e.g. generate `percent0`, `percent1`, etc. Joins already do this with join IDs (`id` -> `id0`, etc).

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
```diff
@@ -3264,6 +3232,45 @@ public RelNode visitRareTopN(RareTopN node, CalcitePlanContext context) {
...
+      }
+      RexNode totalWindowOver =
+          PlanUtils.makeOver(
+              context,
+              BuiltinFunctionName.SUM,
+              context.relBuilder.field(countFieldName),
+              List.of(),
+              partitionKeys,
+              List.of(),
+              WindowFrame.rowsUnbounded());
```
Swiddis, line 3283: issue: This has a correctness issue for high-cardinality fields because the total is based on the returned buckets.

The current impl works well for low-cardinality fields, but for high cardinality we get a truncated bucket set. I'm worried about long-tail types of top-N cases (e.g. "top 10 endpoints" from a service with many assets). `rare` would be much more affected in that case.

In principle we should be using the `sum_other_doc_count` value that's returned by aggregations if it's available: https://docs.opensearch.org/latest/aggregations/bucket/terms/#example-response. But fixing may be complex depending on the exact conditions where that value is present, I don't know them off the top but it may only apply to `term` aggregations and not composite aggs, and it means percent aggregations need to be finalized at the coordinator. If it's not available then the "right" way is to submit a whole separate `_count` request but that'll complicate query flow.

If it's a lot to patch, I think we can leave this behavior, but should document the limitation. You can manually work around it by using a subquery or something like eventstats to work with `stats count()` computed ahead of time. (Alternatively, if such a method exists, we could rewrite the query plan to be equivalent to it.)

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
AjimelecGonzalez: Sure, will remove the check.

core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
AjimelecGonzalez: Will look into it and document, I believe that the using the `sum_other_doc_count `value would be the best choice here but will have to deep dive. For now I will add the limitations to the md file for top and rare commands.

AjimelecGonzalez: Did testing by changing the settings for `plugins.query.size_limit`, `plugins.query.buckets`, `search.max_buckets` from 1, 10000, and 50000. And no change to the percentages with rare and top commands. No limitation.

PR #5646, opened 2026-07-22
Push down aggregation on text field without .keyword sub-field

### LINKED ISSUE #5634: [BUG] Non-pushdownable stats/eventstats silently forces a full-scan PIT and fails with an opaque error
## Query Information
**PPL Command/Query:**
```ppl
source = opensearch_dashboards_sample_data_logs | stats values(response) by geo.dest
```

The failure is not specific to `values()`. Any `stats`/`eventstats` whose aggregation **cannot be pushed down** to OpenSearch triggers it. Confirmed triggers:
```ppl
# non-pushdownable aggregation function
source = opensearch_dashboards_sample_data_logs | stats values(response) by geo.dest
source = opensearch_dashboards_sample_data_logs | stats list(response) by geo.dest
# window function (eventstats -> COUNT() OVER (...), no pushdown form)
source = opensearch_dashboards_sample_data_logs | eventstats count() by response
# group-by on a `text` field WITHOUT a .keyword subfield, or a `nested`/array field
source = <index> | stats count by <analyzed_text_field_without_keyword>
```

**Expected Result:**
The aggregation is computed and rows are returned. If an aggregation genuinely cannot be pushed down, it should still execute correctly (and ideally without silently scanning the entire index).

**Actual Result:**
```
java.sql.SQLException: exception while executing query: Error occurred while creating PIT
for internal plugin operation
```
With the diagnostic wrapping from #5631 applied, the previously-hidden root cause is visible:
```
... nested: OpenSearchRejectedExecutionException[Trying to create too many Point In Time
contexts. Must be less than or equal to: [N]. This limit can be set by changing the
[search.max_open_pit_context] setting.]
```
On the stock plugin (without #5631) the underlying cause is discarded and the caller only sees the opaque `Error occurred while creating PIT for internal plugin operation`.

## Dataset Information
**Dataset/Schema Type**
- [x] OpenTelemetry (OTEL)  
- [x] Custom (public reproduction uses the Dashboards sample logs dataset)

**Index Mapping** (public reproduction)
Uses the built-in `opensearch_dashboards_sample_data_logs` index as-is. No custom mapping required — `values()`/`eventstats` refuse pushdown regardless of mapping.

For the field-mapping variant (the production trigger), a group-by field mapped as `text` with **no** `keyword` subfield is enough:
```json
{
  "settings": { "index.max_result_window": 10000 },
  "mappings": {
    "properties": {
      "resource": { "properties": { "attributes": { "properties": { "k8s_label": { "properties": {
        "applicationid": { "type": "text" }
      }}}}}}
    }
  }
}
```
```ppl
source = <that index> | stats count by `resource.attributes.k8s_label.applicationid`
```

## Bug Description
**Issue Summary:**

A `stats`/`eventstats` query whose aggregation cannot be pushed down to OpenSearch silently degrades into a **full-index document scan** on the coordinator, which forces internal Point-in-Time (PIT) creation and can fail the whole query. There are two distinct problems stacked here.

### Root cause #1 — a refused aggregation silently becomes a full-index scan (the real defect)

When `stats ... by X` can be translated into a native OpenSearch composite aggregation, the request is sent with `size: 0` and no PIT is involved. When it **cannot** be translated, `CalciteLogicalIndexScan.pushDownAggregate(...)` returns `null` and Calcite falls back to aggregating client-side over a raw document scan. That fallback scan carries `requestedTotalSize = Integer.MAX_VALUE (2147483647)`.

In `OpenSearchRequestBuilder.build(...)`:
- `sourceBuilder.size() == 0` is **false** (size is MAX_VALUE), so it does not short-circuit;
- `buildRequestWithPit(...)` then evaluates `startFrom + size > maxResultWindow` → `2147483647 > 10000` is always true → `createPit(...)` is called.

Note the scan does **not** issue one `size = MAX_VALUE` request. `buildRequestWithPit` clamps the per-page size to `maxResultWindow` and paginates through the whole index via the PIT (`OpenSearchRequestBuilder.java:130-140`). The net effect is "stream every matching document to the coordinator, PIT-paginated at `max_result_window` per page, and aggregate in-memory in Calcite." On a small index this is invisible; on a large index it is pathological, and the PIT it requires is simply the first place a visible error surfaces (even if PIT creation succeeded, the query would be extremely slow/memory-heavy).

**Refusal path, confirmed in source.** For `values()`/`list()` the refusal is specific — these are *recognized* aggregations, not unknown names:
1. Every PPL UDAF is stamped `SqlKind.OTHER_FUNCTION` (`core/.../calcite/utils/UserDefinedFunctionUtils.java:99`); `values`/`list` are defined via `createUserDefinedAggFunction` (`core/.../expression/function/PPLBuiltinOperators.java:514-522`).
2. `AggregateAnalyzer#createRegularAggregation`'s `OTHER_FUNCTION` case resolves the name back to `BuiltinFunctionName.VALUES` / `LIST` — the lookup **succeeds** (`opensearch/.../request/AggregateAnalyzer.java:559-561`).
3. Its inner `switch` only handles `TAKE`, `FIRST`, `LAST`, `PERCENTILE_APPROX`, `DISTINCT_COUNT_APPROX`; `VALUES`/`LIST` hit the `default` → `throw new AggregateAnalyzerException("Unsupported push-down aggregator ...")` (`AggregateAnalyzer.java:597-599`).
4. `analyze()`'s outer `catch (Throwable)` re-wraps it as the checked `ExpressionNotAnalyzableException` (`AggregateAnalyzer.java:323-325`).
5. `pushDownAggregate`'s `catch (Exception)` returns `null` (`CalciteLogicalIndexScan.java:426-431`); `AggregateIndexScanRule.apply` sees `null`, skips `call.transformTo`, and the `LogicalAggregate` stays a client-side operator over the plain scan (`AggregateIndexScanRule.java:100-104`).

Conditions that refuse aggregation pushdown (all confirmed in code):
- a non-pushdownable aggregation function such as `values()` / `list()` — via the path above;
- group-by field is `text` with no `.keyword` subfield — `OpenSearchTextType.toKeywordSubField(...)` returns `null`, the term-source resolution throws, and the same `catch → null` at `CalciteLogicalIndexScan#pushDownAggregate` fires (a *different* throw than the aggregator switch above);
- group-by field is `nested`/`ARRAY` — a separate early `return null` guard in `pushDownAggregate` (`CalciteLogicalIndexScan.java:392-401`, see #5006);
- the operation is a window function (`eventstats`), which has no aggregation-pushdown form.

By contrast, `count`/`avg`/`sum`/`min`/`max` have dedicated builder cases (`AggregateAnalyzer.java:481-528`), so `analyze()` returns non-null and `OpenSearchRequestBuilder#pushDownAggregation` sets `sourceBuilder.size(0)` (`OpenSearchRequestBuilder.java:214`), routing `build()` to the no-PIT path. (Nuance: `min`/`max` on non-numeric/non-temporal fields push down as a `top_hits` aggregation rather than a native min/max metric agg — still `size:0`, still no PIT.)

### Root cause #2 — the PIT failure is reported with an opaque, cause-less message

When the forced `createPit` fails, `OpenSearchNodeClient#createPit` (and the legacy `PointInTimeHandlerImpl`) discard the underlying cause and throw a fixed string `"Error occurred while creating PIT for internal plugin operation"`. They also only recognize a security denial when `OpenSearchSecurityException` is the *direct* cause, so a transport-wrapped denial on a multi-node cluster is flattened into the same opaque string. This makes a resource failure (PIT-context exhaustion), a timeout, and a permission denial all indistinguishable to the operator. Addressed by #5631 (surfaces the root cause and unwraps the full cause chain). #5631 is a diagnostics fix only and does **not** address root cause #1.

## Verification

Re-verified on a live single-node cluster (OpenSearch 3.x, `plugins.calcite.enabled: true`, `opensearch_dashboards_sample_data_logs`, 14,074 docs) at the behavioral, plan (`_explain`), and source level.

**The query is trivial — PIT capacity is the only barrier.** Same query, only `search.max_open_pit_context` changed:

| `search.max_open_pit_context` | Result of `stats values(response) by geo.dest` |
|---|---|
| `0` (node saturated / at limit) | ❌ `Error occurred while creating PIT ...` |
| `300` (headroom) | ✅ **173 rows** |

The successful run returns only 173 buckets — the result is tiny; it fails at `=0` purely because the client-side fallback opens a PIT to read its input documents. Under peak production concurrency a node sits effectively at its limit, reproducing the `=0` case.

**A trailing `| head N` does not prevent the PIT.** `... | stats values(response) by geo.dest | head 200` still fails: `head` bounds the aggregation **output** (173 buckets, under 200 → nothing to trim), but the PIT is allocated for the **input** document scan that runs before the limit applies.

**`_explain` confirms the split at plan level:**

| Query | `AGGREGATION` in PushDownContext | `requestedTotalSize` | client-side agg |
|---|---|---|---|
| `stats count() by geo.dest` | present (pushed down) | — | no |
| `stats values(response) by geo.dest` | absent | 2147483647 | `EnumerableAggregate` |

**Behavior matrix at `max_open_pit_context=0`** (❌ = PIT created then rejected; ✅ = no PIT):

| Query | Outcome |
|---|---|
| `stats count() by geo.dest` | ✅ 173 |
| `stats count() by geo.dest \| head 200` | ✅ 173 |
| `stats avg(bytes) by geo.dest` | ✅ 173 |
| `stats sum(bytes) by geo.dest` | ✅ 173 |
| `stats values(response) by geo.dest` | ❌ PIT |
| `stats values(response) by geo.dest \| head 200` | ❌ PIT |
| `stats list(response) by geo.dest` | ❌ PIT |
| `head 500` (Logs-tab shape) | ✅ 500 |
| `fields response \| head 500` | ✅ 500 |

**Steps to Reproduce:**
1. Enable the Calcite engine: `PUT _cluster/settings {"transient":{"plugins.calcite.enabled":true}}`
2. (To force the PIT failure deterministically on an idle single node) `PUT _cluster/settings {"transient":{"search.max_open_pit_context":0}}`. At production scale this exhaustion happens naturally under peak concurrency; the setting just makes it reproducible locally.
3. Load the `opensearch_dashboards_sample_data_logs` sample dataset.
4. Run `source = opensearch_dashboards_sample_data_logs | stats values(response) by geo.dest`.
5. Observe the `Error occurred while creating PIT ...` failure.
6. Contrast: a bounded document fetch such as `source = opensearch_dashboards_sample_data_logs | fields bytes | head 500` succeeds (size < window ⇒ no PIT), and a pushdownable `source = ... | stats count by response` succeeds (`size: 0` ⇒ no PIT).

**Impact:**
- User-facing query failures on aggregation/statistics views for fields that don't push down (commonly deep OTel fields dynamically mapped as analyzed `text` or `nested`).
- The failure is load- and mapping-dependent, so it presents as intermittent ("only at peak") and is hard to reproduce.
- Because even users with full permissions hit it, and because of root cause #2, it is easily misdiagnosed as a permissions problem when the true cause is a full-scan-forced PIT hitting `search.max_open_pit_context` (or a circuit breaker/timeout).
- Silent degradation to a full-index scan is a latent performance/stability risk independent of the PIT error.

## Environment Information
**OpenSearch Version:** reproduced on OpenSearch 3.x with the Calcite engine enabled (`plugins.calcite.enabled: true`).

**Additional Details:**
- Related diagnostics PR: #5631 (makes the underlying PIT failure visible; not a fix for root cause #1).
- Related nested/array pushdown limitation: #5006.

## Suggested Direction
1. **Preferred (root cause #1):** avoid silently degrading a refused aggregation into a `size = MAX_VALUE` full scan. Options: keep aggregating but bound/paginate sanely; fall back to a pushed-down form where possible (e.g. use a `.keyword` subfield when present); or fail fast with an actionable message instead of an implicit full scan.
2. **Mitigation (root cause #2):** ship #5631 so operators can see whether a PIT failure is exhaustion, timeout, or permission. Separately, verify PITs are always released on query cancellation/failure so abandoned UI queries don't leak PIT contexts and accelerate exhaustion.

penghuo:
### Description
When a text field has no `.keyword` sub-field, PPL queries that group by or aggregate over that field silently fall back to a full `_source` scan with client-side aggregation, even when Calcite pushdown is enabled.

**Fix.** When `getReferenceForTermQuery()` returns `null` for a bare `RexInputRef`, route through `ScriptQueryExpression(rexNode, ...)` to build a Calcite script that reads the field from `_source`. This reuses the exact `SOURCE` path already established by `TermQuery`/`LikeQuery`/`RexStandardizer.visitInputRef` and consumed at runtime by `CalciteScriptEngine.ScriptDataContext.getFromSource` → `SourceLookup.get(...)`. The change is in `AggregateAnalyzer.AggregateBuilderHelper.build` — one branch added.

### Related Issues
Resolves #5634

### Check List
  - Unit tests in `AggregateAnalyzerTest`: `analyze_aggCall_TextWithoutKeyword_countPushesDownAsScript`, `analyze_groupBy_TextWithoutKeyword` (asserts scripted `value_count` / `terms` DSL and `SOURCES`/`DIGESTS` script params).
  - YAML rest tests in `text_agg_pushdown.yml` (10 cases) covering `top`, `stats … by`, `count(field)` — including result correctness and pushed-down DSL assertions — plus baselines for text+keyword and multi-index queries.


integ-test/src/test/resources/expectedOutput/calcite/chart_null_str.yaml
```diff
@@ -25,15 +25,11 @@ calcite:
...
...lciteEnumerableIndexScan(table=[[OpenSearch, opensearch-sql_test_index_bank_with_null_values]], PushDownContext=[[FILTER->AND(IS NOT NULL($1), IS NOT NULL($0)), AGGREGATION->rel#:LogicalAggregate.NONE.[](input=RelSubset#,group={0, 2},avg(balance)=AVG($1))], OpenSearchRequestBuilder(sourceBuilder={"from":0,"size":0,"timeout":"1m","query":{"bool":{"must":[{"exists":{"field":"gender","boost":1.0}},{"exists":{"field":"balance","boost":1.0}}],"adjust_pure_negative":true,"boost":1.0}},"aggregations":{"composite_buckets":{"composite":{"size":1000,"sources":[{"gender":{"terms":{"script":{"source":"{\"langType\":\"calcite\",\"script\":\"rO0ABXQAaXsKICAiZHluYW1pY1BhcmFtIjogMCwKICAidHlwZSI6IHsKICAgICJ0eXBlIjogIlZBUkNIQVIiLAogICAgIm51bGxhYmxlIjogdHJ1ZSwKICAgICJwcmVjaXNpb24iOiAtMQogIH0KfQ==\"}","lang":"opensearch_compounded_script","params":{"utcTimestamp": 0,"SOURCES":[1],"DIGESTS":["gender"]}},"missing_bucket":true,"missing_order":"first","order":"asc"}}},{"age0":{"histogram":{"field":"age","missing_bucket":true,"missing_order":"first","order":"asc","interval":10.0}}}]},"aggregations":{"avg(balance)":{"avg":{"field":"balance"}}}}}}, requestedTotalSize=2147483647, pageSize=null, startFrom=0)])
```
penghuo, line 29: EnumerableAggregate is pushed down as script.

penghuo:
@qianheng-aws  Please take a look PR.

qianheng-aws (approved):
LGTM

PR #5650, opened 2026-07-23
Fix IllegalStateException in UnifiedQueryPlanner.preserveCollation on composite-collation plans

noCharger:
### Description

`UnifiedQueryPlanner.preserveCollation` fails to plan any query whose top `RelNode` carries more than one collation in its trait set. It surfaces as a 500 `Failed to plan query`.

**Root cause.** The method reads the collation via:

```java
RelCollation collation = logical.getTraitSet().getCollation();
```

`RelTraitSet.getCollation()` delegates to `getTrait(RelCollationTraitDef.INSTANCE)`, which assumes a single collation. When a plan derives several collations, Calcite stores them as a `RelCompositeTrait` and `getTrait` refuses to return it, throwing:

```
java.lang.IllegalStateException: Trait index 1 has multiple values in this trait set; use getTraits(RelTraitDef) instead of getTrait(RelTraitDef)
```

Such a composite arises whenever a non-`Sort` top node is ordered by more than one equivalent key — for example `... | sort age | eval x = age`: `x` mirrors `age`, so the top `LogicalProject` is sorted on both `age` and `x` and derives two collations. A plan with a single derived collation reads back fine, which is why this was not caught earlier. (A single multi-column literal-row plan such as `makeresults format=csv data=...` derives one multi-field collation, not a composite, so it does not trigger this.)

**Fix.** Read the collation through the composite-safe `getTraits(RelCollationTraitDef.INSTANCE)` and materialize a `LogicalSort` only when there is exactly one genuine (non-empty) sort collation. A composite (several incidental derived collations) is not a user sort, so the plan is left unwrapped instead of crashing. Behavior for the single-collation case is unchanged.

### Related Issues
N/A

### Check List

### DIFF
--- a/api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
+++ b/api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
@@ -7,9 +7,10 @@ package org.opensearch.sql.api;
 
 import static org.opensearch.sql.monitor.profile.MetricName.ANALYZE;
 
+import java.util.List;
 import lombok.RequiredArgsConstructor;
 import org.apache.calcite.rel.RelCollation;
-import org.apache.calcite.rel.RelCollations;
+import org.apache.calcite.rel.RelCollationTraitDef;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.RelRoot;
 import org.apache.calcite.rel.core.Sort;
@@ -154,9 +155,20 @@ public class UnifiedQueryPlanner {
     }
 
     private RelNode preserveCollation(RelNode logical) {
-      RelCollation collation = logical.getTraitSet().getCollation();
-      if (!(logical instanceof Sort) && collation != RelCollations.EMPTY) {
-        return LogicalSort.create(logical, collation, null, null);
+      if (logical instanceof Sort) {
+        return logical;
+      }
+      // Only a genuine single-sort collation is materialized. A literal-row plan (makeresults
+      // data=) carries several incidental derived collations kept as a RelCompositeTrait; those
+      // are not a user sort, so getTraits returns more than one and we leave the plan unwrapped.
+      // RelTraitSet.getCollation() would instead cast the composite to a single RelCollation and
+      // throw.
+      List<RelCollation> collations =
+          logical.getTraitSet().getTraits(RelCollationTraitDef.INSTANCE);
+      if (collations != null
+          && collations.size() == 1
+          && !collations.get(0).getFieldCollations().isEmpty()) {
+        return LogicalSort.create(logical, collations.get(0), null, null);
       }
       return logical;
     }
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
```diff
@@ -154,9 +155,15 @@ public RelNode plan(String query) {
...
-        return LogicalSort.create(logical, collation, null, null);
+      if (logical instanceof Sort) {
+        return logical;
+      }
+      List<RelCollation> collations =
+          logical.getTraitSet().getTraits(RelCollationTraitDef.INSTANCE);
+      if (collations != null
+          && collations.size() == 1
+          && !collations.get(0).getFieldCollations().isEmpty()) {
+        return LogicalSort.create(logical, collations.get(0), null, null);
```
dai-chen, line 166: So the first collation must be on user specified fields instead of derived ones?

dai-chen (approved):
Minor comment. Thanks for the fix!

api/src/main/java/org/opensearch/sql/api/UnifiedQueryPlanner.java
dai-chen: Also this was copied from https://github.com/opensearch-project/sql/blob/92acbec3ca03fc04095cddfb4da8266b87c821e1/core/src/main/java/org/opensearch/sql/executor/QueryService.java#L1074. Does it have similar issue or not?

noCharger: On the first one, the guard fires only when the top node carries exactly one non-empty collation. That single collation equals what the old `getTraitSet().getCollation()` returned, so the single-collation path keeps its prior behavior. When several equivalent collations appear, say `sort age | eval x = age` where `x` mirrors `age` so the top `Project` derives an ordering on both `age` and `x`, the plan is left unwrapped. Those extra collations are incidental orderings derived from equal expressions. The real user ordering already lives in the inner `sort`, and a `Project` over a `Sort` preserves order, so the final result stays sorted and a top-level sort would be redundant here. That way the guard steps around the ambiguous multi-collation case and avoids leaning on which collation happens to be first.

On the second one, `QueryService` stays safe. It wraps the plan in a `LogicalSystemLimit` for the query-size limit before reading the collation, and `LogicalSystemLimit.create` already folds any composite into one canonized collation first.

```java
List<RelCollation> collations = input.getTraitSet().getTraits(RelCollationTraitDef.INSTANCE);
RelCollation collation = collations == null ? null : collations.get(0);
collation = RelCollationTraitDef.INSTANCE.canonize(collation);
RelTraitSet traitSet = input.getTraitSet().replace(Convention.NONE).replace(collation);
```

So the trait set it reads back holds a single collation and `getCollation()` there stays within one value. `UnifiedQueryPlanner` reads the collation straight off the raw analyzed plan before any `LogicalSystemLimit` is added, which is why it meets the raw composite and throws. So that copied line is fine as it stands in `QueryService`.

PR #5656, opened 2026-07-27
Add a generic, extensible rest-endpoint provider SPI

noCharger:
### Description

Rebuild of the reverted PPL `rest` command (#5599, reverted in #5635) as a **separated, generic, extensible framework**. `rest <name>` exposes read-only, fixed-schema management endpoints as a relational row source behind one operator allow-list, so results feed straight into `where` / `stats` / `head`.

**Key design points**

- **Generic endpoint SPI.** A thin new `ppl-rest-spi` module exposes `RestEndpointProvider` (an `ExtensiblePlugin` `loadExtensions` SPI) plus `RestEndpointDefinition` / `RestEndpointHandler` / `RestEndpointContext` and `Column` / `ColumnType` / `ArgSpec`. New endpoints are registrations, not new grammar keywords — the language stays a single generic `rest <name>` command.
- **Built-in is a uniform SPI client.** `/_cluster/health` ships via `CoreEndpointsProvider`, which fetches through the SPI context's transport node client (`ctx.client()`) exactly like an external plugin — there is no privileged in-repo path. `ColumnType` keeps the SPI module free of any dependency on the sql engine's `ExprType`.
- **Allow-list choke point.** `plugins.ppl.rest.allowed_endpoints` is the single `resolve()` gate; it defaults to `["/_cluster/health"]` (health on, every other endpoint off until an operator opts in). Health stays gated by the cluster's own `cluster:monitor/health` permission.
- **Lazy relational scan.** The handler runs at scan execution, never at planning, so `EXPLAIN` is side-effect free and `where` / `head` / `stats` compose and push down. `RestRequest` mirrors the system-index `OpenSearchSystemIndexScan` behind one shared catalog enumerator.
- **Central redaction seam.** Rows pass through a single `Redactor` (`(endpoint, row) -> row`) at the choke point before coercion; the default is `Redactor.NONE` (no-op). Redaction lives at the platform choke point rather than in the endpoint SPI.
- **Scope.** This PR adds the framework plus the `/_cluster/health` reference endpoint; further endpoints are left to follow-up PRs.

### Related Issues

Rebuild of #5599 (reverted in #5635) as a separated, extensible framework.

### How it works: `handler` + `ctx` flow (bootstrap to fetch)

Each endpoint's `handler` is captured into its `Endpoint` at bootstrap and travels bundled inside it; the per-query `ctx` (args + transport client) is assembled at `search()`; the two meet at `Endpoint.toRows` -> `handler.fetch(ctx)`. The registry and redactor cross the bootstrap-to-query Guice gap through static holders.

```mermaid
flowchart TB
  subgraph BOOT["Phase 1 · Plugin bootstrap (once per node; loadExtensions runs before the Guice injector)"]
    direction TB
    prov["RestEndpointProvider SPI<br/>CoreEndpointsProvider.getEndpoints()<br/>→ RestEndpointDefinition(name, schema, argSpec, handler)"]
    load["SQLPlugin.loadExtensions<br/>loader.loadExtensions(RestEndpointProvider.class)"]
    build["publishRestCommandRegistries()<br/>new RestEndpointRegistry(providers)<br/>each Definition → new Endpoint(def)<br/>★ Endpoint.handler = def.handler()"]
    hold["RestEndpointRegistryHolder.set(registry)<br/>RedactorHolder.set(Redactor.NONE)"]
    prov -->|discovered| load
    load -->|providers| build
    build -->|publish| hold
  end
  subgraph QUERY["Phase 2 · Query execution (per query: rest '/_cluster/health' ...)"]
    direction TB
    gt["OpenSearchStorageEngine.restTable()<br/>registry = RestEndpointRegistryHolder.get()<br/>redactor = RedactorHolder.get()"]
    src["new RestCatalogSource(registry, spec, client, redactor)<br/>★ endpoint = registry.resolve(spec.endpoint) «allow-list gate»"]
    scan["CalciteEnumerableCatalogScan<br/>source.createRequest()"]
    req["RestCatalogSource.createRequest()<br/>→ new RestRequest(client, endpoint, spec, redactor)"]
    en["OpenSearchCatalogEnumerator<br/>request.search()"]
    se["RestRequest.search()<br/>★ ctx = RestEndpointContext.of(spec.args, nodeClient)<br/>endpoint.toRows(ctx, redactor)"]
    tr["Endpoint.toRows(ctx, redactor)<br/>★★ handler.fetch(ctx) «handler + ctx meet»<br/>→ redactor.redact(endpoint, raw) → coerce → rows"]
    gt --> src
    src --> scan
    scan --> req
    req --> en
    en --> se
    se --> tr
  end
  hold -.->|"registry / redactor (static bridge: bootstrap → query)"| gt
  build ==>|"handler travels bundled inside Endpoint"| tr
```

### Testing

- Unit tests green on JDK 21: rest endpoint registry, catalog source, cross-plugin extensibility, the centralized `Redactor` seam, storage engine, and settings.
- `CalcitePPLRestIT` and `RestCommandSecurityIT` cover `/_cluster/health` plus the allow-list and per-arg gates.

### Check List
  - [x] New functionality has a user manual doc added (`docs/user/ppl/cmd/rest.md`).


core/src/main/java/org/opensearch/sql/utils/SystemIndexUtils.java
```diff
@@ -34,6 +37,121 @@ public static Boolean isSystemIndex(String indexName) {
...
+    }
+    byte[] bytes = new byte[h.length() / 2];
+    for (int i = 0; i < bytes.length; i++) {
+      bytes[i] =
+          (byte)
+              ((Character.digit(h.charAt(2 * i), 16) << 4)
+                  + Character.digit(h.charAt(2 * i + 1), 16));
+    }
+    return new String(bytes, StandardCharsets.UTF_8);
+  }
```
songkant-aws, line 129: Could we leverage java.util.HexFormat?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestEndpointRegistry.java
```diff
@@ -0,0 +1,251 @@
...
+        }
+        m.put(definition.name(), new Endpoint(definition));
+      }
+    }
+    this.registry = m;
+  }
+
+  public java.util.Set<String> endpointNames() {
+    return registry.keySet();
+  }
```
songkant-aws: Nit: Consider avoiding overly small helper methods, especially when they are used only in unit tests.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchStorageEngine.java
```diff
@@ -35,10 +44,37 @@ public Collection<FunctionResolver> getFunctions() {
...
+  private Table restTable(String name) {
+    RestSpec spec = decodeRestSpec(name);
+    RestEndpointRegistry registry = RestEndpointRegistryHolder.get();
+    if (registry == null) {
+      throw new IllegalStateException(
+          "the rest command endpoint registry is not initialized on this node");
+    }
+    registry.resolve(spec.getEndpoint());
+    List<String> allowed = settings.getSettingValue(Settings.Key.PPL_REST_ALLOWED_ENDPOINTS);
+    if (allowed == null || !(allowed.contains("*") || allowed.contains(spec.getEndpoint()))) {
```
songkant-aws: Should "\*\" apply to extension-provided endpoints? With the extensible registry, a cluster configured with ["\*\"] will automatically enable endpoints added by a newly installed or upgraded RestEndpointProvider, without an explicit allow-list change. This seems to weaken the endpoint-by-endpoint opt-in and security-review boundary. Could "*" be limited to built-in endpoints, with external endpoints requiring explicit names?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestEndpointRegistry.java
```diff
@@ -0,0 +1,251 @@
...
+  public RestEndpointRegistry(List<RestEndpointProvider> providers) {
+    Map<String, Endpoint> m = new LinkedHashMap<>();
+    for (RestEndpointProvider provider : providers) {
+      for (RestEndpointDefinition definition : provider.getEndpoints()) {
+        if (m.containsKey(definition.name())) {
+          LOG.warn(
+              "rest endpoint [{}] is already registered; ignoring the duplicate from provider [{}]",
+              definition.name(),
+              provider.getClass().getName());
+          continue;
```
songkant-aws, line 69: Endpoint names are global, while the operator allow-list identifies only the endpoint name and not the contributing provider. If two external providers register the same name, the current first-wins behavior binds that allow-list entry to whichever provider is loaded first. Was this ambiguity considered? Would it be safer to disable only the conflicting endpoint, while allowing the node and the remaining endpoints to start normally?

ahkcs (approved):

ppl-rest-spi/README.md
```diff
@@ -0,0 +1,80 @@
+# ppl-rest-spi
+
+The extension point for the PPL `rest` command. Any OpenSearch plugin can contribute read-only,
+fixed-schema management endpoints that become queryable as `rest '<name>'`, without touching the
+PPL grammar and without depending on the sql plugin's internals. The built-in `/_cluster/health`
+endpoint is just one client of this same contract.
```
penghuo: Could u make the statement more generic. It is not just extension point for the PPL. This solution provide a generic a way to let other plugin to integrate PPL without introduce new command. 

`The built-in `/_cluster/health` endpoint is just one client of this same contract.` this is just example use case. mention it in example section.

ppl-rest-spi/README.md
```diff
@@ -0,0 +1,80 @@
...
+| `RestEndpointHandler` | `List<Map<String, Object>> fetch(RestEndpointContext)`, run at scan execution (not planning), so the scan is lazy and `EXPLAIN` is side-effect free. |
+| `RestEndpointContext` | The validated `args()` plus an optional core `NodeClient` (`client()`) for a handler that issues its own read-only transport action. |
+
+## Add an endpoint
+
+1. Depend on this module `compileOnly`, and declare the sql plugin as an extended plugin so
+   `loadExtensions` discovers your provider:
+
+   ```gradle
+   dependencies { compileOnly project(':ppl-rest-spi') }   // or the published artifact
```
penghuo: It should be clear step by step guide. 
Other plugins need published artifact, right?

ppl-rest-spi/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+   ```yaml
+   plugins.ppl.rest.allowed_endpoints: ["/_cluster/health", "/_my/thing"]
+   ```
+
+Then `rest '/_my/thing' verbose=true | where count > 0 | stats sum(count)` composes like any scan.
+
+## Rules and guarantees
+
+- Endpoints are read-only. The handler produces rows; the sql side coerces each value to the
```
penghuo: sql -> ppl

ppl-rest-spi/README.md
```diff
@@ -0,0 +1,80 @@
...
+  declared `ColumnType` and rejects an uncoercible value with a clear error.
+- Endpoint names are global. If two providers register the same name, the first registration wins
```
penghuo: In case all the endpoint is regestered at plugins.ppl.rest.allowed_endpoints. It should no conflict, right?

ppl-rest-spi/README.md
```diff
@@ -0,0 +1,80 @@
...
+  and the duplicate is ignored with a logged warning.
+- An endpoint must be both registered by a provider and present in `allowed_endpoints`; anything
+  else is rejected before any transport call.
+- Redaction of sensitive values is applied centrally at the row-shaping choke point; a provider
```
penghuo: what does this mean?

ppl-rest-spi/README.md
```diff
@@ -0,0 +1,80 @@
...
+| Type | Role |
+|---|---|
+| `RestEndpointProvider` | Your entry point: `List<RestEndpointDefinition> getEndpoints()`. |
+| `RestEndpointDefinition` | One endpoint: `name()`, `schema()`, `argSpec()`, `handler()`. Build with `RestEndpointDefinition.builder()`. |
+| `Column` / `ColumnType` | A fixed output column. `ColumnType` is `STRING`, `INTEGER`, `LONG`, `DOUBLE`, or `BOOLEAN`. The schema is pinned before execution. |
+| `ArgSpec` | The query args the endpoint accepts and each arg's allowed value domain; an unknown or out-of-domain arg is rejected. `ArgSpec.NONE` accepts none. |
+| `RestEndpointHandler` | `List<Map<String, Object>> fetch(RestEndpointContext)`, run at scan execution (not planning), so the scan is lazy and `EXPLAIN` is side-effect free. |
+| `RestEndpointContext` | The validated `args()` plus an optional core `NodeClient` (`client()`) for a handler that issues its own read-only transport action. |
+
+## Add an endpoint
```
penghuo, line 80: Love this on boarding section.

ppl-rest-spi/src/main/java/org/opensearch/sql/spi/rest/Redactor.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * The single redaction hook for the {@code rest} command: given one raw response row (column name
+ * to value) for an endpoint, it returns the row to emit with sensitive values masked in one pass.
+ * Not a {@code loadExtensions} SPI (one central owner, not an open registration surface); the
+ * default is {@link #NONE} (no-op), applied once per row at the choke point before values are
+ * coerced to the schema types.
+ */
+@FunctionalInterface
+public interface Redactor {
```
penghuo: extension Plugin can control the response in handler, right? if yes, does Redactor required?

core/src/main/java/org/opensearch/sql/utils/SystemIndexUtils.java
penghuo: +1

integ-test/build.gradle
```diff
@@ -419,6 +423,8 @@ testClusters {
         plugin(getArrowFlightRpcPlugin())
         plugin(getAnalyticsEnginePlugin())
         plugin ":opensearch-sql-plugin"
+        // Composite-default cluster: PPL queries route to the analytics engine unless excluded.
+        setting 'cluster.pluggable.dataformat', 'composite'
```
penghuo, line 427: This PR should not releated to analytics engine?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchStorageEngine.java
```diff
@@ -35,10 +44,37 @@ public Collection<FunctionResolver> getFunctions() {
...
       return new OpenSearchIndex(client, settings, name);
     }
   }
+
+  private Table restTable(String name) {
+    RestSpec spec = decodeRestSpec(name);
+    RestEndpointRegistry registry = RestEndpointRegistryHolder.get();
+    if (registry == null) {
+      throw new IllegalStateException(
+          "the rest command endpoint registry is not initialized on this node");
```
penghuo: this is not good user facing error.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchStorageEngine.java
```diff
@@ -35,10 +44,37 @@ public Collection<FunctionResolver> getFunctions() {
...
+    } else if (isSystemIndex(name)) {
+      return new OpenSearchCatalogTable(new SystemIndexCatalogSource(client, name), settings);
     } else {
       return new OpenSearchIndex(client, settings, name);
     }
   }
+
+  private Table restTable(String name) {
+    RestSpec spec = decodeRestSpec(name);
+    RestEndpointRegistry registry = RestEndpointRegistryHolder.get();
```
penghuo, line 56: nit: Why use static class? It is hard to mock and test. Does singleton better?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/CoreEndpointsProvider.java
```diff
@@ -0,0 +1,84 @@
...
+                    Column.of("cluster_name", STRING),
+                    Column.of("status", STRING),
+                    Column.of("number_of_nodes", INTEGER),
+                    Column.of("number_of_data_nodes", INTEGER),
+                    Column.of("active_primary_shards", INTEGER),
+                    Column.of("active_shards", INTEGER),
+                    Column.of("relocating_shards", INTEGER),
+                    Column.of("initializing_shards", INTEGER),
+                    Column.of("unassigned_shards", INTEGER),
+                    Column.of("timed_out", BOOLEAN)))
```
penghuo: It may not required to expose the schema. single string column should be fine. Then use spath command to extract.

ppl-rest-spi/README.md
penghuo: @vamsimanohar  please take a look onboarding section and SPI part.

penghuo (changes requested):

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchStorageEngine.java
noCharger: Removed the "*" wildcard entirely. `plugins.ppl.rest.allowed_endpoints` is now the single source of truth

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/rest/RestEndpointRegistry.java
noCharger: Replaced first-wins with a safer policy: when two external providers register the same endpoint name, both are disabled and a WARN is logged. Fixed in 31e529750.

ppl-rest-spi/README.md
noCharger: same as https://github.com/opensearch-project/sql/pull/5656#discussion_r3705010810

ppl-rest-spi/src/main/java/org/opensearch/sql/spi/rest/Redactor.java
noCharger: removed Redactor abstraction, so it can be defined as part of the handler logic

integ-test/build.gradle
noCharger: It is required for `AnalyticsEngineCompatIT` running on a AE-enabled cluster to assert that rest still routes through Calcite.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchStorageEngine.java
noCharger: It's already a lazy-init singleton. It can't be constructor-injected because the registry is built from `loadExtensions` during node bootstrap, before the Guice injector exists; this mirrors the existing `AnalyticsExecutorHolder`. It stays testable via set(...) (see OpenSearchStorageEngineTest).

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/OpenSearchStorageEngine.java
noCharger: removed this dead code

ppl-rest-spi/README.md
noCharger: updated the maven artifact publish

penghuo (approved):
Thanks!

Issue #5610, opened 2026-07-06
[FEATURE] Define a schema-conflict and mismatched-data handling policy for multi-index queries


dai-chen:
**Is your feature request related to a problem?**

Querying across multiple indices (e.g. a wildcard index pattern like `source=logs-*`) has no consistent policy for handling two related but distinct problems:

1. **Schema-merge-time conflicts**: the same field name has genuinely incompatible mapping types across backing indices (e.g. `object` in some, scalar `text` in others). `MergeRuleHelper` currently resolves conflicts via an ordered rule chain:
    - `DeepMergeRule` only merges struct-vs-struct sub-properties.
    - `TextKeywordConflictRule` only resolves scalar-vs-scalar (text/keyword) conflicts.
    - Any other mismatch falls through to `LatestRule`, which silently picks whichever backing index's mapping was processed last, with no validation, error or coercion.
2. **Execution-time mismatched data**: even once a unified schema is settled on, an individual document's actual value can still fail to match the declared type for that field — not because the document is malformed, but because it came from a differently-mapped index (e.g. dynamic mapping drift across backing indices, or one index legitimately storing the field as a plain string while another stores it as a nested object). `OpenSearchExprValueFactory` blindly casts each document's raw value to the declared type with no defensive handling, so a document from a mismatched-mapping shard throws:
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.util.Map
  at org.opensearch.sql.opensearch.data.utils.ObjectContent.map(ObjectContent.java:81)
  at org.opensearch.sql.opensearch.data.value.OpenSearchExprValueFactory.parseStruct(OpenSearchExprValueFactory.java:376)
  at org.opensearch.sql.opensearch.data.value.OpenSearchExprValueFactory.parse(OpenSearchExprValueFactory.java:218)
  at org.opensearch.sql.opensearch.data.value.OpenSearchExprValueFactory.construct(OpenSearchExprValueFactory.java:189)
  at org.opensearch.sql.opensearch.response.OpenSearchResponse.lambda$handleAggregationResponse$2(OpenSearchResponse.java:263)
  at org.opensearch.sql.opensearch.storage.scan.OpenSearchIndexEnumerator.moveNext(OpenSearchIndexEnumerator.java:139)
```

**What solution would you like?**

[...]

PR #5657, opened 2026-07-27
Add an opt-in partial-result mode for aggregations on text/keyword mapping conflicts

ahkcs:
### Description

On the Calcite PPL path, an aggregation grouped on a field that is mapped **`keyword` in some indices of a wildcard pattern and `text` in others** cannot use native pushdown. The multi-index type merge collapses the field to `text`-without-`.keyword`, which has no doc values, so the aggregation runs as a per-document `_source` script over **every document** — correct, but a full-index scan that is orders of magnitude slower on a wide pattern.

This PR adds an **opt-in mode** that returns a fast, *partial* answer instead: it aggregates over only the subset of indices where the field is natively aggregatable (`keyword`) and attaches a warning naming the ones it excluded.

So the choice becomes **complete-but-slow** (default) vs **fast-but-partial** (opt-in) — both correct, differing in coverage and speed.

### How it works

1. **Warning channel.** Successful PPL JSON responses gain an optional `warnings: [{type, message, detail}]` array, emitted only when non-empty (existing responses are byte-for-byte unchanged):
   ```json
   "warnings": [{
     "type": "PARTIAL_RESULT",
     "message": "Results exclude 1 of 2 indices due to a text/keyword mapping conflict on [applicationid].",
     "detail": "[applicationid] is not mapped as keyword in every queried index, so these indices were excluded from the aggregation: [logs-text]. Map [applicationid] as keyword across all indices to include them."
   }]
   ```
2. **Partial-result plan** (`PartialResultAggregatePushdown`). When the mode is on and the group key is a text/keyword conflict, the scan is narrowed to the aggregatable index subset and the aggregation pushed down over just that subset (`size = 0`, no PIT). The partitioning logic is unit-tested in isolation.
3. **Per-request override.** A `partial_result` boolean in the query body (mirroring `profile`) overrides the cluster setting for one query; absent → cluster setting decides.

### Behavior

Cluster setting `plugins.query.partial_result.on_mapping_conflict.enabled` (default `false`):

| Query | Off (default) | On |
| --- | --- | --- |
| `stats count() by <conflict field>` | complete result, slow (`_source` scan of all docs) | fast result over the `keyword` subset + `PARTIAL_RESULT` warning |
| no-conflict / single-index aggregation | complete, no warning | complete, no warning (unchanged) |
| any of the above with `format=csv` | as above | falls through to the complete result (CSV has no warning channel) |

### Key points

- **Opt-in, default off.** A partial result is knowingly incomplete, so it never happens silently; with the setting off the change is behavior-preserving.
- **Never degrades silently.** Only the JSON shape carries `warnings`, so CSV/RAW/VIZ fall through to the complete result rather than dropping data unannounced.
- **Deterministic selection.** Keep the `keyword` group whenever one exists; otherwise the `text`-with-`.keyword` group; always exclude bare `text`. The result never depends on how many indices of each type match.
- **Calcite PPL path only**; V2/legacy untouched.

**Not in scope:** recovering an *excluded but aggregatable* group (`text`-with-`.keyword` alongside a `keyword` group) — that needs a per-group split-and-union, a larger separate change. This is why the warning recommends mapping the field as `keyword` everywhere.

### Check List
- [x] New functionality includes testing (unit + integration).
- [x] New functionality has been documented (`docs/user/admin/settings.rst`).


anasalkouz:
1. Is this only applicable for non-mustang?
2. Is this only limited to text vs keyward use-case? can we extend the scope?
3. Shall we have a role on the inspect query feature to suggest customer to enable this parital result flag to optimize performance if the query fails to push down?
4. Can we have performance benchmark for the 3 cases? with no pushdown, with text pushdown, and with partial results?

ahkcs:
## Performance benchmark: partial results vs. today vs. scripted text pushdown (#5646)

Comparing three responses to the mapping-conflict PIT-exhaustion case (an aggregation groups on a field mapped `keyword` in some indices of a wildcard pattern and `text` in others):

- **A — today.** The type merge collapses the field to `text`-without-doc-values, aggregate pushdown is lost, and the engine scans every document per shard — opening a Point-In-Time (PIT) context on every shard — and aggregates client-side. Complete answer; trips `search.max_open_pit_context` on wide patterns.
- **B — scripted text pushdown (#5646).** Routes the group key through a Calcite `_source` script pushed down with `size=0`. No PIT. Complete answer.
- **C — partial results (this PR).** Narrows the scan to the aggregatable (`keyword`) subset, pushes down natively (`size=0`, no PIT), returns a partial answer plus a `PARTIAL_RESULT` warning naming the excluded indices.

These do **not** compute the same thing, so every latency figure is paired with a completeness column.

### Test setup

| | |
|---|---|
| Cluster | Single node, OpenSearch `3.8.0-SNAPSHOT`, 2 GB heap (raised from the 512 MB dev default so A fails on PIT, not the query memory circuit breaker) |
| Engine | Calcite path (`plugins.calcite.enabled=true`) — the only path in scope for this PR |
| A & C | Same build (this PR); A = `partial_result:false`, C = `partial_result:true` |
| B | #5646 build, separate run on identical seeded data |
| Iterations | 30 measured + 5 warmup per query, serial (clean per-query latency + exact PIT deltas) |
| Latency | Client-side wall-clock of the `_plugins/_ppl` call |
| PIT/query | Delta of the cumulative `point_in_time_total` node stat |

**Datasets** (deterministic, seed = 42):
- **wide** — 40 `keyword` + 4 bare-`text` indices, 2 shards each (**88 shards**), 5,000 docs/index (220,000 total). A wide wildcard pattern where 88 shards exceeds any realistic PIT limit.
- **small** — 1 `keyword` + 1 `text`, 1 shard each, 20,000 docs/index. Control below the PIT limit.
- **flat** — same as small but the conflict field is top-level (`appid`), not nested. (See the note on B.)

Conflict field for wide/small is a nested `resource.attributes.applicationid`; for flat it is top-level `appid`.

### Latency p50 / p90 / p99 (ms)

"Today" (A) has two modes on the same query, decided by whether the shard count exceeds `search.max_open_pit_context`:

| Query | A: PIT opened, **under** limit (no 500) | A: PIT limit **exceeded** | B: #5646 (script) | C: partial (this PR) | Completeness of C | PIT/query (A) |
|---|---|---|---|---|---|---|
| `wide` stats  | 441 / 463 / 494 | **FAIL — 500** | 112 / 150 / 271 | **11 / 12 / 13** | 91.2% (200k/219k) | **88** |
| `wide` top    | 439 / 455 / 486 | **FAIL — 500** | 123 / 149 / 293 | **16 / 18 / 21** | 91.2% | **88** |
| `small` stats | 90 / 110 / 115  | 92 / 111 / 114 | 32 / 36 / 43    | **5 / 6 / 7**    | 50% (20k/40k) | 2 |
| `small` top   | 96 / 116 / 122  | 94 / 102 / 113 | 37 / 44 / 53    | **11 / 14 / 15** | 50% | 2 |
| `flat` stats  | 58 / 78 / 82    | 56 / 63 / 75   | 20 / 23 / 42    | **4 / 5 / 5**    | 50% (20k/40k) | 4 |
| `flat` top    | 60 / 67 / 86    | 57 / 62 / 80   | 27 / 30 / 31    | **10 / 12 / 14** | 50% | 4 |

The "under limit" column used `max_open_pit_context=500`; "exceeded" used `=10`. A only fails where shard count crosses the limit (wide, 88 shards). Small/flat stay under and complete — but still open PITs and run 8–20× slower than C. Error rate in the exceeded regime: A = 100% on wide, C = 0% everywhere (never opens a PIT).

### Completeness (sum of `count()` across all buckets)

| Dataset | A (complete) | B | C (partial) | C completeness |
|---|---|---|---|---|
| wide, nested field  | 219,328 | 220,000 but **1 `null` bucket** (grouping lost) | 200,000 | **91.2%** |
| small, nested field | 40,000  | 40,000 but **1 `null` bucket** | 20,000 | **50%** |
| flat field          | 40,000  | 40,000, **50 buckets (correct)** | 20,000 | **50%** |

### Why the latencies differ (mechanism)

A leaves the aggregate *above* the scan (`explain` shows `requestedTotalSize=2147483647`): every matching document is streamed out of every shard over PIT cursors into the coordinator JVM and counted there — cost scales with **document count**. B and C fuse the aggregate into the scan (`size=0`), so the count runs inside each shard and only bucket results cross the wire — cost scales with **bucket count**, and no PIT is opened. B groups on a per-document `_source` script; C groups on native `keyword` doc values, which is why C stays ~2–5× ahead of B even where both push down.

### Takeaways

1. **When it runs, C is fastest** (~34× vs A, ~10× vs B on wide) — but that speed *is* the partial answer: it excludes the non-aggregatable indices. On wide that is an 8.8% undercount; where the text indices hold half the data, 50%. Always accompanied by the `PARTIAL_RESULT` warning.
2. **In the low-PIT-budget regime, A fails outright** (100% errors on wide). B and C never open a PIT.
3. **B is complete and PIT-free on flat fields** and is the natural default there. On the **nested** dotted field, B in its current state grouped all documents into a single `null` bucket (complete count, grouping lost) — worth verifying whether the scripted `_source` reader resolves nested dotted paths. This PR's producer resolves the nested path.

C is intended as an opt-in escape hatch (default off) for the widest patterns / lowest PIT budgets where a knowingly-partial, clearly-warned answer is preferable to a slow scan or a 500 — complementary to, not competing with, a complete-answer pushdown fix.

<sub>Single-node, laptop-scale absolutes; the ratios and the PIT / completeness / error-rate columns are the transferable results.</sub>

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
```diff
@@ -44,6 +44,8 @@ public enum Key {
     CALCITE_PUSHDOWN_ROWCOUNT_ESTIMATION_FACTOR(
         "plugins.calcite.pushdown.rowcount.estimation.factor"),
     CALCITE_SUPPORT_ALL_JOIN_TYPES("plugins.calcite.all_join_types.allowed"),
+    CALCITE_PARTIAL_RESULT_ON_MAPPING_CONFLICT(
+        "plugins.calcite.partial_result.on_mapping_conflict.enabled"),
```
penghuo: calcite is internal impl. plugins.query.partial_result.on_mapping_conflict.enabled is better.

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/AggregateIndexScanRule.java
```diff
@@ -98,6 +98,12 @@ protected void apply(
       @Nullable LogicalProject project,
       CalciteLogicalIndexScan scan) {
     AbstractRelNode newRelNode = scan.pushDownAggregate(aggregate, project);
+    if (newRelNode == null) {
+      // Normal pushdown failed (e.g. a text/keyword mapping conflict across a wildcard pattern that
+      // would otherwise fall back to a per-shard scan and exhaust PIT contexts). If partial results
+      // are enabled, push the aggregation over the aggregatable index subset instead and warn.
+      newRelNode = scan.tryPartialResultAggregate(aggregate, project);
```
penghuo: why add a new function instead of reuse pushDownAggregate?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -431,6 +434,87 @@ public AbstractRelNode pushDownAggregate(Aggregate aggregate, @Nullable Project
...
+      return null;
+    }
+    try {
+      List<String> outputFields = aggregate.getRowType().getFieldNames();
+      List<String> bucketNames = outputFields.subList(0, aggregate.getGroupSet().cardinality());
+      // getIndexNames() returns the raw pattern (e.g. ["logs-*"]) for a wildcard, not the resolved
+      // concrete indices; getIndexMappings expands it server-side to a map keyed by concrete index
+      // name, which is what the partitioning classifies.
+      Map<String, IndexMapping> mappings =
+          osIndex.getClient().getIndexMappings(osIndex.getIndexName().getIndexNames());
```
penghuo: avoid call indexMapping again.

ahkcs:
> 1. Is this only applicable for non-mustang?
> 2. Is this only limited to text vs keyward use-case? can we extend the scope?
> 3. Shall we have a role on the inspect query feature to suggest customer to enable this parital result flag to optimize performance if the query fails to push down?
> 4. Can we have performance benchmark for the 3 cases? with no pushdown, with text pushdown, and with partial results?

1. Yes, currently it's only applicable for Calcite path.
2. Today it's deliberately scoped to the text/keyword conflict, it can be extended, and the shape generalizes cleanly if we have more partial result use cases.
3. We can add that recommendation/suggestion
4. link for performance benchmarking: https://github.com/opensearch-project/sql/pull/5657#issuecomment-5111106792

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
penghuo: We should have multiple settings?

1. Does mapping conflict allowed?
* Yes (default, PPL current behavior). I suspect text/keyword conflict not the only case. 
* No: Fast Fail.

2. Does aggregation on text field allowed?
* Yes: Use script text aggregation.
* No: There are multiple choice. Ignore index has text field or Fast Fail or use null value instead of fetch from text field.

ahkcs: Thanks for the suggestion, updated the naming

### Current behavior isn't uniform

`MergeRuleHelper` tries three rules, first match wins:

| Conflict                                                   | Rule                      | Today                                                                                                                                |
| ---------------------------------------------------------- | ------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
| STRUCT/ARRAY, same core type                               | `DeepMergeRule`           | real merge, correct                                                                                                                  |
| text ↔ keyword                                             | `TextKeywordConflictRule` | collapses to `text`-without-`.keyword` → no doc values → no pushdown → PIT per shard → **500** past the limit                        |
| **everything else** (long↔integer, ip↔keyword, date↔long…) | `LatestRule`              | `target.put(key, source)` while iterating a `toUnmodifiableMap` — **the last mapping yielded wins, and that order isn't guaranteed** |

So you're right that text/keyword isn't the only case. For the rest, the merged type is **arbitrary per index set** — stable for a given set, but two identically-shaped sets can disagree. Verified on 12 freshly-created, identical `long`+`integer` conflicts:

```
describe        → 10 sets resolved to `int`, 2 to `bigint`
docs 5000000000 + 7:
  int sets      → [7, 705032704]      ← 64-bit value truncated, 200 OK, no warning
  bigint sets   → [7, 5000000000]     ← correct
```

`long`+`integer` → `int` is *narrowing*, so this isn't least-restrictive widening. Filtering is honored correctly, so it's a type-resolution defect, not a predicate bug. Its own issue, independent of this PR.

### Settings 
1. **`allow` needs defining.** "Current behavior" means three different things above. It should mean *merge **and warn***, not "merge and sometimes silently truncate" — making this PR's warning channel the substrate for the whole matrix. Fast-fail is genuinely useful: for a `LatestRule` conflict, failing loudly beats silent truncation.

### Proposal: make it discoverable, not a flag users must know

- **Backend `analyze`** — detect the failed-pushdown signature (aggregate above the scan, `requestedTotalSize=2147483647`) and emit a recommendation naming the field, the conflict, and the remedy. `AnalyzeResponse` already has a `recommendations` field, so this is additive.
- **Frontend `inspect query`** — render it with the honest trade-off (faster, no PIT, *partial*).

### Scope

**This PR:** warning channel + `PARTIAL_RESULT` producer + single opt-in boolean (default off, inert)

**Follow-ups:** (1) `analyze` recommendation + `inspect query`; (2) `LatestRule` arbitrary-type/truncation bug;

opensearch/src/main/java/org/opensearch/sql/opensearch/planner/rules/AggregateIndexScanRule.java
ahkcs: Updated, rule is back to one call

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
ahkcs: Updated — now `osIndex.getIndexMappings()`, reusing the fetch getFieldTypes() already makes. No second call.

common/src/main/java/org/opensearch/sql/common/setting/Settings.java
dai-chen: @ahkcs Just wonder is this solving same problem as https://github.com/opensearch-project/sql/issues/5610 or different? Is this also supposed to be part of https://github.com/opensearch-project/sql/issues/4349 instead of handling case by case?

ahkcs: Hi Chen, for this PR, it's solving one of the problems in #5610 (merge rule). And the optimization in this PR is mainly for performance, which is a different path from #4349 since 4349 is about tolerance and this PR is about improving performance

docs/user/admin/settings.rst
```diff
@@ -236,6 +236,53 @@ Result set::
       }
     }
 
+plugins.query.partial_result.on_mapping_conflict.enabled
```
penghuo: Mark as experimental.

docs/user/admin/settings.rst
```diff
@@ -236,6 +236,53 @@ Result set::
...
+========================================================
+
+Version
+-------
+3.8
```
penghuo: since 3.9

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -428,7 +448,81 @@ public AbstractRelNode pushDownAggregate(Aggregate aggregate, @Nullable Project
         LOG.debug("Cannot pushdown the aggregate {}", aggregate, e);
       }
     }
-    return null;
+    return allowPartialFallback ? tryPartialResultAggregate(aggregate, project) : null;
```
penghuo: Question, Any case worth to try?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -428,7 +448,81 @@ public AbstractRelNode pushDownAggregate(Aggregate aggregate, @Nullable Project
...
+   * e.g. an OpenSearch Dashboards toggle) takes precedence when present; otherwise the cluster
+   * setting decides.
+   */
+  private boolean isPartialResultEnabled() {
+    Boolean override = QueryContext.getPartialResultOverride();
+    if (override != null) {
+      return override;
+    }
+    return (Boolean)
+        osIndex.getSettings().getSettingValue(Settings.Key.PARTIAL_RESULT_ON_MAPPING_CONFLICT);
```
penghuo: Consolidate setting and query paramater could in single method. QueryContext.isPartialResultEnabled

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -428,7 +448,81 @@ public AbstractRelNode pushDownAggregate(Aggregate aggregate, @Nullable Project
...
+      Aggregate aggregate, @Nullable Project project, List<String> bucketNames) {
+    if (!isPartialResultEnabled()) {
+      return null;
+    }
+    // A format with no warnings channel (CSV/RAW/VIZ) must not silently drop indices.
+    if (!QueryContext.isWarningsSupported()) {
+      return null;
+    }
+    try {
+      // Reused from the field-type fetch, so no extra round trip.
```
penghuo: Remove this comments.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -397,7 +409,15 @@ public AbstractRelNode pushDownAggregate(Aggregate aggregate, @Nullable Project
         if (LOG.isDebugEnabled()) {
           LOG.debug("Cannot pushdown the aggregate due to bucket contains array (nested) type");
         }
-        return null;
+        return allowPartialFallback ? tryPartialResultAggregate(aggregate, project) : null;
```
penghuo: Question, Any case worth to try?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PartialResultAggregatePushdown.java
```diff
@@ -0,0 +1,162 @@
...
+    // indices of each type match. Never keep both groups: their mix would re-collapse to a
+    // conflict.
+    List<String> keptIndices;
+    if (!keywordIndices.isEmpty()) {
+      keptIndices = keywordIndices;
+      excludedIndices.addAll(textKeywordIndices);
+    } else if (!textKeywordIndices.isEmpty()) {
+      keptIndices = textKeywordIndices;
+    } else {
+      return null; // no aggregatable subset -> partial mode can't help
```
penghuo: What if indices has keyword and int field in aggregation? Indices include int fileld will be excluded?

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
```diff
@@ -428,7 +448,81 @@ public AbstractRelNode pushDownAggregate(Aggregate aggregate, @Nullable Project
...
+   * is aggregatable, push the aggregation over just that subset, and record a warning naming the
+   * excluded indices. Only runs behind the opt-in setting and only when the response format can
+   * carry the warning ({@link QueryContext#isWarningsSupported}); returns {@code null} otherwise.
+   * Partitioning lives in {@link PartialResultAggregatePushdown}.
+   */
+  private AbstractRelNode tryPartialResultAggregate(
+      Aggregate aggregate, @Nullable Project project) {
+    List<String> outputFields = aggregate.getRowType().getFieldNames();
+    return tryPartialResultAggregate(
+        aggregate, project, outputFields.subList(0, aggregate.getGroupSet().cardinality()));
```
penghuo: Does this partial results change support | stats sum(a) by abs(b). where b could be keyword/text.

docs/user/admin/settings.rst
ahkcs: Updated

docs/user/admin/settings.rst
ahkcs: Updated

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
ahkcs: Removed

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
ahkcs: Updated

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
ahkcs: neither adds a case, removed

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
ahkcs: neither adds a case, removed

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PartialResultAggregatePushdown.java
ahkcs: Updated. Before the fix, resolveBucketMapping treated anything that wasn't keyword or text-with-.keyword as `NOT_AGGREGATABLE`, so an int field landed in the excluded bucket.

I added a `CONFLICTING_TYPE` classification for a group field that maps to a non-text type (int, date, boolean, etc.). When any index resolves to that, plan() returns null and partial mode bails — it leaves the query on the normal path rather than excluding the aggregatable index. Only a genuine bare-text field is still excludable, which is the text/keyword collapse this fallback is meant for.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/CalciteLogicalIndexScan.java
ahkcs: Partial results only kicks in for a **bare group field** (`... by b`), not for an expression over one (`... by abs(b)`).

The mechanism is per-index mapping partitioning: `resolveBucketMapping` looks up the group field in each index's `_mapping` to decide keyword vs text. For `abs(b)` the group key is a computed column, not a stored field, so the lookup finds nothing in any index → `plan()` returns `null` → partial mode bails. The query then runs the normal way and returns the **complete** result, with `abs(b)` evaluated per-document (script over `_source`) — no data dropped, no partial-result warning.

So to your example `stats sum(a) by abs(b)` with `b` keyword/text: it's not *handled* by partial results, but it's not broken — you get the full result, just not the fast/partial path. Verified on a keyword+text pair:
- `... by b` → partial result over the keyword subset + warning
- `... by upper(b)` → complete result over all indices, no warning (partial mode declined)

Supporting partial results for an expression key would mean reasoning about whether the expression is computable per index and partitioning on that — a separate, larger change.

penghuo: I do not think the `plugins.query.partial_result.on_mapping_conflict.enabled` setting scope should limited only on raw field on group by list.  The following query should support partial results, right?
```
-- | eval b=abs(bb) | stats sum(a) by b
-- | spath input=body | stats sum(a) by b
```

ahkcs: Broadened it. The partitioning traces each group key through the eval `Project` to the scan fields it reads, so `eval b = lower(city) | stats sum(a) by b` now gets partial results just like `... by city` (multi-field expressions require every field it reads to be aggregatable). ITs added.

```
source=%s | eval g = concat(city, region) | stats count() by g | sort g
```

will return partial results

Two things to note:
 `abs(bb)` needs a numeric `bb`, and a numeric-vs-keyword conflict isn't a text/keyword collapse — that bails by design (excluding an aggregatable numeric index would drop valid data); 
 
Also `spath` is different — `b` is extracted from `_source`, not a mapped field, so there's no per-index mapping to partition on;

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PartialResultAggregatePushdown.java
penghuo: The current logic still narrowed, e.g. it does not cover int with text.
I think rules is "if a index include a non_aggregatable field, ignore that index".

ahkcs: Updated — it partitions by aggregatability now: drop any index where the group field is non-aggregatable (text family incl. `text`+`.keyword`, or absent), keep the aggregatable ones (keyword/numeric/date/boolean/ip) + warn. So **int+text** now keeps int and excludes text.

On the original keyword+int: both are aggregatable, so nothing's dropped — it's not a partial-results case but a genuine type conflict with no lossless merge (LatestRule last-write-wins). After testing, this conflict currently crashes the node / 500s on the normal path (assertion when the arbitrary merged type doesn't fit the values) — verified live, and independent of the values (int-like keyword strings like "5" crash the same as "prod").

Partial-results declines here, so it inherits that pre-existing behavior. It's #5610, out of scope for this PR but worth tracking separately.

Added tests in `PartialResultAggregatePushdownTest` + `CalcitePartialResultOnMappingConflictIT`.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PartialResultAggregatePushdown.java
```diff
@@ -0,0 +1,158 @@
...
+    // merged type is arbitrary -> leave it to the normal path.
+    if (aggregatableGroups.size() != 1) {
+      return null;
+    }
+    if (excludedIndices.isEmpty()) {
+      return null; // homogeneous already -> pushdown would not have failed
+    }
+
+    List<String> keptIndices = aggregatableGroups.values().iterator().next();
+    excludedIndices.sort(null);
```
penghuo: nit: why sort on list? if it is for orderd error message, keep it in when build error message.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PartialResultAggregatePushdown.java
```diff
@@ -0,0 +1,158 @@
...
+      Map<String, OpenSearchDataType> flatMapping, List<String> bucketNames) {
+    List<String> tokens = new ArrayList<>();
+    for (String field : bucketNames) {
+      OpenSearchDataType type = flatMapping.get(field);
+      if (type == null) {
+        return null; // field absent here -> not aggregatable
+      }
+      MappingType mappingType = type.getMappingType();
+      if (mappingType == MappingType.Keyword) {
+        tokens.add("kw");
```
penghuo: "kw" is as same as "t:" + mappingType, not need to different it.

penghuo (approved):

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PartialResultAggregatePushdown.java
ahkcs: moved it into `buildWarning`. `plan()` now leaves `excludedIndices` in insertion order; only the message sorts a copy, since ordering is just for a stable/readable warning

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/scan/PartialResultAggregatePushdown.java
ahkcs: Agreed — dropped the special case.

Issue #4766, opened 2025-11-10
[FEATURE] Add mvcombine command in PPL


srikanthpadakanti:
**Is your feature request related to a problem?**
Currently, OpenSearch SQL (PPL) does not provide a native way to combine identical field values from consecutive rows/events into multivalue fields for downstream analytics, data transformation, and aggregation.

**What solution would you like?**
Introduce an `mvcombine` command to:
- Combine values of a specified field from consecutive rows/events into a multivalue field for each group with identical values in other fields.
- Support an optional delimiter parameter to customize value separation.
- Ensure that only consecutive rows/events with matching non-specified fields are combined.
- Work efficiently and be compatible with streaming/distributed queries.

**What alternatives have you considered?**
- Post-processing outside OpenSearch SQL, which adds overhead and complexity.
- Using array aggregation, scripting, or transforms, which do not guarantee strictly consecutive combination and may be less intuitive.

**Do you have any additional context?**
This capability would enable more flexible analytics, summarization, and event correlation tasks within PPL and support common log/data prep workflows and dashboarding.

Reference documentation: [mvcombine command reference](https://help.splunk.com/en/splunk-cloud-platform/search/search-reference/10.1.2507/search-commands/mvcombine)

---

**Examples:**
1. **Basic Example**
   - Input:
     | user | action | id |
     |------|--------|----|
     | joe  | login  | 1  |
     | joe  | login  | 2  |
     | joe  | logout | 3  |
   - Query: `... | mvcombine id`
   - Output:
     | user | action | id      |
     |------|--------|---------|
     | joe  | login  | [1, 2]  |
     | joe  | logout | [3]     |
2. **Delimiter Option**
   - Query: `... | mvcombine id delimiter=";"`
   - Output:
     | user | action | id     |
     |------|--------|--------|
     | joe  | login  | 1;2    |
     | joe  | logout | 3      |
3. **Non-Consecutive Events (Edge Case)**
   - Only consecutive rows/events with identical values in all other fields are combined. Sorting/grouping may be required to combine non-adjacent matches.
   - Input:
     | user | action | id |
     |------|--------|----|
     | joe  | login  | 1  |
     | sam  | login  | 5  |
     | joe  | login  | 2  |
   - Output:
     | user | action | id   |
     |------|--------|------|
     | joe  | login  | [1]  |
     | sam  | login  | [5]  |
     | joe  | login  | [2]  |

dai-chen:
[...]

PR #5663, opened 2026-07-30
Emit list() for mvcombine so it runs on the analytics engine

noCharger:
### Description

`mvcombine` lowered to Calcite `ARRAY_AGG(<field>) FILTER (<field> IS NOT NULL)`, which the analytics-engine (DataFusion) route cannot execute: `ARRAY_AGG` is not in the engine's aggregate function set, so the plan is rejected during marking. Mapping it into the engine is not viable either — Calcite `ARRAY_AGG` infers a nullable-element `ARRAY` while the engine's `array_agg` infers a NOT NULL element, and the two cannot be reconciled during the aggregate rewrite.

The fix stays on the SQL side: `performArrayAggAggregation` now emits the existing PPL `list()` aggregate (`PPLBuiltinOperators.LIST`), which the analytics engine already supports end to end (it maps to the engine's `array_agg` / `list_merge` operators). `ListAggFunction` filters nulls internally, so the explicit `IS NOT NULL` filter is dropped. No analytics-engine (core) change is required.

**Behavior change (please review):** the combined field is now `ARRAY<VARCHAR>` — `list()` renders elements as strings and retains at most 100 values per group, per its documented contract. For string target fields (the documented `mvcombine` use) output is unchanged; numeric/other targets are now stringified and capped. On the Calcite/Lucene path this differs from the previous unbounded, type-preserving `ARRAY_AGG`, so flagging in case reviewers prefer to gate it.

### Bug reproduce (analytics-engine / DataFusion route)

Composite/parquet index, `analytics.planner.prefer_metadata_driver=false` (forces DataFusion):

```ppl
source=test_data | fields category, status | mvcombine status
```

#### Before (`ARRAY_AGG`) — rejected on the analytics engine

Logical plan:

```
LogicalProject(category=[$0], status=[$1])
  LogicalAggregate(group=[{0}], status=[ARRAY_AGG($1) FILTER $2])
    LogicalProject(category=[$0], status=[$7], $f2=[IS NOT NULL($7)])
      LogicalTableScan(table=[[opensearch, test_data]])
```

Physical plan: none — rejected during analytics-engine marking, before execution.

Output:

```json
{
  "error": {
    "details": "No enum constant org.opensearch.analytics.spi.AggregateFunction.ARRAY_AGG",
  },
  "status": 500
}
```

#### After (`LIST`) — runs on DataFusion

Logical plan:

```
LogicalProject(category=[$0], status=[$1])
  LogicalAggregate(group=[{0}], status=[LIST($1)])
    LogicalProject(category=[$0], status=[$7])
      LogicalTableScan(table=[[opensearch, test_data]])
```

Physical plan (`profile=true`; every node `viableBackends=[[datafusion]]`, stage `chosen_backend=datafusion`):

```
OpenSearchSort(fetch=[10000], viableBackends=[[datafusion]])
  OpenSearchProject(category=[$0], status=[$1], viableBackends=[[datafusion]])
    OpenSearchAggregate(group=[{0}], status=[LIST($1)], mode=[SINGLE], viableBackends=[[datafusion]])
      OpenSearchProject(category=[$0], status=[$7], viableBackends=[[datafusion]])
        OpenSearchTableScan(table=[[test_data]], viableBackends=[[datafusion]])
```

Output:

```
+----------+--------------------------+
| category | status                   |
|----------+--------------------------|
| A        | ["active", "active"]     |
| B        | ["inactive", "active"]   |
| C        | ["pending"]              |
+----------+--------------------------+
```

### Related Issues

Related to #4766 ([FEATURE] Add `mvcombine` command in PPL) — extends it to the analytics-engine (DataFusion) route.

### Check List

  - `CalcitePPLMvCombineTest` logical-plan + Spark SQL expectations updated for the `LIST` aggregate. `CalciteMvCombineCommandIT` is unchanged (it already asserts string-array output on a string field).
  - `docs/user/ppl/cmd/mvcombine.md` gains a Limitations note (string elements, 100-value cap, unordered).

### DIFF
--- a/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/CalciteRelNodeVisitor.java
@@ -73,7 +73,6 @@ import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexVisitorImpl;
 import org.apache.calcite.rex.RexWindowBounds;
 import org.apache.calcite.sql.SqlKind;
-import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.fun.SqlTrimFunction;
 import org.apache.calcite.sql.type.ArraySqlType;
@@ -4498,13 +4497,10 @@ public class CalciteRelNodeVisitor extends AbstractNodeVisitor<RelNode, CalciteP
       RelBuilder relBuilder, int targetIndex, String targetName, List<RexNode> groupExprs) {
 
     final RexNode targetRef = relBuilder.field(targetIndex);
-    final RexNode notNullTarget = relBuilder.isNotNull(targetRef);
-
+    // list() rather than Calcite ARRAY_AGG: list() carries the ARRAY<VARCHAR> contract the
+    // analytics-engine (DataFusion) route supports, and drops nulls itself (no filter needed).
     final RelBuilder.AggCall aggCall =
-        relBuilder
-            .aggregateCall(SqlLibraryOperators.ARRAY_AGG, targetRef)
-            .filter(notNullTarget)
-            .as(targetName);
+        relBuilder.aggregateCall(PPLBuiltinOperators.LIST, targetRef).as(targetName);
 
     relBuilder.aggregate(relBuilder.groupKey(groupExprs), aggCall);
   }
--- a/docs/user/ppl/cmd/mvcombine.md
+++ b/docs/user/ppl/cmd/mvcombine.md
@@ -108,6 +108,10 @@ Error: Query returned no data
 
 =======
 
+## Limitations
+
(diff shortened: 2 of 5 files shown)


noCharger:
Closed in favor of a type-preserving fix on the analytics-engine (DataFusion) side.

This PR emitted the PPL `list()` aggregate, which runs on the DataFusion route but downgrades the combined field to `ARRAY<VARCHAR>` (elements stringified) with a 100-value cap on *all* engines — a semantic regression for non-string fields.

Instead, `mvcombine` keeps emitting the type-preserving Calcite `ARRAY_AGG` (mainline; already correct on Calcite/Lucene), and the DataFusion route is taught to execute it via a dedicated type-preserving array-agg operator (element type = source type, no VARCHAR cast). Java-only in the analytics engine, no Rust change (the `array_agg`/`list_merge` UDAFs are already type-generic).

Verified live on the DataFusion route: `mvcombine` over an integer field returns `ARRAY<INT>` (`[30, 30]`, not `["30","30"]`) with `chosen_backend=datafusion`; `list()`/`values()` remain `ARRAY<VARCHAR>`. That change will be raised as a separate opensearch-project/OpenSearch (core) PR.

PR #5669, opened 2026-08-01
fix: allow LIST/VALUES aggregates to accept DATE/TIMESTAMP/IP/BINARY field types

### LINKED ISSUE #22581: [BUG] analytics-engine LIST/VALUES aggregates reject DATE/TIMESTAMP/IP/BINARY despite allowing them
### Bug

The PPL `LIST` and `VALUES` aggregates reject field types that their own signature lists as allowed. Every call over a `DATE`/`TIMESTAMP`/`IP`/`BINARY` field returns a self-contradictory `400`:

```
"reason": "Invalid Query",
"details": "Aggregation function LIST expects field type
   {[BYTE]|[SHORT]|[INTEGER]|[LONG]|[FLOAT]|[DOUBLE]|[STRING]|[BOOLEAN]|[DATE]|[TIME]|[TIMESTAMP]|[IP]|[BINARY]},
   but got [TIMESTAMP]",
```

`[TIMESTAMP]` is in the allowed set, yet it is rejected. Same for `[DATE]`, `[IP]`, `[BINARY]`.

### Steps to reproduce

1. Index a doc with a date field:
   ```
   PUT /t/_doc/1?refresh
   { "ts": "2004-07-09 10:17:35" }
   ```
2. Run a `list` aggregate over it via the analytics-engine PPL route:
   ```
   POST /_plugins/_ppl
   { "query": "source=t | stats list(ts)" }
   ```
3. Response is `400 Invalid Query` with the `details` shown above. Expected: a list of the field's values.

Or via the sandbox IT:
```
./gradlew ':sandbox:qa:analytics-engine-rest:integTest' \
  --tests 'org.opensearch.analytics.qa.ListAggregateMultiTypeIT.testListBinary'
```

### Failing tests (`:sandbox:qa:analytics-engine-rest:integTest`)

`DatetimeCoverageIT.testListFunctionDateUsesSpace`, `testListFunctionTimestampUsesSpace`; `ListAggregateMultiTypeIT.testListBinary`, `testListDate`, `testListDateNanos`, `testListIp`; `ListAggregateMultiShardIT.testListBinaryTwoShards`, `testListDateTwoShards`, `testListIpTwoShards`; `TwoShardAggregationIT.testReduceCorrectnessAcrossTwoShards`; `TwoShardOversamplingAggregationIT.testReduceCorrectnessAcrossTwoShards`.

### Notes

Fails deterministically on `main` (reproduced on unrelated PRs, e.g. #22553). The `VALUES` variant lists both `[BINARY]` and `[BINARY,INTEGER]` as allowed but still rejects `[BINARY]`, suggesting the operand check compares full type identity rather than the `SqlTypeName`. Likely area: `sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java` / `OpenSearchValuesRule.java`.

waterWang:
### Description

This PR fixes a type-matching bug in the PPL aggregate function validation that caused LIST and VALUES aggregate functions to reject DATE, TIMESTAMP, IP, and BINARY field types, even though these types were explicitly listed in the allowed signatures.

### Root Cause

The `typesMatch()` method in `PPLTypeChecker.java` was returning `false` when one operand was a UDT (via `AbstractExprRelDataType`) and the other was a plain `RelDataType`. Database-sourced fields (e.g., TIMESTAMP from a table) arrive as plain `RelDataType`, but the allowed `SCALAR_TYPES` list in `PPLOperandTypes.java` uses UDT variants for temporal/special types (e.g., `TIMESTAMP_UDT`, `DATE_UDT`, `IP_UDT`, `BINARY_UDT`).

This caused the validation to fail with:
```
Aggregation function LIST expects field type {[BYTE]|[SHORT]|[INTEGER]|[LONG]|[FLOAT]|[DOUBLE]|[STRING]|[BOOLEAN]|[DATE]|[TIME]|[TIMESTAMP]|[IP]|[BINARY]}, but got [TIMESTAMP]
```

### Fix

Changed `typesMatch()` to fall back to `SqlTypeName` comparison when one operand is a UDT and the other is a plain `RelDataType`, instead of returning `false`.

### Related Issues

Fixes opensearch-project/OpenSearch#22581

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
@@ -563,10 +563,9 @@ public interface PPLTypeChecker {
         && actual instanceof AbstractExprRelDataType<?> actUdt) {
       return expUdt.getUdt() == actUdt.getUdt();
     }
-    if (expected instanceof AbstractExprRelDataType<?>
-        || actual instanceof AbstractExprRelDataType<?>) {
-      return false;
-    }
+    // Fall back to SqlTypeName comparison when one operand is a UDT and the other is a plain
+    // RelDataType. This covers database-sourced fields (e.g. TIMESTAMP) that arrive as plain
+    // RelDataType against allowed signatures using UDT variants (e.g. TIMESTAMP_UDT).
     return expected.getSqlTypeName() == actual.getSqlTypeName();
   }
 


dai-chen (commented):
Please add IT to verify.

PR #5671, opened 2026-08-03
Document that PPL fetch_size limits aggregation buckets

ahkcs:
### Description

`fetch_size` is applied to the **final** rows of a PPL query, so on an aggregating query it limits the number of aggregation **buckets** returned rather than sampling documents. The existing docs described it only as "limits the number of rows returned," which reads like DSL `size` — where aggregation results come back in full.

This adds a short section covering:

- `fetch_size` is equivalent to appending `| head <fetch_size>`, so for `stats`/`timechart`/`top`/`rare`/`patterns ... mode=aggregation` it truncates buckets.
- Aggregate **values** stay correct — unlike DSL `terminate_after`, which undercounts because documents are never collected. But unlike DSL `size`, aggregation results *are* truncated. It matches neither analogue exactly, which is what made the behavior easy to misread.
- Bucket counts grow with the queried time range rather than data volume, so a fixed limit truncates unexpectedly on wide ranges.

Raised in review of opensearch-project/OpenSearch-Dashboards#12511, where this behavior caused Discover to silently drop buckets from PPL visualizations.

### Issues Resolved

None; documentation gap found during review of OpenSearch-Dashboards#12511.

### Check List

Docs-only change. `docs/user/interfaces/` is not covered by the doctest suite, so no queries here are executed as assertions; the example values were confirmed by hand against a local 3.8.0-SNAPSHOT cluster.

### DIFF
--- a/docs/user/ppl/limitations/limitations.md
+++ b/docs/user/ppl/limitations/limitations.md
@@ -62,7 +62,7 @@ For the following functionalities, the query will be forwarded to the V2 query e
     * ML  
     * Kmeans  
 * `show datasources` and command  
-* SQL queries with `fetch_size` parameter (cursor-based pagination). Note: PPL's `fetch_size` (response size limiting, no cursor) is supported in Calcite Engine.
+* SQL queries with `fetch_size` parameter (cursor-based pagination). Note: PPL's `fetch_size` (response size limiting, no cursor) is supported in Calcite Engine. It limits the final rows of the query, so on an aggregating query it limits the number of buckets returned.
 
 
 ## Malformed Field Names in Object Fields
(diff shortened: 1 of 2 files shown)


docs/user/ppl/limitations/limitations.md
```diff
@@ -62,7 +62,7 @@ For the following functionalities, the query will be forwarded to the V2 query e
     * ML  
     * Kmeans  
 * `show datasources` and command  
-* SQL queries with `fetch_size` parameter (cursor-based pagination). Note: PPL's `fetch_size` (response size limiting, no cursor) is supported in Calcite Engine.
+* SQL queries with `fetch_size` parameter (cursor-based pagination). Note: PPL's `fetch_size` (response size limiting, no cursor) is supported in Calcite Engine. It limits the final rows of the query, so on an aggregating query it limits the number of buckets returned.
```
penghuo, line 65: As this is PPL limitations doc, Remove following sentence

> SQL queries with `fetch_size` parameter (cursor-based pagination). Note:

PR #5674, opened 2026-08-04
Match analytics types in aggregate signatures

ahkcs:
### Description
Treat analytics-engine schema markers as their equivalent PPL semantic types when validating UDT-backed aggregate signatures. This restores `LIST`/`VALUES` for timestamp, date, time, IP, and binary fields routed through the analytics engine.

The regression was exposed by OpenSearch sandbox CI after #5633: SQL-private UDTs and analytics-api types rendered the same name but failed class-based matching.

### Testing
- `./gradlew :core:test --tests org.opensearch.sql.expression.function.PPLTypeCheckerTest spotlessCheck`
- OpenSearch analytics REST tests for `DatetimeCoverageIT`, `ListAggregateMultiShardIT`, `ListAggregateMultiTypeIT`, `TwoShardAggregationIT`, and `TwoShardOversamplingAggregationIT`

Signed-off-by: Kai Huang <ahkcs@amazon.com>

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
@@ -552,22 +552,50 @@ public interface PPLTypeChecker {
     };
   }
 
-  /**
-   * Compares two RelDataTypes for signature matching. Two UDTs match if they share the same {@link
-   * ExprUDT} tag — comparing {@code getClass()} is unsafe because addCharsetAndCollation collapses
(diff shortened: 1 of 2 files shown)


ahkcs:
@sandeshkr419 This fixes the SQL snapshot regression currently blocking `sandbox-check` on opensearch-project/OpenSearch#22611. The analytics compatibility check is green, and I also verified the exact failing seed/test plus the affected OpenSearch datetime and list aggregation suites locally. Could you take a look?

PR #5675, opened 2026-08-04
Accept plain Calcite types against UDT operand signatures

ahkcs:
## Description

`PPLOperandTypes.SCALAR_TYPES` declares `DATE`/`TIME`/`TIMESTAMP`/`IP`/`BINARY` operands as UDTs, but `PPLTypeChecker.typesMatch` rejected a pair outright whenever only one side extended `AbstractExprRelDataType`:

```java
if (expected instanceof AbstractExprRelDataType<?>
    || actual instanceof AbstractExprRelDataType<?>) {
  return false;
}
```

The analytics engine builds its row types from plain Calcite types (`date` → `TIMESTAMP(3)`, `date_nanos` → `TIMESTAMP(9)`, `ip` and `binary` → `VARBINARY`) plus markers that extend Calcite's `AbstractSqlType`, not `AbstractExprRelDataType`. Every such operand therefore failed the check, so `list(<date field>)` and `values(<ip field>)` were rejected.

The resulting error contradicted itself, because `getAllowedSignatures()` renders via the UDT tag while `getActualSignature()` renders via `convertRelDataTypeToExprType` — both print `TIMESTAMP`:

```
Aggregation function LIST expects field type
{[BYTE]|...|[DATE]|[TIME]|[TIMESTAMP]|[IP]|[BINARY]}, but got [TIMESTAMP]
```

This maps the UDT tag to the `SqlTypeName`s a backend would emit for the same logical type. Comparing backing types would not work, since the UDTs are all VARCHAR-backed.

Note the mapping is expressed over `SqlTypeName` rather than by reusing `OpenSearchTypeFactory.convertAnalyticsEngineRelDataTypeToExprType`: `analytics-api` is a `compileOnly` dependency of `core`, so touching those marker classes throws `NoClassDefFoundError` on any runtime classpath that lacks them. An earlier revision of this fix did exactly that and broke 168 `:ppl:test` cases.

## Testing

New `PPLUdtSignatureMatchTest` covers the `ANY_SCALAR` checker behind `list(<field>)`:
- plain `TIMESTAMP(3)`/`TIMESTAMP(9)`, `DATE`, `TIME`, `VARBINARY` are accepted
- UDT operands still match (no behavior change on the existing path)
- plain scalars (`INTEGER`, `BIGINT`, `VARCHAR`, `BOOLEAN`) still match
- arrays and maps are still rejected, so `ANY_SCALAR` has not been widened into a catch-all

Verified the test fails without the production change (3 of 6 cases) and passes with it.

| Suite | Result |
|---|---|
| `:core:test` | pass (`--rerun-tasks`) |
| `:ppl:test` | pass (`--rerun-tasks`) |
| `:opensearch:test` | pass (`--rerun-tasks`) |
| `:core:spotlessCheck`, `:ppl:spotlessCheck` | pass |

### Downstream effect

This clears 11 pre-existing `sandbox-check` integration-test failures in `opensearch-project/OpenSearch`, which consumes `opensearch-sql-plugin:3.8.0.0-SNAPSHOT`:

- `ListAggregateMultiTypeIT` — `testListDate`, `testListDateNanos`, `testListIp`, `testListBinary`
- `ListAggregateMultiShardIT` — `testListDateTwoShards`, `testListIpTwoShards`, `testListBinaryTwoShards`
- `DatetimeCoverageIT` — `testListFunctionDateUsesSpace`, `testListFunctionTimestampUsesSpace`
- `TwoShardAggregationIT`, `TwoShardOversamplingAggregationIT` — `testReduceCorrectnessAcrossTwoShards`

Those failures reproduce identically on unrelated branches, confirming they are not caused by any single PR.

## Check List
- [x] Public documentation issue/PR created — not applicable; this restores intended behavior with no user-facing API change.

### DIFF
--- a/core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
+++ b/core/src/main/java/org/opensearch/sql/expression/function/PPLTypeChecker.java
@@ -557,17 +557,49 @@ public interface PPLTypeChecker {
    * ExprUDT} tag — comparing {@code getClass()} is unsafe because addCharsetAndCollation collapses
    * ExprDateType/ExprTimeType/ExprTimeStampType/ExprBinaryType down to ExprSqlType, so different
    * UDTs would appear equal. Plain types match by SqlTypeName.
+   *
+   * <p>A UDT signature also accepts the equivalent plain Calcite type. Signatures such as {@code
+   * PPLOperandTypes.ANY_SCALAR} declare temporal/IP/BINARY operands as UDTs, but the analytics
+   * engine builds its row types from plain Calcite types plus its own markers, which extend
+   * Calcite's {@code AbstractSqlType} rather than {@link AbstractExprRelDataType}. Matching on
+   * class identity alone made {@code list(<date field>)} fail with an error that listed the very
+   * type it had just rejected.
    */
   private static boolean typesMatch(RelDataType expected, RelDataType actual) {
     if (expected instanceof AbstractExprRelDataType<?> expUdt
         && actual instanceof AbstractExprRelDataType<?> actUdt) {
       return expUdt.getUdt() == actUdt.getUdt();
     }
-    if (expected instanceof AbstractExprRelDataType<?>
-        || actual instanceof AbstractExprRelDataType<?>) {
-      return false;
+    if (expected instanceof AbstractExprRelDataType<?> expUdt) {
+      return matchesPlainType(expUdt.getUdt(), actual);
+    }
+    if (actual instanceof AbstractExprRelDataType<?> actUdt) {
+      return matchesPlainType(actUdt.getUdt(), expected);
     }
     return expected.getSqlTypeName() == actual.getSqlTypeName();
+  }
+
+  /**
+   * Whether a plain Calcite type is the non-UDT spelling of {@code udt}. The UDTs themselves are
+   * all VARCHAR-backed, so this maps the tag to the {@link SqlTypeName}s a backend would produce
+   * for the same logical type instead of comparing backing types.
+   */
+  private static boolean matchesPlainType(ExprUDT udt, RelDataType plain) {
+    return switch (udt) {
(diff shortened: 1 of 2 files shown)


ahkcs:
### End-to-end verification against the analytics-engine QA suite

Built this branch's plugin zip and pointed `sandbox/qa/analytics-engine-rest` at it, so the only variable is the SQL plugin. Same OpenSearch commit (`f98b7665d2d`), same machine, JDK 25.

Confirmed the fix is in the shipped artifact first — `matchesPlainType` appears in the `core-3.8.0.0-SNAPSHOT.jar` bytecode inside the built zip, not just in source.

**A/B on `ListAggregateMultiTypeIT`:**

| SQL plugin zip | Result |
|---|---|
| Published `3.8.0.0-SNAPSHOT` (no fix) | **13 tests, 4 failed** |
| Built from this branch | **13 tests, 0 failed** |

The 4 baseline failures are exactly the reported ones — `testListDate`, `testListDateNanos`, `testListIp`, `testListBinary` — each with the self-contradictory error:

```
Aggregation function LIST expects field type
{...|[DATE]|[TIME]|[TIMESTAMP]|[IP]|[BINARY]}, but got [TIMESTAMP]
```

**All five previously-failing QA classes, with this branch's zip:**

| Class | Tests | Failures |
|---|---|---|
| `DatetimeCoverageIT` | 68 | 0 |
| `ListAggregateMultiTypeIT` | 13 | 0 |
| `ListAggregateMultiShardIT` | 13 | 0 |
| `TwoShardAggregationIT` | 1 | 0 |
| `TwoShardOversamplingAggregationIT` | 1 | 0 |
| **Total** | **96** | **0** |

That covers all 11 `sandbox-check` failures in `opensearch-project/OpenSearch` (the two `TwoShard*` classes each aggregate many per-function reduce checks, including `list_ts_date`, `list_client_ip`, and `values_client_ip`).

Swiddis (approved):

penghuo (approved):

Issue #5405, opened 2026-05-05
[RFC][FEATURE] PPL/SQL Query Linter & Static Analyzer


RyanL1997:
### 1.1 Today: validation is post-execution

In OpenSearch SQL/PPL, every form of feedback the user gets about query quality lands **after** the query has already been dispatched (or, in the editor, after the parser fails outright). There is no advisory layer that runs *before* execution to warn the user about queries that are syntactically valid but semantically suboptimal or wrong.

The current state, by stage:

| Stage | Where it runs | What it catches | What it misses |
| --- | --- | --- | --- |
| ANTLR parse | OSD Monaco worker + backend | Syntax errors only — rendered as `MarkerSeverity.Error` red squigglies | Anything past tokens/grammar |
| `SemanticCheckException` | Backend `Analyzer`, **at execution time** | Hard semantic failures (e.g., *"can't resolve field X"*) | Soft mistakes that don't fail the resolver |
| Cluster execution | Backend executor | Engine-level failures, OOMs, timeouts | Wrong-but-completing results |
| **Pre-execution advisory** | — | **— (does not exist today) —** | Everything semantically suboptimal |

Everything between *"ANTLR parsed it"* and *"the cluster ran it"* is invisible until after the user clicks **Run**. The Monaco editor's `MarkerSeverity.Warning` channel is currently unused.

### 1.2 What that means in practice

This is forgiving on input but unforgiving on output. The same query the user wrote is dispatched, possibly scans a year of data, and returns either an error, an empty result, or a *wrong* result. Examples we see in the wild:

- `... | where service = "checkout" | stats count() by user_id` on a 5-billion-doc index — runs to completion, OOMs the coordinator at the `stats` stage.
- `... | where message like "%5xx%"` against a `keyword`-mapped field — silently returns zero rows. The user assumes "no errors == no matches".
- `source = logs-* | head 100` against a wildcard pattern matching zero indices — empty result, no warning, indistinguishable from "no data in the time range".
- `... | sort @timestamp` after `stats` — sorts the wrong post-aggregation columns, producing non-deterministic output.
- A typo in a field name (`stauts` vs `status`) — `SemanticCheckException` thrown, but only after the user submits and waits.

None of these are syntax errors. None *necessarily* throw `SemanticCheckException` early enough to be useful. The user discovers the problem only after the query has either run wrong or failed at the cluster.

### 1.3 Where peers are

| Tool | Pre-execution advisory | Auto-fix | Cost preview |
[...]

PR #5678, opened 2026-08-05
feature(ci): validate PPL linter compatibility across grammar surfaces

Hanyu-W:
### Description

Adds `[Linter] PPL grammar compatibility`, which validates the SQL candidate
runtime grammar with the paired OpenSearch Dashboards production headless
linter.

- Pairs SQL `main` or `X.Y` with the same OSD branch and records both SHAs.
- Exports the grammar directly from `:ppl`; no cluster or backend execution.
- Runs 24 trigger/control cases for the 12 default-enabled catalog rules from
  [OSD #12526](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12526).
- Enforces `coveredRuleIds == catalogRuleIds - excludedRuleIds`. The four
  default-off headless rules and two explain-backed rules are excluded with
  reasons; missing or stale classifications fail.
- Uses exact OSD diagnostic counts. Error-recovered parse trees remain valid.

`command-suggestion` is the thirteenth default-enabled check, but it is a
syntax-channel feature rather than a catalog detector and is out of scope.

Supports #5405.

### Testing

- `node --test scripts/ppl-lint/__tests__/*.test.mjs` (37 passing)
- `./gradlew :ppl:test --tests org.opensearch.sql.ppl.autocomplete.PPLGrammarBundleExporterTest`
- `actionlint .github/workflows/ppl-lint-grammar-compatibility.yml`
- `git diff --check`

### Check List

- [x] New functionality includes testing and documentation.
- [x] No new PPL command or public API is introduced.
- [x] Commits are signed per the DCO.

### DIFF
--- a/.github/workflows/ppl-lint-grammar-compatibility.yml
+++ b/.github/workflows/ppl-lint-grammar-compatibility.yml
@@ -7,53 +7,36 @@ on:
       - '[0-9]+.[0-9]+'
     paths:
       - build.gradle
-      - settings.gradle
       - ppl/build.gradle
       - ppl/src/main/antlr/OpenSearchPPLLexer.g4
       - ppl/src/main/antlr/OpenSearchPPLParser.g4
       - ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/GrammarBundle.java
       - ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
-      - plugin/build.gradle
-      - plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
-      - plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
+      - ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleExporter.java
       - scripts/ppl-lint/**
       - scripts/ppl-lint-rule-validation.sh
       - .github/workflows/ppl-lint-grammar-compatibility.yml
-      - docs/dev/ppl-lint-grammar-compatibility-ci*.md
   push:
     branches:
       - main
       - '[0-9]+.[0-9]+'
     paths:
       - build.gradle
-      - settings.gradle
       - ppl/build.gradle
       - ppl/src/main/antlr/OpenSearchPPLLexer.g4
       - ppl/src/main/antlr/OpenSearchPPLParser.g4
       - ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/GrammarBundle.java
       - ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleBuilder.java
-      - plugin/build.gradle
-      - plugin/src/main/java/org/opensearch/sql/plugin/SQLPlugin.java
-      - plugin/src/main/java/org/opensearch/sql/plugin/rest/RestPPLGrammarAction.java
+      - ppl/src/main/java/org/opensearch/sql/ppl/autocomplete/PPLGrammarBundleExporter.java
       - scripts/ppl-lint/**
       - scripts/ppl-lint-rule-validation.sh
       - .github/workflows/ppl-lint-grammar-compatibility.yml
-      - docs/dev/ppl-lint-grammar-compatibility-ci*.md
   workflow_dispatch:
     inputs:
-      osd_repo:
-        description: Optional OpenSearch Dashboards repository override.
-        required: false
-        type: string
       osd_ref:
-        description: Optional OpenSearch Dashboards ref override.
+        description: Optional ref in opensearch-project/OpenSearch-Dashboards.
         required: false
         type: string
-      allow_release_line_mismatch:
-        description: Development-only bypass for exact release-line validation.
-        required: false
-        default: false
-        type: boolean
 
 permissions:
   contents: read
@@ -82,9 +65,7 @@ jobs:
           EVENT_NAME: ${{ github.event_name }}
           TARGET_BRANCH: ${{ github.base_ref || github.ref_name }}
           SQL_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
-          REQUESTED_OSD_REPO: ${{ inputs.osd_repo }}
           REQUESTED_OSD_REF: ${{ inputs.osd_ref }}
-          REQUESTED_BYPASS: ${{ inputs.allow_release_line_mismatch }}
         run: |
           set -euo pipefail
           sql_raw=$(sed -nE 's/.*opensearch_version = System\.getProperty\("opensearch\.version", "([^"]+)"\).*/\1/p' build.gradle)
@@ -92,13 +73,12 @@ jobs:
             { echo "::error::Invalid SQL product version: $sql_raw"; exit 1; }
           sql_version="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}.${BASH_REMATCH[3]}"
           sql_line="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}"
-          bypass=$([[ "$EVENT_NAME" == workflow_dispatch && "$REQUESTED_BYPASS" == true ]] && echo true || echo false)
           case "$TARGET_BRANCH" in
             main) canonical_ref=main ;;
             *)
               [[ "$TARGET_BRANCH" =~ ^[0-9]+\.[0-9]+$ ]] ||
                 { echo "::error::Target branch must be main or an exact X.Y release branch"; exit 1; }
-              [[ "$TARGET_BRANCH" == "$sql_line" || "$bypass" == true ]] ||
+              [[ "$TARGET_BRANCH" == "$sql_line" ]] ||
                 { echo "::error::SQL $sql_raw does not match target branch $TARGET_BRANCH"; exit 1; }
               canonical_ref="$TARGET_BRANCH"
               ;;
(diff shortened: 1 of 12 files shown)


RyanL1997 (changes requested):
This PR needs a reconstruction:
- first we need to clearify the scope of this testing: it should be catching the grammar changes that impacts the rules we are having for linter in OSD
- the CI workflows should handle the version dynamically - e.g. at the future versions this change will be in the next release version 3.9 and when that happen, we should automatically picking up both SQL version and OSD version from that branch instead of main
- No need to implement more ITs for the existing grammar this is out of the scope of this change. It should be ONLY focusing on catching the linter related grammar change by the OSD validator, not the SQL plugin itself.

scripts/ppl-lint/validate-osd-grammar.mjs
```diff
@@ -0,0 +1,463 @@
...
+      ruleId: grammarCase.ruleId,
+      kind: grammarCase.kind,
+      query: grammarCase.query,
+      expectedCount: grammarCase.expectedCount,
+    };
+    try {
+      if (api.buildTree) {
+        const parse = api.buildTree(grammarCase.query, grammar);
+        if (!hasParseTree(parse)) throw new Error('candidate parser produced no parse tree');
+        if (hasErrorNode(parseTreeOf(parse))) {
```
RyanL1997: **This asserts a contract OSD deliberately does not hold.** `buildRuntimeTree` in `headless_ppl_lint.ts` documents the opposite: it keeps error-recovered trees on purpose because "the lint rules are written to walk it best-effort", citing `eval x = <field> + 1` as a query the engine accepts but the runtime ATN cannot fully parse.

So a grammar change where the parser recovers but every lint rule still behaves correctly fails here, carrying no information about linter breakage. The diagnostic-count comparison below is the real signal — I would drop this gate or make it non-fatal.

Also `node.constructor?.name === 'ErrorNode'` is name-based type detection, fragile across antlr4ng versions and under bundling.

.github/workflows/ppl-lint-grammar-compatibility.yml
```diff
@@ -0,0 +1,389 @@
...
+          distribution: temurin
+          java-version: 21
+          cache: gradle
+
+      - name: Capture candidate runtime grammar
+        if: ${{ steps.capability.outputs.available == 'true' }}
+        timeout-minutes: 25
+        run: |
+          set -euo pipefail
+          ./gradlew :opensearch-sql-plugin:run --no-daemon > ppl-grammar-cluster.log 2>&1 &
```
RyanL1997: **The cluster is not needed.** `PPLGrammarBundleBuilder.build()` is a pure static function over the generated ANTLR classes — no `Settings`, no `NodeClient`, no cluster state, and `getBundle()` is a static singleton. The bundle is a build artifact of the `.g4` files.

Booting the plugin (25-minute step timeout, 120x5s readiness poll, port-9200 guard, PID traps, cluster log) to `curl` static JSON is the biggest cost and flake surface in the PR. A small `JavaExec` in `:ppl` that writes `ppl-grammar-bundle.json` replaces all of it, deletes ~150 lines of bash across this file and the wrapper script, and takes the job from ~45 min to a few. It also moots the analyzer findings about PID reuse and trap handling.

.github/workflows/ppl-lint-grammar-compatibility.yml
```diff
@@ -0,0 +1,389 @@
...
+            }' > "$GITHUB_WORKSPACE/resolved-target.json"
+          echo "osd_version=$osd_version" >> "$GITHUB_OUTPUT"
+          echo "osd_release_line=$osd_line" >> "$GITHUB_OUTPUT"
+
+      - name: Detect headless grammar capability
+        id: capability
+        run: |
+          set -euo pipefail
+          module=.ci/OpenSearch-Dashboards/src/plugins/data/public/antlr/opensearch_ppl/headless_ppl_lint
+          if [[ -f "${module}.ts" || -f "${module}.js" ]]; then
```
RyanL1997: **Weakest of three disagreeing capability probes — this is the false-green path.**

This accepts only `.ts` and `.js`. The wrapper (`ppl-lint-rule-validation.sh:386-391`) also accepts bare and `.mjs`. The adapter (`validate-osd-grammar.mjs:127-130`) accepts bare via `existsSync`, which additionally matches a **directory**.

So if OSD refactors `headless_ppl_lint.ts` into `headless_ppl_lint/index.ts`, this returns false, the adapter writes `skipReason: osd-headless-grammar-api-unavailable`, and the job exits 0 — green, indistinguishable from the legitimate old-branch skip, while the API is present and working.

Suggest deleting this step and letting the adapter be the single source of truth; it already handles the absent-module case at `:437-440` and returns 0 with the skip report.

scripts/ppl-lint/validate-osd-grammar.mjs
```diff
@@ -0,0 +1,463 @@
...
+    if (sqlLine !== branch || osdLine !== branch) {
+      fail(
+        `Exact release target ${branch} requires SQL and OSD ${branch}.x versions; ` +
+          `got SQL ${target.sql.version} and OSD ${target.osd.version}.`
+      );
+    }
+  }
+}
+
+function modulePath(root, name) {
```
RyanL1997: Third variant of the capability probe: bare via `existsSync` (matches a directory), `.js`, `.ts` — no `.mjs`.

This is the most permissive of the three and, being colocated with the code that actually loads the module, the only one that can be correct by construction. Make it the only one — see the comment on `ppl-lint-grammar-compatibility.yml:211`.

scripts/ppl-lint-rule-validation.sh
```diff
@@ -0,0 +1,493 @@
...
+  --cases "$GRAMMAR_CASES"
+  --target "$TARGET"
+  --osd-root "$OSD_ROOT"
+  --osd-sha "$OSD_SHA"
+  --report "$REPORT"
+  --summary "$SUMMARY"
+)
+
+headless_module_exists() {
+  [[ -f "$OSD_ROOT/$HEADLESS_MODULE" ||
```
RyanL1997: Second variant of the capability probe: bare, `.ts`, `.js`, `.mjs` — but all via `-f`, so unlike the adapter it never matches a directory.

Three probes, three different answers for the same OSD checkout. A `.mjs` module is available here and unavailable to the workflow; a directory module is available to the adapter and unavailable to both of these. Collapse to the adapter's check.

scripts/ppl-lint/validate-osd-grammar.mjs
```diff
@@ -0,0 +1,463 @@
...
+  if (grammar.grammarHash && grammar.grammarHash !== bundle.grammarHash) {
+    fail('Deserialized grammar hash does not match the bundle.');
+  }
+  return { bundle, grammar };
+}
+
+function loadCases(file, catalogIds) {
+  const document = readJson(file, 'grammar cases');
+  const rawCases = Array.isArray(document) ? document : object(document, 'case document').cases;
+  if (!Array.isArray(rawCases) || !rawCases.length) fail('At least one grammar case is required.');
```
RyanL1997: **No coverage floor.** This requires at least one case, and the loop below requires trigger+control per selected rule — but nothing requires a minimum number of *rules*.

Truncate `grammar-cases.json` to a single rule and the report reads `rules: {selected: 1, passed: 1}` and goes green. Combined with the missing reverse-coverage check against the OSD catalog, deleting a failing case is a valid way to make CI pass.

Asserting the covered set against `catalogIds` minus an explicit allow-list fixes both at once.

Hanyu-W: Fixed in a7e3b2f8a. The adapter now enforces coveredRuleIds == catalogIds - excludedRuleIds, with a required reason for every exclusion. Deleting a required rule’s cases, adding an unclassified catalog rule, leaving a stale exclusion, overlapping covered/excluded sets, or omitting a trigger/control pair now fails structurally. The required set is the default-enabled catalog rules from OSD #12526.

scripts/ppl-lint-rule-validation.sh
Hanyu-W: Removed the shell capability probe as well. The local wrapper now follows the same path as CI and always delegates the availability decision to the adapter's resolver/loader.

scripts/ppl-lint/validate-osd-grammar.mjs
Hanyu-W: Consolidated resolution and loading in the adapter. Only a true failure to resolve the headless entry point becomes a legacy skip; import failures, transitive dependency failures, and invalid exports now fail structurally.

.github/workflows/ppl-lint-grammar-compatibility.yml
Hanyu-W: Deleted the workflow-level file probe and its conditional skip path. Every run now delegates capability and validation to the adapter, so the workflow cannot independently produce a false-green absence decision.

.github/workflows/ppl-lint-grammar-compatibility.yml
Hanyu-W: Replaced the development cluster and REST capture with :ppl:exportPplGrammarBundle, backed directly by PPLGrammarBundleBuilder.getBundle(). The workflow and local wrapper no longer use port 9200, polling, curl, PID traps, or cluster logs.

scripts/ppl-lint/validate-osd-grammar.mjs
Hanyu-W: Removed the parse-cleanliness gate and the ErrorNode name check. The adapter now relies only on lintQueryWithBundle() diagnostic behavior, so an error-recovered tree passes when the production OSD rules still behave correctly.

Hanyu-W:
@RyanL1997 Thank you for the feedback. I have pushed changes to address the review, please re-review when you get a chance. Github Actions are currently down preventing me from testing the ci, will update this comment with CI testing results once Github Actions is fixed.

EDIT: Github Actions is up, this is what the CI looks like now:
image

RyanL1997 (approved):

PR #5681, opened 2026-08-05
fix: push down date range filters wrapped by redundant date conversions (both operands, v2 + Calcite)

### LINKED ISSUE #5680: [BUG] Date range filter on a date field falls back to a per-document script instead of a native range when the field is wrapped in timestamp() / CAST(... AS TIMESTAMP)
## Query Information

**PPL Command/Query:**
```ppl
# Filtering a `date`-mapped field where the field is wrapped in timestamp()/CAST(... AS TIMESTAMP).
# This wrapped shape is what clients such as the Grafana OpenSearch data source generate
# automatically for their dashboard time filter.
source=my-logs
| where timestamp(event_time) >= cast('2024-01-15 12:00:00' as timestamp)
    and timestamp(event_time) <= cast('2024-01-15 15:00:00' as timestamp)
| stats count() as c

# The logically-identical BARE-field form behaves correctly (pushes down to a native range):
# source=my-logs
# | where event_time >= '2024-01-15 12:00:00' and event_time <= '2024-01-15 15:00:00'
# | stats count() as c
```

**Expected Result:**
The wrapped-field range predicate should push down to a native OpenSearch `range` query (BKD/points accelerated), exactly like the bare-field form. `POST _plugins/_ppl/_explain` should show a `range` query on `event_time`.

**Actual Result:**
The predicate is emitted as a per-document `ScriptQueryBuilder` (`"lang": "opensearch_compounded_script"`) whose decoded body is `gte(timestamp(event_time), cast_to_timestamp(...))`, i.e. it calls `TypeCastOperators.castToTimestamp` → `LocalDateTime.parse` for **every scanned document**. `_explain` shows a `script` filter instead of `range`. Because the script has no index acceleration, the timestamp is string-parsed per document across all shards, saturating the search thread pool on large indices. The serialized script also embeds a per-request timestamp, so each execution is a unique script that must be recompiled, which can trip the script-compilation-rate circuit breaker and surface as `all shards failed`.

## Dataset Information

**Dataset/Schema Type**
- [ ] OpenTelemetry (OTEL)
- [x] Custom (details below) — time-series log records with a `date`-typed timestamp field

**Index Mapping**
```json
{
  "mappings": {
    "properties": {
      "event_time":  { "type": "date" },
      "event_type":  { "type": "keyword" },
      "request_id":  { "type": "text" },
      "group_id":    { "type": "keyword" }
    }
  }
}
```

**Sample Data**
```json
{
  "event_time": "2024-01-15T13:30:00.000Z",
  "event_type": "type_a",
  "request_id": "req-000042",
  "group_id": "group-42"
}
```

## Bug Description

**Issue Summary:**
When a PPL/SQL predicate compares a `date`-mapped field to a timestamp literal **and the field is wrapped in a date function** (`timestamp(<field>)` or `CAST(<field> AS TIMESTAMP)`), the SQL engine does not push the comparison down to a native `range` query. It falls back to a per-document script filter. The same filter written against the bare field pushes down to `range` and is orders of magnitude cheaper. Wrapping an already-`date`/`timestamp`-typed field in `timestamp()`/`CAST(... AS TIMESTAMP)` is a no-op for a range comparison (monotonic, range-preserving), so it should push down identically.

**Steps to Reproduce:**
1. Create an index with a plain `date` field (mapping above) and index a few documents (sample above).
2. Run `_explain` on the **bare-field** form and confirm it pushes down to `range`:
   ```json
   POST _plugins/_ppl/_explain
   { "query": "source=my-logs | where event_time >= '2024-01-15 12:00:00' and event_time <= '2024-01-15 15:00:00' | stats count() as c" }
   ```
   → `{"range":{"event_time":{...}}}`.
3. Run `_explain` on the **wrapped-field** form and observe the fallback to a script filter:
   ```json
   POST _plugins/_ppl/_explain
   { "query": "source=my-logs | where timestamp(event_time) >= cast('2024-01-15 12:00:00' as timestamp) and timestamp(event_time) <= cast('2024-01-15 15:00:00' as timestamp) | stats count() as c" }
   ```
   → two `script` filters (`"lang": "opensearch_compounded_script"`). Both queries return the same count; only the execution plan differs.

**Impact:**
On large time-series indices the per-document `LocalDateTime.parse` runs as a linear scan across all shards, pinning search threads at high CPU and driving allocation/GC pressure; repeated executions can trip the script-compilation-rate breaker (`all shards failed`). Because common clients (e.g. the Grafana OpenSearch data source) auto-generate the wrapped `timestamp(<timeField>)` time filter, users hit this without writing it themselves, so the impact is broad for time-series dashboards and alerts backed by PPL/SQL.

## Environment Information

**OpenSearch Version:** 2.19 (reproduced on the released distribution). Reproducible via both PPL and SQL; independent of security settings.

**Additional Details:**

### Root cause
In `opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java`, `canSupport()` only allows push-down when the left operand is a bare `ReferenceExpression`:
```java
public boolean canSupport(FunctionExpression func) {
  return (func.getArguments().size() == 2)
          && (func.getArguments().get(0) instanceof ReferenceExpression)
          && (func.getArguments().get(1) instanceof LiteralExpression
              || literalExpressionWrappedByCast(func))
      || isMultiParameterQuery(func);
}
```
A `timestamp(field)` / `cast(field as timestamp)` left operand is a `FunctionExpression`, so `canSupport` returns false and `FilterQueryBuilder.visitFunction` falls back to `buildScriptQuery(func)`.

### Proposed fix
Fold a redundant date/time cast on the field side to the underlying field reference so the predicate pushes down to a native range. In `LuceneQuery`:
```java
/**
 * True if the operand is a date/time cast (or timestamp()/date()/time() builtin) applied to a
 * reference whose field type is already an OpenSearchDateType. Such a wrap is redundant for a
 * range comparison and can be unwrapped so the predicate pushes down instead of scripting.
 */
protected boolean referenceWrappedByRedundantDateCast(Expression arg) {
  if (arg instanceof FunctionExpression) {
    FunctionExpression fn = (FunctionExpression) arg;
    FunctionName name = fn.getFunctionName();
    boolean isDateCast =
        name.equals(BuiltinFunctionName.CAST_TO_TIMESTAMP.getName())
            || name.equals(BuiltinFunctionName.CAST_TO_DATE.getName())
            || name.equals(BuiltinFunctionName.CAST_TO_TIME.getName())
            || name.equals(BuiltinFunctionName.TIMESTAMP.getName())
            || name.equals(BuiltinFunctionName.DATE.getName())
            || name.equals(BuiltinFunctionName.TIME.getName());
    return isDateCast
        && fn.getArguments().size() == 1
        && fn.getArguments().get(0) instanceof ReferenceExpression
        && fn.getArguments().get(0).type() instanceof OpenSearchDateType;
  }
  return false;
}

private ReferenceExpression unwrapReference(Expression arg) {
  if (arg instanceof ReferenceExpression) {
    return (ReferenceExpression) arg;
  }
  return (ReferenceExpression) ((FunctionExpression) arg).getArguments().get(0);
}
```
`canSupport()` accepts the wrapped-but-redundant left operand:
```java
public boolean canSupport(FunctionExpression func) {
  return (func.getArguments().size() == 2)
          && (func.getArguments().get(0) instanceof ReferenceExpression
              || referenceWrappedByRedundantDateCast(func.getArguments().get(0)))
          && (func.getArguments().get(1) instanceof LiteralExpression
              || literalExpressionWrappedByCast(func))
      || isMultiParameterQuery(func);
}
```
`build()` unwraps before building the range:
```java
public QueryBuilder build(FunctionExpression func) {
  ReferenceExpression ref = unwrapReference(func.getArguments().get(0));
  Expression expr = func.getArguments().get(1);
  ExprValue literalValue =
      expr instanceof LiteralExpression ? expr.valueOf() : cast((FunctionExpression) expr, ref);
  return doBuild(ref.getAttr(), ref.type(), literalValue);
}
```
The fold is restricted to `OpenSearchDateType` references, so it only applies where the cast is genuinely redundant and range-preserving. The equivalent normalization (`timestamp($dateRef) → $dateRef`) should also be added to the Calcite/v3 predicate path.

### Validation
Built from source and run as single-node OpenSearch 2.19 containers on the same host (100k docs, `event_time` mapped as native `date`, 8 iterations each). Three builds are compared so the behavior change can be traced end to end:

- **2.19 (pre-#3615)** — a clean released 2.19 with neither the limit-pushdown change (PR #3615) nor this fold. Included as a sanity check of the pre-upgrade behavior.
- **2.19 + #3615** — adds the limit-pushdown correctness change (PR #3615), no date-cast fold.
- **2.19 + #3615 + fold** — adds the date-cast fold proposed here.

`_explain` for the wrapped-field predicate is a `script` (`opensearch_compounded_script`, per-document `castToTimestamp`) on both the pre-#3615 and the #3615 builds, and a native `range` on the fold build.

Full numbers across query shapes. `DATE` = `timestamp(event_time) >= cast(… as timestamp) and timestamp(event_time) <= cast(… as timestamp)`; `SEL` = a selective term filter (e.g. `event_type="type_a" or event_type="type_b"`). Each cell is `plan · median latency`:

| Query shape | 2.19 (pre-#3615) | 2.19 + #3615 | 2.19 + #3615 + fold |
|---|---|---|---|
| `where DATE \| head 10000 \| where SEL \| stats count()` | `script` · 51 ms | `script` · 228 ms | `range` · 177 ms |
| `where DATE \| where SEL \| head 10000 \| stats count()` | `script` · 44 ms | `script` · 46 ms | `range` · 36 ms |
| `where DATE \| where SEL \| stats count()` (no `head`) | `script` · 36 ms | `script` · 33 ms | `range` · 25 ms |
| `where DATE \| stats count()` (pure date) | `script` · 248 ms | `script` · 249 ms | `range` · 99 ms |

Each build holds its own independently randomized data, so absolute totals differ slightly: the `DATE` window holds ~37k docs and the selective result is ~1.5k.

Reading the table:
- **The date predicate is a per-document `script` on both the pre-#3615 and #3615 builds; the fold is what turns it into a native `range`.** The pure-date row — no selective filter to front the script — is the clearest view of the per-document cost: ~248 ms → 99 ms (~2.5×) here, widening with data volume and shard count.
- **The head-before shape (`… | head 10000 | where SEL | …`) is where behavior changed across #3615.** Pre-#3615, the trailing selective filter was pushed below the `head`, so it narrowed the script's candidate set and head-before ran fast (51 ms) and returned the same count as the other shapes (~1473). #3615 correctly stops pushing a filter that follows a `head`/`LIMIT` (a limit-then-filter correctness fix), so the selective filter now runs after the limit: head-before returns a truncated-then-filtered count (~407) and the date `script` is left to run unnarrowed over the whole window (228 ms). This is the regression a user sees after upgrading to a 2.19 patch that includes #3615.
- **The fold fixes that regression without undoing #3615's correctness fix.** With the date predicate pushed down to a native `range`, head-before drops to 177 ms and every shape uses `range`; results match the #3615 build.
- **`head` placed after the filters (head-after / no-head) is unaffected across all three builds** — the selective filter fronts the script there, so those shapes are already cheap and stay cheap; they simply move from `script` to `range` with the fold.

## Screenshots
Not applicable — the issue is fully reproducible from the CLI and is demonstrated by the `_explain` plan output shown under **Actual Result** and **Steps to Reproduce** (native `range` for the bare field vs `opensearch_compounded_script` for the wrapped field).

burnthm:
### Description

A PPL/SQL range comparison on a `date`-mapped field can fall back to a per-document script instead of pushing down to a native OpenSearch `range` query, whenever either operand is wrapped in a redundant date/time conversion. The scripted path calls `castToTimestamp` → `LocalDateTime.parse` for **every scanned document**; with no BKD/points acceleration this string-parses the timestamp per document across all shards and saturates the search thread pool. The serialized script also embeds a per-request timestamp, so each execution recompiles and can trip the script-compilation-rate breaker.

`canSupport()` gates push-down on both operands, and each side has its own gap. **This PR fixes both:**

**1. Field operand — fold a redundant conversion on the filtered field.** Push-down required the left operand to be a bare field reference, so `timestamp(<field>)` or `CAST(<field> AS TIMESTAMP)` disqualified it. Wrapping a field that is already of that same date/time type is a no-op, so the wrap is folded back to the field reference.

**2. Value operand — resolve a bound wrapped in a date conversion.** Push-down accepted the right operand only as a literal or a `CAST_TO_*(literal)`. A bound written as `timestamp('...')` therefore scripted *even with a bare field on the left*:

```
where `event_time` >= timestamp('2026-08-04 17:42:40')
```

This is the shape clients emit for a time-range bound. The Grafana OpenSearch data source builds its PPL time filter exactly this way — [`ppl_request.go`](https://github.com/grafana/opensearch-datasource/blob/main/pkg/opensearch/client/ppl_request.go) formats ``where `%s` >= timestamp('%s')``, i.e. **bare field, conversion around the literal**. PPL coerces the string argument first, so the operand arrives as `timestamp(cast_to_timestamp('...'))`: the function name is `timestamp`, which is not in `castMap`, and its argument is a cast rather than a literal, so both halves of the existing check fail. The value operand is now resolved to the cast to evaluate plus the literal to evaluate it on, accepting the `timestamp()`/`date()`/`time()` builtins alongside `CAST_TO_*`. The cast is *resolved* rather than evaluated directly so `castMap` still parses the literal against the field's declared date formats.

Fix 1 applies to both engines (`LuceneQuery` for v2, `PredicateAnalyzer` for Calcite/v3) because the push-down decision lives in a different place on each and the v2 change alone has no effect on 3.x, which uses Calcite by default. Fix 2 is v2-only — Calcite already pushed that shape down.

**Both folds apply only where the conversion is genuinely a no-op.** Three guards, each with a negative test:

1. **The conversion target must equal the type it is applied to.** `date(<timestamp field>)` truncates the time component (so `date(ts) <= '2024-01-15'` is *not* `ts <= '2024-01-15'`), and `time(<timestamp field>)` extracts the time of day, which is not even monotonic in the timestamp.
2. **Only `CAST` and the `timestamp()`/`date()`/`time()` conversion operators qualify.** Result type alone is insufficient — other single-argument functions also return a date/time type while changing the value, `LAST_DAY` being the clearest example.
3. **A conversion that changes the type of an inner cast on the value operand** is not a no-op and stays on the script path.

Behaviour is unchanged throughout: only the execution plan changes, from `script` to `range`.

### Validation

**The value-operand fix, on a real cluster.** `birthdate > timestamp('2016-12-08 00:00:00')` (the client-generated shape) now emits two native `range` queries on `birthdate` with no script clause — the same plan as the bare-field form. Pinned by `ExplainIT.testFilterTimestampWrappedBoundPushDownExplain`.

**v2 and v3 return identical doc counts** for the field-operand fold. Both engines run against an identical 100k-doc dataset (fixed RNG seed and insertion order; `event_action` histogram verified equal on both nodes). Nodes: `2.19.0` + v2 fold, `3.7.0` + Calcite fold.

| query shape | v2 (2.19) | v3 (3.7) | counts |
|---|---|---|---|
| pure date, `timestamp()` wrap | RANGE / 37507 | RANGE / 37507 | match |
| pure date, `CAST(..)` wrap | RANGE / 37507 | RANGE / 37507 | match |
| pure date, bare field (control) | RANGE / 37507 | RANGE / 37507 | match |
| + selective filter, no `head` | RANGE / 1488 | RANGE / 1488 | match |
| + selective filter, `head` after `where` | RANGE / 1488 | RANGE / 1488 | match |
| + selective filter, `head` before `where` | RANGE / 385 | RANGE / 385 | match |

Every shape plans a native range on both engines and returns the same count, **including the bare-field control** — so the fold reproduces bare-field semantics rather than merely agreeing with itself.

**Performance** (single node, 100k docs, identical dataset, median of 5). On the unpatched node the bare-field form of the same filter runs in 25 ms while the wrapped form takes 628 ms — a ~25× penalty purely from the wrap. After the fix the wrapped form is back to bare-field parity:

| Query shape | Unpatched | Patched | Speedup |
|---|---|---|---|
| pure date (`timestamp()` wrap) | `script` 628 ms | `range` 20 ms | 31× |
| pure date (`CAST` wrap) | `script` 550 ms | `range` 18 ms | 31× |
| pure date (bare field, control) | `range` 25 ms | `range` 18 ms | — |
| + selective filter, no `head` | `script` 53 ms | `range` 19 ms | 2.8× |
| + selective filter, `head` after `where` | `script` 122 ms | `range` 58 ms | 2.1× |
| + selective filter, `head` before `where` | `script` 396 ms | `range` 218 ms | 1.8× |

Shapes with a selective filter look milder only because that filter pushes down natively and leads the Lucene conjunction, so the per-document script runs on a small matching subset (~1.5k of 100k).

**Guard cases** were verified on a 3-doc deterministic index: `timestamp(ts)` / `CAST(ts AS TIMESTAMP)` flip `script` → `range` with unchanged counts, while `date(ts)`, `time(ts)` and `last_day(d)` stay on the script path with unchanged counts.

### Related Issues
Resolves #5680

### Check List
  - v2 `FilterQueryBuilderTest`: both folds produce a `range`; a type-changing conversion on either operand, and a cast over a non-date field, stay on the script path.
  - v3 `PredicateAnalyzerTest`: the fold produces a `RangeQueryBuilder`; `date()` over a timestamp field and `last_day()` stay on the script path.
  - `ExplainIT` (inherited by `CalciteExplainIT` and `CalciteNoPushdownIT`, so each covers v2, Calcite-with-pushdown and Calcite-without-pushdown): the wrapped-bound and wrapped-field shapes push down to a native range; `last_day()` does not.
  - `CastFunctionIT`: the wrapped and bare forms return identical rows.


burnthm:
On v3, a comparison whose field side is a CAST(... AS TIMESTAMP) / timestamp(field) (SqlKind.CAST, handled by supportedRexCall/visitCall → toCastExpression) won't fold to a bare field reference, so it falls back to a ScriptQueryBuilder instead of emitting a native rangeQuery, which is the same per-document parsing cost this issue describes, just on the Calcite engine. Users on the v3 engine (or with plugins.calcite.enabled) would still hit the regression.

Proposed follow-up: in PredicateAnalyzer, when the operand of a comparison is a CAST/date-builtin over a field whose OpenSearch type is already an OpenSearchDateType, unwrap it to the underlying field ref before building the range (mirroring referenceWrappedByRedundantDateCast/unwrapReference here), with the same guardrail: only unwrap when the inner field is genuinely date-typed so non-order-preserving conversions still use the script path.

Can handle this as a separate follow-up PR to keep this one focused on the v2 path, or fold it into this PR if reviewers prefer a single change.

burnthm:
## v3 (Calcite) fix added, plus a correctness fix to the v2 path

Following up on my earlier note about the Calcite engine — I've folded the v3 change into this PR rather than a separate one, since the two halves share the same rule and the same guard. Two new commits:

| Commit | What |
|---|---|
| `fix: only fold a date cast when it does not change the type` | Correctness fix to the **v2** path (details below) |
| `fix: push down date filter wrapped by redundant date cast (Calcite)` | The **v3** `PredicateAnalyzer` counterpart |

### 1. Latent correctness bug in the original v2 commit (please look here first)

My first version checked "is this a date/time cast?" and "is the operand a date-typed field?" but never that the two **matched**. That folds away casts which are real conversions, silently changing results:

- `date(<timestamp field>)` truncates the time component, so `date(ts) <= '2024-01-15'` is **not** `ts <= '2024-01-15'`. Against docs at `2024-01-15 08:00` and `2024-01-15 23:00` the predicate correctly matches **2** rows; folded to the bare field it matched **0**.
- `time(<timestamp field>)` extracts the time of day, which isn't even monotonic in the timestamp (23:00 on day 1 sorts after 01:00 on day 2), so no range rewrite is valid at all.

Fixed by mapping each cast function to the type it produces and folding only when that target equals the field's own type — i.e. only when the cast is genuinely a no-op. The intended case (`timestamp()` / `CAST(... AS TIMESTAMP)` over a `date`-mapped field, which is `TIMESTAMP` in OpenSearch) still pushes down. Both engines now carry the same guard, each with a negative test.

### 2. The v3/Calcite half

`PredicateAnalyzer` only pushes a comparison down to a native range when the operand is a bare field reference, so on the Calcite engine the wrapped form falls back to a per-document script exactly as it does on v2. Since 3.x uses Calcite by default, **the v2 fold alone has no effect there** — I confirmed this on a stock 3.7.0 node before writing the v3 change.

One implementation gotcha worth flagging for reviewers: date/time values are modelled as UDTs whose backing `SqlTypeName` is `VARCHAR`, so `call.getType().getSqlTypeName()` **never** returns `TIMESTAMP`/`DATE`/`TIME`. My first attempt checked `getSqlTypeName()` and was silently dead code — the patched node behaved exactly like stock. The UDT (`EXPR_TIMESTAMP`/`EXPR_DATE`/`EXPR_TIME`) is what identifies the cast target, matching the existing `instanceof ExprSqlType` → `getUdt()` idiom already in that file.

### 3. Validation — v2 and v3 return identical doc counts

Both engines were run against an **identical** 100k-doc dataset (fixed RNG seed and insertion order; `event_action` histogram verified equal on both nodes: `break_enter=9935, cue_message=1965, gar=1946, impression=76143, other=10011`). Nodes: `2.19.0` + v2 fold, and `3.7.0` + the Calcite fold.

| query shape | v2 (2.19) | v3 (3.7) | counts |
|---|---|---|---|
| pure date, `timestamp()` wrap | RANGE / 37507 | RANGE / 37507 | match |
| pure date, `CAST(..)` wrap | RANGE / 37507 | RANGE / 37507 | match |
| pure date, bare field (control) | RANGE / 37507 | RANGE / 37507 | match |
| + selective filter, no `head` | RANGE / 1488 | RANGE / 1488 | match |
| + selective filter, `head` after `where` | RANGE / 1488 | RANGE / 1488 | match |
| + selective filter, `head` before `where` | RANGE / 385 | RANGE / 385 | match |

Every shape plans a native range on both engines and returns the same count — **including the bare-field control**, so the fold reproduces bare-field semantics rather than merely agreeing with itself. (Latency isn't compared across the two lines here; that isn't apples-to-apples across major versions. Same-engine before/after numbers are in #5680.)

The type-match guard was checked separately on a 3-doc deterministic index, stock vs patched: `timestamp(ts)` / `CAST(ts AS TIMESTAMP)` flip `script` → `range` with unchanged counts, while `date(ts)` and `time(ts)` stay on the script path with unchanged counts.

### 4. Automated review suggestions

- **Explicit parentheses in `canSupport()`** — incorporated. It also documents that `isMultiParameterQuery` is an intentional alternative path rather than an accident of precedence.
- **Null checks on `getFunctionName()` / `inner.type()`** — not taken as written. `x instanceof T` is already `false` for `null`, so `inner.type() != null && inner.type() instanceof ...` is redundant. The `getFunctionName()` concern is now moot: the boolean chain became a map lookup, and an unknown (or null) key simply returns `null`, which is already handled.
- **`unwrapReference` reachable via the `isMultiParameterQuery` branch** — worth a note: base `build()` already assumed a bare reference at operand 0 before this PR (it did a direct `(ReferenceExpression)` cast), so this isn't a new failure mode; the change converts a would-be `ClassCastException` into an explicit `IllegalStateException`. In practice relevance functions override `build()`, so that path isn't reached. Happy to restructure if you'd prefer the precondition enforced differently.

`./gradlew :opensearch:spotlessCheck :opensearch:test` passes on the final state.

### Open question for maintainers

Should the Calcite half ship here or as its own PR? I kept it together because the type-match guard needs to be identical on both paths, but I'm happy to split it if you'd rather review them separately.

dai-chen (commented):
Thanks for the changes! A high level question: I checked https://github.com/grafana/opensearch-datasource/blob/main/pkg/opensearch/client/ppl_request.go but don't see it wraps timestamp field. Could you double check if any simplification or more changes required in this PR?

burnthm:
> Thanks for the changes! A high level question: I checked https://github.com/grafana/opensearch-datasource/blob/main/pkg/opensearch/client/ppl_request.go but don't see it wraps timestamp field. Could you double check if any simplification or more changes required in this PR?

Yes, you're right @dai-chen. The Grafana code points to a different timestamp behavior and `range` that needs proper pushdown, testing, and verification so that it is properly accounted for:
```go
timeFilter := fmt.Sprintf(" where `%s` >= timestamp('%s') and `%s` <= timestamp('%s')", timeField, from, timeField, to)
```
Especially since this was the original issue being raised 😅.

## Correction: Grafana wraps the *bound*, not the field



I originally decoded the `script` type query that results from the Grafana client as: 
- Filter 1: `gte( timestamp(event_time), cast_to_timestamp("2026-08-04 17:42:40") )`
- Filter 2: `lte( timestamp(event_time), cast_to_timestamp("2026-08-04 20:42:40") )`

i.e. a **time-range filter on `event_time` for the dashboard window (now-3h → now)**, where the **field is wrapped in the `timestamp()` builtin**: `timestamp(event_time) >= cast('…17:42:40') AND timestamp(event_time) <= cast('…20:42:40')`Bare backticked field on the left, `timestamp()` around the literal on the right. 

I re-decoded the serialized script from example slow logs running the sample query and walked the resulting decoded `script_query` object graph in stream order: element 0 of the comparison's argument list is a plain `ReferenceExpression(attr="event_time", OpenSearchDateType/TIMESTAMP)` — not a function wrapping it. The `timestamp(cast_to_timestamp("2026-08-04 17:42:40"))` chain is element 1. So what actually ran was:

```
event_time>= timestamp(cast_to_timestamp('2026-08-04 17:42:40'))
```

which matches the `ppl_request.go` client reference. My earlier claim that the data source wraps the date field was a misreading/mis-decoding of that payload. The issue description has been corrected.

## Why it still scripted — and what I'd missed

With a bare field on the left, the blocker is the **value operand**. `canSupport()` accepts it only via `literalExpressionWrappedByCast`, which requires the function to be in `castMap` (keys are `CAST_TO_*` only — `timestamp` is not among them) **and** its argument to be a bare `LiteralExpression`. PPL coerces the string first, so the operand is `timestamp(cast_to_timestamp(literal))` and fails both halves → `buildScriptQuery`.

I confirmed on a 2.19 node with only the original (field-side) fold applied:

| predicate | plan |
|---|---|
| `` `ts` >= timestamp('…') `` — client-generated shape | **SCRIPT** |
| `timestamp(ts) >= cast('…' as timestamp)` — what the fold covered | RANGE |
| `ts >= '…'` (control) | RANGE |

So your question surfaced an important gap: the PR did not cover the shape that caused the reported impact. The reported symptom was unchanged (the CPU is burned by scripted `castToTimestamp` per document either way), but the fix was landing on the wrong operand.

## What the PR now contains — two fixes, one per operand

1. **Field operand** — fold a redundant conversion on the filtered field (`timestamp(<field>)` / `CAST(<field> AS TIMESTAMP)` where the field is already that type). Applies to both engines, since the v2 change alone has no effect on 3.x.
2. **Value operand** (new commit) — resolve a bound wrapped in a date conversion. Accepts the `timestamp()`/`date()`/`time()` builtins alongside `CAST_TO_*`, and because applying `timestamp()` to an already-timestamp value is a no-op, `timestamp(cast_to_timestamp(literal))` evaluates the inner cast alone. The cast is *resolved* rather than evaluated directly so `castMap` keeps parsing the literal against the field's declared date formats. This one is v2-only — Calcite already pushed the shape down, consistent with the regression being reported on the v2 engine.

Verified on a real cluster: `birthdate > timestamp('2016-12-08 00:00:00')` now emits two native `range` queries on `birthdate` with no script clause — byte-identical to the plan for the bare-field form. Pinned by `ExplainIT.testFilterTimestampWrappedBoundPushDownExplain`, which is inherited by `CalciteExplainIT` and `CalciteNoPushdownIT`, so it covers v2, Calcite-with-pushdown, and Calcite-without-pushdown.

## On simplification

Both folds are deliberately narrow, because two earlier iterations of the redundancy check were wrong and I'd rather the guards stay explicit:

- keying off the result type alone folded `LAST_DAY` (single date argument, returns `DATE`) over a `DATE`-typed field, turning `last_day(d) = '2024-01-31'` into a range on the raw field and returning 1 row where 2 are correct;
- not requiring the conversion target to match the operand type folded `date(<timestamp field>)`, turning `date(ts) <= '2024-01-15'` into `ts <= '2024-01-15'` — 0 rows where 2 are correct.

burnthm:
A few followup notes on https://github.com/opensearch-project/sql/pull/5681/commits/04be09594e80e9d49017b36b3702ff6b238c6f00. 

1. Since the bound fix resolves `timestamp()` via the equivalent `CAST_TO_TIMESTAMP` handler, I checked
that equivalence directly: evaluating both functions on the same literals on an unpatched node gives
identical results for all 9 inputs I tried (canonical, date-only, sub-second, ISO, minute-precision,
malformed, numeric), including identical exceptions. The two-argument `timestamp(a, b)` form is a
different function and stays on the script path.

2. One thing worth flagging: `RangeQuery.value()` renders a date bound using the field's first declared
format, so when that format has no sub-second component the bound loses precision. On an unpatched
node a `'…12:00:00.123'` bound already returns the wrong count for the plain-string and `cast()`
forms; `timestamp()` avoided it only by falling back to the script. The changes here bring the behavior of pattern, ``ts` >= timestamp('…')`, in line with other pushed down patterns also using `range`.

3. The only behavior change beyond the plan is a TIME-typed field: `t >= time(...)`
fails on an unpatched node with `invalid to get timestampValue from value of type TIME` and now
returns the correct count. And a literal-first comparison (`timestamp('…') <= ts`) still scripts,
because v2 selects the query by function name and assumes the field is the first operand — supporting
it means inverting the comparison, which Calcite already does via `swap()`.

opensearch/src/main/java/org/opensearch/sql/opensearch/request/PredicateAnalyzer.java
```diff
@@ -365,6 +372,14 @@ static RexUnknownAs getNullAsForSearch(RexCall search) {
 
     @Override
     public Expression visitCall(RexCall call) {
+      // Fold a redundant date/time cast over a field of the same date/time type -- e.g.
+      // timestamp(<timestamp field>) or CAST(<timestamp field> AS TIMESTAMP) -- to the bare field
+      // reference. Such a wrap is a no-op, so the enclosing comparison can push down to a native
+      // range query instead of falling back to a per-document script.
+      if (isRedundantDateCastOverField(call)) {
```
dai-chen, line 379: I think for the redundant cast on field, it should be done earlier, e.g., V3's CalciteRel/RexNodeVisitor or V2's Analyzer as part of logical planning/optimization. For V3, is this not optimized due to time related type are UDT in our engine? Because I thought this optimization should be in Calcite builtin rule.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -105,7 +238,7 @@ protected boolean literalExpressionWrappedByCast(FunctionExpression func) {
    * @return query
    */
   public QueryBuilder build(FunctionExpression func) {
-    ReferenceExpression ref = (ReferenceExpression) func.getArguments().get(0);
+    ReferenceExpression ref = unwrapReference(func.getArguments().get(0));
```
dai-chen, line 241: Same as commented above and should be done earlier I think.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -86,15 +87,147 @@ private boolean isMultiParameterQuery(FunctionExpression func) {
...
+            .get(name)
+            .equals(DATE_CAST_TARGET_TYPES.get(innerFn.getFunctionName()))) {
+      return Map.entry(innerFn.getFunctionName(), literal);
     }
-    return false;
+    return null;
+  }
+
+  /** Date/time cast functions mapped to the type each one produces. */
+  private static final Map<FunctionName, ExprCoreType> DATE_CAST_TARGET_TYPES =
```
dai-chen, line 175: Create single BiMap instead since this is small.

opensearch/src/main/java/org/opensearch/sql/opensearch/storage/script/filter/lucene/LuceneQuery.java
```diff
@@ -86,15 +87,147 @@ private boolean isMultiParameterQuery(FunctionExpression func) {
...
+   * <p>Besides the plain {@code CAST_TO_*(literal)} form, this accepts the {@code
+   * timestamp()}/{@code date()}/{@code time()} builtins, which are what clients typically emit for
+   * a time-range bound (for example the Grafana OpenSearch data source builds its PPL time filter
+   * as {@code where `field` >= timestamp('...')}). PPL coerces the string argument first, so the
+   * operand arrives as {@code timestamp(cast_to_timestamp('...'))}; applying {@code timestamp()} to
+   * a value that is already a timestamp is a no-op, so the inner cast alone is evaluated. The cast
+   * is resolved rather than evaluated here so that {@link #castMap} keeps parsing the literal
+   * against the field's declared date formats.
+   */
+  private Map.Entry<FunctionName, LiteralExpression> resolveLiteralOperand(
```
dai-chen, line 138: Could you remind me what's the gap in the existing unwrap logic in `cast` method? Can we consider simplifying the fix or extracting this to a new class?

PR #5683, opened 2026-08-05
Strengthen AnalyticsEngineSecurityIT coverage and fix routing

finnegancarroll:
### Routing fix for analytics engine security plugin tests

The SQL plugin routes queries to the analytics engine based on `isAnalyticsIndex()`, which checks whether the target index uses the composite data format. Previously, this check only worked for **concrete index names** — it performs a direct `Metadata.index(name)` lookup which does not resolve wildcards, aliases, or multi-index expressions. Queries using `source = analytics_*` or `source = my_alias` would silently fall back to the legacy PPL backend because the lookup returned null for non-concrete names.

This meant wildcard, alias, and some multi-index security tests were passing but testing the **legacy PPL path's** security behavior, not the analytics engine's. The concrete-index tests (e.g., `testPPLQueryDeniedWithSearchPermissionOnly`) correctly validated analytics engine routing, but the wildcard/alias ALLOW tests masked the routing gap with `try/catch` fallbacks.

The fix adds `cluster.pluggable.dataformat=composite` (plus `cluster.composite.secondary_data_formats=["lucene"]`) to the `testClusters.analyticsEngineSecurityIT` configuration. This activates the cluster-level fast path in `isAnalyticsIndex()`, ensuring ALL queries route through the analytics engine regardless of source format.

To prevent future silent fallback regressions, we add three **routing guard tests** that use `profile=true`. The analytics engine returns a `"profile"` key containing engine-specific execution details (stage timings, physical plans, DataFusion metrics) that are distinct from the legacy PPL profile output. If a query silently falls back to the legacy path, the response structure will differ and these tests will fail:

| Test | What it guards |
|------|---------------|
| `testRoutingGuardConcreteIndex` | Concrete index name routing |
| `testRoutingGuardWildcardSource` | Wildcard source routing |
| `testRoutingGuardAlias` | Alias source routing |

### Test assertion enhancements

Removed `try/catch(ResponseException)` fallbacks that previously masked non-403 errors (including 500s) as passing tests. ALLOW tests now assert HTTP 200, validate non-empty `datarows`, and check actual row content (e.g., `assertContainsName("alice")`). DENY tests assert 403 status AND validate that the error body references the denied action or contains "no permissions". Also validates that 403 responses do not leak internal stack traces or node IDs.

### New test coverage

| Test | Description |
|------|-------------|
| `testRoutingGuardConcreteIndex` | Proves analytics engine handles concrete index queries (profile key present) |
| `testRoutingGuardWildcardSource` | Proves analytics engine handles wildcard source queries |
| `testRoutingGuardAlias` | Proves analytics engine handles alias queries |
| `testPPLCommaSourceAllAuthorized` | Comma-separated source, user authorized on all indices |
| `testPPLCommaSourcePartiallyAuthorized` | Comma-separated source with one forbidden index → deny |
| `testPPLMultiIndexDeniedWithBackticksAuthorizedFirst` | Backtick-quoted index names must not bypass FGAC |
| `testPPLMultiIndexDeniedWithUnauthorizedFirst` | Unauthorized index first in comma list must still deny |
| `testPPLQueryAllowedViaConcreteIndexForAliasUser` | Alias grant implies concrete index access |
| `testPPLQueryDeniedWithoutClusterPermission` | Missing cluster:admin/opensearch/ppl → 403 |
| `testPPLQueryMultiBackingAliasAllowed` | Multi-backing alias, user covers all backing indices |
| `testPPLQueryMultiBackingAliasDenied` | Multi-backing alias, user has no access |
| `testPPLQueryMultiBackingAliasPartialAccessDenied` | Multi-backing alias, user covers only some backing indices |
| `testPPLQueryWithWildcardSourceAllowed` | Wildcard source expanding to authorized indices only |
| `testPPLQueryWithWildcardSourceDenied` | Wildcard source, user unauthorized on all matching indices |
| `testPPLQueryWithWildcardSourcePartialAccessDenied` | Wildcard expands to include forbidden index → deny |
| `testSQLQueryAllowedWithWildcardPermission` | SQL parity: wildcard role permission matches concrete index |
| `testSQLQueryDeniedWithWildcardPermissionOnNonMatchingIndex` | SQL parity: wildcard role does not match forbidden index |
| `testSQLQueryAllowedViaAlias` | SQL parity: alias as table name |
| `testSQLQueryWithExactAnalyticsPermission` | SQL parity: exact analytics/query permission is sufficient |
| `testDeniedResponseContainsActionName` | 403 body references the denied action for user actionability |
| `testDeniedResponseDoesNotLeakInternalDetails` | 403 body contains no stack traces or internal node info |

### DIFF
--- a/integ-test/src/test/java/org/opensearch/sql/security/AnalyticsEngineSecurityIT.java
+++ b/integ-test/src/test/java/org/opensearch/sql/security/AnalyticsEngineSecurityIT.java
@@ -7,17 +7,24 @@ package org.opensearch.sql.security;
 
 import java.io.IOException;
 import java.util.Locale;
+import org.json.JSONArray;
 import org.json.JSONObject;
 import org.junit.Test;
 import org.opensearch.client.Request;
 import org.opensearch.client.RequestOptions;
 import org.opensearch.client.Response;
 import org.opensearch.client.ResponseException;
+import org.opensearch.sql.legacy.TestUtils;
 
 /**
  * Integration tests for analytics engine index-level authorization via the production SQL plugin
  * PPL endpoint. Verifies that queries on composite (analytics-engine-backed) indices are subject to
  * the {@code indices:data/read/analytics/query} permission check.
+ *
+ * <p>IMPORTANT: The testClusters config sets {@code cluster.pluggable.dataformat=composite} so that
+ * ALL queries route through the analytics engine regardless of how the source is specified
+ * (wildcard, alias, comma-separated). Without this, {@code isAnalyticsIndex()} only recognizes
+ * concrete index names via per-index metadata lookup.
  */
 public class AnalyticsEngineSecurityIT extends SecurityTestBase {
 
@@ -25,6 +32,7 @@ public class AnalyticsEngineSecurityIT extends SecurityTestBase {
   private static final String FORBIDDEN_INDEX = "analytics_forbidden_test";
   private static final String TEST_INDEX_2 = "analytics_security_extra";
   private static final String TEST_ALIAS = "analytics_alias";
+  private static final String MULTI_ALIAS = "analytics_multi_alias";
 
   private static final String ALLOWED_USER = "analytics_allowed_user";
   private static final String ALLOWED_ROLE = "analytics_allowed_role";
@@ -38,6 +46,10 @@ public class AnalyticsEngineSecurityIT extends SecurityTestBase {
   private static final String ALIAS_ROLE = "analytics_alias_role";
   private static final String EXACT_PERM_USER = "analytics_exact_perm_user";
   private static final String EXACT_PERM_ROLE = "analytics_exact_perm_role";
+  private static final String NO_CLUSTER_PERM_USER = "analytics_no_cluster_perm_user";
+  private static final String NO_CLUSTER_PERM_ROLE = "analytics_no_cluster_perm_role";
+  private static final String MULTI_ALIAS_USER = "analytics_multi_alias_user";
+  private static final String MULTI_ALIAS_ROLE = "analytics_multi_alias_role";
 
   private static boolean initialized = false;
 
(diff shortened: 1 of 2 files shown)


integ-test/src/test/java/org/opensearch/sql/security/AnalyticsEngineSecurityIT.java
```diff
@@ -240,225 +283,336 @@ public void testPPLQueryDeniedForForbiddenIndex() throws IOException {
...
+  // is entirely REST-based with no cluster-level transport action. This is pre-existing behavior
+  // in both the legacy SQL path and the analytics engine SQL path.
+
+  // ===========================================================================================
+  // ROUTING GUARD TESTS — Prove queries execute on analytics engine, not legacy PPL
+  // ===========================================================================================
+
+  @Test
+  public void testRoutingGuardConcreteIndex() throws IOException {
+    // The analytics engine's profile output contains a "plan" object with "query_id" — a field
```
dai-chen, line 495: Shall we do the same for AE SQL path?

dai-chen (approved):
Minor comment. Thanks for the changes!

integ-test/src/test/java/org/opensearch/sql/security/AnalyticsEngineSecurityIT.java
finnegancarroll: thanks for taking a look! added sql routing guards for these cases as well.

dai-chen:
Retrying failed AE CI which seems failed transiently.

dai-chen:
@finnegancarroll The AE CI is failing on error below consistently. It seems unrelated to this PR but do we have plan to fix?

```
> A failure occurred while executing de.undercouch.gradle.tasks.download.internal.DefaultWorkerExecutorHelper$DefaultWorkAction
   > de.undercouch.gradle.tasks.download.org.apache.hc.client5.http.ClientProtocolException: Forbidden (HTTP status code: 403, URL: https://ci.opensearch.org/ci/dbc/feature-build-opensearch/feature-datafusion/latest/linux/x64/tar/builds/opensearch/plugins/1-analytics-engine-3.9.0.zip)
```

finnegancarroll:
> @finnegancarroll The AE CI is failing on error below consistently. It seems unrelated to this PR but do we have plan to fix?

It looks like it's failing to pull the 3.9 AE artifacts published by core. I see core is on 3.9 so maybe bad timing and just needs another retry.

PR #5695, opened 2026-08-10
Lower frontend-private rel nodes before the analytics-engine handoff (fix dedup 500)

ahkcs:
### Description

PPL `dedup` followed by any pipe stage fails with a 500 on the analytics-engine route:

```
source=idx | dedup category | fields category
→ 500 RuntimeException: There was internal problem at backend
```

Server-side: `IllegalStateException: Project rule encountered unmarked child [LogicalDedup]` at `OpenSearchProjectRule.onMatch`. Deterministic for every `dedup … | <anything>` shape.

Resolves #22671.

> **Supersedes [opensearch-project/OpenSearch#22675](https://github.com/opensearch-project/OpenSearch/pull/22675).** That PR fixed the same issue on the analytics-engine side by having the engine reflect over foreign rel nodes to discover and run their convert rules. Per review feedback, the frontend should own its own rules rather than the engine repairing plans it receives — this PR takes that direction. The companion QA IT (`DedupCommandIT`) will land as a small OpenSearch-core PR.

#### Root cause

`dedup` is planned as a `ROW_NUMBER() OVER (PARTITION BY keys)` + Filter composite. `CalciteToolsHelper.optimize` — run on both the Lucene and analytics routes — then collapses that composite into a `LogicalDedup` via `PPLSimplifyDedupRule`, so that `DedupPushdownRule` can push dedup into a Lucene scan.

On the **Lucene path**, physical planning (Volcano) lowers the `LogicalDedup` back (or pushes it into the scan), so it disappears. On the **analytics path**, the optimized `RelNode` goes straight to the analytics engine with no Volcano step, so the `LogicalDedup` survives to the engine's marking phase, which rejects it as an unmarked child.

`LogicalDedup` exists solely to enable the Lucene pushdown (`DedupPushdownRule` is its only consumer); the analytics engine has no equivalent pushdown and cannot plan the node.

#### Fix

Add `CalciteToolsHelper.optimizeForAnalytics`, which runs the same HEP rules as `optimize` **minus `PPLSimplifyDedupRule`**. `RestUnifiedQueryAction` calls it on the analytics route, so the dedup composite is never collapsed and the analytics engine receives the standard `ROW_NUMBER` window form it can plan directly.

- **No collapse-then-expand.** Rather than generate a `LogicalDedup` and convert it back, the analytics route simply doesn't generate it — the rule that creates it (only useful for the Lucene pushdown) is skipped.
- **Lucene path untouched.** `optimize` is unchanged; the Lucene route keeps `PPLSimplifyDedupRule` and its dedup pushdown.
- **Script-based thread-pool routing preserved.** Optimization still runs before dispatch (`ScriptDetector` inspects the optimized plan), so worker-pool routing is unaffected.

### Testing

| test | before | after |
|---|---|---|
| `CalcitePPLDedupTest.testOptimizeForAnalyticsKeepsRowNumberForm` (`optimize` produces `LogicalDedup`; `optimizeForAnalytics` keeps `ROW_NUMBER`, no `LogicalDedup`, predicate preserved) | — | pass |
| `CalcitePPLDedupTest.testOptimizeForAnalyticsMatchesOptimizeWithoutDedup` (dedup-free plan identical under both variants) | — | pass |
| `CalcitePPLDedupTest` (full class, regression) | 12/12 | 14/14 pass |
| Live analytics-engine cluster, parquet-backed index: `dedup category \| fields category`, `dedup category \| stats count()`, `dedup 2 category \| fields category` | 500 `unmarked child [LogicalDedup]` | 200, correct rows |

### Check List
- [x] New functionality has been documented (method Javadoc + PR).

### DIFF
--- a/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
+++ b/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
@@ -234,7 +234,7 @@ public class RestUnifiedQueryAction {
                     plan = addFetchSizeLimit(plan, planContext, fetchSize);
                     plan = addQuerySizeLimit(plan, planContext);
                     plan =
-                        org.opensearch.sql.calcite.utils.CalciteToolsHelper.optimize(
+                        org.opensearch.sql.calcite.utils.CalciteToolsHelper.optimizeForAnalytics(
                             plan, planContext);
                     RelNode finalPlan = plan;
                     Runnable executeTask =
--- a/core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
+++ b/core/src/main/java/org/opensearch/sql/calcite/utils/CalciteToolsHelper.java
@@ -556,9 +556,27 @@ public class CalciteToolsHelper {
   private static final HepProgram HEP_PROGRAM =
       new HepProgramBuilder().addRuleCollection(hepRuleList).build();
 
+  // PPLSimplifyDedupRule collapses the ROW_NUMBER window form of dedup into a LogicalDedup so
+  // DedupPushdownRule can push it into a Lucene scan. The analytics engine has no such pushdown and
+  // cannot plan a LogicalDedup, so its optimization runs the standard window form directly.
+  private static final HepProgram ANALYTICS_HEP_PROGRAM =
+      new HepProgramBuilder()
+          .addRuleCollection(
+              hepRuleList.stream()
+                  .filter(rule -> rule != PPLSimplifyDedupRule.DEDUP_SIMPLIFY_RULE)
+                  .toList())
+          .build();
+
   public static RelNode optimize(RelNode plan, CalcitePlanContext context) {
     Util.discard(context);
     HepPlanner planner = new HepPlanner(HEP_PROGRAM);
+    planner.setRoot(plan);
+    return planner.findBestExp();
+  }
+
+  public static RelNode optimizeForAnalytics(RelNode plan, CalcitePlanContext context) {
+    Util.discard(context);
+    HepPlanner planner = new HepPlanner(ANALYTICS_HEP_PROGRAM);
     planner.setRoot(plan);
     return planner.findBestExp();
   }
--- a/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLDedupTest.java
+++ b/ppl/src/test/java/org/opensearch/sql/ppl/calcite/CalcitePPLDedupTest.java
@@ -14,6 +14,7 @@ import org.apache.calcite.test.CalciteAssert;
 import org.junit.Assert;
 import org.junit.Test;
 import org.opensearch.sql.calcite.plan.rule.PPLSimplifyDedupRule;
+import org.opensearch.sql.calcite.utils.CalciteToolsHelper;
 
 public class CalcitePPLDedupTest extends CalcitePPLAbstractTest {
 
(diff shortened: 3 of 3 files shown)


core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsPlanLowering.java
```diff
@@ -0,0 +1,57 @@
...
+ * CalciteToolsHelper.optimize}. On the Lucene/Calcite path Volcano's {@code RelNode.register} hook
+ * then fires the matching convert/pushdown rules and the node disappears. The analytics route never
+ * runs that Volcano step, so the node would survive to the analytics-engine marking phase, which
+ * rejects it as an "unmarked child".
+ *
+ * <p>Rather than have the analytics engine reflect over foreign nodes to discover their rules, the
+ * SQL plugin owns this contract: it explicitly runs its own convert rules here, so the analytics
+ * engine only ever receives standard Calcite nodes. {@link #LOWERING_RULES} is the explicit
+ * registry — add a rule when a new frontend-private rel needs a standard-shape lowering on the
+ * analytics route.
```
penghuo: Remove AI wording. It explain why did it in PPL, not analytics.

core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsPlanLowering.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.hep.HepPlanner;
+import org.apache.calcite.plan.hep.HepProgram;
+import org.apache.calcite.plan.hep.HepProgramBuilder;
+import org.apache.calcite.rel.RelNode;
+import org.opensearch.sql.calcite.plan.rule.PPLDedupConvertRule;
+
+/**
+ * Lowers frontend-private rel nodes into standard Calcite shapes before a plan is handed to the
```
penghuo: What is Lower? Standarding naming?

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -236,6 +237,10 @@ private void doExecute(
                     plan =
                         org.opensearch.sql.calcite.utils.CalciteToolsHelper.optimize(
                             plan, planContext);
+                    // optimize() collapses dedup back into a frontend-private LogicalDedup (for the
+                    // Lucene pushdown path); the analytics engine can't mark that node, so lower it
+                    // (and any other frontend-private rel) into a standard Calcite shape here.
+                    plan = AnalyticsPlanLowering.lower(plan);
```
penghuo: LogicalDedup is added becuase Pushdown needed it? If true, The other options is in RelNodeVisitor avoid generated LogicalDedup when use analytics engine.

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
```diff
@@ -236,6 +237,9 @@ private void doExecute(
                     plan =
                         org.opensearch.sql.calcite.utils.CalciteToolsHelper.optimize(
                             plan, planContext);
+                    // Lower PPL rel nodes (e.g. LogicalDedup) that the analytics engine cannot
+                    // mark.
+                    plan = AnalyticsPlanLowering.lower(plan);
```
dai-chen: 1. At line 237, we've triggered optimizations already?
2. I think both lines changed the previous boundary (SQL/PPL generates raw plan and backend like AE responsible for optimizations). Shall we move both into AnalyticsExecutionEngine and later if needed explicitly call new UnifiedQueryPlanner.optmize or UnifiedQueryOptimizer API?

mch2: I'd like to re-think our SQL → AE integration. We've accumulated a lot of one-offs in AE, substrait adapters that exist to accommodate PPL specific operators and a test time circular dependency since AE's integration tests now need the SQL plugin installed to produce plans and test e2e.

Thinking AE should publish a contract and have front-ends like SQL/PPL compile and optimize to it. AE validates against it at the boundary and rejects up front instead of accepting whatever arrives and failing late in substrait conversion or out on a data node.

core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsPlanLowering.java
ahkcs: removed

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Good call — done. Rather than generate a `LogicalDedup` and convert it back, the analytics route now just doesn't generate it: `PPLSimplifyDedupRule` only exists to collapse the `ROW_NUMBER` composite into a `LogicalDedup` for the Lucene `DedupPushdownRule`, and the analytics engine has no such pushdown. Added `CalciteToolsHelper.optimizeForAnalytics` (the same rules minus that one) and call it on the analytics route, so the engine gets the standard `ROW_NUMBER` window form directly.

core/src/main/java/org/opensearch/sql/executor/analytics/AnalyticsPlanLowering.java
ahkcs: The Lower naming is gone. Per your other comment, the analytics route now just skips the rule that generates `LogicalDedup` instead of converting it back

plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java
ahkcs: Reworked since your comment — dropped lowering; the route now calls `optimizeForAnalytics` (rules minus PPLSimplifyDedupRule), so no `LogicalDedup` is generated.

Line 237 already optimized (predates this PR, #5656 — just swapped it). And it can't move into `AnalyticsExecutionEngine` yet — dispatchTask() routes thread pools via ScriptDetector on the optimized plan before execute(). 

Agree the boundary cleanup + Marc's AE-contract idea are worth a follow-up.

penghuo (approved):

PR #5696, opened 2026-08-11
fix: Prevent InvalidTypeIdException when Prometheus metric has 'type'…

### LINKED ISSUE #5684: [BUG] PromQL queries fail with InvalidTypeIdException when metric has a label named "type"
**What is the bug?**
Any PromQL query that returns metrics containing a label named `"type"` fails with:
```
Could not resolve subtype of [simple type, class org.opensearch.sql.directquery.transport.model.datasource.PrometheusResult]: missing type id property 'type'
```
For example, `avg by (type) (some_metric)` or simply querying any metric that has a `type` label in its label set causes the error. All other label names work correctly.
The root cause is in `DataSourceResult.java` (line 17):
```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
```
This Jackson annotation declares `"type"` as the polymorphic type discriminator for `DataSourceResult` subtypes. When the Prometheus response contains a metric label called `"type"` anywhere in the JSON structure (e.g., `{"metric": {"type": "some_value"}, "values": [...]}`), Jackson mistakenly interprets it as the type discriminator instead of treating it as data, causing deserialization to fail.
1. Set up a Prometheus data connection in OpenSearch
2. Have any metric with a label named `type` (e.g., via relabeling, OTel transform processor, or native metric exposition)
3. In OpenSearch Dashboards, open the Metrics Explorer or PromQL editor
4. Execute a query that includes the `type` label in results, e.g.:
   - `avg by (type) (my_metric)`
   - Or simply select a metric that has a `type` label
5. Observe the `InvalidTypeIdException` error
Reproduced with both VictoriaMetrics and vanilla Prometheus as backends.
The query should execute successfully and return results with the `type` label as a regular metric label, just like any other label name (e.g., `instance`, `job`, `container`). The label name `"type"` is a perfectly valid Prometheus label and should not conflict with internal deserialization mechanics.
- OpenSearch version: 3.7.0
- Tested with: Prometheus (vanilla) and VictoriaMetrics as data sources
- OpenSearch Dashboards with `query_enhancements` plugin
**Affected file:** `direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/datasource/DataSourceResult.java`
```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
@JsonSubTypes({@JsonSubTypes.Type(value = PrometheusResult.class, name = "prometheus")})
public interface DataSourceResult {}
```
**Proposed fix (identified with assistance of AI):**
Change the `property` value in the `@JsonTypeInfo` annotation to a name that cannot collide with valid Prometheus label names. For example:
```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "@datasource_type")
@JsonSubTypes({@JsonSubTypes.Type(value = PrometheusResult.class, name = "prometheus")})
public interface DataSourceResult {}
```
Alternative approaches:
- Use `JsonTypeInfo.As.WRAPPER_OBJECT` instead of `As.PROPERTY` to avoid any flat-field collision
- Use `JsonTypeInfo.As.EXISTING_PROPERTY` with an explicit dedicated field not exposed in the metric labels
The key constraint is that `"type"` is a valid and commonly-used Prometheus metric label (e.g., `kube_pod_info{type="..."}`) and must not be reserved by the deserialization framework.

nagendramohan:
… label

Add @JsonTypeInfo(use = JsonTypeInfo.Id.NONE) on PrometheusResult to override the parent DataSourceResult interface's polymorphic type handling. This prevents Jackson from interpreting a metric label named 'type' as the polymorphic type discriminator, which caused InvalidTypeIdException during deserialization.

The type dispatch is already handled explicitly via switch statement in ExecuteDirectQueryActionResponse, so polymorphic type annotations are not needed on the concrete class.

Resolves opensearch-project#5684

### Description
  PromQL queries fail with `InvalidTypeIdException` when a metric contains a label named `type`. This happens because the `DataSourceResult`
  interface uses `@JsonTypeInfo(property = "type")` for polymorphic deserialization, and Jackson interprets the metric's `type` label as the
  type discriminator.
  
  **Fix:** Add `@JsonTypeInfo(use = JsonTypeInfo.Id.NONE)` on `PrometheusResult` to override the parent's polymorphic type handling. The type
  dispatch is already handled explicitly via switch statement in `ExecuteDirectQueryActionResponse`, so annotation-based polymorphism is unnecessary on the concrete class.
  
 Includes a regression test with a metric containing `"type": "gauge"` label.
  
  ### Related Issues
  Resolves #5684
  
  ### Check List
  - [ ] API changes companion pull request
  [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
  
  For more information on following Developer Certificate of Origin and signing off your commits, please check
  [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/ExecuteDirectQueryActionResponse.java
+++ b/direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/ExecuteDirectQueryActionResponse.java
@@ -131,19 +131,11 @@ public class ExecuteDirectQueryActionResponse extends ActionResponse {
     Map<String, DataSourceResult> parsedResults = new HashMap<>();
 
     try {
-      // Add type to JSON if it doesn't already have it
-      final String resultWithType;
-      if (!rawResult.contains("\"type\":")) {
-        resultWithType = addTypeFieldToJson(rawResult, dataSourceType);
-      } else {
-        resultWithType = rawResult;
-      }
-
       DataSourceResult result;
       // Parse based on the determined data source type
       switch (dataSourceType.toLowerCase()) {
         case "prometheus":
-          result = OBJECT_MAPPER.readValue(resultWithType, PrometheusResult.class);
+          result = OBJECT_MAPPER.readValue(rawResult, PrometheusResult.class);
           break;
           // Add cases for other data source types as they're implemented
         default:
@@ -162,16 +154,5 @@ public class ExecuteDirectQueryActionResponse extends ActionResponse {
     }
 
     return parsedResults;
-  }
-
-  /**
-   * Adds a type field to the JSON string for proper polymorphic deserialization.
-   *
-   * @param rawJson The raw JSON string without a type field
-   * @param type The type to add
-   * @return Modified JSON string with type field
-   */
-  private String addTypeFieldToJson(String rawJson, String type) {
-    return rawJson.replaceFirst("\\{", "{\"type\":\"" + type + "\",");
   }
 }
--- a/direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/datasource/DataSourceResult.java
+++ b/direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/datasource/DataSourceResult.java
@@ -5,15 +5,14 @@
 
 package org.opensearch.sql.directquery.transport.model.datasource;
 
-import com.fasterxml.jackson.annotation.JsonSubTypes;
-import com.fasterxml.jackson.annotation.JsonTypeInfo;
-
 /**
  *
  * @opensearch.experimental
  *
  * Interface for results from various data sources.
+ *
+ * <p>Concrete result types are dispatched by the {@code dataSourceType} string carried alongside
+ * the serialized payload in the transport protocol (see {@code ExecuteDirectQueryActionResponse}),
+ * so no in-JSON Jackson type discriminator is used here.
  */
-@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
-@JsonSubTypes({@JsonSubTypes.Type(value = PrometheusResult.class, name = "prometheus")})
 public interface DataSourceResult {}
--- a/direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/datasource/PrometheusResult.java
+++ b/direct-query/src/main/java/org/opensearch/sql/directquery/transport/model/datasource/PrometheusResult.java
@@ -7,6 +7,7 @@ package org.opensearch.sql.directquery.transport.model.datasource;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import java.util.List;
 import java.util.Map;
@@ -22,6 +23,7 @@ import lombok.Setter;
 @Getter
 @Setter
 @JsonTypeName("prometheus")
+@JsonTypeInfo(use = JsonTypeInfo.Id.NONE)
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class PrometheusResult implements DataSourceResult {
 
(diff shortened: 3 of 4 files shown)


nagendramohan:
Friendly ping — this fixes a crash when Prometheus metrics contain a field named type (issue #5684). Small change: one annotation + regression test. Happy to address any feedback.

lezzago:
### Code review — findings

The fix (adding `@JsonTypeInfo(use = JsonTypeInfo.Id.NONE)` to `PrometheusResult` to stop Jackson from treating a metric's `type` label as a polymorphic discriminator) is functionally correct for the reported case, and the regression test validates it. A few points worth considering before merge — the first two are the substantive ones.

---

**1. Transport round-trip becomes asymmetric across mixed-version nodes (correctness).**
`writeTo()` serializes via `OBJECT_MAPPER.writeValueAsString(result)`. Previously `PrometheusResult` inherited `@JsonTypeInfo(As.PROPERTY, property="type")` from `DataSourceResult`, so serialization emitted a root `"type":"prometheus"`. With `Id.NONE`, a patched node now emits **no** root `type`. But the `StreamInput` constructor still reads via `readValue(resultJson, PrometheusResult.class)` — the same polymorphic path that made `addTypeFieldToJson(...)` necessary in the first place. So during a rolling upgrade, a **new** node's payload (no root `type`) fed to an **old** node fails deserialization (`missing type id` / `InvalidTypeIdException`) → `RuntimeException: Failed to deserialize Prometheus result`.

Scope: one-directional (new→old only; old→new is fine since `Id.NONE` + `ignoreUnknown` drop the stray root `type`), and this is `@opensearch.experimental`. Bounded, but worth a conscious call.

**2. The fix leaves the polymorphism/injection machinery as dead code.**
With `Id.NONE`, the root `type` injected by `addTypeFieldToJson(...)` becomes an unknown property that `@JsonIgnoreProperties(ignoreUnknown = true)` silently drops — so the injection has no effect. Both `readValue` sites target `PrometheusResult.class` directly (never the `DataSourceResult` interface), which means `@JsonSubTypes` / `@JsonTypeName` and the interface's `@JsonTypeInfo` are now vestigial too. A cleaner fix would remove the polymorphism + `addTypeFieldToJson` scaffolding entirely rather than layer `Id.NONE` on top of it. (Non-blocking, but it removes a latent trap for the next maintainer.)

**3. The `contains("\"type\":")` guard is a fragile whole-string substring scan** (`ExecuteDirectQueryActionResponse#parseResult`). Any metric label named `type` flips the branch — it's harmless now only because of point 2, but it no longer expresses meaningful intent. Cleaning up point 2 removes this too.

**4. Test coverage gap.** The new regression test exercises only the raw-string constructor → `parseResult` path. It doesn't cover `writeTo()` → `new ExecuteDirectQueryActionResponse(StreamInput)`, which is exactly the path affected by point 1. A round-trip test with a `type` label in the metric would give real coverage there.

**5. Nit:** the test file still ends without a trailing newline (`\ No newline at end of file`). This is pre-existing on `main`, not introduced here, but since the file is being touched anyway, `./gradlew spotlessApply` would clean it up.

nagendramohan:
Thanks for the thorough review @lezzago — the breakdown was really helpful.

I ended up taking your suggestion in #2 and reworking the fix rather than layering Id.NONE on top of the existing machinery. The new commit removes the polymorphism scaffolding entirely:

- Dropped @JsonTypeInfo/@JsonSubTypes from DataSourceResult and @JsonTypeName/@JsonTypeInfo(Id.NONE) from PrometheusResult
- Deleted addTypeFieldToJson() and the contains("\"type\":") guard, and now read PrometheusResult directly

The reason this is safe: the concrete result type is already carried as the separate dataSourceType string and dispatched via the switch in both
writeTo() and the StreamInput constructor, so the in-JSON type discriminator was never actually exercised through the interface. Removing it is what
resolves #2 and #3 — the fragile substring guard goes away with it.

For #4, I added a writeTo() → StreamInput round-trip test (testStreamSerializationPreservesTypeMetricLabel) that carries a metric label literally named type, so the transport path is now covered directly. And #5 — ran spotlessApply.

That leaves your #1, which I want to be upfront about. This does change the serialized shape (no root type in the result JSON), so there's a bounded,
one-directional mixed-version window during a rolling upgrade: a new node's payload sent to an old node could fail to deserialize (old → new is fine,
since the stray type is just ignored). It only affects the result-JSON shape — dispatch is unaffected because it keys off the separate dataSourceType
string — and the API is @opensearch.experimental.

My read is that's an acceptable tradeoff given the experimental status, but I'd rather not make that call unilaterally. Are you comfortable with it
as-is, or would you prefer I add a compatibility guard (e.g. keep tolerating/emitting the root type for a transition period)? Happy to go either way.

nagendramohan:
@lezzago Can you please review this one Let me know if there are further comments :)

robertpaschedag:
Gentle ping...

